PR #109, opened 2019-01-08
Fix Java Make Clean.

DavidMansolino:
The make clean for the Java controller was broken from the main compilation process because the Java wrapper is deleted first and therefore the clean rule was never defined.


DavidMansolino:
test me

PR #111, opened 2019-01-10
Fix SUMO exporter

fabienrohrer:
SUMO exporter may crash on the recently-added RoadLine nodes, with this error:

```
  File "exporter.py", line 49, in <module>
    road = Road(roadString, roadType)
  File "/Users/fabien/repos/webots/resources/sumo_exporter/road.py", line 63, in __init__
    self.wayPoints = apply_spline_subdivison_to_path(self.wayPoints, splineSubdivision)
  File "/Users/fabien/repos/webots/resources/sumo_exporter/math_utils.py", line 10, in apply_spline_subdivison_to_path
    points.append((points[pointsNumber - 1][0] + (points[pointsNumber - 1][0] - points[pointsNumber - 2][0]),
IndexError: list index out of range
```

It's because the simplistic .wbt parser implemented in the SUMO exporter does not support subnodes.

The solution presented here simply drop this case, which is the only one for now. The ultimate solution would be to have a real .wbt parser, such as the under-development [WEPT](https://github.com/fabienrohrer/wept).


fabienrohrer:
test me

fabienrohrer:
👍

fabienrohrer:
👎 not sure this is a good way to do. There are other cases were it can block ('appearance / pavementAppearance / startRoadsLine')

fabienrohrer:
I finally choose to implement a better wbt node extractor. This should solves all the issues reported here.

fabienrohrer:
test me

fabienrohrer:
👍

DavidMansolino:
👀

PR #112, opened 2019-01-10
Fix default floor creation

fabienrohrer:
The appearance of the default Floor created by the OSM importer is invalid, because the "texture" field has been replaced by the "appearance" field in R2019a.


fabienrohrer:
test me

PR #118, opened 2019-01-11
Transfer all Textures to GPU at Load

### LINKED ISSUE #84: Transfer all Textures to GPU at Load
**Describe the Bug**
Currently all the wren textures are loaded, setup and bind when the world is loaded, but OpenGL actually transfer them to the GPU only the first time they are used. Which can cause slowdowns while the simulation is running (because of culling new objects/textures can enter the scene, see for example https://www.gamedev.net/forums/topic/370069-texture-prefetch-preloading-solved/).

**Steps to Reproduce**
This is easily visible in big worlds when the viewpoint far field is set to non-zero. Moving the viewpoint initially produces jaggy motions and we can see the GPU memory increasing while we move. Then, once we have 'viewed' all the objects, the motion becomes smooth and GPU memory becomes stable.

**Expected behavior**
Performing an initial rendering without culling will force the transfer of all the textures to the GPU: this is currently in test here: cb98047a6c5564df1574f9bf05591c8371e02a26

DavidMansolino:
Fix #84.

### DIFF
--- a/include/wren/scene.h
+++ b/include/wren/scene.h
@@ -42,9 +42,8 @@ void wr_scene_apply_pending_updates(WrScene *scene);
 
 /* The 'materialName' parameter is optional (can be set to NULL for default), and if set, force
 the renderables to use the named material */
-void wr_scene_render(WrScene *scene, const char *material_name, bool culling = true);
-void wr_scene_render_to_viewports(WrScene *scene, int count, WrViewport **viewports, const char *material_name,
-                                  bool culling = true);
+void wr_scene_render(WrScene *scene, const char *material_name, bool culling);
+void wr_scene_render_to_viewports(WrScene *scene, int count, WrViewport **viewports, const char *material_name, bool culling);
 
 void wr_scene_set_ambient_light(const float *ambient_light);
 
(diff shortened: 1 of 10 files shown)


include/wren/scene.h
```diff
@@ -42,8 +42,9 @@ void wr_scene_apply_pending_updates(WrScene *scene);
 
 /* The 'materialName' parameter is optional (can be set to NULL for default), and if set, force
 the renderables to use the named material */
-void wr_scene_render(WrScene *scene, const char *material_name);
-void wr_scene_render_to_viewports(WrScene *scene, int count, WrViewport **viewports, const char *material_name);
+void wr_scene_render(WrScene *scene, const char *material_name, bool culling = true);
+void wr_scene_render_to_viewports(WrScene *scene, int count, WrViewport **viewports, const char *material_name,
+                                  bool culling = true);
```
omichel: It's normally not possible to define a default argument in C.

DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
👍

PR #121, opened 2019-01-15
Fix Emitter-Receiver Matlab API

DavidMansolino:
The Matlab Emitter-Receiver API is incomplete.

https://www.cyberbotics.com/doc/reference/emitter?version=fix-emitter-reciever-matlab-api#emitter-functions


DavidMansolino:
👍

PR #132, opened 2019-01-23
Reduce 3D view resolution on Retina displays

fabienrohrer:
Fix #46

PR #141, opened 2019-01-28
Fixed screenshots, optimized movies on Retina displays

omichel:
Fixed screenshot resolution bug on Retina displays introduced in #132.
Optimized movie creation on Retina displays.

### DIFF
--- a/src/webots/gui/WbWrenWindow.cpp
+++ b/src/webots/gui/WbWrenWindow.cpp
@@ -269,47 +269,44 @@ void WbWrenWindow::resizeWren(int width, int height) {
   emit resized();
 }
 
-void WbWrenWindow::flipImageBuffer(unsigned char *buffer, int width, int height, int channels) {
-  // flip vertically the image (about 2x faster than QImage::mirrored())
-  const int lineSize = width * channels;
-  const int halfHeight = height / 2;
-  unsigned char *srcPtr = buffer;
-  unsigned char *dstPtr = &buffer[(height - 1) * lineSize];
-  for (int y = 0; y < halfHeight; y++) {
-    for (int x = 0; x < lineSize; x++) {
-      const uchar d = dstPtr[x];
-      dstPtr[x] = srcPtr[x];
-      srcPtr[x] = d;
-    }
-    srcPtr += lineSize;
-    dstPtr -= lineSize;
-  }
+void WbWrenWindow::flipAndScaleDownImageBuffer(const unsigned char *source, unsigned char *destination, int sourceWidth,
+                                               int sourceHeight, int channels, int scaleDownFactor) {
+  // flip vertically the image and scale it down (about 3x faster than QImage::mirrored(), QImage::scaled())
+  const int h = sourceHeight / scaleDownFactor;
+  const int sourceLineSize = channels * sourceWidth;
+  const int destinationLineSize = sourceLineSize / scaleDownFactor;
+  for (int y = 0; y < h; y++)
+    for (int x = 0; x < destinationLineSize; x++)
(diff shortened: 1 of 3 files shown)


omichel:
test me

omichel:
👍

fabienrohrer:
The algorithm is now working (right dimension and color), but I have a new clang warning:

```
gui/WbWrenWindow.cpp:278:25: warning: cast from 'const unsigned char *' to 'const uint32_t *' (aka 'const unsigned int *') increases required alignment from 1 to 4 [-Wcast-align]
  const uint32_t *src = (const uint32_t *)source;  // assuming a pixel is coded as four bytes (RGBA)
                        ^~~~~~~~~~~~~~~~~~~~~~~~
gui/WbWrenWindow.cpp:279:19: warning: cast from 'unsigned char *' to 'uint32_t *' (aka 'unsigned int *') increases required alignment from 1 to 4 [-Wcast-align]
  uint32_t *dst = (uint32_t *)destination;
                  ^~~~~~~~~~~~~~~~~~~~~~~
2 warnings generated.
```

fabienrohrer:
test me

omichel:
Thank you for testing. I don't have such a warning with gcc... Any idea on how to fix?
We could also choose to silence it on the mac (with -Wcast-align).

src/webots/gui/WbWrenWindow.cpp
```diff
@@ -269,47 +269,46 @@ void WbWrenWindow::resizeWren(int width, int height) {
...
+                                               int sourceHeight, int scaleDownFactor) {
+  // flip vertically the image and scale it down (about 3x faster than QImage::mirrored(), QImage::scaled())
+  const int h = sourceHeight / scaleDownFactor;
+  const int w = sourceWidth / scaleDownFactor;
+  const int yFactor = scaleDownFactor * sourceWidth;
+  const uint32_t *src = (const uint32_t *)source;  // assuming a pixel is coded as four bytes (RGBA)
+  uint32_t *dst = (uint32_t *)destination;
+  for (int y = 0; y < h; y++)
+    for (int x = 0; x < w; x++)
+      dst[(h - 1 - y) * w + x] = src[y * yFactor + x * scaleDownFactor];
```
DavidMansolino: This is not respecting our coding style ([CS360](https://github.com/omichel/webots/wiki/CPP-Coding-Style#cs360-single-if-while-or-for-statement-must-be-written-without-brackets-the-statement-part-should-be-put-on-a-separate-line-in-case-of-nested-block-keep-brackets)).

fabienrohrer:
The only workaround I found for now is the following:

- The `(unsigned char *) => (int *)` cast shows this warning.
- The `(void *) => (int *)` cast does not show this warning. (We are using this in other part of the code)
- Changing the problematic lines like above fixes the issue:

```
  const uint32_t *src = (const uint32_t *)((void *)source);
  uint32_t *dst = (uint32_t *)((void *)destination);
```

This is not perfect, but I think that it's well commented, it's better than removing `-Wcast-align`.
Is it ok like this?

src/webots/gui/WbWrenWindow.cpp
fabienrohrer: I fix also this in my next commit.

omichel:
Yes, that's fine for me. The comment should mention that the char * pointer is aligned on an int boundary because it corresponds the beginning of the memory chunk of the rendered image.

omichel:
OK, your comment is very good. That's all right for me.

omichel:
👍

fabienrohrer:
Perfect, I'm just relaunching Jenkins tests, and I will approve it.

fabienrohrer:
test me

fabienrohrer:
test me

fabienrohrer:
test me

Issue #127, opened 2019-01-22
Webots crashes on startup


fabienrohrer:
**Describe the Bug**

On some hardware, Webots R2019a crashes on startup.

**Screenshots**

![capture](https://user-images.githubusercontent.com/866788/51541404-05d0bf80-1e59-11e9-9bfd-7c6b0f6e0e92.JPG)

_(from ticket #324)_ 

**System**
 
It seems to affect mainly Windows users, without guarantee for now.
The bug have been reproduced with the following GPU:
  - AMD Radeon(TM) R7 M440 (0x6900)
  - AMD Radeon 6600M and 6700M Series
  - NVIDIA Corporation - GeForce 840M/PCIe/SSE2 - 3.3.0 NVIDIA 376.54
  - NVIDIA Corporation - GeForce 540M - 3.3.0

fabienrohrer:
A user fails the procedure at step 2. He can run "empty.wbt" but not "geometric_primitives.wbt" whatever the preferences of step 3.

DavidMansolino:
It seems the problem comes from HDR background. Non HDR background + PBR + Bloom + GTAO is working fine.

DavidMansolino:
The problem seems to be the backing of the specular irradiance, disabling it fixes the crash.

DavidMansolino:
Reducing the resolution of the specular irradiance cubemap texture from 512 to 128 (and also the number of mimaps from 10 to 8) fixes the problem, I can now run all the worlds of the guided tour.

Since this resolution is not yet linked to the texture quality preference I propose to link it to it, so that it will work in low-quality texture mode.
Then we will need a way to detect these problematic GPUs and switch automatically to the low texture quality mode (the GPU I am currently testing have 1Gb of dedicated memory which is not so low).

DavidMansolino:
I have made a few comparisons to check the impact on the rendering (please make sure to compare them at full resolution and not only the small preview):

| resolution | metalness 1 | metalness 0 |
| ------------- | ---------------- | --------------- |
|     512       | ![512](https://user-images.githubusercontent.com/2461619/51622148-4f430c80-1f36-11e9-84f3-cca97ce81b71.png) | ![512_2](https://user-images.githubusercontent.com/2461619/51622154-51a56680-1f36-11e9-9f80-7c2c302294cb.png) |
[...]

PR #145, opened 2019-01-29
 Improve OpenGL version detection

fabienrohrer:
Add simple OpenGL version check to understand better issues similar to #127.

### DIFF
--- a/src/webots/gui/WbWrenWindow.cpp
+++ b/src/webots/gui/WbWrenWindow.cpp
@@ -289,8 +289,7 @@ void WbWrenWindow::flipAndScaleDownImageBuffer(const unsigned char *source, unsi
   const int w = sourceWidth / scaleDownFactor;
   const int yFactor = scaleDownFactor * sourceWidth;
 
-  // - The `unsigned char *` to `int *` cast is possible assuming that a pixel is coded as four bytes (RGBA) aligned on an `int
-  // *` boundary.
+  // - The `unsigned char *` to `int *` cast is possible assuming that a pixel is coded as four bytes (RGBA) aligned on an `int *` boundary.
   // - A preliminary `unsigned char *` to `void *` cast is required to by-pass "cast-align" clang warnings.
   const uint32_t *src = (const uint32_t *)((void *)source);
   uint32_t *dst = (uint32_t *)((void *)destination);
--- a/src/webots/gui/WbWrenWindow.cpp
+++ b/src/webots/gui/WbWrenWindow.cpp
@@ -69,8 +69,10 @@ WbWrenWindow::WbWrenWindow() :
 
   setSurfaceType(QWindow::OpenGLSurface);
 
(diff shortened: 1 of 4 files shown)


fabienrohrer:
test me

fabienrohrer:
On my VMWare environment, I can bound OpenGL version to 2.1 to simulate this fatal error:

capture d ecran 2019-01-29 a 10 55 52

fabienrohrer:
👍

src/webots/widgets/WbMessageBox.cpp
```diff
@@ -69,6 +69,14 @@ void WbMessageBox::critical(const QString &message, QWidget *parent, const QStri
     WbLog::error(message);
 }
 
+void WbMessageBox::fatal(const QString &message, QWidget *parent, const QString &title) {
+  if (gEnabled)
+    QMessageBox::critical(parent, title, message);
+  else
+    WbLog::error(message);
```
DavidMansolino: Instead of duplicating the code you can call `WbMessageBox::critical`.

src/webots/gui/WbWrenWindow.cpp
```diff
@@ -69,8 +69,10 @@ WbWrenWindow::WbWrenWindow() :
 
   setSurfaceType(QWindow::OpenGLSurface);
 
+  const WbVersion openGLTargetVersion(3, 3);
```
DavidMansolino: That's just an idea but maybe because the target version is imposed by WREN we should encode it on the WREN side.

src/webots/widgets/WbMessageBox.cpp
fabienrohrer: My concern is that WbMessageBox::critical does not quit Webots.
I just remark it's the purpose of WbLog::fatal.
If it's ok I will drop my modifications and use WbLog::fatal instead.

fabienrohrer:
test me

fabienrohrer:
test me

src/webots/gui/WbWrenWindow.cpp
fabienrohrer: It seems that neither WREN nor GLAD are currently linked to a specific OpenGL version. We only check at the WREN initialization that some OpenGL extensions are available. But Webots/Qt set OpenGL 3.3, because they manage the windows, the OpenGL initialization and the shaders. Shaders are the (only?) main reason we need OpenGL 3.3.

It has some sense, and this PR is correct according to this state.

If we would like to change this, it's not related with this PR and would require a new GitHub issue and a stronger refactor.

fabienrohrer:
👍

src/webots/gui/WbWrenWindow.cpp
fabienrohrer: Correction: glad.h is actually linked with an OpenGL version when generating glad.h

PR #154, opened 2019-02-04
Fixed gcc7/cygwin flush of stdout/stderr

### LINKED ISSUE #124: Controller Print not Displayed on Windows
**Describe the Bug**
Some prints in the controller are sometimes not displayed on Windows.

**Steps to Reproduce**
The following controller doesn't print anything:
```
#include <stdio.h>
#include <webots/robot.h>

#define TIME_STEP 16

   wb_robot_init();
     printf("hello world!\n");
   }    
   wb_robot_cleanup();    
   return 0; 
 }
```
Adding a `fflush(stdout);` after the print solves the issue.

**System**
 - Operating System: Windows 10. This is not reproducible on Linux.

omichel:
Fixes #124.
I tested with C and C++ controllers with GCC 7 and GCC 8 and it worked in all cases.

### DIFF
--- a/include/controller/c/webots/robot.h
+++ b/include/controller/c/webots/robot.h
@@ -67,7 +67,9 @@ int wb_robot_init_msvc();  // internally, this function just calls wb_robot_init
 
 int wb_robot_step(int duration);  // milliseconds
 
-#ifdef __CYGWIN__  // in that case, we need to flush explicitly the stdout/stdin streams otherwise they are buffered
+#ifdef __CYGWIN__  // In that case, we need to flush explicitly the stdout/stdin streams otherwise they are buffered
+// We cannot call fflush from the libController as libController is compiled with gcc8 and won't flush the stdout/stderr
+// of a gcc7 (cygwin) compiled binary. Therefore, we need to perform the fflush in a gcc7 compiled code, e.g., in a macro here.
 #define wb_robot_step(d) (fflush(NULL) ? wb_robot_step(d) : wb_robot_step(d))
 #endif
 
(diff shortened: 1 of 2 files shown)


omichel:
👍

DavidMansolino:
👀

include/controller/c/webots/robot.h
```diff
@@ -65,7 +65,12 @@ int wb_robot_init_msvc();  // internally, this function just calls wb_robot_init
 #define wb_robot_init() (setvbuf(stdout, NULL, _IONBF, 0), setvbuf(stderr, NULL, _IONBF, 0), wb_robot_init_msvc())
 #endif
 
-int wb_robot_step(int duration);                                                                // milliseconds
+int wb_robot_step(int duration);  // milliseconds
+
+#ifdef __CYGWIN__  // in that case, we need to flush explicitly the stdout/stdin streams otherwise they are buffered
+#define wb_robot_step(d) (fflush(NULL) ? wb_robot_step(d) : wb_robot_step(d))
```
fabienrohrer: I don't understand the reason to create a macro for this, rather than simply put the fflush statement into the wb_robot_step() function core. This line looks like a hack (difficult to understand and "hidden"). If there are reasons for this, a comment would be good.

omichel:
👍

include/controller/c/webots/robot.h
fabienrohrer: Sorry to insist, but in this case, wouldn't it be cleaner to build libController with gcc7 as we spoke yesterday?

omichel: Good idea. Let me try that...

omichel: For some reason, it doesn't work: the streams are not flushed when the libController and the controller are compiled with GCC7/cygwin. It seems GCC7/cygwin blocks the DLLs from being able to flush, regardless if the DLL was compiled with GCC7/cygwin or GCC8. So, I have no better idea than my original one.

PR #155, opened 2019-02-05
Fix macOS build

fabienrohrer:
`WbSysInfo::isAmdLowEndGpu()` should not be defined on macOS, otherwise it leads to build issues:

```
core/WbSysInfo.cpp:555:17: error: out-of-line definition of 'isAmdLowEndGpu' does not match any declaration in namespace 'WbSysInfo'
bool WbSysInfo::isAmdLowEndGpu(QOpenGLFunctions *gl) {
                ^~~~~~~~~~~~~~
# compiling gui/WbWheelEvent.cpp
core/WbSysInfo.cpp:557:16: error: use of undeclared identifier 'gpuDeviceId'
  quint32 id = gpuDeviceId(gl);
               ^
2 errors generated.
```

### DIFF
--- a/src/webots/core/WbSysInfo.cpp
+++ b/src/webots/core/WbSysInfo.cpp
@@ -550,8 +550,6 @@ int WbSysInfo::intelGPUGeneration(QOpenGLFunctions *gl) {
   return 0;
 }
 
-#endif
-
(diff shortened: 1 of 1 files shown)


fabienrohrer:
test me

fabienrohrer:
👍

PR #169, opened 2019-02-07
Add a script to clean Webots preferences

fabienrohrer:
Cleaning the Webots preferences is a difficult task (especially on macOS and Windows). I think it would deserve a helper script to do so.

### DIFF
--- a/src/bin/clean_webots_preferences.sh
+++ b/src/bin/clean_webots_preferences.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+# Cleanup all Webots preferences.
+
+if [ "$(uname)" == "Darwin" ]; then
+  for n in `find $HOME/Library/Preferences -name com.cyberbotics.* | sed 's:.*/::;s:.plist::'`; do
+    echo Remove "$n";
+    defaults remove $n;
+    rm $HOME/Library/Preferences/$n.plist;
+  done
+else
+  echo "Unsupported OS"
+  exit 1
+fi


fabienrohrer:
It would be better to do it in Python to simply support windows registry modifications like this: https://stackoverflow.com/questions/15128225/python-script-to-read-and-write-a-path-to-registry

fabienrohrer:
👍

scripts/preferences_cleaner/preferences_cleaner.py
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env python3
```
DavidMansolino: This is not yet supported by th etest suite, you should extend this mechanism: https://github.com/cyberbotics/webots/blob/c4ce84200caa0725655d1538076099d2fc036a22/tests/sources/test_license.py#L70

fabienrohrer: Mmmm, I finally changed this to `python`. The version is tested later and described in the README.md. It sounds better to me.

fabienrohrer:
👍

omichel:
This script is apparently not distributed in the installation packages. Is that correct? Therefore I wonder whether we should mention it in the change log.

omichel:
OK, good idea.

fabienrohrer:
👍

fabienrohrer:
I applied your suggestions as-is.

👍

PR #182, opened 2019-02-08
Fix wrong include and shared libray paths

omichel:
This is some clean-up.


omichel:
👍

fabienrohrer:
Just a question; I'm wondering how this could work outside an msys environment, i.e. when running Webots from the icon. Did you test this?

omichel:
Yes, it works as MSYS2 considers that its root folder is `C:\Program Files\Webots\msys64`.

DavidMansolino:
I am currently creating a package to test this (ant other PRs).

projects/samples/howto/controllers/vision/Makefile
```diff
@@ -27,7 +27,7 @@ USE_C_API = true
 ifeq ($(OSTYPE),windows)
   OPENCV_VERSION := $(subst .,,$(shell opencv_version))
   INCLUDE = -I/mingw64/include/opencv4
-  LIBRARIES = -L"$(WEBOTS_HOME)/msys64/mingw64/bin" -L/mingw64/bin -lopencv_core$(OPENCV_VERSION) -lopencv_imgproc$(OPENCV_VERSION)
+  LIBRARIES = -L/mingw64/bin -lopencv_core$(OPENCV_VERSION) -lopencv_imgproc$(OPENCV_VERSION)
```
DavidMansolino: I can't neither run or recompile this controller from the release (both works fine from the development environement):
```
make
make: opencv_version: Command not found
# updating vision.d
make: opencv_version: Command not found
# compiling vision.cpp
vision.cpp:25:10: fatal error: opencv2/core/core.hpp: No such file or directory
 #include <opencv2/core/core.hpp>
          ^~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make: *** [C:/Program Files/Webots/resources/Makefile.include:691: build/release/vision.o] Error 1
```
capture

omichel:
I believe the errors you got are not related with the changes in the Makefile and are not a regression. I just added a comment in the vision.cpp file to explain how to proceed to use this controller. Distributing the OpenCV dependency on Windows is the only way to fix this problem, but that would add an extra 100MB+ to the package, which seems overkill for just one controller example. So, I would propose to keep things like this.

👍

DavidMansolino:
I fixed a few others paths I will create a package to make sure it works fine.

DavidMansolino:
test me

DavidMansolino:
test me

PR #195, opened 2019-02-12
Deny Transform to Differentialwheels

DavidMansolino:
It should not be possible to transform any node into a DifferentialWheels node from Webots.


DavidMansolino:
test me

DavidMansolino:
:+1:

PR #196, opened 2019-02-12
Fix texture downscaling (alternative)

### LINKED ISSUE #171: Improve linear_motor.wbt Node hierarchy
**Describe the Bug**
The node hierarchy of the linear motor simulation should be improved. It does contain useless group node and empty transform node.

Furthermore with the AMD radeon R5 the ruler is to much transparent.
Nvidia GTX 980:
linear_motor

Radeon R5:
linear_motor

fabienrohrer:
Fix #171

### DIFF
--- a/src/webots/nodes/WbImageTexture.cpp
+++ b/src/webots/nodes/WbImageTexture.cpp
@@ -103,50 +103,47 @@ static void blurred(QImage &result, int radius) {
   int rgba[4];
   unsigned char *p;
 
-  int i1 = 0;
-  int i2 = 3;
-
-  for (int col = c1; col <= c2; col++) {
+  for (int col = c1; col <= c2; ++col) {
     p = result.scanLine(r1) + col * 4;
-    for (int i = i1; i <= i2; i++)
+    for (int i = 0; i < 4; ++i)
       rgba[i] = p[i] << 4;
 
     p += bpl;
-    for (int j = r1; j < r2; j++, p += bpl)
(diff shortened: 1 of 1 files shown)


fabienrohrer:
This solution works but is unfortunately very slow.

omichel:
Why is it slower than what is done by Qt? Is it a different algorithm?

fabienrohrer:
Yes, this is a different algorithm but the idea is globally the same.

As expected, in Qt, the image smooth-scale algorithm is split into several optimized parts, the main entry point use 2D transform matrix:
http://code.qt.io/cgit/qt/qtbase.git/tree/src/gui/image/qimage.cpp#n2767
I think the scale is done in this function:
http://code.qt.io/cgit/qt/qtbase.git/tree/src/gui/painting/qimagescale.cpp#n540
I think that the blur (which is actually the part creating alpha issue for sure) is done in this function:
https://code.woboq.org/qt5/qtbase/src/widgets/effects/qpixmapfilter.cpp.html#_Z12qt_blurImageP8QPainterR6QImagedbbi
The blur is actually done using a QPainter which is also a big part of Qt.

I also tried to blur using QPainter and qt_blurImage(), but it suffers of the same alpha bug (showing it comes from this part).

fabienrohrer:
I'm trying another approach about to apply a blur at the same time as the scale down.

fabienrohrer:
Drop for #197

PR #203, opened 2019-02-13
Optimized image scale down performance

omichel:
Avoiding the use of float and conversion from float to int allow us to speed up the image scale down algorithm. Before this PR, it took 127 ms on my PC, with this PR, it takes 69 ms.

### DIFF
--- a/src/webots/maths/WbImage.cpp
+++ b/src/webots/maths/WbImage.cpp
@@ -19,36 +19,33 @@
 
 #define CLAMP(v, low, high) ((v) < (low) ? (low) : ((v) > (high) ? (high) : (v)))
 
-static const float gaussBlur[3][3] = {
-  // 3x3 convolution matrix to blur.
-  {1.0f / 16.0f, 2.0f / 16.0f, 1.0f / 16.0f},
-  {2.0f / 16.0f, 4.0f / 16.0f, 2.0f / 16.0f},
-  {1.0f / 16.0f, 2.0f / 16.0f, 1.0f / 16.0f},
-};
-
 WbImage *WbImage::downscale(int width, int height, int xBlurRadius, int yBlurRadius) {
+  // 3x3 convolution matrix to blur.
+  static const int gaussBlur[3][3] = {
+    {1, 2, 1},
+    {2, 4, 2},
+    {1, 2, 1},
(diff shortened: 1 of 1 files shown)


omichel:
test me

fabienrohrer:
Just to inform you also that there is a huge difference if Webots is built in debug or in release mode. The compiler optimizer do a very good job on this loop.

src/webots/maths/WbImage.cpp
```diff
@@ -19,36 +19,33 @@
...
+        unsigned char acc = 0;
         for (int u = -1; u < 2; ++u) {
-          const float x = CLAMP(widthRatio * i + u * xBlurRadius, 0, mWidth - 1);
+          const int x = CLAMP(mWidth * i / width + u * xBlurRadius, 0, mWidth - 1);
           for (int v = -1; v < 2; ++v) {
-            const float y = CLAMP(heightRatio * j + v * yBlurRadius, 0, mHeight - 1);
+            const int y = CLAMP(mHeight * j / height + v * yBlurRadius, 0, mHeight - 1);
             const int srcIndex = CLAMP((int)(x + y * mWidth) * channels + c, 0, srcSize);
-            pixels[destIndex] += (unsigned char)(gaussBlur[u + 1][v + 1] * mData[srcIndex]);
+            acc += gaussBlur[u + 1][v + 1] * mData[srcIndex] / 16;
```
fabienrohrer:
```suggestion
            acc += gaussBlur[u + 1][v + 1] * mData[srcIndex] >> 4;
```

omichel: It makes no difference as the compiler automatically translates `/ 16` into `>> 4` for integers. I measured the performance of both cases, and cannot observe any difference.

omichel: And I believe `/ 16` is more readable.

fabienrohrer: Thank you for the test. It's indeed quite obvious that the compiler can deal with this.

omichel:
My performance tests were measured in release mode.

omichel:
👍

omichel:
Thank you for the tests.

PR #212, opened 2019-02-13
Fix SimpleBuilding Construction Type.

DavidMansolino:
This texture was renamed because used in another PROTO too.


DavidMansolino:
:+1:

PR #239, opened 2019-02-22
Improve Connector Doc

### LINKED ISSUE #238: Connector Improvement Proposal
**Is your feature request related to a problem? Please describe.**

There are some limitations of the controller that are annoying me (I encountered twice the problem this week and I know @fabienrohrer had it too this week):
  1. It should be possible to insert a Connector node in a non-robot node (especially for `passive` connectors this makes a lot of sense).
  2. It should be possible to have one of the two connectors that is static (e.g. its parent doesn't have any physics), this is useful for example when plugin something in a wall.
  3. The doc is not very clear about the fields if they should be defined in both connector (e.g. should I put the same `distanceTolerance` or `tensileStrength` to both connector or just one?).

**Describe the solution you'd like**
  1. This should be quite obvious to implement, probably just adjusting the insertion rules will do the job.
  2. This would require to connect the dynamic connector to the environment (this is quite obvious to do, but the snapping mechanism might be a bit more tricky to adjust).
  3. We should check in the code and improve the doc

**Describe alternatives you've considered**
  1. Adding a dummy robot without controller is a simple/durty workaround.
  2. The static connector could be a Solid with a very high mass (but it will decrease the simulation stability).
  3. The only alternative is to test the behavior.

DavidMansolino:
Partially fix #238.

https://www.cyberbotics.com/doc/reference/connector?version=improve-connector-doc

### DIFF
--- a/docs/reference/connector.md
+++ b/docs/reference/connector.md
@@ -111,7 +111,7 @@ This can be used to simulate the rupture of the docking mechanism.
 The tensile force corresponds to a force that pulls the two connectors apart (in the negative *z*-axes direction).
 When the tensile force exceeds the tensile strength, the link breaks.
 Note that if both connectors are locked, the effective tensile strength corresponds to the sum of both connectors' `tensileStrength` fields.
-The default value -1 indicates an infinitely strong docking mechanism that does not break no matter how much force is applied.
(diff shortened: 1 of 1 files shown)


DavidMansolino:
test me

DavidMansolino:
:+1:

PR #247, opened 2019-02-26
Hotfix cleanup Nao

omichel:
Fixed PEP8 for nao_demo_python.py controller, removed useless DEF in Nao proto.

👍


DavidMansolino:
:eyes:

PR #254, opened 2019-02-27
Feature THREE.js in webots.js

stefaniapedrazzi:
Use THREE js instead of x3dom.

Use cases:
* HTML5 animation/model
* webots stream
* robotbenchmark

Tasks:
* [x] Create scene:
  * [x] on streaming connect
  * [x] on world reload
* [x] Animation:
  * [x] load x3d file
  * [x] apply transform position changes (rotation and translation)
  * [x] apply texture transform translation changes
  * [x] apply material emissiveColor and diffuseColor changes
* [x] Add object
* [x] Delete objects:
  * [x] clear scene
  * [x] delete single object
* [x] extend x3d.js:
  * [x] load textures from stream (and reuse them)
  * [x] handle USE DEF mechanism
    * [x] basic load: geometries, appearance, material, transform
  * [x] define missing nodes import:
    * [x] Viewpoint
    * [x] Background
    * [x] WorldInfo
    * [x] Geometries:
      * [x] Box
      * [x] Cone
      * [x] Cylinder
      * [x] Sphere
      * [x] Elevationgrid
      * [x] IndexedLineSet
      * [x] PointSet (missing shader)
    * [x] TextureTransform
    * [x] CubeMap
    * [x] Lights:
      * [x] DirectionalLight
      * [x] SpotLight
      * [x] PointLight
* [x] object selection
* [x] follow object
* [x] navigation in the scene
* [x] restore windows (robot window, console, editor, help window)
* [x] Some performance improvements:
  * [x] render scene only when needed ( viewpoint changed, texture loaded, objects added/deleted, pose changed)
  * [x] cache dictionary mapping from Webots id to THREEjs object
  * [x] GPUPicker updates

Other tasks:
* [x] export changes in baseColor and emissiveColor for PBRAppearance (same as for Appearance)


stefaniapedrazzi:
Default sphere texture mapping is correct (same as with x3dom) even if it doesn't match the one in Webots: see #281.

stefaniapedrazzi:
test me

stefaniapedrazzi:
I'm using terser (https://github.com/terser-js/terser) to minify the webots.js library and I added the commands to the main Makefile.

Here is a debug index.html file that can be used instead of the one in `resources/web/streaming_viewer` that use the full library.

```
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml">
  <head>
    <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
    <title>Streaming viewer</title>
    <link rel='stylesheet' href='https://code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css'/>
    <link type="text/css" rel="stylesheet" href='https://www.cyberbotics.com/jquery-ui-layout/1.4.4/layout-default.css' />
    <link type="text/css" rel="stylesheet" href='../wwi/wwi.css' />
    <link type="text/css" rel="stylesheet" href='style.css' />
    <script src='https://code.jquery.com/jquery-3.1.1.min.js' integrity='sha256-hVVnYaiADRTO2PzUGmuLJr8BLUSjGIZsDYGmIJLv2b8=' crossorigin='anonymous'></script>
    <script src='https://code.jquery.com/ui/1.12.1/jquery-ui.min.js' integrity='sha256-VazP97ZCwtekAsvgPBSUwPFKdrwD3unUfSGVYrahUqU=' crossorigin='anonymous'></script>
    <script src='https://www.cyberbotics.com/jquery-ui-layout/1.4.4/jquery.layout.js'></script>
    <script src='https://www.cyberbotics.com/jquery-dialogextend/2.0.4/jquery.dialogextend.min.js'></script>
    <script src='https://cdn.jsdelivr.net/ace/1.2.6/min/ace.js'></script>
    <script src="https://cdnjs.cloudflare.com/ajax/libs/three.js/100/three.js"></script>
    <script src="https://cdn.jsdelivr.net/gh/brianxu/GPUPicker/GPUPicker.js"></script>
    <script src="../wwi/util.js"></script>
    <script src="../wwi/dialog_window.js"></script>
    <script src="../wwi/console.js"></script>
    <script src="../wwi/help_window.js"></script>
    <script src="../wwi/robot_window.js"></script>
    <script src="../wwi/toolbar.js"></script>
    <script src="../wwi/editor.js"></script>
    <script src="../wwi/context_menu.js"></script>
    <script src="../wwi/viewpoint.js"></script>
    <script src="../wwi/selector.js"></script>
    <script src="../wwi/mouse_events.js"></script>
    <script src="../wwi/texture_manager.js"></script>
    <script src="../wwi/resource_manager.js"></script>
    <script src="../wwi/server.js"></script>
    <script src="../wwi/stream.js"></script>
    <script src="../wwi/x3d.js"></script>
    <script src="../wwi/x3d_scene_manager.js"></script>
    <script src="../wwi/animation.js"></script>
    <script src="../wwi/webots.js"></script>
    <script src="setup_viewer.js"></script>
  </head>
  <body>
    <div style="padding-left:10px">
      <h1>Webots streaming viewer</h1>
      <p>
        Connect to:
        <input id="IPInput" type="text" value="localhost" />
        :
        <input id="PortInput" type="text" value="1234" />
        <input id="ConnectButton" type="button" value="Connect" onclick="connect()" />
      </p>
    </div>
    <div id="playerDiv">
    </div>
  </body>
</html>
```

stefaniapedrazzi:
I just realized that the current method to handle DEF/USE nodes only works when loading the scene but not when adding/removing object.
I think that the best way to handle DEF/USE is to do it in Webots and send to webots.js the DEF node id instead of the use name.

And I still have to fix the display/hide of sensors visualization in highway_driving.wbt that currently doesn't work.

But maybe we could start with a first review.

stefaniapedrazzi:
test me

stefaniapedrazzi:
Compilation is failing because 'tenser' to compress webots.js library is not found.
We can discuss on monday if we want to use it and in this case I will install it on the jenkins machine.

stefaniapedrazzi:
:+1:

resources/web/streaming_viewer/index.html
```diff
@@ -5,14 +5,16 @@
...
-    <link type="text/css" rel="stylesheet" href='https://www.cyberbotics.com/wwi/R2019a/wwi.css' />
+    <link type="text/css" rel="stylesheet" href='../wwi/wwi.css' />
     <link type="text/css" rel="stylesheet" href='style.css' />
     <script src='https://code.jquery.com/jquery-3.1.1.min.js' integrity='sha256-hVVnYaiADRTO2PzUGmuLJr8BLUSjGIZsDYGmIJLv2b8=' crossorigin='anonymous'></script>
     <script src='https://code.jquery.com/ui/1.12.1/jquery-ui.min.js' integrity='sha256-VazP97ZCwtekAsvgPBSUwPFKdrwD3unUfSGVYrahUqU=' crossorigin='anonymous'></script>
     <script src='https://www.cyberbotics.com/jquery-ui-layout/1.4.4/jquery.layout.js'></script>
     <script src='https://www.cyberbotics.com/jquery-dialogextend/2.0.4/jquery.dialogextend.min.js'></script>
     <script src='https://cdn.jsdelivr.net/ace/1.2.6/min/ace.js'></script>
-    <script src="https://www.cyberbotics.com/wwi/R2019a/webots.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/three.js/100/three.js"></script>
```
fabienrohrer: Why not adding the GPUPicker dependencies (and even THREEjs?) inside terser?

resources/web/wwi/animation.js
```diff
@@ -0,0 +1,204 @@
...
+    if (!this.playSlider || !this.sliding)
+      return;
+
+    var w = event.target.clientWidth - 66; // size of the borders of the slider
+    var x = event.clientX - event.target.getBoundingClientRect().left - 48; // size of the left border (including play button) of the slider
+    var value = 100 * x / w;
+    if (value < 0)
+      value = 0;
+    else if (value >= 100)
+      value = 99.999;
```
fabienrohrer: Hard coded value like this would deserve a comment.

resources/web/wwi/server.js
```diff
@@ -0,0 +1,96 @@
...
+      webots.User1Id = '';
+    if (typeof webots.User1Name === 'undefined')
+      webots.User1Name = '';
+    if (typeof webots.User1Authentication === 'undefined')
+      webots.User1Authentication = '';
+    if (typeof webots.User2Id === 'undefined')
+      webots.User2Id = '';
+    if (typeof webots.User2Name === 'undefined')
+      webots.User2Name = '';
+    if (typeof webots.CustomData === 'undefined')
```
fabienrohrer: There are currently 2 syntaxes to check against 'undefined':

```
typeof data === 'undefined'
```

or 

```
data === undefined
```

The first syntax is often preferred. My concern is just to be coherent. https://stackoverflow.com/questions/3390396/how-to-check-for-undefined-in-javascript

resources/web/wwi/animation.js
stefaniapedrazzi: This is old code that I simply moved but not modified. I will check if I can understand why this was done.

resources/web/streaming_viewer/index.html
stefaniapedrazzi: We can discuss this during the meeting.
But if we do it for THREEjs, GPUPicker, why not doing it as well for all the other dependencies?

resources/web/wwi/animation.js
```diff
@@ -0,0 +1,204 @@
...
+Animation.prototype = {
+  constructor: Animation,
+
+  init: function(onReady) {
+    var that = this;
+    this.onReady = onReady;
+    var xmlhttp = new XMLHttpRequest();
+    xmlhttp.open('GET', this.url, true);
+    xmlhttp.overrideMimeType('application/json');
+    xmlhttp.onreadystatechange = function() {
```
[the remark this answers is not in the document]
stefaniapedrazzi: Yes I know.
But I'm not using ECS6 because I had the impression that it is not widely supported by all the browser.

stefaniapedrazzi:
For testing I changed the streaming viewer and animation to use the local webots.js (before it was using the published one) and that is why the icons are currently not found. Not sure if we want to continue using the published or local ones. If we decide for the local resources I will fix the icons.

> I'm not sure about the status of the shadows. There are several references in the code, but they are no more displayed on screen. (Shouldn't we completely drop them for now? Mainly in the animation dialog box.)

Shadows are supposed to work and they did until my last modifications. I will check why they are no longer working and restore them.
I think that until we do not implement ambient occlusion it is better to keep the default ones instead of completely disabling them.

resources/web/wwi/animation.js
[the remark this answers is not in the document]
fabienrohrer: I think ECS6 can be used without fear. It is well supported on every up-to-date browser: https://kangax.github.io/compat-table/es6/
I think we should rather think about using Babel in case of compatibility issues.
Using this workflow, the source code will be much cleaner.

resources/web/streaming_viewer/index.html
fabienrohrer: Yes, we should choose between everything or nothing :-)

stefaniapedrazzi:
> C: Just to not forget: the robot component still uses the old viewer. But we could do that later in another PR.

How does the robot component load the webots.js dependency?


I removed the debug changes that were disabling the shadows and now they work.
Otherwise I addressed all the comments but the ones related with babel (ES6 and dependencies).

fabienrohrer:
> How does the robot component load the webots.js dependency?

The doc loads webots.js, interactions with X3Dom are coded inside webots.js:
https://github.com/omichel/webots/blob/master/docs/dependencies.txt#L28
https://github.com/omichel/webots/blob/master/docs/js/viewer.js#L601-L879

We should do this in another PR.

stefaniapedrazzi:
> The doc loads webots.js, interactions with X3Dom are coded inside webots.js:
https://github.com/omichel/webots/blob/master/docs/dependencies.txt#L28
https://github.com/omichel/webots/blob/master/docs/js/viewer.js#L601-L879
We should do this in another PR.

Ok, thank you.
Yes, we should change dependencies.txt and upload the new files only when we have a stable new webots.js (or webots.min.js).

stefaniapedrazzi:
Task: send DEF node id instead of DEF and USE names
* [x] at load
* [ ] on update:
  * [ ] USE turned into DEF
  * [ ] 
* [ ] handle updates
 * [ ] cloned nodes (e.g. Group, Transform, Shape)
 * [ ] nested nodes (e.g. TextureTransform, Material)

fabienrohrer:
@stefaniapedrazzi I have issues with DEF-USE on most of the worlds in the current state.

stefaniapedrazzi:
I will check if I don't have any uncommitted changes because for me everything works fine..

stefaniapedrazzi:
I just committed all my local changes. Now the USE/DEF should work correctly.

fabienrohrer:
Thank you ^^

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
:+1:

omichel:
:eyes:

resources/web/wwi/mouse_events.js
```diff
@@ -0,0 +1,368 @@
...
+      if (this.intersection == null) {
+        var cameraPosition = new THREE.Vector3();
+        this.sceneManager.viewpoint.camera.getWorldPosition(cameraPosition);
+        params.distanceToPickPosition = cameraPosition.length();
+      } else
+        params.distanceToPickPosition = this.intersection.distance;
+      if (params.distanceToPickPosition < 0.001) // 1 mm
+        params.distanceToPickPosition = 0.001;
+      // FIXME this is different from webots. We need to understand why the same formula doesn't work.
+      // THREEjs Camera fov is half of Webots Viewpoint fov.
```
omichel: I believe it is important to understand why we don't use the same formula here, as this may hide a more fundamental problem.

resources/web/wwi/resource_manager.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+function ResourceManager() {
```
omichel: What does this class do? A comment would be welcome. Also the name `ResourceManager` is not very useful. What kind of resources are we talking about? A better name could be `wwi_url.js`.

resources/web/wwi/texture_manager.js
```diff
@@ -0,0 +1,124 @@
+/* global THREE */
+'use strict';
+
+function TextureManager() {
```
omichel: The `Manager` keyword is too vague in my opinion, also the method names are repeating the `texture` keyword, which we should try to avoid. Maybe we can call this class `TextureLoader` instead?

resources/web/wwi/video_manager.js
```diff
@@ -0,0 +1,75 @@
+'use strict';
+
+function VideoManager(parentObject, mouseEvents, stream) {
```
omichel: Same comment here, I would remove the `Manager` keyword and call this class simply `Video` and remove `video` from the method names.

resources/web/wwi/texture_manager.js
```diff
@@ -0,0 +1,124 @@
...
+  if (!TextureManager.instance) {
+    TextureManager.instance = this;
+    this.textures = [];
+    this.loadingTextures = [];
+    this.loadingCubeTextureObjects = [];
+    this.streamingMode = false;
+    this.onTextureLoad = undefined;
```
omichel: There are several remaining `undefined` (missing the single quotes). We should search and replace all of them.

resources/web/wwi/x3d_scene_manager.js
```diff
@@ -0,0 +1,400 @@
+/* global THREE, Selector, TextureManager, Viewpoint */
+/* global convertStringToVec2, convertStringToVec3, convertStringToQuaternion, convertStringTorgb */
+/* global createDefaultGeometry, createDefaultMaterial */
+'use strict';
+
+function X3dSceneManager(domElement) {
```
omichel: I would remove the `Manager` keyword here as well.

resources/web/streaming_viewer/index.html
```diff
@@ -5,14 +5,16 @@
     <title>Streaming viewer</title>
     <link rel='stylesheet' href='https://code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css'/>
     <link type="text/css" rel="stylesheet" href='https://www.cyberbotics.com/jquery-ui-layout/1.4.4/layout-default.css' />
-    <link type="text/css" rel="stylesheet" href='https://www.cyberbotics.com/wwi/R2019a/wwi.css' />
+    <link type="text/css" rel="stylesheet" href='../wwi/wwi.css' />
```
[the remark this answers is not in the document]
stefaniapedrazzi: Yes, I changed this mainly to make the tests easier, otherwise you have to change it manually (for animations every time you export one).

resources/web/wwi/texture_manager.js
stefaniapedrazzi: Here is correct to use `undefined` without quotes.
You use quotes only when checking for the type.

resources/web/wwi/resource_manager.js
stefaniapedrazzi: The idea was to have a class that handles all the references to (local) resources like images, etc. without having to hide these paths some other code.

I would prefer a more generic name than `wwi_url.js`.

resources/web/wwi/mouse_events.js
stefaniapedrazzi: Yes, but note that this was exactly the same with x3dom: we were not using the same formula as in Webots.

stefaniapedrazzi:
test me

resources/web/wwi/resource_manager.js
stefaniapedrazzi: Maybe we could call this class `StandardPath` as in Webots

omichel: To me, it looks more related with URLs rather than paths, isn't it? What about `DefaultUrl`?

resources/web/wwi/mouse_events.js
stefaniapedrazzi: The only difference I found w.r.t. Webots was that in webots.js the partial translation was applied based on the relative mouse move distance (difference from previous position) but in Webots the distance to the initial mouse position when the drag was started.
I changed the behavior in webots.js and in the world I tested I get the same behavior.

stefaniapedrazzi:
test me

stefaniapedrazzi:
I think I addressed all the comments.

:+1:

fabienrohrer:
Globally, this PR is working very well, I use it since a week or so.

There are issues, but their resolution may wait depending on our priorities.

I found a bug which disturbs me: when streaming `tinkerbots_demo.wbt` there are issues with the occlusion map. They seem to be partially exported but without the `ImageTexture.url` field, causing issue in the `TextureLoader._onImageLoaded()` function and rendering issues:

Capture d’écran 2019-04-24 à 16 09 53

I tried to solve it, but didn't found the reason yet. Could you look at this, Stefania?

stefaniapedrazzi:
yes, I will look at the issue with the occlusion map.

stefaniapedrazzi:
> The last thing that worries me about the code is the use of singleton classes.
I believe we should generally avoid singletons, see https://stackoverflow.com/questions/137975/what-is-so-bad-about-singletons for a discussion.

There are only two singleton classes:. DefaultUrl and TextureLoader.
I read the discussion you mentioned.
Both classes have internal fields that have to exists for the whole execution of webots.js and they don't cause any dependency issue.
In this case I think that it is much better and easier to maintain if we access these classes instances directly rather than having to retrieve them from other objects as we already do in other cases.
Of course it is also possible and not very complicated to remove the singleton, but the number of class arguments will increase, we will have to add additional references to nested objects and pay attention that they are initialized at correct place.

stefaniapedrazzi:
> I found a bug which disturbs me: when streaming tinkerbots_demo.wbt there are issues with the occlusion map. They seem to be partially exported but without the ImageTexture.url field, causing issue in the TextureLoader._onImageLoaded() function and rendering issues:

Note that for the Tinkerbots PBRAppearance in Webots the `occlusionMapUrl` is not set and thus the `ImageTexture.url` field of the occlusion ImageTexture is empty.
So the exported occlusion map "ImageTexture" is correct.

I didn't change the export behavior and I'm not sure it is good to export the ImageTexture if no url is set. This will for sure make the url field value updates easier. And on the other hand, these additional "useless" export could be avoided by removing empty ImageTexture nodes from the world.

I will now check why the appearance in THREE.js is not the expected one.

omichel:
> There are only two singleton classes:. DefaultUrl and TextureLoader.
> I read the discussion you mentioned.
> Both classes have internal fields that have to exists for the whole execution of webots.js and they don't cause any dependency issue.
> In this case I think that it is much better and easier to maintain if we access these classes instances directly rather than having to retrieve them from other objects as we already do in other cases.
> Of course it is also possible and not very complicated to remove the singleton, but the number of class arguments will increase, we will have to add additional references to nested objects and pay attention that they are initialized at correct place.

I actually don't agree on all the arguments of the discussion thread I mentioned.
I don't like singletons because I believe they shouldn't be a class, but rather utility functions.
In Webots, we replaced some singleton classes with C++ namespaces which is I believe the right way to get rid of singletons.
In Javascript, I am not sure of what would be the best way to achieve this, but in any case, I believe these singletons should not be objects that can be instantiated and deleted, because it makes no sense. Rather, I believe that turning them into utility functions, possibly with the equivalent of a namespace, that will initialize persistent internal values if needed, is the way to go.

stefaniapedrazzi:
> Rather, I believe that turning them into utility functions, possibly with the equivalent of a namespace, that will initialize persistent internal values if needed, is the way to go.

If you turn them into namespace, then you will have to add global variable.
This is would mean to just add one global variable for the DefaultUrl class and it would not be so bad.
But in the TextureLoader class we have many internal fields that doesn't have to be global variables and it is very important to have an instance of it.

So if you really prefer it I can change DefaultUrl.
But I don't agree that turning TextureLoader into a namespace or integratin it into another class would be a better solution.

omichel:
For the record, we just had a voice discussion about singleton and agreed to turn the DefaultUrl into a function and add helper functions to hide the TextureLoader singleton from the API, so that the user of TextureLoader don't have to care about the singleton life cycle and don't consider it as an object any more.

stefaniapedrazzi:
test me

stefaniapedrazzi:
I'm still checking why ambient occlusion maps are not working.
If I disable it seems to work fine.

stefaniapedrazzi:
But I get the exact same appearance setting up a simple box with the Tinkerbots Base appearance in the online [three.js/editor](https://threejs.org/editor/).
So it seems the aoMap effect is different from the one in wren and it has to be adjusted to achieve the same appearance like in Webots.

stefaniapedrazzi:
test me

stefaniapedrazzi:
:+1:

fabienrohrer:
@stefaniapedrazzi Could you upload the required files on the Cyberbotics FTP if it's time to do it? https://www.cyberbotics.com/wwi/R2019b/...

stefaniapedrazzi:
I just uploaded the R2019b files.

PR #260, opened 2019-02-28
Fix tire creaseAngle

fabienrohrer:
Edges should be invisible on a tire. The current computation is complex and do not take into consideration the tire curvature.

| develop | this branch |
| --- | --- |
| wheel before | wheel after |
| session_1_wheel_before | session_1_wheel_after |


fabienrohrer:
test me

fabienrohrer:
👍

PR #241, opened 2019-02-25
Fix default quit button on webots.js

stefaniapedrazzi:
The quit button was not visible by default.
I also fixed other ESLint errors.

PR #261, opened 2019-02-28
webots.js: remove initialization of showRevert and showQuit in wrong method

stefaniapedrazzi:
Given that `webots.showRevert` and `show.Quit` properties are not set to the simulation server, they doesn't have to be initialized in `webots.Server` constructor.

In any case this initialization was fixed in #241 and it is no longer needed.


stefaniapedrazzi:
test me

stefaniapedrazzi:
:+1:

PR #272, opened 2019-03-08
Fix E-puck Light Sensors Calibration.

DavidMansolino:
Since the introduction of PBR we did set the ambient intensity of the default TexturedBackground to 1.0, so that non-PBR objects are ambient illuminated in a similar way than the PBR ones are by the background.
Since the introduction of HDR the intensity range of light has increased.

These two reasons imply that the calibration of the e-puck light sensors had to be updated.

I checked all the e-puck worlds and now the behavior of the light sensors are correct, the sensors facing the light return 0, the ones on the opposite side return ~4095 and the ones facing not exactly the light return an intermediate value.


DavidMansolino:
test me

DavidMansolino:
:+1:

PR #279, opened 2019-03-13
Improve Sumo Interface Error Message

DavidMansolino:
We had recently several questions about where should exactly be the SUMO network files.


DavidMansolino:
test me

DavidMansolino:
:+1:

PR #282, opened 2019-03-14
Fixed minimum mass of WoodenBox which was too high (for balsa for example)

omichel:
Ditto.

### DIFF
--- a/projects/objects/factory/containers/protos/WoodenBox.proto
+++ b/projects/objects/factory/containers/protos/WoodenBox.proto
@@ -45,11 +45,13 @@ Solid {
       size.z = min_size
     end
 
-    -- we want the mass for a 1000cm3 box be comprised between 0.1 and 10 kg inclusive
-    -- and the mass for a 27m3 box be comprised between 90 and 9000 kg, hence the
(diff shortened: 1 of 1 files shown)


omichel:
test me

omichel:
👍

projects/objects/factory/containers/protos/WoodenBox.proto
```diff
@@ -45,11 +45,13 @@ Solid {
       size.z = min_size
     end
 
-    -- we want the mass for a 1000cm3 box be comprised between 0.1 and 10 kg inclusive
-    -- and the mass for a 27m3 box be comprised between 90 and 9000 kg, hence the
+    -- we want the mass for a 1000cm3 box be comprised between 0.01 and 10 kg inclusive
+    -- and the mass for a 27m3 box be comprised between 9 and 9000 kg, hence the
```
DavidMansolino: The density of balsa if `170kg/m³` the previous range was already correct:
   - 1000cm³ => 0.001m³ => 0.17kg
   - 27m3 => 4590kg
Fruthermore I don't understand why we use this two ranges as example because they doesn't match...

DavidMansolino: Edit: the box may be empty.

PR #286, opened 2019-03-14
Intermodal containers

fabienrohrer:
- [x] Models
- [x] Icons
- [x] Doc
- [x] Log

containers


fabienrohrer:
test me

fabienrohrer:
👍

projects/objects/freight/protos/IntermodalOfficeContainer.proto
```diff
@@ -0,0 +1,120 @@
+#VRML_SIM R2019a utf8
+# license: Copyright Cyberbotics Ltd. Licensed for use only with Webots.
+# license url: https://cyberbotics.com/webots_assets_license
+# An intermodal container is a large standardized shipping container, designed and built for intermodal freight transport, meaning these containers can be used across different modes of transport – from ship to rail to truck – without unloading and reloading their cargo.
+# This container contains an office.
```
omichel: I would avoid copy/pasting descriptions (this is painful to read in the user guide).
```suggestion
# An intermodal container with an office room.
```

projects/objects/freight/protos/IntermodalOfficeContainer.proto
```diff
@@ -0,0 +1,120 @@
...
+
+PROTO IntermodalOfficeContainer [
```
omichel: I am wondering if this is the best name, as google image provides better results for simply "office container" rather than "intermodal office container", so I believe we could simplify the name to just `OfficeContainer`.

fabienrohrer:
I took into consideration your comments except the rename of `IntermodalOfficeContainer` which appears to be less precise to me. The "intermodal" keyword is there to define the dimensions of the container. Let me know if you would like to change it and I will do it.

omichel:
I checked google images for "intermodal office container" and "office container". It's true that the intermodal version is more precise on the container size, so that's probably a better choice.

PR #306, opened 2019-03-26
Fix Track X3D export.

### LINKED ISSUE #305: Crash when Exporting Robot with Track to X3D
**Describe the Bug**
If Webots is started with the `--enable-x3d-meta-file-export` option it will crash when exporting to X3D a robot with track nodes.

DavidMansolino:
Fix #305

### DIFF
--- a/src/webots/nodes/utils/WbWorld.cpp
+++ b/src/webots/nodes/utils/WbWorld.cpp
@@ -421,9 +421,7 @@ void WbWorld::createX3DMetaFile(const QString &filename) const {
       const WbBaseNode *deviceBaseNode = dynamic_cast<const WbBaseNode *>(device);
       const WbJointDevice *jointDevice = dynamic_cast<const WbJointDevice *>(device);
       const WbMotor *motor = dynamic_cast<const WbMotor *>(jointDevice);
-      if (motor && !jointDevice->joint())  // special Track motor case
-        deviceObject.insert("type", "Track");
-      else if (deviceBaseNode)
+      if (deviceBaseNode)
         deviceObject.insert("type", deviceBaseNode->nodeModelName());
 
       if (jointDevice && jointDevice->joint()) {  // case: joint devices.
@@ -447,6 +445,8 @@ void WbWorld::createX3DMetaFile(const QString &filename) const {
             WbVector3 v = m.translation();
             if (!v.almostEquals(WbVector3()))
               deviceObject.insert("transformOffset", v.toString(WbPrecision::FLOAT_MAX));
+            if (motor)
+              deviceObject.insert("track", "true");
             break;
           } else {
             const WbAbstractTransform *transform = dynamic_cast<const WbAbstractTransform *>(parent);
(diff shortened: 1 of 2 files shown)


stefaniapedrazzi:
:eyes:

stefaniapedrazzi:
Oops.. there was no thumb up..

DavidMansolino:
Indeed, it's not ready (sorry I forget to put 'draft' PR and once it is created you can't change anymore :-/ )

DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
:+1:

DavidMansolino:
(result visible here: https://www.cyberbotics.com/doc/guide/telemax?version=feature-telemax-robot)

stefaniapedrazzi:
:eyes:

DavidMansolino:
> Maybe it would be better to list the Track motors in the LinearMotor section with all the other motors given that it is a normal LinearMotor node.

This is what I did initially but the problem is that any node in this section will have the slider to move it, and since our viewer doesn't support moving track node it looks a bit strange to have this useless slider. I will improve it to not show the slider in this case (even so the goal of this PR was initialyl to hotfix the crash and not extend our viewer).

stefaniapedrazzi:
Alternatively if it doesn't look consistent to not have sliders you could maybe just change the section name from "Track" to "Track motors".

DavidMansolino:
> Alternatively if it doesn't look consistent to not have sliders you could maybe just change the section name from "Track" to "Track motors".

That's the first things I tried but our viewer is probably checking for the 'motor' string as it was displaying sliders for this new category too :rofl:

DavidMansolino:
Now they are in the LinearMotor section but without the slider:
Capture du 2019-03-26 16-25-44

DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
:+1:

stefaniapedrazzi:
:eyes:

src/webots/nodes/utils/WbWorld.cpp
```diff
@@ -445,6 +445,8 @@ void WbWorld::createX3DMetaFile(const QString &filename) const {
             WbVector3 v = m.translation();
             if (!v.almostEquals(WbVector3()))
               deviceObject.insert("transformOffset", v.toString(WbPrecision::FLOAT_MAX));
+            if (motor)
```
stefaniapedrazzi: It doesn't seem correct to assume that if a device has a motor but it is not a joint, then it is a `Track` device.
A `Propeller` has a motor too, so with the current code the `Propeller` motor is tagged with the `track` flag.
I think that here an additional type check is needed to set the track flag correctly just for Track motors.

DavidMansolino: Very good point, I did forget the propeller case.

DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
:+1:

PR #308, opened 2019-03-26
Fix disabled Fog when visibilityRange is 0

### LINKED ISSUE #307: Fog.visibilityRange 0 doesn't disable the fog as stated in the documentation
If the `Fog.visibilityRange` is 0.0, all the objects (but not the textured background) are completed blended by fog.

But the documentation states that value 0.0 should disable the fog node:
https://github.com/omichel/webots/blob/master/docs/reference/fog.md#L16

stefaniapedrazzi:
Fix #307: as stated in the docs if the `Fog.visibilityRange` is 0, then the Fog node is disabled.

### DIFF
--- a/src/webots/nodes/WbFog.cpp
+++ b/src/webots/nodes/WbFog.cpp
@@ -142,13 +142,15 @@ void WbFog::updateVisibilityRange() {
 void WbFog::applyChangesToWren() {
   assert(isFirstInstance());
 
+  WrSceneFogType fogType = mWrenFogType;
   float density = 0.0f;
   if (mVisibilityRange->value() > 0.0)
     density = 1.0 / mVisibilityRange->value();
-
+  else
+    fogType = WR_SCENE_FOG_TYPE_NONE;
   const float color[] = {static_cast<float>(mColor->red()), static_cast<float>(mColor->green()),
                          static_cast<float>(mColor->blue())};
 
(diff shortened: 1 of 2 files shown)


stefaniapedrazzi:
test me

stefaniapedrazzi:
:+1:

src/webots/nodes/WbFog.cpp
```diff
@@ -142,13 +142,15 @@ void WbFog::updateVisibilityRange() {
 void WbFog::applyChangesToWren() {
   assert(isFirstInstance());
 
+  WrSceneFogType fogType = mWrenFogType;
   float density = 0.0f;
   if (mVisibilityRange->value() > 0.0)
     density = 1.0 / mVisibilityRange->value();
-
+  else
+    fogType = WR_SCENE_FOG_TYPE_NONE;
```
DavidMansolino: This makes perfectly sense but is not 100% coherent with the doc, as in the doc we say it should be 0.0 and not <= 0.0 (we should probably improve the doc).

stefaniapedrazzi: Negative values are not supported and if the visibility value is negative it is automatically set to 0. So it should not be possible that `mVisibilityRange` is less than 0.0.
https://github.com/omichel/webots/blob/75eca6004007b2d3b865e6a0aafffc9d6a2aba18/src/webots/nodes/WbFog.cpp#L135

In the doc it is already specified thant visibilityRange is in the range from 0 to inf, so it doesn't seem necessary to change it.

PR #310, opened 2019-03-27
Fix Light Doc.

DavidMansolino:
ditto

### DIFF
--- a/docs/reference/light.md
+++ b/docs/reference/light.md
@@ -2,9 +2,9 @@
 
 ```
 Light {
-  SFFloat ambientIntensity 0       # [0, 1]
+  SFFloat ambientIntensity 0       # [0, inf)
   SFColor color            1 1 1   # any color
-  SFFloat intensity        1       # [0, 1]
+  SFFloat intensity        1       # [0, inf)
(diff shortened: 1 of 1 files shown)


DavidMansolino:
This was not updated when introducing HDR.

DavidMansolino:
:+1:

stefaniapedrazzi:
The documentation of the `ambientIntensity` is outdated. I'm just writing an issue about this.

DavidMansolino:
Yes indeed, it does influence non-PBR object only.

stefaniapedrazzi:
> Yes indeed, it does influence non-PBR object only.

Yes, but even for non-PBR objects it doesn't work as documented.

stefaniapedrazzi:
By the way it seems that `ambientIntesity` cannot be greater than 1:

```
WARNING: PointLight: Invalid 'ambientIntensity' changed to 1. The value should be in range [0, 1].
```

DavidMansolino:
> By the way it seems that ambientIntesity cannot be greater than 1:

It should probably... but this should not be changed on this branch, I will keep therefore keep ambientIntesity bounded on this branch.

DavidMansolino:
:+1:

PR #312, opened 2019-03-27
Add Support for Normal in IndexedFaceSet

DavidMansolino:
This is a trial of adding normal support for the indexedFaceSet.
Remainings task:
  - [x] Normal per face.
  - [x] Triangle mesh cash.
  - [x] Doc.
  - [x] Tests.


DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
I have checked with the normals from the universal robot URDF model and it is working as expected. Unfortunately, the result is better with the crease angle than with the normal from the URDF (which are probably not set manually), but at least it proves that the normal in Webots are working as expected.

DavidMansolino:
:+1:

DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
:+1:

omichel:
:eyes:

DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
@omichel I definitely can't reproduce the problem that you have with the test suite that is not failing when expected. I tested, on my Linux environment, Windows environment and on Jenkins.
In all the cases if I remove the normals or I move the shape the test fails, the only minor exception is if I don't move the shape enough (because since it reflects the upper background as long as the camera doesn't see the border of the shape the image remains the same).

omichel:
The tests are now failing when they should on Windows. Thank you for investigating this.

PR #313, opened 2019-03-27
Fix GPU memory reading on AMD graphics cards

omichel:
As reported by a user the GPU memory reading on AMD graphics cards was broken, probably due to an update of the AMD driver. This PR fixes the problem.


omichel:
test me

omichel:
Messed-up with bad merge, restarting from a clean branch.

PR #316, opened 2019-03-28
Fixed the display of pi and added explanation about formula

### LINKED ISSUE #309: HTML PI encoding not interpreted in code block
In https://www.cyberbotics.com/doc/reference/pointlight the PI character is still displayed as `&pi;`.

Probably it is correct that in the code block everything is written as is and it is not interpreted, but in this case it would be better to just show `pi` instead of `&pi;`.

omichel:
Fix #309.
Preview: https://www.cyberbotics.com/doc/reference/pointlight?version=fix-pi-display


omichel:
👍

PR #336, opened 2019-04-04
Cleanup threadingPolicy.

### LINKED ISSUE #268: General/threadingPolicy Preference Usefulness
**Describe the Bug**
The ``General/threadingPolicy`` preference is not documented anywhere and it is not possible to change it from the user interface.
Does someone know what it is/was used for? Is it still used (in this case it should be documented, at least on the wiki if used for debugging) or can we remove it?

https://github.com/omichel/webots/blob/0aee13196d44be0aea6232113224f4b8255efb9e/src/webots/core/WbPreferences.cpp#L67
https://github.com/omichel/webots/blob/0aee13196d44be0aea6232113224f4b8255efb9e/src/webots/engine/WbSimulationWorld.cpp#L323

DavidMansolino:
Fix #268.


DavidMansolino:
test me

DavidMansolino:
:+1:

PR #345, opened 2019-04-08
 Fix controller builds using a custom Java Makefile.

fabienrohrer:
When compiling a Java controller in the GUI having a custom Makefile, `make -jX` is called. The first rule of the Makefile is called in this default case. Currently, the first rule is `clean`; this mechanism is broken.


fabienrohrer:
test me

fabienrohrer:
👍

PR #356, opened 2019-04-09
Convert textures from PNG to JPEG to save disk space and network bandwith

omichel:
Textures images shouldn't be stored as PNG unless they need an alpha channel or the equivalent JPEG file is smaller than the PNG (rare). The JPEG images created in this PR are based on a 95% JPEG quality, which is very high and the human eye can hardly see the difference from the original PNG.


omichel:
test me

omichel:
👍

fabienrohrer:
According to our conventions, this PR should target the 'revision' branch (this is not an urgent doc modification).

fabienrohrer:
(otherwise, everything seems to work well)

omichel:
> According to our conventions, this PR should target the 'revision' branch (this is not an urgent doc modification).

I am not sure where it is specified, it should be on [this page](https://github.com/omichel/webots/wiki/Git-workflow), but nothing is specified about that.

This PR is not a new feature, neither a bug fix, it may be considered as a cleanup, so for me, it is unclear whether it should go to the master branch or the revision branch.

fabienrohrer:
We discussed about this something like 2 weeks ago and we agreed. Don't you remember?
We can come back on this decision if you would like.

fabienrohrer:
It is a matter of what is the status of the master branch for the users. If we don't respect this, there is no need of a revision branch.

omichel:
Whatever we decide, it should be documented on the wiki.
Let's re-discuss this on Monday.

DavidMansolino:
I agree with Fabien, we already discussed about this (maybe more 1month ago than 2 weeks) and for me it was clear, except doc fix everything should go to revision (we should at any time be able te compile from master a release package that doesn't need to be tested). But you are right Olivier, this should be clearly documented.

omichel:
Replaced by #359 targeting the revision branch.

PR #357, opened 2019-04-10
Fix check for Hinge2Joint ancestor when inserting a Muscle

stefaniapedrazzi:
Muscle are not available for Hinge2Joint, but the check was mistakenly checking the whole ancestor hierarchy instead of stopping at the first joint instance.


stefaniapedrazzi:
test me

stefaniapedrazzi:
:+1:

DavidMansolino:
:eyes:

PR #366, opened 2019-04-15
Improve default SpotLight.beamWidth value

stefaniapedrazzi:
Currently the default values match the VRML ones but it is not a valid setup in Webots given that there is the constraint that `beamWidth` has to be less or equal to `cutOffAngle`.


stefaniapedrazzi:
test me

stefaniapedrazzi:
Wrong target branch.

PR #367, opened 2019-04-15
Improve default SpotLight.beamWidth default value

stefaniapedrazzi:
Currently the default values match the VRML ones but it is not a valid setup in Webots given that there is the constraint that `beamWidth` has to be less or equal to `cutOffAngle`.

### DIFF
--- a/change_logs/ChangeLog.html
+++ b/change_logs/ChangeLog.html
@@ -41,6 +41,7 @@ Released on xx xxth, 2019
 <li>Connector nodes can now be inserted in Solid nodes.</li>
 <li>Connector nodes can now be part of the static environment.</li>
 <li>Improved the infra-red DistanceSensor model by taking into account the roughness and occlusion of the surface.</li>
+<li>Improved the default SpotLight.beamWidth value.</li>
 </ul>
 <li><b>Bug fix</b></li>
 <ul>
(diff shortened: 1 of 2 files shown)


stefaniapedrazzi:
test me

stefaniapedrazzi:
:+1:

DavidMansolino:
:eyes:

stefaniapedrazzi:
> I am wondering if we should not use a 'nicer' value (e.g. something which makes the light 'smooth' e.g. 0.7) to help the user designing nice simulations?

Yes, I also though about it.
But then I was hesitating in choosing an arbitrary value so I simply chose the value that is automatically set in the current version so that the behaviour doesn't change.
0.7 is a good value.

stefaniapedrazzi:
test me

omichel:
This contradicts VRML'97 default value: http://www.lighthouse3d.com/vrml/tutorial/index.shtml?slight
We should therefore add some code to import/export to VRML'97 and handle correctly the default VRML97 value.

stefaniapedrazzi:
> We should therefore add some code to import/export to VRML'97 and handle correctly the default VRML97 value.

Just note that we never handled correctly the default VRML 97 value.
The version with beamWidth 0.785 was working exactly as before.

stefaniapedrazzi:
Adding code to correctly export VRML'97 is very quick, but the import code is more complicated.
So at the end I think that it is better to keep the default VRML value even if it is not supported in Webots and resetting the beamWidth to default value will throw an error as this is a very minor issue.

omichel:
A good compromise might be to set a good default value when we create a new SpotLight node (we did something similar for `Box.size`).

stefaniapedrazzi:
Yes, this is easy done.
I can commit it if you want.

But it doesn't solve the original issue that is that the default value (the one set by reset) is not supported by Webots.

PR #368, opened 2019-04-15
Fix Windows Packaging on Develop

DavidMansolino:
We do not distribute the CPP version of the ROS controller on Windows anymore.


DavidMansolino:
:+1:

PR #369, opened 2019-04-16
Babel, IE and JS dependencies

fabienrohrer:
- [x] Use [Babel](https://babeljs.io/) to be able to support ES6 feature on IE.

ie and babel


fabienrohrer:
JS Minifier tool comparison in our example:

| Tool | File size (in comparison to the original file) |
| --- | --- |
| [reference] |  162Ko |
| terser |  120.6Ko (-25%) |
| babel-minify | 96.4Ko (-40%) |

`babel-minify` gives slightly better results.

fabienrohrer:
Babel seems to do perfectly the job.
Without it, the current state of this project doesn't work on IE11 (for example because of the use of default parameters, etc.). With it, it works directly.
I think we should go on in this direction, and use widely ES6.

fabienrohrer:
I didn't measured any performance lost using the `webots.min.js` produced by Babel. It is certainly safe and for sure simpler to use it everywhere as-is.

fabienrohrer:
I only have issues with the EC6 Array.includes() function on IE11. Babel propose to install the `polyfill` module to support this, but I have issues with its installation. As we are using this only twice, I think we could avoid this EC6 feature.

fabienrohrer:
I think this PR gives a good basis to go on with this project.
I didn't tried to merge JS dependencies into a single one yet.
Once merged, we could adopt EC6 features (like the classes :-)) in other PR (probably one PR per feature).

fabienrohrer:
👍

resources/web/wwi/Makefile
```diff
@@ -34,9 +34,34 @@ JS_SOURCES = \
...
+NPM_EXISTS = $(shell which npm 2> /dev/null)
+
+ifeq ($(NPM_EXISTS),)
+release debug profile update:
+	@echo "# Please install nodejs to build webots.min.js..."
+	@echo "- Linux: sudo apt install npm"
+	@echo "- macOS: brew install node"
+	@echo "- Windows: pacman -S mingw-w64-x86_64-nodejs"
+else
+release debug profile update: install $(JS_SOURCES)
```
omichel: This Makefile when invoked twice in a row seems to repeat useless operations and is very verbose (tested on Windows):
```
$ make
npm install babel-minify @babel/core @babel/cli @babel/preset-env
npm WARN saveError ENOENT: no such file or directory, open 'D:\msys64\home\Olivier\webots\resources\web\wwi\package.json'
npm WARN enoent ENOENT: no such file or directory, open 'D:\msys64\home\Olivier\webots\resources\web\wwi\package.json'
npm WARN wwi No description
npm WARN wwi No repository field.
npm WARN wwi No README data
npm WARN wwi No license field.
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.7 (node_modules\fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.2.7: wanted {"os":"darwin","arch":"any"} (current: {"os":"win32","arch":"x64"})

+ babel-minify@0.5.0
+ @babel/cli@7.4.3
+ @babel/core@7.4.3
+ @babel/preset-env@7.4.3
updated 4 packages in 2.673s
# Compressing webots.min.js
```

resources/web/wwi/Makefile
```diff
@@ -34,9 +34,31 @@ JS_SOURCES = \
...
+  babel-minify \
+  @babel/core \
+  @babel/cli \
+  @babel/preset-env
+
+NPM_EXISTS = $(shell which npm 2> /dev/null)
+
+ifeq ($(NPM_EXISTS),)
+release debug profile update:
+	@echo "# Please install nodejs to build 'webots.min.js' using the instructions on the GitHub wiki."
```
fabienrohrer: > We should update the wiki pages about Webots installation regarding the optional packages (node.js) and/or update the scripts (e.g., for example on Windows, msys64_installer.sh should contain mingw-w64-x86_64-nodejs as an optional package.

You're right. I just simplified instructions, added this dependency in the windows installation script, and added the following Wiki sections (difficult to do better because we don't have branches on the wiki):

- https://github.com/omichel/webots/wiki/Linux-Optional-Dependencies#nodejs
- https://github.com/omichel/webots/wiki/macOS-Optional-Dependencies#nodejs

fabienrohrer:
> This Makefile when invoked twice in a row seems to repeat useless operations and is very verbose (tested on Windows)

I think I also solved this:
I use now a standard `npm` "package.json" to define the required dependencies (this allow to have a better control on them, like to define which version we would like,  and on future Babel presets if any, and to remove the warnings mentioned above).
The remaining stdout is also removed (by the "silent" argument, and by piping stdout to `/dev/null`).

I also solved the `Array.includes` issue I have yesterday (I understood how to include Babel's polyfill correctly.

fabienrohrer:
👍

resources/web/wwi/Makefile
```diff
@@ -32,11 +32,30 @@ JS_SOURCES = \
...
+else
+release debug profile update: install $(JS_SOURCES)
+	@echo "# compressing webots.min.js"
+	@node_modules/.bin/babel $(JS_SOURCES) -o webots.max.js --presets "@babel/preset-env"
+	@node_modules/.bin/minify webots.max.js -o webots.min.js
+	@rm webots.max.js
+endif
+
+install:
```
omichel: I believe one rule is missing here, as the install takes place every time I type make (and takes a few seconds to complete on Windows), which shouldn't be the case.

fabienrohrer: I think I found a solution to this issue using a temporary file (named install.token) which determines when the build occured.

As a proof of success, please examine the following command serie:

```shell
$ make uninstall && echo "Make 1" && make && echo "Make 2" && make && echo "Make 3" && touch package.json && make
# uninstalling npm dependencies
Make 1
# uninstalling npm dependencies
# installing npm dependencies
# compressing webots.min.js
Make 2
# compressing webots.min.js
Make 3
# uninstalling npm dependencies
# installing npm dependencies
# compressing webots.min.js
```

fabienrohrer:
👍

resources/web/wwi/Makefile
omichel: Make 2 should not recompile `webots.min.js`. I am fixing this and will commit something soon.

omichel:
I am now fully satisfied with the behavior of this Makefile. May I let you check it?

fabienrohrer:
It's working as expected, better than before, thank you.

@stefaniapedrazzi Could you also check this PR? I think it's an important stage we should all agree.

stefaniapedrazzi:
test me

PR #375, opened 2019-04-18
THREEjs: Support environment maps

fabienrohrer:
- [x] Enable environment map
- [x] Fix cubemap load (the first load doesn't work)
- [x] Fix cube map orientation

### Comparison with vs without environment map

with:without environment map

### Comparison with Webots

pbr a
pbr b

### DIFF
--- a/resources/web/wwi/x3d_scene.js
+++ b/resources/web/wwi/x3d_scene.js
@@ -350,10 +350,10 @@ X3dScene.prototype = {
 
   _setupEnvironmentMap: function() {
     var backgroundMap;
-    if (typeof this.scene.background !== 'undefined' && this.scene.background.constructor.name === 'CubeTexture')
+    if (typeof this.scene.background !== 'undefined' && this.scene.background.isCubeTexture)
       backgroundMap = this.scene.background;
     this.scene.traverse(function(child) {
-      if (child instanceof THREE.Mesh && child.material && child.material.constructor.name === 'MeshStandardMaterial') {
+      if (child.isMesh && child.material && child.material.isMeshStandardMaterial) {
         var material = child.material;
         material.envMap = backgroundMap;
       }
--- a/resources/web/wwi/x3d_scene.js
+++ b/resources/web/wwi/x3d_scene.js
@@ -84,6 +84,7 @@ X3dScene.prototype = {
     this.objectsIdCache = {};
     this.useNodeCache = {};
     this.root = undefined;
+    this.scene.background = undefined;
     this.onSceneUpdate();
     this.render();
   },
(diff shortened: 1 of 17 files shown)


fabienrohrer:
Supporting well the background and environment map orientation is not straight forward, because there is a mismatch by definition between the Webots and ThreeJS orientations. ThreeJS doesn't provide the API to rotate the background (it's hard coded in the shader system),  and the Webots one has been based on VRML and already released.
The solution proposed here is to rotate the top/bottom images by 180° and invert the left/right, front/back textures. It's working well.

fabienrohrer:
👍

omichel:
Flipping the images is an expensive operation. Can't we avoid that and change the texture coordinates instead to achieve the same result?

fabienrohrer:
Unfortunately no, there are no texture coordinates here. For example, the reflections coordinates are dealt inside a generated shader. I'm less sure about the skybox, but I didn't found the background box yet.

I don't see better solutions without patching ThreeJS to support cube map flipping, or changing the background and reflection orientations in Webots (which currently seems to match the X3D spec).

About performance, the operation is currently taking 8 ms in an async method.

fabienrohrer:
I would recommend to go on with this solution as it solves more things than others.

resources/web/wwi/x3d_scene.js
```diff
@@ -345,6 +348,18 @@ X3dScene.prototype = {
     });
   },
 
+  _setupEnvironmentMap: function() {
+    var backgroundMap;
+    if (typeof this.scene.background !== 'undefined' && this.scene.background.constructor.name === 'CubeTexture')
+      backgroundMap = this.scene.background;
+    this.scene.traverse(function(child) {
+      if (child instanceof THREE.Mesh && child.material && child.material.constructor.name === 'MeshStandardMaterial') {
```
stefaniapedrazzi: Why not using `Mesh.isMesh` function instead of `instanceof`?
The rest of the code is currently using the `is<Type>` methods because personally I found that it makes the code more readable.
But apart from personal preferences it is important to be consistent and use the same method (whatever) everywhere.

Same for the material and background:  `MeshStandardMaterial.isMeshStandardMaterial` and `CubeTexture.isCubeTexture`.

`CubeTexture.isCubeTexture` is not documented but I just checked and it is set in the code: https://github.com/mrdoob/three.js/blob/master/src/textures/CubeTexture.js#L23
The `CubeTexture` type check would also have to be replaced in the rest of the code because I didn't found the `isCubeTexture` before.

fabienrohrer: Thank you, I go ahead with your proposition.

fabienrohrer:
👍

omichel:
I believe it is important to check whether the background and environment map orientation in Webots comes from X3D (or VRML) or not. If not, we should probably change it in Webots to match the one of ThreeJS.

PR #389, opened 2019-04-25
Fix release of the OpenCV packages on Windows

### LINKED ISSUE #376: OpenCV example crashes on startup on Windows
"el" on Discord reports us that `vision.wbt` crashes on startup on Windows:

> when i run the vision.wbt world it says this
INFO: vision: Starting: ""D:\Programs\Webots\projects\samples\howto\controllers\vision\vision.exe""
WARNING: vision: The process crashed some time after starting successfully.
WARNING: 'vision' controller crashed.

I looked quickly at this issue, and it seems that OpenCV libraries are [well installed](https://github.com/omichel/webots/blob/master/src/install_scripts/msys64_installer.sh#L28
), but not [well released](https://github.com/omichel/webots/blob/master/src/packaging/files_msys64.txt).
I didn't test more yet.

fabienrohrer:
Fix #376

### DIFF
--- a/change_logs/ChangeLog.html
+++ b/change_logs/ChangeLog.html
@@ -20,6 +20,7 @@ Released on xx xxth, 2019
 </ul>
 <li><b>Bug fixes</b></li>
 <ul>
+<li>Fixed missing OpenCV dependencies on Windows preventing the vision controller to run and to compile.</li>
 <li>Fixed PNG images with a useless alpha layer that were falsly considered as transparent by Webots causing alpha sorting problems.</li>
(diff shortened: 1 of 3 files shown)


omichel:
Adding all `opencv*.dll` yields to adding 46 files (totaling 36MB) whereas only `libopencv_core` and `libopencv_imgproc` are needed (8MB). Also `libquadmath-0.dll` is a dependency of these two OpenCV libraries which was missing.

fabienrohrer:
👀 
Thank you for this, I had planned to care of it :-)

fabienrohrer:
👍

PR #391, opened 2019-04-26
Added Python3 path before msys2 path to prevent python conflicts

omichel:
This fixes a problem on Windows where the MSYS2 python command was used instead of the official Python3, thus causing a failure of the main build process during the compilation of the Python3 wrapper (Python.h was not found).

👍


fabienrohrer:
👀

PR #393, opened 2019-04-29
Use ES6 classes

fabienrohrer:
Migrate to the use of ES6 classes.


fabienrohrer:
👍

stefaniapedrazzi:
:eyes:

resources/web/wwi/x3d.js
```diff
@@ -1,29 +1,27 @@
-/* global THREE, ActiveXObject, TextureLoader */
+/* global THREE, ActiveXObject, TextureLoader, Float32Array, Uint16Array */
```
stefaniapedrazzi: Given that we decided to use ES6, I would rather add "es6" =  true in the eslint json file `.eslintrc.json`.

fabienrohrer: Note that this linter issue is also present in the `develop` branch. I'm checking if we could already change .eslintrc.json.

stefaniapedrazzi: Yes, I know.
I changed `.eslintrc.json` locally but I didn't commit it.
I was about to do it now, but I think it is better to do it in this branch instead of ignoring them.

fabienrohrer: Oh I see, yes I will do it here then.

fabienrohrer:
👍

Issue #396, opened 2019-04-29
Crash copying and pasting an appearance node of the Floor PROTO node


stefaniapedrazzi:
**Describe the Bug**
Webots crashes when copying and pasting an appearance node in the Floor PROTO appearance field.

**Steps to Reproduce**
1. Go to humanoid_sprint.wbt
2. Copy the appearance field of the GRASS Floor node
3. Paste the clipboard node to the appearance field of the TRACK Floor node
4. -> Webots crashes

**Additional context**
This bug arise because the parentNode is accessed after becoming invalid as a consequence of the node regeneration triggered from https://github.com/omichel/webots/blob/d64d6427cbdb1a6f3db27db2dcae8596771521ca/src/webots/nodes/utils/WbNodeOperations.cpp#L152

stefaniapedrazzi:
I cannot reproduce the crash on the master branch.

PR #398, opened 2019-04-29
Fix pasting in a PROTO regenerating field

stefaniapedrazzi:
Addresses #396: resetting the field to NULL before the copy to reset the selection was triggering the node regeneration too early.


stefaniapedrazzi:
test me

stefaniapedrazzi:
:+1:

PR #399, opened 2019-04-29
Hotfix reusing an already loaded texture in webots.js

stefaniapedrazzi:
Fix TextureLoader interface to propagating the texture if it was already loaded and cached.


stefaniapedrazzi:
:+1:

PR #406, opened 2019-04-30
Use ES6 features in loops

fabienrohrer:
Review the loops with ES6 features in mind (`Array.forEach`, `let`, `in`).

Several loops are already using this formalism.


omichel:
Beware that "let" is not supported by IE. You should test if babel handles this properly.

fabienrohrer:
Yes, it's precisely the role of Babel to deal with such cases (as classes, etc.). I test all the current PRs on IE since we migrated to Babel:

Capture d’écran 2019-04-30 à 10 44 18

fabienrohrer:
👍

Issue #281, opened 2019-03-14
Wrong sphere texture mapping in WREN


stefaniapedrazzi:
**Describe the Bug**
The texture mapping for a sphere displayed in Webots doesn't match the X3D texture mapping.

**Steps to Reproduce**
1. Go to 'textured_shapes.wbt'
2. Export HTML5 model
3. Open the exported model in a browser
4. The texture mapping in the browser and in Webots for the sphere are different in the top and bottom area

**Expected behavior**
The sphere texture mapping in WREN should match the X3D one.

**Screenshots**
Webots:
![textured_shapes_webots](https://user-images.githubusercontent.com/5910449/54345907-4c19f200-4644-11e9-83c0-d47759e5ba14.png)
x3dom:
![Screenshot from 2019-03-14 10-28-41](https://user-images.githubusercontent.com/5910449/54345800-15dc7280-4644-11e9-85df-e14bf5bf608b.png)
view3dscene:
![textured_shapes_view3dscene](https://user-images.githubusercontent.com/5910449/54345793-0d843780-4644-11e9-927f-a5d24b83fd6d.png)

fabienrohrer:
The fix should probably target R2019b instead, because it may have consequences for the users to fix the Webots mapping. Is this true?

stefaniapedrazzi:
Yes, i will target 2019b.
I don't think that users will have to change anything, but sensors can behave slightly differently.

stefaniapedrazzi:
The mismatch comes from the different triangulation of the sphere:

![sphere_triangulation](https://user-images.githubusercontent.com/5910449/54348520-78843d00-4649-11e9-8729-605aef9531ef.png)

The WREN implementation (icosphere) is probably better as all the triangles are about the same size. But it doesn't match the sphere (uv sphere shape used by X3D (x3dom) and THREE.js.

I checked in Blender and both primitives `UV Sphere` and `Icosphere` are available.
If we want to keep the icosphere, then we have to export it as IndexedFaceSet and not as a Sphere geometry.

omichel:
Yes, let's do this: export our Sphere primitive as mesh for THREE.js.

DavidMansolino:
Linked to #254

stefaniapedrazzi:
During the meeting we discussed to add a flag to the Sphere node specifying if the mapping/triangulation type: icosphere or uv-sphere.

* Webots: by default creates an icosphere
* VRML import: it imports a uv-sphere
* VRML/X3D export:
[...]

PR #407, opened 2019-04-30
Fix icosphere vs. UV sphere import/export

stefaniapedrazzi:
Addresses #281: Webots uses icosphere, but the VRML/X3D format use UV sphere causing mismatches during import and export.

Tasks:
* Webots:
  * [x] add `Sphere.ico` field defining if the Sphere is a Icosphere or a UV sphere
  * [x] VRML import: import UV sphere by default
  * [x] Webots Sphere creation: create icosphere by default
  * [x] fix Pen device painting on the textured sphere
* WREN:
  * [x] WREN: add a new function to create a UV sphere
  * [x] fix icosphere texture mapping to match UV mapping
* Export:
  * [x] VRML export: export sphere as IndexedFaceSet
  * [x] X3D export: export ico field and use it in webots.js to create a THREE.IcosahedronGeometry instead of a THREE.SphereGeometry
* Documentation
  * [x] update documentation of Sphere node


src/webots/nodes/WbSphere.cpp
```diff
@@ -116,13 +107,17 @@ bool WbSphere::areSizeFieldsVisibleAndNotRegenerator() const {
...
-    writer << " subdivision=\'" << 8 * mSubdivision->value() << ',' << 8 * mSubdivision->value() << "\'";
+  if (writer.isX3d()) {
+    writer << " subdivision=\'" << mSubdivision->value() << ',' << mSubdivision->value() << "\'";
+    writer << " ico=\'" << (mIco->value() ? "true" : "false") << "\'";
+  }
 }
 
 bool WbSphere::sanitizeFields() {
-  if (WbFieldChecker::checkIntInRangeWithIncludedBounds(this, mSubdivision, 1, 6, 1))
-    return false;
```
stefaniapedrazzi: I don't fully understand why it is needed to return false (and skip WREN mesh creation) if a field has not a correct value and it is reset to a default value.
The problem if that if the mesh is not created, the next time you edit/select or interact with the node Webots crashes.

omichel: This mechanism is indeed questionable. It is present in many geometric primitives. We should either keep it everywhere or remove it everywhere. But if we do so, it should be in done in a different PR. If you believe it is wrong, you should open an issue (question) about it.

fabienrohrer: This mechanism is difficult to understand, but useful: this is because `WbFieldChecker` functions [actually change the field value](https://github.com/omichel/webots/blob/master/src/webots/nodes/utils/WbFieldChecker.cpp#L255) if required, and so triggers itself the mesh recreation (if it's not the case, there is maybe a signal-slot issue). Therefore returning is important.

stefaniapedrazzi: I will inspect the issue once completed this PR and open issue if it is required.
But it would help to better understand why the WREN creation is skipped.

omichel: And also, the method name is probably not well chosen. Instead of `checkIntInRangeWithIncludedBounds`, it should probably be named something like `clampIntInRangeWithIncludedBounds`.

stefaniapedrazzi: The `WbFieldChecker` always changes the field value for every node, but only for the geometries we have this more complex workflow that causes crashes with WREN.
Anyway I will check this later, and if it possible to simplify it and make it more robust I think it would be better.

stefaniapedrazzi:
Maximum `subdivision` value for the icosphere in the code and in the doc doesn't match.
I think it is better to change the code and limit the maximum number of faces to 20480, i.e. subdivision 5 and not 6 (81920 faces).

stefaniapedrazzi:
test me

stefaniapedrazzi:
Texture mapping of icosphere in Webots doesn't match the texture mapping of the icosahedron in THREEjs, where it is about the same as for the UV sphere.

In Webots icosphere the texture is stretched (icosphere left and UV sphere right):
texture_mapping

We can keep the current texture mapping and use a custom texture mapping for THREEjs in `webots.js` or change the texture mapping in Webots.

omichel:
I would change the texture mapping in Webots to make it more standard (similar to the one of THREEjs)

fabienrohrer:
The VRML or X3D specs is ambiguous relatively to the choice between linear or radial texture mapping:

> When a texture is applied to a sphere, the texture covers the entire surface ...

- http://graphcomp.com/info/specs/sgi/vrml/spec/part1/nodesRef.html#Sphere
- http://www.web3d.org/documents/specifications/19775-1/V3.3/Part01/components/geometry3D.html#f-Spherenode

As you propose, radial mapping seems more logical.

stefaniapedrazzi:
Note VRML/X3D use UV spheres and not icospheres.
I checked with different VRML viewers and the new Webots UV sphere matches what is displayed in the other viewers (including THREEjs).

For icosphere there is no VRML/X3D reference to follow.

omichel:
> For icosphere there is no VRML/X3D reference to follow.

This is why should probably follow the implementation of THREEjs for icosphere texture mapping.

stefaniapedrazzi:
Yes, I agree.
It was just to specify that icospheres don't exist in VRML/X3D.

fabienrohrer:
tessellation (ico vs uv) and texture mapping are not related: you can do a linear or radial mapping both on ico or uv spheres (but there will be issues on joints in some cases, influencing the choice). VRML or X3D does specifies the mapping, but does not specifies the tessellation. We are free to choose both.

stefaniapedrazzi:
All the VRML/X3D viewer I found use the UV tessellation and mapping.
So even if you could be theoretically free to choose what we want, this is not possible in practice if we want to be compatible with other software.

fabienrohrer:
I agree :-)

omichel:
To me, it is important that when we switch the "ico" field of a Sphere from TRUE to FALSE, the texture mapping on the sphere remains unchanged and hence the sphere remains visually very similar.

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
:+1:

omichel:
:eyes: By the way, why did you set the default value of `Sphere.ico` to `FALSE`?

stefaniapedrazzi:
> By the way, why did you set the default value of Sphere.ico to FALSE?

Because we want to create a UV sphere when reading a VRML file.

docs/reference/sphere.md
```diff
@@ -17,12 +18,23 @@ The `radius` field specifies the radius of the sphere (see [this figure](#sphere
 %end
 
 The `subdivision` field controls the number of faces of the rendered sphere.
-Spheres are rendered as icosahedrons with 20 faces when the subdivision field is set to 0.
-If the subdivision field is 1 (default value), then each face is subdivided into 4 faces, making 80 faces.
+If the `ico` field is TRUE, thus the sphere is rendered as an icosahedron, and the `subdivision` field is 1 (default value), then each face is subdivided into 4 faces, making 80 faces.
```
omichel:
```suggestion
If the `ico` field is TRUE and the `subdivision` field is 0, the sphere is rendered as an icosahedron with 20 faces. If the `subdivision` field is 1 (default value), then each face of the icosahedron is subdivided into 4 faces, resulting in a 80 faces icosahedral.
```

omichel:
> Because we want to create a UV sphere when reading a VRML file.

That breaks badly backward compatibility. Couldn't we instead of that, set the default value to TRUE and change it to FALSE when importing VRML?

stefaniapedrazzi:
The backward compatibility it is not broken.
For spheres stored in files older than R2019b the ico field is automatically set to TRUE.

stefaniapedrazzi:
> I am sorry, but I believe that setting the default value of ico to TRUE is better if we want to encourage users to use icosahedra instead of UV spheres.

In practice nothing changes for users in setting different default values:
* if they create a Sphere from Webots then it will by default an icosphere
* if they load an old Webots file then the world will use icosphere

On the other hand this implementation is more consistent with all the other nodes, where the default values in `<Node>.wrl` defines the VRML specification.
But I will check if there is a simple way to change the values in case of VRML files.

Also note that the major change that breaks the backwards compatibility is not the tessellation but different texture mapping used for icospheres so that UV and ICO sphere texture mapping is as similar as possible.

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
:+1:

@omichel: I changed the default values as you requested.

Just note that now the following exact same node definition will produce different results if stored in a VRML or Webots file
```
Sphere {
  radius 0.1
}
```

stefaniapedrazzi:
:-1: 

Some textured spheres have to be adapted:
- SoccerBall
- ..

stefaniapedrazzi:
For the `SoccerBall`, it can be fixed by changing the texture (the one in the picture is just a test one) and increasing the subdivision value.
Otherwise if we want to keep the number of faces low, we should define it as IndexedFaceSet.

Here is the a icosphere with  subdivision 3.
Further reducing the subdivision will result in more curved lines on the top/bottom area (left ball view)
soccer_ball

omichel:
That looks good to me. Changing the texture is the way to go (if we can find a good one).

stefaniapedrazzi:
test me

stefaniapedrazzi:
:+1:

docs/reference/sphere.md
```diff
@@ -17,12 +18,23 @@ The `radius` field specifies the radius of the sphere (see [this figure](#sphere
...
 With a subdivision field set to 2, 320 faces will be rendered, making the sphere very smooth.
 A maximum value of 5 (corresponding to 20480 faces) is allowed for this subdivision field to avoid a very long rendering process.
-A value of 10 will turn the sphere appearance into a black and white soccer ball.
+If the `ico` field is FALSE, then the sphere is rendered as a UV sphere, and the `subdivision` value has to be in the range [3, 32].
+In this case the `subdivision` field specifies the number of rings and segments in which the sphere is subdivided.
+If the `subdivision` field is set to 3, then the sphere will have 3 rings and 3 segments, making 9 faces.
+With a `subdivision` field set to 32, the faces are 1024.
+
+The `ico` field specifies whether the sphere is rendered as an icosahedron (TRUE) or a UV sphere (FALSE).
+By default [Sphere](#sphere) nodes created in Webots are icosahedrons with `subdivision` value of 1, whether sphere geometries imported from VRML are UV spheres with a `subdivision` value of 24.
```
omichel:
```suggestion
By default [Sphere](#sphere) nodes created in Webots are icosahedrons with `subdivision` value of 1, whereas sphere geometries imported from VRML are UV spheres with a `subdivision` value of 24.
```

src/webots/nodes/utils/WbVisualBoundingSphere.cpp
```diff
@@ -40,31 +40,6 @@ void WbVisualBoundingSphere::enable(bool enabled, const WbBaseNode *node) {
 void WbVisualBoundingSphere::createSphere(
```
omichel: Shouldn't we totally get rid of this dumb method?
The same question applies also to `WbVisualBoundingSphere::show()` and `WbVisualBoundingSphere::clear()`.

stefaniapedrazzi: these methods were very useful for debugging.
I don't know if now with wren there is a better method to debug the bounding spheres.
My idea was to restore these methods, but not in this PR.

omichel: OK, in that case the comment shouldn't be TODO_WREN, but some explanation saying that these methods are useful and should be uncommented for debugging purposes.

stefaniapedrazzi: The problem is that they have not been ported to WREN, this is why there is the `TODO_WREN`.

omichel:
Is it normal to have the soccer ball look like this?
soccer_ball

omichel:
It looks very bad comparing to the RobocupSoccerBall:
nao_robocup

stefaniapedrazzi:
No, the first SoccerBall screenshot is the one with old texture. I will check if I missed to update some textures.

stefaniapedrazzi:
I checked and I get the correct SoccerBall when inserting it from the Add Node dialog.

stefaniapedrazzi:
soccer_ball

omichel:
Sorry, my fault, I was not up-to-date with your branch. Recompiling now...

omichel:
OK, it looks very good to me indeed. Sorry for the confusion.
Maybe we should simply update the SoccerBall proto icon so that it better resembles the new one.

stefaniapedrazzi:
Yes, I will update the icon.

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
:+1:

PR #421, opened 2019-05-06
Update Webots version to R2019a revision 2

stefaniapedrazzi:
Ditto.


stefaniapedrazzi:
test me

stefaniapedrazzi:
:+1:

stefaniapedrazzi:
Personally I think that changing the version on the branch should be the first thing we do.
Especially for major version change and for guaranteeing compatibility this is very important.

DavidMansolino:
> Personally I think that changing the version on the branch should be the first thing we do.

I fully agree, we should do this right after the release.

PR #427, opened 2019-05-06
Fix wwi_url value in webots.js

stefaniapedrazzi:
Fix value of `wwi_url` in case a query was added in the script include declaration as in robotbenchmark use case.
This PR also fixed a problem with the format of the `menu.png` url.

### DIFF
--- a/resources/web/wwi/editor.js
+++ b/resources/web/wwi/editor.js
@@ -41,7 +41,7 @@ class Editor extends DialogWindow { // eslint-disable-line no-unused-vars
       saveShortcut = 'Ctrl-S';
     else // macOS
       saveShortcut = 'Cmd-S';
-    this.menu.innerHTML = '<input type="image" id="webotsEditorMenuImage" width="17px" src="' + DefaultUrl.getImageUrl('menu') + '">' +
+    this.menu.innerHTML = '<input type="image" id="webotsEditorMenuImage" width="17px" src="' + DefaultUrl.wwiUrl() + 'images/' + 'menu.png">' +
                           '<div id="webotsEditorMenuContent">' +
                           '<div id="webotsEditorSaveAction" class="webotsEditorMenuContentItem" title="Save current file">Save<span style="float:right"><i><small>' + saveShortcut + '</small></i></span></div>' +
                           '<div id="webotsEditorSaveAllAction" class="webotsEditorMenuContentItem" title="Save all the files">Save All</div>' +
--- a/resources/web/wwi/editor.js
+++ b/resources/web/wwi/editor.js
@@ -41,7 +41,7 @@ class Editor extends DialogWindow { // eslint-disable-line no-unused-vars
       saveShortcut = 'Ctrl-S';
     else // macOS
       saveShortcut = 'Cmd-S';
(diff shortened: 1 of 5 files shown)


stefaniapedrazzi:
:+1:

resources/web/wwi/editor.js
```diff
@@ -41,7 +41,7 @@ class Editor extends DialogWindow { // eslint-disable-line no-unused-vars
       saveShortcut = 'Ctrl-S';
     else // macOS
       saveShortcut = 'Cmd-S';
-    this.menu.innerHTML = '<input type="image" id="webotsEditorMenuImage" width="17px" src="' + DefaultUrl.getImageUrl('menu') + '">' +
+    this.menu.innerHTML = '<input type="image" id="webotsEditorMenuImage" width="17px" src="' + DefaultUrl.wwiUrl() + 'images/' + 'menu.png">' +
```
omichel: Why not using getImageUrl here?

omichel: I see, it doesn't return the correct string... But then, why don't we remove this getImageUrl method which doesn't seem to be very useful. Is it used in many places?

stefaniapedrazzi: It is used in two places.
I can change it and remove the `url(..)` part so that it can be used here as well.

stefaniapedrazzi:
:+1:

omichel:
I would rather get rid of this getImageUrl() function which is not very useful...

stefaniapedrazzi:
> I would rather get rid of this getImageUrl() function which is not very useful...

I don't agree.
I think it is much better to hard code the url/path to images folder only in one place instead of hard coding it many times everywhere in the code.

omichel:
Regarding `getImageUrl()`, I don't like this kind of method for the following reasons:

1. It doesn't do what its name suggests (e.g., it is specific to PNG and will fail with a JPEG).
2. It is hiding a simple piece of code, harming the understanding of the code using this method: we are just adding a simple prefix and a suffix to a string. To understand what it does, the user should identify and open the corresponding source file, search for the method and check the body of this method. This would be unnecessary if we would simply append an absolute reference to the wwi default url, e.g., `"/images/icon.png"` instead of passing `"icon"` to `getImageUrl()`.

stefaniapedrazzi:
We can change the method so that it return the url to the images `wwi/images` folder if you prefer.
But even if it is a very simple function I really think that it is better to store the `wwi/images/` path in a single place in the the code.
Hard coding it every time we need to access a resource file will result in a big mess and not in a well structured code. And if you will need to change the file structure it will require much more work that simply changing it in a simple position

omichel:
> We can change the method so that it returns the url to the images wwi/images folder if you prefer.

Yes, that would be better.

> But even if it is a very simple function I really think that it is better to store the wwi/images/ path in a single place in the the code.
> Hard coding it every time we need to access a resource file will result in a big mess and not in a well structured code. And if you will need to change the file structure it will require much more work that simply changing it in a simple position

I agree, but this is a trade-off between readability and maintenance. From both points of views, it is not a big deal and won't cause us any big readability or maintenance trouble in any case. So, you suggestion to return the path to the images folder seems to be a good compromise.

PR #422, opened 2019-05-06
Update Webots version to R2019b

stefaniapedrazzi:
Ditto.

PR #428, opened 2019-05-06
Run test_suite in Batch Mode

DavidMansolino:
This will prevent Webots been stuck on the welcome dialog on jenkins each time we update the version (see  #422).


DavidMansolino:
test me

DavidMansolino:
:+1:

Issue #431, opened 2019-05-07
WbGeometry::sanitizeFields doesn't work at load


stefaniapedrazzi:
The `WbGeometry::sanitizeFields` function returns FALSE and skips the WREN mesh creation if a field was invalid and had to be reset to default.
This was done because the field change due to the reset should automatically trigger the WREN mesh recreation.

Unfortunately this mechanism doesn't work at world load and the mesh is not created at all causing other issues and crashes.

stefaniapedrazzi:
Additionally in a normal case where you change a field like the WbSphere::subdivision, the `sanitizeFields` function is called 3 times from the `updateSubdivision` method:
* https://github.com/omichel/webots/blob/revision/src/webots/nodes/WbSphere.cpp#L179 
* https://github.com/omichel/webots/blob/revision/src/webots/nodes/WbSphere.cpp#L139
* https://github.com/omichel/webots/blob/revision/src/webots/nodes/WbSphere.cpp#L154

PR #432, opened 2019-05-07
Fix calls to geometries sanitizeFields

stefaniapedrazzi:
Address #431: change the way the geometry `sanitizeFields` function is called so that meshes are created after reset at load and the calls to `sanitizeFields` function on parameter change are reduced to 1.

This is done by
* moving `sanitizeFields` outside (just before) the `buildWrenMesh` function
* call `sanitizeFields` only when needed, i.e. at load or when a geometry field is changed

### DIFF
--- a/src/webots/nodes/WbBox.cpp
+++ b/src/webots/nodes/WbBox.cpp
@@ -71,6 +71,8 @@ void WbBox::createWrenObjects() {
   WbGeometry::createWrenObjects();
   WbGeometry::computeWrenRenderable();
 
+  sanitizeFields();
+
   const bool createOutlineMesh = isInBoundingObject();
   mWrenMesh = wr_static_mesh_unit_box_new(createOutlineMesh);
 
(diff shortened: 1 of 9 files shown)


stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
:+1:

omichel:
That looks very good, but I have a question: Should we also change the `sanitizeFields()` method to return `void` instead of `bool` and hence simplify it?

stefaniapedrazzi:
> Should we also change the sanitizeFields() method to return void instead of bool and hence simplify it?

At the end I'm not sure that changing the return value to void will really simplify the workflow.
Because the buildWrenMesh method will be called twice (once when the fields checked for the first time and once due to the field value reset) or we need to introduce an additional method to block the field update triggered by sanitizeFields but still emit the `changed` field value signal that is also used by other functionalities.

It is feasible to disconnect and connect the slots, but the current solution seems fine.

omichel:
But is this return value used at some place?

stefaniapedrazzi:
Yes, it is used in all the `update` slot functions.

omichel:
:eyes:

PR #435, opened 2019-05-07
Migrate robot components to THREEjs

### LINKED ISSUE #433: Upgrade the robot components to the new webots.min.js
Porting the robot components to the new THREEjs viewer becomes important.

It's currently broken in some branches:
https://github.com/omichel/webots/pull/430

fabienrohrer:
Fix #433

- [x] Use the new `webots.min.js` viewer
- [x] Restore the functionalities
    - [x] sliders
    - [x] leds
    - [x] device origins
    - [x] Robot size estimation to set the device origin size
    - [x] motor reset
    - [x] view reset
- [x] Increase rendering quality using textured background and pbr
- [x] Regenerate the robot components (required for some new features like PBR and USE-DEF)
- [x] Side improvements
    - [x] Viewpoint fieldOfView seems wrong
    - [x] Fix robot components on the develop branch (update dependencies.txt)
    - [x] Add API to webots.json (pre/postRender callbacks, etc.)
    - [x] Fix static background texture load
    - [ ] ~~Fix IndexedLineSet of Khepera1 => to be fixed in #455~~
    - [x] Add the Tinkerbots to components.json (now this is possible)
    - [x] Implement Cylinder/Cone side/top/bottom flags


fabienrohrer:
It starts to work well:
Preview: https://cyberbotics.com/doc/guide/robots?version=feature-threejs-robot-component

fabienrohrer:
It's now working much better. PBR is renders much better with environment maps:
https://cyberbotics.com/doc/guide/robots?version=feature-threejs-robot-component

It remains to fix:
- [ ] the viewpoint initialization
- [ ] a bug with quaternions (the angle is inverted when reseting some robots)
- [ ] The Khepera1 IndexedLineSet causes a crash in webots.min.js

resources/web/wwi/x3d.js
```diff
@@ -134,6 +134,7 @@ THREE.X3DLoader = class X3DLoader {
   parseTransform(transform) {
     var object = new THREE.Object3D();
     object.userData.x3dType = 'Transform';
+    object.userData.id = getNodeAttribute(transform, 'id', '');
```
stefaniapedrazzi: The `id` is already stored in the `object.name` field and there is a function to retrieve the object by this value.
So this is a duplication of the same data, why do you need it?

fabienrohrer: Thank you, I was not aware of this. I will look if I could reuse `object.name`.

stefaniapedrazzi: I could not use the `userData` field nor the `Object3D.traverse` because these are only available for THREE.Object3D nodes but not for others nodes, like the THREE.Texture.

stefaniapedrazzi:
I don't get any crash when streaming the khepera.wbt simulation. I will check the robot component.
But there is an issue with the DistanceSensor USE nodes and the IndexedLineSet are only displayed for the DEF node.

resources/web/wwi/x3d.js
fabienrohrer: Using `object.name` and `getObjectByCustomId()` is indeed much better. I take the advantage of this to rename this function and change its parameters order (to simplify the calls):

```
getObjectById(id, object=undefined, context=undefined)
```

fabienrohrer:
Just discussed with Stefania, she can't reproduce the khepera issue, I'm going forward with this.

fabienrohrer:
Cylinder and cones side/bottom/top flags are not implemented in webots.min.js, this is required to support the Khepera2 correctly.

fabienrohrer:
Viewpoint field of view is now fixed; `three.js` defines the fov vertically rather than vertically:

fov

Cylinder/Cone bottom/side/top flags are now supported as in Webots:

cylinder-cone

stefaniapedrazzi:
By the way I checked and the issue with the khepera1 IndexedLineSet USE/DEF is not reproducible in the develop branch.

stefaniapedrazzi:
I checked and the USE/DEF and clone mechanism works correctly. I don't know why the LineSegments are not rendered. Note that I get the same issue even using basic nodes instead of USE/DEF.
And once after rotating the viewpoint the missing IndexedLineSet appeared, so it could be a rendering issue rather than a node instantiation issue.

fabienrohrer:
I fixed the rotation bug by giving the initial angle though the meta file.

fabienrohrer:
Khepera1 is not working because of a bug in the `THREE.Object3d.clone()` function in threejs < r103, used here:

https://github.com/omichel/webots/blob/develop/resources/web/wwi/x3d.js#L66

The clone function alterates the line set material from `MeshStandardMaterial` to `LineBasicMaterial`:

lineset bug

This bug is reported here: https://github.com/mrdoob/three.js/issues/15974
.. and fixed in r103: https://github.com/mrdoob/three.js/pull/15062/files
(I tested and upgrading solves the issue).

I recommend to accept this PR with this bug, and to upgrade threejs in another PR: because this has side effects (GPUPicker cause warnings, etc.).

fabienrohrer:
test me

fabienrohrer:
👍

fabienrohrer:
Note for the reviewer:

- the shadows are still perfectible (some artifacts are present on some worlds) but give a better result.
- Khepera1 doesn't work for now: cf #455 

👍

fabienrohrer:
👎 
The OP3 has still issues with the recent modifications we did on the `X3DScene.getObjectById()` function.

fabienrohrer:
(this is fixed)

👍

stefaniapedrazzi:
:eyes:

resources/web/wwi/x3d_scene.js
```diff
@@ -226,23 +237,27 @@ class X3dScene { // eslint-disable-line no-unused-vars
...
+    if (!object ||
+        (skipBoundingObject && typeof object.userData !== 'undefined' && object.userData.x3dType === 'Switch')) {
+      context.parent = undefined;
+      return undefined;
+    }
+
     if (Array.isArray(object)) {
-      for (let i = 0, l = object.length; i < l; i++) {
-        var o = this.getObjectByCustomId(object[i], id, context);
+      for (let i = 0; i < object.length; i++) {
```
stefaniapedrazzi: why did you changed the for loop instructions?
note that it is common to store the length of an array so that the code it is slightly faster.

omichel: I tested on my Firefox, and it turns out that caching the length in a local variable is faster. Try it yourself here: https://jsperf.com/array-length-performance. I am getting this:
image

More info here: https://stackoverflow.com/questions/9592241/javascript-is-the-length-method-efficient

fabienrohrer: In fact, my modification is not important; I will roll it back. I was fearing about the scope of `l` but it was correct. For consistency concern, we are using a lot more the second syntax.

resources/web/wwi/x3d_scene.js
```diff
@@ -49,17 +49,24 @@ class X3dScene { // eslint-disable-line no-unused-vars
...
+    if (typeof this.postRender === 'function')
+      this.postRender(this.scene, this.viewpoint.camera);
   }
 
   resize() {
     var width = this.domElement.clientWidth;
     var height = this.domElement.clientHeight;
     this.viewpoint.camera.aspect = width / height;
+    if (this.viewpoint.camera.fovX)
+      this.viewpoint.camera.fov = this.viewpoint.camera.fovX / this.viewpoint.camera.aspect;
```
stefaniapedrazzi: The correct formula is:
```suggestion
      this.viewpoint.camera.fov = 2.0 * Math.atan(Math.tan(0.5 *  this.viewpoint.camera.fovX) / this.viewpoint.camera.aspect);

File: resources/web/wwi/x3d_scene.js (hunk 2 above)
Comment (fabienrohrer) [reply resources/web/wwi/x3d_scene.js]: I propose that I will roll back this modification, and then create a PR to use this pattern everywhere.

**Comment by fabienrohrer** [comment]
👍

File: resources/web/wwi/x3d.js [hunk 4]
```diff
@@ -674,11 +692,33 @@ THREE.X3DLoader = class X3DLoader {
     var radius = getNodeAttribute(cylinder, 'radius', '0');
     var height = getNodeAttribute(cylinder, 'height', '0');
     var subdivision = getNodeAttribute(cylinder, 'subdivision', '32');
-    var openEnded = getNodeAttribute(cylinder, 'bottom', 'true').toLowerCase() !== 'true';
-    // var openSided = getNodeAttribute(cylinder, 'side', 'true').toLowerCase() === 'true' ? false : true;
-    // var openTop = getNodeAttribute(cylinder, 'top', 'true').toLowerCase() === 'true' ? false : true;
-    // set thetaStart = Math.PI / 2 to match X3D texture mapping
```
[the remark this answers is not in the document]
fabienrohrer: Ok, I just restored it.

resources/web/wwi/x3d.js
```diff
@@ -895,18 +935,19 @@ THREE.X3DLoader = class X3DLoader {
...
       this.scene.viewpoint.camera.near = near;
       this.scene.viewpoint.camera.far = far;
     } else {
       console.log('Parse Viewpoint: error camera');
       // Set default aspect ratio to 1. It will be updated on window resize.
-      this.scene.viewpoint.camera = new THREE.PerspectiveCamera(fov, 1, near, far);
+      this.scene.viewpoint.camera = new THREE.PerspectiveCamera(0.785, 1, near, far);
     }
+    this.scene.viewpoint.camera.fovX = fov;
+    this.scene.viewpoint.camera.fov = fov / this.scene.viewpoint.camera.aspect; // to be updated at each window resize.
```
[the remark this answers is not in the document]
fabienrohrer: Thank you, you're right. I finally used the exact same function we have in Webots and factorized it.

fabienrohrer:
👍

resources/web/wwi/x3d.js
```diff
@@ -1073,3 +1075,17 @@ function convertStringTorgb(s) {
...
+
+function horizontalToVerticalFieldOfView(hFov, aspectRatio) {
+  // Units of the angles: radians.
+  // reference: WbViewpoint::updateFieldOfViewY()
+  var tanHalfFieldOfViewY = Math.tan(0.5 * hFov);
+  // According to VRML standards, the meaning of mFieldOfView depends on the aspect ratio:
+  // the view angle is taken with respect to the largest dimension
+  if (aspectRatio < 1.0)
+    return hFov;
+  else {
```
stefaniapedrazzi: Minor notes: 
- in webots.js, if aspectRatio < 1.0 there is no need to compute tanHalfFieldOfViewY
- this `else` is not needed because of the previous `return`

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

resources/web/wwi/x3d.js
fabienrohrer: I just simplified it a lot ^^

fabienrohrer:
👍

src/webots/nodes/utils/WbWorld.cpp
```diff
@@ -432,6 +432,7 @@ void WbWorld::createX3DMetaFile(const QString &filename) const {
           deviceObject.insert("minPosition", motor->minPosition());
           deviceObject.insert("maxPosition", motor->maxPosition());
           deviceObject.insert("position", motor->position());
+          deviceObject.insert("initialPosition", motor->joint()->solidEndPoint()->rotation().angle());
```
omichel: I am fearing we are adding redundant information here. The previous version of the robot web component didn't have this information and could nevertheless display the range of motors motion properly.

fabienrohrer: This may look as an issue, but I have done this on purpose. If this value is not given, we have to mathematically retrieve it from the threejs object3d rotation. Aside from the fact that this conversion is complex and time consuming, this operation is also certainly impossible: I think some information is lost during the load because of the eulerAxisAndAngle to quaternion implying that the inverse operation is not equivalent. It was possible with X3Dom because they probably have another rotation system closer from our rotation system.

I think the current way is much simpler, efficient and robust.

omichel: I am sure it is not impossible to compute because we could compute it in the previous version. Maybe we should not compute it from the threejs object3d rotation, but earlier, from the x3d data we receive? Also, the computation should be done only once at load time and therefore it's performance impact is low.

fabienrohrer: As told above, I think that the eulerAxisAndAngle->quaternion->eulerAxisAndAngle is not giving the same result, so doing as before is impossible.

It's indeed certainly to store the transform information in x3d.js during the load in the userData. But in this case I will pollute every other threejs projects with unneeded information. What I did here has the advantage to only have this information for the robot components which is the only location for now where we need this.

omichel: My idea was not to store the transform information for every node in x3d.js during the load in the userData, but rather to compute the initialPosition of the motor (or whatever value we need) at this point.

fabienrohrer: Computing the initial position of the motor after the threejs load is not possible because some information is lost.

omichel: All right. I didn't noticed this information was in the robot web component specific x3dmeta file. Given it's specific to the web component, I believe this redundancy is acceptable if it simplifies the design.

fabienrohrer: Ok thank you

Issue #411, opened 2019-05-02
Cylinder solid mesh and outline orientation doesn't match


stefaniapedrazzi:
**Describe the Bug**
The orientation of the Cylinder outline (used for the bounding box) doesn't match the solid mesh one.

**Steps to Reproduce**
1. Go to `geometric_primitives.wbt`
2. Transform the `CYLINDER` Transfomr node to Solid
3. add the DEF name `GEOM` to the Cylinder geometry
4. set the `CYLINDER` Solid.boundingObject field to `USE GEOM`
5. Change the Cylinder.subdivision field of `GEOM` to 3 (or any odd number)
-> the bounding box outline doesn't match the solid mesh

**Expected behavior**
The bounding box outline and the solid mesh should match.

**Screenshots**
![cylinder_bbox](https://user-images.githubusercontent.com/5910449/57059202-65d6cd80-6cb4-11e9-93a0-b5791089fe2b.png)

PR #436, opened 2019-05-08
Fix Cylinder boundingObject representation

stefaniapedrazzi:
Address #411.

Tasks:
* [x] fix orientation of Cylinder outline mesh
* [x] set minimum subdivision (16) value for boundingObject Cylinder and Capsule
    -> make it clear that the ODE shape will be a real cylinder even if the subdivision value used for the graphical mesh is low

### DIFF
--- a/change_logs/ChangeLog.html
+++ b/change_logs/ChangeLog.html
@@ -17,6 +17,7 @@ Released on xx xxth, 2019
 <li><b>Enhancements</b></li>
 <ul>
 <li>Converted many PNG textures to JPEG to save disk space and network bandwidth.</li>
+<li>Improved boundingObject representation of Capsule and Cylinder geometries by setting a minimum subdivision value.</li>
 </ul>
 <li><b>Bug fixes</b></li>
 <ul>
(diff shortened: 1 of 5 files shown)


stefaniapedrazzi:
test me

stefaniapedrazzi:
I was wondering if we had to warn the user about the minimum subdivision value used for bounding objects.
Changing it silently didn't seem very user-friendly, on the other hand having lot of warnings at load or when adding nodes was bad as well.

So the current implementation is:
* no warnings on node finalization
* no warnings for USE nodes
* a warning is displayed if the subdivision field is changed after the node creation (for example when explicitly setting it from the scene tree)

stefaniapedrazzi:
:+1:

DavidMansolino:
:eyes:

DavidMansolino:
It is working fine.

But in case of USE node, the warning is sometimes displayed when in my point of view it should not. For example if you have a Shape=>Cylinder in the children field of a Solid node and then you use the cylinder in the bounding object the warning is displayed.

You can simply reproduce this in the `khepera3_gripper` world, change the 'CYLINDER_LOAD.CYLINDER_SHAPE.Cylinder.subdivision` field, the warning will be displayed.

Wouldn't it be possible to skip the warning in case of USE node?

stefaniapedrazzi:
Yes, the purpose was to not print warnings in the case of USE nodes.
I tested with a simple example of USE Shape and geometry and it was working as expected.

I will check the khepera world.

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

PR #441, opened 2019-05-09
Hinge2Joint Muscle.

### LINKED ISSUE #410: Extend muscle visualization to Hinge2Joint
The BioRob Lab would be interested in having an extended the muscle visualization that also applied to `Hinge2Joint` (even if applied only to the first rotation axis).
Currently only `SliderJoint` and `HingeJoint` nodes supports the `Muscle` node.

I will evaluate the complexity of this new functionality.
A priori this is the current status:
1. the muscle mesh geometry only depends on the endPoint Solid position (should work for Hinge2Joint as well)
2. the muscle mesh material depends on the force extracted from the motor (could work only for a single axis in case of Hinge2Joint)
3. the muscle node is currently defined as a subnode of the Motor node (could work only for a single axis in case of Hinge2Joint)

We already discussed some time ago about the possibility to move the muscle definitions outside the Motor node. This is not a priority now but it will simplify point 3.

DavidMansolino:
Fix  #410

Beta: https://cyberbotics.com/files/repository/beta/webots-R2019b-beta2-x86-64.tar.bz2
doc: https://www.cyberbotics.com/doc/reference/muscle?version=feature-hinge2-joint-muscle


DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
Status: waiting user feedback

DavidMansolino:
I did run a complete travis test (including test suite: https://travis-ci.com/omichel/webots/builds/115773574)

:+1:

stefaniapedrazzi:
:eyes:

src/webots/nodes/WbMuscle.cpp
```diff
@@ -46,18 +46,22 @@ const int SUBDIVISION = 16;            // ellipsoid subdivision
...
 void WbMuscle::init() {
-  mMaxRadius = findSFDouble("maxRadius");
+  mVolume = findSFDouble("volume");
   mStartOffset = findSFVector3("startOffset");
   mEndOffset = findSFVector3("endOffset");
   mColors = findMFColor("color");
   mCastShadows = findSFBool("castShadows");
   mVisible = findSFBool("visible");
 
+  // deprected field
```
stefaniapedrazzi:
```suggestion
  // deprecated field
```

src/webots/nodes/WbMuscle.cpp
```diff
@@ -46,18 +46,22 @@ const int SUBDIVISION = 16;            // ellipsoid subdivision
...
+  mMaxRadius = findSFDouble("maxRadius");
+  if (mVolume->value() == 0.01 && mMaxRadius->value() != 0.0)
+    warn(tr("'maxRadius' field replaced by the 'volume' field, use this new field instead."));
```
stefaniapedrazzi: For consistency with other deprecation messages something like the following would be better:

```suggestion
    warn(tr("'maxRadius' field is deprecated, please use the 'volume' field instead."));
```

```suggestion
    warn(tr("Deprecated 'maxRadius' field, please use the 'volume' field instead."));
```

DavidMansolino:
I did change the starting points and volume of the muscle, now they overlap much less (but ideally we should be able to make them transparent).

DavidMansolino:
👍

stefaniapedrazzi:
:eyes:

PR #444, opened 2019-05-10
Document RosInertialUnit Axes Switch

DavidMansolino:
Since it took me quite some time to find the answer to https://stackoverflow.com/questions/56074321/quaternion-calculation-in-rosinertialunit-cpp-of-webots-ros-default-controller I think it is good to document it.


DavidMansolino:
test me

DavidMansolino:
:+1:

PR #447, opened 2019-05-13
Fix broken if conditions in webots.js parsePointSet

stefaniapedrazzi:
Fix `X3dLoader.parsePointSet` method broken when improving check for undefined variables.


stefaniapedrazzi:
test me

stefaniapedrazzi:
:+1:

PR #450, opened 2019-05-14
Test Multiple Python Version with Travis

DavidMansolino:
It is possible to launch the test with multiple instances in parallel using different versions of Python.


DavidMansolino:
:+1:

PR #461, opened 2019-05-16
Macos Travis Test

DavidMansolino:
Travis doesn't support officially tests written in Python, but I have been able to make it works by using the 'generic' language and manually installing python with brew.


DavidMansolino:
Fully functional on macos :smiley: 

:+1:

Issue #74, opened 2018-12-13
Support Equirectangular Cubemaps on the Web


tn12787:
Equirectangular `Cubemap` nodes aren't exported to X3D right now.

DavidMansolino:
Linked to #254

PR #473, opened 2019-05-20
Support Equirectangular Cubemaps in webots.min.js

stefaniapedrazzi:
Address #74.

Extend `webots.min.js`:
* [x] support HDR image format
* [x] support equirectangular background
* [x] fix HDR background orientation
* [x] merge three.js/examples/ scripts in `webots.min.js`


stefaniapedrazzi:
test me

stefaniapedrazzi:
Screenshot from 2019-05-23 10-18-19

I added the EquirectangularToCubeGenerator.js script to our repository so that we can adapt the shaders and flip the texture horizontally. We can propose a change to three.js but I would wait until the other proposal is accepted to know how they prefer to proceed.

:+1:

fabienrohrer:
We will have issue to let this other proposal be accepted, it is widely refused by the threejs dev team. Therefore your current solution is the best for now. Let's discuss of this next Monday.

fabienrohrer:
:eyes:

fabienrohrer:
I can run this branch in HTML export. It is good. I like the modifications on the Makefile goes in the right direction.

I have lot of questions 😅 :

- It seems the feature is not working in streaming mode, but only in "local" mode (.html export). Is there some limitations there?
- If I understand well, the HDR background is converted to CubeTexture in JS. This may have a performance cost. Did you have an idea of this cost?
- 512x512 faces is very small for a background. Is this on purpose? Performance? I would say at least 1024x1024 would be better to avoid ugly pixelization.
- Do you know if the background pixel format is in float and allows to go beyond 1.0? This is really important for the following post processing.
- It seems antialiasing is lost, but I'm not sure why. In https://github.com/omichel/webots/pull/408/files, I solved this by adding an FXAA shader as in Webots.

stefaniapedrazzi:
> It seems the feature is not working in streaming mode, but only in "local" mode (.html export). Is there some limitations there?

For me it works in streaming mode too.
Which world did you test?
Did you check that you are using the correct local webots.min.js?

> If I understand well, the HDR background is converted to CubeTexture in JS. This may have a performance cost. Did you have an idea of this cost?

We do exactly the same in WREN so I didn't check the costs.

> 512x512 faces is very small for a background. Is this on purpose? Performance? I would say at least 1024x1024 would be better to avoid ugly pixelization.

The texture width is used to define the resolution.
A minimum resolution of 512x512.
I checked in a previous version and it was working as expected. I will check again.

> Do you know if the background pixel format is in float and allows to go beyond 1.0? This is really important for the following post processing.

I didn't check.
But the background uses the HDR loaded data texture that has RGBAFormat and RGBEEncoding, so the values should be float values.

> It seems antialiasing is lost, but I'm not sure why. In https://github.com/omichel/webots/pull/408/files, I solved this by adding an FXAA shader as in Webots.

I was checking this but I didn't find the reason yet.

fabienrohrer:
> Which world did you test?

youbot.wbt
Could you try this one? If it works I may have fail something.

> Did you check that you are using the correct local webots.min.js?

Yes, but I changed index.html to do so.

> I was checking this but I didn't find the reason yet.

I had issues on the other branch because of the use of EffectComposer which is incompatible with antialiasing. But it seems not to be the case here.

stefaniapedrazzi:
> youbot.wbt
Could you try this one? If it works I may have fail something.

Ok, I see. I have a test streaming texture folder containing the background image and thus the texture is loaded from file instead of using the streamed one.
I will fix it.

stefaniapedrazzi:
The streaming issue is fixed.
Waiting for the background (#483) and texture streaming improvements to finalize this PR.

stefaniapedrazzi:
I found this thread that seems to describe the issue we have with antialiasing:
https://discourse.threejs.org/t/why-is-a-custom-fbo-not-antialiased/1329

omichel:
Does our webots.js work with a WebGL 2 context? If so, we could query if WebGL 2 is available and use it if available (and thus have anti-aliasing working), otherwise we could fall back to WebGL 1 with no anti-aliasing. That seems reasonable.

stefaniapedrazzi:
I checked and webots.js seems to work with WebGL 2 context.
So I will switch to it and add a fallback to WebGL 1.

If later we will found some issues we can switch back to WebGL 1 and pply the FXAA pass.

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
The last issue is to improve the background resolution but I didn't find yet where the HDR equirectangular background texture is limited.
I tried to load a equirectangular JPEG image and it gives a better resolution.

resources/web/wwi/equirectangular_to_cube_generator.js
```diff
@@ -0,0 +1,109 @@
...
+  var boxMesh = new THREE.Mesh(new THREE.BoxBufferGeometry(1, 1, 1), getShader());
+  boxMesh.material.side = THREE.BackSide;
+  scene.add(boxMesh);
+
+  class EquirectangularToCubeGenerator {
+    constructor(sourceTexture, options) {
+      options = options || {};
+
+      this.sourceTexture = sourceTexture;
+      this.resolution = options.resolution || 512;
```
fabienrohrer: > I didn't find yet where the HDR equirectangular background texture is limited.

Not sure, but isn't it here?

stefaniapedrazzi: No, this line simply set a lower resolution, i.e. if `options.resolution > 512` then `this.resolution = options.resolution`

stefaniapedrazzi:
:+1:

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
The difference between the results with HDR and JPEG images is due to the linear magFilter: in case of JPEG images the result is similar to what we have in Webots.

Given that we plan to replace HDR images by JPEG images, the current status could be enough.

:+1:

fabienrohrer:
👀

stefaniapedrazzi:
@fabienrohrer has many warnings on his system with WebGL 2. I will revert the change and use FXAA pass instead.

fabienrohrer:
Thank you, it seems indeed to be a more reliable flow, and more compliant with the future developments including the post-processing branch (which uses already the FXAA shader without issues).

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
:+1:

resources/web/wwi/Makefile
```diff
@@ -39,8 +40,27 @@ BABEL_JS_SOURCES = \
...
+  pmrem/PMREMGenerator.js \
+  shaders/CopyShader.js \
+  shaders/DepthLimitedBlurShader.js \
+  shaders/FXAAShader.js \
+  shaders/GammaCorrectionShader.js \
+  shaders/SAOShader.js \
+  shaders/UnpackDepthRGBAShader.js \
+  postprocessing/EffectComposer.js \
+  postprocessing/RenderPass.js \
+  postprocessing/SAOPass.js \
```
fabienrohrer: The dependency list should be minimized (it contains several unnecessary files for this PR, like the SAO file).

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
:+1:

PR #478, opened 2019-05-21
Fix Various Python Regex.

DavidMansolino:
Python 3 interprets string literals as Unicode strings, and therefore the `\d` is treated as an escaped Unicode character causing the following pep8 warning: `DeprecationWarning: invalid escape sequence \d`


DavidMansolino:
test me

PR #481, opened 2019-05-21
Optimize and simplify the CubeMap rotation in `three.js`

### LINKED ISSUE #387: Cubemap orientation
Following #375, the background orientation (and so environment map reflections) in THREEjs and Webots are rotated by 180° vertically. This is currently solved in JS by flipping the top/bottom images and inverting the right/left and front/back textures.

We could explore now a better solution: to break the Webots background orientation.
First step is to determine our compliance with X3D to determine if this is a good idea or not.

fabienrohrer:
Fix #387 in case https://github.com/mrdoob/three.js/pull/16507 is accepted.


fabienrohrer:
Status: wait on https://github.com/mrdoob/three.js/pull/16507 merge and new threejs release.

fabienrohrer:
Won't fix it this way.

PR #486, opened 2019-05-22
Travis Package Creation Test

DavidMansolino:
The goal of this PR is to test every day the development environment installation procedure, complete compilation process, and package creation:
  - This is done on a daily basis and not on every commit to not slow done the development and PR approval process.
  - The test is done on the 'revision' branch, but if it works fine I propose to extend it to the develop branch too.
  - Thanks to this new test we will then be very close from creating a nightly build of Webots using Travis.
  - For now, the only missing step is the PROTO cache creation as it requires the GUI (we can probably remove this constraint but this is out of the scope of this PR).
  - This PR already allowed me to fix various code issues (see diff).

Expected result when running the test: https://travis-ci.org/omichel/webots/builds/535699990?utm_source=github_status&utm_medium=notification


DavidMansolino:
test me

DavidMansolino:
:+1:

fabienrohrer:
👀 (macOS modifications)

PR #487, opened 2019-05-22
Generalize the modal window to the entire documentation

fabienrohrer:
- [x] Use the Blog modal window in the entire documentation
- [x] Simplify their initialization (create the modal div only once, instead of once per image)
- [x] Support thumbnail maximization (the original image is loaded in the modal window in case of thumbnails).

Preview: https://cyberbotics.com/doc/guide/samples-environments?version=feature-modal-window-generalization


fabienrohrer:
test me

fabienrohrer:
👍

fabienrohrer:
My first attempt was to use the ladybug GIF but it was too small to be good looking: its white background is problematic.
Therefore the present solution was the quickest to have a good looking solution.

But if you agree, I can spend more time to include the lady bug animation.

omichel:
Yes, that would be nice. Thank you.

fabienrohrer:
Ok, it was easier than I thought, because I have a CSS rule for this image:

Capture d’écran 2019-05-23 à 10 04 01

omichel:
Although I cleared my cache, I don't get the nice rounded white rectangle visible on your screenshot... I also tried to purge webots-doc.css on jsDelivr to no avail...

fabienrohrer:
Yes, webots-doc.css has not been refreshed yet.

I just discussed with the jsdelivr dev team, they told be the purge is not working in case we use the branch name in the url. In this case, we need to wait that the cache is cleared on their server and that can take hours. They plan to solve this, but they don't told me delays.

fabienrohrer:
Sorry for the inconvenience, but meanwhile to solve this, this PR could be tested locally (my last modification is very safe).

👍

fabienrohrer:
The cache has just been updated :-D This is now testable in the preview url.

omichel:
OK, that looks great. Thank you. But can't we avoid duplicating this load animation file? Why not referring to the original load_animation.gif file instead of copying it?

fabienrohrer:
The doc is supposed to work offline. All resources should be included.

omichel:
OK, in that case, shouldn't we have `webots.js` use the `load.gif` of the doc instead of https://github.com/omichel/webots/blob/master/resources/web/wwi/images/load_animation.gif? That would avoid a redundancy of this file.

fabienrohrer:
In my opinion, creating a bidirectional dependency link between webots.min.js and the doc for a such minor topic is a bad idea.

omichel:
OK, I understand. But at least we should use the same file name (as it has the same content).

fabienrohrer:
Fair enough, I change this.

fabienrohrer:
test me

fabienrohrer:
👍

fabienrohrer:
😃

PR #494, opened 2019-05-24
Fix Geometry Update not Awakening Solids

### LINKED ISSUE #409: ElevationGrid does not update boundingobject
**Bug description**
When an elevationGrid is defined, it works fine with respect to collision. However, when a supervisor is used to change the height parameters of the elevationGrid, the visual appearance is updated correctly, but the elevationGrid's collision no longer works properly. The elevationgrid seems still be using its old bounding object, rather than the new one which should follow the updated height values. I also noticed, that before changing height parameters with the supervisor, if you change the position of the elevationgrid and then apply the supervisor's height update, then the problem is no more present and the elevation grid's collision works as expected.

**Steps to reproduce**
Using a supervisor, change the height field parameters of an already defined elevationgrid and then have something collide with the elevationgrid.

**Expected behavior**
I would expect the collision of the elevationgrid to work with the new height values from the supervisor.

**System**
 - Graphics Card:

DavidMansolino:
Attempt to fix #409

### DIFF
--- a/change_logs/ChangeLog.html
+++ b/change_logs/ChangeLog.html
@@ -38,6 +38,7 @@ Released on xx xxth, 2019
 <li>Fixed geometry not created on world loading if invalid geometry fields are written in the file.</li>
 <li>Fixed mismatch between boundingObject outline and graphical representation of a Cylinder if the subdivision field is set to an odd value.</li>
 <li>Fixed 'getRangeImageArray' function in the Python API returning always None.</li>
+<li>Fixed runtime changes of kinematic solid bounding objects not taken into account when colliding with sleeping dynamic solids.</li>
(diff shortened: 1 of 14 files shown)


DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
:+1:

omichel:
:eyes:

Issue #175, opened 2019-02-07
Impossible to Recompile Qt Robot-Windows from the Windows Release 


DavidMansolino:
**Describe the Bug**
Recompilation of Qt robot-windows from the Windows release fails with the following error:
```
# linking robotis-op2_window.dll
build/release/Viewer.o:Viewer.cpp:(.text+0x160): undefined reference to `QArrayData::deallocate(QArrayData*, unsigned long, unsigned long)'
build/release/Viewer.o:Viewer.cpp:(.text+0x160): relocation truncated to fit: R_X86_64_PC32 against undefined symbol `QArrayData::deallocate(QArrayData*, unsigned long, unsigned long)'
build/release/Viewer.o:Viewer.cpp:(.text+0x1b0): undefined reference to `QArrayData::deallocate(QArrayData*, unsigned long, unsigned long)'
build/release/Viewer.o:Viewer.cpp:(.text+0x1b0): relocation truncated to fit: R_X86_64_PC32 against undefined symbol `QArrayData::deallocate(QArrayData*, unsigned long, unsigned long)'
build/release/Viewer.o:Viewer.cpp:(.text$_ZN5QListI7QStringED2Ev[_ZN5QListI7QStringED2Ev]+0x97): undefined reference to `QArrayData::deallocate(QArrayData*, unsigned long, unsigned long)'
build/release/Viewer.o:Viewer.cpp:(.text$_ZN5QListI7QStringED2Ev[_ZN5QListI7QStringED2Ev]+0x97): relocation truncated to fit: R_X86_64_PC32 against undefined symbol `QArrayData::deallocate(QArrayData*, unsigned long, unsigned long)'
build/release/TransferWidget.o:TransferWidget.cpp:(.text+0x1f3): undefined reference to `QArrayData::deallocate(QArrayData*, unsigned long, unsigned long)'
build/release/TransferWidget.o:TransferWidget.cpp:(.text+0x1f3): relocation truncated to fit: R_X86_64_PC32 against undefined symbol `QArrayData::deallocate(QArrayData*, unsigned long, unsigned long)'
build/release/TransferWidget.o:TransferWidget.cpp:(.text+0x30d): undefined reference to `QArrayData::deallocate(QArrayData*, unsigned long, unsigned long)'
[...]

PR #506, opened 2019-05-28
Windows: distribute patched gcc-9 instead of gcc-7

### LINKED ISSUE #288: Flashback occurs when using a physical plugin
**Describe the Bug**
I want to realize that when two robots move forward in my project, they do not collide, but pass through their bodies. The engineer told me that using physical plug-ins may achieve this effect, and let me first familiar with samples / howto / Worlds / physics.wbt, but I just commented on one of the lines of code: 

//dWebotsConsolePrintf("Initialization: ROBOT = %p %p\n", robot_geom, robot_body);

The webots quit directly, and when they re-entered, they executed the project as soon as they entered, and immediately quit.

**System**
 - Windows 10

omichel:
Fixes #288 and #175.

I finally found the cause of the bug preventing gcc-8 to link when it was installed in a folder with a space. The problem comes from `windres` which fails at executing in a folder containing a space. `windres` is called at link time by `ld` to link the resulting executable with a file called `default-manifest.o`. If this file is not present, then `ld` doesn't attempt to use it and doesn't fail. It seems this `default-manifest.o` file is not very important and according to my tests, everything works fine without it. See https://github.com/msys2/MSYS2-packages/issues/454 for details. Removing this file from the distribution actually fixes gcc-8 link and hence allow us to use gcc-8 instead of gcc-7 in the distribution.


omichel:
👍

omichel:
My PR on MSYS2 with a patch fixing gcc was accepted, merged and a new gcc package is available. See https://github.com/msys2/MINGW-packages/pull/5450.
Therefore I can simplify this PR.

👎

omichel:
👍

DavidMansolino:
👀

DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
test me

DavidMansolino:
test me

omichel:
Thank you for the tests!

PR #508, opened 2019-05-28
Let the streaming server to serve textures

fabienrohrer:
- [x] Let live QTcpSocket and QWebSocketServer on the same port using `QWebSocketServer::handleConnection()`
- [x] Serve textures
- [x] Update `webots.min.js` accordingly


fabienrohrer:
test me

fabienrohrer:
Without speaking of cache, the loading time of a streaming client is improved from 35% to 55% locally on my computer.

Cache is also working (but not performance improvement is visible locally or on a local computer):
cache

fabienrohrer:
👍

resources/web/wwi/webots.js
```diff
@@ -190,7 +190,8 @@ webots.View = class View {
           this.server = new Server(this.url, this, callback);
           this.server.connect();
         } else { // url expected form: "ws://cyberbotics2.cyberbotics.com:80"
-          this.stream = new Stream(this.url, this, finalizeWorld);
+          var httpServerUrl = this.url.replace(/ws/, 'http'); // Serve the texture images.
```
omichel: This seems to work fine with http, but not https. We should quickly extend it to https as well (possibly in a separate PR) otherwise, it will be a regression.

resources/web/wwi/stream.js
```diff
@@ -108,14 +109,6 @@ class Stream { // eslint-disable-line no-unused-vars
...
       data = data.substring(data.indexOf(':') + 1).trim();
       this.view.updateWorldList(currentWorld, data.split(';'));
-    } else if (data.startsWith('image')) {
-      // Extract texture url: the url should only contain escaped ']' characters.
-      var urlPattern = /[^\\]\]/g; // first occurrence of non-escaped ']'
-      var match = urlPattern.exec(data);
-      var textureUrlEndIndex = match.index + 1;
-      var textureUrl = data.substring(data.indexOf('[') + 1, textureUrlEndIndex).replace(/\\]/g, ']');
-      data = data.substring(data.indexOf(':', textureUrlEndIndex) + 1);
-      TextureLoader.loadFromUri(data, textureUrl);
```
stefaniapedrazzi: This is the only call to `TextureLoader.loadFromUri`, therefore it should be removed from the TextureLoader class.

resources/web/wwi/webots.js
fabienrohrer: This is deserving a comment, but this line supports `https` (it replaces the first `ws` by `http`, so it can change `wss` => `https`). Nevertheless, I need to test more seriously SSL connections.

resources/web/wwi/stream.js
```diff
@@ -2,17 +2,18 @@
...
-    this.url = url;
+  constructor(wsServer, httpServer, view, onready) {
+    this.wsServer = wsServer;
+    this.httpServer = httpServer;
     this.view = view;
     this.onready = onready;
     this.socket = null;
     this.videoStream = null;
-    TextureLoader.setStreamingMode(true);
+    TextureLoader.setStreamingMode(true, this.httpServer);
```
stefaniapedrazzi: The `httpServer` in this class seems to be only used to setup the `TextureLoader`.
And this is the only call to `TextureLoader` in the `Stream` class.

I think it would be better to move this instruction in `webots.js` given that the textures are no longer streamed.
This way we can also remove the unneeded dependency of `Stream` on `TextureLoader`.
Also note that the first argument of `setStreamingMode` is no longer used and could be completely removed.

stefaniapedrazzi:
I tested both streaming from Webots and running robotbenchmark locally and now they work both correctly.

stefaniapedrazzi:
When connecting with SSL it doesn't work.
The solution seems to override the QTcpServer so that it creates a QSslSocket connection instead of a QTcpSocket connection.
With this solution I can currently connect to the test simulation server machine using the streaming_viewer.html and WSS protocol. But I'm still checking why passing through the simulation server the connection times out.

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
At the end the remaining issue was only due to the port, maybe some Webots instances were not closed properly. Now it works correctly also using the streaming server.

fabienrohrer:
👍

PR #515, opened 2019-05-31
Fix UnicodeDecodeError when importing OSM on windows.

omichel:
Use the codecs to correctly read in utf-8 file, preventing UnicodeDecodeError


DavidMansolino:
:eyes:

Issue #510, opened 2019-05-29
Animation: initial state not restored when looping


stefaniapedrazzi:
**Describe the Bug**
When looping an animation created for the robot_programmig.wbt simulation, the initial state is not restored when moving to step 0.
I checked with other simulations, for example the hoap2_sumo.wbt, and it works correctly.

**Steps to Reproduce**
1. Go to 'robot_programming.wbt'
2. Record an animation of the robot that first doesn't move for 600 ms and then moves to the target (see attached controller)
3. Run it in the browser
4. When the animation restarts, the robot state is not re-initialized until the 600 ms are elapsed.

**Expected behavior**
The robot initial state should already be restored at step 0.

robot_programming.wbt test controller:
```
"""Simple robot controller."""

from controller import Robot

# Define the target motor position in radians.
target = 11
# Get pointer to the robot.
robot = Robot()

robot.step(600);

# Set the target position of the left and right wheels motors.
robot.getMotor("motor.left").setPosition(target)
robot.getMotor("motor.right").setPosition(target)
```

stefaniapedrazzi:
I compared to the animation stored on robotbenchmark.net, and there it works correctly because the initial state is stored in the json file at time step 0.

stefaniapedrazzi:
The main issue is that objects fields are now updated separately, i.e. the JSON update line for a transform can contain only the rotation and not the translation value.

Because of this, the optimized mechanism to run the simulation backwards is broken.
The current mechanism loops through the steps only until the object id is found and then breaks the search.
Given that we cannot known if other field updates for the same object are missing we need to search through the whole steps.

I will now evaluate the slow down removing the optimization and using the "new" exported JSON file format.

PR #518, opened 2019-06-03
Fix animation

stefaniapedrazzi:
Addresses #510.
Fix jumping to a previous animation pose:
* store initial state in JSON file (only for nodes that will change during the animation recording)
* remove optimized search mechanism in `webots.min.js` and search in all the steps for updates

Other improvements:
* remove duplicated pause and play images used for the animation
* fix other wrong/unused exported lines
* restore correct appearance of pendulum in inverted_pendulum.wbt (broken when switching to PBRAppearance)


stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
test me

stefaniapedrazzi:
When measuring the performance of jumping from one pose to the other in this branch and on robotbenchmark, the slow down of removing the loop optimization is quite big (in pick and place jumping from start to end takes 157ms instead of 15ms) but on my PC the update time is acceptable.

I will try to speed it up by storing in the animation JSON file only the ids that changes and not all the possible ids.

stefaniapedrazzi:
:+1:

fabienrohrer:
👀

PR #500, opened 2019-05-27
Cleanup Remove 'environmentMap' Field of PBRAppearance Node

DavidMansolino:
Fix first part of https://github.com/omichel/webots/issues/483#issuecomment-496243286

PR #519, opened 2019-06-03
Fix references to the dropped SphereGrid.cubeMap field

fabienrohrer:
`SphereGrid.cubeMap` has been dropped in #500 but few references remain.

This prevents `physically_based_rendering.wbt` to be loaded correctly.


fabienrohrer:
👍

PR #528, opened 2019-06-04
Reduce gap between Webots rendering and webots.min.js

fabienrohrer:
- [x] Adjust the global ambient light to shift the rendering by some offset.
- [x] Deal with sRGB map for raw colors, textures and background colors as in Webots.
- [x] Remove the `threejs` gamma management which is not compliant with the Webots rendering pipeline.
- [x] Add a final pass dealing with exposure and gamma correction, the same as the Webots one.
- [x] Support the Viewpoint.exposure field.
- [x] Support colored background as an environment map rather than an AmbientLight (as in Webots) by generating a CubeTexture.
- [x] Reduce the gap between the PointLight/SpotLight attenuation management and the threejs lights.
- [x] Fix roughness rendering: multiply its contribution by 2 give results close to the Webots ones.
- [x] Adjust rendering with HDR backgrounds


fabienrohrer:
Some (working well) results:

apartment
colored_background_metalness1
colored_background_metalness1_with_quadratic_attenuation
tinkerbots
youbot

fabienrohrer:
For sure, there is still a gap between the Webots rendering and the threejs one. The remaining things are more and more tedious to solve with the threejs API. But this PR solves a lot of issues and reduce for sure drastically this gap.

fabienrohrer:
👍

stefaniapedrazzi:
:eyes:

stefaniapedrazzi:
I'm getting this with the robotbenchmark square_path world (green walls with three.js) do you have the same?

Screenshot from 2019-06-12 10-51-16

fabienrohrer:
No it's working like a charm locally on my computer:

Capture d’écran 2019-06-12 à 11 13 49 (2)

Are you testing: projects/samples/robotbenchmark/square_path/worlds/square_path.wbt
Any modification?

stefaniapedrazzi:
Sorry, it was a strange cache problem.

fabienrohrer:
I prefer this 😸

resources/web/wwi/x3d.js
```diff
@@ -823,15 +829,26 @@ THREE.X3DLoader = class X3DLoader {
...
     var intensity = parseFloat(getNodeAttribute(light, 'intensity', '1'));
     var location = convertStringToVec3(getNodeAttribute(light, 'location', '0 0 0'));
     var radius = parseFloat(getNodeAttribute(light, 'radius', '100'));
     var castShadows = getNodeAttribute(light, 'castShadows', 'false').toLowerCase() === 'true';
 
-    var lightObject = new THREE.PointLight(color.getHex(), intensity);
-    lightObject.decay = attenuation.x;
+    var lightObject = new THREE.PointLight(color.getHex());
+
+    // Tradeoff to let cohabit VRML light attenuation and the threejs light "physically correct mode".
```
stefaniapedrazzi: we should be consistent in the way we write three.js. In another file we use `three.js`.

I propose to change it here, but it is also fine to change the other occurrence(s).

```suggestion
    // Tradeoff to let cohabit VRML light attenuation and the three.js light "physically correct mode".
```

fabienrohrer: We should certainly use `three.js`, this is the terminology used on their doc: https://threejs.org/docs/index.html#manual/en/introduction/Creating-a-scene

fabienrohrer:
> The only thing that I personally find that it is not very intuitive and consistent is that PBRAppearance colors are exported with a different format than other colors. But if this is the best way to achieve a full match it is ok.

Indeed, it's difficult to do better here, this is the Webots implementation.

fabienrohrer:
👍

omichel:
> The only thing that I personally find that it is not very intuitive and consistent is that PBRAppearance colors are exported with a different format than other colors.

I don't understand this comment. It seems `PBRAppearance.baseColor` uses the same format as `Appearance.diffuseColor` in X3D files or did I miss something?

stefaniapedrazzi:
`Appearance.diffuseColor` is in linear space, `PBRAppearance.baseColor` is in sRGB space.
In the X3D file they seems to be the same but they need to be handled differently and this is what I think it is not very good.

PR #536, opened 2019-06-05
Speed-up cppcheck CI tests

omichel:
This PR makes that cppcheck tests only the C/C++ files that are modified in a branch (compared to the parent branch).


omichel:
Travis tests now take 3 to 5 minutes.
👍

omichel:
👍

DavidMansolino:
:eyes:

tests/sources/test_cppcheck.py
```diff
@@ -101,57 +167,35 @@ def setUp(self):
...
         command = self.cppcheck + ' --enable=warning,style,performance,portability --inconclusive --force -q '
         command += '--inline-suppr --suppress=invalidPointerCast --suppress=useStlAlgorithm -UKROS_COMPILATION '
         # command += '--xml '  # Uncomment this line to get more information on the errors
-        command += '--std=c++03 --output-file=' + self.reportFilename
-        for source in self.projectsSkippedDirs:
-            command += ' -i\"' + os.path.normpath(self.WEBOTS_HOME + '/' + source) + '\"'
-        for source in self.projectsSourceDirs:
-            command += ' \"' + os.path.normpath(self.WEBOTS_HOME + '/' + source) + '\"'
+        command += '--std=c++03 --output-file=\"' + self.reportFilename + '\"'
+        if os.path.isfile(files_diff):
```
DavidMansolino: The content of this if is almost duplicated from the on of the `test_sources_with_cppcheck` function, wouldn't it be possible to make a function to avoid duplication?

tests/sources/test_cppcheck.py
```diff
@@ -101,57 +167,35 @@ def setUp(self):
...
+            added = False
+            file = open(files_diff, 'r')
```
DavidMansolino:
```suggestion
            with open(files_diff, 'r') as file:
               ...
```

tests/sources/test_cppcheck.py
```diff
@@ -101,57 +167,35 @@ def setUp(self):
...
+                                break
+            file.close()
+            if not added:
+                return
+        else:
+            for source in projectsSkippedDirs:
+                command += ' -i\"' + source + '\"'
+            for source in projectsSourceDirs:
+                command += ' \"' + source + '\"'
+        os.chdir(self.WEBOTS_HOME)
```
DavidMansolino: Is this really required? This looks dangerous and I am not sure to understand why you need it?

omichel: That's because I pass all the files and folder to `cppcheck` as relative to `WEBOTS_HOME` and not absolute. This is very useful to reduce the size of the command line which can be very long and sometimes exceeds the maximum length allowed on Windows, e.g., 8192 chars.

omichel: To improve this, I can save the previous `cwd` and restore at after the command has been issued.

tests/sources/test_clang_format.py
```diff
@@ -88,7 +88,8 @@ def test_sources_are_clang_format_compliant(self):
         files_diff = os.path.join(self.WEBOTS_HOME, 'tests', 'sources', 'files_diff.txt')
         sources = []
         if os.path.isfile(files_diff):
-            file = open(files_diff, 'r')
+            with open(files_diff, 'r') as file:
+                ...
```
DavidMansolino: 😉

tests/sources/test_cppcheck.py
DavidMansolino: In most of the cases the tests are already run in the correct folder because we call them with `python -m unittest discover -s ...`. But I can indeed reproduce the issue when calling directly this test.

I think it would indeed be safer to restore the `cwd`, I am a bit afraid that otherwise it can 'pollute' the next test when launched with `unittest`.

omichel:
👍

omichel:
👍

omichel:
👍

tests/sources/test_clang_format.py
```diff
@@ -21,7 +21,8 @@
 import os
 import subprocess
 import sys
```
DavidMansolino:
```suggestion
```

tests/sources/generate_diff.py
```diff
@@ -0,0 +1,30 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Generate a list of modified files with respect to the parent branch.
+This script is used only by Travis."""
+import os
+import subprocess
+import sys
+
+branch = os.getenv('TRAVIS_BRANCH')  # branch targeted by the pull request
```
DavidMansolino: I just read the documentation of this variable:
>for push builds, or builds not triggered by a pull request, this is the name of the branch.
for builds triggered by a pull request this is the name of the branch targeted by the pull request.
for builds triggered by a tag, this is the same as the name of the tag (TRAVIS_TAG).

This means that if one day we will want to re-enable branch push test, this script will brake the test because it will test the diff with itself and therefore not test anything (without notifying us).

omichel:
👍

PR #540, opened 2019-06-06
Readd selection and context menu functionlity on webots.min.js for mobile devices

stefaniapedrazzi:
Fix touchEnd callback re-adding the selection and long click functionality.


stefaniapedrazzi:
:+1:

PR #558, opened 2019-06-12
Fix openVR includes installation.

DavidMansolino:
The includes were copied into `include/openvr/headers/*.h` instead of `include/openvr/*.h`.


DavidMansolino:
:+1:

PR #559, opened 2019-06-12
Fix Makefile call order

fabienrohrer:
- [x] Fix Makefile call order: ROS libraries download may occur after the ROS controller build on mac
- [x] Fix `make` call on the `./projects` directory which cleans instead of build: `cleanse` should never be the first rule.
- [x] Fix the pico rule of `cleanse` on macOS


fabienrohrer:
👍

PR #561, opened 2019-06-12
Improve variable name

fabienrohrer:
Following a comment of @omichel

### DIFF
--- a/projects/robots/telerob/telemax_pro/controllers/telemax_demo/telemax_demo.c
+++ b/projects/robots/telerob/telemax_pro/controllers/telemax_demo/telemax_demo.c
@@ -25,7 +25,7 @@
 #define TIME_STEP 32
 
 void noop(int duration) {
-  // Wait without applying any command during 'duration' simulation steps.
+  // Wait without applying any command during 'duration' milliseconds.
   int counter = 0;
   while (counter < duration) {
     counter += TIME_STEP;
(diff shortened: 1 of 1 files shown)


fabienrohrer:
👍

projects/robots/telerob/telemax_pro/controllers/telemax_demo/telemax_demo.c
```diff
@@ -24,10 +24,10 @@
 
 #define TIME_STEP 32
 
-void noop(int n_steps) {
-  // Wait without applying commands during 'n_steps' steps.
+void noop(int duration) {
+  // Wait without applying any command during 'duration' simulation steps.
```
omichel:
```suggestion
  // Wait without applying any command during 'duration' milliseconds.
```

fabienrohrer: Your suggestion is wrong, the unit of the argument is a number of simulation steps, not milliseconds. Please check the function content.

fabienrohrer: Therefore `n_steps` is better than `duration` IMO

omichel: I don't think so: `counter` is accumulating `TIME_STEP` which is expressed in milliseconds and then compared to `duration` which should therefore be milliseconds as well.

fabienrohrer: Oops, you're right. Sorry for this.

fabienrohrer:
👍

PR #570, opened 2019-06-14
Fix Release

### LINKED ISSUE #568: Release broken on mac and linux
The release is broken on mac and linux:
```
# creating webots.deb version R2019b
Missing file: projects/robots/robotis/darwin-op/libraries/python27/libnagers.so
```

This is because the release script is assuming that all the `*.so` and `*.dylib` should start with `lib` which is not the case for libraries generated by swig.

DavidMansolino:
Fix #568

### DIFF
--- a/src/packaging/webots_distro.c
+++ b/src/packaging/webots_distro.c
@@ -636,17 +636,22 @@ static void create_file(const char *name, int m) {
         type |= (TYPE_LINUX | TYPE_MAC | TYPE_WINDOWS);
       if (type & TYPE_DLL && (mode == MAC || mode == DEB)) {  // prefix "lib" to the basename
         j = i - 1;
-        while (buffer[j] != '/' && j >= 0) {
-          buffer[j + 3] = buffer[j];
+        while (buffer[j] != '/' && j >= 0)
           j--;
+        if (buffer[j + 1] != '_') {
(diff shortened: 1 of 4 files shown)


src/packaging/webots_distro.c
```diff
@@ -636,17 +636,22 @@ static void create_file(const char *name, int m) {
...
-        if (j >= 0) {
-          buffer[++j] = 'l';
-          buffer[++j] = 'i';
-          buffer[++j] = 'b';
-          i += 3;
-        } else
-          fprintf(stderr, "DLL parsing: Reached the beggining of the string without finding a '/' character.");
       }
       if (mode == ISS) {
         if (type & TYPE_EXE) {
```
DavidMansolino: By the way, this script is really horrible we should urgently rewrite it from scratch (in Python :wink:):
  - half of the script is for handling strings and pathes...
  - it is under-documented.
  - it is not robust at all and contains tons of exception/specific cases.
  - there is no point to have a compiled executable for such kind of script.
  - it requires input from other script in Python.

DavidMansolino:
Distribution is now fixed on mac an linux: https://travis-ci.com/omichel/webots/jobs/208073529

DavidMansolino:
:+1:

PR #573, opened 2019-06-17
Fix insertion of USE PROTO into parent PROTO parameter

### LINKED ISSUE #332: DEF node not found after inserting node from Add node dialog
**Describe the Bug**
Wrong DEF node check after inserting USE node from Add node dialog.

**Steps to Reproduce**
1. Go to maze_runner.wbt
2. Select the second Window node
3. Delete the Window.wallAppearance node
4. Open Add new.. dialog and choose the WINDOW_WALL_APPEARACE USE node (again)
-> You will get an error in the console telling that no matching DEF is found and the USE node is turned into a DEF node.

**Expected behavior**
The DEF node exists and is found at load and when opening the Add node dialog. Thus it should be correctly found after setting it from the Add node dialog as well.

stefaniapedrazzi:
Fix #332: there is no need to check the validity of internal nodes of a USE node.


stefaniapedrazzi:
:+1:

omichel:
:eyes:

PR #575, opened 2019-06-17
Fix ffmpeg dependencies

omichel:
This should help us reducing the risk of breaking the ffmpeg-based movie creation in Windows distributions. We had several times the problem caused by updates of the DLL dependencies of `ffmpeg.exe` on Windows.


omichel:
👍

DavidMansolino:
👀

src/packaging/copy_msys64_files.py
```diff
@@ -83,6 +83,12 @@ def list_dependencies(package):
                 print('# \033[1;31m' + l + ' is already included\033[0m')
             else:
                 files.append(l)
+print("# processing ffmpeg dependencies (DLLs)")
+sys.stdout.flush()
+script = os.getenv('WEBOTS_HOME') + "\\src\\packaging\\ffmpeg_dependencies.sh"
```
DavidMansolino:
```suggestion
script = os.path.join(os.environ['WEBOTS_HOME'], 'src', 'packaging', 'ffmpeg_dependencies.sh')
```

PR #578, opened 2019-06-17
Update Makefile.windows

DavidMansolino:
#558 was still not perfect, because in some case as the `$(TARGET_PATH)/openvr_api.dll` rule create the `$(WEBOTS_HOME_PATH)/include/openvr` directory, the `$(WEBOTS_HOME_PATH)/include/openvr` rule was not called.


DavidMansolino:
👍

omichel:
:eyes:

PR #584, opened 2019-06-18
Make travis per commit tests on one platform only

DavidMansolino:
The goal of this PR is to test on linux-python27 and mac only during the nightly build or if triggered with the `travis_sources` keyword).


DavidMansolino:
This time we can't make the test faster than this...

:+1:

PR #586, opened 2019-06-18
Fix Robotis library "make clean" rule

fabienrohrer:
`projects/robots/robotis/darwin-op/libraries/python/managers.i` is wrongly removed during a `make clean`.


fabienrohrer:
👍

PR #588, opened 2019-06-18
macOS Fix crash on startup

fabienrohrer:
Due to our recent modifications on the libode build, the RPATH in libode was wrong causing a crash on the Webots statup when installed from a package.


fabienrohrer:
This is also causing crashes in physics plugins.

fabienrohrer:
👍

PR #589, opened 2019-06-18
Add moose controller in the Makefile list

fabienrohrer:
The `moose` controller is not build in the main build process.


fabienrohrer:
👍

PR #590, opened 2019-06-18
Fix URe wbproj

fabienrohrer:
This perspective is not consistent with the other ones of the guided tour, and the controller is not displayed in the text editor.


fabienrohrer:
👍

fabienrohrer:
👍

PR #594, opened 2019-06-18
Fix warning about deprecated wallTileSize in camera_auto_focus

fabienrohrer:
The warning is:

```
ERROR: 'RectangleArena.proto': Lua error: The 'wallTileSize' field is deprecated, using the new 'wallApperance' field instead.
```


fabienrohrer:
👍

PR #596, opened 2019-06-19
Enable MacOS Nightly Builds

DavidMansolino:
Remaning tasks:
  - [x] Fix installation image.
  - [x] Fix java version.


src/packaging/webots_distro.c
```diff
@@ -942,7 +943,9 @@ static void create_file(const char *name, int m) {
       fprintf(fd, "' | osascript\n");
       fprintf(fd, "chmod -Rf go-w \"/Volumes/%s\"\n", application_name);
       fprintf(fd, "sync\n");
+      fprintf(fd, "sleep 5\n");
       fprintf(fd, "sync\n");
+      fprintf(fd, "sleep 5\n");
```
DavidMansolino: Adding sleeps here fixes the creation of the package on travis for macOS.

I am currently testing the package creation on my fork before merging this: https://travis-ci.org/DavidMansolino/webots-test/builds/547575865

omichel: Sounds like sorcery... Thank you Apple! 🙈

DavidMansolino: It seems to work fine: https://github.com/DavidMansolino/webots-test/releases/tag/nightly_19_6_2019

DavidMansolino:
Only remaining issue: the osascript used to set the installation image is causing a timeout, the result is that when installing the package we don't see the installation image. Not sure how to fix this, I tried to increase default timeout but it didn't change, not sure we can afford using finder on travis.

omichel:
It may be worth checking this: https://github.com/andreyvit/create-dmg

DavidMansolino:
> It may be worth checking this: https://github.com/andreyvit/create-dmg

In fact I am currently trying this option right now :smile: It seems indeed to simplify everything!

DavidMansolino:
I have successfully created the package with `create-dmg`, but it doesn't help:
  - It appears they are also using Finder: https://github.com/andreyvit/create-dmg/blob/master/support/template.applescript#L2 which cause the exact same problem for creating the installation window.
  - They have also sometimes a disk detach failure (which was apparently now fixed in our home-made version of the script), they aslo have arbitrary wait/delay call: https://github.com/andreyvit/create-dmg/blob/master/support/template.applescript#L61

DavidMansolino:
:+1: 

The result can be tested on my fork: https://github.com/DavidMansolino/webots-test/releases/tag/nightly_21_6_2019

fabienrohrer:
👀

DavidMansolino:
:+1:

DavidMansolino:
Test suite passed on my fork: https://travis-ci.org/DavidMansolino/webots-test/builds/550135521?utm_source=github_status&utm_medium=notification

PR #617, opened 2019-06-19
Fix Nightly Version in Splash Screen

DavidMansolino:
We should not show the commit ID in the splash screen:
Capture_decran_2019-06-19_a_14 04 44


DavidMansolino:
:+1:

PR #618, opened 2019-06-19
Fix camera_noise_mask to render more easy the red blob detection

### LINKED ISSUE #609: camera_noise_mask.wbt: red box is not recognized
In R2018b, the controller in `camera_noise_mask.wbt` was able to recognize the green, the blue and the red boxes.
On develop, the red box is no longer recognized.

fabienrohrer:
Fix #609 

Since R2018b, an HDR background has been added to this world, changing the global light, and render more difficult the red blob detection.

I added a central light, change the pink color of this block to red, and increase the blob surface by rotating slightly the block. Should be much more robust now.

camera_red_blob

### DIFF
--- a/projects/samples/devices/worlds/camera.wbt
+++ b/projects/samples/devices/worlds/camera.wbt
@@ -15,9 +15,10 @@ TexturedBackgroundLight {
 }
 RectangleArena {
 }
-DirectionalLight {
-  direction 1 -1 -1
-  intensity 0.4
+PointLight {
+  attenuation 0 0 1
+  intensity 0.6
+  location 0 0.6 0
 }
 DEF GREEN_BOX Solid {
   translation -0.05 0.05 -0.25
@@ -36,7 +37,7 @@ DEF GREEN_BOX Solid {
   name "green box"
   boundingObject Shape {
     appearance PBRAppearance {
-      roughness 1.1102230246251565e-16
+      roughness 0
       metalness 0
     }
     geometry USE BOX0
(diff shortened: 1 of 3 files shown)


fabienrohrer:
👍

omichel:
:eyes: (I just noticed that `camera_motion_blur.wbt` has the same problem: only the green blob is detected)

fabienrohrer:
I just observed that `camera.wbt` skips also time to time the red blob.

fabienrohrer:
I will use the same setup in these 3 worlds, in this PR.

omichel:
I saw that on this PR, that `camera_noise_mask.wbt` still skips the red or the green blob in rare cases. I let it run on Windows for a while and I noticed that sometimes, the red or green blob was skipped. But I think it is good that it happens as it looks realistic, and demonstrates the potential effect of a random noise on a camera image with a weak image processing algorithm. That is very realistic and we should keep it like that.

omichel:
We should however avoid that some blobs are skipped in 100% of the cases as it currently happens with `camera_motion_blur.wbt`.

fabienrohrer:
Are you sure of this? As a user looking for a camera example and without comments (and it's difficult to explain a random behavior), I would see this as a bug. This seems a weird decision to me.

=> I have the patch ready. Please confirm me you would like this random behavior.

fabienrohrer:
FYI just discussed with Olivier, and it seems to be better to fix everything and homogenize these 3 worlds as I have a patch which is ready for this.

fabienrohrer:
👍

omichel:
:eyes:

fabienrohrer:
Great, thank you for the tests.

PR #622, opened 2019-06-20
Fix Nightly Release

DavidMansolino:
The nightly release was published twice, once because of the CRON table and once because the release creates a tag.


DavidMansolino:
:+1:

PR #625, opened 2019-06-20
Improved Telemax Overlays

### LINKED ISSUE #610: Improve camera overlays settings for telemax.wbt
I would maybe reduce the default size of the telemax camera overlays, because when running Webots on a 15.6'' screen the 3D and the robot itself are not visible.

telemax

DavidMansolino:
Fix #610

It is a bit sad that now on high-resolution screens the camera overlays look very small, but since it is the first world of the guided tour it should work on every setup.


DavidMansolino:
:+1:

stefaniapedrazzi:
:eyes: 

Probably the best solution for all the setup will be to not store the overlay size as a percentage of the original image size, but as a percentage of the 3D scene size (similarly on how the overlay position works).

PR #629, opened 2019-06-20
Fix Tutorial Number

DavidMansolino:
Since we are going to release really soon, I think we can afford to merge this in develop instead of master.


DavidMansolino:
:+1:

PR #636, opened 2019-06-24
Fix HDR environment maps when disconnecting/connecting the streaming viewer

### LINKED ISSUE #628: Mismatch Between Webots and Webots.js
**Describe the Bug**
In the `telemax.wbt` world, if you connect to the streaming server after the led gets on, the rendering is hugely different:

Capture du 2019-06-20 16-43-58

fabienrohrer:
Fix #628 

In such case, the `_setupEnvironmentMap()` function was called without argument after the world load. The simplest solution seems to simplify the `_setupEnvironmentMap()` function in order to have no arguments, and to rely on the background scene (as it's already done for the other backgrounds).

- [ ] Once merged, update `webots.min.js` on the FTP


fabienrohrer:
👍

stefaniapedrazzi:
:eyes:

PR #642, opened 2019-06-24
Always use 'play' in cookie storing the animation status

stefaniapedrazzi:
Previously `this.gui` was supporting the values `play` and `pause`, but when changing the icon from `play.png` to `real_time.png`,  `play` is internally converted to `real_time` in animation.js.
In order to keep compatibility and given that `play` is more descriptive it is good to continue using it from outside the Animation class.

This PR adds a function to return the status of the Animation and convert back `real_time` to `play`, needed on robotbenchmark to store the animation status in the `animation_gui` cookie.

### DIFF
--- a/resources/web/wwi/animation.js
+++ b/resources/web/wwi/animation.js
@@ -24,6 +24,10 @@ class Animation { // eslint-disable-line no-unused-vars
     xmlhttp.send();
   }
 
+  getStatus() {
+    return this.gui === 'real_time' ? 'play' : 'pause';
+  }
+
   // private methods
   _setup(data) {
     this.data = data;
--- a/resources/web/wwi/animation.js
+++ b/resources/web/wwi/animation.js
@@ -26,6 +26,7 @@ class Animation { // eslint-disable-line no-unused-vars
 
   // Return the animation status: play or pause.
   // This should be used to store the current animation status and restore it later when calling webots.View.setAnimation().
+  // This is for example used in robotbenchmark.net benchmark page.
   getStatus() {
     return this.gui === 'real_time' ? 'play' : 'pause';
   }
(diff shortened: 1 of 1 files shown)


stefaniapedrazzi:
:+1:

fabienrohrer:
:eyes:

fabienrohrer:
Sorry to quibble, but as this is used outside the `webots` repo, I think this deserves a comment: what is using this? why not directly using `this.gui`?

stefaniapedrazzi:
> Sorry to quibble, but as this is used outside the webots repo, I think this deserves a comment: what is using this? why not directly using this.gui?

robotbenchmark is using this in `footer.php` to save the status of the animation and restore it when you load a benchmark page.

We can directly use `this.gui`, but then we have to modify the old code to save and retrieve the animation status so that it uses everywhere `real_time` instead of `play`.
But personally I prefer `play` rather than `real_time` because I think it is much more descriptive.

fabienrohrer:
I understand this, I only ask to add a comment in this function explaining that: I fear to read this function in 3 months, see that it is unused here, and remove it.

stefaniapedrazzi:
I just added a comment

stefaniapedrazzi:
:+1:

fabienrohrer:
Could we refer in the comment that it's used in footer.php of robot benchmark?

stefaniapedrazzi:
> Could we refer in the comment that it's used in footer.php of robot benchmark?

I can add it. But personally I think it is bad to refer to specific applications.
The best documentation is that giving that we have a method to set the animation status at start we also need a method to store it.

stefaniapedrazzi:
I added the reference to robotbenchmark but not to the specific file: testing will easily point out the problem.
But note that we should always test any modifications to robotbenchmark too, independently if there are comments referring to it or not.

:+1:

PR #643, opened 2019-06-24
macOS: Fix controller build in case WEBOTS_HOME contains a space

fabienrohrer:
This issue occurs when the user explicitly rename/move `/Applications/Webots.app` to something containing a space (e.g. `/Applications/Webots R2019b.app`)


fabienrohrer:
👍

PR #648, opened 2019-06-25
Delete Old Nightly Builds

DavidMansolino:
When creating a new nightly build all the nightly build older than 2days12hours are removed (tag too). This way we should have always only the last 3 nightly builds.


DavidMansolino:
:+1:

PR #408, opened 2019-04-30
THREEjs post processing

fabienrohrer:
Explore:

- SSAO
- Bloom
- FSAA
- Tone and Gamma

Capture d’écran 2019-04-30 à 15 35 40
Capture d’écran 2019-04-30 à 15 36 58

PR #651, opened 2019-06-25
Bloom in webots.min.js

fabienrohrer:
Aim: Add the bloom post processing in webots.min.js

Apply and improve #408, now possible thanks to #528

- [x] Replicate the bloom rendering pass in `webots.min.js`
- [x] Map the `Viewpoint.bloomThreshold` with `webots.min.js`
- [x] Adjust parameters to reduce the rendering gap.
- [x] Log

Around this PR:

- [x] Fix Viewpoint.bloomThreshold runtime update which was not applied.
- [x] Add a `debug` mode to the `webots.min.js` Makefile in order to build a `webots.debug.js` file which is not minified and obfuscated (very helpful to debug).

### DIFF
--- a/change_logs/ChangeLog.html
+++ b/change_logs/ChangeLog.html
@@ -9,7 +9,20 @@
 
 <h1>Webots change log</h1>
 
-<h2><a href='https://cyberbotics.com/doc/blog/Webots-2018-b-release'>Webots R2019b</a></h2>
+<h2><a href='https://cyberbotics.com/doc/blog/Webots-2020-a-release'>Webots R2020a</a></h2>
+Released on December ??th, 2019
+
+<ul>
+<li><b>New Features</b></li>
+<ul>
+<li>Improved the Webots online 3D viewer: 'webots.min.js'</li>
+<ul>
+<li>Improved support of the Webots rendering pipeline: supported the Bloom post-processing effect.</li>
+</ul>
+</ul>
+</ul>
+
+<h2><a href='https://cyberbotics.com/doc/blog/Webots-2019-b-release'>Webots R2019b</a></h2>
 Released on June 25th, 2019
 
 <ul>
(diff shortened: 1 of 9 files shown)


fabienrohrer:
The bloom is applied in a very similar way as the Webots one. I copied almost exactly the shaders we have in Webots.

Here are some results:

| Bloom threshold | Webots | webots.min.js |
|---|---|---| 
| Bloom 0 | webots_bloom_0 | threejs_bloom_0  | 
| Bloom 1 | webots_bloom_1 | threejs_bloom_1 | 

Where to look at:

where to look at

As the bloom input is dealt by `three.js` and has not floating point values bigger than 1.0, then the twinkling effect appearing on saturated edges (the most visible effect of the bloom as implemented in Webots) is almost not visible in this implementation:

| threejs | Webots |
|---|---|
| threejs | webots |

Going further is another story: we have to deal with HDR in the threejs pipeline.
I think the current status is good enough to be merged as it reduces the rendering diff between Webots and threejs.

fabienrohrer:
👍

stefaniapedrazzi:
I'm getting this error with some of the robotbenchmark worlds, like the `square_path` and `robot_programming`. Do you have the same?

```
Uncaught TypeError: (intermediate value)(intermediate value)(intermediate value).now is not a function
    at mn.getDelta (webots.min.js?R2019b:1)
    at THREE.EffectComposer.render (webots.min.js?R2019b:1)
    at e.value (webots.min.js?R2019b:1)
    at e.onTextureLoad (webots.min.js?R2019b:1)
    at e.value (webots.min.js?R2019b:1)
    at webots.min.js?R2019b:1
    at HTMLImageElement.i (webots.min.js?R2019b:1)
```

stefaniapedrazzi:
Sorry, forget my previous comment. It is three.js version 105 that causes the issue.

fabienrohrer:
Phew 😄

PR #654, opened 2019-06-25
Fix default blog page

omichel:
This is allow us to automatically update the default blog page.

### DIFF
--- a/docs/doc.php
+++ b/docs/doc.php
@@ -15,6 +15,10 @@
     header("Location: $request_uri");
     exit();
   }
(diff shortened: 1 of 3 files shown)


omichel:
👍

fabienrohrer:
It seems to work well. I like the fact to remove the "silent" argument which was difficult to understand.

Shouldn't we target `master` instead? I guess we will deploy it as soon as possible.

omichel:
I re-targeted to the master branch.
:+1:

PR #667, opened 2019-06-27
Feature Several Branches in Nightly Builds

DavidMansolino:
This way we will be able to perform the nightly builds both on develop and revision and to put both in the same nightly build.


DavidMansolino:
:+1:

PR #672, opened 2019-06-27
Fix Connector Behavior in ROS Controller

DavidMansolino:
Reported here: https://stackoverflow.com/questions/56775714/using-the-webots-connector-device-with-ros


DavidMansolino:
:+1:

PR #673, opened 2019-06-27
webots.min.js: Ambient occlusion

fabienrohrer:
Aim: Find a simple way to add ambient occlusion in webots.min.js

- [ ] Replicate the Webots GTAO in Webots
- [ ] Log


fabienrohrer:
`threejs` SAO and SSAO techniques are unfortunately much less advanced from our GTAO technique.

I can disable the occlusions on the background, but their is a halo effect I cannot remove. Here is an (exaggerated) illustration of the unwanted effect:
 
Capture d’écran 2019-07-02 à 09 57 29

We should certainly go in the GTAO direction.
SAO and SSAO techniques are way more simplistic than GTAO (one shader vs four shaders with history management).
I would say that it will cost us at least 2 weeks of dev with lot of incertitude.

omichel:
OK, let's go ahead with GTAO. It is certainly very important to have good ambient occlusion.

fabienrohrer:
### Analysis of GTAO rendering pipeline

#### Framebuffer outputs of the GTAO shaders in Webots

| shader order | shader pass name | result |
|---|---|---|
| 1 | gtao.frag | gtao_frag_output |
| 2 | gtao_spatial_denoise.frag | gtao_spatial_denoise_output |
| 3 | gtao_temporal_denoise.frag | gtao_temporal_denoise_output |
| 4 | gtao_combine.frag | gtao_combine_output0 |

#### Special shader pass inputs:

- `gtao.frag / inputTextures[2]`: gtao_input2_zoomed (zoomed) gtao_input2_zoomed (original)

fabienrohrer:
Wrong direction.

PR #683, opened 2019-07-01
Feature build on appveyor

omichel:
AppVeyor VM only has OpenGL 1.1, however it seems [possible](https://github.com/Wangscape/Wangscape/pull/121/files) to install a better OpenGL.
AppVeyor VM is Windows 2012 server and misses `C:\Windows\System32\XInput9_1_0.dll` (which is available since Windows Vista).


omichel:
👍

DavidMansolino:
:eyes:

src/packaging/publish_release.py
```diff
@@ -67,13 +67,17 @@
 
 if not releaseExists:
     print('Creating release "%s" with tag "%s" on commit "%s"' % (title, tag, options.commit))
+    if tag:
+        draft = True
+    else:
+        draft = False
```
DavidMansolino: You can replace these line by (but if you think it is more understandable like this it's fine for me too):
```
draft = True if tag else False
```

projects/robots/robotis/darwin-op/plugins/robot_windows/robotis-op2_window/SSH.cpp
```diff
@@ -92,13 +92,13 @@ int SSH::openSSHSession(const QString &IPAddress, const QString &username, const
 int SSH::verifyKnownHost() {
   int state;
```
DavidMansolino: The changes to this file are not compiling on Ubuntu 16.04.

DavidMansolino: I did just launch a Travis test to make sure it is not only on my environment.

omichel:
👍

PR #627, opened 2019-06-20
Hotfix rounding fields values on WBT write

stefaniapedrazzi:
Fix #582: it seems that using `QDoubleSpinBox::text().toDouble()` instead of `QDoubleSpinBox::value()` solves the issue.

PR #684, opened 2019-07-01
Hotfix rounding issues

### LINKED ISSUE #582: Rounding issues saving float values to WBT file
There are rounding issues writing float values to world file.
This only occurs for some specific values, for example 1.6 but not 1.5.

**Steps to Reproduce**
1. Open new world
2. Change `Viewpoint.far` to 1.6
3. Save the world
-> in the WBT file the far value is exported as `far 1.5999999999999998`

**Expected behavior**
The export field value should be `far 1.6` and not `far 1.5999999999999998`

**System**

stefaniapedrazzi:
Fix #582: improve fix in #627 by changing the rounding method so that it uses the `WbPrecision::doubleToString` rounded value.

Using the string value to round reduces differences when saving the value to file and make it more robust to the intrinsic floating noise.
Note that for almost all the values the two rounding methods are equivalent.

This is the same principle it was used in #627 where the value was extracted from the spin box text, but now it has a higher precision that fixes the reset to default issue.

### DIFF
--- a/src/webots/maths/WbPrecision.cpp
+++ b/src/webots/maths/WbPrecision.cpp
@@ -53,23 +53,6 @@ QString WbPrecision::doubleToString(double value, Level level) {
   }
 }
 
-const double WbPrecision::epsilon(Level level) {
-  switch (level) {
-    case DOUBLE_MAX:
-      return std::numeric_limits<double>::epsilon();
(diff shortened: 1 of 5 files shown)


stefaniapedrazzi:
:+1:

omichel:
:eyes:

PR #686, opened 2019-07-01
Avoid Nighlty Release During Weekend

DavidMansolino:
We don't want to create nightly release during the week-end.


DavidMansolino:
:+1:

PR #691, opened 2019-07-03
Fixed Travis source tests

omichel:
That should fix the Jenkins source test with Python 2.7.
It also fixes the test on Windows which was failing.


omichel:
👍

DavidMansolino:
In such case, it is probably worse starting the appropriate test before putting the thumb-up (I am starting it right now).

omichel:
All the tests are now passing.
👍

PR #709, opened 2019-07-11
Fix Texture Picking

DavidMansolino:
The wrong pixel was taken, which was in some cases even causing crashes (e.g. is uv = (1.0, 1.0)).


DavidMansolino:
:+1:

PR #716, opened 2019-07-15
Merge master into revision

omichel:
This should fix the change log link in the GitHub releases (we should also merge revision into develop subsequently).


omichel:
Oops. There is a small conflict...

PR #721, opened 2019-07-15
Speed-up Inno Setup compression

### LINKED ISSUE #707: AppVeyor Nighlty Build failure
Travis returned a failure on this nightly build on revision, however no failure is visible in the log output and the nightly release was correctly uploaded:

https://ci.appveyor.com/project/omichel/webots-sj08b?fullLog=true

omichel:
This should fix #707.

Changing the compression ratio from "max" to "fast" and avoiding to compress `jpg` and `png` files (which are already compressed) allow us to lower the Inno Setup compilation time from 10:13 minutes down to 3:13 minutes on my computer. However, the resulting `webots_setup.exe` is 7% bigger (1,523 MB instead of 1,417 MB). On the appveyor VM, the Inno Setup compilation takes about 15 minutes, so the gain should be even better.

I believe this is a good compromise.

### DIFF
--- a/src/packaging/webots_distro.c
+++ b/src/packaging/webots_distro.c
@@ -224,6 +224,9 @@ static void copy_file(const char *file) {
       fprintf(fd, "Source: \"%s\"; DestDir: \"{app}\\%s\"", file2, dest2);
       if (file2[i + 1] == '.')
         fprintf(fd, "; Attribs: hidden");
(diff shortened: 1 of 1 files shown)


omichel:
👍

omichel:
Meanwhile, I also asked to the appveyor support team to extend our 60 minutes limit to 90 minutes. I am waiting for their answer...

omichel:
I just got an answer from appveyor support team and they extended our account from 60 to 90 minutes. So that should be largely sufficient.

DavidMansolino:
Very good news !

PR #724, opened 2019-07-15
webots.min.js: fix viewpoint reset and other improvements

### LINKED ISSUE #720: webots.min.js: viewpoint not reset when clicking the reset simulation button
After clicking on the reset simulation button, the scene is reset but the viewpoint is not immediately reset.
Only on first step it is reset to the initial position and orientation.

Additionally the reset icon has a wrong vertical line (on develop)
Screenshot from 2019-07-15 16-12-38

stefaniapedrazzi:
Fix  #720 (not the icon issue).

I also fixed some other issues:
* fix selection reset when moving the viewpoint
* avoid two consecutive costly rendering when applying an animation pose (one for the scene updates and one for the viewpoint change)
* fix buttons behavior in benchmark record windows


stefaniapedrazzi:
:+1:

fabienrohrer:
👀

PR #733, opened 2019-07-17
Improve Mavic 2 Pro helix by setting Propeller.fastHelix

fabienrohrer:
- [x] Add a fastHelix representation
- [x] Calibrate the maxVelocity to 5000 [rad/s]
- [x] Fix drone control accordingly
- [x] Open the `Propeller.fastHelixThreshold` field
    - [x] Doc: https://cyberbotics.com/doc/reference/propeller?version=enhancement-mavic-helix
    - [x] Log: https://cyberbotics.com/doc/reference/changelog-r2019?version=enhancement-mavic-helix

mavic_2_pro_helix


fabienrohrer:
After calibrating the Propeller correctly, the hard-coded threshold to switch from a slow helix to a fast helix is too small to be switched in my simulation. So this hard-coded value is clearly too high in some cases. I simply open this as a new field (`Propeller.fastHelixThreshold`) to be able to reduce it in the Mavic simulation, and set it to the default value to not break the backward compatibility.

ezgif-2-fa3ad3720d3c

fabienrohrer:
👍

fabienrohrer:
Ok, I will look what I can do with this, but in another PR.

DavidMansolino:
I agree for the contrast, but not for the darkness, if you look on the official website, the robot is quite dark: https://www.dji.com/ch/mavic-2

fabienrohrer:
Ok, thank you for your comments.

I have done a quick search, and it seems that the color I choose is already a good tradeoff. It may appear too dark or too bright because this plastic part is a bit reflective.

In the screenshot you post, the picture has been done in a well-lighted studio, probably with a strong white balance, and it appears lighter. But in real pictures of the robot, it looks much darker. Here is a small sample of what we can found on Google Images: (red: darker, green: lighter, blue: quite equal):

mavic color

About contrast, it depends a lot of what is mirrored into the metallic parts because of the reflections:

mavic_2_pro_metallic_parts

So if you both agree, I think it's best to keep it as-is.

omichel:
Yes, but we should make sure that such metallic details are well visible in the demo simulation and video. Maybe we could improve the lightning conditions or viewpoint to show a more contrasted view of the drone, showing such details.

fabienrohrer:
Yes, I can look at this. It seems that the metallic parts are too reflective, causing the contrast issues.

PR #736, opened 2019-07-18
Merge revision into develop

DavidMansolino:
No conflicts but I had to update various PROTO headers: be5323d


DavidMansolino:
:+1:

PR #739, opened 2019-07-19
Fix External Controller

DavidMansolino:
Once an external controller was exiting the `INFO: 'extern' controller exited successfully.` message was never displayed, it was not possible to re-launch the external controller and the simulation was blocked (because waiting for ht controller).


DavidMansolino:
:+1:

PR #740, opened 2019-07-22
Improve Mavic 2 PRO materials and lights

fabienrohrer:
.. to reduce the differences between real pictures and Webots.


fabienrohrer:
| Webots (if there are doubts) | Real world  |
| --- | --- |
| mavic_2_pro |  looking-down-birds-eye-view-onto-a-flying-dji-mavic-2-pro-personal-drone-flying-very-low-near-to-the-ground-T5FFAH |
| mavic_2_pro_1 | mavic-2-pro-blog |

fabienrohrer:
Better rendering could be achieved if `Background.intensity` would be open (a branch about this is pending in develop)

fabienrohrer:
👍

Issue #749, opened 2019-07-24
webots.min.js: improve performance


stefaniapedrazzi:
The new robotbenchmark version based on PBR appearances and three.js takes longer to load a scene. Additionally on `obstacle_avoidance` benchmark, rotating the viewpoint doesn't work smoothly when the simulation is just loaded (it works fine once all the objects in the scene has been completely rendered at least once).

stefaniapedrazzi:
I did some tests on the `obstacle_avoidance` benchmark and both issues are strictly related to textures:
* PBRAppearances have high resolution (2048x2048)
* Some of the PBRAppearances nodes are used many times (for example 30 times the `VarnishedPine` one)

I tried two different changes separately (just to measure the performance):
* reduce the resolution of big textures from 2048x2048 to 1024x1024
* reuse the same THREE.Texture object in `webots.min.js` for all the `ImageTexture` instances having the same texture filename (not taking care of other parameters that could change like the texture transform or the wrap)

With any of these two changes I could achieve a loading time of the scene comparable to the current robotbenchmark.net one.
The viewpoint rotation issue was not completely solved (still not always smooth) but it was much better than before and the viewpoint rotation didn't blocked completely.

omichel:
It seems we should really avoid the duplication of identical textures and reuse them as much as possible.

stefaniapedrazzi:
Yes, reusing images it is not enough and we also have to add a mechanism to reuse textures.

But I would also inspect if it is really necessary to use so big resolutions or if we could achieve the same with smaller images.

stefaniapedrazzi:
The difference you get in the scene appearance when reducing the textures size is exactly the same as reducing the texture quality from "High" to "Medium".
There is clearly a difference in the appearance when looking closer to the object, but depending on the object if you are far enough there is no difference.

In Webots:
![texture_1](https://user-images.githubusercontent.com/5910449/62025972-f486b800-b1d9-11e9-9e8c-464cc4e88c33.png)
![texture_2](https://user-images.githubusercontent.com/5910449/62025975-f5b7e500-b1d9-11e9-819a-b7603f25f1c7.png)

In Firefox:
![Screenshot from 2019-07-29 08-35-16](https://user-images.githubusercontent.com/5910449/62026665-ecc81300-b1db-11e9-99d3-32d4024ecd2e.png)



In Webots we have an option in the preferences to reduce the quality/resolution of the textures.
But at the moment we don't have any similar option for `webots.min.js` and even setting 
[...]

PR #751, opened 2019-07-25
Enhancement webots js textures

stefaniapedrazzi:
Addresses #749:

Improve scene load and rendering performance:
* [x] cache and reuse Texture objects when possible, i.e. when the texture transform, wrap etc. matches.
* [x] if textures are down scaled due to the texture quality settings then export the down scaled images to the web interface instead of the original ones
* [x] fix navigation in the scene not smooth on scene load -> disable frustum culling until all the textures are loaded
* [x] improve animation: automatically start it on page load only when all the textures are loaded to avoid slowing down the animation due to the texture computations and renderings.

~~Bug fix:~~
* [x] ~~fix issue due to GpuPicker not supporting undefined `scene.background` (change `undefined` to `null`)~~ _edit: fixed in #766_


stefaniapedrazzi:
In the `obstacle_avoidance` benchmark creating a single `THREE.Texture` instance, if the texture parameters matches, reduces the instances for the `VarnishedPine` PBRAppearance (4 textures with resolution 2048x2048) from 30 to just 5.

stefaniapedrazzi:
The rendering clearly takes long when new objects enter the viewpoint frustum (for example the windows and the professor table in the obstacle_avoidance benchmark).
If two objects share the same textures (for example the windows on the left and right wall), only the first object will slow down the scene navigation. Given that just the textures and not the geometries are reused it seems that the problem are again textures and materials setup.

I tried different solutions on scene load to try to  avoid this slow down when navigating into the scene:
* set the `Object3D.frustumCulling` of each object to true
* call `WebGLRenderer.compile()`
but on my system I don't see any difference.

omichel:
That's unfortunately another drawback of using Threejs versus using Wren.js. Resolving this problem in Wren.js would have been easier (e.g., similar to what we did in Webots to fix the problem). However, there should be a way in Threejs to force load the textures. On Webots/Wren, I recall we disabled the culling a forced one rendering, which caused Wren to load all the textures. Can't we do something similar in Threejs?

stefaniapedrazzi:
> However, there should be a way in Threejs to force load the textures. On Webots/Wren, I recall we disabled the culling a forced one rendering, which caused Wren to load all the textures. Can't we do something similar in Threejs?

This is what the `frustumCulling = true` should be supposed to do. But it doesn't solve the issue.

omichel:
That sounds odd, I would rather disable the frustum culling instead of enabling it to ensure all the objects are rendered.

stefaniapedrazzi:
> That sounds odd, I would rather disable the frustum culling instead of enabling it to ensure all the objects are rendered.

This is the way to disable/enable frustum culling on three.js:
https://github.com/mrdoob/three.js/issues/3949

omichel:
Yes, it should be set to `false`, not `true`, isn't it?

stefaniapedrazzi:
Yes. I wrote the wrong value in the doc.
But I tested with `false` and it doesn't solve the issue.

stefaniapedrazzi:
oops finally I had a typo in the code. That's working.

stefaniapedrazzi:
:+1:

fabienrohrer:
👀

resources/web/wwi/texture_loader.js
```diff
@@ -184,6 +265,43 @@ class _TextureLoaderObject {
...
       this.onTextureLoad();
+
+    this._evaluatePendingData();
+  }
+
+  _setTimeout() {
+    // Set texture loading timeout.
+    // If after some time no new textures are loaded, the hasPendingData variable is automatically
+    // reset to false in order to handle not found textures.
+    // The `this.loadingTextures` dictionary is not reset so that it is still possible to recore late textures.
```
omichel: What does "recore" mean? I guess it's a typo.

stefaniapedrazzi: yes. I don't remember what I meant, but `load` is fine.

PR #756, opened 2019-07-25
Fix the insertion of a PROTO containing a BallJoint

fabienrohrer:
Following this discussion: https://stackoverflow.com/questions/57142908/why-cant-i-insert-a-rotationalmotor-node-in-device-field-of-balljoint-node-in?noredirect=1#comment100870939_57142908

The `BallJoint` case and the `BallJoint.device3` case were not validated correctly in `WbNodeUtilities::validateExistingChildNode()` causing the impossibility to insert a PROTO containing a BallJoint.


fabienrohrer:
👍

fabienrohrer:
👍

PR #759, opened 2019-07-25
Improve OSM importer

fabienrohrer:
- [x] Support more cases for OSM lanes.forward/backward tags

### DIFF
--- a/resources/osm_importer/webots_objects/road.py
+++ b/resources/osm_importer/webots_objects/road.py
@@ -113,21 +113,27 @@ class Road(WebotsObject):
         if 'layer' in self.tags:
             self.layer = float(self.tags['layer'])
 
-        if 'lanes:forward' in self.tags:
-            self.forwardLanes = int(self.tags['lanes:forward'])
-            self.backwardLanes = self.lanes - self.forwardLanes
-            assert self.forwardLanes > 0
-            assert self.lanes - self.forwardLanes >= 0
+        try:
+            if 'lanes:forward' in self.tags:
+                self.forwardLanes = int(self.tags['lanes:forward'])
+                self.backwardLanes = self.lanes - self.forwardLanes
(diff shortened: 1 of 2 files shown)


resources/osm_importer/webots_objects/road.py
```diff
@@ -113,21 +113,27 @@ def parse_tags(self):
...
-            else:
-                self.forwardLanes = self.lanes - self.backwardLanes
-            assert self.backwardLanes > 0
-            assert self.lanes - self.backwardLanes >= 0
+        try:
+            if 'lanes:forward' in self.tags:
+                self.forwardLanes = int(self.tags['lanes:forward'])
+                self.backwardLanes = self.lanes - self.forwardLanes
+                assert self.forwardLanes > 0
+                assert self.lanes - self.forwardLanes >= 0
```
DavidMansolino: Are these assert still useful now that they are in a try-except ?

fabienrohrer: I think yes, because the try-except checks if it's an int, but the assert checks it's positive. However asserts are not ideal because they stop the execution of the program.

DavidMansolino: Yes indeed but will the assert will have an effect inside try-except ?

fabienrohrer: A Python `assert` throws an `AssertionError`, and not a `ValueError` which is catched here. So the behavior remain unchanged.

However, I think we should deal these 2 cases the same way, probably by not stopping the execution of the program, and display a simple warning rather.

DavidMansolino: Thank you for the explanation, you are completely right, I did indeed miss the `ValueError`.

fabienrohrer:
It has solved the user issue.

👍

PR #766, opened 2019-07-29
webots.min.js: Fix background color

fabienrohrer:
- [x] Fix JS crash in case there is no background node (cf. above to reproduce the bug)
- [x] Fix the default Background color (black in Webots, white in threejs)
- [x] Log

### To reproduce the bug

- Open `physically_based_rendering.wbt`
- Remove the background
- Open the browser client
- Get this error when picking:
        Capture d’écran 2019-07-29 à 15 34 17

Note:  `threejs` do the assumption that the background is null at several locations, for example [here](https://github.com/mrdoob/three.js/blob/r104/src/scenes/Scene.js#L31).

### DIFF
--- a/docs/reference/changelog-r2019.md
+++ b/docs/reference/changelog-r2019.md
@@ -14,6 +14,7 @@ Released on XXX YYth, 2019.
   - Bug fixes
     - Webots online 3D viewer (`webots.min.js`)
       - Fixed [ElevationGrid](elevationgrid.md) normals.
+      - Fixed [Background](background.md) default color.
     - Fixed `realistic_village.wbt` materials.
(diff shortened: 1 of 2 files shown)


fabienrohrer:
Default background:

Capture d’écran 2019-07-29 à 15 49 41

stefaniapedrazzi:
> Fix JS crash in case there is no background node (cf. above to reproduce the bug)

I also fixed this big here #751 : it was enough to replace `undefined` by `null` was enough.

fabienrohrer:
👍

stefaniapedrazzi:
:eyes:

PR #767, opened 2019-07-29
webots.min.js: Fix fog type

fabienrohrer:
- [x] Fix fog type which was blocked to `Linear` because of a typo
- [x] Log


fabienrohrer:
👍

PR #775, opened 2019-08-05
New HDR background

fabienrohrer:
I think we missed an HDR outdoor background take at noon with patterns in the sky (clouds) to increase reflection effects on the reflective materials.

This one seems to solve this issue:

city

Comparison on a car:

| HDR background | Result |
| --- | --- |
| `moutains` (default) | mountains |
| `noon_cloudy_countryside` (new) | noon_cloudy_countryside |


fabienrohrer:
👍

PR #787, opened 2019-08-07
Modal images dimension

fabienrohrer:
Currently, the modal images are limited with the doc images max-height = 500px.

- https://cyberbotics.com/doc/guide/samples-environments

When the image is big, If the window is too small, the image is cut, it is too big the image is limited to a height of 500px.
Setting explicitly the max dimension as a percentage of the parent modal window solves these 2 issues:

- https://cyberbotics.com/doc/guide/samples-environments?version=enhancement-modal-window-image-size


fabienrohrer:
👍

PR #789, opened 2019-08-07
Fix Various Code Issues

DavidMansolino:
All these issues are reported by DeepCodeAI (an AI that is scanning all the public Github repo and bug fixes to identify common errors and propose a fix: https://www.deepcode.ai)

I fixed most of the security/bug/performance issues with python. But there are many other less important or in other languages.


DavidMansolino:
:+1:

PR #790, opened 2019-08-07
fix-appartment-note

DavidMansolino:
Fix https://github.com/omichel/webots/pull/656#pullrequestreview-271861879


DavidMansolino:
:+1:

PR #799, opened 2019-08-12
visual_tracking benchmark: fix URI of camera/display images

stefaniapedrazzi:
Fix the image URI sent to the robot window: in python 3 `base64.b64encode` creates a  bytes literal object that converted to string has the format `b'<value>'`.
The additional `b` and quotes was making the URI invalid.


stefaniapedrazzi:
:+1:

PR #800, opened 2019-08-12
Fix read controller python shebang

omichel:
The python shebang of controller programs was totally ignored by Webots causing some misleading behaviors: although the shebang was indicating to run python 3, Webots may silently run Python 2.7. This PR fixes this problem by reading the standard shebang of controller programs to determine which version of python to use. On Windows, shebangs are not supported, so a warning is displayed in case of version mismatch.

### DIFF
--- a/src/webots/control/WbController.cpp
+++ b/src/webots/control/WbController.cpp
@@ -490,7 +490,7 @@ void WbController::setProcessEnvironment() {
     QFile pythonSourceFile(mControllerPath + name() + ".py");
     if (pythonSourceFile.open(QIODevice::ReadOnly)) {
       QTextStream in(&pythonSourceFile);
-      QString line = in.readLine();
+      const QString &line = in.readLine();
       if (line.startsWith("#!")) {
 #ifndef _WIN32
         if (line.startsWith("#!/usr/bin/env "))
--- a/src/webots/control/WbController.cpp
+++ b/src/webots/control/WbController.cpp
@@ -485,6 +485,33 @@ void WbController::setProcessEnvironment() {
     if (mPythonCommand.isEmpty())
       mPythonCommand = WbLanguageTools::pythonCommand(
         mPythonShortVersion, WbPreferences::instance()->value("General/pythonCommand", "python").toString());
+    // read the python shebang (first line starting with #!) to possibly override the python command
+    QString expectedVersion;
+    QFile pythonSourceFile(mControllerPath + name() + ".py");
+    if (pythonSourceFile.open(QIODevice::ReadOnly)) {
+      QTextStream in(&pythonSourceFile);
+      QString line = in.readLine();
+      if (line.startsWith("#!")) {
+#ifndef _WIN32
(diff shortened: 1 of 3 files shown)


omichel:
👍

omichel:
👍

stefaniapedrazzi:
:eyes:

src/webots/control/WbController.cpp
```diff
@@ -485,6 +485,33 @@ void WbController::setProcessEnvironment() {
     if (mPythonCommand.isEmpty())
       mPythonCommand = WbLanguageTools::pythonCommand(
         mPythonShortVersion, WbPreferences::instance()->value("General/pythonCommand", "python").toString());
+    // read the python shebang (first line starting with #!) to possibly override the python command
+    QString expectedVersion;
+    QFile pythonSourceFile(mControllerPath + name() + ".py");
+    if (pythonSourceFile.open(QIODevice::ReadOnly)) {
+      QTextStream in(&pythonSourceFile);
+      QString line = in.readLine();
```
stefaniapedrazzi: `line` can be const
```suggestion
      const QString &line = in.readLine();
```

src/webots/control/WbController.cpp
```diff
@@ -485,6 +485,33 @@ void WbController::setProcessEnvironment() {
...
+#else  // Windows: check that the version specified in the shebang corresponds to the version of Python installed
+        expectedVersion = line.mid(line.lastIndexOf("python", -1, Qt::CaseInsensitive) + 6);
+        bool mismatch = false;
+        int l = expectedVersion.length();
+        if (l == 1 && expectedVersion[0] != mPythonShortVersion[0])
+          mismatch = true;
+        if (l >= 3 && (expectedVersion[0] != mPythonShortVersion[0] || expectedVersion[2] != mPythonShortVersion[1]))
+          mismatch = true;
+        if (mismatch)
+          warn(tr("Python shebang requests python%1, but current path points to Python%2")
```
stefaniapedrazzi: first "python" is written with lower case p and the second with upper case P
```suggestion
          warn(tr("Python shebang requests python%1, but current path points to python%2")
```

src/webots/control/WbController.cpp
```diff
@@ -485,6 +485,33 @@ void WbController::setProcessEnvironment() {
     if (mPythonCommand.isEmpty())
       mPythonCommand = WbLanguageTools::pythonCommand(
         mPythonShortVersion, WbPreferences::instance()->value("General/pythonCommand", "python").toString());
+    // read the python shebang (first line starting with #!) to possibly override the python command
+    QString expectedVersion;
```
stefaniapedrazzi: This is only used in the Windows case, couldn't this declaration be moved to line 501?

src/webots/control/WbController.cpp
omichel: That was on purpose because the shebang uses all lowercase letter but the Windows installation folder is usually named "Python27" or "Python37".

omichel:
👍

PR #802, opened 2019-08-13
Fixed gitignore files for robotis python libraries

omichel:
👍


DavidMansolino:
I know that it is not striclty usefull because they are not generated on Windows but we should probably apply this fix for python 3.5 and 3.6 too, this will simply the maintenance to have identic gitignore.

PR #811, opened 2019-08-15
Fix Python37 not released on Linux

DavidMansolino:
We document that Python37 is available on Linux (https://www.cyberbotics.com/doc/guide/using-python#introduction), but that was not actually the case.


DavidMansolino:
:+1:

PR #812, opened 2019-08-16
Improve Vehicle Light Intensity 

DavidMansolino:
Example in the night:
city_night

Example with the indicators:
city

ezgif-4-6e99c8d069f9


DavidMansolino:
:+1:

PR #813, opened 2019-08-16
Fix possible dll conflict on windows

omichel:
Following a user bug report (support ticket 2649), this fix guarantees that we use the correct version of libssl on Windows.


omichel:
👍

PR #815, opened 2019-08-20
webots.min.js: fix 'this' object in static methods

stefaniapedrazzi:
Remove calls  to `this` object in static methods of `dialog_window.js`.
This was causing issues when calling `webots.alert` and `webots.confirm` functions.


stefaniapedrazzi:
:+1:

PR #817, opened 2019-08-20
webots.min.js: Improve history management of the console

fabienrohrer:
Fix https://github.com/omichel/robotbenchmark/issues/56

- [x] Add an upper limit to the console history (100 logs)
- [x] Add a clear button

Result (with an history limit of 5):

Capture d’écran 2019-08-20 à 10 24 43


fabienrohrer:
👍

stefaniapedrazzi:
:eyes:

stefaniapedrazzi:
It works fine.

But I'm wondering if moving the clear button on the window title/tool bar (next to the maximize button) or in a fixed position (like the right-bottom corner) wouldn't be more clean.

With this solution I personally feel that there are some weird behaviors:
* the clear button moves
* if you scroll up to the beginning of console text, the clear button is not visible.

I also think that just the icon with a tooltip is enough and the text "Clear" is not needed.
We don't have any other button with icon+text in the interface, but just icons or just text (in the editor menu).

fabienrohrer:
I had initially the same feeling and tried to add the button in the titlebar or in a jQuery footer. Every solutions I found was about to hack jQuery to force the button or footer creation after the dialog load. This was producing at best some code which is difficult to maintain and which is very dependent on the jQuery version. We should not go in this direction.

I will now give a second try about to create a custom menu tool bar. To be consistent with the editor, it may be better to add it on top of the dialog, and with an icon only as you suggest.

fabienrohrer:
I finally choose the solution about to mimics the editor GUI:

Capture d’écran 2019-08-20 à 14 00 14

fabienrohrer:
👍

stefaniapedrazzi:
:eyes:

fabienrohrer:
I just added a log entree for this.

👍

PR #824, opened 2019-08-21
Add the fall penality only once

fabienrohrer:
Fix https://github.com/omichel/robotbenchmark/issues/327

When a the robot falls, the 30s time penalty is added every milliseconds. Soon enough, the timer (supporting 100 minutes) is overloaded causing weird results.

Adding the penalty only once fix this issue.


fabienrohrer:
👍

stefaniapedrazzi:
I'm not sure that adding the penalty only once will be fair and will prevent cheating: a user could make the robot fall to exploit instabilities and jump to a further location.

What do you think?
What about setting a maximum time to 100 minutes?

fabienrohrer:
Yes, setting the penalty to the maximum amount of time is also a solution.

fabienrohrer:
In this case, the penaliy becomes more a disqualification criteria. I'm changing this PR target to develop.

fabienrohrer:
Migrated to #829

PR #826, opened 2019-08-21
Update naoqisim repo path

DavidMansolino:
Prepare the naoqisim migration to cyberbotics organization


DavidMansolino:
:+1:

DavidMansolino:
(The migration should be done before merging this PR, but this doesn't prevent approving this PR)

PR #830, opened 2019-08-21
Fixed the installation of npm on Linux

omichel:
This PR makes that travis install will the `npm` package the same was we do it manually following the new explanations here: https://github.com/omichel/webots/wiki/Linux-installation.

It is simpler and greatly facilitates the build on snapcraft.io.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,20 +83,21 @@ jobs:
             - libfreetype6-dev
             - libxml2-dev
             - libzzip-0-13
-            - libssl-dev
+            - libssl1.0-dev
             - libboost-dev
             - libjpeg8-dev
             - libavcodec-extra
             - libpci-dev
             - libgd-dev
             - libtiff5-dev
-            - libssh-dev
+            - libssh-gcrypt-dev
             - libzip-dev
             - python-pip
             - libreadline-dev
             - libassimp-dev
             - libpng12-dev
             - libav-tools
+            - npm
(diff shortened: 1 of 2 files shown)


omichel:
:+1:

DavidMansolino:
(I just started the test including the compilation process)

omichel:
Apparently ubuntu 16 (on travis) only has `libssl-dev`, but on ubuntu 16, `libssl-dev` doesn't conflict with `npm` which is fine this way.

👍

DavidMansolino:
Ok thank you for the test. We should then also update: https://github.com/omichel/webots/wiki/Linux-installation

omichel:
Correct. This is done now.
👍

fabienrohrer:
👀

.travis.yml
```diff
@@ -90,13 +90,14 @@ jobs:
             - libpci-dev
             - libgd-dev
             - libtiff5-dev
-            - libssh-dev
+            - libssh-gcrypt-dev
```
fabienrohrer: I don't understand why we need only `libssh-gcrypt-dev` here and we ask users to install `libssh-dev` in the wiki. Isn't `libssh-gcrypt-dev` sufficient for every cases?

fabienrohrer: Forget my comment `libssh-gcrypt-dev` is mentioned in the wiki.

PR #831, opened 2019-08-22
webots.min.js: support texture urls from jsdelivr

fabienrohrer:
Aim: improve browser cache usage in-between simulations thanks to the use of jsdelivr for the texture assets inside Webots.


fabienrohrer:
This should be resolved in a more ambitious way. To have an asset/resource online server.

PR #843, opened 2019-08-30
Add missing setMode function to doc

DavidMansolino:
This function was not documented for Python, Java and Matlab.


DavidMansolino:
👍

PR #848, opened 2019-09-02
Fix path to python versions of robotis managers

stefaniapedrazzi:
Fix issue https://github.com/omichel/robotbenchmark/issues/349: to support multiple versions of python we now also have to link to the specific python folder other than the generic 'python' folder kept for backward compatibility.


stefaniapedrazzi:
:+1:

PR #855, opened 2019-09-03
Feature snapcraft build

omichel:
Modifications needed in Webots to allow us to publish a snap package hopefully with strict confinement.


omichel:
I confirm that using POSIX shared memory on macOS is a very good solution to work around the SYSV shared memory limits we have been experiencing so far. I will port the POSIX shared memory to macOS as well in this PR since I already did it for Linux and it works like a charm.

fabienrohrer:
I tested the POSIX shared memory on macOS and it seems to work well.

omichel:
Thank you for the tests. I will update the documentation accordingly.

omichel:
This PR is the first milestone aiming at preparing Webots for a snap distribution. It was successfully tested with the current snap (currently available on the candidate channel). It allows to build and publish Webots as a snap with strict confinement. Some issues remain as listed in #912, but they should be addressed in subsequent PRs on https://github.com/cyberbotics/webots-snap and on this repository.

👍

src/webots/Makefile
```diff
@@ -482,6 +482,7 @@ OTHER_SOURCES = \
   WbControllerPlugin.cpp \
   WbDataPacket.cpp \
   WbDevice.cpp \
+	WbDesktopServices.cpp \
```
stefaniapedrazzi: Small indentation issue:
```suggestion
  WbDesktopServices.cpp \
```

stefaniapedrazzi:
I get an error running `webots_distro.c` on Linux:
```
ln: failed to create symbolic link 'libopenal.so.1': File exists
```

The following line has to be removed given that the file `libopenal.so.1` is now already copied from the dependencies package.
https://github.com/cyberbotics/webots/blob/e837ffa64a7361706a2d183e0763466ae931ddca/src/packaging/webots_distro.c#L1112

stefaniapedrazzi:
There are two nested pre-processor `#ifdef __WIN32` conditions.
The inner one should be removed:
https://github.com/cyberbotics/webots/blob/e837ffa64a7361706a2d183e0763466ae931ddca/src/webots/core/WbSysInfo.cpp#L55

src/webots/core/WbDesktopServices.cpp
```diff
@@ -0,0 +1,31 @@
...
+#endif
+
+bool WbDesktopServices::openUrl(const QString &url) {
+#ifdef __linux__
+  QProcess process;
+  process.setProgram("xdg-open");  // we need to use xdg-open to be snap compliant
+  process.setArguments(QStringList() << url);
+  return process.startDetached();
+#else
+  return QDesktopServices::openUrl(url);
```
stefaniapedrazzi: Compilation error on Windows: `QDesktopServices::openUrl` expects a `const QUrl &` argument and not a `const QString &` argument.

probably the best solution would be to change it to:
```suggestion
  return QDesktopServices::openUrl(QUrl(url));
```
and to fix the calls of WbDesktopServices::openUrl() to remove when possible the duplicated `openUrl(QUrl(<string>).toString())` instructions.
For example in https://github.com/cyberbotics/webots/blob/e837ffa64a7361706a2d183e0763466ae931ddca/src/webots/core/WbFileUtil.cpp#L273

omichel: `WbDesktopServices::openUrl()` should receive a `QString` argument to work with the `connect(label, &QLabel::linkActivated, &WbDesktopServices::openUrl);` connections. So we should probably keep it like that (or create new method only for the connect callbacks that would convert `QString` to `QUrl` and call WbDesktopServices::openUrl()`.

stefaniapedrazzi: Yes, I don't suggested to change the `WbDesktopServices::openUrl()` function, but to change when possibible the use of this function from `openUrl(QUrl(<string>).toString())` to ``openUrl(<string>))

stefaniapedrazzi: Ok, I see what do you mean.
So then I would at least remove a duplicated interpretation of the url:

```suggestion
  return QDesktopServices::openUrl(QUrl(url, QUrl::StrictMode));
```

omichel: For some reason, I cannot commit your suggestion from this page (the popup menu is disabled). I will commit from git.

stefaniapedrazzi: It seems it also needed to include `QtCore/QUrl`.

omichel: All this should be fixed with my last commit.

omichel:
I checked that the URLs, movies and screenshots are opening correctly under Linux (tested on Ubuntu 18.04).

👍

omichel:
I will :wink:

omichel:
Unfortunately, the unit test of travis is currently broken, so it is likely to fail. However, @DavidMansolino may have found the problem. So let's wait a little bit...

omichel:
OK, I am running the travis_unit test with the patch of @DavidMansolino...

omichel:
✔️ 👍 (hehe)

omichel:
👍

stefaniapedrazzi:
:eyes:

src/packaging/files_core.txt
```diff
@@ -157,7 +157,7 @@ lib/qt/plugins/platforminputcontexts/libcomposeplatforminputcontextplugin.so [li
 lib/qt/plugins/platforminputcontexts/libibusplatforminputcontextplugin.so [linux]
 lib/qt/plugins/platforms/ [linux,mac]
 lib/qt/plugins/platforms/libqcocoa.dylib [mac]
-lib/qt/plugins/platforms/libqxcb.so [linux]
+lib/qt/plugins/platforms/lib*.so [linux]
```
DavidMansolino: Is there an updated version of the snap that I can test on fedora?

omichel: It's coming...

omichel:
For the record, I updated the Qt-5.13.1 dependency package for Linux to add `lib/qt/plugins/platforms/libqwayland-egl.so` which was missing there.

src/webots/core/WbDesktopServices.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include <QtCore/QUrl>
+#include <QtGui/QDesktopServices>
+#endif
+
+bool WbDesktopServices::openUrl(const QString &url) {
+#ifdef __linux__
+  QProcess process;
+  process.setProgram("xdg-open");  // we need to use xdg-open to be snap compliant
+  process.setArguments(QStringList() << url);
+  return process.startDetached();
```
DavidMansolino: Is there a reason for not using simply the static member?
```
return  QProcess::startDetached("xdg-open", QStringList() << url)
```

omichel: No, that would be simpler indeed.

PR #856, opened 2019-09-04
Robotbenchmark pause simulation on complete

stefaniapedrazzi:
Related to https://github.com/omichel/robotbenchmark/pull/352.
* Pause the simulation from benchmark supervisor when the benchmark is completed:
    * [x] update all the benchmark supervisor controllers
    * [x] add close callback to webots.confirm dialog 
    * [x] use a different message to notify `webots.min.js` client if the simulation is paused by Webots/controller or by the client itself 

* Synchronize files with robotbenchmark repo


stefaniapedrazzi:
:+1:

PR #861, opened 2019-09-05
Fix handling mouse wheel event while running another mouse event

stefaniapedrazzi:
Fix issue https://github.com/omichel/robotbenchmark/issues/351:
the wheel event was overwriting the `MouseEvents.moveParams` and deleting the `scaleFactor` variable. Due to this when moving the mouse the computed `zoomScale` value was NaN and the viewpoint position was broken.

Steps to reproduce:
1. start tilting/zooming the viewpoint by pressing the middle/wheel button
2. move the wheel (without releasing the wheel button)
3. move the mouse (without releasing the wheel button)


stefaniapedrazzi:
:+1:

fabienrohrer:
👀

PR #863, opened 2019-09-06
Improved life cycle of the Webots temporary folder

### LINKED ISSUE #859: The temp folder is not cleared properly
Webots uses a temporary folder to store several files while running.
When Webots quits normally, it deletes its own temporary folder.
However, if it crashes (or the the system crashes or is rebooted), this temporary folder is not deleted.
So, many temporary folders can therefore accumulate and saturate the disk over time.
To work around this, we implemented this quick and dirty trick: in the geter of the temp folder, we check if only one instance of Webots is running. If this is the case, then we delete all the webots temp folders except for the current one. This is not good because in case we are on a simulation server, it is very unlikely that only one instance of Webots is running at a time. Therefore the temp folder will never get cleaned-up.

A better approach would be the following: when we start Webots, we check all the webots temp folder and we delete the ones which are older than one hour. We should also add a timer in Webots that will touch its own temp folder every hour to prevent another instance of Webots from deleting it.

Such a mechanism has two advantages on the previous one:

1. It removes the need to determine if only one instance of Webots is running. We achieved this by a complicated non cross-platform approach which causes confinement issues with snap.
2. It will be more efficient on simulation servers where the cleanup could occur even if several instances of Webots are running.

Please let me know what do you think about this proposal and if you agree, I will implement it.

omichel:
Fixes #859.


omichel:
👍

DavidMansolino:
:eyes:

src/webots/core/WbStandardPaths.cpp
```diff
@@ -205,20 +216,18 @@ const QString &WbStandardPaths::webotsTmpPath() {
...
-      filters << "webots-*";
-      tmpDir.setNameFilters(filters);
-
-      QFileInfoList list = tmpDir.entryInfoList();
-      for (int i = 0; i < list.size(); ++i) {
-        QDir dirToRemove(list.at(i).absoluteFilePath());
-        dirToRemove.removeRecursively();
+    // cleanup old and unused tmp directories
+    const QString &tmpRoot = webotsTmpPath.left(webotsTmpPath.chopped(1).lastIndexOf('/'));
+    const QDir directory(tmpRoot);
```
DavidMansolino: This part looks a bit too 'C' style to me :wink: 

```
    QDir directory(webotsTmpPath);
    directory.cdUp();
```

Seems simpler and clearer.

src/webots/core/WbStandardPaths.cpp
```diff
@@ -232,6 +241,11 @@ const QString &WbStandardPaths::webotsTmpPath() {
         WbLog::fatal(QObject::tr("Cannot create a directory in the Webots temporary directory \"%1\"").arg(webotsTmpPath));
 #endif
     }
+
+    // write a new live.txt file in the webots tmp folder every hour to prevent any other webots process to delete it
+    static QTimer timer;
+    QTimer::connect(&timer, &QTimer::timeout, liveWebotsTmpPath);
+    timer.start(3600 * 1000);
```
DavidMansolino: I would update it every 30min, having the exact same update and time threshold seems dangerous.

omichel:
Thank you for the review. I took into account your comments.
:+1:

DavidMansolino:
:eyes:

PR #865, opened 2019-09-09
Fix regular expression pattern for Solid.name

### LINKED ISSUE #841: SFString Field Starting with '*' Trigger Qt Warning
**Describe the Bug**
If a field start with a `*` the following warning is displayed in the MSYS console:
```
Warning: QRegularExpressionPrivate::doMatch(): called on an invalid QRegularExpression object
```

**Steps to Reproduce**
1. Open the ``khepera3_gripper.wbt`` world.
2. Select a box and add `*` at the beginning of the name field.
3. Save and quit Webots.
4. Re-open Webots => The warnign is displayed in the console

**Expected behavior**
Not sure if we want to deny the use of ``*`` at start, but probably it would be better to remove this warning.

**System**

stefaniapedrazzi:
Fix #841: escape characters in Solid.name that have special meaning when used as a regular expression pattern string.


stefaniapedrazzi:
:+1:

DavidMansolino:
:eyes:

PR #872, opened 2019-09-10
Use POSIX shared memory instead of SYSV shared memory on Linux and macOS

omichel:
By default Qt uses SYSV shared memory on Linux and macOS. This has several drawbacks:
- on macOS: the default size of SYSV shared memory segments is very small preventing many camera-based simulation to work out-of-the-box. We ask users to modify their system settings to work around this problem, which is not very user friendly.
- on Linux: the SYSV shared memory segments are not compatible with snap strict confinement, hence preventing us to package Webots as a snap package with strict confinement (which is better than the classic confinement).

### DIFF
--- a/docs/reference/changelog-r2019.md
+++ b/docs/reference/changelog-r2019.md
@@ -15,8 +15,6 @@ Released on XXX YYth, 2019.
   - New Samples
     - Added new samples about the [Accelerometer](../guide/samples-devices.md#accelerometer-wbt) and [Brake](../guide/samples-devices.md#brake-wbt) devices.
     - Added a `village_center` world.
-  - Dependency Updates
-    - Windows: upgraded to Qt 5.12.4.
   - Enhancements
     - Improved the intensity and color of the bus, truck and car vehicle lights.
     - macOS and Linux: Use POSIX shared memory segments to fix the limit problems on macOS and allow for snap packaging with strict confinement on Linux.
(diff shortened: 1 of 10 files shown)


omichel:
:+1:

fabienrohrer:
👀

fabienrohrer:
Unfortunately, every camera are failing systematically, even if I clean the shared memory before (`./src/bin/clean_shm.sh`).

Capture d’écran 2019-09-10 à 09 51 13

Do you want to go ahead with this on Mac?

docs/reference/changelog-r2019.md
```diff
@@ -17,6 +17,7 @@ Released on XXX YYth, 2019.
     - Added a `village_center` world.
   - Enhancements
     - Improved the intensity and color of the bus, truck and car vehicle lights.
+    - macOS and Linux: Use POSIX shared memory segments to fix the limit problems on macOS and allow for snap packaging with strict confinement on Linux.
```
fabienrohrer: I'm wondering if mentioning here `snap` is useful, it seems to be unrelated.

omichel:
Oops. I tested only on Linux and it works fine. Unfortunately, I don't have a development environment ready on macOS, so it may take a while for me test investigate on macOS. Can I let you check it? I believe I didn't change anything from #855 which you tested successfully.

docs/reference/changelog-r2019.md
omichel: It's related. Without POSIX shared memory, there is no way to create a strictly confined snap.

fabienrohrer:
Just fixed an issue on macOS. Everything seems to work well now.

omichel:
👍

PR #877, opened 2019-09-11
webots.min.js: hotfix single node load into existing scene

### LINKED ISSUE #857: webots.min.js: grass becomes white on world reset in highway_driving
After resetting the highway_driving world the grass texture disappears:
Screenshot from 2019-09-04 09-44-34

stefaniapedrazzi:
Fix #857:
* [x] fix streaming a PBRAppearance node insertion event: PBRAppearance it is currently exported as two siblings nodes for X3D compatibility (Appearance and PBRAppearance) but XML parser doesn't support multiple root nodes
* [x] fix loading new geometry and appearance (and TextureTransform) nodes into an existing three.js scene

### DIFF
--- a/resources/web/wwi/texture_loader.js
+++ b/resources/web/wwi/texture_loader.js
@@ -15,6 +15,10 @@ var TextureLoader = {
     return new THREE.Texture();
   },
 
+  applyTextureTransform: function(texture, transformData) {
+    this._getInstance().applyTextureTransform(texture, transformData);
+  },
+
   createColoredCubeTexture: function(color, width = 1, height = 1) {
     // Create an off-screen canvas.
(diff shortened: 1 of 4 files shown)


stefaniapedrazzi:
Because of the different nodes structure in three.js and Webots/x3d for some nodes it is not straightforward to correctly insert/remove correctly them.

These are the other nodes that are not handled:  `ImageTexture`, `Coordinate`, `TextureCoordinate`, `Normal`, `Color`,  `Material`.
As soon as this is not causing issues in our simulations I would keep it like this for the moment.

stefaniapedrazzi:
:+1:

fabienrohrer:
👀

Issue #834, opened 2019-08-27
Performance issues in HDR backgrounds


fabienrohrer:
### Main observation

Since the migration to the equirectangular HDR backgrounds, we noticed a significant performance drop in Webots and in streaming.

To illustrate this, the time to (re)load `humanoid_sprint.wbt` in Webots (develop/release) is:

- `4341 ms` with an equirectangular HDR backgrounds
- `2746 ms` with a cubic PNG background (with a comparable resolution) => -40%

This result depends on the GPU quality, so it is certainly even worst on low-end computers.

The current pipeline in case of an HDR background consumes lot of memory (typically `8196x4096x4 = 150 Mo` per background copy), CPU and GPU time:

```
+-Webots---------------------------------------------------------+
|                                                                |
| +-1-----------------+      +-6-------+      +-2-------------+  |
| |Equirectangular HDR| +--> |Cubic HDR| +--> |Irradiance Maps|  |
| +--------+--CPU/GPU-+      +-----GPU-+      +-----------GPU-+  |
|          |                                                     |
+----------------------------------------------------------------+
           |
       Download
       (cached)
           |      +-webots.min.js--------------------------------------------+
           |      |                                                          |
           |      | +-1-----------------+      +-6-------+      +-1-------+  |
           +------->+Equirectangular HDR| +--> |Cubic HDR| +--> |PMREM Map|  |
                  | +-----------CPU/GPU-+      +-----GPU-+      +-----GPU-+  |
                  |                                                          |
                  +----------------------------------------------------------+
```

### Possible actions

The following actions could prevent us to do several operations at load (in Webots and in webots.min.js):
- Dropping the equirectangular feature
- Precomputing the irradiance or PMREM maps
- Merging at best the irradiance or PMREM maps and share them between Webots and webots.min.js

#### Possible action plan

- [x] Step 1: Remove the equirectangular feature
    - [x] Drop the support of the equirectangular cubemap
    - [x] Drop the CubeMap node, and use the standard `VRML Background *Url` fields: http://lighthouse3d.com/vrml/tutorial/index.shtml?back
    - [x] Split our equirectangular HDR backgrounds to cubic HDR textures.
- [x] Step 2: Precompute and merge the irradiance maps (cf. #505)
    - [x] Precompute the irradiance map(s)
    - [x] Merge them if possible
    - [x] Share them with webots.min.js
- [x] Step 3: Release all our backgrounds as cubic HDR textures with irradiance maps.

Background node after this plan:

```
[...]

PR #880, opened 2019-09-11
Precompute irradiance maps

fabienrohrer:
Step 2 of #834

- [x] Precompute irradiance maps
    - [x] Script to precompute maps
        - [x] Diffuse maps
        - [x] Specular maps
    - [x] Modify Background to support irradiance maps as input
- [x] Convert existing backgrounds
- [x] Use the irradiance maps in `webots.min.js`
- [x] Doc
- [x] Log


fabienrohrer:
Current status of the diffuse irradiance maps export:

| ID (Webots / OpenGL) | Webots (apitrace) | script |
| --- | --- | --- |
| right / +x | mountains_diffuse_positive_x | mountains_right dm |
| left / -x | mountains_diffuse_negative_x | mountains_left dm |
| top / +y | mountains_diffuse_positive_y | mountains_top dm |
| bottom / -y | mountains_diffuse_negative_y | mountains_bottom dm |
| front / +z | mountains_diffuse_positive_z | mountains_front dm |
| back / -z | mountains_diffuse_negative_z | mountains_back dm |

- results are very close but still not perfect. The specular part (in good progress) will certainly allow me to fix the remaining bugs.

fabienrohrer:
I master now the pipeline and can push my precomputed irradiance maps in Webots:

pick_and_place

Currently I generate the first specular irradiance map using a Python script, and then I use automatic mipmap generation to create roughness LOD levels (rather than using a shader to "blur" them)

comparison

But this blurness is required. Otherwise roughness does not act realistically (to shiny as in the screenshot above).

I think that keeping the baking but starting it from the precomputed HDR background is the way to go.

Probable facts (but to be determined after trial):
- The initial precomputed HDR background could be simply a downscaled version of the HDR background.
- I suspect that the precomputed diffuse map could simply be one of the last LOD of the baked specular map (rendering its precomputing useless, which matches with what `threejs` does)

fabienrohrer:
I can now use a single irradiance map for diffuse and specular reflections with the same quality level as the one we have in Webots. The diffuse light changes a bit, but I think it's minor and justifiable.

In Webots, the background load (including irradiance map load, baking, and appliance on reflections) has been reduced from 924ms to 320ms (-70%). The background size has been reduced from `6 x 1.04` Mo to `6 x (0.3 + 0.26)` Mo (-45%).

Even with 98% of quality, the JPEG artifacts are too much visible on backgrounds to be acceptable. This is mainly because these images are often stretched (from 512x512 to the view3d resolution), and because gamma if applied. I prefer PNG for these files. Here is some comparison:

| JPEG (98%) | PNG |
| --- | --- |
| jpeg 98 | png |

On the Webots side, I'm still not satisfied with the gamma management. For an obscure reason, I need to apply gamma in the precomputed backgrounds. To be fixed.

fabienrohrer:
Forgot the screenshot :-)

pick_and_place_1

fabienrohrer:
I found a satisfying solution for the background gamma: to deal with it in the shader which renders the background.

fabienrohrer:
👍

src/webots/nodes/WbBackground.cpp
```diff
@@ -323,101 +321,113 @@ void WbBackground::applySkyBoxToWren() {
...
+
+    mIrradianceCubeTexture =
+      wr_texture_cubemap_bake_specular_irradiance(cm, WbWrenShaders::iblSpecularIrradianceBakingShader(), w);
+    wr_texture_cubemap_disable_automatic_mip_map_generation(mIrradianceCubeTexture);
+  } else {
+    wr_texture_delete(WR_TEXTURE(mIrradianceCubeTexture));
+    mIrradianceCubeTexture = NULL;
+
+    // Fallback: a cubemap is found but no irradiance map: bake a small irradiance map to have right colors.
+    // Reflections won't be good in such case.
```
DavidMansolino: This is a nice backup and works surprisingly not so bad.

resources/wren/shaders/pbr_stencil_ambient_emissive.frag
```diff
@@ -75,12 +75,13 @@ vec3 getIBLContribution(IBLInfo iblInputs, vec3 n, vec3 reflection) {
...
   vec2 brdf = texture(inputTextures[5], vec2(iblInputs.NdotV, iblInputs.perceptualRoughness)).rg;
-  if (material.cubeTextureFlags.x > 0.0 && material.cubeTextureFlags.y > 0.0) {
+  if (material.cubeTextureFlags.x > 0.0) {
     float mipCount = 7.0;
     float lod = (iblInputs.perceptualRoughness * mipCount);
+    // A single irradiance map is used for the diffuse and specular reflections.
     // invert z components of sample vectors due to VRML default camera orientation looking towards -z
-    diffuseLight = texture(cubeTextures[0], vec3(n.xy, -n.z)).rgb;
-    specularLight = textureLod(cubeTextures[1], vec3(reflection.xy, -reflection.z), lod).rgb;
+    diffuseLight = textureLod(cubeTextures[0], vec3(n.xy, -n.z), 4).rgb;
```
DavidMansolino: Is the `4` empirical or is there any reason behind?

DavidMansolino:
It was so obvious that I did forget to mention it, but the source tests are not passing (due to duplicated textures) and it would be nice to launch the test suite too.

fabienrohrer:
@DavidMansolino I think I addressed all the issues you reported here.

fabienrohrer:
The pros/cons balance of this PR is quite equilibrated. It should be accepted informed.

### PROs

- the Webots and webots.min.js loading performance gain (as described above)
- the reduced memory footprint (on disk, in Webots and in webots.min.js)
- this solution is certainly more friendly with low-end GPUs.
- the previous diffuse reflections were inconsistent on top and on bottom of the cubemap (a kind of shadow artifact is present). This PR solves this.

### CONs

- Background user should define 2 maps instead of 1. (This is attenuated by the fact that our backgrounds are generally encapsulated, and by the fact that this architecture gives more flexibility)
- The Bloom post-processing is not available anymore in the background, because now it contains only values between 0 and 1.
- I cannot guarantee that the diffuse reflections are still physically based (but as mentioned in the pros list, the previous diffuse maps are also certainly buggy.)

fabienrohrer:
👍

DavidMansolino:
:eyes:

src/webots/nodes/WbBackground.cpp
```diff
@@ -311,113 +277,139 @@ void WbBackground::applySkyBoxToWren() {
...
+      QSize textureSize = imageReader.size();
 
-      bool alpha = false;
-      for (int i = 0; i < 6; i++) {
-        QImageReader imageReader(textureUrls[i]);
-        QSize textureSize = imageReader.size();
+      if (textureSize.width() != textureSize.height())
+        throw tr("The texture '%1' is not a square image (its width doesn't equal its height).").arg(imageReader.fileName());
+      if (i > 0 && textureSize.width() != edgeLength)
+        throw tr("Texture dimension mismatch between '%1' and '%2'").arg(lastFile).arg(imageReader.fileName());
```
DavidMansolino: Are those checks really usefull?
I don't see any good reason for keeping this constraint.

projects/objects/backgrounds/protos/TexturedBackground.proto
```diff
@@ -38,7 +38,7 @@ PROTO TexturedBackground [
       texture == "noon_cloudy_countryside" or
       texture == "noon_park_empty" or
       texture == "mountains"
-    local extension = isHDR and "hdr" or "jpg"
+    local extension = isHDR and "png" or "jpg"
     local path = "textures/cubic"
   }%
   Background {
```
DavidMansolino: Now that we do support reflections without visual background it would be interesting to add a new field to specify if the *Url fields should be set or not (but if you prefeer I can do this in another PR).

src/webots/nodes/WbBackground.cpp
fabienrohrer: Yes, they are useful. Without, it looks like this 😉 :

physically_based_rendering

Wren loads these images as a cubemap which offers useful accessor abilities, but also this constraint about dimension.

projects/objects/backgrounds/protos/TexturedBackground.proto
fabienrohrer: This is a good idea. But as you tell, this is a new feature that I prefer to implement in another PR.

fabienrohrer:
👍 @stefaniapedrazzi or @omichel Could you review this PR, focusing on the web part, please?

stefaniapedrazzi:
Yes, I can review this PR.
I will try to do it tomorrow.

PR #884, opened 2019-09-12
Improve linux launcher termination

stefaniapedrazzi:
Related to https://github.com/omichel/robotbenchmark/issues/348:
in the streaming server application it happens that Webots launcher is killed but not the `webots-bin` process.
In the current version the termination and interruption signals are already caught and propagated to `webots-bin`, but with the current workflow it is possible that the launcher exits before the `webots-bin` child process terminates (see articles referenced just below). 

Resources:
- https://unix.stackexchange.com/questions/146756/forward-sigterm-to-child-in-bash/444676#444676
- http://veithen.io/2014/11/16/sigterm-propagation.html


stefaniapedrazzi:
I Installed the new launcher on the beta.robotbenchmark.net and I will continue testing it for a while to check if it really solves the issue.

I also added a Python script to cleanup the servers, kill old Webots instances and notify about orphan "webot-bin" processes.

stefaniapedrazzi:
It happened a couple of times today that the simulation_server.py script blocked waiting for a Webots instance termination. I added a timeout to `Popen.wait` in `simulation_server.py` to avoid it but it seems  that sometimes there are still issues killing Webots.

stefaniapedrazzi:
It seems to work much better than the previous version.
Probably I'm the only one testing `beta.robotbenchmark.net` in the last days/weeks and the server has been restarted a couple of times since this version of the scripts is running, but currently there aren't any ghost webots processes coming from the beta.

:+1:

resources/web/server/simulation_server.py
```diff
@@ -243,11 +243,16 @@ def kill_webots(self):
...
-            self.webots_process.terminate()
-            self.webots_process.wait()
+            if sys.platform == 'darwin':
+                self.webots_process.kill()
+            else:
+                self.webots_process.terminate()
+                try:
+                    self.webots_process.wait(5)  # set a timeout (seconds) to avoid blocking the whole script
+                except subprocess.TimeoutExpired:
+                    logging.warning('[%d] ERROR killing Webots [%d' % (id(self), self.webots_process.pid))
```
omichel:
```suggestion
                    logging.warning('[%d] ERROR killing Webots [%d]' % (id(self), self.webots_process.pid))
```

resources/web/server/kill_old_webots_instances.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python3
+
+import psutil
+import re
+import time
+import os
+
+'''
+Kill all instances of streaming Webots processes fulfilling the following conditions:
+- created since 3 days
```
omichel:
```suggestion
- created 3 days ago or earlier
```

resources/web/server/kill_old_webots_instances.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python3
+
+import psutil
+import re
+import time
+import os
+
+'''
+Kill all instances of streaming Webots processes fulfilling the following conditions:
```
omichel: Is this script supposed to work on any platform? If not we should explain on which platforms it works.

resources/web/server/kill_old_webots_instances.py
```diff
@@ -0,0 +1,78 @@
...
+    for root, subdirs, files in os.walk(path):
+        for filename in files:
+            fileStatsObj = os.stat(os.path.join(root, filename))
+            if fileStatsObj.st_mtime < timeout:
+                return True
+    return False
+
+
+currentTime = time.time()
+creationTimeLimit = currentTime - 86400 * 3  # days in seconds
```
omichel:
```suggestion
creationTimeLimit = currentTime - 86400 * 3  # 3 days in seconds
```

resources/web/server/kill_old_webots_instances.py
```diff
@@ -0,0 +1,78 @@
...
+modificationTimeLimit = currentTime - 86400 * 1  # days in seconds
```
omichel:
```suggestion
modificationTimeLimit = currentTime - 86400 * 1  # 1 day in seconds
```

stefaniapedrazzi:
> Is the kill_old_webots_instance.py script called manually or from another script?

The `kill_old_webots_instance.py`script is called manually.
I would check when deploying the new robotbenchmark version if we still have many running webots instances and in that case add a scheduler in the simulation_server.py script that runs this every day or so.

stefaniapedrazzi:
:+1:

PR #888, opened 2019-09-16
Fix removing the second node from node_list in libController

### LINKED ISSUE #883: Wrong Node Ref After Removing Penultimate Node in MF Field
**Describe the Bug**
I'm working on a project where I need a supervisor (in C++) that creates nodes, keeps a pointer on them to modify them and deletes them. I've discovered that if I delete the penultimate node created, then the pointer on the ultimate node is corrupted, and if I try to access it, I got the error "Error: wb_supervisor_node_remove() called with NULL or invalid 'node' argument." 

**Steps to Reproduce**
1. start a new world
2. create a Robot node with Supervisor field at True 
3. set the controller field with the controller "testBugSupervisor" attached
4. it will simulate the behaviour to get the errror

**Expected behavior**
It only occurs with the suppresion of the penultimate nodes created and doesn't matter how the node has been imported in the world,. 

**System**
 - Operating System: Windows 7 pro
 - Graphics Card: NVIDIA GeForce GTX 1060 6GB

[testBugSupervisor.zip](https://github.com/cyberbotics/webots/files/3604792/testBugSupervisor.zip)

stefaniapedrazzi:
Fix #883: fix deleting the second node of the `node_list` list in libController `supervisor.c`. 

Previously both the first and second nodes were wrongly removed from the list.

### DIFF
--- a/docs/reference/changelog-r2019.md
+++ b/docs/reference/changelog-r2019.md
@@ -37,8 +37,9 @@ Released on XXX YYth, 2019.
     - Webots now reads the Python shebang of controller programs to determine which version of Python to execute.
     - External controllers now wait if started before Webots.
     - Fixed warnings printed in the terminal if a [Solid](solid.md).name field contains characters with special meaning in regular expressions.
+    - Fixed invalid node references in controllers after deleting nodes from Webots or from the [Supervisor](supervisor.md) API (thanks to @chilaire).
     - Linux: Fixed missing Python3.7 controller API.
-    - Windows: Fixed possible DLL conflict with libssl-1_1-x64.dll and libcrypto-1_1-x64.dll.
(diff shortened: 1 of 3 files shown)


stefaniapedrazzi:
:+1:

stefaniapedrazzi:
I also fixed a Webots crash when trying to delete twice the same node or deleting the root node.

:+1:

fabienrohrer:
Would it be simple to add tests for these cases?

stefaniapedrazzi:
Yes it shouldn't be complicated. I'm not sure how the test_suites handles a Webots crash but it should report the error as expected.

stefaniapedrazzi:
:+1:

DavidMansolino:
> I'm not sure how the test_suites handles a Webots crash but it should report the error as expected.

It depends in which test category the crash happen, but it is reported as a failure in any case if a crash happens.

:eyes:

PR #889, opened 2019-09-16
Fix CppCheck issues

stefaniapedrazzi:
Cppcheck 1.86 reports:
```
[src/webots/maths/WbPolygon.cpp:33] -> [src/webots/maths/WbPolygon.cpp:87]: (style) Variable 'det' is reassigned a value before the old one has been used.
[src/webots/nodes/WbPhysics.cpp:148] -> [src/webots/nodes/WbPhysics.cpp:155]: (style) Condition 'size==2' is always true
```


stefaniapedrazzi:
:+1:

Issue #891, opened 2019-09-17
Wrong Rendering if TextureCoordinate are Equal


DavidMansolino:
**Describe the Bug**
If the texture coordinate of the 3 vertices of a triangle are equal the rendering is not homogeneous as expected but looks completely wrong.

**Screenshots**
![empty](https://user-images.githubusercontent.com/2461619/65041292-c199ae80-d956-11e9-8771-6d8a70feee27.png)

**System**
 - Operating System:Linux Ubuntu 18.04
 - Graphics Card: NVIDIA GeForce GTX 970 

**Additional context**
This is reproducible with the following world:
```
#VRML_SIM R2019b utf8
WorldInfo {
}
Viewpoint {
  orientation -0.5610947722608096 -0.7525912045976002 -0.3446434901516687 1.2782008704957197
  position -0.23236190965843073 1.1022376123073423 0.9162881547270028
}
TexturedBackground {
}
TexturedBackgroundLight {
}
Shape {
  appearance CementTiles {
  }
  geometry IndexedFaceSet {
    coord Coordinate {
      point [
        0 0 0
        1 0 0
        1 0 1
      ]
    }
    texCoord TextureCoordinate {
      point [
        0 0
        0 0
        0 0
      ]
    }
    ccw FALSE
    coordIndex [
      0, 1, 2, -1
    ]
    normalIndex [
      0, 1, 2, -1
    ]
  }
}
```

DavidMansolino:
The problem is visible only if using a normal map texture.

stefaniapedrazzi:
This issue is also visible with the Laptop PROTO

Webots screenshot:
![laptop](https://user-images.githubusercontent.com/5910449/65127805-864fbc00-d9f8-11e9-93be-d22b1de7d718.png)

stefaniapedrazzi:
I will check if I can identify where does the issue come from.

stefaniapedrazzi:
The problematic function is `cotangentFrame` in `pbr.frag`:
https://github.com/cyberbotics/webots/blob/bd8e1aff86dac1802ad42c2690e19fca50f6375f/resources/wren/shaders/pbr.frag#

For example if I change the `invmax` value to 1.0, the issue disappears.
But I'm still debugging to understand where is exactly the source of the problem and how to prevent it.

PR #892, opened 2019-09-18
webots.min.js: texture mapping rendering issue

### LINKED ISSUE #752: webots.min.js: rendering issues on some systems
On my Linux desktop I get some issues when rendering specific objects, for example:
* `TruckSimple` in the `highway_driving` benchmark
Screenshot from 2019-07-25 11-20-03

* both `BungalowStyleHouse` objects in the `humanoid_marathon`
Screenshot from 2019-07-25 11-10-13

I tested both on Chrome and Firefox and I get the same issue.
The black dots are visible both the in the robotbenchmark animation and simulation.

@fabienrohrer tested on his macOS Chrome/Firefox and works correctly.

 - Graphics Card: NVIDIA GeForce RTX 2060 6 GB

stefaniapedrazzi:
Fix #752: add some small offset to default texture coordinates to avoid rendering issues if the multiple texture coordinates of a face are exactly the same.

Fix "flat" texture mapping in `TextureParallelepiped` to solve rendering issues with doors in `avoid_obstacles.wbt`.


stefaniapedrazzi:
It just remain to fix the rendering issue with the Laptop PROTO (similar issue also visible in Webots see #891).

stefaniapedrazzi:
It is very likely that three.js has the same bug in their shaders as the one fixed in wren (https://github.com/cyberbotics/webots/pull/895).

I'm checking if I can create a simple example to report the issue.

:+1:

DavidMansolino:
:eyes:

PR #898, opened 2019-09-20
Fixed compilation of lua which was displaying a warning

omichel:
The warning was:
```
make[1]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
```


omichel:
It is now faster to compile lua (uses multiple threads) and doesn't display the warning any more. I tested on Windows and Linux. On macOS, the package is pre-compiled, so there is no such problem.

👍

DavidMansolino:
(I launched the compilation+test suite on Travis)

PR #910, opened 2019-09-25
Merge revision into develop

stefaniapedrazzi:
Fix conflicts in:
* `dependencies/Makefile.linux`

PR #915, opened 2019-09-26
Debug test suite

DavidMansolino:
Fix nighlty failure on revision.


DavidMansolino:
It seems related to Qt:
```
RuntimeError: Error when getting the Webots version (the command "/home/travis/build/cyberbotics/webots/webots --version" exited with "134" and output "Info: Could not load the Qt platform plugin "xcb" in "" even though it was found.

Fatal: This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: xcb.

/home/travis/build/cyberbotics/webots/webots: line 44: 10000 Aborted                 (core dumped)
```

DavidMansolino:
I tried to recreate the Qt package from Ubuntu 16.04 but doesn't solve the issue.

DavidMansolino:
Now the same problem is reproducible on develop too, it seems therefore definitely linked to the Qt update (that's the main change of the latest merge: https://github.com/cyberbotics/webots/pull/910/files).

DavidMansolino:
Adding `libxkbcommon-x11-dev` to the Travis dependencies fixes the issue :smiley:

DavidMansolino:
will be included in #855

PR #926, opened 2019-09-30
E-puck 2 long-range distance sensor

Willem3141:
This PR implements the long-range distance sensor (time-of-flight sensor) that the E-puck 2 is equipped with. I added the ToF sensor to the PROTO, so it works in simulation, and to the WiFi remote control. I also added it to the robot window:

image

For the E-puck 1, where the ToF sensor is not present, the new sensor is simply hidden from the robot window. I also added a picture of the E-puck 2, which the robot window now shows instead of the E-puck 1 if you're working with a E-puck 2.

Please note that I do not have access to an E-puck 1, so I have not been able to test if I did not accidentally break remote-controlling that (although I did test simulating an E-puck 1, and that works fine). I also did not edit the cross-compilation (I have no experience with that, plus it does not work for the E-puck 2 anyway if I'm not mistaken?)

Please let me know if you have any questions / comments!


omichel:
Thank you for this great contribution. I am going to review it and test it. :eyes:

omichel:
I am testing this PR and when running the e-puck2.wbt simulation, I am getting a warning in the Webots console:
```
[e-puck] Error: wb_distance_sensor_get_value() called for a disabled device! Please use: wb_distance_sensor_enable().
```
Do you get the same?

Also, there is no 3D shape for the new ToF sensor, maybe we should add one in the model of the e-puck2?

Willem3141:
> Do you get the same?
Ah, oops, thanks for catching that! I tested only with my customized controller (which enabled the ToF sensor).

> Also, there is no 3D shape for the new ToF sensor, maybe we should add one in the model of the e-puck2?
I can try to add it, but the sensor is really tiny, so I don't know if it is necessary?

omichel:
I just fixed these two minor problems. I believe we should be able to merge your contribution with that. I will ask for a review from another webots developer before we can merge it.
:+1:

omichel:
👍

Willem3141:
Thanks! :tada:

PR #931, opened 2019-10-01
Fix libraries names in Changelog

stefaniapedrazzi:
Fix underscores in libraries name that are interpreted as italic text.


stefaniapedrazzi:
:+1:

PR #935, opened 2019-10-01
Fix macOS package creation

fabienrohrer:
`cp` on travis fails with this warning:

    cp: the -R and -r options may not be specified together.

- [x] Fix `cp` calls. As `-a` is a shortcut to `-pPR` and  `-r` is equivalent to `-R` (but not documented on macOS), `-a` is equivalent to `-ar` and less problematic.


fabienrohrer:
👍

PR #936, opened 2019-10-01
Fix C3D file not defined.

### LINKED ISSUE #933: Invalid controller arguments for C3dViewer PROTO
The controller arguments generated by the C3dViewer PROTO are wrong and are not correctly handled by the `c3d_viewer.py` controller.

In the PROTO the arguments list is generated without checking if the values are empty or not:
https://github.com/cyberbotics/webots/blob/32b7a6a33f51acea3d5f5a1611c24d2805b1d7f8/projects/humans/c3d/protos/C3dViewer.proto#L25-L35

So if the c3dFile value is empty the resulting arguments list received by the controller will be:
```
[c3d_viewer] ['c3d_viewer.py', '1', '0.28', '0', '0.28', '-0.21', '0', '0.22', '1', '-1', '0.25']
```
The c3dFile value is not in the list and the playbackSpeed is taken by the controller as the file name.

The result is that a default PROTO will print this strange message in the console:
```
INFO: c3d_viewer: Starting controller: python3 -u "c3d_viewer.py"  1 0.28 0 0.28 -0.21 0 0.22 1 -1 0.25
[c3d_viewer] '1' does not exist.
WARNING: 'c3d_viewer' controller exited with status: 1.
```

DavidMansolino:
Fix  #933

### DIFF
--- a/projects/humans/c3d/controllers/c3d_viewer/c3d_viewer.py
+++ b/projects/humans/c3d/controllers/c3d_viewer/c3d_viewer.py
@@ -78,7 +78,7 @@ timestep = int(supervisor.getBasicTimeStep())
 enableValueGraphs = []
 
 # parse arguments
-if len(sys.argv) < 3:
+if len(sys.argv) < 3 or sys.argv[1] == 'None':
     sys.exit('C3D file not defined.')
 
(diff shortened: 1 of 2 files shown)


DavidMansolino:
:+1:

PR #940, opened 2019-10-01
Removed Qt Wayland platform plugin

omichel:
Wayland is not needed as we can disable it from an environment variable in the snap environment.

👍


omichel:
Yes.

PR #944, opened 2019-10-01
Fixed compilation of Qt robot windows on snap

omichel:
👍


omichel:
Actually, the creation of soft links for Qt5 libraries was not consistent with other shared library we distribute, it was wrong and somehow redundant. This PR actually cleans this up and fixes missing soft links that are needed to link with Qt libraries.

src/packaging/webots_distro.c
```diff
@@ -1084,30 +1084,6 @@ static void create_file(const char *name, int m) {
       fprintf(fd, "mkdir usr/local/bin\n");
       fprintf(fd, "ln -s /usr/local/%s/webots usr/local/bin/webots\n", application_name_lowercase_and_dashes);
       fprintf(fd, "cd %s/debian/usr/local/%s/lib\n", distribution_path, application_name_lowercase_and_dashes);
```
DavidMansolino: We should have removed this one too.

PR #956, opened 2019-10-02
Fix OSM importer dependencies installation

DavidMansolino:


DavidMansolino:
This seems to fix the dependencies installation, I still need to test with the importer.

DavidMansolino:
:+1:

PR #958, opened 2019-10-02
Fix C3D runtime issue.

DavidMansolino:
The example C3D file provided was causing issues because it is missing the `POWERS`, `MOMENTS` and `FORCES` arrays.


DavidMansolino:
:+1:

PR #967, opened 2019-10-02
Fix OpenAl dependency

DavidMansolino:


fabienrohrer:
Thank you.

fabienrohrer:
It seems to be correct.

PR #972, opened 2019-10-03
Fix libopenal soft link

### LINKED ISSUE #964: Issues with Travis and OpenAL library
In a justified way, @omichel tried to simplify the OpenAL dependency on Linux. Unfortunately, `Travis` seems to break this dependency when creating the tar.bz2 file. We had to revert this patch for the `R2019b.rev1` release, but it would be interesting to dig further in this direction: https://github.com/cyberbotics/webots/pull/963

stefaniapedrazzi:
Fix #964.

Note that it is not necessary to create a package to check the issue but this can be inspected directly on the development environment by checking that the soft link points to the library with a relative path and not an absolute path.

### DIFF
--- a/dependencies/Makefile.linux
+++ b/dependencies/Makefile.linux
@@ -34,7 +34,7 @@ qt-clean:
 qt: $(WEBOTS_HOME)/lib/libQt5Core.so.$(QT_VERSION)
 
 $(WEBOTS_HOME)/lib/libQt5Core.so.$(QT_VERSION): $(WEBOTS_DEPENDENCY_PATH)/$(QT_PACKAGE)
-	tar xvjfm $(WEBOTS_DEPENDENCY_PATH)/$(QT_PACKAGE) -C $(WEBOTS_HOME)
+	tar xfm $(WEBOTS_DEPENDENCY_PATH)/$(QT_PACKAGE) -C $(WEBOTS_HOME)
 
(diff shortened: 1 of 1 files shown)


omichel:
It is not consistent with what we do for other libraries such as libssl, libcrypto, etc.
I believe it's cleaner to copy all the files (including the correct soft links) from the dependency package instead of re-creating them.

stefaniapedrazzi:
Yes.
This is only the fixed version of the current work flow.

omichel:
My previous contribution was not making this absolute link and we still had the problem on travis (but not on a local machine)... Also, I believe the link to openal is done the wrong way: openal.so should be a soft link pointing to openal.so.1 which should point to the binary file (openal.so.1.x.y) and not the other way round. This is the case for the Qt5 libraries and any other library. Again, that's a argument for copying the binaries and links from the dependency package and not recreating the links.

omichel:
I just tested to build a release from this branch. I deleted the travis cache before creating the release and... tadam... it worked: the openal links are now correct.

I believe the soft links are now created consistently and the right way (copy from the dependency packages, the binary file being the one with complete version number, other being soft links).

I will create a new PR for introducing MD5 hash checks to fix the dependency change problem.
Meanwhile, can merge this PR which fixes our tarballs (and possibly deb) nightly builds.

:+1:

DavidMansolino:
:eyes:

PR #973, opened 2019-10-04
Fixed extern controllers on snap

### LINKED ISSUE #950: Snap package limitations
Following-up #912, here are the remaining issues with the snap package:

- External controllers don't work.
- Impossible to use a different version of Python, impossible to install new pip modules.
- On OpenSuse (and probably other system), the compilation of C/C++ controllers raises exceptions from lsb_release about missing apt-cache (this is not problem specific to the snap package, the tarball package has likely the same problem).
- MATLAB controllers are not functional.
- On some systems other than Ubuntu 18.04, the speakers/sound are not working. That's probably because the user need to connects manually the webots snap to pulseaudio.
- e-puck cross-compilation is not working. However, we will deprecate this soon.

These limitations (with possible workarounds) should be documented in the user guide.

omichel:
Fixes #950.

This PR makes it easy to run extern controllers with the snap package (no need to be administrator). Basically, the webots launcher and libController were modified so that the snap version of webots can run extern controllers without any special setup. The user may override the `WEBOTS_TMPDIR` and/or `TMPDIR` environment variables before launching webots and the extern controllers if needed. Otherwise, they will be automatically setup to a default value (`$SNAP_USER_COMMON/tmp` = `~/snap/webots/common/tmp` for the snap version or `/tmp` for the non-snap version) that guarantees it will work smoothly with both the snap and non-snap versions of webots.

Additionally, the launcher now cleans-up the temporary folder and pipe files in case webots crashed without cleaning them up.

The new documentation page is available here for review: https://cyberbotics.com/doc/guide/installation-procedure?version=fix-extern-controller-on-snap

Finally, once this PR is merged, the https://cyberbotics.com/download page should be updated to add a link to the [Webots page](https://snapcraft.io/webots) on the snap store.

Also, as soon as we get an answer from https://lauchpad.net, we will be able to change the snap author from "Olivier Michel" to "Cyberbotics".

### DIFF
--- a/docs/guide/installation-procedure.md
+++ b/docs/guide/installation-procedure.md
@@ -14,32 +14,18 @@ The [chapter](language-setup.md) covers the set up of these tools.
 
 Webots will run on most recent Linux distributions running glibc2.11.1 or earlier.
 This includes fairly recent Ubuntu, Debian, Fedora, SuSE, RedHat, etc.
-Webots comes in two different package types: `.deb` and `.tar.bz2` (tarball).
-The `.deb` package is aimed at the latest LTS Ubuntu Linux distribution whereas the tarball package includes many dependency libraries and it is therefore best suited for installation on other Linux distributions.
-These packages can be downloaded from our [website](https://cyberbotics.com/download).
-
-> **Note**: Some of the following commands requires the `root` privileges.
-You can get these privileges by preceding all the commands by the `sudo` command.
-
-<!-- -->
+Webots comes in three different package types: `.deb` (Debian package), `.tar.bz2` (tarball package) and `.snap` (snap package).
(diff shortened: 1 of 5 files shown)


omichel:
After more than 2 days of trial and failure, the snap was finally uploaded on snapcraft.io!
You can install it from the candidate channel with `snap install --candidate webots`.

✔️ 👍

omichel:
I improved the documentation with respect to `WEBOTS_HOME` and `LD_LIBRARY_PATH`.

:+1:

stefaniapedrazzi:
Thank you. i'm reviewing it
:eyes:

stefaniapedrazzi:
The documentations is fine :+1:

omichel:
👍

stefaniapedrazzi:
:eyes:

omichel:
(note: the last version of the snap was uploaded to the snap store, version 11, on the candidate channel).

PR #974, opened 2019-10-04
Fix clock not published in synchronized mode.

DavidMansolino:
Fix https://stackoverflow.com/questions/58232623/how-to-sync-robot-in-ros

The clock need to be published non-stop (which was not the case when the controller was waiting for a call to the step service).


DavidMansolino:
The user confirmed that this fixes its issue.

:+1:

PR #976, opened 2019-10-07
Qt5 update on Windows

omichel:
MSYS2 updated Qt over the week-end.

You should run again the `src/install_scripts/qt_windows_installer.sh` script that will proceed automatically to the update.
I tested this update and it caused no compilation warning and Webots seems happy with it (we already used this version of Qt on Linux).

👍


omichel:
Note: I am not going to merge this PR right now, because I would like to merge #973 before to release the snap version without this change in the change log.

Issue #928, opened 2019-10-01
The buttons of the web interface look bad


omichel:
The height of the fast button is not the same as the other run and step buttons.
The shape of the run button is different from the shape of the other buttons, etc.
![image](https://user-images.githubusercontent.com/1264964/65939827-3aa10780-e427-11e9-865f-70f18da0aef7.png)
We should reuse the icons of Webots which are better designed. That would provide also more consistency.
![image](https://user-images.githubusercontent.com/1264964/65940089-e5192a80-e427-11e9-9cf2-390b205f9f8e.png)

stefaniapedrazzi:
When the fast button has been added it seems that has been copied directly from the Webots icons without taking into account the style used in the web interface.

stefaniapedrazzi:
I also don't like very much the appearance of the new world selection drop-down list and I think it should be improved.

PR #979, opened 2019-10-07
webots.min.js: enhancement web interface toolbar

stefaniapedrazzi:
Address #928:
* [x] improve simulation mode buttons and use the same icons as the Webots ones
* [x] improve world selection drop-down list


stefaniapedrazzi:
:+1:

omichel:
I am wondering if the behavior of the "fast" button in the web interface is the same as in Webots (e.g., black display and run as fast as possible) or is it equivalent to the "run" button?

stefaniapedrazzi:
The "fast" button in the web interface sets the Webots simulation mode to "fast".
https://github.com/cyberbotics/webots/blob/f743e2f0d03c6bfaa6a01b6bd169ba39b3124148/src/webots/gui/WbStreamingServer.cpp#L412

stefaniapedrazzi:
I agree that the choice of the icon is arguable (fast vs. run) given that in the web interface the scene is rendered in any case.

But in any case the icon should match the button name.

stefaniapedrazzi:
Ok, so I will rename the button (and everything related to it) to run and change the icon.
Note that currently this button is not visible in any of the default application we provide but only for custom solution.

stefaniapedrazzi:
I changed the target branch to develop because changing the flag name from `webots.showFast` to `webots.showRun` could break the compatibility.

:+1:

omichel:
:eyes:

PR #983, opened 2019-10-08
New geometry: rectangle

michou214:
Add a single/double sided and adjustable rectangle geometry

Parameters :
-  **size x y** where x and y have to be positive
- **doubleSided TRUE/FALSE** to see only one side or both


michou214:
:+1:

projects/objects/geometries/protos/Rectangle.proto
```diff
@@ -0,0 +1,57 @@
+#VRML_SIM R2019b utf8
+# license: Copyright Cyberbotics Ltd. Licensed for use only with Webots.
+# license url: https://cyberbotics.com/webots_assets_license
```
omichel: I believe we should release that under the Apache 2.0 license.

DavidMansolino: This is a more general question, for now, all of our 'home-made' PROTO are using this custom license.
If we want to change this we should probably define (and document) a clear rule and switch them all at the same time.

omichel: OK, let's do it separately.

michou214:
:+1:

PR #984, opened 2019-10-08
Fix appveyor build

omichel:
nodejs is not available any more from msys, see https://github.com/msys2/MINGW-packages/issues/5848
Therefore we need to install it in a different way.
I updated the Wiki accordingly: https://github.com/cyberbotics/webots/wiki/Windows-Optional-Dependencies#nodejs


omichel:
I could successfully create a Windows package from a tag on appveyor on this branch.
✔️ 👍

DavidMansolino:
:eyes:

PR #985, opened 2019-10-08
Display Name field after node type in tree view in gui

HefnySco:
I am a recent user to webots. and I luv it.
I used it  to implement quaddcopter and tricopter in Ardupilot.
the problem I found is that I always want to know which arm of the quad try and error I just added field name to the Tree View nodes.
I am not a QT developer so I just try to figure out which field to change, but for example we need to make it an optional not mandatory view.


omichel:
Hi @HefnySco, thank you for this interesting contribution!
I am going to review it today :eyes:

omichel:
@HefnySco: what do you think of #992? Does it fulfill your need?

HefnySco:
> @HefnySco: what do you think of #992? Does it fulfill your need?

more straight forward implementation.

PR #989, opened 2019-10-09
Fixed Java version on Travis and User Guide

### LINKED ISSUE #988: Inconsistent Java version
The java version has been changed in the linux installation instructions: https://github.com/cyberbotics/webots/wiki/Linux-Optional-Dependencies/_compare/a787c2d67b0c01558001110859e086d9d7f650ca...b4cdb914519eb157e74eb2a0403ab0b0ad52a57f

But this has not been reflected in Travis, this is bad as it means that we are not using the same development environment locally than the one used for testing and releasing.

omichel:
Fixes #988.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,7 @@ jobs:
       install: skip  # bundle install is not required
       language: python
       python: 3.7
-      jdk: openjdk8
+      jdk: openjdk11
       addons: &apt_packages
         apt:
           packages:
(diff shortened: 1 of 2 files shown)


omichel:
Since Travis currently runs Ubuntu 16.04, `openjdk-11-jdk` is not available there. So, we should stick to `openjdk-8-jdk` until we test on Travis with Ubuntu 18.04.

PR #992, opened 2019-10-09
Fix better scene tree naming

omichel:
Following-up on the good idea from @HefnySco proposed in #985, this implementation takes care of displaying only device names to avoid over-cluttering the scene tree.

### DIFF
--- a/docs/reference/changelog-r2019.md
+++ b/docs/reference/changelog-r2019.md
@@ -5,6 +5,7 @@ Released on ???
 
   - Enhancements
     - Webots now wait for extern controllers if the `Robot.synchronization` field is set to `TRUE`.
+    - Device names are displayed in the scene tree next to the node name.
   - Dependency Updates
     - Windows: upgraded to Qt 5.13.1.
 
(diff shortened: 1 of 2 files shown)


HefnySco:
This is much straight forward.
Thank you for implementing it.

omichel:
👍

stefaniapedrazzi:
Why do you display only the name for the devices and not for robots?
This doesn't seem very consistent given that the "name" field is inherited from the Robot node.

omichel:
Because we don't want to "pollute" too much the scene tree with information that is not very useful. It is very relevant and convenient to display device names when you program a robot. However, having the robot name displayed is much less useful.

HefnySco:
@stefaniapedrazzi @omichel 
why not adding a check-box in the "view" menu.
to display all or non. that is how we limit polluted view for ppl. who are not interested and switch to full view to keep it consistent from gui view point. -just a thought-.

omichel:
At some point, we are going to perform a more in-depth re-design of the scene tree in Webots, so we will take your advice into account, but for now, it is okay to proceed with such a compromise.

PR #994, opened 2019-10-11
Check MD5 for dependencies

### LINKED ISSUE #965: Better dependency management
Modifying the dependencies on the FTP in a single directory is for sure a bad practice.
The side effects may be big, it bypasses the review process, and it has no versioning control. Updating the package can break the previous Webots versions.

Simply adding a supplementary directory specifying the Webots version would already help. Probably even better solutions may be found.

omichel:
Fixes #965.
In this PR, the dependencies Makefile checks the MD5 sum of each Webots dependency and fails if the check fails. The output of the Makefile was also cleaned-up, so that it is now much more readable.

### DIFF
--- a/dependencies/Makefile.linux
+++ b/dependencies/Makefile.linux
@@ -29,119 +29,160 @@ clean:
 cleanse: $(PACKAGES_CLEAN)
 
 qt-clean:
-	rm -rf $(WEBOTS_DEPENDENCY_PATH)/webots-qt-*-linux64-release.tar.bz2 $(WEBOTS_HOME)/$(QT_PACKAGE)* $(WEBOTS_HOME)/lib/libQt* $(WEBOTS_HOME)/lib/libicu* $(WEBOTS_HOME)/lib/qt $(WEBOTS_HOME)/include/qt $(WEBOTS_HOME)/bin/qt/lrelease $(WEBOTS_HOME)/bin/qt/lupdate $(WEBOTS_HOME)/bin/qt/moc $(WEBOTS_HOME)/resources/web/local/qwebchannel.js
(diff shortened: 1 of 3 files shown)


omichel:
✔️ 👍

dependencies/Makefile.linux
```diff
@@ -29,119 +29,160 @@ clean:
...
+	@tar xfm $(WEBOTS_DEPENDENCY_PATH)/$(QT_PACKAGE) -C $(WEBOTS_HOME)
 
 $(WEBOTS_DEPENDENCY_PATH)/$(QT_PACKAGE):
-	rm -f $(WEBOTS_DEPENDENCY_PATH)/$(QT_PACKAGE)
-	wget $(DEPENDENCIES_URL)/$(QT_PACKAGE) -P $(WEBOTS_DEPENDENCY_PATH)
-	touch $(WEBOTS_DEPENDENCY_PATH)/$(QT_PACKAGE)
+	@rm -f $(WEBOTS_DEPENDENCY_PATH)/$(QT_PACKAGE)
+	@echo "# downloading $(QT_PACKAGE)"
+	@wget -qq $(DEPENDENCIES_URL)/$(QT_PACKAGE) -P $(WEBOTS_DEPENDENCY_PATH)
+	@echo "e004cc3d47fd359131fb931303214c9d  $(QT_PACKAGE)" > $(QT_PACKAGE).md5
```
DavidMansolino: I would put the MD5 value at the beginning of the Makefile close to the package names so that it is easy to check and update them.

DavidMansolino: Ok let's keep it like this so that if it fails the md5 is displayed as error instead of the variable.

fabienrohrer:
On macOS, `md5sum` is not installed by default. This issue is raised:

```
# downloading opencv-2.4.3.tar.bz2
/bin/sh: md5sum: command not found
```

It seems that `md5` should be used instead.

omichel:
Oops. May I let you fix the problem and test it on macOS?

fabienrohrer:
Yes, sure.

fabienrohrer:
Not sure how travis deals with this. Installing brew `md5sha1sum` doesn't solve the problem. The `md5sum` coming from there does not have the `--status` option.

=> I think the most standard solution is to change `Makefile.mac` to use `md5` instead.

omichel:
Yes, probably. I was confident it would work on macOS since the travis source test on macOS passed.

PR #999, opened 2019-10-14
Fix possible crash with external controllers.

DavidMansolino:
At the destruction of the world, it happens that the robots are removed, and therefore the controllers also which causes issues with the `mRobotsWaitingExternController` list which is then referring to non-existing robot anymore which causes issues when removing them from the list due to connects.

No changelog entry as this was never released.


DavidMansolino:
:+1:  :heavy_check_mark:

PR #1003, opened 2019-10-14
Fix snap external controllers

DavidMansolino:
On Ubuntu 16.04 the system libraries are not the same that the one distributed within the snap package which causes conflicts when running external controllers.

Problematic libraries:
  - lipng16 VS 12
  - /lib/x86_64-linux-gnu/libz.so.1: version `ZLIB_1.2.9' not found (required by /snap/webots/current/usr/lib/x86_64-linux-gnu/libpng16.so.16)

https://www.cyberbotics.com/doc/guide/running-extern-robot-controllers?version=fix-snap-external-controllers#running-extern-robot-controller-with-the-snap-version-of-webots


DavidMansolino:
:+1:

omichel:
We should probably target the master branch with this PR.

Issue #1004, opened 2019-10-14
Improve Matlab error message in the SNAP


DavidMansolino:
In the snap package, we should at least improve the error message when trying to start a Matlab controller to say that this should be done using extern controller instead of the default message saying that Matlab should be added to the PATH.

DavidMansolino:
Fixed in #1005

PR #1005, opened 2019-10-14
Added an explanation message for MATLAB controllers / snap

omichel:
Addresses #1004.


src/webots/control/WbController.cpp
```diff
@@ -783,6 +783,12 @@ void WbController::startPython() {
 }
 
 void WbController::startMatlab() {
+#ifdef __linux__
+  if (qgetenv("SNAP_NAME") == "webots") {
```
DavidMansolino: That's a detail, but we should probably create a `bool isSnap()` function in WbSysInfo (similar to the already existing `isRootUser` function) as we use this test at several places in the code.

omichel:
👍

PR #1008, opened 2019-10-15
Fixed clang format version problem

omichel:
Five day ago, MSYS2 upgraded clang to version 9.0.0 (and thus clang-format). Since then, the clang-format formatting on Windows is different and causes the source test to fail on travis and appveyor. I found that with clang-format 9.0.0, we should simply add an option (`AllowAllConstructorInitializersOnNextLine: false`) to fix the problem and have the same formatting as previously. However, this option is not recognized by versions of clang-format older than 9.0.0, causing a run-time error. Therefore, we have to handle two different `.clang-format` configuration files: one for Windows with MSYS2 and the other for Linux, macOS and Appveyor (which uses an older version of clang-format).


omichel:
✔️ 👍

stefaniapedrazzi:
:eyes:

PR #829, opened 2019-08-21
Fix race stop watch

fabienrohrer:
Fix https://github.com/omichel/robotbenchmark/issues/327

When a the robot falls, the 30s time penalty is added every milliseconds. Soon enough, the timer (supporting 100 minutes) is overloaded causing weird results.

Adding a disqualifying penalty fix this issue.

PR #1009, opened 2019-10-16
Fix exporting first translation/rotation field change

stefaniapedrazzi:
Fix issue with digit translation export described in #829.

Bug Description:
The initial translation( and rotation) field value is not explicitly set and thus it is automatically set to `[0, 0, 0]`.
So if an object, whose initial position was different from `[0, 0, 0]`, is moved to the position `[0, 0, 0]` the change is not detected and not exported.


stefaniapedrazzi:
:+1:

fabienrohrer:
:eyes:

PR #1006, opened 2019-10-15
Implemented tornado on_message function

trackq:
Solving this error

```
2019-10-14 08:42:56,855 [ERROR  ]  Uncaught exception GET / (172.20.0.1)
HTTPServerRequest(protocol='http', host='localhost:1999', method='GET', uri='/', version='HTTP/1.1', remote_ip='172.20.0.1')
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/tornado/websocket.py", line 649, in _run_callback
    result = callback(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/tornado/websocket.py", line 427, in on_message
    raise NotImplementedError
NotImplementedError
```

PR #1015, opened 2019-10-17
Fix exception in session_server.py script

stefaniapedrazzi:
The `WebSocketHandler.on_message` should be overridden to prevent raising the `NotImplementedError` exception (see #1006).


stefaniapedrazzi:
:+1:

PR #1021, opened 2019-10-21
Fix Travis not Publishing Releases

DavidMansolino:
This explains why the releases are not uploaded on Github.


DavidMansolino:
:+1:

PR #1025, opened 2019-10-21
Add support for multiple table types in documentation

stefaniapedrazzi:
Currently it is only possible to create tables for the programming language because the tab option is stored in a generic "tab" GET value.
The aim of this PR is to improve the current system so that it will support multiple table options (at least programming language and operating system).

Main changes:
- when defining a `%tab-component` in the MD file, we should also specify the table title (e.g. "language", "os")
- each table type has a specific query value in the URL with `tab-` prefix(e.g. "tab-language", "tab-os")


stefaniapedrazzi:
Unfortunately we probably have to wait until the last commit will be available on jsdelivr.

:+1:

omichel:
:eyes: I am testing from https://www.cyberbotics.com/doc/guide/running-extern-robot-controllers?version=feature-multiple-tables-in-documentation#environment-variables.
When I select a language and an os, the language and os parameters are properly added to the URL. However, when I navigate to a different page (following a link in the doc menu), I loose the language and os parameters from the URL (but not the version parameter), so that when I come back to the original page I need to select os and language again.

stefaniapedrazzi:
> When I select a language and an os, the language and os parameters are properly added to the URL. However, when I navigate to a different page (following a link in the doc menu), I loose the language and os parameters from the URL (but not the version parameter), so that when I come back to the original page I need to select os and language again.

Strange, I'm not getting the same (I just cleared the cache).

This is the behavior before the fix in 1c88665 but should be fixed now.

And before the last commit 29088c2 (the current version delivered by jsdelivr), there are errors when selecting a tab.

stefaniapedrazzi:
Ok, this is correct.
To test we have to wait until jsdelivr updates the files to the latest version.

Alternatively in Chrome developer tools it is possible to modify the JS files while running them and apply manually the latest commit 29088c2

omichel:
I cleared the jsdelivr cache as explained here: https://github.com/omichel/cyberbotics/wiki/How-to-purge-jsDelivr-cache and it now works as expected.

DavidMansolino:
for me it is not working as expected, @omichel can you open this link: https://www.cyberbotics.com/doc/guide/running-extern-robot-controllers?version=feature-multiple-tables-in-documentation&tab-language=python#environment-variables
do you see the `python` tab as expected?

stefaniapedrazzi:
Yes, I will check tomorrow

> I cleared the jsdelivr cache as explained here: https://github.com/omichel/cyberbotics/wiki/How-to-purge-jsDelivr-cache and it now works as expected.

Good. Yesterday even after purging the resources were not updated...

stefaniapedrazzi:
The problem comes from `docs/doc.php` that doesn't parse correctly the version now that the tab key changed.

stefaniapedrazzi:
Escaping the HTML special characters before extracting the queries breaks the regular expression and make it more difficult (especially for `&` that is converted to `&amp;`).

To me it seems better to first extract the book, page, branch, etc. and then make escape the HTML special characters before using these values.

stefaniapedrazzi:
:+1:

omichel:
:eyes:

stefaniapedrazzi:
I just fixed a couple of links that were still not working correctly:
* `tab` query with version (page not loaded):
    https://www.cyberbotics.com/doc/guide/running-extern-robot-controllers?version=master&tab=python
* old `tab` query in new `viewer.js` (tab option was ignored):
    https://www.cyberbotics.com/doc/guide/running-extern-robot-controllers?version=feature-multiple-tables-in-documentation&tab=python

Then I uploaded the the `request_method.js` file so that I could remove the test code from `viewer.js`.

stefaniapedrazzi:
I tried to purge jsdelivr but it still doesn't work for me..

Maybe it is better to merge this PR directly in master.

omichel:
I could purge the jsdelivr cache with https://purge.jsdelivr.net/gh/cyberbotics/webots@feature-multiple-tables-in-documentation/docs/js/viewer.js

omichel:
You can check it at https://cdn.jsdelivr.net/gh/cyberbotics/webots@feature-multiple-tables-in-documentation/docs/js/viewer.js

omichel:
I updated https://github.com/omichel/cyberbotics/wiki/How-to-purge-jsDelivr-cache which was still referring to omichel/webots instead of cyberbotics/webots. Maybe that's the problem you experienced?

omichel:
Shall I update `doc.php` on the server with the new version you committed here?

stefaniapedrazzi:
> Shall I update doc.php on the server with the new version you committed here?

Yes, thank you.

> I updated https://github.com/omichel/cyberbotics/wiki/How-to-purge-jsDelivr-cache which was still referring to omichel/webots instead of cyberbotics/webots. Maybe that's the problem you experienced?

No, I'm sure I used the correct URL.
And I just checked but I'm still getting the old version.

omichel:
I am definitely getting the new version... Did you clear your browser cache?

stefaniapedrazzi:
> I am definitely getting the new version... Did you clear your browser cache?

yes, even after clearing the cache (both in Chrome and Firefox) I'm still getting the old version.
But if I download the file using `wget`, then it works and i'm getting the new one..

stefaniapedrazzi:
> So that's clearly a cache issue.

Yes, but I'm not sure it is a cache issue at my browser level.
I tried from a virtual machine and I get the exact same old file.

PR #1027, opened 2019-10-21
Fix logging messages in the console before GUI initialization

### LINKED ISSUE #825: Streaming server setup errors not printed in the console
The WbStreamingServer instance is created and initialized before the WbConsole instance, so the socket and tcp server errors are only printed on the terminal.

This is a problem because in case of applications like robotbenchmark.net where Webots is started from a script, the terminal is not available and the error logs are lost.

We should at least redirect the stderr of Webots to a file when starting it from `simulation_server.py` if it is not worth to change the instructions order to print the errors in the Webots console.

stefaniapedrazzi:
Fix #825: store log messages emitted before WbConsole instance and print them when the GUI initialization is over.


stefaniapedrazzi:
:+1:

Issue #1028, opened 2019-10-22
Cppcheck 1.89 Update


DavidMansolino:
**Describe the Bug**
Cppcheck 1.89 is giving many new issues, it is now available on mac OS on Travis, but we should update on Linux too:
  - [x] Fix/remove errors: https://travis-ci.com/cyberbotics/webots/jobs/248075574#L909
  - [x] Update on Linux too.

stefaniapedrazzi:
I fixed the errors in #1031.
But I'm not sure what should be done to update the cppcheck version on Linux for Travis.

DavidMansolino:
I will check this part.

stefaniapedrazzi:
Thank you

PR #1031, opened 2019-10-22
Fix cppcheck 1.89 errors

stefaniapedrazzi:
Address #1028


stefaniapedrazzi:
Yesterday evening I ran the `travis_unit` test and everything passes except for the unit test on ubuntu 18.04 where the `mirror_supervisor` failed.

So maybe we could start reviewing this PR.

The source code of ODE was not tested, so I removed the include files of ODE too that are now throwing many errors (mainly parameters that could be const).

:+1:

DavidMansolino:
:eyes:

stefaniapedrazzi:
Sorry I was reviewing for the last time before merging and I found that the new method `WbClipboard::deleteInstance` was incomplete.

:+1:

stefaniapedrazzi:
I run the test_suite locally and found another issue with the clipboard.

:+1:

PR #1033, opened 2019-10-23
Fixed Display.imageNew from array in Python

### LINKED ISSUE #1020: Python API imageNew() of Display
I think I found a bug in `Display.newImage()`.

In `resources/languages/python/controller.i`, `imageNew()` is defined as follows:
```
%rename (__internalImageNew) imageNew(int width, int height, const void *data, int format) const;
...
%extend webots::Display {
  %pythoncode %{
    def imageNew(self, data, format, width=None, height=None):
      if isinstance(data, list):
        return self.__internalImageNew(len(data), len(data[0]), data, format)
      elif width is None or height is None:
        raise TypeError('imageNew : width and height must be given if data is not a list')
      else:
        return self.__internalImageNew(width, height, data, format)
    def drawPolygon(self, x, y):
      self.__internalDrawPolygon(x, y, min(len(x), len(y)))
    def fillPolygon(self, x, y):
      self.__internalFillPolygon(x, y, min(len(x), len(y)))
  %}
}
```

I guess it's intended to support easy access to the image like `image[x][y][ch]` as `Camera.getImageArray()`, but `Displey.imageNew()` passes the given data directly to the `const void *data` parameter, which expects a row-major data.

It should rearrange the given data to column-major before passing the data to C's `imageNew()`.

omichel:
Fixes #1020.

Here is the code I used to test it:
```python
from controller import Robot, Display


class CameraDisplay(Robot):
    timeStep = 32
    motors = []

    def __init__(self):
        super(CameraDisplay, self).__init__()
        self.camera = self.getCamera('camera')
        self.display = self.getDisplay('display')
        self.camera.enable(self.timeStep)
        self.motors.append(self.getMotor('right wheel motor'))
        self.motors.append(self.getMotor('left wheel motor'))
        self.motors[0].setPosition(float('inf'))
        self.motors[1].setPosition(float('inf'))
        self.motors[0].setVelocity(1)
        self.motors[1].setVelocity(-1)

    def run(self):
        while True:
            a = self.camera.getImageArray()
            if a:
                image = self.display.imageNew(a, Display.RGB)
                self.display.imagePaste(image, 0, 0)
                self.display.imageDelete(image)
            if self.step(self.timeStep) == -1:
                break


controller = CameraDisplay()
controller.run()
```

### DIFF
--- a/docs/reference/changelog-r2019.md
+++ b/docs/reference/changelog-r2019.md
@@ -12,6 +12,7 @@ Released on ???
     - Fixed `simulation_server.py` script to work with Python3.
     - Fixed exporting first translation and rotation fields change during animation recording and simulation streaming.
     - Fixed displaying streaming server initialization errors in the Webots console.
+    - Fixed bug in Python Display.imageNew() when passing an image array: rearranged image data from column-major order (thanks to Inbae Jeong).
 
 ## Webots R2019b Revision 1
 Released on October 3rd, 2019.
--- a/docs/reference/changelog-r2019.md
+++ b/docs/reference/changelog-r2019.md
@@ -12,7 +12,7 @@ Released on ???
     - Fixed `simulation_server.py` script to work with Python3.
     - Fixed exporting first translation and rotation fields change during animation recording and simulation streaming.
     - Fixed displaying streaming server initialization errors in the Webots console.
-    - Fixed bug in Python Display.imageNew() when passing an image array: rearranged image data from column-major order (thanks to Inbae Jeong).
+    - Fixed bugs in Python Display.imageNew() when passing an image array: rearranged image data from column-major order and memory leak (thanks to Inbae Jeong).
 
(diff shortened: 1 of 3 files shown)


omichel:
Shall we also fix this line?
https://github.com/cyberbotics/webots/blob/ffbda7ae254c99bdbf72b5bab576ff8f68ea2c67/resources/languages/python/controller.i#L90

and replace it with:
```
%typemap(in) const double * {
```

I don't understand the purpose of `[ANY]` and I am afraid the `freearg` won't work on it.

omichel:
👍

DavidMansolino:
> I don't understand the purpose of [ANY]

http://www.swig.org/Doc3.0/SWIGDocumentation.html#Typemaps_nn17

not 100% sure, but I would say `%typemap(in) const double * {` is indeed better.

:eyes:

PR #1037, opened 2019-10-23
Fix problematic lsb_release in Makefile.os.include

### LINKED ISSUE #1001: Makefile.os.include calls lsb_release for all linux versions
The `Makefile.os.include` calls the `lsb_release` command on linux even in the non-ubuntu case which displays nasty errors in the console when compiling any code (even if the compilation works fine):

https://github.com/cyberbotics/webots/blob/c4ce84200caa0725655d1538076099d2fc036a22/resources/Makefile.os.include#L61

If we are not on Ubuntu we should not call this command and leave `UBUNTU_VERSION` empty.

DavidMansolino:
Fix #1001

### DIFF
--- a/resources/Makefile.os.include
+++ b/resources/Makefile.os.include
@@ -58,7 +58,9 @@ ifeq ($(OSTYPE),linux)
     OSARCH = x86_64
   endif
   ifeq ($(OSTYPE),linux)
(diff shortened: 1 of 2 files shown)


DavidMansolino:
:+1:

omichel:
:eyes:

PR #1039, opened 2019-10-24
Set the timeout to 30 seconds instead of 10 to minimize timeout failures on Travis

omichel:


omichel:
👍
✔️ https://travis-ci.com/cyberbotics/webots/builds/133334198

PR #1042, opened 2019-10-28
Fix cleanse rule if npm is not installed

### LINKED ISSUE #986: Mandatory npm Dependency
**Describe the Bug**
The npm dependency is documented as optional: https://github.com/cyberbotics/webots/wiki/Linux-Optional-Dependencies#nodejs

However, the main compilation process fails if npm is not installed. It should just be skipped if npm is missing, as it is already the case for java and python for example.

fabienrohrer:
Fix #986 😸


fabienrohrer:
👍

PR #1044, opened 2019-10-29
macOS: Update Qt

fabienrohrer:
- [x] Update scripts
- [x] Fix RPATH issues
- [x] Upload package on the FTP and update MD5
- [x] Fix crash on startup
- [x] Fix warning
- [x] Log

### DIFF
--- a/src/install_scripts/qt_mac_installer.sh
+++ b/src/install_scripts/qt_mac_installer.sh
@@ -1,5 +1,5 @@
 WEBOTS_PATH=~/webots
-QT_VERSION=5.12.5
+QT_VERSION=5.13.1
 
 # prepare Webots
 cd $WEBOTS_PATH
--- a/src/install_scripts/qt_mac_installer.sh
+++ b/src/install_scripts/qt_mac_installer.sh
@@ -1,20 +1,24 @@
 #!/bin/bash
 
-WEBOTS_PATH=~/webots
+if [[ -z "${WEBOTS_HOME}" ]]; then
+  echo "WEBOTS_HOME is not defined."
+  exit 1
+fi
+
 QT_VERSION=5.13.1
 
 # prepare Webots
-cd $WEBOTS_PATH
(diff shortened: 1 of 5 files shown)


fabienrohrer:
Notes for the reviewer:

- I choose Qt 5.12.5 on macOS instead of Qt 5.13.1 because it's not yet available as a precompiled package on macOS. Qt 5.12.5 is the [last LTS](https://en.wikipedia.org/wiki/Qt_version_history) and I don't want to rebuild manually Qt if I can avoid this.
- I dropped the general OpenGL setup in main.cpp. It is causing issue in QWebEngine (mainly, a warning complaining about the use of core profile). The target OpenGL format (OpenGL 3.3, etc.) is set in WbWrenWindow and it seems to be sufficient.

fabienrohrer:
👍 ✔️

omichel:
Just a preliminary question, it seems there is a pre-compiled binary of Qt 5.13.1 for mac here: https://www.qt.io/offline-installers : Is there a problem with it?

fabienrohrer:
Good point, I missed this page. Usually I use the online installer which doesn't include this package:

Capture d’écran 2019-10-30 à 10 31 34

I will check if this is equivalent.

fabienrohrer:
Fortunately, all the issues I solved for Qt 5.12 are present in Qt 5.13 too.
It remains only to wait on the distrib test results: https://travis-ci.com/cyberbotics/webots

src/install_scripts/qt_mac_installer.sh
```diff
@@ -1,5 +1,5 @@
-QT_VERSION=5.11.2
-WEBOTS_PATH=~/develop/webots
+WEBOTS_PATH=~/webots
```
omichel: Wouldn't it be more generic/scalable to use WEBOTS_HOME here?

fabienrohrer:
👍

PR #1045, opened 2019-10-30
Fix PEP8 Test Failure.

DavidMansolino:
This fixes the nightly test failure.

### DIFF
--- a/tests/sources/test_pep8.py
+++ b/tests/sources/test_pep8.py
@@ -37,6 +37,10 @@ skippedDirectories = [
     'projects/languages/ros/controllers/ros_python/python'
 ]
 
+if sys.version_info[0] < 3:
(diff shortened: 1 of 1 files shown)


fabienrohrer:
Thank you, mea culpa.

This is because I intensively used Python 3 for this script. By the way, I think this will become an issue soon because we will drop Python 2 soon: to be discussed on Monday.

DavidMansolino:
> This is because I intensively used Python 3 for this script. By the way, I think this will become an issue soon because we will drop Python 2 soon: to be discussed on Monday.

We will probably drop the Python2.7 test on Travis at the same time and keep only the Python3.x ones.

DavidMansolino:
:+1: :heavy_check_mark:

PR #1049, opened 2019-10-31
Preferences cleaner script: Minor improvements

fabienrohrer:
- [x] When running in Python2, a cryptic message was displayed because of a syntax error (about stderr).
- [x] On macOS:
    - [x] `unlink.missing_ok` argument was only available since Python 3.8.
    - [x] the assert was causing issue in case the preference file has been removed, but not applied by the system. Ignoring this is more robust.


fabienrohrer:
👍

fabienrohrer:
No it is not. I have done this only to display a correct error message when running from Python2, and telling Python2 is not supported for this script ;-)

PR #1051, opened 2019-10-31
Convert old backgrounds to HDR

fabienrohrer:
Step 3 of #834

- [x] Convert old backgrounds to HDR
- [ ] Adjust their exposure
- [x] Update TexturedBackground
- [ ] Check/Update doc
- [ ] Log


fabienrohrer:
This PR has no sense.

Converting an non HDR background to HDR is not possible, because some information is missing. It's like to convert a greyscale picture to a colored one.
it's not just a matter to apply a function on the pixels as I thought first: the result would be the same as modifying the Background.IBLStrength.

PR #1056, opened 2019-11-01
Fix visual cpp templates

omichel:
The problem was showing up when starting Webots from a MSYS2 console.
Because the MSYS2 console overrides the `TMP` and `TEMP` environment variables with `/tmp` which is not a valid Windows path, some helper applications (including Visual Studio) could be confused if relying on these environment variables. Therefore, it is safer to clear these environment variables on Windows before starting any helper application. Clearing them actually resets them to their initial default value, e.g., `C:\Users\Olivier\AppData\Local\Temp`, which is what most helper applications expect.

Also, the Visual Studio files generated by the "new controller" wizard were not respecting the carriage return of Windows. This PR also fixes this problem.


omichel:
👍

PR #1062, opened 2019-11-04
Fixed creation of package on Windows

omichel:
The creation of the Windows package was broken when the Makefile was called locally from the MSYS console.


omichel:
I could create a package locally and on appveyor from this branch.
👍 ✔️

PR #1066, opened 2019-11-05
webots.min.js: compile library as module

stefaniapedrazzi:
`webots.min.js` is a standalone library (including three.js, minified, including babel resources, etc.).
But to include it in another JS application (like the robot designer) we just need to concatenate together all the files.

### DIFF
--- a/resources/web/wwi/.gitignore
+++ b/resources/web/wwi/.gitignore
@@ -3,7 +3,7 @@
 /node_modules
 /package-lock.json
 /webots.min.js
-/webots-module.min.js
+/webots.lib.js
 /webots.debug.js
 /adapter.js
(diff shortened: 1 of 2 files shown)


stefaniapedrazzi:
:+1:

fabienrohrer:
It seems there is a naming issues in the diff: what is the name of the target? `webots-module.js`, `webots-module.min.js` or `webots.lib.js`? (I have a preference for the last one)

stefaniapedrazzi:
It's not minified, so it should not be `webots-module.min.js`.
For the other two options, I considered both and for me it is the same.
I will change to `webots.lib.js`.

stefaniapedrazzi:
:+1:

PR #1067, opened 2019-11-06
Dummy PR to inspect mirror test

stefaniapedrazzi:


stefaniapedrazzi:
I can reproduce a failure with the mirror test on Windows on the Acer laptop (not systematically but about 1/10 runs).

stefaniapedrazzi:
I could confirm that the failure of the test on my Windows is related to the order in which the 2 camera devices (test supervisor and mirror) are initialized and rendered:
* if the mirror camera is handled first -> test passes
* if the test camera is handled first -> test fails

Now I will check why the order is not always the same.
It could be that the controllers are still not always processed in the same order.

stefaniapedrazzi:
Issue should be fixed in #1083

PR #1068, opened 2019-11-06
Fix Windows build

omichel:
This PR fixes a broken dependency in MSYS2. The `qtwebkit` package was broken after a `qt5` update. Copying the old `libicu??64.dll` files in `/mingw64/bin` fixes the problem until it is fixed by @Alexpux. See https://github.com/msys2/MINGW-packages/issues/5934 for more information about it.


omichel:
👍 ✔️

DavidMansolino:
:eyes:

PR #1060, opened 2019-11-04
Fix none Controllers

DavidMansolino:
Fix #1047

The bug is also visible on the first world of the guided tour.

PR #1073, opened 2019-11-07
improve IK world: use a "none" controller for the root robot

fabienrohrer:
Thanks to https://github.com/cyberbotics/webots/pull/1060 fix.


fabienrohrer:
👍

PR #1075, opened 2019-11-07
Fix qt platform plugin trouble

omichel:
In some cases, Webots was unable to start because of a different version of the platform plugins. A typical case is the following:

1. I install Webots R2019b.
2. I update MSYS2 and get a new version of qt5.
3. I compile and start Webots from MSYS2.
4. I start my installed version of Webots.
5. Webots fails to start complaining it cannot load the qt platform plugin. This is because it attempts to load the new ones which were somehow cached by Qt when I launched Webots from MSYS2.

The solution is to set the `QT_QPA_PLATFORM_PLUGIN_PATH` in `main.cpp` to point to the correct location of the platform plugins (the ones compatible with the current Webots binary), before the initialization of Qt.

Once this PR is merged into the revision branch, the revision branch should be merged very carefully in the develop branch (see FIXME comments about it).


omichel:
👍

omichel:
👍

PR #1080, opened 2019-11-08
Removed useless include statement

omichel:
It was wrongly introduced in #1075.
👍


DavidMansolino:
Be careful when merging this in develop.

Issue #1061, opened 2019-11-04
Extend TexturedBackground PROTO


DavidMansolino:
As a follow-up of #880 we should improve the `TexturedBackground` PROTO:
  - [ ] Add a skycolor field (set to black by default).
  - [ ] Add a bolean field to define if we want to see the background (color part) or not.
  - [ ] Eventually add a bolean field to define if we want to see the reflections or not.
  - [ ] Other?

fabienrohrer:
Fixed in #1082

PR #1082, opened 2019-11-11
Implement new TexturedBackground fields

fabienrohrer:
- [x] Implement new fields for TexturedBackground (cf. #1061 )
- [x] Deprecate non-HDR backgrounds
    - [x] Fix worlds accordingly
    - [x] Add new HDR backgrounds to replace the old ones:
        - [x] `dusk`
        - [x] `mars`
- [x] Doc
- [x] Log


fabienrohrer:
👍

stefaniapedrazzi:
:eyes:

docs/guide/object-backgrounds.md
```diff
@@ -43,24 +43,28 @@ TexturedBackground {
 
 - `texture`: Defines the texture of the background.
 
+- `luminosity`: Is `Background.luminosity`.
+
+- `skybox`: Specifies if the background is used to define the textured skybox.
```
stefaniapedrazzi: I have the feeling that this sentence is not very clear. 
I would use "texture" instead of "background" term because even if the `skybox` field is disabled there is a uniform colored background.

```suggestion
- `skybox`: Specifies if the `texture` field is used to define the skybox.
```

But then I'm also wondering if  `skybox` is a good field name, given that from the description (and `fallbackSkyboxColor` field) it seems that even if this field is set to FALSE a skybox still exists.

docs/guide/object-backgrounds.md
```diff
@@ -43,24 +43,28 @@ TexturedBackground {
...
+
+- `reflections`: Specifies if the background is used in the material reflections.
+
+- `fallbackSkyboxColor`: Defines the background color in case the skybox is not defined.
```
stefaniapedrazzi: As mentioned in the previous comment, when reading these documentation I'm a little bit confused about the use of the terms "background", "skybox", and "texture".

I think that would be better to rename this field simply `backgroundColor`.

The description is ok like this only if no skybox is created in case the texture is not defined or disabled.
Otherwise instead of "the skybox is not defined" we should use "the `skybox` field is FALSE"

fabienrohrer:
I fully agree with your comments about renaming, I was not very confident with them. I tried to take into account your comments. I reduced the use of the `skybox` keyword, and rely the descriptions on the other fields. I also rename `skybox` to `background` for consistency. Is it better like this?

fabienrohrer:
> Given that the web_component_studio_template.wbt background changed, we should probably also update all the web component exported models.

I just tried, but the diff is big for other reasons. I would like to update them in another PR once this one is merged if you agree.

fabienrohrer:
👍

stefaniapedrazzi:
Thank you, yes the documentation is better like this.
The `background` field name seems still too generic. But unfortunately I don't have many ideas about how to improve it. Maybe at the end `skybox` was still slightly better.

(Note that at line 52 `skybox`is still sued instead of `background`)

By the way I checked the benchmarks and they work fine too.

fabienrohrer:
I think I would go for `skybox` for the bool, and `skyColor` for the MFColor (to match the Background node).

stefaniapedrazzi:
Yes, this seems good.

fabienrohrer:
👍

projects/objects/backgrounds/protos/TexturedBackground.proto
```diff
@@ -2,66 +2,100 @@
 # license: Copyright Cyberbotics Ltd. Licensed for use only with Webots.
 # license url: https://cyberbotics.com/webots_assets_license
 # tags: static
-# Background textured with a skybox.
+# Background textured with a background.
```
stefaniapedrazzi: is this correct or should it be
```suggestion
# Background textured with a skybox.
```

fabienrohrer: I think this line would deserve a better description, the role of this PROTO has increased.

fabienrohrer:
👍

docs/guide/object-backgrounds.md
```diff
@@ -2,23 +2,20 @@
 
 ## TexturedBackground
 
-Background textured with a skybox.
+`TexturedBackground` provides a set of background textures, and apply them to the background skybox and to the reflections on the 3D objects.
```
stefaniapedrazzi: The description is fine but there is a typo

```suggestion
`TexturedBackground` provides a set of background textures, and applies them to the background skybox and to the reflections on the 3D objects.
```

fabienrohrer:
Ok, thank you ^^

👍

PR #1083, opened 2019-11-11
Fix determism in cameras rendering order

stefaniapedrazzi:
Try to fix `mirror` test error.

Change the way the cameras are rendered by looping through the robots so that the order is always the same.
Previously the order was determined by the order in which the set sampling period command was received from the controllers.


stefaniapedrazzi:
:+1: :heavy_check_mark:

The travis_unit test succeded but I've never been able to reproduce the error when running the tests manually from travis.

On my Windows (were the issue was not systematically reproducible) I've no longer been able to get a failure.

But the ultimate test will be the nightly build.

PR #1093, opened 2019-11-12
Update web components

fabienrohrer:
Update components, principally to have the new HDR background.


fabienrohrer:
👍

stefaniapedrazzi:
:eyes:

stefaniapedrazzi:
Changes are fine.

Just a question about the quacopter:
https://www.cyberbotics.com/doc/guide/mavic-2-pro?version=update-web-components

previously it seems we manually fixed the X3D file so that only the slow helices and not the fast  ones were shown.

Then we should also update the `web_component_studio_supervisor.py` controller because it doesn't work with Python3.

fabienrohrer:
> Just a question about the quacopter:
https://www.cyberbotics.com/doc/guide/mavic-2-pro?version=update-web-components
previously it seems we manually fixed the X3D file so that only the slow helices and not the fast ones were shown.

You're right, we should certainly change this here.

> Then we should also update the web_component_studio_supervisor.py controller because it doesn't work with Python3 (I can open an issue for this).

For sure, but it's not related with this PR. Could you open an issue about this, and assign it to me?

fabienrohrer:
👍

Issue #1096, opened 2019-11-13
Drop support of Python 2 for R2020a


fabienrohrer:
We warn users since one year that Python 2 is deprecated, and won't be maintained after January 1 2020:
- https://pythonclock.org
- https://github.com/cyberbotics/webots/blob/revision/resources/languages/python/controller.i#L797

To be done before R2020a release:
- [ ] Make sure all our scripts are working with Python3, and don't bother with Python 2 maintenance.
    - [ ] web component: #1095
    - [ ] image_tool scripts #1097
    - [ ] ...
- [ ] Remove Python 2 from our build system.
- [ ] Remove Python 2 from test suite.
- [ ] Homogenize our shebang to `#!/usr/bin/env python3`
- [ ] Update doc

fabienrohrer:
It's not the time to drop Python 2 yet.

PR #1097, opened 2019-11-13
image scripts: Support Python 3 and drop Python 2

fabienrohrer:
`hdr.py` was not compliant with Python 3. Supporting also Python 2 is difficult and useless: #1096


fabienrohrer:
👍

scripts/image_tools/images/hdr.py
```diff
@@ -19,9 +19,11 @@
```
DavidMansolino: We should probably add the python3 shebang

fabienrohrer: done

fabienrohrer:
👍

scripts/image_tools/images/hdr.py
omichel: Was it really done? It doesn't seem so...

fabienrohrer: The Python 3 shebang has been added to the `scripts/image_tools/*py` scripts which are designed to be executed. Adding a shebang to `hdr.py` which is a Python file used by the scripts is meaningless.

So this seems to be correctly solved, yes.

PR #1100, opened 2019-11-13
Drop Python 2

fabienrohrer:
Implement #1096

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -18,6 +18,7 @@ install:
 build: off
 
 test_script:
+  - python --version
   - set WEBOTS_HOME=%APPVEYOR_BUILD_FOLDER%
   - python -m unittest discover -s %APPVEYOR_BUILD_FOLDER%\\docs\\tests
   - ps: if ($env:APPVEYOR_REPO_TAG -ne "true") { python $env:APPVEYOR_BUILD_FOLDER\\tests\\sources\\generate_diff.py $env:APPVEYOR_REPO_COMMIT $env:APPVEYOR_REPO_NAME }
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -18,7 +18,6 @@ install:
 build: off
 
 test_script:
-  - python --version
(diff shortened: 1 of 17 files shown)


fabienrohrer:
We should also certainly upgrade to Python 3.8 on every OS, but I prefer to do this in another PR, once this one is merged.

stefaniapedrazzi:
> We should also certainly upgrade to Python 3.8 on every OS, but I prefer to do this in another PR, once this one is merged.

Note that the default Python version on Ubuntu 18.04 is 3.6.
This is the one we should make sure that is supported on Linux.

stefaniapedrazzi:
Default Python version on Ubuntu 16.04 is 2.7, I'm not sure it is good to completely drop the support of python2 if we still support Ubuntu 16.04.

fabienrohrer:
The aim of this PR is also to evaluate the feasibility of this drop. I'm also still not sure this is a good idea.

We could discuss of this on our next meeting.

But for sure, I really would like to go forward with this topic. Ubuntu 16.04 could install Python 3.5 easily, and following our release logics, we will drop 16.04 support in 5 months, when 20.04 LTS will be released.

stefaniapedrazzi:
Yes, I just wanted to point out some issues that this change could cause.
I think it is good to discuss it on Monday.

> But for sure, I really would like to go forward with this topic. Ubuntu 16.04 could install Python 3.5 easily, and following our release logics, we will drop 16.04 support in 5 months, when 20.04 LTS will be released.

Of course at some point users won't never use Python 2.7. But personally I think that it is too early to do it in R2020a.
Changing the documentation is ok, but at least I would keep the precompiled libcontroller so that it can still work out-of-the box

fabienrohrer:
You're probably right, but it's annoying, because the maintenance complexity will remain for months. Anyway, I'm stopping there, and we could decide what to do on Monday. I will try to create a pros/cons list.

fabienrohrer:
It's not the time to drop Python 2 yet.

PR #1108, opened 2019-11-18
webots.min.js: Fix variable scope

stefaniapedrazzi:
If you reset the simulation while it is running, the simulation should automatically run when the reset is completed.


stefaniapedrazzi:
:+1:

PR #1114, opened 2019-11-20
webots.min.js: restore video streaming code

stefaniapedrazzi:
After porting to three.js the video streaming code is no longer working.
Here I'm restoring it while inspecting the MJPEG streaming.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ jobs:
   include:
     - &source_tests_staging
       stage: Sources Tests
+      if: type = cron OR NOT commit_message IN (travis_distrib_only)
       name: Ubuntu 16.04 / Python 3.8
       git:
         depth: 3
--- a/resources/web/streaming_viewer/index.html
+++ b/resources/web/streaming_viewer/index.html
@@ -6,14 +6,49 @@
     <title>Streaming viewer</title>
     <link rel='stylesheet' href='https://code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css'/>
     <link type="text/css" rel="stylesheet" href='https://www.cyberbotics.com/jquery-ui-layout/1.4.4/layout-default.css' />
-    <link type="text/css" rel="stylesheet" href='https://www.cyberbotics.com/wwi/R2020a/wwi.css' />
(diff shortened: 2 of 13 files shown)


fabienrohrer:
Are you sure you want to restore this feature? I think we dropped it for good reasons, the streaming of 3D data is more modern, efficient (in term of streaming resources) and gives more possibilities (like active steaming to multiple clients).

I fear to scatter our resources to maintain several streaming systems, and I don't see how to integrate this in our roadmap discussed in the last alignment meeting. For me, improving our existent system is more promising.

fabienrohrer:
Or is it just a matter to stream camera images?

stefaniapedrazzi:
Olivier proposed to give it another try using directly MJPEG streaming instead of other utilities like gstreamer.
It is working smoothly for other services like AWS and if it works for our application as well it would be a much simpler development than implement web WREN.

fabienrohrer:
I think it's a pity to not discuss this together and not try to integrate this in a bigger scope (webots.io, etc.). This development was not a part of our last alignment meeting results. The long term objectives should dictate the developments, not the contrary. I urge to create a plan for webots.io (next Monday?).

Typically, AWS uses virtual machines and existing streaming applications to stream apps. Wouldn't be a better direction?

stefaniapedrazzi:
I restored the video streaming of the simulation (this time using directly MJPEG streaming) with the simulation mode control and interactions with the scene (other functionalities like the robot windows, context menu, etc. are not working).
This is a testing version where the X3D streaming is skipped on the Webots side rather than integrating the two streaming methods, but it is enough for testing.
It can be tested using the `streaming_viewer` from this branch.
The MJPEG is streamed at a fixed port 8889.

stefaniapedrazzi:
I did some initial performance tests connecting to the server (using the streaming_viewer).
I had to set a maximum frame rate (currently set to 20 FPS just for testing) because of the latency when interacting with the scene (rotating and moving the viewpoint).

Now the video is played smoothly and the latency is quite small for me.

I will try now to make it work directly from beta.robotbenchmark.net so that it can easily be tested from any systems.

stefaniapedrazzi:
This test branch has been replaced by #1352.

Issue #1107, opened 2019-11-15
Version Check should Rely on Github


DavidMansolino:
Instead of maintaining and duplicating the current Webots version on cyberbotics.com we should probably directly use the version file from the Github repo:

https://github.com/cyberbotics/webots/blob/c4ce84200caa0725655d1538076099d2fc036a22/src/webots/core/WbWebotsUpdateManager.cpp#L50

PR #1115, opened 2019-11-20
Switch the service to retrieve the Webots version from Cyberbotics to GitHub

fabienrohrer:
Fix https://github.com/cyberbotics/webots/issues/1107
Use the GithHub release REST API: https://developer.github.com/v3/repos/releases (using raw or jsdelivr seems to be bad long term choices)

### DIFF
--- a/src/webots/core/WbWebotsUpdateManager.cpp
+++ b/src/webots/core/WbWebotsUpdateManager.cpp
@@ -78,7 +78,7 @@ void WbWebotsUpdateManager::downloadReplyFinished() {
   }
 
   if (!success) {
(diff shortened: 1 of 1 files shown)


fabienrohrer:
👍

DavidMansolino:
👀

DavidMansolino:
Not related to this PR but I just figured out that it is strange to have the 'Check for Updates' in the 'Tools' menu on Windows, it will probably suit better in the 'Help' menu (more standard, and intuitively I was looking for it in this menu).

fabienrohrer:
Very true, I'm opening a new PR with this modification right now..

PR #1116, opened 2019-11-21
Move `Check for updates...` menu

fabienrohrer:
- [x] Linux/Windows: move `Check for updates...` menu from `Tools` to `Help`
- [x] Remove the `WbGuidedTour::isAvailable()` which looks like a vestige of the license system. If the `guided_tour.txt` file is not present, it means that Webots is not installed correctly. In such case, the guided tour box can be open, but shows an "Internal error" label.


fabienrohrer:
👍

DavidMansolino:
👀

PR #1117, opened 2019-11-25
Fix Nao.selfCollision

fabienrohrer:
Reported here: https://stackoverflow.com/questions/59020010/nao-robot-sinking-into-floor-webots

Fix `Nao.selfCollision`. Internal forces were present due to the feet bounding objects causing weird shrink in floors. The touch sensors bounding objects were overlapping with the foot body.

The present Transform should be equivalent to the previous one.


fabienrohrer:
| Before | After |
| --- | --- |
| before | after |

fabienrohrer:
👍

W4hrsch3inlich:
Im facing a problem, where the bottom foot bounding object keeps lowering its position.

When i start the simulation:
Normal
After a while:
Bug

Any suggestions why this appears and how it could be fixed ?

omichel:
I cannot reproduce this with the `nao_demo.wbt` of Webots R2020b revision 1, after making the robot walk for a few steps. Can you provide clear step-by-step instructions describing how we could reproduce this problem? Also, please open a complete and precise new bug report instead of writing a message in this PR.

PR #1120, opened 2019-11-25
Escape characters before sending message to the robot window

### LINKED ISSUE #1065: String sent to robot-window not escaped
**Describe the Bug**
Most of the strings sent to the robot window (such as device name, robot name, robot model, etc.) are not escaped which causes the robot-window to fail at load.

**Steps to Reproduce**
1. Open the robotis_op3 world.
2. Change the name of the robot to `ROBOTIS' OP3`.
3. Start the simulation.
4. Open the robot-window.
5. See error:

```
[javascript] Uncaught SyntaxError: missing ) after argument list (generic.html:1)
```

stefaniapedrazzi:
Fix #1065:
the single quote `'` and backslash `\` characters need to be escaped before sending a message to the robot window.

The single quote character in the message will interfere with the message delimiters.
The backslash character is interpreted and has to be converted to the corresponding characters to avoid it (we have similar issues when using it in a QRegularExpression).

### DIFF
--- a/docs/reference/changelog-r2019.md
+++ b/docs/reference/changelog-r2019.md
@@ -22,6 +22,7 @@ Released on ???
     - Fixed displaying streaming server initialization errors in the Webots console.
     - Fixed bugs in Python Display.imageNew() when passing an image array: rearranged image data from column-major order and memory leak (thanks to Inbae Jeong).
     - Fixed [Nao.selfCollision](../guide/nao.md) due to overlapping bounding objects in feet (thanks to Sheila).
+    - Fixed errors sending messages containing single quote (') and backslash (\) to the robot windows.
 
 ## Webots R2019b Revision 1
 Released on October 3rd, 2019.
--- a/resources/projects/plugins/robot_windows/generic/generic.js
+++ b/resources/projects/plugins/robot_windows/generic/generic.js
@@ -121,18 +121,18 @@ function addDeviceType(type) {
 }
 
 function addDevice(device) {
-  var div = '<div id="' + device.name + '" class="device">';
+  var div = '<div id="' + device.htmlName + '" class="device">';
   div += '<h2>';
   if (device.type !== 'RotationalMotor' && device.type !== 'LinearMotor' && device.type !== 'DifferentialWheels')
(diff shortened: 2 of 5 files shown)


stefaniapedrazzi:
:+1:

DavidMansolino:
:eyes:

stefaniapedrazzi:
Thank you I will check to also fix the case where a device or robot name contains `"`.

This is a different problem than `'` and `\`.
The `"` is escaped and it is sent correctly to the robot window.
The issue is that it is contained in a JSON string  and thus should be prepended by an additional backslash character in order to be handled correctly.
This problem cannot be solved on the Webots side, but it is a libController issue.

stefaniapedrazzi:
Supporting double quotes in devices names in the HTML robot window is more complicated that just escaping, because we cannot use double quotes in HTML.
But in the HTML the escaped character `&quot;` is interpreted, so you need to use both the escaped and not escaped value depending if you are generating the item or requesting it.

stefaniapedrazzi:
clang-format test fails on Windows (version 9) but passed on Ubuntu and macOs.
I will try to adjust the clang-format setup file but ideally it would be better to use the same version on all the platforms..

fabienrohrer:
Actually... :-)
On macOS, I also have issues with `.clang-format-9` and `AllowAllConstructorInitializersOnNextLine`. It seems this rule is not implemented (while documented). It would be good to open an issue or PR about this..

stefaniapedrazzi:
Note that in the current clang-format-9 setup file `AllowAllConstructorInitializersOnNextLine` is also set to false.
Thus at least the current result is consistent.
I'm now trying to change this value to true and check if it works (maybe you already did and assessed that it doesn't)

DavidMansolino:
> On macOS, I also have issues with .clang-format-9 and AllowAllConstructorInitializersOnNextLine

Which version of clang-format are you using? Because on Travis when we did set this up it was working fine with this.

stefaniapedrazzi:
I triggered manually the `webots-test-sources-revision` test on this branch and now it passed.
I also installed clang-format 9 locally on my Ubuntu and tested and it is passing as well.
I don't know why it failed yesterday.

stefaniapedrazzi:
:+1:

DavidMansolino:
Unfortunately, it seems that this is causing issues on Windows:
https://ci.appveyor.com/project/omichel/webots-sj08b

PR #1122, opened 2019-11-25
Fix overwriting DISPLAY variable in simulation server.py 

stefaniapedrazzi:
Port fix from develop to revision branch.


stefaniapedrazzi:
:+1: 

Note that I had to rename the branch because a branch name ending with `.py` is wrongly interpreted as python file and check by the test_pep8 (that will fail).

PR #1123, opened 2019-11-25
Fixed PEP8 Test not Ignoring '.git' Folder

DavidMansolino:
The PEP8 test is the only test that was looking for files in this folder.
This will allow us to make branches ending with `.py` (I tested on purpose with the name of this branch) and speed up the test.


DavidMansolino:
:+1:

PR #934, opened 2019-10-01
Merge revision into develop

DavidMansolino:
Conflicts:
```
	modifié des deux côtés :  dependencies/Makefile.linux
	supprimé par eux :        projects/robots/telerob/telemax_pro/protos/TelemaxBattery.proto
	supprimé par eux :        projects/robots/telerob/telemax_pro/protos/TelemaxPro.proto
	supprimé par eux :        projects/robots/telerob/telemax_pro/protos/TelemaxTrack.proto
	supprimé par eux :        projects/robots/telerob/telemax_pro/protos/TelemaxWheel.proto
	supprimé par eux :        projects/robots/telerob/telemax_pro/protos/TelemaxZoomCamera.proto
	supprimé par eux :        projects/robots/telerob/telemax_pro/worlds/.telemax.wbproj
	supprimé par eux :        projects/robots/telerob/telemax_pro/worlds/telemax.wbt
	modifié des deux côtés :  resources/Makefile.java.include
	modifié des deux côtés :  src/install_scripts/qt_linux_installer.sh
	modifié des deux côtés :  src/packaging/files_core.txt
	modifié des deux côtés :  src/packaging/webots_distro.c
	modifié des deux côtés :  src/webots/Makefile
```

PR #1131, opened 2019-11-27
Fixed Windows Packaging

DavidMansolino:
The windows package was broken on develop since a while because of a bad merge: #934


DavidMansolino:
I just tried to create and install the package and it now works fine.

👍

DavidMansolino:
:+1:

PR #1132, opened 2019-11-28
Fix Robot.supervisor field value in curriculum/advanced_pattern_recognition.wbt

stefaniapedrazzi:
Probably when removing the Supervisor node, the supervisor capability was set to the wrong robot causing errors when running the `curriculum/advanced_pattern_recognition.wbt` simulation.


stefaniapedrazzi:
:+1:

PR #1135, opened 2019-11-28
SICK LD-MRS: change the way to manage the overlapping layers

fabienrohrer:
According [these specifications](https://cdn.sick.com/media/docs/3/03/803/Operating_instructions_LD_MRS_3D_LiDAR_sensors_en_IM0032803.PDF), the SICK LD-MRS can has supplementary and overlapping layers with a shifted and bigger range (It was not a mode of operation as I thought at first).
The current PR is about to add another internal Lidar (which can be optionally switched on) to manage this long range case.

- [x] Add a supplementary internal Lidar to manage the long-range case.
- [x] Fix the vertical offset in case of a 85° scan (9.5° to 7.5°)
- [x] Drop the `SickLdMrs.measurementLayer` which becomes pointless (the number of layers can be extracted from the `SickLdMrs.type`
- [x] Update doc accordingly: [Preview](https://cyberbotics.com/doc/guide/lidar-sensors?version=fix-ld-mrs#sick-ld-mrs)
    - [x] Warn about limitations (constant horizontal resolution and constant vertical azimuth)

### DIFF
--- a/docs/guide/lidar-sensors.md
+++ b/docs/guide/lidar-sensors.md
@@ -154,21 +154,32 @@ The `resolution` field specifies the number of points returned per layer per sca
 #### SICK LD-MRS
 
 The [SICK LD-MRS](https://www.sick.com/us/en/detection-and-ranging-solutions/3d-lidar-sensors/ld-mrs/c/g91913) is a multi-layer lidar designed for harsh outdoor environments.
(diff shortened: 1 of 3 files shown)


fabienrohrer:
👍

DavidMansolino:
:eyes:

PR #1136, opened 2019-11-28
Fix Reset of Hidden Kinematic Parameters

### LINKED ISSUE #916: The reset button doesn't reset the motor positions properly
**Steps to Reproduce**
1. Set Webots in `pause` mode.
2. Open the first world of the guided tout (Telemax PRO)
3. Press the reset button.
4. The motor positions are messed up.

**Expected behavior**
The reset button should restore the motor position like if we would use the revert button.

DavidMansolino:
Fix #916


DavidMansolino:
:+1: :heavy_check_mark:

PR #1148, opened 2019-12-03
Improve documentation about Kinematics mode

### LINKED ISSUE #929: maxForce/Torque is taken into account in kinematics mode
**Describe the Bug**

In kinematics mode, I expect that maxForce/Torque shouldn't affect the motor speed because the objects have no mass, but it seems to be the case.
Is this true?

**Steps to Reproduce**

1. Open `projects/samples/devices/worlds/motor.wbt`
2. Remove every Physics nodes of the robot, and save this simulation.
3. Playing with the maxTorque of the RotationalMotor affects the speed of the motor.

**Reference**

https://stackoverflow.com/questions/58173637/rotationalmotors-of-double-joint-robot

fabienrohrer:
Fix #929

### DIFF
--- a/docs/reference/linearmotor.md
+++ b/docs/reference/linearmotor.md
@@ -26,5 +26,3 @@ It is expressed in *Newton* (N).
 The `wb_motor_set_available_force` function can be used to change the *available force* at run-time.
 The value of `maxForce` should always be zero or positive (the default value is 10 N).
 A small `maxForce` value may result in a motor being unable to move to the target position because of its weight or other external forces.
-
-    In [*kinematics* mode](physics.md#how-to-use-physics-nodes), `maxForce` is used to clamp the joint acceleration in case `Motor.acceleration` is not defined.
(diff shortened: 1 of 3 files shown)


fabienrohrer:
👍

stefaniapedrazzi:
:eyes:

docs/reference/linearmotor.md
```diff
@@ -26,3 +26,5 @@ It is expressed in *Newton* (N).
 The `wb_motor_set_available_force` function can be used to change the *available force* at run-time.
 The value of `maxForce` should always be zero or positive (the default value is 10 N).
 A small `maxForce` value may result in a motor being unable to move to the target position because of its weight or other external forces.
+
+    In [*kinematics* mode](physics.md#how-to-use-physics-nodes), `maxForce` is used to clamp the joint acceleration in case `Motor.acceleration` is not defined.
```
stefaniapedrazzi: I'm not sure this sentence describes precisely what it happens.

Is the clamp done only if the acceleration is not defined?
from the code it seems that the clamp is applied if the acceleration is defined but it is bigger than the force/torque.
And then if the acceleration is not defined, it seems that the acceleration is not clamped but set to `maxForce`.
https://github.com/cyberbotics/webots/blob/faa4a29d04c5c21863c23d05d266fe56626671fb/src/webots/nodes/WbMotor.cpp#L299-L300

fabienrohrer: Very true, I will rephrase this.

fabienrohrer:
👍

PR #1153, opened 2019-12-05
Document login to old Webots releases

fabienrohrer:
Preview: https://www.cyberbotics.com/doc/guide/general-faq?version=feature-document-old-release-login#can-i-still-use-a-webots-version-before-the-r2019a-release


fabienrohrer:
👍

PR #1166, opened 2019-12-09
Regenerate documentation

stefaniapedrazzi:
* [x] Update objects documentation
    * [x] update the MD files
    * [x] add the images:
        * [x] "images/appearances/Copper.png" file.
        * [x] "images/appearances/CorrugatedPlates.png" file.
        * [x] "images/appearances/CorrugatedPvc.png" file.
        * [x] "images/appearances/DryMud.png" file.
        * [x] "images/appearances/FormedConcrete.png" file.
        * [x] "images/appearances/Pcb.png" file.
        * [x] "images/objects/geometries/Rectangle/model.png" file.
* [x] Generate thumbnails


DavidMansolino:
About the image, they should be generated automatically by the icon studio :wink:

stefaniapedrazzi:
:+1:

DavidMansolino:
:eyes:

docs/guide/appearances.md
```diff
@@ -177,10 +177,10 @@ Derived from [PBRAppearance](../reference/pbrappearance.md).
...
 ```
 CarpetFibers {
-  SFColor  colorOverride    1 1 1
-  SFString type             "wooly"
-  SFNode   textureTransform NULL
-  SFFloat  IBLStrength      1
+  SFColor                                   colorOverride    1 1 1
+  SFString                                  type             "wooly"
+  SFNode                                    textureTransform NULL
+  SFFloat                                   IBLStrength      1
```
omichel: Why did you add so many spaces here?

stefaniapedrazzi: This is the output of the script.

The SFString type in the original PROTO file lists the possible options {"wooly", etc.}.
If the line length is not too long, the field options characters are replaced by spaced.
Otherwise if the line is long, the option characters are removed.

I assume this was implemented to have a nice indentation both in the documentation and in the PROTO file.

DavidMansolino: If I am not wrong the script is indeed supposed to remove the SFString enumeration as it is supposed to go in the field description instead to keep the header simple in the doc.
However, the corresponding spaces are supposed to be removed from the header in the doc (maybe this is somehow broken).
But anyway this comment is completely out of the scope of this PR (which consists only on running the script and no manual edition). We should clearly accept this PR as is and open an issue about improving/fixing the script and do this AFTER the rush of the release.

stefaniapedrazzi: > However, the corresponding spaces are supposed to be removed from the header in the doc (maybe this is somehow broken).

No, I checked it some time ago and there is an explicit rule that removes the spaces only if the line is shorter than 40 characters:
https://github.com/cyberbotics/webots/blob/a782e2e0df926a3f33edac219a264562b035829b/docs/guide/generate_objects_doc.py#L131

So this was a wanted feature when implemented and not a bug.
But of course this is strange and can be modified.

omichel: Yes, please open an issue about this.

PR #1172, opened 2019-12-10
Improve doc about old license system

fabienrohrer:
❗️This PR targets `master`.

I have now more information about the supported versions. I think it's important to document this quickly.

=> Webots >= 8.0 and < 8.4 are not supported anymore because the modules packages are currently not restored.


fabienrohrer:
👍

PR #1176, opened 2019-12-10
macOS: Fix new clang warnings coming from swig

fabienrohrer:
Since the last clang update, `-Wno-maybe-uninitialized` has been replaced by `-Wno-uninitialized`. Lot of warnings are raised with the previous argument.


fabienrohrer:
👍

PR #1177, opened 2019-12-11
Improved receiver warnings

fabienrohrer:
It is required to call `wb_receiver_get_queue_length()` before `wb_receiver_get_data()`, but the warning in case this is not fulfilled is cryptic: `invalid device tag`.

The aim of this PR is simply to improve this warning to be self explanatory.

### DIFF
--- a/src/lib/Controller/api/receiver.c
+++ b/src/lib/Controller/api/receiver.c
@@ -303,7 +303,7 @@ const void *wb_receiver_get_data(WbDeviceTag tag) {
   if (rs && rs->queue)  // if queue not empty
     result = rs->queue->data;
   else if (rs)
-    fprintf(stderr, "Error: wb_receiver_get_data(): the receiver queue length should be get first.\n");
+    fprintf(stderr, "Error: wb_receiver_get_data(): the receiver queue is empty.\n");
   else  // rs->queue == NULL
     fprintf(stderr, "Error: wb_receiver_get_data(): invalid device tag.\n");
   robot_mutex_unlock_step();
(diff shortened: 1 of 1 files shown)


fabienrohrer:
👍

src/lib/Controller/api/receiver.c
```diff
@@ -302,7 +302,9 @@ const void *wb_receiver_get_data(WbDeviceTag tag) {
   Receiver *rs = receiver_get_struct(tag);
   if (rs && rs->queue)  // if queue not empty
     result = rs->queue->data;
-  else
+  else if (rs)
+    fprintf(stderr, "Error: wb_receiver_get_data(): the receiver queue length should be get first.\n");
```
DavidMansolino:
```suggestion
    fprintf(stderr, "Error: wb_receiver_get_data(): the receiver queue is empty.\n");
```

fabienrohrer:
Much better, thank you!

👍

DavidMansolino:
> Why not making nested tests instead? That would be more straight forward to understand.

This is only my point of view, but this is much more cleaner than nested conditions. In general nested conditions and loops tend to make the code more difficult to understand and error prone.

fabienrohrer:
Very subjective to me and pointless. I'm going forward.

omichel:
Sorry, I don't agree on that one.

1. This test is really unnecessarily complicated.
2. Why do we need to test twice the same value (e.g., `rs`)?
3. This test is inconsistent with other similar tests in the same file.

I will submit a better fix for this.

PR #1180, opened 2019-12-11
Fix Reset

### LINKED ISSUE #1178: Reset button does not restart the controllers
Clicking the reset button on the GUI restarts smoothly the simulation, but the controllers are not restarted. Certainly related with the modifications on the supervisor functions.

DavidMansolino:
Fix #1178.

This issue was caused bedause the `QAction::triggered` signal has an optional argument that is causing a problem with the optional argument of the slot.

### DIFF
--- a/src/webots/app/WbApplication.hpp
+++ b/src/webots/app/WbApplication.hpp
@@ -60,7 +60,7 @@ public:
   void worldReload();
 
   // reset the simulation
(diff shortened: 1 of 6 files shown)


DavidMansolino:
:+1: :heavy_check_mark:

fabienrohrer:
👀

PR #1181, opened 2019-12-11
Fix memory leak

fabienrohrer:
- [x] Fix memory leak in the camera recognition algorithm, causing a memory tsunami in worlds using this feature :-)


fabienrohrer:
I run Apple memory leak Instruments in some worlds of the guided tour and I didn't found other major memory leak (it's difficult to guarantee there is no one left).

👍 ✔️

PR #1187, opened 2019-12-12
Improve rendering of soccer worlds

### LINKED ISSUE #1182: darwin-op/soccer.wbt only partially working
Probably because of the change in the world appearance, the darwin-op soccer.wbt simulation is not fully working.

The robot kicks the ball once, but then it is not able to detect the ball and it continues to turn around.
Previously it was able to detect the ball and move to it to kick it again.

fabienrohrer:
Soccer worlds have saturated lights with exaggerated shadows and low reflections.
This causes issues in the color detection algorithm of the DARwIn-OP (the dark side of the ball is too different from its bright side).

- [x] Fix #1182
- [x] Reduce the aldebo of the soccer texture (exaggeratedly saturated in my point of view)
- [x] Reduce its dimension, pointlessly high
- [x] Choose a background with more reflections
- [x] Adapt light and background luminosity


fabienrohrer:
| Before | After |
| --- | --- |
| before | after_2 |

fabienrohrer:
👍

DavidMansolino:
:eyes:

Issue #1162, opened 2019-12-09
webots.min.js: highway_driving benchmark is running very slow on Firefox


stefaniapedrazzi:
The `highway_driving` benchmark simulation with webots.min.js based on three.js is running very slow on Firefox (70.0.1 (64-bit)) but it runs much faster on Chrome (Version 78.0.3904.97 (Official Build) (64-bit)).

We reported other issues on Firefox not reproducible on Chrome (see https://github.com/omichel/robotbenchmark/issues/360).

stefaniapedrazzi:
Even after fixing the HDR reflections the issue with the speed of the highway_driving benchmark is not fixed.

stefaniapedrazzi:
I can make it run smoothly on Firefox by skipping some rendering, for example rendering the scene at most every 20 ms.
I still have to inspect if on Firefox the scene is rendered more often than in Chrome and if yes where does this come from.

stefaniapedrazzi:
I'm testing again and the issue is not completely fixed.
The first simulation I load runs very slow.
Any simulation started later runs in real-time.

To debug I also tried to connect with VNC to the simulation server, but as soon as I connect the simulation speed increases.

I can reproduce the issue both on Chrome/Linux and Firefox/Linux and also with any benchmark (tested visual_tracking, pick_and_place, inverted_pendulum).
I will try to debug with an uncompressed version of webots.min.js.

stefaniapedrazzi:
The web simulation runs slowly because it receives updates with a big time step.
I though it could be a problem with a specific port (the first one chosen).
But now I cannot reproduce the issues.

I will try again later.

stefaniapedrazzi:
The problem was caused by the screen lock of the simulation server machine.

PR #1188, opened 2019-12-12
webots.min.js: hotfix rendering frequency

stefaniapedrazzi:
Addresses #1162: on Firefox the render function is called too often causing a slow down of the scene update.

All the system will benefit from having a maximum rendering frequency and skipping the rendering if called too often.
To avoid missing some single render call, I also added a timeout.


stefaniapedrazzi:
:+1:

fabienrohrer:
👀

stefaniapedrazzi:
Ok, thank you.
I will add the comment and the constant.

resources/web/wwi/x3d_scene.js
```diff
@@ -86,6 +86,12 @@ class X3dScene { // eslint-disable-line no-unused-vars
   }
 
   render() {
+    if (this.nextRenderingTime && this.nextRenderingTime > (new Date()).getTime()) {
+      if (!this.renderingTimeout)
+        this.renderingTimeout = setTimeout(() => this.render(), 40);
```
omichel: I was wondering if we shouldn't use the exact delay here instead of 40, e.g.
```
const dt = this.nextRenderingTime - (new Date()).getTime();
```

stefaniapedrazzi: Yes, I can change it.
But the timeout of 40 is so small that I doubt there is a visual difference.

omichel: I agree, there won't probably be any visual difference, but it looks better to me to use the actual value of the delay rather than assuming it's 40 ms. Do you agree?

stefaniapedrazzi: yes

stefaniapedrazzi:
:+1:

resources/web/wwi/x3d_scene.js
```diff
@@ -86,6 +86,18 @@ class X3dScene { // eslint-disable-line no-unused-vars
   }
 
   render() {
+    // Set maximum rendering frequency.
+    // To avoid slowing down the simulation rendering the scene too often, the last rendring time is checked
```
omichel:
```suggestion
    // To avoid slowing down the simulation rendering the scene too often, the last rendering time is checked
```

PR #1192, opened 2019-12-13
Cleanup

omichel:
Better fix for #1177:

1. Made all the tests on the device tag simple and consistent with each other.
2. Systematic of the `__FUNCTION__` macro to avoid typos in function names.
3. Fixed another couple of similar problems where the device tag was not tested properly.
4. Fixed double initialized variables.
5. Removed unnecessary nested tests.


omichel:
👍

omichel:
✔️

src/lib/Controller/api/receiver.c
```diff
@@ -191,145 +191,162 @@ void wb_receiver_init(WbDevice *d) {
...
   Receiver *rs = receiver_get_struct(tag);
-  if (rs) {
+  if (!rs)
+    fprintf(stderr, "Error: %s(): invalid device tag.\n", __FUNCTION__);
+  else {
     rs->enable = true;
     rs->sampling_period = sampling_period;
-  } else
-    fprintf(stderr, "Error: wb_receiver_enable(): invalid device tag.\n");
+  }
```
DavidMansolino: What is the advantage of inverting the condition here?

src/lib/Controller/api/receiver.c
```diff
@@ -191,145 +191,162 @@ void wb_receiver_init(WbDevice *d) {
 
 void wb_receiver_enable(WbDeviceTag tag, int sampling_period) {
   if (sampling_period < 0) {
-    fprintf(stderr, "Error: wb_receiver_enable() called with negative sampling period.\n");
+    fprintf(stderr, "Error: %s() called with negative sampling period.\n", __FUNCTION__);
```
DavidMansolino: Unfortunately, this introduces a non-homogeneity with the other devices, ideally, we should change this in the whole C API.

src/lib/Controller/api/receiver.c
omichel: In general, it is always better to rule out bad cases (errors) before addressing the normal case. This is also more consistent with similar tests in this file.

src/lib/Controller/api/receiver.c
omichel: Yes, that would be a good practice. I can do it in this PR.

src/lib/Controller/api/receiver.c
DavidMansolino: > In general, it is always better to rule out bad cases (errors) before addressing the normal case.

This is clearly true if the negative implies a return, in other cases that's more arguable, but this is a detail.

src/lib/Controller/api/receiver.c
DavidMansolino: That would be nice, thank you.

src/lib/Controller/api/compass.c
```diff
@@ -135,10 +135,13 @@ const double *wb_compass_get_values(WbDeviceTag tag) {
   Compass *compass = compass_get_struct(tag);
   if (compass) {
     if (compass->sampling_period <= 0)
-      fprintf(stderr, "Error: wb_compass_get_values() called for a disabled device! Please use: wb_compass_enable().\n");
+      fprintf(stderr,
+              "Error: %s() called for a disabled device! Please use: "
+              "wb_compass_enable().\n",
+              __FUNCTION__);
```
DavidMansolino: Same here.

src/lib/Controller/api/robot.c
```diff
@@ -726,19 +736,23 @@ WbDeviceTag wb_robot_get_device_by_index(int index) {
   if (index >= 0 && index < wb_robot_get_number_of_devices())
     return (WbDeviceTag)index + 1;  // the first item is the robot and not a device
   else {
-    fprintf(stderr, "Error: The index of wb_robot_get_device_by_index() is out of the bounds.\n");
+    fprintf(stderr,
+            "Error: %s(): the index of wb_robot_get_device_by_index() is out "
+            "of the bounds.\n",
+            __FUNCTION__);
```
DavidMansolino: This seems wrong.

src/lib/Controller/api/distance_sensor.c
```diff
@@ -112,7 +112,7 @@ void wb_distance_sensor_init(WbDevice *d) {
```
DavidMansolino: One is missing line 98:
```
void wbr_distance_sensor_set_value(WbDeviceTag t, double value) {
  DistanceSensor *ds = distance_sensor_get_struct(t);
  if (ds)
    ds->value = value;
  else
    fprintf(stderr, "Error: wbr_distance_sensor_set_value(): invalid device tag.\n");
}
```

omichel:
I believe I fixed all the line-length related problems.
👍

PR #1194, opened 2019-12-16
Fix background export of robot component.

### LINKED ISSUE #1193: Avoid exporting default background in web_component_studio
**Describe the Bug**
It is not required to export the default (mountain) background for the robot component. The `web_component_studio` script should therefore be improved to not export it.

DavidMansolino:
Fix #1193


DavidMansolino:
:+1:

PR #1196, opened 2019-12-16
Rename 'physics' fields into 0enablePhysics'

### LINKED ISSUE #1195: Consistency in PROTO field names for physics
It seems that most of the time the field name `enablePhysics` is used, but there are a couple of PROTO nodes that uses `physics` instead.
If possible it would be better to always use the same field name.

Here are the few sensors using `physics`:
* Robotis LDS-01
* SickLdMrs
* MultiSenseS21

Given that it is new, maybe it would be good to fix at least the Robotis LDS-01.

DavidMansolino:
Fix #1195

### DIFF
--- a/docs/guide/appearances.md
+++ b/docs/guide/appearances.md
@@ -1373,7 +1373,7 @@ RoughPolymer {
 
 ### RoughPolymer Field Summary
 
-- `baseColor`: Defines the base color of the materil.
+- `baseColor`: Defines the base color of the material.
 
 - `textureTransform`: Defines an optional 2d texture transform.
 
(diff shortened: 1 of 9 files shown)


docs/guide/appearances.md
```diff
@@ -1373,7 +1373,7 @@ RoughPolymer {
 
 ### RoughPolymer Field Summary
 
-- `baseColor`: Defines the base color of the materil.
+- `baseColor`: Defines the base color of the material.
```
DavidMansolino: Not directly related to this PR but I saw this while regeneratign the doc.

DavidMansolino:
:+1:

fabienrohrer:
👀

fabienrohrer:
I just checked successfully that these fields were not used:

```
 find . -name "*.wbt" | xargs grep physics | grep FALSE
```

PR #1200, opened 2019-12-16
Update TurtleBot3 Burger splash screen to include the lidar.

DavidMansolino:
Update TurtleBot3 Burger splash screen to include the lidar.


DavidMansolino:
:+1:

PR #1209, opened 2019-12-17
Fix Visual Studio Controller Launch

### LINKED ISSUE #1179: Launching Controller from Visual Studio is not Working
**Describe the Bug**
Using the wizard it is possible to create a controller which then compiles fine with Visual Studio but then when trying to start it from Visual Studio (as an extern controller) it fails to start the controller:
Capture

I am almost sure that this was working with the Webots R2019b.

DavidMansolino:
Fix #1179


DavidMansolino:
(I just launched a package creation from this branch to make sure it is working fine).

DavidMansolino:
I have checked the package and it now works perfectly with Visual Studio, even when Visual Studio is not launched by Webots.

DavidMansolino:
👍

PR #1214, opened 2019-12-19
Avoid including XC16 in the snap release.

DavidMansolino:
The XC16 folder was not included in the latest snap released (removed manually), this PR fixes the issue once for all.


DavidMansolino:
:+1:

PR #1223, opened 2019-12-20
Fix libpng dependency

DavidMansolino:
Remove all the libController dependencies (libtiff, libpng and libjpeg).


DavidMansolino:
:+1 :heavy_check_mark: (tested on all the OS, release available here: https://github.com/cyberbotics/webots/releases/tag/untagged-12f168001778e7a5527c)

src/lib/Controller/util/g_image.c
```diff
@@ -298,129 +135,101 @@ static int g_image_png_save(GImage *img, const char *filename) {
...
 }
 
+struct ImageData {
+  unsigned char **target_data;
+  unsigned long *target_data_size;
+};
+
+void g_image_save_to_jpeg_buffer_callback(void *context, void *data, int size) {
+  if (!*(((struct ImageData *)context)->target_data))
+    *(((struct ImageData *)context)->target_data) = (unsigned char *)malloc(size);
```
omichel: When I see a malloc and no corresponding free, I fear a memory leak... Are you sure this memory is free? Where? Can you add a comment in the code about it?

src/lib/Controller/Makefile
```diff
@@ -33,38 +33,24 @@ CFLAGS      = -fPIC -Wall
...
+SHARED_LIBS = -lm -lpthread -ldl -llzma -lrt
 TARGET      = $(WEBOTS_CONTROLLER_LIB_PATH)/libController.so
 endif
 
 ifeq ($(OSTYPE),windows)
 CFLAGS      = -Wall -mwindows
 INCLUDE     = -I$(WEBOTS_PATH)/include/controller/c -I$(WEBOTS_PATH)/include -I/mingw64/include
 LD_FLAGS    = -shared -mwindows --def Controller.def
-STATIC_LIBS = -static -lpng -lz -ljpeg -ltiff -llzma -lzstd
+STATIC_LIBS = -static -lz -llzma -lzstd
```
omichel: I believe these libraries are dependencies of png, jpeg or tiff and we should remove them as well.

src/lib/Controller/util/g_image.c
DavidMansolino: Yes I am sure that the memory is free. As the user of this function has to provide the pointer he is then responsible to free it (this is not new and was already working the exact same way before). This is done here: https://github.com/cyberbotics/webots/blob/master/src/lib/Controller/api/default_robot_window.c#L507
But I will add a note explaining this in the header file next to the function.

src/lib/Controller/Makefile
DavidMansolino: You're right I tested and it seems we don't need to link with them anymore (but we still need to include them in the package as Webots still relies on them).

DavidMansolino:
:+1: :heavy_check_mark:

DavidMansolino:
Wiki updated: https://github.com/cyberbotics/webots/wiki/Linux-installation/_compare/aeeecf0ccfc1a64eba4dca709a7f4f0558f4e7bb...34439308e82f9de811cb0220a956fd71daa25e7b

omichel:
Also here: https://github.com/cyberbotics/webots/wiki/Windows-Optional-Dependencies/_compare/0e7c6b29a8042852a7c30b23da10ebf95eb61694...e628a435f59ba79c80ef84c164d9b95e785e4b46

DavidMansolino:
Thank you, I missed this one.

PR #1227, opened 2020-01-03
Update copyright dates.

DavidMansolino:
Update the date to 2020 in the copyrights.
This was the cause of the nightly test failures.


DavidMansolino:
:+1:    (some issues in some test are remaining but this PR is already quite long in term of modified files).

PR #1228, opened 2020-01-03
Fixed unclosed files in test.

DavidMansolino:
The tests are sometimes failing with Python3 because of unclosed files.


DavidMansolino:
:+1:   (to be merged after #1227)

PR #1230, opened 2020-01-03
Fix cppcheck failure

DavidMansolino:
An update of cppcheck on Windows has raised some new issues.


DavidMansolino:
(I can't reproduce anymore)

PR #1232, opened 2020-01-03
Remove Problematic xvfb Hack.

DavidMansolino:
This hack was causing problem when compiling the snap package on travis:
https://travis-ci.com/cyberbotics/webots-snap/jobs/272051733#L6574

Furthermore, removing it seems good because it remove snapcraft specific compilation instructions/code.


DavidMansolino:
:+1:

PR #1244, opened 2020-01-08
Remove OpenCV

### LINKED ISSUE #911: Drop OpenCV release
It is not our responsibility to release libraries such as OpenCV.
We should not release it anymore, it requires maintenance and since it is in the lib/controller folder it does pollute the controller environment.

Instead of a sample using OpenCv we should write a clear how-to doc about integrating external libraries and showcase it using open-cv.

DavidMansolino:
Fix #911


DavidMansolino:
:+1: :heavy_check_mark:

PR #1246, opened 2020-01-08
Fixed SUMO and ROS dependencies makefiles.

DavidMansolino:
These makefiles where not removing the archive once downloaded and exctracted (worse the archive was included in the release).


DavidMansolino:
:+1: :heavy_check_mark:

PR #1248, opened 2020-01-09
Fix wb_supervisor_add_force_with_offset()

fabienrohrer:
The offset is wrongly computed, probably because of the ODE rotation which is not correctly taken into account. Using `dBodyAddForceAtPos()` instead of `dBodyAddForceAtRelPos()` simplifies and fixes this issue.

### DIFF
--- a/docs/reference/changelog-r2020.md
+++ b/docs/reference/changelog-r2020.md
@@ -7,6 +7,7 @@ Released on XXX YYYrd, 2020.
     - Improved the SUMO interface to use the version of SUMO already installed (if any).
     - Improved the SUMO interface to be compatible with the latest version (1.4).
   - Bug fixes
+    - Fix the [wb_supervisor_node_add_force_with_offset()](supervisor.md#wb_supervisor_node_add_force_with_offset) function.
     - Fix default controller of the [ABB IRB 4600/40](../guide/irb4600-40.md) robot.
   - Cleanup
     - OpenCV is not distributed within Webots anymore.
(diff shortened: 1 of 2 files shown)


fabienrohrer:
[✔️](https://travis-ci.com/cyberbotics/webots/builds/143698198)👍

docs/reference/changelog-r2020.md
```diff
@@ -7,6 +7,7 @@ Released on XXX YYYrd, 2020.
     - Improved the SUMO interface to use the version of SUMO already installed (if any).
     - Improved the SUMO interface to be compatible with the latest version (1.4).
   - Bug fixes
+    - Fix the [`wb_supervisor_node_add_force_with_offset()`](supervisor.md#wb_supervisor_node_add_force_with_offset) function.
```
DavidMansolino:
```suggestion
    - Fixed the direction of the force applied with the [`wb_supervisor_node_add_force_with_offset()`](supervisor.md#wb_supervisor_node_add_force_with_offset) function.
```

src/webots/nodes/utils/WbSupervisorUtilities.cpp
```diff
@@ -794,13 +794,17 @@ void WbSupervisorUtilities::handleMessage(QDataStream &stream) {
       WbNode *const node = getProtoParameterNodeInstance(WbNode::findNode(id));
       WbSolid *const solid = dynamic_cast<WbSolid *>(node);
       if (solid) {
-        WbVector3 force(fx, fy, fz), offset(ox, oy, oz);
+        const WbMatrix4 &solidMatrix = solid->matrix();
+
+        WbVector3 offset = solidMatrix * WbVector3(ox, oy, oz);
```
omichel:
```suggestion
        const WbVector3 offset = solidMatrix * WbVector3(ox, oy, oz);
```

fabienrohrer:
👍

fabienrohrer:
> Thank you, this seems much cleaner, but what about the C_SUPERVISOR_NODE_ADD_FORCE case?

Just saw this comment.. This case should work smoothly, the computation of the offset is much simpler in this case and seems to be correctly done.

PR #1256, opened 2020-01-10
Improve Hall Environment

DavidMansolino:
Improve the hall world according to omichel/cyberbotics#31


DavidMansolino:
:+1:

DavidMansolino:
These DEF SHAPE / USE SHAPE are made to be used in the Bounding object (which is the case), the shape are most of the time not equivalent from on solid to the other.
But I just saw that there are some appearances that can be defined.

DavidMansolino:
:+1:

PR #1272, opened 2020-01-14
Improve Motion-Blur Sample.

DavidMansolino:
Given the simplicity of this simulation and the low resolution of the camera, we can afford to decrease the time step to have a better motion-blur effect.
Master:
camera_motion_blur_1
This branch:
camera_motion_blur


DavidMansolino:
:+1:

PR #1275, opened 2020-01-15
Added missing units to parameter descriptions

### LINKED ISSUE #1260: UI/API: Unit Declaration
For the most part, the API/guide is well written and structured. However, sometimes the default values for properties are nowhere stated.
For example, the position sensor's resolution field simply has a float value, but not even the documentation of the position sensor specifies what unit is used. Other properties like a motor's minPosition and maxPosition also state no unit in the guide.
Ideally, each field would simply state the unit directly in app's UI.

omichel:
Fixes #1260.

### DIFF
--- a/docs/reference/motor.md
+++ b/docs/reference/motor.md
@@ -58,7 +58,7 @@ With a small *P*, more simulation steps are needed to reach the target position,
     `wb_motor_set_control_pid` function.
 
 - The `minPosition` and `maxPosition` fields specify *soft limits* for the target position.
-They are expressed in *meters* (m) for linear motors and in *radians* (rad) for rotational motors.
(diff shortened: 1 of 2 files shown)


omichel:
👍

docs/reference/motor.md
```diff
@@ -58,6 +58,7 @@ With a small *P*, more simulation steps are needed to reach the target position,
     `wb_motor_set_control_pid` function.
 
 - The `minPosition` and `maxPosition` fields specify *soft limits* for the target position.
+They are expressed in *meters* (m) for linear motors and in *radians* (rad) for rotational motors.
```
DavidMansolino:
```suggestion
They are expressed in *meters* (m) for [LinearMotor](linearmotor.md) and in *radians* (rad) for  [RotationalMotor](rotationalmotor.md) .
```

PR #1276, opened 2020-01-16
Improve Supervisor set/get Velocity function

DavidMansolino:
A user reported that this was not clear in the documentation.


DavidMansolino:
:+1:

PR #1283, opened 2020-01-20
Fixed red planet color

### LINKED ISSUE #1269: sojourner.wbt doesn't look like Mars any more
Because of recent changes in the textures / background, the ground color of sojourner.wbt is not red any more, but grey. It's a pity as it should refer to the red planet (Mars)...

omichel:
Fix #1269.
Before this PR:
sojourner_before
After this PR:
sojourner_new

### DIFF
--- a/projects/objects/rocks/protos/Rock10cm.proto
+++ b/projects/objects/rocks/protos/Rock10cm.proto
@@ -8,7 +8,7 @@ PROTO Rock10cm [
   field SFRotation rotation         0 1 0 0
   field SFString   name             "rock 10 cm"
   field MFString   texture          "textures/rock.jpg"  # Defines the texture used for the rock.
-  field SFColor    color            1 1 1                # Is `Solid.chidlren.Shape.appearance.baseColor`.
+  field SFColor    color            1 1 1                # Defines the base color of the rock
   field SFNode     physics          NULL                 # Is `Solid.physics`.
   field SFBool     locked           FALSE                # Is `Solid.locked`.
 ]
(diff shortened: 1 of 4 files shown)


omichel:
👍

projects/objects/rocks/protos/Rock10cm.proto
```diff
@@ -8,6 +8,7 @@ PROTO Rock10cm [
   field SFRotation rotation         0 1 0 0
   field SFString   name             "rock 10 cm"
   field MFString   texture          "textures/rock.jpg"  # Defines the texture used for the rock.
+  field SFColor    color            1 1 1                # Is `Solid.chidlren.Shape.appearance.baseColor`.
```
DavidMansolino: This comment does not follow our PROTO guidelines:
> If the field is connected directly to a **root** node with an IS, the comment should be # Is RootNodeType.fieldname.

omichel:
👍

projects/objects/rocks/protos/Rock10cm.proto
```diff
@@ -8,6 +8,7 @@ PROTO Rock10cm [
   field SFRotation rotation         0 1 0 0
   field SFString   name             "rock 10 cm"
   field MFString   texture          "textures/rock.jpg"  # Defines the texture used for the rock.
+  field SFColor    color            1 1 1                # Defines the base color of the rock
```
stefaniapedrazzi: Just a minor note: here the final `.` is missing in the field description

DavidMansolino: I will add a test for this in our test suite :wink:

DavidMansolino:
This planet is actually not so red:
Capture du 2020-07-27 08-31-17
Left: the first image published by the NASA that made people think the planet is red. Right: actual color of the planet.
> https://www.numerama.com/sciences/638963-mars-la-planete-pas-si-rouge-que-ca.html

PR #1285, opened 2020-01-20
Fix Camera Image Save to JPEG

DavidMansolino:
When saving a BGRA image to jpeg the resulting image was wrong, this is visible in the robot-window of the e-puck2.wbt world.


DavidMansolino:
:+1: :heavy_check_mark:

PR #1288, opened 2020-01-21
Add a Stadium Background

DavidMansolino:
Before:

After:

Not use with the robotis-op2 soccer world because it messes up the image processing of the controller.


fabienrohrer:
omg, thank you for this: https://github.com/cyberbotics/webots/pull/1288/commits/c2ba4a79c38b93eb3411b9b01a3d57b3d1e53bb9 😅

DavidMansolino:
:+1:

stefaniapedrazzi:
It looks better.
But please check that the robotis-op2 soccer simulation still works after modifying the color of the field because we recently had problems with this simulation that was broken.

omichel:
I just checked it. The stadium background was missing and I added it. Now the robot gets finally attracted by the orange color of the seats in the stadium... But for the first few seconds, it can see the ball and can kick it. I am playing with the initial position of the robot to improve it a little bit...

omichel:
I finally removed the stadium background from the dawin-op soccer demo otherwise it quickly looses the ball and is attracted by the orange color of the stadium. I adjusted the initial position of the robot and viewpoint so that it makes a nice course.

PR #1301, opened 2020-01-27
Improve support for mail servers in session server script

stefaniapedrazzi:
**Description**
Some parameters are missing in the session server script configuration file to use mail servers that need authentication. 

**Related Issues**
This pull-request fixes issue https://github.com/omichel/robotbenchmark/issues/362

**Tasks**
  - [x] Add config parameters
  - [x] Update documentation

**Documentation**
https://cyberbotics.com/doc/guide/web-simulation?version=enhancement-session-server-script#session-server


stefaniapedrazzi:
I think it is quite common that the email address can also be used as username.
This is for example also the case for gmail.

But I can add an optional username parameter, and if not set use the sender address.

PR #1303, opened 2020-01-27
Documentation: fix extracting tabs if version is not specified

### LINKED ISSUE #1296: Broken docs Links with Tab
**Describe the Bug**
When entering a link to the doc with the tab-os or tab-language set, the page is not found and the index is displayed:
https://cyberbotics.com/doc/guide/running-extern-robot-controllers?tab-os=macos&tab-language=python

stefaniapedrazzi:
Fix #1296.

The code to extract queries if the version is not specified has some issues:
* `$version` is not defined 
* regular expression should be adjusted

Moreover the code to parse the query is quite complex.
Using PHP built-in functions to parse the URL and the query string it is possible to simplify it.

URL to check the functionality (after uploading the new version on the server):
https://cyberbotics.com/doc/guide/running-extern-robot-controllers?tab-os=macos&tab-language=python

### DIFF
--- a/docs/doc.php
+++ b/docs/doc.php
@@ -61,7 +61,7 @@
   if ($branch === '') {
     # get HEAD commit SHA, to ensure that when master is updated the latest version is cached by the CDN
     ini_set('user_agent', $repository); # every GitHub request needs a valid user agent header
-    $githubHead = file_get_contents("https://api.github.com/repos/cyberbotics/webots/git/refs/heads/master");
+    $githubHead = @file_get_contents("https://api.github.com/repos/cyberbotics/webots/git/refs/heads/master");
     // failed request / github is down
     if ($githubHead === FALSE)
       $cacheUrl = "https://cdn.jsdelivr.net/gh/$repository/webots@master";  // fall back to dev URL at worst
--- a/docs/doc.php
+++ b/docs/doc.php
@@ -16,7 +16,7 @@
     exit();
   }
   if ($request_uri == "/doc/blog/index") {
-    header("Location: /doc/blog/Webots-2019-b-release");
+    header("Location: /doc/blog/Webots-2020-a-release");
     exit();
   }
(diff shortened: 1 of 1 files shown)


docs/doc.php
```diff
@@ -16,60 +16,33 @@ function endsWith($haystack, $needle) {
     exit();
   }
   if ($request_uri == "/doc/blog/index") {
-    header("Location: /doc/blog/Webots-2019-b-release");
+    header("Location: /doc/blog/Webots-2020-a-release");
```
stefaniapedrazzi: This link should be automatically updated (when updating the release version) or we should document to update it just before the release.

stefaniapedrazzi:
I think it would be good to test this PR directly in production.
I can upload the new version on the website when the reviewer is ready.

docs/doc.php
omichel: Normally, it's done here: https://github.com/cyberbotics/webots/blob/master/src/bin/new_version.sh#L51
Do you think this doesn't work?

omichel:
Can you please upload it on the server so that I can test it now?

stefaniapedrazzi:
I just uploaded it.

docs/doc.php
stefaniapedrazzi: I will check if it works.

stefaniapedrazzi: I checked and it works fine (also on the develop branch it is correctly set).
So then I don't know why it was wrongly set here.

omichel:
I just noticed that the web page was displaying a warning. I fixed it in 25e4d3f54b228dbdb9d31f2301155c67392f8c29. This was already fixed in the previous `doc.php` online, but apparently this patch was not committed on github.

stefaniapedrazzi:
I also found out that I deleted the instruction setting the repository in case of fork.

stefaniapedrazzi:
If I try to load a documentation with an invalid version I'm still getting an error that quickly disappears, is it this the error that you made silent or was like this even before?

omichel:
No, I silenced another error that was happening on the master branch. We could also silence this one the same way, but it's less important to me.

PR #1306, opened 2020-01-28
Fix WGET issue with language on macOS.

DavidMansolino:
**Description**
If the `LANG` is not set, wget on Mac somehow tries to compute it from the computer location and language, which in my case gives: `en_CH.UTF-8` which doesn't exists and therefore raises many warnings.


fabienrohrer:
👀

resources/web/wwi/Makefile
```diff
@@ -53,6 +53,17 @@ THEEEJS_VERSION = 104
...
+WGET=wget -qN -O
+
+null :=
+space := $(null) $(null)
+WEBOTS_HOME_PATH=$(subst $(space),\ ,$(strip $(subst \,/,$(WEBOTS_HOME))))
+include $(WEBOTS_HOME_PATH)/resources/Makefile.os.include
+
+ifeq ($(OSTYPE),darwin)
+ WGET=LANG=en_US.UTF-8 wget -qN -O
+endif
```
omichel: I would prefer that we don't override variables if possible, that makes the code more readable, e.g.:
```
ifeq ($(OSTYPE),darwin)
 WGET=LANG=en_US.UTF-8 wget -qN -O
else
  WGET=wget -qN -O
endif
```

resources/web/wwi/Makefile
```diff
@@ -103,13 +114,13 @@ webots.lib.js:
...
 	@mkdir -p $(DEPENDENCIES_PATH)
-	@wget -qN -O $@ $(THREEJS_URL)
+	@$(WGET) $@ $(THREEJS_URL)
 
 $(LOCAL_THREEJS_EXAMPLE_SOURCES):
 	$(eval FILENAME=$(@:$(DEPENDENCIES_PATH)/%=%))
 	@echo "# downloading three.js $@"
 	@mkdir -p $(dir $@)
-	wget -qN -O $@ $(THREEJS_EXAMPLE_URL)$(FILENAME)
+	$(WGET) $@ $(THREEJS_EXAMPLE_URL)$(FILENAME)
```
fabienrohrer: This line should be silenced:

```suggestion
	@$(WGET) $@ $(THREEJS_EXAMPLE_URL)$(FILENAME)
```

PR #1311, opened 2020-01-29
Support Python 3.8 on Windows

omichel:
It was quite tricky to fix this one, but I believe it is now done the right way.
I updated the [wiki](https://github.com/cyberbotics/webots/wiki/Windows-Optional-Dependencies/_compare/e628a435f59ba79c80ef84c164d9b95e785e4b46...84206c8ebf40a1b147a9cac2349efe966a6d9d80) accordingly.


resources/languages/python/controller.i
```diff
@@ -22,6 +22,18 @@
 #define SWIG_PYTHON_2_UNICODE
 %}
 
+%pythonbegin %{
+import sys
+import os
+if os.name == 'nt' and sys.version_info >= (3, 8, 0):  # we need to explicitly list the folders containing the DLLs
```
DavidMansolino:
```suggestion
if os.name == 'nt' and sys.version_info >= (3, 8):  # we need to explicitly list the folders containing the DLLs
```

omichel:
✔️ [appveyor webots-nightly-master](https://ci.appveyor.com/project/omichel/webots-master/builds/30453877)

PR #1312, opened 2020-01-29
Fixed flake8 warnings

omichel:
Flake8 doesn't like naming a variable `l`...

### DIFF
--- a/src/packaging/copy_msys64_files.py
+++ b/src/packaging/copy_msys64_files.py
@@ -48,8 +48,7 @@ skip_paths = ['/usr/share/', '/mingw64/bin/zlib1.dll']
 for p in dependencies:
     print("# processing " + p)
     sys.stdout.flush()
-    l = subprocess.check_output(['pacman', '-Qql', p]).decode().strip().split('\n')
-    for f in l:
+    for f in subprocess.check_output(['pacman', '-Qql', p]).decode().strip().split('\n'):
         skip = False
         for g in skip_paths:
             if f.startswith(g):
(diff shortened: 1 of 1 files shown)


DavidMansolino:
:eyes:

omichel:
The warnings are visible in atom:
image
Not sure what else you want to improve in this script. Can you elaborate?

DavidMansolino:
Thank you for the explanation, I have update my flake8 module and can now see them in atom. We should probably add this to the test suite.

About the improvements, to me, this code looks a bit like if we are writing C style in Python. Here are the main issues I see:
```
d = list_dependencies('make')
d += list_dependencies('coreutils')
d += list_dependencies('mingw-w64-x86_64-gcc')
d += list_dependencies('mingw-w64-i686-gcc')
# remove duplicate packages
seen = set()
dependencies = []
for item in d:
    if item not in seen:
        seen.add(item)
        dependencies.append(item)
```
This part is over-complicated and can be replaced by something like:
```
dependencies = list(set(  # The use of a set make sur to avoid duplication
    list_dependencies('make') +
    list_dependencies('coreutils') +
    list_dependencies('mingw-w64-x86_64-gcc') +
    list_dependencies('mingw-w64-i686-gcc')
))
```

*** 

Many prints are followed by a flush:
```
    print("# processing " + p)
    sys.stdout.flush()
```
This is again not the Python way to do, here is a more simple version to makes sure the print is flushed (which is maybe even not needed):
```
print('# processing ' + p, flush=True)
```

***

One of the file manipulations is encapsulated in a `with open() as f:` which is good, but the second one is wrong.

***

Several final `/` are removed with this instructions `[:-1]`, this is not immediately understandable and can cause problem if one of the paths is missing the `/`, instead, it can easily be replaced by `rstrip('/')`

***

One of the `subprocess.check_output` is nicely encapsulated in a function that makes it simple to understand, but not the other.

***

This code contains many cryptic variables such as `p`, `d`, `g`, `w`, etc. this is completely against our coding style.

***

Finally, a few more comments before each 'step' of the script wouldn't hurt.

omichel:
Thank you for the detailed answer and hints. I believe I addressed all these points.

PR #1315, opened 2020-02-03
Simulation server script: connect to check port availability

stefaniapedrazzi:
Using the internal list of ports in use is to pick an available port is not sufficiently robust.
The only effective and robust way to check the status of a port is to try to connect to it.


stefaniapedrazzi:
Sorry I introduced a bug when improving the code.

Issue #1316, opened 2020-02-03
Webots wont import proto with connector


Acwok:
**Describe the Bug**
Webots wont import proto with connector as child.

**Steps to Reproduce**
1. Create a proto based on a solid with a connector as child
2. try to import it in Webots with the add new button
3. the proto doesn't appear in the list

Tested on webots 2019b revision1, webots 2020a, webots 2020a revision1

**Proto example:**

```
#VRML_SIM R2019b utf8
# license: Apache License 2.0
# license url: http://www.apache.org/licenses/LICENSE-2.0

PROTO Test [
    field  SFVec3f     translation  0 0 0
    field  SFRotation  rotation     0 1 0 0
    field  SFString    name         "test"
]
{
    Solid {
        translation IS translation
        rotation IS rotation
        children [
            Connector {       
            }
        ]
    }
}
```

DavidMansolino:
Thank you for this bug report, I can indeed reproduce this. The PROTO can only be inserted in a Robot node whereas it should be possible to insert it as a root node.

DavidMansolino:
This problem was fixed in #1317, the fix will be available in the next version of Webots (and the nightly builds).

PR #1317, opened 2020-02-04
Fixed Importation of Connectors in PROTO.

DavidMansolino:
**Description**
Webots wont import proto with connector as child.

**Related Issues**
This pull-request attempts to fix issue  #1316

### DIFF
--- a/src/webots/app/WbProtoCachedInfo.cpp
+++ b/src/webots/app/WbProtoCachedInfo.cpp
@@ -193,7 +193,7 @@ WbProtoCachedInfo *WbProtoCachedInfo::computeInfo(const QString &protoFileName) 
   tokenizer.rewind();
   while (tokenizer.hasMoreTokens()) {
     WbToken *token = tokenizer.nextToken();
-    if (token->isIdentifier() && WbNodeUtilities::isDeviceTypeName(token->word())) {
+    if (token->isIdentifier() && WbNodeUtilities::isDeviceTypeName(token->word()) && token->word() != "Connector") {
       protoInfo->mContainsDevices = true;
       break;
     }
--- a/src/webots/app/WbProtoCachedInfo.cpp
+++ b/src/webots/app/WbProtoCachedInfo.cpp
@@ -37,7 +37,7 @@
 
 static QString gFileHeader = "Webots Proto Cache File version 1.4";
 
-WbProtoCachedInfo::WbProtoCachedInfo(const QString &protoFileName) : mContainsDevices(false), mBaseType("UNKNOWN") {
+WbProtoCachedInfo::WbProtoCachedInfo(const QString &protoFileName) : mNeedsRobotAncestor(false), mBaseType("UNKNOWN") {
   mHexProtoFileHash.clear();
   QFileInfo info(protoFileName);
   mAbsoluteProtoFileName = info.absoluteFilePath();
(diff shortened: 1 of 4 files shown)


stefaniapedrazzi:
The Connector is a Device, so the expected behavior is that it should only be inserted in a node with a top Robot node.
Is it documented that the Connector it's special and doesn't behave like all the other devices and can be inserted in a Solid object?

DavidMansolino:
> The Connector is a Device, so the expected behavior is that it should only be inserted in a node with a top Robot node.

I don't think that it is somewhere stated that Devices can only be inserted in a node with a top Robot node.

> Is it documented that the Connector it's special and doesn't behave like all the other devices and can be inserted in a Solid object?

It is documented that connectors are linked to Solid (there is no notion of Robot node on the doc).

For me, this is clearly a bug as using base nodes you are allowed to insert a Connector in a root Solid.
We have an exception especially for this: 

https://github.com/cyberbotics/webots/blob/master/src/webots/nodes/utils/WbNodeUtilities.cpp#L1524

This seems to perfectly make sense since it is possible to have passive connectors, they don't need any control (and therefore no need of a Robot node).

stefaniapedrazzi:
Ok, thank you for the details.
I just think that for users having clear and simple rules is better.

So if we do not allow to insert devices in a non robot object (as it is implemented since ever in the code) we should document it.
Then we should also document any exception to this rule (in a separate PR).

stefaniapedrazzi:
:eyes:

src/webots/app/WbProtoCachedInfo.cpp
```diff
@@ -193,7 +193,7 @@ WbProtoCachedInfo *WbProtoCachedInfo::computeInfo(const QString &protoFileName)
   tokenizer.rewind();
   while (tokenizer.hasMoreTokens()) {
     WbToken *token = tokenizer.nextToken();
-    if (token->isIdentifier() && WbNodeUtilities::isDeviceTypeName(token->word())) {
+    if (token->isIdentifier() && WbNodeUtilities::isDeviceTypeName(token->word()) && token->word() != "Connector") {
       protoInfo->mContainsDevices = true;
```
stefaniapedrazzi: Just a minor note: maybe we should change the member and getter name to be more consistent with the code and instead of `mContainsDevice` change it to something like `mNeedsRobotAncestor`.

DavidMansolino: Good point. Are we allowed to change this string in the proto cache file as well or it is not good to do this in a minor version?

stefaniapedrazzi: I think it would be better to do it in a major version.
If I'm not wrong the only way to force an automatic update of the Proto Cache version if the PROTO file doesn't change is to increase the Proto Cache version.
if we do not increate the version, the cache of users' PROTO nodes will contain a wrong value.

stefaniapedrazzi: Another solution will be to support both old and new label in the revision and change it late. but it seems too much work for a small detail.

DavidMansolino: Ok, I will then change only the getter and variable name on the Webots side. and then fix this in develop once merged.

stefaniapedrazzi:
:eyes:

PR #1322, opened 2020-02-04
Feature office chair

DavidMansolino:
This PR adds an office chair:

hall


omichel:
I noticed the proto file is huge (1.59 MB). How many triangles are there? Shouldn't we reduce it a little bit?

DavidMansolino:
Strange indeed, because it is supposed to have ~2k triangles, but maybe I did forget to apply one of modifier, I will check.

DavidMansolino:
I reduced slightly the number of triangles (one of the modifiers was indeed not applied) and I removed some of the texture coordinate (size decreases by a factor > 3) .

PR #1328, opened 2020-02-05
Fix Display

DavidMansolino:
**Description**
The Display device is completely wrong in the management of image format, we are lucky that it works, but as soon as an image has a transparency layer the result is wrong.

The Display was still based on the old RGBA image format instead of BGRA.


src/lib/Controller/api/display.c
```diff
@@ -761,8 +763,8 @@ WbImageRef wb_display_image_load(WbDeviceTag tag, const char *filename) {
     return NULL;
   }
   GImage *gi = g_image_new(filename);
-  if (gi->failed || (gi->data_format != G_IMAGE_DATA_FORMAT_ABGR && gi->data_format != G_IMAGE_DATA_FORMAT_RGB)) {
-    fprintf(stderr, "Error: wb_display_image_load(): the \"%s\" image is unreadable.\n", filename);
```
omichel: Why did you remove this error message?

docs/reference/display.md
```diff
@@ -703,7 +704,7 @@ They should be deleted with the `wb_display_image_delete` function when they are
 Finally, note that both the main display image and the clipboard images have an alpha channel.
 
 The `wb_display_image_new` function creates a new clipboard image, with the specified `with` and `height`, and loads the image `data` into it with respect to the defined image `format`.
-Three images format are supported: `WB_IMAGE_RGBA` which is similar to the image format returned by a `Camera` device and `WB_IMAGE_RGB` or `WB_IMAGE_ARGB`.
+Five images format are supported: `WB_IMAGE_BGRA` which is similar to the image format returned by a `Camera` device and `WB_IMAGE_RGB`, `WB_IMAGE_RGBA`, `WB_IMAGE_ARGB`,  or `WB_IMAGE_ABGR`.
 `WB_IMAGE_RGBA` and `WB_IMAGE_ARGB` are including an alpha channel respectively after and before the color components.
```
omichel: But also `WB_IMAGE_BGRA` and `WB_IMAGE_ABGR` (maybe we can remove this sentence which is not very helpful).
```suggestion
```

src/lib/Controller/api/display.c
DavidMansolino: Oups that's a mistake, I will restore it.

DavidMansolino:
(I fogot to add the :heavy_check_mark: travis_unit : https://travis-ci.com/cyberbotics/webots/builds/147522941)

PR #1330, opened 2020-02-05
Responsive doc

fabienrohrer:
## Aim

Apply minor modifications to decrease a lot the width threshold to render correctly the documentation, and be supported on mobile platforms.

## Positive side effect

The doc looks better inside Webots.

## Result

https://cyberbotics.com/doc/reference/accelerometer?version=feature-responsive-doc

|  | Before | Now |
| --- | --- | --- |
| iPhone | doc-before | doc-after |
| Inside Webots | embedded-doc-before | embedded-doc-after |

## Notes

I know there are still things which could be improved, but I would prefer to go quickly on this topic. This should be seen as a quick hotfix.
I think the doc would deserve a strong CSS refactor. Dropping jquery and using Bulma in the doc would be awesome.


fabienrohrer:
The idea is to have a single width threshold under which we almost remove horizontal margin and padding of big components.

PR #1333, opened 2020-02-06
Fixed possible warning due to negative QTimer duration

omichel:
The problem was reported on [stackoverflow](https://stackoverflow.com/questions/60080991/warning-qobjectstarttimer-timers-cannot-have-negative-intervals-in-webots-ro). This patch guarantees that we never pass a negative value to the `QTimer`.

### DIFF
--- a/src/webots/engine/WbSimulationWorld.cpp
+++ b/src/webots/engine/WbSimulationWorld.cpp
@@ -54,7 +54,7 @@ WbSimulationWorld::WbSimulationWorld(WbProtoList *protos, WbTokenizer *tokenizer
   mSimulationHasRunAfterSave(false) {
   if (mWorldLoadingCanceled)
     return;
-  mSleepRealTime = (int)basicTimeStep();
+  mSleepRealTime = basicTimeStep();
 
   WbSimulationState::instance()->resetTime();
   // Reset random seed to ensure reproducible simulations.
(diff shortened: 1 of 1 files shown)


src/webots/engine/WbSimulationWorld.cpp
```diff
@@ -175,9 +175,11 @@ void WbSimulationWorld::step() {
     //              (if the real-time mode is enabled, of course)
     // mean *= 0.90;
 
-    if (mean > timeStep && mSleepRealTime > 0.0)
+    if (mean > timeStep && mSleepRealTime > 0.0) {
       mSleepRealTime -= 0.03 * timeStep;
-    else if (mean < timeStep)
+      if (mSleepRealTime < 0)
+        mSleepRealTime = 0.0;
```
DavidMansolino: From a control point of view introducing the clamp here is not very good as it introduced a bias in the positive direction...
Personally I would have instead clamped this just before use.
```
mTimer->start(qMax(mSleepRealTime, 0));
```

omichel: Yes, that's true. However, this case is extremely rare and we should also rename the `mSleepRealTime` member variable into a more complicated name to reflect that it's a proportional control parameter (and hence can be negative). For simplicity, I'd prefer to keep things like that.

Issue #1350, opened 2020-02-11
Controlling robotiq 2 finger gripper: joint didn't move other parts of gripper 


ipa-rwu:
Hello everyone,

the main goal is to control robotiq 2 finger gripper from moveit.

I am trying to use robotiq 2 finger 85 gripper in webots. I used ```urdf2webots``` transfer gripper urdf which is in this  [repository](https://github.com/StanleyInnovation/robotiq_85_gripper/tree/master/robotiq_85_description/urdf) to ".proto" file which you can find in this [link](https://github.com/ipa-rwu/ur5_2f_webots/blob/master/protos/Robotiq85Gripper.proto). 

Now I can connect it with moveit. I sent a goal from rviz as you can see this figure. It moves perfect in rviz.
![Screenshot from 2020-02-11 15-06-36](https://user-images.githubusercontent.com/54799760/74249026-d2253980-4ce8-11ea-8138-ac703103c2c9.png)
![Screenshot from 2020-02-11 15-06-48](https://user-images.githubusercontent.com/54799760/74249035-d6515700-4ce8-11ea-8d88-6ed86841f5be.png)

But there is only one joint moves in webots. It supposes "gripper_finger1_joint" move other parts. But it didn't. As you can see this picture. 
![Screenshot from 2020-02-11 15-06-54](https://user-images.githubusercontent.com/54799760/74249465-7c9d5c80-4ce9-11ea-8303-78614aadf051.png)
Is it model probleam? I have no ideas to config proto files.

Anyone has ideas?

DavidMansolino:
There is indeed probably an issue with your model.
Would you be interested in adding your model to Webots?
If so, please open a PR to add your PROTO to this folder https://github.com/cyberbotics/webots/tree/master/projects/devices/robotiq/protos
Once the PR is open I will help you fixing your model.

ipa-rwu:
Thank you very much. I already did.

DavidMansolino:
Perfect, I will have a look and let you know directly on the PR.

DavidMansolino:
Linked to #1365

ipa-rwu:
Thank you for releasing a new version. I tried the model for robotiq 2 finger gripper which was created for 2020a but is not working for 2020b yet. You can find in this [link](https://github.com/cyberbotics/webots/tree/ipa-rwu-patch-1/projects/devices/robotiq/protos)

DavidMansolino:
What is not working exactly?
[...]

PR #1351, opened 2020-02-11
proto model for robotiq 2finger 85 gripper 

ipa-rwu:
**Description**
robotiq 2 finger 85 gripper

**Related Issues**
This pull-request fixes issue  #1350

**Tasks**
  - [ ] Task 1
  - [ ] Task 2

**Documentation**
https://cyberbotics.com/doc/guide/getting-started-with-webots?version=ipa-rwu-patch-1

**Screenshots**
TODO


DavidMansolino:
Very nice model!
It seems that the problem you are facing is caused by some missing joint (at the location indicated by the red arrows):
Capture du 2020-02-11 17-21-12
Probably that the urdf2webots importer is not handling correctly closed loop. I will check this tomorrow morning.

DavidMansolino:
@ipa-rwu, I have fixed the missing joint issue. Let me know if this solves your issue.
If it works, I would suggest improving your model for better performance and stability.
The next step would be to improve the bounding objects, instead of using the meshes directly it would be better to use sets of basic geometries (such as cylinder, boxes, etc.).

ipa-rwu:
@DavidMansolino Thank you very much. I tried your model. The good things is that "gripper_finger1_joint" can move a half part of gripper. As you can see below:

Screenshot from 2020-02-12 09-39-00

But According to this gripper, "gripper_finger1_joint" should control both part. As you can see in rviz.
Screenshot from 2020-02-12 09-40-41

I found a example where "finger_joiont" should be "gripper_finger1_joint" in our case.
display

DavidMansolino:
Unfortunately, in Webots a motor can control only on joint, can't you simply apply the same command on the `gripper_finger2_joint` motor?
Note also, that from the original URDF file it seems there are two distinct joints.

ipa-rwu:
@DavidMansolino Good to know. I will try to rewrite the driver.

ipa-rwu:
@DavidMansolino I check the urdf of this gripper, 
`
        <joint name="${prefix}gripper_finger1_joint" type="revolute">
            <parent link="${prefix}gripper_base_link"/>
            <child link="${prefix}gripper_finger1_knuckle_link"/>
            <axis xyz="0 0 1"/>
            <origin rpy="${pi} 0.0 0.0" xyz="0.05490451627 0.03060114443 0.0"/>
            <limit lower="0.0" upper="0.804" velocity="2.0" effort="1000"/>
        </joint>

        <joint name="${prefix}gripper_finger2_joint" type="revolute">
            <parent link="${prefix}gripper_base_link"/>
            <child link="${prefix}gripper_finger2_knuckle_link"/>
            <axis xyz="0 0 1"/>
            <origin rpy="0.0 0.0 0.0" xyz="0.05490451627 -0.03060114443 0.0"/>
            <limit lower="-3.14" upper="3.14" velocity="100.0" effort="1000"/>
            <mimic joint="${prefix}gripper_finger1_joint"/>
`
Is it possible do it in webots?

DavidMansolino:
Well spotted, I didn't saw the 'mimic joint' part.
Unfortunately, this is not yet possible in Webots.
However, as a workaround, you can implement this in the interface controller between Webots and ROS.

ipa-rwu:
@DavidMansolino I can move these two joints now. But I found it is different between the one in rviz and in webots. Do you have any ideas? 
Screenshot from 2020-02-13 09-11-10
Screenshot from 2020-02-13 09-11-38

DavidMansolino:
This is probably due to the fact that not only `gripper_finger1_joint` and `gripper_finger2_joint` are linked, but they are also linked to `gripper_finger1_inner_knuckle_joint`, `gripper_finger2_inner_knuckle_joint`, `gripper_finger1_finger_tip_joint` and `gripper_finger2_finger_tip_joint`.

By the way, we added the feature of linking joints together to our roadmap (#1359).

DavidMansolino:
Moved to https://github.com/cyberbotics/community-projects/pull/16

PR #1352, opened 2020-02-12
MJPEG streaming

stefaniapedrazzi:
Implement the MJPEG based streaming alongside with the X3D streaming.

Tasks
  - [x] implement MJPEG streaming in Webots and refactor the `WbStreamingServer` delete `WbMultimediaStreamer` class: move code specific for x3d and mjpeg streaming in two different classes inheriting from the `WbStreamingServer` class
  - [x] add functions to block showing the handles, matter centers, etc. in the Webots 3D scene
  - [x] update video streaming code in `webots.min.js`
  - [x] add robot windows in web interface for multimedia streaming
  - [x] add context menu in the web interface for multimedia streaming
  - [x] add edit controller functionality and robot windows in web interface for multimedia streaming
  - [x] fix all scene refresh issues in the web interface for multimedia streaming
  - [x] fix scene interaction from touch devices
  - [x] update documentation
  - [x] optimizations

Note:
* viewpoint_control benchmark no longer working -> viewpoint data have to be sent from Webots and simulation has to run

Tests:
  - [x] Firefox
  - [x] Chrome
  - [x] Edge
  - [x] Safari
  - [x] Android
  - [x] iOS (doesn't work on robotbenchmark.net)
    - bug moving jQuery UI window (see #1496)


stefaniapedrazzi:
Only optimizations are missing but it would be good to have a first review.

The changes in this branch are deployed on beta.robotbenchmark.net.
(for Webots installation the version based on master, i.e. before 5e9beb6, is installed on the simulation server)

omichel:
https://beta.robotbenchmark.net is working very well (from my home).

DavidMansolino:
I tested a few simulations of https://beta.robotbenchmark.net, with my slow home network, it is not very nice to use, but when switching to 4G it is perfect 😃

omichel:
@DavidMansolino: is it better to use the X3D version from your slow network connection?

DavidMansolino:
The X3D version runs much more smoothly on my slow network connection but it is way more slower to load.

resources/web/wwi/multimedia_client.js
```diff
@@ -0,0 +1,383 @@
...
+    this.lastWidth = this.domElement.width;
+    this.lastHeight = this.domElement.height;
+  }
+
+  resize(width, height) {
+    this.requestNewSize();
+  }
+
+  processServerMessage(data) {
+    if (data.startsWith('multimedia: ')) {
```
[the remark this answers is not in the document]
stefaniapedrazzi: This is an old message that was already implemented, and all the messages currently uses this protocol.
But in general using the colon is not causing any issues in the existing messages.
I introduced a couple of more complex messages where I had to sent a JSON objects, because when you start to send more than one string I agree that it is complicated with this plain format.
In any case the messages needs to be synchronized in Webots and in webots.min.js, so it will be in any case error prone even with JSON objects.

We can decide to move all the messages to JSON objects, but this should apply to all of them (including the ones for x3d streaming) and will be break completely the backwards compatibility.
Because if it is tricky and error prone for all the mjpeg streaming messages, than it is the same for all the other messages.

src/webots/gui/WbX3dStreamingServer.cpp
```diff
@@ -0,0 +1,250 @@
...
+void WbX3dStreamingServer::startX3dStreaming(QWebSocket *client) {
+  try {
+    if (WbWorld::instance()->isModified() || mX3dWorldGenerationTime != WbSimulationState::instance()->time())
+      generateX3dWorld();
+    sendWorldToClient(client);
+    // send the current simulation state to the newly connected client
+    const QString &stateMessage = simulationStateString();
+    if (!stateMessage.isEmpty())
+      client->sendTextMessage(stateMessage);
+    // TODO
```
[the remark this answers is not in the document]
stefaniapedrazzi: I checked and this function has already been fully ported to the new class. I just forgot to remove the TODO.

resources/web/wwi/wwi.css
```diff
@@ -331,3 +331,17 @@ button::-moz-focus-inner {
...
+#remoteScene {
+  background: white;
+  width: 100%;
+  height: calc(100% - 48px);
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  -o-user-select: none;
+  user-select: none;
+  // get rid of whitespace after #remoteScene image
```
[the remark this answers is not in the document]
stefaniapedrazzi: it works  correctly on my system, but maybe it won't on some other system.
I will change it.

resources/web/wwi/x3d_scene.js
```diff
@@ -439,6 +439,72 @@ class X3dScene { // eslint-disable-line no-unused-vars
     return undefined;
   }
 
+  getRobotWindows() {
+    var windows = [['worldInfoWindow', this.worldInfo.window]];
```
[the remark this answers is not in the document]
stefaniapedrazzi: Yes.
Previously we didn't use the `let` keyword. I didn't check the whole code to change `var` into `let`.
This is an old piece of code that I just moved to a new function.

stefaniapedrazzi:
Thank you for the tests!
Yes, for the mjpeg streaming we need a good network connection.
Probably it can still be slightly improved but for slow connections the x3d option will remain better.

> I was wondering if we should introduce some generic mechanism to define what user interaction is allowed / not allowed in a specific simulation, e.g., is the user allowed to apply a force/torque, move objects while the simulation is running, etc.

Do you mean in Webots or from the Web interface?
Yes, at some point we will need it at least in Webots to support different cases.

> Also, it seems some renderings and image generations could be skipped, cached or otherwise optimized, but I assume you still need to iterate on the optimization.

Do you have any example in particular?
I didn't touch or inspect the way Webots requests new renderings. This should already be optimized for Webots as well. If not it should be fixed in a separate PR.
But I already added some code so that the images are created and sent only when needed.
I didn't inspect the mjpeg streaming protocol options.

omichel:
> Do you have any example in particular?
No, that was just some suspicion, I didn't fully check it, so probably I am wrong.

stefaniapedrazzi:
Remaining TODOs following the weekly meeting discussion:
* [ ] detect slow connections and reduce image resolution / JPEG compression
* [ ] fix complex messages and use JSON objects instead

stefaniapedrazzi:
Image comparison detects equal images only if just the background (tested with textured background) is visible. If some objects are visible (but nothing moves) they are not detected as the same images using Qt hash functionalities (or JPEG comparison).
If we would like to really try to go on with this improvement we should inspect other hashing methods, like perceptual hashing, that finds similar images (instead of equal images).

DavidMansolino:
Can you trying with the ambient occlusion disabled just to make sure that we understand where this difference comes from?

stefaniapedrazzi:
Yes, it's the ambient occlusion. If I disable it then the images comparison works as you would expect.

DavidMansolino:
Thank you for the confirmation.

omichel:
I believe we should try to implement perceptual hashing later if really needed.

omichel:
Alternatively, we could reset the ambient occlusion random seed at each rendering so that images are actually identical.

DavidMansolino:
> Alternatively, we could reset the ambient occlusion random seed at each rendering so that images are actually identical.

I am not sure this will work, as far as I understand the ambient occlusion relies on the fact that the 'noise' is not the same at each iteration and use the outputs of the previous frames to improve the final rendering (this is an optimization that allows to compute only a 'sub-part' of the occlusion at each frame).

stefaniapedrazzi:
Here is an article reviewing the different methods for interactive live streaming. It is also considering the browser native MJPEG implementation that we are using and listing the advantages and limitations:
https://link.springer.com/article/10.1007/s11042-017-4556-6#ref-CR25

omichel:
Thank you for the link. So, I believe we should investigate the following:

- Try to use [QTcpSocket::bytesWritten](https://doc.qt.io/qt-5/qiodevice.html#bytesWritten) to see whether it can help to determine if the connection is jamming: if the number of bytes written doesn't match the number of bytes sent, we should consider it's jamming and hence we should reduce the image quality. Otherwise we should increase the image quality.

If that doesn't work, we have two options:

1. Implement a WebSocket Javascript interface that will receive binary JPEG images from Webots, display them and acknowledge the receipt. This will allow Webots to know for sure if the image is received and adapt the image quality and frame rate accordingly.

2. Add a menu in the web interface allowing the user to set the image quality.

On the long term, I would like to avoid (2.) as it is not user friendly.

stefaniapedrazzi:
I already investigated using QTcpSocket::bytesWritten. The number is reset immediately when the message is sent and not when the client receives or handles the images. So I could not get any useful data out of it.

omichel:
But isn't that sufficient?
When you have a very slow connection, is `byteWriten` called immediately with the number of bytes sent using the `write` method?

omichel:
I am thinking about a another possibility: on the client side, at the initialization stage, we could measure how much time does it take to download a large sample image from Webots (like a splash screen) and send back this information to Webots, expressed in MB/s. Then depending on this MB/s value, Webots could choose an image quality for the M-JPEG streaming.

stefaniapedrazzi:
> But isn't that sufficient?
When you have a very slow connection, is byteWriten called immediately with the number of bytes sent using the write method?

I will check again with tests on the streaming server, but from my preliminary tests the byteWritten value was not giving any information about the client network speed. Maybe it will give info about the server upload speed.

> I am thinking about a another possibility: on the client side, at the initialization stage, we could measure how much time does it take to download a large sample image from Webots (like a splash screen) and send back this information to Webots, expressed in MB/s. Then depending on this MB/s value, Webots could choose an image quality for the M-JPEG streaming.

Yes. That would be something to try.
I was also trying to get the network speed sending an image from Webots to the client through the web socket (once or continuously like every 5 seconds or so).

stefaniapedrazzi:
Finally I could get some useful information about the connection speed using the `QTcpSocket::bytesWritten` and `QTcpSocket::bytesToWrite` functions.
It seems that the method I used to limit my network connection was not fully working. Now I can clearly see that with slow connection the QTcpSocket write buffer size continuously increases.

stefaniapedrazzi:
I setup beta.robotbenchmark to use the MJPEG streaming.
But if started from the streaming_server.py, `QTcpSocket::bytesToWrite` value always returns 0 and thus it is not possible to detect slow connections.
If I start manually Webots from a terminal on the simulation server machine and connect using the `streaming_viewer` then it works. The only difference in the starting command is that in this second case SSL is not used.

omichel:
So that would mean that the connection speed evaluation doesn't work on SSL connections?

stefaniapedrazzi:
I'm inspecting it, but it's probably just a matter of using a different method for QSslSocket.

omichel:
Great! :eyes:

omichel:
It works really nicely and smoothly! 😄
I am testing on Windows 10 with Firefox and Chrome.

However, when I resize the window, the 3D view and toolbar don't get resized.
On Chrome, resizing the window also breaks the 3D navigation: it doesn't work any more.
On Firefox, it seems more robust.
Also the fullscreen works only in Firefox, not in Chrome.
Can you reproduce these problems?

omichel:
EDIT: Firefox and Chrome behave the same regarding the fullscreen mode: if you resize the window, then it is not possible to enter the fullscreen mode. But if you don't resize the window, it works: you enter and quit the fullscreen mode nicely.

omichel:
EDIT: finally Firefox and Chrome behave exactly the same, also regarding the 3D navigation which is broken on both browser after you resize the window.

stefaniapedrazzi:
Thank you for the review.
I will fix the resizing issues.

stefaniapedrazzi:
I fixed the resize but I still have to check if the updated images is sent correctly.
Then I would also fix the worlds combo box that it is not disabled when the toolbar is disabled.

omichel:
The resize now seems to work fine.
However, I noticed some annoying bug: when navigating in the 3D view, if I drag the mouse and release the mouse button above the info window or above the toolbar, my button release event is not taken into account and although I do not press the button any more, the viewpoint is still rotating while I move the mouse.

stefaniapedrazzi:
I could reproduce the drag issue only in the streaming viewer and not on robotbenchmark on localhost. It should be fixed now.

I also fixed the disabling/enabling of the world select on the toolbar.

omichel:
I cannot test https://beta.robotbenchmark.net, it seems to be down (no answer from session server).

omichel:
Thank you. I confirm the resize issue is fixed.
Also the drag behavior is now better: no mouse up event is lost any more and the behavior is now similar to the X3D version.

But it is strange that when the drag continue over a robot window, over the toolbar or outside of the navigator window, it is not taken into account any more and the viewpoint doesn't move. It is certainly possible to improve this behavior to achieve something like [this](https://eyes.nasa.gov/curiosity/), but it should probably be done in a separate PR.

src/webots/gui/WbMultimediaStreamingServer.cpp
```diff
@@ -0,0 +1,393 @@
...
+    }
+    mLimiter->resetStop();
+    return;
+  }
+
+  double bytes;
+  if (mSentImagesCount > 0)
+    bytes = ((double)mAverageBytesToWrite) / mSentImagesCount;
+  else
+    bytes = 0;
```
omichel, line 168:
```suggestion
  const double bytes = (mSentImagesCount > 0) ? ((double)mAverageBytesToWrite) / mSentImagesCount : 0;
```

src/webots/gui/WbX3dStreamingServer.cpp
```diff
@@ -0,0 +1,256 @@
...
+      startX3dStreaming(client);
+    // else streaming is started once the world loading is completed
+    return;
+  } else if (message == "reset") {
+    // reset nodes visibility
+    QWebSocket *client = qobject_cast<QWebSocket *>(sender());
+    foreach (WbBaseNode *node, WbWorld::instance()->viewpoint()->getInvisibleNodes())
+      client->sendTextMessage(QString("visibility:%1:1").arg(node->uniqueId()));
+    resetSimulation();
+    QString state = WbAnimationRecorder::instance()->computeUpdateData(true);
```
omichel, line 93:
```suggestion
    const QString &state = WbAnimationRecorder::instance()->computeUpdateData(true);
```

src/webots/gui/WbX3dStreamingServer.cpp
```diff
@@ -0,0 +1,256 @@
...
+  }
+}
+
+void WbX3dStreamingServer::sendUpdatePackageToClients() {
+  sendActivityPulse();
+
+  if (mWebSocketClients.size() > 0) {
+    qint64 currentTime = QDateTime::currentMSecsSinceEpoch();
+    if (mLastUpdateTime < 0.0 || currentTime - mLastUpdateTime >= 1000.0 / WbWorld::instance()->worldInfo()->fps()) {
+      QString state = WbAnimationRecorder::instance()->computeUpdateData(false);
```
omichel, line 126:
```suggestion
      const QString &state = WbAnimationRecorder::instance()->computeUpdateData(false);
```

src/webots/gui/WbX3dStreamingServer.cpp
```diff
@@ -0,0 +1,256 @@
...
+    }
+  }
+}
+
+bool WbX3dStreamingServer::prepareWorld() {
+  try {
+    bool regenerationRequired = mX3dWorldReferenceFile != WbWorld::instance()->fileName() || mX3dWorldGenerationTime != 0.0;
+    if (!regenerationRequired) {
+      // if a non static procedural PROTO is used we need to regenerate the world anyway
+      QList<WbProtoModel *> models = WbProtoList::current()->models();
```
omichel, line 143:
```suggestion
      const QList<WbProtoModel *> models = WbProtoList::current()->models();
```

src/webots/gui/WbX3dStreamingServer.cpp
```diff
@@ -0,0 +1,256 @@
...
+        }
+      }
+      mLastUpdateTime = currentTime;
+    }
+  }
+}
+
+bool WbX3dStreamingServer::prepareWorld() {
+  try {
+    bool regenerationRequired = mX3dWorldReferenceFile != WbWorld::instance()->fileName() || mX3dWorldGenerationTime != 0.0;
```
omichel, line 140:
```suggestion
    const bool regenerationRequired = mX3dWorldReferenceFile != WbWorld::instance()->fileName() || mX3dWorldGenerationTime != 0.0;
```

src/webots/gui/WbX3dStreamingServer.cpp
```diff
@@ -0,0 +1,256 @@
...
+  if (node->isProtoParameterNode()) {
+    // PROTO parameter nodes are not exported to X3D or transmitted to webots.min.js
+    foreach (WbNode *nodeInstance, node->protoParameterNodeInstances())
+      propagateNodeAddition(nodeInstance);
+    return;
+  }
+
+  WbStreamingServer::propagateNodeAddition(node);
+
+  WbBaseNode *baseNode = static_cast<WbBaseNode *>(node);
```
omichel, line 193: Could be declared as const.

src/webots/gui/WbX3dStreamingServer.cpp
```diff
@@ -0,0 +1,256 @@
...
+  world->write(writer);
+  mX3dWorld = worldString;
+  mX3dWorldTextures = writer.texturesList();
+  mX3dWorldGenerationTime = WbSimulationState::instance()->time();
+  mX3dWorldReferenceFile = WbWorld::instance()->fileName();
+  mLastUpdateTime = -1.0;
+}
+
+void WbX3dStreamingServer::sendWorldToClient(QWebSocket *client) {
+  qint64 ret = client->sendTextMessage(QString("model:") + mX3dWorld);
```
omichel, line 231:
```suggestion
  const qint64 ret = client->sendTextMessage(QString("model:") + mX3dWorld);
```

src/webots/gui/WbX3dStreamingServer.cpp
```diff
@@ -0,0 +1,256 @@
...
+  if (ret < mX3dWorld.size())
+    throw tr("Cannot sent the entire world");
+
+  QString state = WbAnimationRecorder::instance()->computeUpdateData(true);
+  if (!state.isEmpty())
+    sendWorldStateToClient(client, state);
+
+  QList<WbRobot *> robots = WbWorld::instance()->robots();
```
omichel, line 239:
```suggestion
  const QList<WbRobot *> robots = WbWorld::instance()->robots();
```

stefaniapedrazzi:
> But it is strange that when the drag continue over a robot window, over the toolbar or outside of the navigator window, it is not taken into account any more and the viewpoint doesn't move. It is certainly possible to improve this behavior to achieve something like this, but it should probably be done in a separate PR.

Note that this is the exact same behavior as in robotbenchmark.net.
So I agree that it should be handled in a separate PR. I will open it.

This is because with the current method we are not getting any event when the mouse is over the jQuery window or toolbar. Adding events listeners, it should be possible to get the mouse move events when over a jQuery element. No idea if JS supports processing mouse events outside the navigator window.

resources/web/streaming_viewer/index.html
```diff
@@ -6,24 +6,24 @@
...
+    <script src="https://www.cyberbotics.com/wwi/R2020a/webots.mjpeg.js"></script>
     <script src="setup_viewer.js"></script>
   </head>
   <body>
     <div style="padding-left:10px">
       <h1>Webots streaming viewer</h1>
       <p>
         Connect to:
-        <input id="IPInput" type="text" value="localhost" />
+        <input id="IPInput" type="text" value="cyberbotics1.epfl.ch" />
```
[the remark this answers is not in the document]
stefaniapedrazzi: This are debug values in case you wanted to test the streaming directly from the streaming server.
But these changes have to be removed before merging.

resources/web/streaming_viewer/index.html
```diff
@@ -6,24 +6,24 @@
     <title>Streaming viewer</title>
     <link rel='stylesheet' href='https://code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css'/>
     <link type="text/css" rel="stylesheet" href='https://www.cyberbotics.com/jquery-ui-layout/1.4.4/layout-default.css' />
-    <link type="text/css" rel="stylesheet" href='https://www.cyberbotics.com/wwi/R2020b/wwi.css' />
+    <link type="text/css" rel="stylesheet" href='https://www.cyberbotics.com/wwi/R2020a/wwi.mjpeg.css' />
```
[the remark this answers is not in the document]
stefaniapedrazzi: These are debug values. 
Before merging this branch it was bad to modify existing files (like wwi.css).
Once this PR is ready to be merged I will revert it.

omichel:
> No idea if JS supports processing mouse events outside the navigator window.

At least it works [here](https://eyes.nasa.gov/curiosity/).

src/webots/gui/WbX3dStreamingServer.cpp
stefaniapedrazzi: `regenerationRequired` is changed at line 146, so it cannot be changed to const.

stefaniapedrazzi:
I fixed all the comments.

src/webots/gui/WbMultimediaStreamingServer.cpp
```diff
@@ -111,7 +111,7 @@ bool WbMultimediaStreamingServer::isNewFrameNeeded() const {
     return true;
 
   const qint64 msecs = mUpdateTimer.elapsed();
-  return msecs >= mFrameRate;  // maximum update frame rate
+  return msecs >= mImageUpdateTimeStep;  // maximum update frame rate
```
omichel, line 114:
```suggestion
  return msecs >= mImageUpdateTimeStep;  // maximum update time step
```

src/webots/gui/WbStreamingServer.cpp
```diff
@@ -206,7 +206,7 @@ void WbStreamingServer::onNewTcpData() {
   const QString &line(socket->peek(1024));  // Peek the request header to determine the requested url.
   QStringList tokens = QString(line).split(QRegExp("[ \r\n][ \r\n]*"));
   if (tokens[0] == "GET") {
-    const QString &requestedUrl(tokens[1].replace(QRegExp("^/"), ""));
+    QString requestedUrl(tokens[1].replace(QRegExp("^/"), ""));
```
omichel, line 209: Why did you removed the const qualifier here?

docs/guide/web-streaming.md
```diff
@@ -6,7 +6,20 @@ Webots can be used as a Web streaming server, i.e., to stream a simulation to se
 In this mode the user can watch an already running simulation and navigate into the scene but cannot interact with the simulation, i.e. controlling the execution of the simulation and modifying the robot controller program.
 Please refer to the [Web simulation](web-simulation.md) documentation to setup a platform where users can run individual simulations.
 
-The mechanism is similar to the [web animation export](web-animation.md) except that the `X3D` file and the animation are sent on the fly to the Web browser clients.
+Two different streaming modes are available:
+- `x3d` mode (default) where the simulation is streamed using a mechanism similar to the [web animation export](web-animation.md) except that the `X3D` file and the animation are sent on the fly to the web browser clients.
```
omichel, line 10:
```suggestion
- `x3d` (default): the simulation is streamed using a mechanism similar to the [web animation export](web-animation.md) except that the `X3D` file and the animation are sent on the fly to the web browser clients.
```

docs/guide/web-streaming.md
```diff
@@ -6,7 +6,20 @@ Webots can be used as a Web streaming server, i.e., to stream a simulation to se
...
+- `mjpeg` mode where the Webots scene rendered on the server is directly sent as `MJPEG` image to the web browser client.
```
omichel, line 11:
```suggestion
- `mjpeg`: the Webots scene rendered on the server is directly sent as an `MJPEG` image to the web browser client.
```

src/webots/gui/WbStreamingServer.cpp
stefaniapedrazzi: That's a mistake. I just restored it.

stefaniapedrazzi:
:+1:

PR #1363, opened 2020-02-14
Fix universal robot ros python version

DavidMansolino:
**Description**
This controller should use python 2.7 as ROS1 supports only python 2.7 for now.


tests/sources/test_license.py
```diff
@@ -170,10 +169,11 @@ def test_sources_have_license(self):
                             (source, APACHE2_LICENSE_CPP)
                     )
                 elif source.endswith('.py') or source.endswith('Makefile'):
+                    for pythonHeader in PYTHON_OPTIONAL_HEADERS:
+                        if content.startswith(pythonHeader):
```
omichel:
```suggestion
                        if content.startswith(pythonHeader + '\n'):
```

PR #1367, opened 2020-02-14
Fixed a mistake in the table for IDs

LarsVaehrens:
**Description**
Changed palm_finger_1_joint_sensor to finger_1_joint_1_sensor in the third row of the second column, because palm_finger_1_joint_sensor was mentioned in two places.


omichel:
FYI, here is the test page for this PR: https://cyberbotics.com/doc/guide/gripper-actuators?version=LarsVaehrens:patch-1

Issue #681, opened 2019-06-30
Unable to run matlab example


sxbug:
**Describe the Bug**
env：Win10，webots 2019b and MATLAB 2017b. 
Using webots to open its own matlab example, click to run it , the MATLAB window is blocked, there is frozen, can only be  closed by windows' TaskManager .  I have set WEBOTS_HOME, WEBOTS_CONTROLLER_NAME and WEBOTS_VERSION parameters .            
Later, I came up with a solution: empty the content of launcher. m,  then run this example.  After pop up the MATLAB window  stop by error,  I copy the content of launcher. m into the matlab window and run it again. This Webots example can run now, but there are still errors, as if there is some libraries not loaded.

**Steps to Reproduce**
1. open webots2019b
2. load C:\Program Files\Webots\projects\languages\matlab\worlds\e-puck_matlab.wbt
3. run it
4. matlab frozen

**Expected behavior**
Should be the normal operation of the program

**Screenshots**

**System**
 - win10 64bit Chinese version
 - Graphics Card: AMD Radeon R9 M370X

DavidMansolino:
This is most probably related to a conflicting library in your `PATH`, please have a look at #63.

sxbug:
I had clear windows registry ,and reinstalled matlab, but frozen too.

when I clear launcher.m, and try again. after pause, I  paste origin launcher.m to matlab window 

it can run,and some error msg:
[matlab] Error: �Ҳ���ָ���ĳ�����
[matlab]  (dynamic library)
[matlab] Error: remote control initialisation failed
[matlab] Error: Cannot load the "C:/Program Files/Webots/projects/robots/gctronic/e-puck/plugins/remote_controls/e-puck_bluetooth/e-puck_bluetooth.dll" remote control library.
[matlab] Running Matlab sample Webots controller.

DavidMansolino:
This is indeed a known issue caused by conflicting libraries between Matlab and the e-puck remote-control, however, you can safely ignore these errors they will not prevent you from controlling the robot from Matlab.

sxbug:
But I can not find a way to ignore these errors, because when I start this simulate ,the matlab is halt and ONLY a black window. What can I do? Thanks

DavidMansolino:
Have you tried the other Matlab samples which don't have dedicated temote_control plugins, e.g. youbot_matlab?

sxbug:
I had try youbot_matlab, It's the same. Matlab is still stuck.

[...]

PR #1378, opened 2020-02-18
Fixed Wrong Matlab Paths

DavidMansolino:
**Description**
Some path have not been updated in the matlab library when moving the controller libraries.

**Related Issues**
Related to #681


DavidMansolino:
:heavy_check_mark:: tested with the latest version of Matlab (R2019b Update 4) on both Ubuntu 16.04 and Windows 10.

Issue #1384, opened 2020-02-18
Emitter/Receiver communication between nodes not working after controller and simulation reset


tsampazk:
**Description**
Main Issue:
I'm using a supervisor node to reset a robot controller and the simulation. The supervisor node and the robot node communicate with receivers/emitters. They work ok up until i reset the robot controller. In the dumbed down example below the robot is sending a counter to the supervisor,
who resets the simulation and the robot controller when it receives the value 10.

After resetting the robot controller and it is restarted and properly broadcasting, it seems that the supervisor receiver stops receiving the messages (the queue length stays at 0). After several steps (12 for me in the example below) the supervisor receiver starts to pick up the old messages (after reset) in order, thus they get badly desynchronized. The send method in the robot emitter properly returns 1.

What i found:
The issue seems to happen when the simulation is also reset via the supervisor.simulationReset()
method is called (it is irrelevant if its called before or after resetting the robot controller. If removed, the expected behavior happens.

Side issue:
Moreover, after the robot controller reset, the supervisor receives an extra value that it probably shouldn't. According to the docs, the sent messages are received instantly and the emitter/receiver queues are cleared by the simulationReset() method.

Based on the main + side issues my guess is that the receiver/emitter queues are messed up after the simulationReset() method.

Note:
In the very first step, the supervisor doesn't receive a value, probably depending on the order the step functions of the two controllers are called.

**Steps to Reproduce**
World .wbt file:
```
#VRML_SIM R2020a utf8
WorldInfo {
}
Viewpoint {
  orientation -0.6931032800836722 0.6931032800836722 0.1980295085953349 0.75
  position 1.2 1.6 2.3
}
TexturedBackground {
}
TexturedBackgroundLight {
}
RectangleArena {
}
DEF ROBOT Robot {
  translation 0.16 0.08 0
  children [
    Receiver {
    }
    Emitter {
    }
  ]
  controller "robot_controller"
}
DEF SUPERVISOR Robot {
  translation -0.22 0.08 0.1
  children [
    Receiver {
    }
    Emitter {
    }
  ]
  name "supervisor"
  controller "supervisor_controller"
  supervisor TRUE
}
```
robot_controller.py
```
from controller import Robot


class RobotEmitter:
    def __init__(self):
        self.robot = Robot()

        self.timestep = int(self.robot.getBasicTimeStep())

        self.emitter = self.robot.getEmitter("emitter")

        self.data = 0

    def get_timestep(self):
        return self.timestep

[...]

PR #1385, opened 2020-02-19
Fix Receiver Reset

DavidMansolino:
**Description**
Since the time is reset to 0 but the controller is not restarted, we should reset the sensor to make sure the `needToRefresh` returns true after the next step even if the sensor is not re-enabled (which is not the case since the controller is not restarted).

**Related Issues**
This pull-request fixes issue #1384

### DIFF
--- a/src/webots/nodes/WbRobot.hpp
+++ b/src/webots/nodes/WbRobot.hpp
@@ -141,6 +141,7 @@ signals:
   void controllerChanged();
   void controllerRestarted();
   void controllerExited();
+  void robotResetted();
   void toggleRemoteMode(bool enable);
   void sendToJavascript(const QByteArray &);
   void appendMessageToConsole(const QString &message, bool useStdout);
--- a/src/webots/nodes/WbRobot.hpp
+++ b/src/webots/nodes/WbRobot.hpp
@@ -141,7 +141,7 @@ signals:
   void controllerChanged();
   void controllerRestarted();
   void controllerExited();
-  void robotResetted();
+  void wasReset();
   void toggleRemoteMode(bool enable);
   void sendToJavascript(const QByteArray &);
   void appendMessageToConsole(const QString &message, bool useStdout);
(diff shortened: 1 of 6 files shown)


src/webots/nodes/WbRobot.hpp
```diff
@@ -141,6 +141,7 @@ public slots:
   void controllerChanged();
   void controllerRestarted();
   void controllerExited();
+  void robotResetted();
```
omichel: This is not English, we should probably use `robotReset` or `robotWasReset` instead.

src/webots/nodes/WbSensor.cpp
```diff
@@ -65,6 +66,14 @@ void WbSensor::toggleRemoteMode(bool enabled) {
...
-void WbSensor::connectToRobotSignal(const WbRobot *robot) {
-  connect(robot, &WbRobot::toggleRemoteMode, this, &WbSensor::toggleRemoteMode, Qt::UniqueConnection);
+void WbSensor::connectToRobotSignal(const WbRobot *robot, bool connectRemoteMode) {
+  if (connectRemoteMode)
+    connect(robot, &WbRobot::toggleRemoteMode, this, &WbSensor::toggleRemoteMode, Qt::UniqueConnection);
+  connect(robot, &WbRobot::robotResetted, this, &WbSensor::reset);
+}
+
+void WbSensor::reset() {
+  mLastUpdate = -9999;
```
omichel: Why not simply `-1` or `-INFINITY`?

src/webots/nodes/WbRobot.hpp
omichel: Also, the use of the word robot is overkill as the class is already named Robot...

src/webots/nodes/WbSensor.cpp
DavidMansolino: -1 is not possible because it has to be at least smaller than -timeStep, but infinity is probably better, I just copied the value from the constructor, but I will try changing this in both parts.

PR #1391, opened 2020-02-20
Fix AppVeyor Tag Build

### LINKED ISSUE #1390: AppVeyor Tag Release Triggered by Nightlies
**Describe the Bug**
The `webots-tag` test is triggered by the nightly tags, this should be avoided: https://ci.appveyor.com/project/omichel/webots-9ufow/history

DavidMansolino:
Attempt to fix  #1390.

### DIFF
--- a/.appveyor-build-tag.yml
+++ b/.appveyor-build-tag.yml
@@ -6,6 +6,9 @@ branches:
   except:
     - /nightly_.*/
 
+on:
+  APPVEYOR_REPO_TAG: true
+
 skip_non_tags: true
 
 clone_depth: 1


.appveyor-build-tag.yml
```diff
@@ -6,6 +6,9 @@ branches:
   except:
     - /nightly_.*/
```
DavidMansolino: The problem is that this will filter branches and not tags. But unfortunately it seems not possible to filter tags: https://help.appveyor.com/discussions/questions/2174-exclude-tags-by-name

PR #1393, opened 2020-02-21
Fix macOS Travis CI

DavidMansolino:
**Description**
Travis tests are currently broken because of some bug in Homebrew.
The bug was introduced in https://github.com/Homebrew/homebrew-bundle/pull/643
This is already reported on the Homebrew repository: https://github.com/Homebrew/homebrew-bundle/issues/646
And here is a thread about this in the Travis forum: https://travis-ci.community/t/macos-build-fails-because-of-homebrew-bundle-unknown-command/7296

The workaround, for now, seems to ask for a homebrew update before using it, however, this slows down quite a lot the test (~7 minutes longer, we should, therefore, remove this as soon as the Travis Image is fixed).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,6 +62,7 @@ jobs:
       language: generic     # Travis doesn't support python on macos yet
       addons:
         homebrew:
+          update: true
           packages:
             - clang-format
             - cppcheck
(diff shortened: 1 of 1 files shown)


DavidMansolino:
:heavy_check_mark:: travis_release

PR #1397, opened 2020-02-21
Add libxkbcommon-x11-0 to dependency of debian-package

junjihashimoto:
**Description**
When we install webots of debian-package, it does not work for not exiting libxkbcommon-x11-0.
https://github.com/cyberbotics/HsWebots/runs/451825815?check_suite_focus=true#step:3:685

**Tasks**
  - [x] Add a dependency of  libxkbcommon-x11-0 to DEBIAN/control-file.


junjihashimoto:
Thank you for your reviewing.

PR #1401, opened 2020-02-21
Feature add metal stainless steel cable appearance

omichel:
This is useful to create metallic cables.


DavidMansolino:
Thank you, this is a very nice material, I am just wondering if the `normalMapFactor` and `occlusionMapStrength` value should not be increased.
Current:
A_1
Increased normal (~3):
A_2
Increased normal (~3) + increased occlusion (~5):
A_3_5

omichel:
Thank you for the comment, it looks indeed much better this way.

Issue #1396, opened 2020-02-21
Support for external meshes


omichel:
In order to facilitate import from different 3D formats and to ease the maintenance of PROTO and world files, we should support external meshes with various format: DAE, STL, OBJ, FBX, etc.
The `Mesh` node would be a geometry node (like `IndexedFaceSet`) and could be defined as follow:

```
Mesh {
  MFString url [ ]       # URL of a 3D file in DAE, STL, OBJ or FBX format (similar to ImageTexture node)
  SFFloat creaseAngle 0  # used if no normals are provided (similar to IndexedFaceSet node)
  SFBool normalPerVertex TRUE # used if normals are provided (similar to IndexedFaceSet node)
}
```

Appearances and other 3D data found in DAE files should be ignored. Multiple meshes should be merged together.

In addition to the `Mesh` node, the `Shape` node should be extended to support a new MFString field called `url`, this field will allow specifying an external file (such as DAE, STL, OBJ, FBX) from which all the meshes and appearances are used.

DavidMansolino:
Partially implemented in #1419.
A second PR should implement the `creaseAngle` and `normalPerVertex` fields (the Mesh node was already implemented in a way that it will be easy to add them).

omichel:
I don't believe it's needed to add the `creaseAngle` and `normalPerVertex` to the `Mesh` node as most 3D files format (like OBJ) should include pre-computed normals. Using pre-computed normals makes that Webots loads meshes twice as fast as with a crease angle computation, which we want to encourage.

PR #1419, opened 2020-02-28
Feature Mesh Node

DavidMansolino:
**Description**
This PR implements the `Mesh` node.

**Related Issues**
This pull-request fixes the first part of #1396

**Tasks**
  - [x] Add Mesh node.
  - [x] Add WREN part.
  - [x] Add ODE part.
  - [x] Merge meshes if more than one mesh is in the file.
  - [x] Create a commun parent class for IdnexedFaceSet and Mesh.
  - [x] Add the `meshes` folder to the concept of Webots project.
  - [x] Sample: https://cyberbotics.com/doc/guide/samples-geometries?version=feature-mesh-node#geometric_primitives-wbt
  - [x] Recompile assimp for Mac including importers.
  - [x] Documentation:
    - https://cyberbotics.com/doc/reference/node-chart?version=feature-mesh-node
    - https://cyberbotics.com/doc/reference/mesh?version=feature-mesh-node
  - [x] Test.
    - [x] Pen.
    - [x] DistanceSensor.
  - [x] Changelog.

As this PR is already quite big, I will limit the following features in the first step:
  - no crease angle field for now.
  - support 'only' the following formats (we may easily add some more in the future if needed):
    - blend
    -  dae
    - fbx
    - obj
    - stl
  - I deliberately didn't add a mesh in the physics_primitives.wbt to avoid encouraging the user to use Meshes as bounding objects.

Except this, everything will be functional (including streaming server).


DavidMansolino:
If a file contains several meshes, they are now correctly merged (including respective rotation, scaling and translation):
image
wb

DavidMansolino:
As this PR is already quite big, I will limit the following features in the first step:
  - no crease angle field for now.
  - support 'only' the following format (we may easily add some more in the future if needed): 
    - blend
    - dae
    - fbx
    - obj
    - stl
  - I deliberately didn't add a mesh in the `physics_primitives.wbt` to avoid encouraging the user to use Meshes as bounding objects.

Except this, everything will be functional (including streaming server).

DavidMansolino:
:heavy_check_mark:: tag creation => release
  - https://travis-ci.com/github/cyberbotics/webots/builds/153481233
  - https://ci.appveyor.com/project/omichel/webots-9ufow/builds/31493672
  - https://github.com/cyberbotics/webots/releases/tag/untagged-681517c9c02141a3f341

tests/api/worlds/pen_mesh.wbt
```diff
@@ -0,0 +1,131 @@
...
+        }
+      }
+      geometry Mesh {
+        url [
+          "meshes/suzanne.obj"
+        ]
+      }
+    }
+  ]
+  name "solid(1)"
```
omichel:
```suggestion
  name "board"
```

DavidMansolino:
> That's a big one...

Yes, I will ask for a second review once you are done.

src/webots/nodes/WbMesh.cpp
```diff
@@ -0,0 +1,415 @@
...
+
+  // create the arrays
+  int current_coord_index = 0;
+  double coord_data[3 * totalVertices];
+  int current_normal_index = 0;
+  double normal_data[3 * totalVertices];
+  int current_tex_coord_index = 0;
+  double tex_coord_data[2 * totalVertices];
+  int current_index_index = 0;
+  unsigned int index_data[3 * totalFaces];
```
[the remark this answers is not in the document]
DavidMansolino: There is no reason at all, I was writing non-camel case python code in parallel of this part and get confused, I will change it.

stefaniapedrazzi:
:eyes:

src/webots/nodes/WbTriangleMeshGeometry.cpp
```diff
@@ -0,0 +1,697 @@
...
+
+  WbGeometry::preFinalize();
+
+  mMeshKey.set(this);
+  WbTriangleMeshCache::useTriangleMesh(this);
+}
+
+void WbTriangleMeshGeometry::postFinalize() {
+  WbGeometry::postFinalize();
+}
```
stefaniapedrazzi: It is not needed to override the `WbGeometry::postFinalize` method if you then just call the base class method.

src/webots/nodes/WbTriangleMeshGeometry.hpp
```diff
@@ -0,0 +1,141 @@
...
+#define WB_TRIANGLE_MESH_GEOMETRY_HPP
+
+#include "WbGeometry.hpp"
+#include "WbTriangleMeshCache.hpp"
+
+#include <unordered_map>
+
+class WbCoordinate;
+class WbNormal;
+class WbTextureCoordinate;
```
stefaniapedrazzi: Are this class include really needed?
It doesn't seem these classes are used here

src/webots/nodes/WbTriangleMeshGeometry.cpp
```diff
@@ -0,0 +1,697 @@
...
+#include "WbResizeManipulator.hpp"
+#include "WbSFBool.hpp"
+#include "WbSFDouble.hpp"
+#include "WbSFNode.hpp"
+#include "WbSimulationState.hpp"
+#include "WbTextureCoordinate.hpp"
+#include "WbTransform.hpp"
+#include "WbTriangleMesh.hpp"
+#include "WbWrenMeshBuffers.hpp"
+#include "WbWrenRenderingContext.hpp"
```
stefaniapedrazzi: It seems this is a copy and paste from `WbIndexedFaceSet`, but here some of these includes are useless (for example WbCoordinate, WbMFInt, etc.)

Also in `WbIndexedFaceSet.cpp` some of the include are no longer needed because the code has been moved here.

src/webots/nodes/WbMesh.cpp
```diff
@@ -0,0 +1,415 @@
...
+#include "WbFieldChecker.hpp"
+#include "WbMFString.hpp"
+#include "WbNodeUtilities.hpp"
+#include "WbRay.hpp"
+#include "WbResizeManipulator.hpp"
+#include "WbSimulationState.hpp"
+#include "WbTransform.hpp"
+#include "WbTriangleMesh.hpp"
+#include "WbUrl.hpp"
+#include "WbVector2.hpp"
```
stefaniapedrazzi: Could you please check again these includes?
Some of them seems useless (for example WbNodeUtilities, but maybe others too)

src/webots/nodes/WbMesh.cpp
```diff
@@ -0,0 +1,415 @@
...
+        assert(face.mNumIndices == 3);
+        indexData[currentIndexIndex++] = face.mIndices[0] + indexOffset;
+        indexData[currentIndexIndex++] = face.mIndices[1] + indexOffset;
+        indexData[currentIndexIndex++] = face.mIndices[2] + indexOffset;
+      }
+
+      indexOffset += mesh->mNumVertices;
+    }
+
+    // add all the childrens of this node to the queue
```
stefaniapedrazzi:
```suggestion
    // add all the children of this node to the queue
```

src/webots/nodes/WbMesh.cpp
```diff
@@ -0,0 +1,415 @@
...
+  const QByteArray meshPath = path().toUtf8();
+  return WbTriangleMeshCache::sipHash13x(meshPath.constData(), meshPath.size());
+}
+
+void WbMesh::exportNodeContents(WbVrmlWriter &writer) const {
+  if (!writer.isVrml()) {
+    WbTriangleMeshGeometry::exportNodeContents(writer);
+    return;
+  }
+  // Export the content as IndexedFaceSet in VRML
```
stefaniapedrazzi:
```suggestion
  // export the content as IndexedFaceSet in VRML
```

just to be consistent with the other comments in this file

src/webots/nodes/WbMesh.cpp
```diff
@@ -0,0 +1,415 @@
...
+  delete[] coordIndex;
+  delete[] normalIndex;
+  delete[] texCoordIndex;
+  delete[] vertex;
+  delete[] normal;
+  delete[] texture;
+}
+
+void WbMesh::rescale(const WbVector3 &scale) {
+}
```
stefaniapedrazzi: This empty method is really needed?
If yes, then there is no need to put in in the CPP file but could be defined directly in the HPP

src/webots/nodes/WbMesh.cpp
```diff
@@ -0,0 +1,415 @@
...
+
+void WbMesh::updateUrl() {
+  // we want to replace the windows backslash path separators (if any) with cross-platform forward slashes
+  int n = mUrl->size();
```
stefaniapedrazzi:
```suggestion
  const int n = mUrl->size();
```

stefaniapedrazzi:
In general it works fine. I tried to import some different free meshes.

But with one particular DAE file (it seems that contains multiple meshes) the meshes are imported correctly but at a wrong position (compared to blender).

DavidMansolino:
Thank you for the tests I can indeed reproduce the issue with the DAE you sent me, the part indicated by the arrow should be 'inside' the gun:
gun
I will check what is hapenning.

DavidMansolino:
Ok, I found and fix the problem, it was a bit tricky, the position was wrong because the mesh was part of an animation and it was exported as DAE with the non-initial position of the animation which was causing the issue.
I managed to say to assimp to exclude animations (among others) which both fix the problem and improve the loading time :smiley:

src/webots/nodes/WbMesh.cpp
DavidMansolino: Good point, I reviewed the headers and severa of them can be removed.

Issue #1417, opened 2020-02-28
Is `Track.textureAnimation` really needed?


omichel:
I am wondering whether we could save the user from setting the `Track.textureAnimation` field. Can't we compute the value of this field automatically? It seems that the motion of the texture should always be in the same way given the orientation of the track.

stefaniapedrazzi:
I'm not sure it can really be computed automatically in such a way that works in all the cases.
Also because a point to take into consideration is that, at least for me as user of Webots, it is much simpler to adapt the `Track.textureAnimation` scale factor rather than modifying the texture mapping that could be very complex.
But I will make some tests to check if there is a way to compute it automatically so that it works for all the possible cases.

Also note, that the `Track.textureAnimation` field also lets the user disable the texture animation, so even if the animation factor is automatically computed it should be replaced by a SFBool field.

omichel:
> Also because a point to take into consideration is that, at least for me as user of Webots, it is much simpler to adapt the Track.textureAnimation scale factor rather than modifying the texture mapping that could be very complex.

But, that would be simply to move the `textureAnimation` values to the `TextureTransform.scale` or so, isn't it?

DavidMansolino:
With @michou214 we just found that there is a bug with this field because the speed of the texture depends on the world time step.
This is because of this line of the `WbTrack::prePhysicsStep` function:
https://github.com/cyberbotics/webots/blob/master/src/webots/nodes/WbTrack.cpp#L617
The duration of the step is not taken into account for the translate of the texture.
Instead it should be:
```diff
diff --git a/src/webots/nodes/WbTrack.cpp b/src/webots/nodes/WbTrack.cpp
index 468139e..6f82b2e 100644
--- a/src/webots/nodes/WbTrack.cpp
+++ b/src/webots/nodes/WbTrack.cpp
@@ -614,7 +614,7 @@ void WbTrack::prePhysicsStep(double ms) {
 
   // texture animation
   if (mTextureTransform) {
-    mTextureTransform->translate(mSurfaceVelocity * mTextureAnimationField->value());
+    mTextureTransform->translate(0.001 * ms * mSurfaceVelocity * mTextureAnimationField->value());
     mTextureTransform->modifyWrenMaterial(mShape->wrenMaterial());
   }
```

With this patch, setting the `textureAnimation` to 1.0 seems to work (i.e. match with the physics behavior) in all cases.

remi-cyberbotics:
:eyes:

[...]

PR #1421, opened 2020-03-02
Fix textureAnimation field and hide them.

michou214:
**Description**
In factory's objects (`ConveyorBlet` and `ConveyorPlatform` PROTOs), I fixed the texture animation by correcting the speed and hide this field in the PROTO since the user don't need to have acces to it.

**Documentation**
File : [docs/guide/object-factory.md](https://cyberbotics.com/doc/guide/object-factory?version=michou214:fix-converoyPlatform)


michou214:
A bug was found during the review. This PR is closed since it will be resolved by #1417

Issue #1439, opened 2020-03-08
Race condition between Supervisor::simulationReset() and Robot::step()


ethanwu10:
**Describe the Bug**
When `Robot::step()` / `wb_robot_step()` is called immediately after `Supervisor::simulationReset()` / `wb_supervisor_simulation_reset()`, and Webots is under relatively high CPU load, the step will wait for the simulation time to reach `duration` ms after the time at which the reset was issued, even if the simulation time has been reset to 0s

**Steps to Reproduce**
1. Load the provided controller code into a supervisor node in a resource-intensive scene (I used the Mavic demo on my laptop—which runs at about 0.5x—and was able to reproduce very reliably)
2. Run the simulation
3. Observe the delay between printed messages - when the race condition occurs, there will be a 5s delay between "Reset" being printed and "Time: 5.056000..." as the simulation clock runs back up to 5s from 0s
4. Observe the simulation time after the controller pauses the simulation - when the race condition occurs, the simulation will stop at 5.088s

**Expected behavior**
The simulation should be stopped at 0.032s, the messages should print with minimal delay between them (should be reproducible when the controller is loaded into a completely empty scene)

**System**
 - Operating System: Arch Linux ([PKGBUILD here](https://github.com/ethanwu10/aur-webots-bin))
 - Graphics Card: Intel Integrated UHD Graphics 620

Controller code:
```c
#include <webots/robot.h>
#include <webots/supervisor.h>
#include <stdio.h>

int main()
{
	wb_robot_init();
	while (wb_robot_step(32) != -1 && wb_robot_get_time() < 5);
	wb_supervisor_simulation_reset();
	printf("Reset\n");
	wb_robot_step(32);
	printf("Time: %f (should be 5.056)\n", wb_robot_get_time());
	wb_robot_step(32);
	printf("Time: %f (should be 0.032)\n", wb_robot_get_time());
	wb_supervisor_simulation_set_mode(WB_SUPERVISOR_SIMULATION_MODE_PAUSE);
	wb_robot_cleanup();
	return 0;
}
```

DavidMansolino:
I can indeed reproduce the issue your are mentioning.
However, the issue is visible only if the controller step (i.e. 32 in your example) is bigger than the simulation step (i.e. 8 in the Mavic demo case).
As a temporary workaround, I would suggest to use the same time step for the supervisor controller and the simulation.

DavidMansolino:
The issue was fixed in #1448.
[...]

PR #1448, opened 2020-03-11
Fix Simulation Reset Timming

DavidMansolino:
**Description**
When the supervisor controller resetting the simulation was using a bigger step value than the world, its step was then very long because the controller 'request time' was stuck at the time before the reset.

**Related Issues**
This pull-request fixes issue #1439


DavidMansolino:

Issue #1454, opened 2020-03-13
Resetting Motorized BallJoint Makes Webots Crash


DavidMansolino:
**Describe the Bug**
Resetting (either from Supervisor or GUI) a motorized balljoint node makes Webots crash.

**Steps to Reproduce**
1. Open the `motor3.wbt` world.
2. Run the simulation.
3. Reset the simulation.

**Expected behavior**
No crash.

**Additional context**
Stack:
```
Thread 1 "webots-bin" received signal SIGABRT, Aborted.
0x00007f490fde7428 in __GI_raise (sig=sig@entry=6)
    at ../sysdeps/unix/sysv/linux/raise.c:54
54	../sysdeps/unix/sysv/linux/raise.c: Aucun fichier ou dossier de ce type.
(gdb) where
#0  0x00007f490fde7428 in __GI_raise (sig=sig@entry=6)
    at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007f490fde902a in __GI_abort () at abort.c:89
#2  0x00007f4910ab1151 in dDebug (num=2, 
    msg=0x7f4910c2e788 "joint type is not AMotor in %s()")
    at ode/src/error.cpp:96
#3  0x00007f4910b10869 in dJointSetAMotorParam (j=0xb1f19d0, parameter=5, 
    value=0) at ode/src/joints/amotor.cpp:340
#4  0x0000000000994121 in WbRotationalMotor::turnOffMotor (this=0x7888b80)
    at nodes/WbRotationalMotor.cpp:73
#5  0x0000000000901e37 in WbMotor::reset (this=0x7888b80)
    at nodes/WbMotor.cpp:533
#6  0x00000000008ad639 in WbJoint::reset (this=0x7888840)
    at nodes/WbJoint.cpp:86
#7  0x000000000089c230 in WbHinge2Joint::reset (this=0x7888840)
    at nodes/WbHinge2Joint.cpp:464
#8  0x00000000007ec0f0 in WbBallJoint::reset (this=0x7888840)
    at nodes/WbBallJoint.cpp:655
#9  0x0000000000888840 in WbGroup::reset (this=0x7886e60)
    at nodes/WbGroup.cpp:271
#10 0x00000000008e7458 in WbMatter::reset (this=0x7886e60)
    at nodes/WbMatter.cpp:150
#11 0x00000000009d8f86 in WbSolid::reset (this=0x7886e60)
```

PR #1455, opened 2020-03-13
Fix motorized ball joint revert

DavidMansolino:
**Description**
Fixed crash when resetting worlds with motorized BallJoint nodes.
The BallJoint is a special case, because the motor control is not applied to the main joint but on an auxiliary joint.

**Related Issues**
This pull-request fixes issue #1454


DavidMansolino:
:heavy_check_mark:: `travis_unit`

PR #1463, opened 2020-03-16
Feature 3d Model Import

DavidMansolino:
**Description**
This PR adds the possibility to import 3D models (blend, dae, fbx, obj, stl) directly in Webots.

Remaining tasks:
   - [x] Doc: https://cyberbotics.com/doc/guide/the-user-interface?version=feature-3d-model-import#file-menu
   - [x] Changelog.
   - [x] Appearance.
   - [x] Integration in Webots GUI.
   - [x] Option (e.g. set bounding objects).
   - [x] Make USE-DEF machanism more robust.
   - [x] Update Linux to assimp 5.0.1. Wiki update:
     - https://github.com/cyberbotics/webots/wiki/Assimp-compilation/_compare/2415b88954335ac2630acfe5bc012984b6689325...ab5644b898e4eaccbabb6f6a3dda3839b929ad50


DavidMansolino:
Since I am recompiling assimp, I am wondering which formats we want to support (I made the importer generic enough so that adding new formats is just changing the filter of the file input dialog), here are the formats available in the latest version (5.0.1) of assimp (this version is available in MSYS and includes all the importer):
```
importer formats: AMF 3DS AC ASE ASSBIN B3D BVH COLLADA DXF CSM HMP IRRMESH IRR
  LWO LWS MD2 MD3 MD5 MDC MDL NFF NDO OFF OBJ OGRE OPENGEX PLY MS3D COB
  BLEND IFC XGL FBX Q3D Q3BSP RAW SIB SMD STL TERRAGEN 3D X X3D GLTF 3MF
  MMD STEP
```

For now I have enabled:
  - Blender
  - FBX
  - STL
  - OBJ

Because they are the one we are using and URDF is using. But it might be interesting to enable others.

@cyberbotics/maintainers what do you thing?

stefaniapedrazzi:
I would also enable X3D (if it is not already).

BVH is interesting for skin / human animation.
And many users are using Autodesk (so 3DS), but they can probably convert their models in one of the other supported formats.

omichel:
I believe we should also add COLLADA (which is widely used, also with URDF), DFX (widely used in AutoCAD) and STEP (widely used as well).

omichel:
According this [article](https://all3dp.com/3d-file-format-3d-files-3d-printer-3d-cad-vrml-stl-obj/), it seems we are pretty good.

DavidMansolino:
✔️: travis_release (https://travis-ci.com/github/cyberbotics/webots/builds/155023952)

docs/guide/the-user-interface.md
```diff
@@ -77,13 +77,25 @@ In order to reset the simulation, the following steps are performed:
...
+Once imported, these objects appear as [Group](../reference/group.md), [Transform](../reference/transform.md), [Solid](../reference/solid.md) or [Shape](../reference/shape.md) nodes at the bottom of the scene tree.
+You can then either turn these objects into Webots nodes (e.g. [Robot](../reference/robot.md)) or cut and paste them into the `children` list of existing Webots nodes.
+The following formats are supported:
+  - [3D Studio mesh](https://wiki.fileformat.com/3d/3ds) files (.3ds).
+  - [AutoCAD DXF](https://en.wikipedia.org/wiki/AutoCAD_DXF) files (.dxf).
+  - [Blender](https://www.blender.org/) files (.blend).
+  - [Biovision Hierarchy](https://en.wikipedia.org/wiki/Biovision_Hierarchy) files (.bhv).
+  - [Collada](https://en.wikipedia.org/wiki/COLLADA) files (.dae).
+  - [Filmbox](https://en.wikipedia.org/wiki/FBX) files (.fbx).
+  - [STEP](https://en.wikipedia.org/wiki/ISO_10303-21) files (.stp).
```
omichel:
```suggestion
  - [STEP](https://en.wikipedia.org/wiki/ISO_10303-21) files (.stp or .step).
```

docs/guide/the-user-interface.md
```diff
@@ -77,13 +77,25 @@ In order to reset the simulation, the following steps are performed:
...
+  - [STL](https://en.wikipedia.org/wiki/STL_(file_format)) files (.stl).
+  - [VRML](https://en.wikipedia.org/wiki/VRML) files (.vrml).
```
omichel:
```suggestion
  - [VRML](https://en.wikipedia.org/wiki/VRML) files (.wrl).
```

src/webots/gui/WbImportWizard.cpp
```diff
@@ -0,0 +1,198 @@
...
+                                    "Filmbox (*.fbx *.FBX);;"
+                                    "STEP (*.stp *.step *.STP *.STEP);;"
+                                    "STL (*.stl *.STL);;"
+                                    "VRML (*.wrl *.WRL);;"
+                                    "Wavefront (*.obj *.OBJ);;"
+                                    "X3D (*.x3d *.X3D)"));
+  if (fileName.isEmpty())
+    return;
+  mFileEdit->setText(fileName);
+  mConclusionLabel->setText(tr("The '%1' file will now be imported at the end of the scene-tree.").arg(fileName));
```
omichel:
```suggestion
  mConclusionLabel->setText(tr("The '%1' file will now be imported at the end of the scene tree.").arg(fileName));
```

DavidMansolino:
> when I tried to import a STEP file, it failed with this error: "WARNING: Invalid data, please verify mesh file (bone weights, normals, ...): Validation failed: A node of the scenegraph is NULL".

I tried it and I have the same issue and can reproduce it with many other step file I can find on internet.
Unfortunately, this error comes from assimp itself. I have tried disabling the validation of the files, then the importation from assimp works fine, but we have to be extremely carful and add a lot of checks to make sure each nodes, parameters, etc. are valid. I tried it and in the case of the step files of the NAS, nothing is working as none of the nodes are valid. I checked and a lot of peoples are saying that the step support of assimp is not functional, see for example: https://github.com/assimp/assimp/issues/1183 and https://github.com/assimp/assimp/issues/2734

Maybe we should unfortunately avoid supporting this format. I checked all the other formats and all the others work fine.
But I think we should also drop DXF as from Wikipedia DXF is getting outdated. Furthermore, DXF is pointless for Webots as it is mainly used for 2D drawing.

omichel:
OK, so let's drop STEP and DXF formats.

DavidMansolino:
(I just launched a travis test to be sure, I will merge once it passes).

Issue #1409, opened 2020-02-26
Inertial Unit Returns Nan if Gravity is Null


DavidMansolino:
**Describe the Bug**
If the `WorldInfo.gravity` field is set to `0 0 0` the inertial unit device always return Nan values.

DavidMansolino:
The problematic part is here:
https://github.com/cyberbotics/webots/blob/010d7cb9c432299631572cf3c5c996073d0bfdf6/src/webots/nodes/WbInertialUnit.cpp#L126-L131
If the `gravity` vector is null we should probably use default gravity.

PR #1466, opened 2020-03-16
Default Y value for gravity now equals 1.0

remi-cyberbotics:
This pull-request fixes issue #1409

### DIFF
--- a/src/webots/nodes/WbInertialUnit.cpp
+++ b/src/webots/nodes/WbInertialUnit.cpp
@@ -128,9 +128,8 @@ void WbInertialUnit::computeValue() {
   const WbVector3 &north = wi->northDirection().normalized();
   WbVector3 minusGravity = -wi->gravity().normalized();
 
-  if (minusGravity.isNan()) {
+  if (minusGravity.isNan())
     minusGravity = WbVector3(0.0, 1.0, 0.0);
-  }
   WbMatrix3 rm(north, minusGravity, north.cross(minusGravity));  // reference frame
   rm.transpose();
   const WbMatrix3 &e = rotationMatrix() * rm;  // extrensic rotation matrix e = Y(yaw) Z(pitch) X(roll) w.r.t reference frame
--- a/src/webots/nodes/WbInertialUnit.cpp
+++ b/src/webots/nodes/WbInertialUnit.cpp
@@ -126,8 +126,11 @@ void WbInertialUnit::computeValue() {
   // get north and -gravity in global coordinate systems
   const WbWorldInfo *const wi = WbWorld::instance()->worldInfo();
   const WbVector3 &north = wi->northDirection().normalized();
-  const WbVector3 &minusGravity = -wi->gravity().normalized();
+  WbVector3 minusGravity = -wi->gravity().normalized();
 
+  if (minusGravity.isNan()) {
+    minusGravity = WbVector3(0.0, 1.0, 0.0);
+  }
   WbMatrix3 rm(north, minusGravity, north.cross(minusGravity));  // reference frame
   rm.transpose();
   const WbMatrix3 &e = rotationMatrix() * rm;  // extrensic rotation matrix e = Y(yaw) Z(pitch) X(roll) w.r.t reference frame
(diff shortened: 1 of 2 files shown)


DavidMansolino:
@remi-cyberbotics if this PR is ready do not forget to ask for a review.

src/webots/nodes/WbInertialUnit.cpp
```diff
@@ -126,8 +126,11 @@ void WbInertialUnit::computeValue() {
   // get north and -gravity in global coordinate systems
   const WbWorldInfo *const wi = WbWorld::instance()->worldInfo();
   const WbVector3 &north = wi->northDirection().normalized();
-  const WbVector3 &minusGravity = -wi->gravity().normalized();
+  WbVector3 minusGravity = -wi->gravity().normalized();
 
+  if (minusGravity.isNan()) {
+    minusGravity = WbVector3(0.0, 1.0, 0.0);
+  }
```
DavidMansolino:
```suggestion
  if (minusGravity.isNan())
    minusGravity = WbVector3(0.0, 1.0, 0.0);
```
https://github.com/cyberbotics/webots/wiki/CPP-Coding-Style#cs360-single-if-while-or-for-statement-must-be-written-without-brackets-the-statement-part-should-be-put-on-a-separate-line-in-case-of-nested-block-keep-brackets

remi-cyberbotics: True, I should set it in my auto-indent plugin...

src/webots/nodes/WbInertialUnit.cpp
```diff
@@ -126,8 +126,10 @@ void WbInertialUnit::computeValue() {
   // get north and -gravity in global coordinate systems
   const WbWorldInfo *const wi = WbWorld::instance()->worldInfo();
   const WbVector3 &north = wi->northDirection().normalized();
-  const WbVector3 &minusGravity = -wi->gravity().normalized();
+  WbVector3 minusGravity = -wi->gravity().normalized();
 
+  if (minusGravity.isNan())
+    minusGravity = WbVector3(0.0, 1.0, 0.0);
```
omichel: Shouldn't we set it to (0, 0, 0) instead?

DavidMansolino: No otherwise the rotation matrix computed just after will not be valid and roll pitch yaw will probably always be 0.

omichel: OK, but did you check that the resulting IMU values computed from this arbitrary unit vector are correct? If we set it to (1.0, 0.0, 0.0) for example, do we get the same result?

DavidMansolino: We will not get the same result with `1.0, 0.0, 0.0` but the `0.0, 1.0, 0.0` vector was chosen to have X=roll, Z=pitch, Y=yaw, which is the default behavior and corresponds to what is displayed in the doc (see this image for example: https://cyberbotics.com/doc/reference/inertialunit#field-summary ).

omichel: Yes, but if `WorldInfo.northDirection` is `0 1 0`, then it will not work...
I believe that `northDirection` and `gravity` should be orthogonal and this constraint is not checked by Webots.
To better model this, we should replace the `WorldInfo.northDirection` and `gravity` vectors with a `northOrientation` rotation and a `gravity` scalar value.

omichel: Let's go ahead with this PR as it is (although it looks wrong to me). I will open an issue about a better way to indicate the north direction and gravity in the `WorldInfo` node.

remi-cyberbotics: Ok so I am merging right now then.

DavidMansolino: I agree, there is definitely something going wrong with the `gravity` and `northDirection` (at least in the context of the inertialUnit), but this PR fixes this the original crash and works now fine except in the case where `gravity` and `northDirection` are parallel.

PR #1471, opened 2020-03-18
Fixed Test Suite Crash.

DavidMansolino:
**Description**
This crash was caused by a rare case when a supervisor changes a procedural field, the node was regenerated and the selection was temporarily set to a node whose mesh was not yet initialized.


DavidMansolino:
:heavy_check_mark:: `travis_unit`

PR #1476, opened 2020-03-19
Fix Test Failure rosdep Missing

DavidMansolino:
**Description**
An update of ROS melodic did break the test suite because the `rosdep` command is not available anymore.


DavidMansolino:
This fixes the compilation but now the ROS test remains stuck :cry:

DavidMansolino:
As I was not able to fix the problem for now, I propose as a hotfix to disable the melodic test in the nightly (the kinetic is still working fine) so that the nightly works again.

PR #1482, opened 2020-03-20
fix python2.7 should be optional to compile webots

### LINKED ISSUE #1470: Mandatory Python2.7 Dependency
**Describe the Bug**
Python2.7 is supposed to be an optional dependency to compile Webots, however, the compilation fails if ti is not available due to this line:
https://github.com/cyberbotics/webots/blob/master/projects/robots/robotis/darwin-op/libraries/Makefile#L31

remi-cyberbotics:
Closes #1470 

This fix will not block the make process when compiling webots with no python2.7 installed.


remi-cyberbotics:
Should I merge now or try to investigate further ?

omichel:
Yes, you can merge it now.

PR #1484, opened 2020-03-20
Add gyroscope to e-puck proto

lukicdarkoo:
**Description**
According to e-puck documentation both versions of e-puck, v1 and v2, are equipped with a gyroscope. However, the gyroscope is not present in e-puck proto.

Support for gyroscope is also implemented for ROS2 e-puck2 driver (physical robot):
https://github.com/cyberbotics/epuck_ros2/issues/5

**Tasks**
  - [x] Add gyroscope to e-puck proto
  - [x] Verify if the lookup tables for accelerometer and gyroscope have to be improved

**Screenshots**
image


lukicdarkoo:
Currently, Webots Gyro and Accelerometer nodes return actual values. However, the sensor's raw format is different. E.g. here how the conversion is done in `epuck_ros2`:
https://github.com/cyberbotics/epuck_ros2/blob/1d51f52f575a9cf57d2871aa7de5cf0012edc57f/epuck_ros2_cpp/src/mpu9250.cpp#L85

Should we go with raw values or stick with actual?

DavidMansolino:
In the simulation model of the e-puck we should return the exact same value than the one returned by the real e-puck API (which I assume are raw value). It should be possible to calibrate the lookup table to do the conversion between real value and raw values.

DavidMansolino:
👀

lukicdarkoo:
I gave myself a bit of freedom on the coding side of the window. Let me know if it has to be corrected. Here is the result:
image

Also, I removed the scaling of accelerometer as we discussed, to keep backwards compatibility and demos were failing.

DavidMansolino:
> I gave myself a bit of freedom on the coding side of the window. Let me know if it has to be corrected. Here is the result:
> image
> 
> Also, I removed the scaling of accelerometer as we discussed, to keep backwards compatibility and demos were failing.

From your preview it looks good to me.

DavidMansolino:
👀

akshattodi:
Can anyone tell me how to get the values of gyroscope x, y, z, axis values with python in webots

Issue #1479, opened 2020-03-19
Restore Melodic Test Suite


DavidMansolino:
**Describe the Bug**
Since an update of ROS melodic (https://discourse.ros.org/t/new-packages-for-melodic-2020-03-18/13275) our ROS test suite is not working anymore.
This has been temporarily fixed in #1476 we should try again in a few days if a new version of ROS melodic has fixed the problem.

DavidMansolino:
Fixed in  #2350

PR #1486, opened 2020-03-23
Restore ROS Melodic Test

DavidMansolino:
**Description**
A new release of Melodic was released this week-end to fix (hopefully) #1479: https://discourse.ros.org/t/new-packages-for-melodic-2020-03-22/13329

Edit 31.03: Seems to work fine when launched manually with a VM of Ubuntu 18.04.


DavidMansolino:
Unfortunately, it doesn't seem to be fixed, but sometimes it takes 2-3 days before the update is available to apt. so I will leave this PR open and retry tomorrow.

DavidMansolino:
Still not working, If it is not working by the end of the week I will install a VM with ROS melodic to better understand what is hapenning.

DavidMansolino:
It seems the simulation and ROS node are running correctly with the correct behaviour, but for some reason, the node doesn't exists which make the test fail because of the timeout.

DavidMansolino:
From this log:
https://travis-ci.com/github/cyberbotics/webots/jobs/316758766#L7341
It seems that both Webots and the complete_test are correctly started, but simulation is not running.

DavidMansolino:
Replaced by #1878

PR #1489, opened 2020-03-24
Fix AppVeyor Tests

DavidMansolino:
**Description**
This is just a test.

### DIFF
--- a/tests/sources/generate_diff.py
+++ b/tests/sources/generate_diff.py
@@ -44,7 +44,7 @@ def github_api(request):
     content = ''
     try:
         response = urlopen(req)
-        content = response.read()
+        content = response.read().decode()
     except HTTPError as e:
         print(request)
         print(e.reason)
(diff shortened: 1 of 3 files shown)


tests/sources/generate_diff.py
```diff
@@ -44,13 +44,13 @@ def github_api(request):
...
         response = urlopen(req)
-        content = response.read()
+        content = response.read().decode()
     except HTTPError as e:
         print(request)
         print(e.reason)
         print(e.info())
     github_api.last_time = time.time()
-    return json.loads(content.decode())
+    return json.loads(content)
```
DavidMansolino: In case of exception, ``content`` is the empty string so it can't be decoded.

DavidMansolino:
The problem was due to the fact that the ``GITHUB_API_KEY`` was not set in AppVeyor and therefore it was using the ip hourly rate (which is very low , only 60) instead of the account rate limit.

omichel:
Well spotted!

PR #1491, opened 2020-03-24
Improve Rotation Edition from GUI

DavidMansolino:
**Description**
I have these changes locally since a long time because it is often useful for me to be able to set the rotation fields using quaternion (especially when using model coming from URDF).

As it might be useful to someone else than me I think it is worth merging it.
This PR adds the possibility to set if we want to specify the rotation using axis-angle or quaternions (we may also add matrix and Euler angles in the future if needed) in the field editor (of course in the scene-tree it remains always expressed as axis-angle):
Screenshot 2020-03-24 at 16 33 34


DavidMansolino:
✔️: travis_release (https://travis-ci.com/github/cyberbotics/webots/builds/154824173)

DavidMansolino:
Both issues are already present on develop and are reproducible when changing any rotation field.
1. If I change for example the 'y' value of this rotation field on develop '0 1 0 0' using the +/- buttons it switch directly to 0 and then -1, to have intermediate value I have to manually edit.
2. On develop when you change a rotation field it is normalised in the tree but not in the editor, you have to unselect the field and re-select it to show the normalised value in the editor.

I will check what I can do for issue 1, and add the button for issue 2 (even so I am a bit afraid that this button will complexity the interface and confuse the user).

omichel:
Yes, these issues were already there. However, it would be nice to take the opportunity of this PR to address them. I believe the +/- buttons should work the same way as for the translation x/y/z fields, i.e., compute the increment/decrement from the number of decimals. I believe a "normalize" button has the advantage of being self explanatory. Users may not understand why the values they enter in the field editor are changed. With a "normalize" button, they will understand that they are normalized, thus explaining the change. Like if they click the "normalize" button, values won't be further changed.

DavidMansolino:
The first issue was a bit tricky to resolve because it was not on the editor side but on the `WbFieldDoubleSpinBox` side, but it is now fixed and acts exactly like the translation.

✔️: travis_unit (https://travis-ci.com/github/cyberbotics/webots/builds/155061315)

PR #1492, opened 2020-03-25
Enhancement webots console ansi codes

### LINKED ISSUE #1481: Add optional timestamps to console messages
When loading a world the console does not get cleaned automatically, it is not a problem but maybe some users would like to associate timestamps to the output, so it matches the moment when they are working on simulations. It should be a setting to match the user's preference.

remi-cyberbotics:
This PR will improve the webots console outputting part, allowing it to support some more ANSI control codes.

This will also improve documentation about the capabilities of the console as well as including some convenient headers in different languages to use those efficiently.

As suggested in the linked issue we could also implements some optional timestamps for the output.

Close #1481


src/webots/gui/WbConsole.cpp
```diff
@@ -410,54 +418,89 @@ void WbConsole::handlePossibleAnsiEscapeSequences(const QString &msg, WbLog::Lev
...
         html += htmlSpan(remains, level);
         handleCRAndLF(html);
         return;
       }
-      if (j != i)
+      if (j != i) {  // Extract text contained between two escape codes if so
         html += htmlSpan(msg.mid(j, i - j), level);
+      }
     }
-    handleCRAndLF(html);
```
remi-cyberbotics: This was never used since we return in the above if when we found the last escape character, otherwise if there was not in the raw string, we pass in the underneath else.

omichel: Well spotted. It seems we could even remove the following `else`.

projects/samples/devices/controllers/console_capabilities/console_capabilities.c
```diff
@@ -0,0 +1,58 @@
+/*
```
DavidMansolino: This controller is missing a ``.gitignore`` file 😉

docs/guide/controller-programming.md
```diff
@@ -452,6 +452,19 @@ if (finished) {
 
 Note that the exit status as well as the value returned by the `main` function are ignored by Webots.
 
+### Console Output
+
+As mentionned above, outputting something from the controller will be redirected by Webots to the console. The Webots console can not be used interactively but still supports some basic ANSI sequences. One can output text with certain style, using color for example.
+
+Currently the console supports the following:
```
omichel:
```suggestion
As mentioned earlier, printing to `stdout` or `stderr` from a controller will be redirected by default to the Webots console.
Note however, that the Webots console doesn't support `stdin` input.
Like most terminals, it supports a few basic [ANSI escape codes](https://en.wikipedia.org/wiki/ANSI_escape_code) for setting text styles and clearing the content of the console:
```

docs/guide/controller-programming.md
```diff
@@ -452,6 +452,19 @@ if (finished) {
...
+  - 8 bit color (foreground and background)
+  - Bold style
+  - Underline style
+  - Clear screen (same as issuing `clear` command in your terminal)
+  - Resetting style
+
+There is a world example in webots: samples/devices/console_capabilities or at `$WEBOTS_HOME/projects/samples/devices/controllers/console_capabilities` to demonstrate how to use those.
```
omichel: I would rename this example to `console` which is less verbose than `console_capabilities` and is more consistent with other device examples.

projects/samples/devices/controllers/console_capabilities/ansi_codes.h
```diff
@@ -0,0 +1,59 @@
+/*
```
omichel: This header file should be useful to other controllers, hence we should include it in the Webots API, and not in a specific controller folder.

projects/samples/devices/controllers/console_capabilities/ansi_codes.h
```diff
@@ -0,0 +1,59 @@
...
+
+// Convenient macros
+#define PRINT_IN_BLACK(x) puts(ANSI_COLOR_FG_BLACK x ANSI_STYLE_RESET)
+#define PRINT_IN_RED(x) puts(ANSI_COLOR_FG_RED x ANSI_STYLE_RESET)
+#define PRINT_IN_GREEN(x) puts(ANSI_COLOR_FG_GREEN x ANSI_STYLE_RESET)
+#define PRINT_IN_YELLOW(x) puts(ANSI_COLOR_FG_YELLOW x ANSI_STYLE_RESET)
+#define PRINT_IN_BLUE(x) puts(ANSI_COLOR_FG_BLUE x ANSI_STYLE_RESET)
+#define PRINT_IN_MAGENTA(x) puts(ANSI_COLOR_FG_MAGENTA x ANSI_STYLE_RESET)
+#define PRINT_IN_CYAN(x) puts(ANSI_COLOR_FG_CYAN x ANSI_STYLE_RESET)
+#define PRINT_IN_WHITE(x) puts(ANSI_COLOR_FG_WHITE x ANSI_STYLE_RESET)
```
omichel: Instead of relying on `puts`, I would base such macros on `printf` which is more common and more powerful and hence rename them into `ANSI_PRINTF_IN_*`. I believe we should use the `ANSI` prefix for all macros in this file to avoid name clash.

docs/guide/controller-programming.md
```diff
@@ -452,6 +452,19 @@ if (finished) {
 
 Note that the exit status as well as the value returned by the `main` function are ignored by Webots.
 
+### Console Output
+
+As mentionned above, outputting something from the controller will be redirected by Webots to the console. The Webots console can not be used interactively but still supports some basic ANSI sequences. One can output text with certain style, using color for example.
+
+Currently the console supports the following:
+  - 8 bit color (foreground and background)
```
omichel: You didn't implemented 8-bit color, but 3-bit color.

docs/guide/controller-programming.md
```diff
@@ -452,6 +452,19 @@ if (finished) {
...
+  - Bold style
+  - Underline style
+  - Clear screen (same as issuing `clear` command in your terminal)
+  - Resetting style
```
omichel:
```suggestion
  - Reset (colors and styles)
```

projects/samples/devices/controllers/console_capabilities/ansi_codes.h
```diff
@@ -0,0 +1,59 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef ANSI_CODES_H
+#define ANSI_CODES_H
+
+#include <stdio.h>
+
+#define ANSI_STYLE_RESET "\x1b[0m"
```
omichel:
```suggestion
#define ANSI_RESET "\x1b[0m"
```

projects/samples/devices/controllers/console_capabilities/ansi_codes.h
```diff
@@ -0,0 +1,59 @@
...
+
+#define ANSI_BOLD_STYLE "\x1b[1m"
```
omichel: The `STYLE` keyword is not useful:
```suggestion
#define ANSI_BOLD "\x1b[1m"
```

projects/samples/devices/controllers/console_capabilities/ansi_codes.h
```diff
@@ -0,0 +1,59 @@
...
+#define ANSI_UNDERLINE_STYLE "\x1b[4m"
```
omichel:
```suggestion
#define ANSI_UNDERLINE "\x1b[4m"
```

projects/samples/devices/controllers/console_capabilities/ansi_codes.h
```diff
@@ -0,0 +1,59 @@
...
+
+#define ANSI_COLOR_BG_BLACK "\x1b[40m"
```
omichel: We try to avoid abbreviations:
```suggestion
#define ANSI_BLACK_BACKGROUND "\x1b[40m"
```
etc.

projects/samples/devices/controllers/console_capabilities/ansi_codes.h
```diff
@@ -0,0 +1,59 @@
...
+#define PRINT_IN_BLACK(x) puts(ANSI_COLOR_FG_BLACK x ANSI_STYLE_RESET)
+#define PRINT_IN_RED(x) puts(ANSI_COLOR_FG_RED x ANSI_STYLE_RESET)
+#define PRINT_IN_GREEN(x) puts(ANSI_COLOR_FG_GREEN x ANSI_STYLE_RESET)
+#define PRINT_IN_YELLOW(x) puts(ANSI_COLOR_FG_YELLOW x ANSI_STYLE_RESET)
+#define PRINT_IN_BLUE(x) puts(ANSI_COLOR_FG_BLUE x ANSI_STYLE_RESET)
+#define PRINT_IN_MAGENTA(x) puts(ANSI_COLOR_FG_MAGENTA x ANSI_STYLE_RESET)
+#define PRINT_IN_CYAN(x) puts(ANSI_COLOR_FG_CYAN x ANSI_STYLE_RESET)
+#define PRINT_IN_WHITE(x) puts(ANSI_COLOR_FG_WHITE x ANSI_STYLE_RESET)
+
+#define CONSOLE_CLEAR puts(ANSI_CLEAR_SCREEN)
```
omichel:
```suggestion
#define ANSI_CLEAR_CONSOLE puts(ANSI_CLEAR_SCREEN)
```

projects/samples/devices/controllers/console_capabilities/console_capabilities.c
```diff
@@ -0,0 +1,58 @@
...
+  printf("%s%sRed background on Blue text%s\n", ANSI_COLOR_BG_RED, ANSI_COLOR_FG_BLUE, ANSI_STYLE_RESET);
+  printf("%s%sGreen background on Yellow%s\n", ANSI_COLOR_BG_GREEN, ANSI_COLOR_FG_YELLOW, ANSI_STYLE_RESET);
+  printf("%s%sYellow background on Magenta text%s\n", ANSI_COLOR_BG_YELLOW, ANSI_COLOR_FG_MAGENTA, ANSI_STYLE_RESET);
+  printf("%s%sCyan background on Red text%s\n", ANSI_COLOR_BG_CYAN, ANSI_COLOR_FG_RED, ANSI_STYLE_RESET);
+  printf("%s%sBlue background on White text%s\n", ANSI_COLOR_BG_BLUE, ANSI_COLOR_FG_WHITE, ANSI_STYLE_RESET);
+  printf("%s%sMangenta background on Green text%s\n", ANSI_COLOR_BG_MAGENTA, ANSI_COLOR_FG_GREEN, ANSI_STYLE_RESET);
+  printf("%s%sWhite background on Black text%s\n", ANSI_COLOR_BG_WHITE, ANSI_COLOR_FG_BLACK, ANSI_STYLE_RESET);
+
+  printf("%s%sBold style green text%s\n", ANSI_COLOR_FG_GREEN, ANSI_BOLD_STYLE, ANSI_STYLE_RESET);
+  printf("%s%sUnderlined style green text%s\n", ANSI_COLOR_FG_GREEN, ANSI_UNDERLINE_STYLE, ANSI_STYLE_RESET);
```
omichel: There is no example of clearing the console here...

projects/samples/devices/worlds/console_capabilities.wbt
```diff
@@ -0,0 +1,196 @@
...
+TexturedBackground {
+}
+TexturedBackgroundLight {
+}
+RectangleArena {
+}
+Robot {
+  translation 0.337657 -7.99361e-15 -0.334416
+  rotation 0 -1 0 -2.570805307179586
+  children [
```
omichel: The robot here is not the interesting part, therefore it should be simplified to the extreme.

projects/samples/devices/worlds/console_capabilities.wbt
```diff
@@ -0,0 +1,196 @@
+#VRML_SIM R2020a utf8
+WorldInfo {
+  info [
+    "The MyBot robot, the robot used in the tutorial of Webots"
+  ]
+  title "MyBot"
```
omichel: This information should be more meaningful.

src/webots/gui/WbConsole.cpp
```diff
@@ -272,29 +272,36 @@ QString WbConsole::htmlSpan(const QString &s, WbLog::Level level) const {
   if (s.isEmpty() || s == "\n")
     return "";
 
-  QString color;
+  QString fgColor;
```
omichel: https://github.com/cyberbotics/webots/wiki/Coding-Style#cs44-no-abbreviations
```suggestion
  QString foregroundColor;
```

src/webots/gui/WbConsole.cpp
```diff
@@ -410,56 +417,94 @@ void WbConsole::handlePossibleAnsiEscapeSequences(const QString &msg, WbLog::Lev
       foreach (const QString code, codes) {
         // the stored sequence may be "0" or "1", "30", "31", "32", etc.
         if (code == "0") {  // reset to default
-          mColor = ansiBlack();
+          mFgColor = ansiBlack();
+          mBgColor.clear();
           mBold = false;
-        } else if (code == "1")  // bold
+          mUnderline = false;
+        } else if (code == "1")
```
omichel: Please do not remove these useful comments

src/webots/gui/WbConsole.cpp
```diff
@@ -410,56 +417,94 @@ void WbConsole::handlePossibleAnsiEscapeSequences(const QString &msg, WbLog::Lev
...
           mBold = true;
-        else if (code.startsWith("3")) {  // foreground color change
+        else if (code == "4")
+          mUnderline = true;
+        else if (code.startsWith("2")) {
+          const char c = code.toLocal8Bit().data()[1];
+          if (c == 'J') {
```
omichel: All these codes should be commented so that we know to what they correspond.

projects/samples/devices/worlds/.console_capabilities.wbproj
```diff
@@ -0,0 +1,10 @@
+Webots Project File version R2020a
```
omichel: This `.wbproj` file should be configured to open the controller program in the editor window.

projects/samples/devices/worlds/.console.wbproj
```diff
@@ -0,0 +1,10 @@
+Webots Project File version R2020a
```
remi-cyberbotics: Is it okay to have a big console output window ? Or do we want to see the code part a little more ?

docs/guide/controller-programming.md
```diff
@@ -452,6 +452,19 @@ if (finished) {
...
+As mentioned earlier, printing to `stdout` or `stderr` from a controller will be redirected by default to the Webots console.
+Note however, that the Webots console doesn't support `stdin` input.
+Like most terminals, it supports a few basic [ANSI escape codes](https://en.wikipedia.org/wiki/ANSI_escape_code) for setting text styles and clearing the content of the console:
+  - 3 bit color (foreground and background)
+  - Bold style
+  - Underline style
+  - Clear screen (same as issuing `clear` command in your terminal)
+  - Reset (colors and styles)
+
+There is a world example in webots: samples/devices/console or at `$WEBOTS_HOME/projects/samples/devices/controllers/console` to demonstrate how to use those.
```
DavidMansolino:
```suggestion
There is a world example in webots: samples/devices/console or at "[WEBOTS\_HOME/projects/samples/devices/controllers/console](https://github.com/cyberbotics/webots/tree/master/projects/samples/devices/controllers/console)" to demonstrate how to use those.
```

docs/guide/controller-programming.md
```diff
@@ -452,6 +452,19 @@ if (finished) {
 
 Note that the exit status as well as the value returned by the `main` function are ignored by Webots.
 
+### Console Output
+
```
DavidMansolino: We should probably mention the new header here.

projects/samples/devices/worlds/console.wbt
```diff
@@ -0,0 +1,33 @@
+#VRML_SIM R2020a utf8
```
DavidMansolino: To me this sample as it is not strictly a device, should rather go in ``projects/samples/howto`` (as it is the case for example for the sample showing how to use the mouse (which is not a Webots device too)).

src/webots/gui/WbConsole.cpp
```diff
@@ -408,58 +414,96 @@ void WbConsole::handlePossibleAnsiEscapeSequences(const QString &msg, WbLog::Lev
...
+      if (i == -1) {  // Previous escape code was the last one found
         const QString remains(msg.mid(j));
         html += htmlSpan(remains, level);
         handleCRAndLF(html);
         return;
       }
-      if (j != i)
+      if (j != i) {  // Extract text contained between two escape codes if so
         html += htmlSpan(msg.mid(j, i - j), level);
+      }
```
DavidMansolino:
```suggestion
      if (j != i)  // Extract text contained between two escape codes if so
        html += htmlSpan(msg.mid(j, i - j), level);
```

projects/samples/devices/worlds/.console.wbproj
DavidMansolino: On my small screen the problem is more the width of the text editor, the 3D view is so small that I don't see the play etc. buttons.

docs/guide/controller-programming.md
```diff
@@ -452,6 +452,19 @@ if (finished) {
...
+As mentioned earlier, printing to `stdout` or `stderr` from a controller will be redirected by default to the Webots console.
+Note however, that the Webots console doesn't support `stdin` input.
+Like most terminals, it supports a few basic [ANSI escape codes](https://en.wikipedia.org/wiki/ANSI_escape_code) for setting text styles and clearing the content of the console:
+  - 3 bit color (foreground and background)
```
omichel:
```suggestion
  - 3-bit color (foreground and background)
```

include/controller/c/webots/utils/ansi_codes.h
```diff
@@ -0,0 +1,59 @@
...
+#define ANSI_PRINTF_IN_BLACK(x, ...) printf(ANSI_BLACK_FOREGROUND x ANSI_RESET "\n", ##__VA_ARGS__)
+#define ANSI_PRINTF_IN_RED(x, ...) printf(ANSI_RED_FOREGROUND x ANSI_RESET "\n", ##__VA_ARGS__)
+#define ANSI_PRINTF_IN_GREEN(x, ...) printf(ANSI_GREEN_FOREGROUND x ANSI_RESET "\n", ##__VA_ARGS__)
+#define ANSI_PRINTF_IN_YELLOW(x, ...) printf(ANSI_YELLOW_FOREGROUND x ANSI_RESET "\n", ##__VA_ARGS__)
+#define ANSI_PRINTF_IN_BLUE(x, ...) printf(ANSI_BLUE_FOREGROUND x ANSI_RESET "\n", ##__VA_ARGS__)
+#define ANSI_PRINTF_IN_MAGENTA(x, ...) printf(ANSI_MAGENTA_FOREGROUND x ANSI_RESET "\n", ##__VA_ARGS__)
+#define ANSI_PRINTF_IN_CYAN(x, ...) printf(ANSI_CYAN_FOREGROUND x ANSI_RESET "\n", ##__VA_ARGS__)
+#define ANSI_PRINTF_IN_WHITE(x, ...) printf(ANSI_WHITE_FOREGROUND x ANSI_RESET "\n", ##__VA_ARGS__)
+
+#define ANSI_CONSOLE_CLEAR printf(ANSI_CLEAR_SCREEN "\n")
```
omichel: I prefer to use MY_MACRO(), with parentheses, because it feels more like I'm calling a function. Otherwise it looks like I'm calling a constant. Also, I setting the verb in second position which makes it more readable.
```suggestion
#define ANSI_CLEAR_CONSOLE() printf(ANSI_CLEAR_SCREEN "\n")
```

projects/samples/devices/controllers/console/console.c
```diff
@@ -0,0 +1,65 @@
...
+
+#include <webots/utils/ansi_codes.h>
+
+#define TIME_STEP 64
+
+int main(int argc, char **argv) {
+  wb_robot_init();
+
+  ANSI_PRINTF_IN_GREEN("This text will not be seen because we are going to clear the console");
+  ANSI_CONSOLE_CLEAR;
```
omichel:
```suggestion
  ANSI_CLEAR_CONSOLE();
```

docs/guide/controller-programming.md
remi-cyberbotics: Just curious about the double quotes here, do we need them ?

DavidMansolino: We don't *need* them, it is just to be coherent with the other references (see for example: https://www.cyberbotics.com/doc/guide/appearances#asphalt).

projects/samples/howto/controllers/console/console.c
```diff
@@ -0,0 +1,65 @@
...
+#include <webots/robot.h>
+
+#include <webots/utils/ansi_codes.h>
+
+#define TIME_STEP 64
+
+int main(int argc, char **argv) {
+  wb_robot_init();
+
+  ANSI_PRINTF_IN_GREEN("This text will not be seen because we are going to clear the console");
```
omichel: It's a pity that we don't see this message at all and maybe skip the fact that the console is actually cleared.
```suggestion
  ANSI_PRINTF_IN_GREEN("This text will disappear in about 4 seconds because we are going to clear the console...");
  wb_robot_step(4096);
```

include/controller/c/webots/utils/ansi_codes.h
```diff
@@ -0,0 +1,59 @@
...
+#define ANSI_PRINTF_IN_BLACK(x, ...) printf(ANSI_BLACK_FOREGROUND x ANSI_RESET "\n", ##__VA_ARGS__)
+#define ANSI_PRINTF_IN_RED(x, ...) printf(ANSI_RED_FOREGROUND x ANSI_RESET "\n", ##__VA_ARGS__)
+#define ANSI_PRINTF_IN_GREEN(x, ...) printf(ANSI_GREEN_FOREGROUND x ANSI_RESET "\n", ##__VA_ARGS__)
+#define ANSI_PRINTF_IN_YELLOW(x, ...) printf(ANSI_YELLOW_FOREGROUND x ANSI_RESET "\n", ##__VA_ARGS__)
+#define ANSI_PRINTF_IN_BLUE(x, ...) printf(ANSI_BLUE_FOREGROUND x ANSI_RESET "\n", ##__VA_ARGS__)
+#define ANSI_PRINTF_IN_MAGENTA(x, ...) printf(ANSI_MAGENTA_FOREGROUND x ANSI_RESET "\n", ##__VA_ARGS__)
+#define ANSI_PRINTF_IN_CYAN(x, ...) printf(ANSI_CYAN_FOREGROUND x ANSI_RESET "\n", ##__VA_ARGS__)
+#define ANSI_PRINTF_IN_WHITE(x, ...) printf(ANSI_WHITE_FOREGROUND x ANSI_RESET "\n", ##__VA_ARGS__)
+
+#define ANSI_CLEAR_CONSOLE() printf(ANSI_CLEAR_SCREEN "\n")
```
omichel:
```suggestion
#define ANSI_CLEAR_CONSOLE() printf(ANSI_CLEAR_SCREEN)
```

remi-cyberbotics:
Last thing would be to implement (optional) timestamps in the console output, and translate the C header to the different languages, should we do it now ?

src/webots/gui/WbConsole.cpp
```diff
@@ -272,29 +272,35 @@ QString WbConsole::htmlSpan(const QString &s, WbLog::Level level) const {
...
+  if (!foregroundColor.isEmpty() || bold || mUnderline) {
     span += " style=\"";
-    if (!color.isEmpty())
-      span += "color:" + color + ";";
+    if (!foregroundColor.isEmpty() && mBackgroundColor.isEmpty())
+      span += "color:" + foregroundColor + ";";
+    else if (!foregroundColor.isEmpty() && !mBackgroundColor.isEmpty()) {
+      span += "color:" + foregroundColor + ";";
+      span += "background-color:" + mBackgroundColor + ";";
+    }
```
omichel: That part of code looks odd to me. Can't we simplify it like that?
```suggestion
    if (!foregroundColor.isEmpty())
      span += "color:" + foregroundColor + ";";
    if (!mBackgroundColor.isEmpty())
      span += "background-color:" + mBackgroundColor + ";";
```

omichel:
As explained in #1481, I would not implement the timestamp in the console.
For the other languages, yes, we should certainly implement it now in this PR.

include/controller/c/webots/utils/ansi_codes.h
```diff
@@ -0,0 +1,59 @@
...
+#define ANSI_YELLOW_FOREGROUND "\x1b[33m"
+#define ANSI_BLUE_FOREGROUND "\x1b[34m"
+#define ANSI_MAGENTA_FOREGROUND "\x1b[35m"
+#define ANSI_CYAN_FOREGROUND "\x1b[36m"
+#define ANSI_WHITE_FOREGROUND "\x1b[37m"
+
+#define ANSI_CLEAR_SCREEN "\x1b[2J"
+
+// Convenient macros
+#define ANSI_PRINTF_IN_BLACK(x, ...) printf(ANSI_BLACK_FOREGROUND x ANSI_RESET "\n", ##__VA_ARGS__)
```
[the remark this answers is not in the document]
remi-cyberbotics: True it is not the same behavior than printf, however if we don't do this way, the user would systematically need to specify ANSI_RESET along with a `"\n"`.
So an example to just output in Red without the `"\n"` in the macro would be:
`ANSI_PRINTF_IN_RED("Red text!%s\n", ANSI_RESET);`
Otherwise just specifying `"\n"` would color the controller name like here:

src/webots/gui/WbConsole.cpp
remi-cyberbotics: True it is more clear, on it!

include/controller/c/webots/utils/ansi_codes.h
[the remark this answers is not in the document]
omichel: Indeed, this is due to some bad management of the [controller name] prefix with respect to colors... A bug that was here before this PR...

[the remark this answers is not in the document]
omichel: I just found it and I believe I can fix it easily. I'll do that on this PR.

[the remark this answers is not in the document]
remi-cyberbotics: Okay Olivier, otherwise I would have implemented it since I tested it a little, but if you are on it...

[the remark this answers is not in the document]
omichel: Bug fixed, commited there.

include/controller/c/webots/utils/ansi_codes.h
remi-cyberbotics: So do we want to print anything before cleaning ?
It sounds stupid put it like this, but I think this macro should keep the newline like when calling the bash function of the same name.
Otherwise we would need a subsequent call to printf to ouptut a `"\n"` and so send the command to the webots console.

omichel: No, it's wrong. If you run webots on with the "--stdout" option, the stdout will come to the Terminal and it will display an empty line with only the controller name. This doesn't happen in Webots (but it should) and is due to this "\n" in ANSI_CLEAR_CONSOLE which is taken into account after the clear screen.

remi-cyberbotics: So a user just issue: `ANSI_CLEAR_CONSOLE();` is his controller code and the screen should get cleaned ?

omichel: Yes.

omichel: But for now, we should forget about these MACROs, the problem is not here, it is in the behavior of the ANSI clear screen which is wrong in the Webots console: `"\x1b[2J\n"` should clear the console and add a new line. Currently that doesn't work in the Webots console, but works in the terminal.

omichel:
@remi-cyberbotics: Can we close this PR?

PR #1497, opened 2020-03-25
Attempt to Fix ConveyorPlatform

### LINKED ISSUE #1483: ConveyorPlatform PROTO object has trouble to carry objects on its belt and is not used anywhere
The ConveyorPlatform PROTO object does not carry elements as it should be, instead they go through the belt and their behavior is undefined. Objects might get carried out after a while or stay stuck on/in the belt.

[Example video](https://cdn.discordapp.com/attachments/687662923044225041/690527086141767680/empty.mp4)

The `animationTexture` field should also be removed from the PROTO file, since it is not used anymore in scenes. Instead it should be automatically computed from the conveyor belt length, so it could be (enventually) modified without altering the animation texture speed.

Following is a world using the element:
```
#VRML_SIM R2020b utf8
WorldInfo {
  info [
    "The MyBot robot, the robot used in the tutorial of Webots"
  ]
  title "MyBot"
}
Viewpoint {
  orientation 0.1157765509215446 0.9791030558957758 0.1671914955739801 4.234990908705313
  position -0.9314227924093998 0.3505959725554629 -0.2764105164625461
}
TexturedBackground {
}
TexturedBackgroundLight {
}
RectangleArena {
  floorSize 2 2
}
ConveyorPlatform {
  translation 0 0.05 0
}
Solid {
  translation -0.361142 0.1 0.140386
  children [
    DEF A_BOX Shape {
      appearance PBRAppearance {
        baseColor 0.333333 1 0
        roughness 1
        metalness 0
      }
      geometry Box {
        size 0.05 0.025 0.05
      }
    }
  ]
  boundingObject USE A_BOX
  physics Physics {
    mass 2
  }
}
```

DavidMansolino:
**Description**
Attempt to fix #1483

@remi-cyberbotics since I am not able to reproduce the issue, can you please check if this fixes the issue?

### DIFF
--- a/projects/objects/factory/conveyors/protos/ConveyorPlatform.proto
+++ b/projects/objects/factory/conveyors/protos/ConveyorPlatform.proto
@@ -58,7 +58,31 @@ PROTO ConveyorPlatform [
             }
           }
         ]
-        boundingObject USE conveyor_belt_BO
+        boundingObject Group {
+          children [
+            USE conveyor_belt_BO
+            DEF belt_1 Transform {
+              translation 0 0 -0.04
+              children [
+                Shape {
+                  geometry Box {
+                    size 0.8 0.047 0.08
+                  }
+                }
+              ]
+            }
+            DEF belt_2 Transform {
+              translation 0 0 0.14
+              children [
+                Shape {
+                  geometry Box {
+                    size 0.8 0.047 0.08
+                  }
+                }
+              ]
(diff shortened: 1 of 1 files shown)


remi-cyberbotics:
Yes it seems it does the trick much better!

DavidMansolino:
Thank you for the confirmation 😃

PR #1501, opened 2020-03-26
Fixed Missing Supervisor Checks.

DavidMansolino:
**Description**
Some Supervisor function were not testing that the robot is actually a Supervisor, which was then causing issue when trying to use the return value (reported by a user on Discord today).

### DIFF
--- a/src/lib/Controller/api/supervisor.c
+++ b/src/lib/Controller/api/supervisor.c
@@ -1372,9 +1372,6 @@ int wb_supervisor_node_get_id(WbNodeRef node) {
 }
 
 static WbNodeRef node_get_from_id(int id) {
-  if (!robot_check_supervisor("node_get_from_id"))
-    return NULL;
-
   robot_mutex_lock_step();
 
   WbNodeRef result = find_node_by_id(id);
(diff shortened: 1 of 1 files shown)


src/lib/Controller/api/supervisor.c
```diff
@@ -1130,10 +1130,16 @@ void wb_supervisor_movie_stop_recording() {
 }
 
 bool wb_supervisor_movie_is_ready() {
+  if (!robot_check_supervisor("wb_supervisor_movie_is_ready"))
```
omichel: I would use the `__FUNCTION__` macro in new code.

src/lib/Controller/api/supervisor.c
```diff
@@ -1345,6 +1372,9 @@ int wb_supervisor_node_get_id(WbNodeRef node) {
 }
 
 static WbNodeRef node_get_from_id(int id) {
+  if (!robot_check_supervisor("node_get_from_id"))
```
omichel: Except here where the function doesn't refer to an API function...

src/lib/Controller/api/supervisor.c
DavidMansolino: As the use of this macro was introduced on develop, this should be done when merging master into develop and not here.

PR #1508, opened 2020-03-30
Enhancement develop console ansi codes

remi-cyberbotics:
This PR transfers the work done on previous PR #1492 

Related documentation link:
https://cyberbotics.com/doc/guide/controller-programming?version=enhancement-rebase-console-ansi-codes


remi-cyberbotics:
This PR is almost finished, I am working on the documentation and I just noticed a different behavior when a Java controller exits:

It seems we are not passing [here](https://github.com/cyberbotics/webots/blob/d2d930bc6b8e4325bd8ffed8c71e58bcf1d0f52c/src/webots/control/WbController.cpp#L604) contrary to C/C++ and Python, what is ongoing ?

Also related, it seems `WBController` crashes when switching from a compiled Java controller to a compiled C, whereas starting only the Java controller works fine.

```
... (C controller code)
INFO: 'console' controller exited successfully.
INFO: MyController: Starting controller: java -classpath "/home/remi/workspace/webots/lib/controller/java/Controller.jar:/home/remi/workspace/webots/projects/samples/howto/controllers/MyController/" -Djava.library.path="/home/remi/workspace/webots/lib/controller/java" MyController 
INFO: MyController: Starting controller: java -classpath "/home/remi/workspace/webots/lib/controller/java/Controller.jar:/home/remi/workspace/webots/projects/samples/howto/controllers/MyController/" -Djava.library.path="/home/remi/workspace/webots/lib/controller/java" MyController 
WARNING: MyController: The process crashed some time after starting successfully.
Picked up _JAVA_OPTIONS: -Dawt.useSystemAAFontSettings=on -Dswing.aatext=true
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f11b902f46c, pid=73546, tid=73551
#
# JRE version: OpenJDK Runtime Environment (13.0.2+8) (build 13.0.2+8)
# Java VM: OpenJDK 64-Bit Server VM (13.0.2+8, mixed mode, tiered, compressed oops, g1 gc, linux-amd64)
# Problematic frame:
# C  [libController.so+0x2646c]
#
# Core dump will be written. Default location: Core dumps may be processed with "/usr/lib/systemd/systemd-coredump %P %u %g %s %t %c %h" (or dumping to /home/remi/workspace/webots/projects/samples/howto/controllers/MyController/core.73546)
#
# An error report file with more information is saved as:
# /home/remi/workspace/webots/projects/samples/howto/controllers/MyController/hs_err_pid73546.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
WARNING: 'MyController' controller crashed.
```

omichel:
Do you get this crash also on the master branch? If so, a new issue should be open to report it as a bug. And, it's not clear when you say `WBController` crashes, do you mean `MyController`?

remi-cyberbotics:
I will try to make it clearer then, and report it if I can reproduce it on master as well.

I am still wondering how can I be sure about a controller has been terminated before outputting some "webots system" text in the console like:
```
make
# updating console.d
# compiling console.c
# linking console
Build finished.
```
Because it seems that when not going through [here](https://github.com/cyberbotics/webots/blob/d2d930bc6b8e4325bd8ffed8c71e58bcf1d0f52c/src/webots/control/WbController.cpp#L604) on a normal exit code that is, the style won't always be reset, which is problematic since many controllers are using loops, and that we might recompile or clean a project while still being in it. How can we avoid this ?

omichel:
We should not care about it. The console is shared by Webots and other controllers, so if a controller changes the foreground color for example, it will affect all the others (output from Webots and other controllers). I would suggest to reset the console only when we clear it from the Tools menu (Ctrl-K) and when we reset or load a world, but not when a controller quits (as this may affect a color set by another controller which doesn't quit).

remi-cyberbotics:
Ok it looks like a better implementation indeed, thank you.

docs/guide/controller-programming.md
```diff
@@ -452,6 +452,47 @@ if (finished) {
...
+As mentioned earlier, printing to `stdout` or `stderr` from a controller will be redirected by default to the Webots console.
+Note however, that the Webots console doesn't support `stdin` input.
+Like most terminals, it supports a few basic [ANSI escape codes](https://en.wikipedia.org/wiki/ANSI_escape_code) for setting text styles and clearing the content of the console:
+  - 3-bit color (foreground and background)
+  - Bold style
+  - Underline style
+  - Clear screen (same as issuing `clear` command in your terminal)
+  - Reset (colors and styles)
+
+There is an example world in webots: samples/howto/console or at `$WEBOTS_HOME/projects/samples/howto/controllers/console` to demonstrate how to use those.
```
remi-cyberbotics, line 466: From here, this part doesn't look good on the documentation, how may we improve it so it won't look like all stuck together ?

https://cyberbotics.com/doc/guide/controller-programming?version=enhancement-rebase-console-ansi-codes#console-output

resources/languages/java/controller.i
```diff
@@ -103,6 +105,12 @@ using namespace std;
 //handling std::string
 %include "std_string.i"
 
+//----------------------------------------------------------------------------------------------
+//  ANSI Support
+//----------------------------------------------------------------------------------------------
+%include <webots/utils/ansi_codes.h>
```
omichel, line 111:
```suggestion
```

resources/languages/python/controller.i
```diff
@@ -36,6 +36,8 @@ if os.name == 'nt' and sys.version_info >= (3, 8):  # we need to explicitly list
 
 %{
 #include <webots/Accelerometer.hpp>
+#include <webots/utils/ansi_codes.h>
```
omichel, line 39:
```suggestion
```

resources/languages/python/controller.i
```diff
@@ -125,6 +127,12 @@ using namespace std;
 %typemap(freearg) const int * {
   free($1);
 }
+//----------------------------------------------------------------------------------------------
+//  ANSI Support
+//----------------------------------------------------------------------------------------------
+%include <webots/utils/ansi_codes.h>
```
omichel, line 133:
```suggestion
```

src/webots/control/WbController.cpp
```diff
@@ -579,18 +573,24 @@ void WbController::appendMessageToBuffer(const QString &message, QString *buffer
...
+    int l = buffer->length();
+    while (buffer->at(count) == '\x1b') {
+      int start = count;
+      count += 2;  // skipping '['
+      while (count < l && buffer->at(count++) < '\x40')
+        ;
+      ansiString += buffer->mid(start, count - start);
+    }
+    // html formatted string ex: "<span style=\"color:#66bb6a;\">Some Text\n</span>"
+    const QString line = ansiString + buffer->mid(count, index + 1 - count);
```
omichel, line 589: Since there is no prefix any more, I am not sure this code is still useful. We can probably remove it.

src/webots/gui/WbConsole.cpp
```diff
@@ -264,37 +264,50 @@ WbConsole::~WbConsole() {
...
 
-void WbConsole::clear() {
+void WbConsole::clear(bool reset) {
   mEditor->clear();
+  if (reset)
+    resetFormat();
+}
+
+void WbConsole::resetFormat() {
+  mForegroundColor = ansiBlack();
```
omichel, line 274: Why not simply clearing it?
```suggestion
  mForegroundColor.clear();
```

src/webots/gui/WbConsole.cpp
```diff
@@ -397,74 +407,106 @@ void WbConsole::handlePossibleAnsiEscapeSequences(const QString &msg, WbLog::Lev
...
+            case '5':
+              mBackgroundColor = ansiMagenta();
+              break;
+            case '6':
+              mBackgroundColor = ansiCyan();
+              break;
+            case '7':
+              mBackgroundColor = ansiWhite();
+              break;
+            case '9':  // 49 - Default background color
```
omichel, line 490: I don't understand this comment. Default color in white isn't it?

src/webots/gui/WbConsole.cpp
```diff
@@ -397,74 +407,106 @@ void WbConsole::handlePossibleAnsiEscapeSequences(const QString &msg, WbLog::Lev
...
+            default:
+              mBackgroundColor = ansiBlack();
```
omichel, line 492: The default case should never happen, hence we should do nothing here.

src/webots/gui/WbConsole.cpp
```diff
@@ -397,74 +407,106 @@ void WbConsole::handlePossibleAnsiEscapeSequences(const QString &msg, WbLog::Lev
...
+              mForegroundColor = ansiCyan();
               break;
             case '7':
-              mColor = ansiWhite();
+              mForegroundColor = ansiWhite();
               break;
             case '9':  // 39 - Default text color
             default:
-              mColor = ansiBlack();
+              mForegroundColor = ansiBlack();
```
omichel, line 463: The default case should never happen hence we should do nothing here.

src/webots/gui/WbConsole.cpp
```diff
@@ -397,74 +407,106 @@ void WbConsole::handlePossibleAnsiEscapeSequences(const QString &msg, WbLog::Lev
...
-      while (msg[i] != 'm')
-        sequence += msg[i++];
-      i++;
-
+      while (msg[i] != '\0' && msg[i] != 'm') {  // 'm' is the final char when supporting only colors
+        sequence += msg[i];
+        if (msg[i] == 'J' && msg[i - 1] == '2')  // "2J" is not terminated by 'm', but is still valid
+          break;
+        i++;
+      }
```
omichel, line 420: It's not a safe way to parse a ANSI sequence. See my contribution in `WbController.cpp` for a safe way (implemented from the Wikipedia documentation, testing for `character < '\x40'` to find the end of the sequence).

docs/guide/controller-programming.md
omichel: See my comment in the review.

resources/languages/java/controller.i
```diff
@@ -20,6 +20,8 @@
 
 %{
 #include <webots/Accelerometer.hpp>
+#include <webots/utils/ansi_codes.h>
```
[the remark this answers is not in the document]
DavidMansolino: If I am not wrong it is, as swig will not generate wrapper for functions in files included in include files.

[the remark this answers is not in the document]
remi-cyberbotics: Yes, as I tried without it - because I found it a bit odd to me - swig still seems to have it included.

src/webots/control/WbController.cpp
remi-cyberbotics: OK the the subrelated methods may be removed as well ?
For example, controllerLogEmitted ?

omichel: If they are not useful any more, yes.

src/webots/gui/WbConsole.cpp
remi-cyberbotics: True, I updated it to some more understandable version, thanks

src/webots/control/WbController.cpp
omichel: This code was not removed... Is it really useful?

omichel, line 589: Here is what I mean:
```suggestion
    const QString line = buffer->mid(0, index + 1);
```

remi-cyberbotics: Oh, I'm sorry I thought you mentioned the part below, I didn't try without that part since I thought that was your commit on this branch, I can still try if you want.

docs/guide/controller-programming.md
```diff
@@ -452,6 +452,53 @@ if (finished) {
...
+Tip: You can `import controller as ctl` and shorten it to `ctl.ANSI_GREEN_FOREGROUND`.
+```python
+import controller as ctl
+
+print(ctl.ANSI_GREEN_FOREGROUND + "Hello World!")
+```
+
+For Java, you may want to create your own style in a constant string for example:
+```java
+import com.cyberbotics.webots.controller.wrapperConstants;
```
omichel, line 489: This is not nice to refer to global wrapper constants...
It's not meaningful with respect to ANSI codes.
I would prefer to see something like: `import com.cyberbotics.webots.controller.AnsiCodes;`
And then use `final String format = AnsiCodes.GREEN_FOREGROUND;`.

remi-cyberbotics: Yes, however how to rename the wrapperConstants package to AnsiCodes regarding the [controller.i](https://github.com/cyberbotics/webots/blob/master/resources/languages/java/controller.i), should I create a new module specifically next to the already existing `%module wrapper` ?

omichel: I don't know yet how to achieve this. I am investigating...

omichel:
I implemented a more object-oriented version of the ANSI codes for C++/Java/Python without the odd `wrapperConstants` import. I also improved the documentation and added usage examples in the `projects/languages` folders. @remi-cyberbotics: can you review it?

docs/guide/controller-programming.md
```diff
@@ -452,6 +452,71 @@ if (finished) {
...
+To demonstrate how to use those, there is an example world and a controller file respectively located in "[WEBOTS\_HOME/projects/samples/howto/worlds/console.wbt](https://github.com/cyberbotics/webots/tree/master/projects/samples/howto/worlds/console.wbt)" and "[WEBOTS\_HOME/projects/samples/howto/controllers/console/console.c](https://github.com/cyberbotics/webots/tree/master/projects/samples/howto/controllers/console/console.c)".
+
+The related C header is located at "[WEBOTS\_HOME/include/controller/c/webots/utils/ansi\_codes.h](https://github.com/cyberbotics/webots/tree/master/include/controller/c/webots/utils/ansi_codes.h)", it contains some useful macros on top of constants, to use it:
+
+%tab-component "language"
+
+%tab "C"
+
+```c
+#include <webots/utils/ansi_codes.h>
```
remi-cyberbotics, line 475: Wouldn't it be more readable with an additional line ?
```suggestion
#include <webots/utils/ansi_codes.h>

```

projects/languages/java/controllers/Slave/Slave.java
```diff
@@ -64,12 +65,11 @@ public void run() {
 
     // perform a simulation steps and leave the loop when the simulation is ended
     while (step(timeStep) != -1) {
-
       // Read sensors, particularly the order of the supervisor
       if (receiver.getQueueLength()>0){
         String message = new String(receiver.getData());
         receiver.nextPacket();
-        System.out.println("I should "+message+"!");
+        System.out.println("I should "+AnsiCodes.RED_FOREGROUND+message+AnsiCodes.RESET+"!");
```
remi-cyberbotics, line 72: Just wondering if it is normal to not add spaces between "+" signs

omichel: No, I just kept the existing coding style for this file (which is not nice I agree), but it's out of the scope of this PR to fix that, as we should fix it in the whole file.

resources/languages/cpp/AnsiCodes.cpp
```diff
@@ -0,0 +1,44 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#define WB_ALLOW_MIXING_C_AND_CPP_API
+#include <webots/utils/ansi_codes.h>
+#include <webots/utils/AnsiCodes.hpp>
+
+using namespace webots;
+
+const std::string AnsiCodes::RESET = ANSI_RESET;
```
remi-cyberbotics, line 21: Can't all of those be inside the AnsiCode.hpp ?

omichel: I don't think so. If you have a better idea, please test it and suggest it here.

remi-cyberbotics: I tried a couple of things without success, It's a pity it's not possible to simply have defines translated to constants automatically by swig...

omichel: Yes, swig has some bad limitations about this.

include/controller/cpp/webots/utils/AnsiCodes.hpp
```diff
@@ -0,0 +1,48 @@
...
+// limitations under the License.
+
+#ifndef ANSICODES_HPP
+#define ANSICODES_HPP
+
+#include <string>
+
+namespace webots {
+  class AnsiCodes {
+  public:
```
DavidMansolino, line 22: Can't we use a namespace instead of a class here?

remi-cyberbotics: I think it is required by swig to be included in Java for example, correct me if I am wrong

DavidMansolino: Yes that's possible indeed. Swig is really picky in this PR, it's a shame to pollute one API because of another one 😞

omichel: I tried hard to use a namespace instead of a class indeed, without any success... 😒. A option would be to make a special file with a class for swig only and a namespace for C++ and Python. Maybe I will try this...

DavidMansolino: Indeed, it seems we have no way to make it work for Java using namespace: http://www.swig.org/Doc1.3/Java.html#namespaces

omichel: Yes, I found that as well (even in the more recent documentation, e.g., [4.0](http://www.swig.org/Doc4.0/Java.html#Java_constants), it is the same)

omichel: OK, so I found a way to have a nice C++ header file (with namespace), but for the Java and Python wrappers, I had to include a dirty hack with a custom `AnsiCodes.hpp` header file that contains both header and implementation code 🙈 so that swig is happy and works nicely.

remi-cyberbotics: I felt every time I tried something similar to make an element working was breaking another one...

DavidMansolino:
👀

PR #1510, opened 2020-03-31
Fix matlab tutorial 6

omichel:
Fixed issues reported in #681.
MATLAB was also unable to handle controller names with a dash like in `e-puck_avoid_collision`, so controller have been renamed and example MATLAB controllers were added along the C examples.


DavidMansolino:
👀

docs/guide/tutorial-4-more-about-controllers.md
```diff
@@ -534,20 +534,17 @@ Reload the world.
...
+>   ps(i) = wb_robot_get_device(convertStringsToChars(ps_names(i)));
+>   wb_distance_sensor_enable(ps(i), TIME_STEP);
 > end
 > ```
 After initialization of the devices, initialize the motors:
 > ```MATLAB
-> left_motor = wb_robot_get_device("left wheel motor");
-> right_motor = wb_robot_get_device("right wheel motor");
+> left_motor = wb_robot_get_device('left wheel motor');
+> right_motor = wb_robot_get_device('right wheel motor');
```
DavidMansolino: Why are you sometimes using simple quote and sometimes using double quotes for string in Matlab?

omichel: Because it's not the same. Double quotes are MATLAB strings while single quotes are char arrays. It turns out our API functions need char arrays, but it's more convenient to make an array of strings and convert them afterwards to char arrays....

omichel:
About the MATLAB programs, I was also hesitating, but I found it very convenient to have it at the same level as the C controller because you can easily use them by cleaning the C controller (make clean), then Webots will use the MATLAB controller. Moreover you can easily compare the MATLAB source code to the C source code. Have these MATLAB controllers here actually don't harm and I found not better solution...

DavidMansolino:
The wbproj are still outdated.

omichel:
Sorry, I forgot this one. Now it should be fine.

DavidMansolino:
References to ``e-puck_go_forward`` and ``e-puck_avoid_collision``should also be updated in the wbproj files.

omichel:
Oops, indeed. Should be fixed now...

PR #1511, opened 2020-03-31
using Ros::fixedNameString() for fixing String in RosDistanceSensor.cpp

klues:
**Description**
using Ros::fixedNameString() in `RosDistanceSensor.cpp` for fixing String instead of own, faulty implementation (`.` not handled).

**Screenshots**
fixes `ros::InvalidNameException` when using `Thymio_II` with ROS: http://prntscr.com/rpzfmc


DavidMansolino:
@klues, feel free to merge this PR.

PR #1512, opened 2020-03-31
Enhancement Variable Truck Size

DavidMansolino:
**Description**
This PR adds the possibility to change the wheelbase of the truck PROTO:
  - [x] Change the wheel location.
  - [x] Change the bounding objects.
  - [x] Change the joint location?
  - [x] Change the mass.
  - [x] Change the Shape.
  - [x] Move LEDs.


DavidMansolino:
The wheelbase and the mass of the Truck are now configurable:
2m
6m

When  changing the wheelbase, the cab is not scaled but only the part between the wheels.

Issue #1434, opened 2020-03-04
Warning OpenGL in console output when toggling "Show contact points"


remi-cyberbotics:
Following the tutorial guide: 
https://www.cyberbotics.com/doc/guide/tutorial-2-modification-of-the-environment?tab-language=c#create-a-ball

It is said at the end of the section that we can toggle View/Optional Rendering/Show Contact Points.
Once the option is checked, *and* the simulation has been run, a looping warning shows up in the linux console ($> ./webots) and says:
```
OpenGL error: GL_INVALID_OPERATION
```

**System**
 - Ubuntu 18.04.4 LTS - 5.3.0-40-generic
 - NVIDIA GeForce RTX 2060 6GB 
 - Webots compiled from source

DavidMansolino:
@lukicdarkoo has the exact same issue.

lukicdarkoo:
The same is happening using `webots_ros2` and Webots compiled from source (branch `develop`, commit 7a1573ca569de6dde14cbce826f0b6c49290d671)

DavidMansolino:
I have been able to get more information about this error:
```
GL_INVALID_OPERATION in glMapBufferRange(length = 0)
GL_INVALID_OPERATION in glUnmapBuffer(buffer is not mapped)
```

This is probably raised internally by the WREN dynamic mesh (WrDynamicMesh).

DavidMansolino:
This seems to be the problematic call:
https://github.com/cyberbotics/webots/blob/bf600fa908cd2035d6b11f5a80f713f6a1349a67/src/wren/DynamicMesh.cpp#L416

PR #1514, opened 2020-04-01
Fix OpenGL error with contact point optionnal rendering

DavidMansolino:
**Description**
When the contact point optional rendering is enabled an OpenGL error was displayed in the console.
This was because an empty mesh was cleared and therefore the `mCoordsDirty` flag was set to true while it should not because the mesh was not changed.

This PR might also slightly improve performance (but it is difficult to measure it) since the dirty flags are not triggering useless updates anymore.

**Related Issues**
This pull-request fixes issue #1434


DavidMansolino:
✔️: travis_unit

PR #1517, opened 2020-04-02
Update changelog-r2020.md

omichel:
A new version of Qt is available on MSYS2.
It generates no new warning when recompiling Webots.


DavidMansolino:
Did you run the 'packaging' test ?

PR #1522, opened 2020-04-06
Fix Add Node Dialog Crash

DavidMansolino:
**Description**
This was causing a crash when using the search bar of the add node dialog if the result was not giving any result in the base node but a result in the USE node.
I am seeing sporadically this bug since a very long time ago but I am happy that I finally managed to reproduce it in a 100% reproducible way.


DavidMansolino:
✔️: travis_unit

PR #1525, opened 2020-04-06
Update System Requirements Doc

remi-cyberbotics:
Supported OS fix/update related documentation.

https://cyberbotics.com/doc/guide/system-requirements?version=fix-doc-supported-os

### DIFF
--- a/docs/guide/system-requirements.md
+++ b/docs/guide/system-requirements.md
@@ -19,9 +19,11 @@ We recommend using a recent version of Linux.
 Webots is provided for Linux 64 (x86-64) systems.
 Since Webots 8.1.0, the Linux 32 (i386) version is no longer provided.
 Webots doesn't run on Ubuntu versions earlier than 12.04.
-- Windows: Webots runs on Windows 10 64-bit, Windows 8.1 64-bit, Windows 8 64-bit and Windows 7 64-bit.
-It is not supported on 32-bit versions of Windows and on older versions including Windows Vista and Windows XP.
-- Mac: Webots runs on macOS 10.13 "High Sierra" and 10.12 "Sierra".
-Webots may work but is not officially supported on earlier versions of macOS.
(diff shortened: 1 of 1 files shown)


docs/guide/system-requirements.md
```diff
@@ -19,9 +19,11 @@ We recommend using a recent version of Linux.
...
 Since Webots 8.1.0, the Linux 32 (i386) version is no longer provided.
 Webots doesn't run on Ubuntu versions earlier than 12.04.
-- Windows: Webots runs on Windows 10 64-bit, Windows 8.1 64-bit, Windows 8 64-bit and Windows 7 64-bit.
-It is not supported on 32-bit versions of Windows and on older versions including Windows Vista and Windows XP.
-- Mac: Webots runs on macOS 10.13 "High Sierra" and 10.12 "Sierra".
-Webots may work but is not officially supported on earlier versions of macOS.
+- Windows: Webots runs on Windows 10 and Windows 8.1.
+It is not supported on older versions (8, 7, Vista or XP) as well as any 32-bit variant.
+- Mac: Webots runs on macOS 10.15 "Catalina" and 10.14 "Mojave".
+Webots may work but is not officially supported on earlier versions of macOS, nor below those mentioned.
```
DavidMansolino, line 25: > on earlier versions of macOS

> nor below those mentioned

Those two sentences means the same no?

remi-cyberbotics: To me "earlier" in this context means the versions which might have been just released

remi-cyberbotics: Or we could just use above and below...

omichel: I would simply drop ", not below those mentioned." which is too verbose to me and doesn't really help.

omichel, line 25: I would make this sentence generic to all OSes:
```suggestion

Webots may work but is not officially supported on earlier versions of the above mentioned operating systems.
```

PR #1535, opened 2020-04-09
enhancement-translate-controller-doc-samples

remi-cyberbotics:
This PR aims to enhance the controller programming documentation. Examples will be translated to different supported languages.

https://cyberbotics.com/doc/guide/controller-programming?version=enhancement-translate-controller-doc-samples


docs/guide/controller-programming.md
```diff
@@ -18,10 +20,67 @@ int main() {
...
+```cpp
+#include <iostream>
+#include <webots/Robot.hpp>
+
+using namespace webots;
+
+int main() {
+  Robot *robot = new Robot();
 
+  while (robot->step(32) != -1)
```
remi-cyberbotics, line 38: Should we use a define for here ? We are talking about those specificities later but the next example is using it nonetheless...

omichel: I would not use a define or variable here as putting the hard-coded value makes it simpler to read (this is a tutorial).

remi-cyberbotics:
I didn't really figure out how we can parse controller arguments with Matlab, this is the only missing example. As far as I know it seems we need to do some wrapping to iterate them but within webots I don't really see how one should do...

I don't think it is necessary to translate snippets since the summary of those is translated for each section.

As for `Controller Termination` section, I don't really see the need to translate anything since we suggest the user to create a function at a specific moment and everything above is pretty explicit I would say.

omichel:
It may not be possible to retrieve the controller arguments from a MATLAB controller... So, it is fine if we don't provide a MATLAB equivalent for that one.

docs/guide/controller-programming.md
```diff
@@ -18,10 +20,67 @@ int main() {
...
+
+%tab "Pyhton"
+```python
+from controller import Robot
+
+robot = Robot()
+
+while robot.step(32) != -1:
+    print("Hello World!")
+    pass
```
omichel, line 55: This is not useful?
```suggestion
```

docs/guide/controller-programming.md
```diff
@@ -51,29 +110,107 @@ When the loop exists, no further communication with Webots is possible and the o
...
 int main() {
   wb_robot_init();
 
   WbDeviceTag ds = wb_robot_get_device("my_distance_sensor");
-  wb_distance_sensor_enable(ds, TIME_STEP);
+  wb_distance_sensor_enable(ds, 32);
 
-  while (wb_robot_step(TIME_STEP) != -1) {
+  while (wb_robot_step(32) != -1) {
     double dist = wb_distance_sensor_get_value(ds);
```
omichel, line 127: It should be `const double value` here as in other languages:
```suggestion
    const double value = wb_distance_sensor_get_value(ds);
```

docs/guide/controller-programming.md
```diff
@@ -51,29 +110,107 @@ When the loop exists, no further communication with Webots is possible and the o
...
 #include <webots/robot.h>
 #include <webots/distance_sensor.h>
 #include <stdio.h>
 
-#define TIME_STEP 32
-
 int main() {
   wb_robot_init();
 
   WbDeviceTag ds = wb_robot_get_device("my_distance_sensor");
```
omichel, line 123: It should be named as in other languages:
```suggestion
  WbDeviceTag sensor = wb_robot_get_device("my_distance_sensor");
```

docs/guide/controller-programming.md
```diff
@@ -51,29 +110,107 @@ When the loop exists, no further communication with Webots is possible and the o
...
+      System.out.println("Sensor value is: " + value);
+    }
+  }
+}
+```
+%tab-end
+
+%tab "MATLAB"
+```MATLAB
+ds = wb_robot_get_device("my_distance_sensor");
```
omichel, line 204:
```suggestion
sensor = wb_robot_get_device('my_distance_sensor');
```

docs/guide/controller-programming.md
```diff
@@ -51,29 +110,107 @@ When the loop exists, no further communication with Webots is possible and the o
...
+wb_distance_sensor_enable(ds, 32);
+
+while wb_robot_step(32) ~= -1
+  dist = wb_distance_sensor_get_value(ds);
```
omichel, line 208:
```suggestion
  value = wb_distance_sensor_get_value(ds);
```

docs/guide/controller-programming.md
```diff
@@ -184,6 +323,103 @@ int main() {
...
+motor = robot.getMotor("my_motor")
+
+F = 2.0   # frequency 2 Hz
+t = 0.0   # elapsed simulation time
+
+while robot.step(TIME_STEP) != -1:
+    pos = sin(t * 2.0 * pi * F)
+    motor.setPosition(pos)
+    t += TIME_STEP / 1000.0
+    pass
```
omichel, line 376:
```suggestion
```

docs/guide/controller-programming.md
```diff
@@ -51,29 +110,107 @@ When the loop exists, no further communication with Webots is possible and the o
...
+
+robot = Robot()
+
+sensor = robot.getDistanceSensor("my_distance_sensor")
+sensor.enable(32)
+
+while robot.step(32) != -1:
+    value = sensor.getValue()
+    print("Sensor value is: ", value)
+    pass
```
omichel, line 175:
```suggestion
```

docs/guide/controller-programming.md
```diff
@@ -184,6 +323,103 @@ int main() {
...
+    final Robot robot = new Robot();
+    final Motor motor = robot.getMotor("my_motor");
+
+    double F = 2.0;   // frequency 2 Hz
+    double t = 0.0;   // elapsed simulation time
+
+    while (robot.step(TIME_STEP) != -1) {
+    double pos = Math.sin(t * 2.0 * Math.PI * F);
+    motor.setPosition(pos);
+    t += (double)TIME_STEP / 1000.0;
```
omichel, line 401:
```suggestion
      double pos = Math.sin(t * 2.0 * Math.PI * F);
      motor.setPosition(pos);
      t += (double)TIME_STEP / 1000.0;
```

docs/guide/controller-programming.md
```diff
@@ -339,10 +577,170 @@ int main() {
...
+    }
+  }
+}
+```
+%tab-end
+
+%tab "MATLAB"
+```MATLAB
+TIME_STEP = 32;
+left_sensor = wb_robot_get_device("left_sensor");
```
omichel, line 716: MATLAB requires single quotes, please check other instances of this mistake:
```suggestion
left_sensor = wb_robot_get_device('left_sensor');
```

docs/guide/controller-programming.md
```diff
@@ -184,6 +323,103 @@ int main() {
...
+%tab "MATLAB"
+```MATLAB
+TIME_STEP = 32;
+motor = wb_robot_get_device("my_motor");
+double F = 2.0;   % frequency 2 Hz
+double t = 0.0;   % elapsed simulation time
+while wb_robot_step(TIME_STEP) ~= -1
+  pos = sin(t * 2.0 * pi * F);
+  wb_motor_set_position(motor, pos);
+    t += TIME_STEP / 1000.0;
```
omichel, line 417:
```suggestion
  t += TIME_STEP / 1000.0;
```

omichel:
About the missing MATLAB example, we should add a simple sentence saying that controller arguments retrieval is not supported in MATLAB, otherwise we don't understand why it's missing.

omichel:
Also, some pieces of source code are not translated in other languages but remain in C.
It looks like this PR was not yet ready for review?
You asked for a review, but it is still marked as draft...

remi-cyberbotics:
I thought it was ready for review, sorry. Unfortunately it seems I used copy paste too much at some time...

I also forgot to remove the draft status beforehand, as for the untranslated examples, like I said in my previous comment, I think it might be irrelevant to translate them all because in the end, the functional code part including all the content of the section is translated. I feel even at the moment some parts are irrelevant already if we are not "thinking in C" for instance:

Just in [Reading Sensors](https://cyberbotics.com/doc/guide/controller-programming?version=enhancement-translate-controller-doc-samples&tab-language=c#reading-sensors), if we switch to any other tab than C, all the following section is almost unrelated since things are done differently in other languages...

as for 'constness' or even `free`ing some illegal code, those are not possible in Python, and so on. How can we avoid inconsistency ?

docs/guide/controller-programming.md
```diff
@@ -51,29 +109,106 @@ When the loop exists, no further communication with Webots is possible and the o
...
 
-  while (wb_robot_step(TIME_STEP) != -1) {
-    double dist = wb_distance_sensor_get_value(ds);
-    printf("sensor value is %f\n", dist);
+  while (wb_robot_step(32) != -1) {
+    const double value = wb_distance_sensor_get_value(sensor);
+    printf("Sensor value is %f\n", value);
   }
 
   wb_robot_cleanup();
```
remi-cyberbotics, line 131: I started to remove those new lines since it is not consistent with the C++ examples, I guess if we want to free anything we should indicate it with a comment

docs/guide/controller-programming.md
```diff
@@ -51,29 +109,105 @@ When the loop exists, no further communication with Webots is possible and the o
...
+  WbDeviceTag sensor = wb_robot_get_device("my_distance_sensor");
+  wb_distance_sensor_enable(sensor, 32);
 
-  while (wb_robot_step(TIME_STEP) != -1) {
-    double dist = wb_distance_sensor_get_value(ds);
-    printf("sensor value is %f\n", dist);
+  while (wb_robot_step(32) != -1) {
+    const double value = wb_distance_sensor_get_value(sensor);
+    printf("Sensor value is %f\n", value);
   }
```
remi-cyberbotics, line 129: I started to remove those new lines since it is not consistent with the C++ examples, I guess if we want to free anything we should indicate it with a comment, no ?

omichel:
I would suggest that if constness raises a problem in some language, we should drop it in these languages only.

docs/guide/controller-programming.md
omichel: I don't understand this comment. Do you mean `const` instead of new lines? Why removing new lines?

remi-cyberbotics: Well, I mean the newline between ` wb_robot_cleanup();` and `return 0;` I don't know why I can't make it more readable in the suggestion...

omichel: I see. This was not a problem for me, but if you believe it's better this way, this is fine for me.

omichel:
We won't be able to avoid language specific things, but at least we should try to be as close as possible between different languages to ease the understanding. For example variable names should be the same.

docs/guide/controller-programming.md
```diff
@@ -98,11 +233,47 @@ The call to the `wb_distance_sensor_get_value` function retrieves the latest val
...
+```cpp
+webots::GPS::const double *getValues() const;
+webots::Accelerometer::const double *getValues() const;
+webots::Gyro::const double *getValues() const;
+```
+%tab-end
+
+%tab "Python"
+```python
+[x, y, z] GPS.getValues()
```
remi-cyberbotics, line 255: I didn't find explicit documentation about how to present methods in class within one line for Python and Java, is my guess right ?

omichel: There is no such convention in Python for the return value. It is usually explained in the description text. Maybe our Python sections should contain the source code of the Python method as usually documented in Python, but also a sentence explaining the return value as it is the case in the official Python documentation.

docs/guide/controller-programming.md
```diff
@@ -98,11 +233,47 @@ The call to the `wb_distance_sensor_get_value` function retrieves the latest val
...
+[x, y, z] Accelerometer.getValues()
+[z, y, z] Gyro.getValues()
+```
```
omichel, line 258: Here is what I mean:

```python
GPS.getValues()
Accelerometer.getValues()
Gyro.getValues()
```
Returns the sensor measurement as an array of 3 floating point numbers: `[x, y, z]`.

docs/guide/controller-programming.md
remi-cyberbotics: So actually it is displayed like you suggested, but it's not really esthetic since the text drops down and disappears as we change from a tab to another. Even if it belongs to the right place, maybe I can suggest to leave it in the tab itself under a comment so it will remain in the "visual tab" and not at the beginning of the following explanation, almost unnoticeable.

Or we can add a new line before the generic explanation text so we really see this line:
`Returns the sensor measurement as an array of 3 floating point numbers: [x, y, z].` is belonging to the tab and not the text...

omichel: Yes you idea of adding it as a comment in the tab seems very good. Let's go ahead with that.

docs/guide/controller-programming.md
```diff
@@ -431,24 +1314,114 @@ In this case the controller should just save the experiment results and quit by
...
+
+%tab "Python"
+```python
+import sys
+
+# terminate only this robot controller
+if finished:
+  saveExperimentsData()
+  del robot
+  sys.exit(0)
```
remi-cyberbotics, line 1411: Is it possible ? If so what about other languages ?

DavidMansolino: It is indeed, but the `del robot` is probably useless as the Python garbage collector will do the job for you 😉

DavidMansolino: About Java, you can probably use something similar with: ``System.exit(0);``.

remi-cyberbotics: So is this sentence relevant for other languages than C ?

> If only one robot controller needs to terminate but the simulation should continue with the other robots, then the terminating robot should call the wb_robot_cleanup function right before quitting:

DavidMansolino: Not really, as this function is called from the deletion of the robot class.
You can probably have a look at this cross-languages samples:
https://cyberbotics.com/doc/guide/tutorial-4-more-about-controllers?tab-language=c++#the-controller-code

docs/guide/controller-programming.md
```diff
@@ -98,11 +243,49 @@ The call to the `wb_distance_sensor_get_value` function retrieves the latest val
...
 const double *wb_accelerometer_get_values(WbDeviceTag tag);
 const double *wb_gyro_get_values(WbDeviceTag tag);
 ```
+%tab-end
+
+%tab "C++"
+```cpp
+webots::GPS::const double *getValues() const;
+webots::Accelerometer::const double *getValues() const;
+webots::Gyro::const double *getValues() const;
```
omichel, line 259: That looks odd... The following looks better, isn't it?
```suggestion
const double *webots::GPS::getValues() const;
const double *webots::Accelerometer::getValues() const;
const double *webots::Gyro::getValues() const;
```

docs/guide/controller-programming.md
```diff
@@ -98,11 +243,49 @@ The call to the `wb_distance_sensor_get_value` function retrieves the latest val
...
+```
+%tab-end
+
+%tab "Python"
+```python
+GPS.getValues()
+Accelerometer.getValues()
+Gyro.getValues()
+
+#Returns the sensor measurement as an array of 3 floating point numbers: `[x, y, z]`.
```
omichel, line 269:
```suggestion
# Returns the sensor measurement as an array of 3 floating point numbers: `[x, y, z]`.
```

docs/guide/controller-programming.md
```diff
@@ -98,11 +243,49 @@ The call to the `wb_distance_sensor_get_value` function retrieves the latest val
...
+```
+%tab-end
+
+%tab "Java"
+```java
+public GPS::public double[] getValues();
+public Accelerometer::public double[] getValues();
+public Gyro::public double[] getValues();
```
omichel, line 277: That looks horrible as well... What about this:
```suggestion
double[] GPS::getValues();
double[] Accelerometer::getValues();
double[] Gyro::getValues();
```

docs/guide/controller-programming.md
```diff
@@ -98,11 +243,49 @@ The call to the `wb_distance_sensor_get_value` function retrieves the latest val
...
+%tab "MATLAB"
+```MATLAB
+[x y z] = wb_gps_get_values(tag)
+[x y z] = wb_accelerometer_get_values(tag)
+[x y z] = wb_gyro_get_values(tag)
+```
+%tab-end
+%end
 
 Each function returns a pointer to three double values.
```
omichel, line 290:
```suggestion
In C and C++, each function returns a pointer to three double values.
```

docs/guide/controller-programming.md
```diff
@@ -113,35 +296,119 @@ Hence accessing the array beyond index 2 is illegal and may crash the controller
...
+z = values[2];
 
 // OK, another way to copy the values
-double a[3] = { pos[0], pos[1], pos[2] };
+double a[3] = { values[0], values[1], values[2] };
 
 // OK, yet another way to copy these values
 double b[3];
-memcpy(b, pos, sizeof(b));
+memcpy(b, values, sizeof(b));
```
omichel, line 318: I believe this is overkill. We should remove it.

docs/guide/controller-programming.md
```diff
@@ -113,35 +296,119 @@ Hence accessing the array beyond index 2 is illegal and may crash the controller
...
+double x, y, z;
+x = values[0];
+y = values[1];
+z = values[2];
+
+// OK, another way to copy the values
+double a[3] = { values[0], values[1], values[2] };
+
+// OK, yet another way to copy these values (C++ 11)
+std::array<double, 3> b = { values[0], values[1], values[2] };
```
omichel, line 339: I would remove that as well.

docs/guide/controller-programming.md
```diff
@@ -113,35 +296,119 @@ Hence accessing the array beyond index 2 is illegal and may crash the controller
...
+
+# OK, to copy the values
+x = values[0]
+y = values[1]
+z = values[2]
+
+# OK, another way to copy the values
+a = []
+a = values
+
```
omichel, line 358: This is pointless and confusing...
```suggestion
```

docs/guide/controller-programming.md
```diff
@@ -113,35 +296,119 @@ Hence accessing the array beyond index 2 is illegal and may crash the controller
...
+# There is no really need to copy differently
```
omichel, line 359:
```suggestion
# There is no really need to copy these values
```

docs/guide/controller-programming.md
```diff
@@ -113,35 +296,119 @@ Hence accessing the array beyond index 2 is illegal and may crash the controller
...
+x = values[0];
+y = values[1];
+z = values[2];
+
+// OK, another way to copy the values
+double []a = values.clone();
+
+// OK, yet another way to copy these values
+double []b;
+System.arraycopy(values, 0, b, 0, values.length);
```
omichel, line 381:
```suggestion
// there is no need to copy these values
```

docs/guide/controller-programming.md
```diff
@@ -113,35 +296,119 @@ Hence accessing the array beyond index 2 is illegal and may crash the controller
...
+% OK, to read the values they should never be explicitly deleted by the controller code
+wb_console_print(sprintf('MY_ROBOT is at position: %g %g %g\n', values(1), values(2), values(3)), WB_STDOUT);
+
+% OK, to copy the values
+x = values[1];
+y = values[2];
+z = values[3];
+
+% OK, another way to copy the values
+a = copy(values);
```
omichel, line 398:
```suggestion
% there is no need to copy these values
```

docs/guide/controller-programming.md
remi-cyberbotics: Ah yes my bad, this is what I wanted to do in the first place besides...

docs/guide/controller-programming.md
remi-cyberbotics: This is why I asked for a suggestion in my very first comment...

docs/guide/controller-programming.md
remi-cyberbotics: You mean all the "extra ways" to copy as well as their equivalent in other languages ?

omichel: Yes. I think it doesn't bring anything helpful, right?

docs/guide/controller-programming.md
```diff
@@ -113,35 +296,119 @@ Hence accessing the array beyond index 2 is illegal and may crash the controller
...
+const double *values = wb_gps_get_values(gps);
 
-pos[0] = 3.5;      // ERROR: assignment of read-only location
-double a = pos[3]; // ERROR: index out of range
-delete [] pos;     // ERROR: illegal free
-free(pos);         // ERROR: illegal free
+values[0] = 3.5;      // ERROR: assignment of read-only location
+double a = values[3]; // ERROR: index out of range
+delete [] values;     // ERROR: illegal free
+free(values);         // ERROR: illegal free
```
omichel, line 411: I would also remove this, as it is not related with Webots, but more C programming...

docs/guide/controller-programming.md
remi-cyberbotics: Yes I never thought so while translating them... A copy is just a copy in the end

docs/guide/controller-programming.md
```diff
@@ -431,24 +1314,93 @@ In this case the controller should just save the experiment results and quit by
...
+if (finished) {
+  saveExperimentData();
+  delete robot;
+  exit(0);
+}
+```
+%tab-end
+%end
 
 Note that the exit status as well as the value returned by the `main` function are ignored by Webots.
```
omichel, line 1405: Exit status and return value of main are the same...
```suggestion
Note that the exit status of a Webots controller ignored by Webots.
```

docs/guide/controller-programming.md
```diff
@@ -431,24 +1314,93 @@ In this case the controller should just save the experiment results and quit by
...
+
+// terminate only this robot controller
+if (finished) {
+  saveExperimentData();
+  delete robot;
+  exit(0);
+}
+```
+%tab-end
+%end
```
omichel, line 1403: I would remove all this and move the `wb_robot_cleanup()`  and `delete robot;` statements in the above code snippet.

remi-cyberbotics:
Thanks for your review @omichel I believe this is really much better looking now, if anyone else has anything to add, please be welcome.

docs/guide/controller-programming.md
```diff
@@ -247,60 +563,265 @@ Instead it stores the data locally and the data is effectively sent when the `wb
...
+```
+%tab-end
+
+%tab "Python"
+```python
+while robot.step(40) != -1:
+  d1 = robot.getDistanceSensor(sensor)
+  d2 = robot.getDistanceSensor(sensor)
+  if (d2 > d1) # WRONG: d2 will always equal d1 here
+    avoidCollision()
```
DavidMansolino, line 639:
```suggestion
    d1 = robot.getDistanceSensor(sensor)
    d2 = robot.getDistanceSensor(sensor)
    if d2 > d1:  # WRONG: d2 will always equal d1 here
        avoidCollision()
```

docs/guide/controller-programming.md
```diff
@@ -247,60 +563,265 @@ Instead it stores the data locally and the data is effectively sent when the `wb
...
+
+%tab "Java"
+```java
+do {
+  readSensors(); // warning: sensor values are undefined on the first iteration
+  actuateMotors();
+} while (robot.step(40) != -1);
+```
+%tab-end
+%end
```
DavidMansolino, line 817: Is it normal that Python and MATLAB are missing here?

docs/guide/controller-programming.md
```diff
@@ -326,29 +847,188 @@ int main() {
...
+    final double left_dist = left_sensor.getValue();
+    final double right_dist = right_sensor.getValue();
+
+    // compute behavior (user functions)
+    final double left = compute_left_speed(left_dist, right_dist);
+    final double right = compute_right_speed(left_dist, right_dist);
+
+    // actuate wheel motors
+    left_motor.setVelocity(left);
+    right_motor.setVelocity(right);
```
DavidMansolino, line 987:
```suggestion
      // read sensors
      final double left_dist = left_sensor.getValue();
      final double right_dist = right_sensor.getValue();

      // compute behavior (user functions)
      final double left = compute_left_speed(left_dist, right_dist);
      final double right = compute_right_speed(left_dist, right_dist);

      // actuate wheel motors
      left_motor.setVelocity(left);
      right_motor.setVelocity(right);
```

docs/guide/controller-programming.md
remi-cyberbotics: Yes, it is stated above that only languages natively supporting the `do...while` statement are demonstrated, this is also completed by a sentence afterwards which emphasis on the need to call `wb_robot_step()` function :)

docs/guide/controller-programming.md
```diff
@@ -247,60 +563,265 @@ Instead it stores the data locally and the data is effectively sent when the `wb
...
+```
+%tab-end
+
+%tab "Python"
+```python
+while robot.step(40) != -1:
+    d1 = robot.getDistanceSensor(sensor)
+    d2 = robot.getDistanceSensor(sensor)
+    if d2 > d1: # WRONG: d2 will always equal d1 here
+      avoidCollision()
```
DavidMansolino, line 639:
```suggestion
        avoidCollision()
```

docs/guide/controller-programming.md
```diff
@@ -247,60 +563,265 @@ Instead it stores the data locally and the data is effectively sent when the `wb
...
 }
 ```
+%tab-end
+
+%tab "Python"
+```python
+while robot.step(40) != -1:
+    d1 = robot.getDistanceSensor(sensor)
+    if robot.step(40) == -1:
+      break
```
DavidMansolino, line 702:
```suggestion
        break
```

docs/guide/controller-programming.md
```diff
@@ -247,60 +563,265 @@ Instead it stores the data locally and the data is effectively sent when the `wb
...
+    d2 = robot.getDistanceSensor(sensor)
+    if d2 > d1:
+      avoidCollision()
```
DavidMansolino, line 705:
```suggestion
        avoidCollision()
```

remi-cyberbotics:
No @DavidMansolino It's me, I just passed again on those earlier and didn't think about hitting `tab` twice (I have to configure my IDE for python...), sorry for that...

Issue #1538, opened 2020-04-14
Problem using group as boundingobject


DengYuelin:
I have a solid with a group as its children, and the group as the boundingobject. The group has contained two transforms, the first has a children shape DEF A, the second children use DEF A, (two identical shape, cylinder, at different place). When I change the radius of A, both cylinders change shape, which is correct. However, only one side of the boundingobject(white border) changed, resulting in different boundingobject on two sides. See the picture for detail.
![Snipaste_2020-04-13_23-38-41](https://user-images.githubusercontent.com/32665692/79193665-f502c580-7ddf-11ea-8b6a-424155dec195.png)

If I save and reload, the two sides will match. However, I intended to change the radius using a program during simulation, so S&L isn't an option.

**System**
I tried this on both systems:
1, Ubuntu 16.04 with webots 2020a-rev1;
2, Windows 10 with webots 2019b-rev1;

**Additional context**
The world I'm using.
[Double_peg_hole.zip](https://github.com/cyberbotics/webots/files/4473560/Double_peg_hole.zip)

stefaniapedrazzi:
Thank you for reporting this bug.
I'm inspecting it and I will fix it as soon as possible.

Meanwhile there are some workarounds.
For example if you use this structure, then resizing the peg works correctly:

```
Solid { 
  childen [
    DEF PEG1 Transform {
      children [
        DEF PEGSHAPE Shape {
          ...
        }
      ]
    }
    DEF PEG2 Transform {
      translation 0.04 0 0
      children [
        USE PEGSHAPE
      ]
    }
  ]
  boundingBox Group {
    children [
      USE PEG1
      Transform {
        translation 0.04 0 0
        children [
          USE PEGSHAPE
        ]
      }
    ]
  }
}
```

stefaniapedrazzi:
This issue is fixed in #1544.

PR #1544, opened 2020-04-15
Fix nested DEF/USE nodes

stefaniapedrazzi:
**Description**
If USE node B has a DEF ancestor A, then USE node A is not updated when DEF node B changes.
Sample world:
```
Solid {
  children [
    Transform {
      children [
        DEF B Shape {
          appearance Appearance {
            material Material {
            }
          }
          geometry Box {
            size 0.1 0.1 0.1
        }
      ]
    }
    DEF A Transform {
      translation 0.4 0 0
      children [
        USE B
      ]
    }
  ]
  boundingObject USE A
}
```

**Related Issues**
This pull-request fixes issue #1538.

**Tasks**
* [x] Update DEF ancestor of USE nodes on field change
* [x] Check if USE node is up-to-date before forwarding the field changes
* [x] Add test

Second task fixes a crash when inserting a Material node into the following DEF A Appearance:
```
DEF G Group {
  children [
    DEF S Shape {
      appearance DEF A Appearance {
      }
    }
    USE S
  ]
}
USE G
```


stefaniapedrazzi:
The original issue described in #1538 is fixed, i.e. when modifying an existing node.
But there are also issues when inserting/deleting nodes.

stefaniapedrazzi:
:heavy_check_mark: travis_unit

DavidMansolino:
:eyes:

tests/api/worlds/nested_def_nodes.wbt
```diff
@@ -0,0 +1,129 @@
...
+              }
+            }
+          ]
+        }
+        DEF T2 Transform {
+          translation 0.4 0 0
+          children [
+            USE S1
+          ]
+        }
```
DavidMansolino, line 44: Adding a `USE T2` here causes an issue:
  - changing the `B1` size works fine.
  - changing the `T2` translation works fine for the shape but not for the bounding object (a save and revert is still required for the update to be effective).

stefaniapedrazzi:
Thank you for testing. I will check these cases too.

stefaniapedrazzi:
The cause of the issue with the Transform.translation change seems the same as the one described in #1569: the field redirection is not correct for USE nodes and it doesn't trigger the update of the USE nodes connected to the upper DEF.
I will review the node creation and field redirection to try to properly fix also these cases.

tests/api/worlds/nested_def_nodes.wbt
stefaniapedrazzi: I added a test for this case below (`USE_CASE_4`): test the update of a USE node containing another USE node.

stefaniapedrazzi:
:heavy_check_mark: travis_unit

Issues reported just above by @DavidMansolino and #1569 are now fixed.

I checked if the issues come from the node instantiation, but there is no field redirection specific to DEF/USE nodes in the node constructor. So it is the `WbNode::notifyFieldChanged` function that have to take care of it.

DavidMansolino:
👀

PR #1461, opened 2020-03-16
Spot model from Boston Dynamics

michou214:
**Description**
Add the Spot robot from Boston Dynamics.

**Tasks**
- [x] Find a model with good licence to reuse it
- [x] Verify size and scale if need
- [x] Blender parts decomposition
- [x] Parts extraction, labelling and remove useless ones
- [x] Meshes reduction and correction
- [x] UV mapping and texture application
- [x] Boundary object correction and optimization
- [x] Rotational axis orientation verification
- [x] Camera orientation verification
- [x] Add sensors
- [x] Weight and center of mass calibration
- [x] PROTO file
  - [x] Header description
  - [x] Clean number x.xxxx and replace xxxe-xxx by 0)
  - [x] Remove transitionStep, rotationStep, rotation with no rotation
  - [x] Texture folder to split in each PROTO + change url 
- [x] World file
  - [x] Header description
  - [x] Arrange items
  - [x] Clean number x.xxxx and replace xxxe-xxx by 0) 
- [ ] Controller creation and behaviour test  (**In progress**)
   - [x] Add the controller in the global Makefile (webots/projects/robots/Makefile)
- [ ] Documentation generation (**In progress**)
  - [x] Update docs/guide/robots.md page
  - [x] Create  docs/guide/spot.md page with all information 
  - [x] Update docs/guide/menu.md page (+ add the new .md files reference)
  - [x] Add modifications in the Changelog (docs/reference/changelog-r2020.md)
  - [ ] Icon, model, thumbnails and world images (verify orientation and shadows)
  - [ ] WEB component generation (verify orientation)
  - [x] Make a Splash image.jpg (ressources/images/splash_images + respect the format JPG in 580x580px or 1080x1080px)
- [ ] Docs test passed
- [ ] Sources test passed
- [ ] Review on Github

**Documentation**
* File: [doc/guide/menu.md](https://www.cyberbotics.com/doc/guide/menu?version=michou214:feature-add-SpotMini-robot)
* File: [doc/reference/changelog-r2020.md](https://cyberbotics.com/doc/reference/changelog-r2020?version=michou214:feature-add-SpotMini-robot)

**Spot Mini from Boston Dynamics**
Changes can be viewed:
* File: [doc/guide/robots.md](https://www.cyberbotics.com/doc/guide/robots?version=michou214:feature-add-SpotMini-robot)
* File: [doc/guide/spot.md](https://www.cyberbotics.com/doc/guide/spot?version=michou214:feature-add-SpotMini-robot)

[...]

PR #1553, opened 2020-04-16
Feature add spot mini robot

DavidMansolino:
**Description**
This PR replaces #1461, by migrating the source branch on this repo instead of a fork.

**Remaining Tasks**
- [x] Update with master.
- [x] _Controller_:
  - [x]  Finish the `go_forward()` function
  - [x]  Clang-format.
- [x] _Documentation:_ 
  - [x] Generate icon and model
  - [x] thumbnails images
  - [x] WEB component
- [x] _New Appearances:_ 
  - [x] https://github.com/cyberbotics/webots/tree/feature-add-SpotMini-robot/projects/robots/boston_dynamics/spot/protos/textures/holes_perforated_steel duplicated with https://github.com/cyberbotics/webots/blob/feature-add-SpotMini-robot/projects/samples/environments/factory/worlds/textures/metal_grid_base_color.png
- [x] _Tests:_ Fix Travis errors
- [x] _Improvement:_ Find a solution to have a real "piston" behaviour in the elbow or remove the `DEF TENDON Solid` node.


DavidMansolino:
I have made some cleanups to this PR, the two main points remaining to fix are:
- [ ] The walking algorithm is not yet working.
- [ ] The robot is not well centred (centre of the robot completely outside of the robot).

DavidMansolino:
My latest commit fixes most of the solid and joint origin but unfortunately it breaks completely the controller, because the joints have moved a lot (they were not inside the robot at all).

DavidMansolino:
I have been able to simplify the controller and make it way more robust but there is still something not realistic at all with the behaviour of this robot, its motions does not make sense at all (it looks like the robot is in water or something similar).
For now, I suspect two possible causes:
  - The bounding objects may be too complex.
  - The mass distribution may be wrong.

omichel:
Can you post a short movie about it?

DavidMansolino:
Here is a video showing the problem:
https://www.youtube.com/watch?v=FdhAOyV8VpY&feature=youtu.be
Their is no spring or damping effect, front and rear legs are doing the exact same motion.

DavidMansolino:
I checked the masses distribution centre of masses and inertia and they are correct.
I also tried to simplify a lot the key bounding objects.
But it doesn't impact the behaviour.

I am now suspecting the legs definition, it seems in the leg too there is coordinate issue (the origin of the leg is completely outside of the leg itself (by more than 1.5 meters), even if the bounding objects, joints and physics are defined accordingly I suspect it may cause issues to ODE.

omichel:
To me, it looks like the mass is too lightweight or the strength of the motors is too high...

DavidMansolino:
The mass is the one of the doc (with is more than 25kg so not so lightweight), but good points about the motors, I will check this.

DavidMansolino:
About the max torque of the motors, it is set to 50Nm per motor, which considering 4 legs and a total mass of 25-30kg seems to make sense.

DavidMansolino:
Correcting the legs is a real nightmare, it is extremely time consuming and error prone 😞

DavidMansolino:
I have fixed all the legs (and sub-parts) origin and it is now working like a charm (realistic motion and much more stable):
https://youtu.be/XI4R89o7yRw

DavidMansolino:
✔️: travis_release (https://travis-ci.com/github/cyberbotics/webots/builds/161386660)

omichel:
:eyes:

docs/guide/spot.md
```diff
@@ -0,0 +1,47 @@
...
+
+Derived from [Robot](../reference/robot.md).
+
+```
+Spot {
+  SFVec3f     translation      0 -1.028 0.46
+  SFRotation  rotation         0 1 0 0
+  SFString    name             "Spot"
+  SFString    model            "Boston Dynamics - Spot"
+  SFString    controller       "moving_demo"
```
omichel, line 24:
```suggestion
  SFString    controller       "spot_moving_demo"
```

projects/robots/boston_dynamics/spot/controllers/spot_moving_demo/spot_moving_demo.c
```diff
@@ -0,0 +1,172 @@
...
+#define NUMBER_OF_JOINTS 12
+#define NUMBER_OF_CAMERAS 5
+
+// Initialize the robot's information
+static WbDeviceTag motors[NUMBER_OF_JOINTS];
+static const char *motor_names[NUMBER_OF_JOINTS] = {
+  "front left leg shoulder abduction motor",  "front left leg shoulder rotation motor",  "front left leg elbow motor",
+  "front right leg shoulder abduction motor", "front right leg shoulder rotation motor", "front right leg elbow motor",
+  "rear left leg shoulder abduction motor",   "rear left leg shoulder rotation motor",   "rear left leg elbow motor",
+  "rear right leg shoulder abduction motor",  "rear right leg shoulder rotation motor",  "rear right leg elbow motor"};
```
omichel, line 41: I am wondering whether we should remove the "leg" keyword from the names here as it seems to be unnecessary to combine with shoulder or elbow, etc.

projects/robots/boston_dynamics/spot/controllers/spot_moving_demo/spot_moving_demo.c
```diff
@@ -0,0 +1,172 @@
...
+static WbDeviceTag cameras[NUMBER_OF_CAMERAS];
+static const char *camera_names[NUMBER_OF_CAMERAS] = {"left head camera", "right head camera", "left flank camera",
+                                                      "right flank camera", "rear camera"};
+static WbDeviceTag leds[NUMBER_OF_LEDS];
+static const char *led_names[NUMBER_OF_LEDS] = {"left top led",          "left middle up led", "left middle down led",
+                                                "left bottom led",       "right top led",      "right middle up led",
+                                                "right middle down led", "right bottom led"};
+
+void step() {
```
omichel, line 50:
```suggestion
static void step() {
```

docs/guide/spot.md
```diff
@@ -0,0 +1,47 @@
...
+- `extension`: Extends the robot with new nodes.
+
+### Samples
+
+You will find the following sample in the folder: "[$WEBOTS\_HOME/projects/robots/boston\_dynamics/spot/worlds](https://github.com/cyberbotics/webots/tree/master/projects/robots/boston_dynamics/spot/worlds)"
+
+### [spot.wbt](https://github.com/cyberbotics/webots/tree/master/projects/robots/boston_dynamics/spot/worlds/spot.wbt)
+
+![spot.wbt.png](images/robots/spot/spot.wbt.thumbnail.jpg) This simulation shows a Spot robot in a simple environment.
+The robot is moving its right arm.
```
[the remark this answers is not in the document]
omichel: I would say right leg...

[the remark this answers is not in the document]
DavidMansolino: Indeed, I didn't updated this part after simplifying the controller.

DavidMansolino:
> Shall we create new PBR appearances for the plastic padded and plastic rough instead of keeping them inside the spot proto folder?

Actually, for now, my rule for creating new appearance is to create one as soon as the texture is used in more than one world/proto, but if you think this textures are nice to showcase we can indeed create new appearances.

omichel:
Yes, I believe plastic are very common appearances that could be useful in many new robots, so I would at least add the plastic rough one.

omichel:
👀

PR #1559, opened 2020-04-17
Fix controllers clear console

### LINKED ISSUE #1546: Loading Warning/Errors not Visible Anymore
**Describe the Bug**
When some of the controllers start they clear the console. This is very dangerous as it will hide all the error messages displayed at the world load.
For example, in my case I had missing textures warnings in the moose.wbt world (due to my environment), but when I started the world it was in realtime mode and therefore the controller started immediately and removed these warnings even before I could see them.
Note also that the message:
```
INFO: XXX: Starting controller
```
Is not visible too.

I agree it is nice to have a clean console when starting the controller, but hiding possible error message may be very confusing for the user that may not understand why the simulation is not behaving as expected.

ghost:
This PR closes #1546 

The more I think about this, the more I believe we should have two buffers:

- One for the Webots "system" messages such as errors for textures loading, or warning for wrong world header ...

- One for the controller itself, located _within_ the system buffer.

This way, an `ANSI_CLEAR_CONSOLE()` command issued from a controller would be allowed to only clear the "controller console", clearing its frame but not the whole system buffer.
So we should just ensure to set the "scroll view area" to the top of the controller frame and only displaying this one, but as soon as the controller exits, if one scrolls up, we should see the previously displayed webots system messages.

I don't know if it would be feasible but it's worth to ask I guess.


DavidMansolino:
That's indeed not a bad idea!
Taking it at the extreme, we can even think about having one tab per controller in the Webots console (+one for the system (i.e. Webots error message)), ideally we should even be able to display more than one tab at the same time (but I am maybe extrapolating a bit too much).
We can maybe talk about this at our next weekly meeting.

remi-cyberbotics:
Thanks @DavidMansolino, why not indeed. Looking forward to know what we'll decide!

remi-cyberbotics:
So after discussion, we agreed to this:

We should implement optional tabs in the Webots console, a bit like in a web browser. By default there would be only one, conserving the actual behavior of the console.

When openning a tab, we could see one or more controllers using this tab, so in this context we will have one tab for the Webots system messages, and one tab for controllers. If we want to have several controllers using the same tab we could use checkboxes at the top of the tab, or any drop-down list with checkable elements.

This way, we could open several tabs for different kind of robot controllers for example, one for ground type and one for aerial type, allowing us to split the system messages from the controller ones at the same time.

Once a layout has been defined, it could be saved to the .wbproj file.

omichel:
There are some other cases that could be fixed (see below). However, once #1595 is implemented, we probably want to keep these `ANSI_CLEAR_CONSOLE()`... So, I will wait until #1595 is merged to complete this PR.
```
samples/devices/controllers/battery/battery.c:    ANSI_CLEAR_CONSOLE();
samples/devices/controllers/gyro/gyro.c:    ANSI_CLEAR_CONSOLE();
samples/devices/controllers/motor/motor.c:    ANSI_CLEAR_CONSOLE();
samples/devices/controllers/position_sensor/position_sensor.c:    ANSI_CLEAR_CONSOLE();
samples/devices/controllers/radar/radar.c:      ANSI_CLEAR_CONSOLE();
samples/devices/controllers/receiver_noise/receiver_noise.c:        ANSI_CLEAR_CONSOLE();
samples/devices/controllers/spherical_camera/spherical_camera.c:    ANSI_CLEAR_CONSOLE();
samples/howto/controllers/console/console.c:  ANSI_CLEAR_CONSOLE();
```

PR #1563, opened 2020-04-17
fix-crash-elevator-grid

### LINKED ISSUE #1556: Webots crashed when creating the ElevatorGrid node
Hello, I have a problem with Webots. Every time when I create the ElevatorGrid node within the solid node, the Webots will breakdown, and I cannot find the solution. I'm using the version of webots-R2020a-rev2. 

1. open an existing world or create a new world.
2. Then create a new solid node.
3. create the ElevatorGrid node in the boundingObject within the solid node.
4. The webots will break down.

**Expected behavior**
I think the ElevatorGrid node should be successfully created and used to create uneven ground. But in fact the webots will breakdown.

**Screenshots**
image
image

**System**
 - Graphics Card: NVIDIA GeForce GTX 1050 Ti, Intel(R) UHD Graphics 630

remi-cyberbotics:
This PR will close #1556


remi-cyberbotics:
Not dereferencing null will now give those messages in webots console:

```
WARNING: Solid > ElevationGrid: Invalid 'xDimension' for use in boundingObject.
WARNING: Solid > ElevationGrid: Invalid 'zDimension' for use in boundingObject.
WARNING: Solid > ElevationGrid: Cannot create the associated physics object.
```

How could it be improved so the grid will be mapped onto the `Solid` node ?

DavidMansolino:
You have the answer of this question in this function: https://github.com/cyberbotics/webots/blob/master/src/webots/nodes/WbElevationGrid.cpp#L447-L467

But these constraints should be added to the documentation of this node: https://www.cyberbotics.com/doc/reference/elevationgrid

DavidMansolino:
And the warnings should be more verbose.

remi-cyberbotics:
So when starting from the ground up, shouldn't we initialize `xDimension` and `zDimension` to two to avoid those warnings when creating our object ?

src/webots/nodes/WbElevationGrid.cpp
```diff
@@ -451,10 +457,10 @@ bool WbElevationGrid::isSuitableForInsertionInBoundingObject(bool warning) const
 
   if (warning) {
     if (mXDimension->value() < 2)
-      warn(tr("Invalid 'xDimension' for use in boundingObject."));
+      warn(tr("Invalid 'xDimension' (should be greater than 2) for use in boundingObject."));
```
omichel, line 460:
```suggestion
      warn(tr("Invalid 'xDimension' (should be greater than 1) for use in boundingObject."));
```

src/webots/nodes/WbElevationGrid.cpp
```diff
@@ -451,10 +457,10 @@ bool WbElevationGrid::isSuitableForInsertionInBoundingObject(bool warning) const
...
 
     if (mZDimension->value() < 2)
-      warn(tr("Invalid 'zDimension' for use in boundingObject."));
+      warn(tr("Invalid 'zDimension' (should be greater than 2) for use in boundingObject."));
```
omichel, line 463:
```suggestion
      warn(tr("Invalid 'zDimension' (should be greater than 1) for use in boundingObject."));
```

src/webots/nodes/WbElevationGrid.cpp
```diff
@@ -57,6 +57,12 @@ void WbElevationGrid::init() {
 
 WbElevationGrid::WbElevationGrid(WbTokenizer *tokenizer) : WbGeometry("ElevationGrid", tokenizer) {
   init();
+
+  // avoid warning
```
omichel, line 61: I am afraid this is not the right way to fix the problem because we don't want to silently override the value set by the user...

omichel:
My proposed patch does a similar job as in the `Box` geometry case: if the `ElevationGrid` is created from the scene tree (GUI), it is initialized by default to a dimension of 2x2. However, if an `ElevationGrid` with a dimension of 0x0 is loaded from a file, these values are not changed. I believe that changing silently the values read from a user world or proto file is not a good idea...

src/webots/nodes/WbElevationGrid.cpp
remi-cyberbotics: True but I was not seeing why an user would like to create such an element and keep those field values to 0.

omichel: I agree there is no point having these values set to 0, but overriding silently values set in a user file (world or proto) is not a good practice in general.

Issue #1520, opened 2020-04-04
JPEG textures not displayed, generate warning about "Unsupported image format"


telramund:
**Describe the Bug**

This only happens on Intel HD Graphics 630 GPU. Another system with NVIDIA GeForce 1050Ti does not have this problem. JPEG textures are not displayed, the following warnings are generated:

WARNING: Texture image size of '/home/smbdy/webots/projects/appearances/protos/textures/chequered_parquetry/chequered_parquetry_base_color.jpg' is not a power of two: rescaling it from -1x-1 to 0x0.
WARNING: Floor (PROTO) > ChequeredParquetry (PROTO) > ImageTexture: Cannot load texture '/home/smbdy/webots/projects/appearances/protos/textures/chequered_parquetry/chequered_parquetry_base_color.jpg': Unsupported image format.
WARNING: Texture image size of '/home/smbdy/webots/projects/appearances/protos/textures/chequered_parquetry/chequered_parquetry_roughness.jpg' is not a power of two: rescaling it from -1x-1 to 0x0.
WARNING: Floor (PROTO) > ChequeredParquetry (PROTO) > ImageTexture: Cannot load texture '/home/smbdy/webots/projects/appearances/protos/textures/chequered_parquetry/chequered_parquetry_roughness.jpg': Unsupported image format.
WARNING: Texture image size of '/home/smbdy/webots/projects/appearances/protos/textures/chequered_parquetry/chequered_parquetry_normal.jpg' is not a power of two: rescaling it from -1x-1 to 0x0.
WARNING: Floor (PROTO) > ChequeredParquetry (PROTO) > ImageTexture: Cannot load texture '/home/smbdy/webots/projects/appearances/protos/textures/chequered_parquetry/chequered_parquetry_normal.jpg': Unsupported image format.
WARNING: Texture image size of '/home/smbdy/webots/projects/appearances/protos/textures/chequered_parquetry/chequered_parquetry_occlusion.jpg' is not a power of two: rescaling it from -1x-1 to 0x0.
WARNING: Floor (PROTO) > ChequeredParquetry (PROTO) > ImageTexture: Cannot load texture '/home/smbdy/webots/projects/appearances/protos/textures/chequered_parquetry/chequered_parquetry_occlusion.jpg': Unsupported image format.

**Steps to Reproduce**

1. Start Webots
2. Open Hexapod sample world (or any other)

**Expected behavior**

Textures should be displayed and no warnings generated

**System**
[...]

PR #1567, opened 2020-04-20
Add libpjeg to tarball package

DavidMansolino:
**Description**
Follow-up of https://github.com/cyberbotics/webots/issues/1520#issuecomment-615372716

Since this library is very small (300ko) and not easy to install on all the distributions of linux, I think we can afford to add it to the tarball package.


DavidMansolino:
:heavy_check_mark: `travis_release`

Issue #1566, opened 2020-04-17
Security issue: Unsafe temporary lib directory in linux launcher


MaxGaukler:
If I read the script correctly, the linux launcher is vulnerable to a library path injection:

https://github.com/cyberbotics/webots/blob/69202baee3c0974f69f0e72d3412d19ed69feaad/src/webots/launcher/webots-linux.sh#L64-L72

The temp dir is silently accepted if it already exists, even if it was created by another user. Therefore one user can escalate permissions to any other user who later launches webots on the same system.

Sketch of exploit:
- create /tmp/webots-lib-dir{1-9999999}
- place some .so file there that is used by webots and would otherwise be loaded from /usr/lib.
- modify that .so file to run exploit code
- make directory world-writable so that webots launcher doesn't fail when writing to path

Sketch of solution:
```
# safely create a temporary directory.
# Note that the following two lines cannot be merged into one because `export` would "hide" the return status of `mktemp`.
WEBOTS_TMP_PATH="$(mktemp -d $TMPDIR/webots-XXXXXX)" || exit 1
export WEBOTS_TMP_PATH
TMP_LIB_DIR="$WEBOTS_TMP_PATH/lib"
mkdir -p $TMP_LIB_DIR
```
At first glance, this sketch should work, however, `snap` and other edge cases should be double-checked.

omichel:
Thank you for this contribution. Please go ahead and open a pull request for that. We will be happy to review it.

MaxGaukler:
There are a few open issues that my PR doesn't yet fix:
- projects/samples/contests/ratslife/controllers/contest_manager/texture_generator.c uses unsafe TMPDIR; should switch to mktdemp(), QTemporaryDir or similar.
- MacOS should be checked, I'd guess the launcher and WbStandardPaths::webotsTmpPath()  are vulnerable
- WbStandardPaths::webotsTmpPath()  has an unsafe fallback for Linux
- WbControledWorld has the possibility of collisions, it uses $TMPDIR/webots_$PID_$random as filename without checking existence (or if it does, there's a TOCTOU race condition). I'm not sure about the semantics of socket files, if there's anything evil I can do if someone else creates a socket with predictable path. Maybe use it for an overwriting attack by placing a symlink?
- Maybe drop all the details in src/lib/Controller/api/system.c and just use QTemporaryDir (with autoRemove=false) or similar?

omichel:
Thank for this in depth analysis which looks very relevant.

[...]

PR #1571, opened 2020-04-20
Use safe temporary directory on Linux

MaxGaukler:
**Description**
The Linux launcher created a predictable temporary directory in an unsafe fashion.

**Related Issues**
This pull-request solves the most critical part of #1566

**Tasks**
Check all non-standard use-cases on Linux
  - [x] snap package
  - [x] external controllers
  - [x] anything else that uses TMPDIR, WEBOTS_TEMP_PATH etc.

### DIFF
--- a/src/lib/Controller/api/system.c
+++ b/src/lib/Controller/api/system.c
@@ -139,11 +139,13 @@ const char *wbu_system_webots_tmp_path() {
   const size_t l = strlen(tmp);
   const char *WEBOTS_PID = getenv("WEBOTS_PID");
   int webots_pid = 0;
+  char random_part[16];
+  random_part[0] = '\0';
   if (WEBOTS_PID && strlen(WEBOTS_PID) > 0)
     sscanf(WEBOTS_PID, "%d", &webots_pid);
-  const size_t path_buffer_size = l + 32;  // enough room to hold tmp + "/webots-XXXX...XXX" (pid_t maximum string length)
+  const size_t path_buffer_size = l + 64;
   char *path_buffer = malloc(path_buffer_size);
-  if (webots_pid == 0) {  // get the webots pid from the most recent "webots-XXX" folder
+  if (webots_pid == 0) {  // get the webots pid from the most recent "webots-<PID>-XXXXXX" folder
     DIR *dir;
     dir = opendir(tmp);
     if (dir) {
@@ -159,8 +161,29 @@ const char *wbu_system_webots_tmp_path() {
             continue;
           if (s.st_mtime < most_recent)
             continue;
-          sscanf(entry->d_name, "webots-%d", &webots_pid);
+          sscanf(entry->d_name, "webots-%d%s", &webots_pid, random_part);
           most_recent = s.st_mtime;
+        }
+      }
+      closedir(dir);
+    }
+  } else {  // get the random part
+    DIR *dir;
+    dir = opendir(tmp);
+    if (dir) {
+      struct dirent *entry;
+      char folder_start[32];
+      sprintf(folder_start, "webots-%d", webots_pid);
+      while ((entry = readdir(dir))) {
+        if (strncmp(entry->d_name, folder_start, strlen(folder_start)) == 0) {
+          struct stat s;
+          snprintf(path_buffer, path_buffer_size, "%s/%s", tmp, entry->d_name);
+          if (stat(path_buffer, &s) < 0)
+            continue;
+          if (!S_ISDIR(s.st_mode))
+            continue;
+          sscanf(entry->d_name, "webots-%d%s", &webots_pid, random_part);
+          break;
         }
       }
       closedir(dir);
@@ -170,7 +193,7 @@ const char *wbu_system_webots_tmp_path() {
     free(path_buffer);
     path_buffer = NULL;
   } else
-    snprintf(path_buffer, path_buffer_size, "%s/webots-%d", tmp, webots_pid);
+    snprintf(path_buffer, path_buffer_size, "%s/webots-%d%s", tmp, webots_pid, random_part);
   WEBOTS_TMP_PATH = path_buffer;
   return WEBOTS_TMP_PATH;
 }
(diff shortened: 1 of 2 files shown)


MaxGaukler:
The remaining tasks should be reviewed by someone with more experience of the internal details and of how to build and package webots; I don't have the necessary knowledge to be sure that it works. The standard case (deb on Ubuntu, just replace the launcher script) works for me.

omichel:
OK, we will review that and check the corresponding boxes.

DavidMansolino:
I checked all the possible issues raised by this change and it seems only this function needs update (because it relies on the fact that the temporary folder include the Webots PID):

https://github.com/cyberbotics/webots/blob/master/src/lib/Controller/api/system.c#L131

MaxGaukler:
Yes, that might be an issue for external controllers because they don't 
receive the WEBOTS_TMP_PATH environment variable.

Another solution on Linux would be to use /run/user/$UID/, which is a 
private temporary directory. However, unlike /tmp, that is a RAM-disk, 
so it could run out of space in extreme cases (huge generated 
textures/PROTOs).

omichel:
This `WEBOTS_PID` is used only in the case of extern controllers and multiple instances of Webots: https://cyberbotics.com/doc/guide/running-extern-robot-controllers#multiple-concurrent-simulations

DavidMansolino:
> Yes, that might be an issue for external controllers because they don't
receive the WEBOTS_TMP_PATH environment variable.

You are completely correct!

> Another solution on Linux would be to use /run/user/$UID/, which is a
private temporary directory. However, unlike /tmp, that is a RAM-disk,
so it could run out of space in extreme cases (huge generated
textures/PROTOs).

I think we can stick with the /tmp solution, we just need to make this function more robust to that it doesn't relies on the pid anymore.
> This `WEBOTS_PID` is used only in the case of extern controllers and multiple instances of Webots: https://cyberbotics.com/doc/guide/running-extern-robot-controllers#multiple-concurrent-simulations

For the `WEBOTS_PID` yes, it is required only in case of multiple instances of Webots, but here the problem is not really `WEBOTS_PID` but rather the fact that the ``wbu_system_webots_tmp_path`` function is expecting a PID (or at least a number) in the temporary folder name (which is not always the case anymore with the `mktemp` function.

mgmax:
You could add some known part to the `mktemp` template, e.g. `mktemp -d /tmp/webots-$PID-XXXXXX`. However, trusting that part may be problematic (?).

MaxGaukler:
You could add some known part to the `mktemp` template, e.g. `mktemp -d /tmp/webots-$PID-XXXXXX`. However, trusting that part may be problematic (?).

DavidMansolino:
> You could add some known part to the `mktemp` template, e.g. `mktemp -d /tmp/webots-$PID-XXXXXX`. However, trusting that part may be problematic (?).

Good to know, this may indeed help a lot. I think about the trust it is quite safe, because at worse, the controller will connect to the wrong instance of Webots but this should not be too critical (as it will not load any library or so from this folder).

src/lib/Controller/api/system.c
```diff
@@ -159,18 +161,39 @@ const char *wbu_system_webots_tmp_path() {
             continue;
           if (s.st_mtime < most_recent)
             continue;
-          sscanf(entry->d_name, "webots-%d", &webots_pid);
+          sscanf(entry->d_name, "webots-%d%s", &webots_pid, random_part);
```
MaxGaukler, line 164: The `random_part` buffer could overflow:
```suggestion
          if (strlen(entry->d_name) > 64) {
              continue; // avoid buffer overflow
          }
          sscanf(entry->d_name, "webots-%d%s", &webots_pid, random_part);
```

DavidMansolino: Thank you for the pointer, our test suite just warn me about the same issue (https://travis-ci.com/github/cyberbotics/webots/jobs/322059423#L275)

DavidMansolino:
I have checked all the use case (normal controller, extern controller with and without WEBOTS_PID defined) both on Linux and macOS and both are working fine.
Our complete test suite is also working: https://travis-ci.com/github/cyberbotics/webots/builds/161427095

src/lib/Controller/api/system.c
omichel: I would rather use:
```suggestion
          sscanf(entry->d_name, "webots-%d-%63s", &webots_pid, random_part);
```

src/lib/Controller/api/system.c
```diff
@@ -159,18 +161,45 @@ const char *wbu_system_webots_tmp_path() {
...
+          struct stat s;
+          snprintf(path_buffer, path_buffer_size, "%s/%s", tmp, entry->d_name);
+          if (stat(path_buffer, &s) < 0)
+            continue;
+          if (!S_ISDIR(s.st_mode))
+            continue;
+          if (strlen(entry->d_name) > 64)
+            continue;
+          // cppcheck-suppress invalidscanf
+          sscanf(entry->d_name, "webots-%d%s", &webots_pid, random_part);
```
omichel, line 191:
```suggestion
          sscanf(entry->d_name, "webots-%d-%63s", &webots_pid, random_part);
```

src/lib/Controller/api/system.c
```diff
@@ -159,18 +161,45 @@ const char *wbu_system_webots_tmp_path() {
...
+      }
+      closedir(dir);
+    }
   }
   if (webots_pid == 0) {
     free(path_buffer);
     path_buffer = NULL;
   } else
-    snprintf(path_buffer, path_buffer_size, "%s/webots-%d", tmp, webots_pid);
+    snprintf(path_buffer, path_buffer_size, "%s/webots-%d%s", tmp, webots_pid, random_part);
```
omichel, line 202:
```suggestion
    snprintf(path_buffer, path_buffer_size, "%s/webots-%d-%s", tmp, webots_pid, random_part);
```

DavidMansolino:
> I believe it's cleaner to use %63s to avoid the buffer overflow problem rather than adding a test and silence the warning.

This can remove the need to silent the warning indeed but it is dangerous (and therefore the test should remain) as if such folder exist it will only partially read the name and then return a non existing folder name.

> Also, it's more readable to include the dash - in the sscanf / sprintf formula.

This is not possible as the dash is not present on MacOS and Windows.

omichel:
I see, so on Windows and mac OS, the random part is an empty string, this is why we should not hard-code the dash. And yes, I understand we should keep the test.

DavidMansolino:
@MaxGaukler thank you for your nice contribution! I am allowing myself to merge this PR.

PR #1573, opened 2020-04-21
Fixed 3D view restoration

### LINKED ISSUE #1572: 3D view disappearing
A couple of users recently reported that the 3D view disappeared and they were unable to get it back, using the `Restore Layout` and/or re-enabling the `3D View` from the `Tools` menu. This may appear indeed when for example the scene tree is large enough and the text editor is also large. When re-enabling the `3D View`, the width of the 3D view may be 0 and one should click on the drag bar to make it visible again. This is counter intuitive and we should ensure that the width and height of the 3D view is not 0 just after enabling it from the `Tools` menu.

omichel:
Fixes #1572.
@DavidMansolino: could you also test it on mac OS?
The testing procedure is:

1. open Webots.
2. ensure the text editor and the scene tree have a large width.
3. hide the 3D view from the Tools menu.
4. show the 3D view again from the same menu.

On the master branch, it will not show the 3D view again.
On this branch, it should work.

### DIFF
--- a/docs/reference/changelog-r2020.md
+++ b/docs/reference/changelog-r2020.md
@@ -5,7 +5,7 @@ Released on XXX.
 
   - New Features
     - macOS and Windows: Added support for Python 3.8.
-    - Added a model of a Mercedes-Benz Sprinter and of a GreenPower EV Star vehicles.
+    - Added a model of a Mercedes-Benz Sprinter.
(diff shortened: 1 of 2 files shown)


DavidMansolino:
👀

omichel:
Well spotted! I believe I just fixed this behavior mismatch.

DavidMansolino:
👀

omichel:
Thank you for the tests!

PR #1577, opened 2020-04-21
Add aperture to ToF sensor

### LINKED ISSUE #1574: E-Puck Distance Sensors are not Correctly Modeled
I observed an inconsistent behaviour of the ToF sensor on the e-puck2 on the physical and simulated robot while making ROS2 node. Field Of View (aperture) is not configured for the ToF. After checking the proto I noticed that FoV is not configured for infra-red sensors either.

**ToF**: According to documentation FoV of ToF sensor used on the e-puck2 is 25 degrees:
http://projects.gctronic.com/epuck2/doc/VL53L0X-Datasheet.pdf (page 25)
while in the simulation the ToF sensor uses the default aperture (90 degrees). Also, `numberOfRays` should be configurable as it is done for infra-red sensors. 

**Infra-red**: Again, the aperture is not configured. According to GCtronic, it is 15 degrees:
https://github.com/gctronic/epuck_driver_cpp/blob/b728d28ca2986dc17030f3d01c3b869d49592d17/src/epuck_driver_cpp.cpp#L1046
but this should be verified. Unfortunately, I am not sure how to estimate it. Let me know if you have an idea? Or should we use the same value as GCtronic?

Furthermore, default value of `distance_sensor_numberOfRays` is 1:
https://github.com/cyberbotics/webots/blob/master/projects/robots/gctronic/e-puck/protos/E-puck.proto#L27
I assume the value is chosen to increase performance, but we maybe should consider increasing it if the performance impact is not too high?

lukicdarkoo:
**Description**
Models the distance sensors better (most noticeably ToF) ~~and allows configurable aperture~~.

**Related Issues**
Fixes #1574

**Tasks**
  - [x] Add aperture field to ToF sensors
  - [x] ~~Allow user to configure ToF and infra-red sensors (not sure if it is a good idea)~~
  - [x] Update documentation

### DIFF
--- a/projects/robots/gctronic/e-puck/protos/E-puckDistanceSensor.proto
+++ b/projects/robots/gctronic/e-puck/protos/E-puckDistanceSensor.proto
@@ -8,7 +8,6 @@ PROTO E-puckDistanceSensor [
   field SFRotation rotation     0 1 0 0
   field SFString   name         "ds"
   field SFInt32    numberOfRays 1         # Is `DistanceSensor.numberOfRays`.
-  field SFFloat    aperture     0.1       # Is `DistanceSensor.aperture`.
 ]
 {
   Solid {
@@ -66,7 +65,6 @@ PROTO E-puckDistanceSensor [
         ]
         type "infra-red"
         numberOfRays IS numberOfRays
-        aperture IS aperture
       }
     ]
     name IS name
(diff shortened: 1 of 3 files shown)


projects/robots/gctronic/e-puck/protos/E-puckDistanceSensor.proto
```diff
@@ -65,7 +65,6 @@ PROTO E-puckDistanceSensor [
         ]
         type "infra-red"
         numberOfRays IS numberOfRays
-        aperture 0.1
```
DavidMansolino, line 68: This will imply that the aperture is ``1.5708``, is this wanted?

PR #1583, opened 2020-04-23
Update motor.md

### LINKED ISSUE #1582: Contradiction in the Motor documentation
https://github.com/cyberbotics/webots/blob/master/docs/reference/motor.md#force-and-torque-control

There is a contradiction between the description and the table.  
The description says that the PID is used when doing velocity control:
> The position (resp. velocity) control described above are performed by the Webots PID-controller

while the table says the opposite:

> 
| &nbsp;                                                                         | position control                                 | velocity control                                 | force or torque control                      |
| ------------------------------------------------------------------------------ | ------------------------------------------------ | ------------------------------------------------ | -------------------------------------------- |
| uses PID-controller                                                            | yes                                              | no                                               | no                                           |

omichel:
Fixes #1582.

### DIFF
--- a/docs/reference/motor.md
+++ b/docs/reference/motor.md
@@ -167,7 +167,7 @@ Finally, in MATLAB you should use the `inf` constant.
 
 ### Force and Torque Control
 
-The position control described above are performed by the Webots PID-controller and ODE's joint motor implementation (see ODE documentation).
+The position control described above is performed by the Webots PID-controller and ODE's joint motor implementation (see ODE documentation).
 As an alternative, Webots also allows the user to directly specify the amount of force (resp. torque) that must be applied by a [Motor](#motor).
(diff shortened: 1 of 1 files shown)


docs/reference/motor.md
```diff
@@ -167,8 +167,8 @@ Finally, in MATLAB you should use the `inf` constant.
 
 ### Force and Torque Control
 
-The position (resp. velocity) control described above are performed by the Webots PID-controller and ODE's joint motor implementation (see ODE documentation).
-As an alternative, Webots does also allow the user to directly specify the amount of force (resp. torque) that must be applied by a [Motor](#motor).
+The position control described above are performed by the Webots PID-controller and ODE's joint motor implementation (see ODE documentation).
```
stefaniapedrazzi, line 170: is it not better to use the singular form?

```suggestion
The position control described above is performed by the Webots PID-controller and ODE's joint motor implementation (see ODE documentation).
```

PR #1585, opened 2020-04-23
Fixed Windows launchers

omichel:
At some point the Windows launchers were failing on machine but I could not understand why. Now they are working fine again and I don't understand why, but I cannot debug any more what was the problem. So, I introduced a better error reporting in these launchers to be able to better understand the problem whenever it strikes back.


DavidMansolino:
👀

Issue #1580, opened 2020-04-22
Classic Theme Broken on MacOS


DavidMansolino:
**Describe the Bug**
When using the system dark mode in macOS if Webots uses the classic theme then the GUI colors are completely broken:

![Screenshot 2020-04-22 at 07 19 14](https://user-images.githubusercontent.com/2461619/79943676-c208a380-8469-11ea-82ef-e541aa37b6e5.png)

PR #1586, opened 2020-04-23
Fix mac classic mode.

DavidMansolino:
**Description**
When using the system dark mode in macOS if Webots uses the classic theme, then the GUI colors are completely broken

**Related Issues**
This pull-request fixes issue #1580


DavidMansolino:
Classic theme is now independent from the system theme on macOS:
Screenshot 2020-04-23 at 11 13 22

PR #1595, opened 2020-04-24
Feature Multi Console

DavidMansolino:
**Description**
This PR is just for investigating and recording several possible solutions to add the support for multi-console (see https://github.com/cyberbotics/webots/pull/1559#issuecomment-616427156).

Doc: https://cyberbotics.com/doc/guide/the-console?version=feature-multi-console

Current design (design only, no implementation), feel free to give your opinion before I start the actual implementation:
Screenshot 2020-04-24 at 16 09 31

  - Consoles are dockable widgets (so that we can move them whereas we want) which are by default stacked as tabs (see the tabs at the bottom).
  - On the top right of each console a dropdown menu allows to select all the controllers (by robot name) + Webots, that we want to display in this console (it is possible for example to display the output of one controller in more than on tab).
  - In the ``Tools`` menu:
    - The `Clear Console` item will be renamed into `Clear Selected Console` (or current instead of selected).
    - A new `Clear all Consoles` item will be added.
    - A new `New Console` item will be added (it will open a new console).
  - The configuration of the consoles will be saved in the *.wbproj file.
  - By default Webots will have only one console which will display all the controllers output + Webots (the behaviour will therefore be exactly the same as currently, except for a few worlds where it makes more sense to split the consoles).

Alternatives:
  - The consoles might be all in one single dockable widget (tab widget), this will allow to move all the console at the same time but we will loose in flexibility (e.g. not possible to see more than one console at the same time).
  - Instead of a dropdown menu for selecting what should be displayed in which console we can use the context menu of the console.
  - Instead of using the robot names to filter controller outputs we can use the controller names (which will make the list shorter but will decrease the filtering granularity).
  - We might also use a Robot.field to determine if it should be in a dedicated console (very bad idea IMO).


DavidMansolino:
A few previews (next step is to be able to show controllers independently):
Multi-console view:
Screenshot 2020-04-29 at 12 33 30
Tabbed consoles (default when a new console is opened):
Screenshot 2020-04-29 at 12 33 58
Selection of what should be displayed:
Screenshot 2020-04-29 at 12 34 17

omichel:
That looks very nice. I have a few comments and questions:

- In the robot console, I believe we should display only message explicitly sent by the robot controller via `printf`, e.g., no `Warning: E-puck (PROTO): 'camera_width' has been...` neither `INFO: khepera3_gripper: Starting controller: ...`. Such messages should be displayed in the Webots console IMO.
- I believe the console filter should be either a drop down menu or a context menu of the console window, but not a dialog window as displayed in your last screenshot.
- About the Webots console, it's nice to distinguish ODE messages from others, but I am wondering how many categories we could make... One for ODE, one for PROTO parsing, one for controllers starting/ending, etc.? Also or alternatively, we could define a log level: DEBUG/INFO/WARNING/ERROR and filter out messages that are below the selected level. Not sure what is the best...

Let me know when you want me to checkout and compile this branch to make further tests and evaluation. These comments are only based on the screenshots you posted here, so they don't rely on a real user experience.

DavidMansolino:
> In the robot console, I believe we should display only message explicitly sent by the robot controller via print

👍 

> no Warning: E-puck (PROTO): 'camera_width' has been...

This should already be the case.

> neither INFO: khepera3_gripper: Starting controller: ...

I was hesitating for this one, anyway it is very easy to change this one, I will move it.

> I believe the console filter should be either a drop down menu or a context menu of the console window, but not a dialog window as displayed in your last screenshot.

My first iteration was with a combobox, but this is a real nightmare for the implementation and unusable for the user as soon as the number of controller increases.
I can try with the context menu but I am afraid it is also going to be difficult to use when the number of controllers increases.

> About the Webots console, it's nice to distinguish ODE messages from others, but I am wondering how many categories we could make... One for ODE, one for PROTO parsing, one for controllers starting/ending, etc.? 

Very good point, I was also thinking about the javascript outputs of the robot windows. I did only ODE for now for a proof of context (and because personally (and some users reported it too) very often I want to hide the ODE warnings).

> Also or alternatively, we could define a log level: DEBUG/INFO/WARNING/ERROR and filter out messages that are below the selected level. Not sure what is the best...

Yes that's on my plans too (log level + the possibility to chose stderr and stdout for controllers). This is also one more argument against a drop-down menu.

> Let me know when you want me to checkout and compile this branch to make further tests and evaluation. These comments are only based on the screenshots you posted here, so they don't rely on a real user experience.

Thank you, it is really nice to have feedback on the screenshots (and this is why I posted them), but remains a lot of details (and important features) and bugs to implement (to be honest this is a long shot PR I consider the current state as a draft only).

DavidMansolino:
I think I implemented most of the missing features and fix all the biggest bugs.

Selection in a context menu:
Screenshot 2020-05-01 at 09 18 12

Per robot selection:
Screenshot 2020-05-01 at 09 18 56

For now both the dialog and context menu are kept (depending on the case one or the other is more simpler to use). But it will be easy to remove one of them.

I have one hesitation about the selection:
- When we select 'All' should we unselect all the others automatically (and vice-versa), this looks like what we expect to see but on the other side it is probably not so nice (e.g. if for debugging you want to display everything just for now run you don't want to 're-create' your selection when you finished debugging but just unselecting 'All').

About the 'system' category, I added only 3 now:
  - ODE errors
  - Javascript
  - Webots: all the others.
These are for me the most important but don't hesitate to suggest more (it should be pretty easy to add).

And I didn't add the log level. I propose to create 4 categories:
  - error: WbLog::ERROR level and controller stderr.
  - warning: WbLog::WARNING level.
  - info: WbLog::INFO and WbLog::Debug level.
  - standard: controller stdout.

> Let me know when you want me to checkout and compile this branch to make further tests and evaluation. These comments are only based on the screenshots you posted here, so they don't rely on a real user experience.

It is not yet perfect, but I think it is now ready for a first behavioural test (please do not look at the code for now).

omichel:
It doesn't compile on Windows:
```
control/WbController.cpp: In member function 'void WbController::start()':
control/WbController.cpp:205:33: error: cannot convert 'const QString' to 'bool'
  205 |                    robot()->name());
      |                    ~~~~~~~~~~~~~^~
      |                                 |
      |                                 const QString
In file included from control/WbController.cpp:24:
core/WbLog.hpp:44:52: note:   initializing argument 2 of 'static void WbLog::warning(const QString&, bool, const QString&)'
   44 |   static void warning(const QString &message, bool popup = false, const
```
It looks like some changes were not pushed to GitHub?

DavidMansolino:
> It looks like some changes were not pushed to GitHub?

Unfortunately not, it is working for me (and for Travis), I will check this.

omichel:
It looks very good so far. A few minor comments:

- I believe that when selecting some items in the filter list, the `All` option should automatically be disabled (and vice-versa). For debugging purposes, it is very easy to create a new console with `All`.
- when a new console is created, it should become the active selected console.
- we may add a top category to select all `System` messages instead of having add one-by-one `Javascript`, `ODE`, etc.
- it would be handy to have the possibility to rename the titles of the consoles (instead of having console (1), (2), (3), etc.).
- even better the default name (if not set by the user) should be the filter value.
- about naming, I propose the following changes:
  - `System` &rarr; `Webots` or `Simulation`?
  - `Webots` &rarr `Others`
  - `ODE errors` &rarr; `ODE` or `Physics`?
  - `Clear Consoles` &rarr; `Clear All Consoles`
- we may also add a `PROTO` category?
- the four log levels you propose seems very good to me.
- not sure: for controllers, shall we add filters for `stderr` and `stdout`? (this may not be a good idea)

DavidMansolino:
Thank you for your tests and very pertinent comments.

> It looks very good so far. A few minor comments:
> 
> * I believe that when selecting some items in the filter list, the `All` option should automatically be disabled (and vice-versa). For debugging purposes, it is very easy to create a new console with `All`.

👍 

> * when a new console is created, it should become the active selected console.

👍 

> * we may add a top category to select all `System` messages instead of having add one-by-one `Javascript`, `ODE`, etc.

👍  (same for the controllers?)

> * it would be handy to have the possibility to rename the titles of the consoles (instead of having console (1), (2), (3), etc.).

I agree been able to rename them would be nice (not sure how to do this, but I will investigate).

> * even better the default name (if not set by the user) should be the filter value.

This is the case for the non-tabbed case (see for example in first picture of my previous message: `Console - Webots | Javascript`). But in the tabbed case we should probably keep the name short (as we don't want to have huge tabs in case several filters are selected).

> * about naming, I propose the following changes:
>   
>   * `System` → `Webots` or `Simulation`?
>   * `Webots` &rarr `Others`
>   * `ODE errors` → `ODE` or `Physics`?
>   * `Clear Consoles` → `Clear All Consoles`

👍 

> * we may also add a `PROTO` category?

What would this category represents? Parsing errors from PROTOs (in that case what about parsing error in worlds?)? Lua outputs?

> * the four log levels you propose seems very good to me.

👍 

> * not sure: for controllers, shall we add filters for `stderr` and `stdout`? (this may not be a good idea)

Do you suggest to have different categories for controllers and system messages? Or to be able to have per log level, e.g.:
```
Filters
       -> All
       -> System
                 -> Webots
                            -> Error
                            -> Warning
                            -> ...
                 -> ...
       -> Controller(s)
                 -> robot1
                            -> Stdout
                            -> Stderr
                 -> ...
```

VS global level:
```
Filters
       -> All
       -> System
                 -> Webots
                 -> ...
       -> Controller(s)
                 -> robot1
                 -> ...
Level
       -> Error
       -> Warning
       -> Info
       -> Standard
```

omichel:
>>        we may add a top category to select all System messages instead of having add one-by-one Javascript, ODE, etc.
> 👍 (same for the controllers?)

Yes, that would be nice as well (and consistent with the other sub-menu).

> This is the case for the non-tabbed case (see for example in first picture of my previous message: Console - Webots | Javascript). But in the tabbed case we should probably keep the name short (as we don't want to have huge tabs in case several filters are selected).

I agree. Maybe if only one filter is selected we should use this one. If several filters are selected, we could use `console`.

> What would this category represents? Parsing errors from PROTOs (in that case what about parsing error in worlds?)? Lua outputs?

I would put together parsing errors from PROTOs and Lua outputs. And I would create a new "World" category for parsing worlds.

> Do you suggest to have different categories for controllers and system messages? Or to be able to have per log level

The per log level option (first one) seems good to me, easier to understand. We should also introduce `All` shortcuts at the top of the sub-menus to select all levels or select both stdout and stderr in the controller case.

DavidMansolino:
Just to summarize (because I am already starting getting lost):
  - [ ]
  - [ ]
  - [ ]
  - [ ]
  - [ ]
  - [ ]

DavidMansolino:
Just to summarize (because I am already starting getting lost):
  - [x] Renaming filters.
  - [x] Disable individual filters once 'all' is selected.
  - [ ] Disable 'All' when an individual filter is selected.
  - [ ] Add a PROTO filter.
  - [ ] Add a World filter.
  - [ ] Add a 'All Controller(s)' filter on top of the controller filters.
  - [ ] Add a 'All Webots' filter on top of the Webots filters.
  - [ ] Make newly created console the active one.
  - [ ] Add log level filters.
  - [ ] Display filter name if only one is active in the tabbed case.
  - [ ] Allows the user to define custom names for consoles.

DavidMansolino:
About allowing the user to rename the console (title), I am not sure how to integrate this in the GUI.
For now the best way I can think of, is to add a 'Set Name' action in the context menu that will open a dialog window allowing to set the name.
Any simpler/more intuitive ideas are welcome.

omichel:
I don't have any better idea... Maybe "Rename" instead of "Set Name".

DavidMansolino:
The renaming of the console is now functional and saved in the wbproj file.

Remains only the PROTO and world filters.
But while implementing I am not so sure anymore it makes sense to make the distinction between these two.
For example with a simple case, with the `Box` node, a warning is raised if one of the component of the of the `size` field is <= 0.
Does it really makes sense to use a different filter for the same warning if it is raised from within a PROTO or a world file?
Maybe adding a kind of 'field validation' filter would make more sense.

omichel:
Yes, we can probably group PROTO and world messages together. But I would maybe name this category "world parsing" instead of "field validation".

DavidMansolino:
This PR is now ready for a new behavioural review (I am cleaning-up the code).

omichel:
:eyes:

omichel:
It is very nice.
However, I found a little problem: when setting several consoles and reloading the simulation, the order of the stacked consoles gets changed.
It would be nice to keep the order set by the user.

DavidMansolino:
> It is very nice.
> However, I found a little problem: when setting several consoles and reloading the simulation, the order of the stacked consoles gets changed.
> It would be nice to keep the order set by the user.

Well spotted! I will try to fix this.

omichel:
In case no Webots filter is set, I would disable the `Levels` menu, as it is meaningless for controllers.
Also, I would rename: `Filters` &rarr; `Filter` and `Levels` &rarr; `Level`. I believe the singular form looks better.

DavidMansolino:
> In case no Webots filter is set, I would disable the Levels menu, as it is meaningless for controllers.

It's not completely meaningless  as there is stdout and stderr.

> Also, I would rename: Filters → Filter and Levels → Level. I believe the singular form looks better.

👍

omichel:
> It's not completely meaningless as there is stdout and stderr.

So, what does correspond to stdout and stderr? Error / Standard? That was not obvious to me on the first sight...
Shall we replace these with `stdout` and `stderr` and disable them when no controller is selected?
Or maybe to it event clearer, add an extra layer in `Level`:

```
Level
   |----->Webots
   |               |-----> All
   |               |-----> Info
   |               |-----> Warning
   |               |-----> Error
   |
   |------> Controller(s)
                       |-------->All
                       |--------> stdout
                       |--------> stderr
```
In that case, we should disable the `Controller(s)` sub-menu if no `Controller(s)` filter is set.
Similarly, we should disable the `Webots` sub-menu if no `Webots` filter is set.

omichel:
Compiling a controller will clear all the consoles, that seems a bit strange. I believe no console should be cleared in that case.

omichel:
I am wondering if we should add another Webots filter for `Compilation`.

DavidMansolino:
> > It's not completely meaningless as there is stdout and stderr.
> 
> So, what does correspond to stdout and stderr? Error / Standard? That was not obvious to me on the first sight...

Yes it's exactly this.
This is what I proposed in my proposal few messages above (https://github.com/cyberbotics/webots/pull/1595#issuecomment-622284029) that you initially approved:
> error: WbLog::ERROR level and controller stderr.
>warning: WbLog::WARNING level.
>info: WbLog::INFO and WbLog::Debug level.
>standard: controller stdout.

> Shall we replace these with `stdout` and `stderr` and disable them when no controller is selected?
> Or maybe to it event clearer, add an extra layer in `Level`:
> 
> ```
> Level
>    |----->Webots
>    |        |-----> All
>    |        |-----> Info
>    |        |-----> Warning
>    |        |-----> Error
>    |
>    |------> Controller(s)
>                  |--------> All
>                  |--------> stdout
>                  |--------> stderr
> ```
> 
> In that case, we should disable the `Controller(s)` sub-menu if no `Controller(s)` filter is set.
> Similarly, we should disable the `Webots` sub-menu if no `Webots` filter is set.

Yes that might indeed be a good solution, I am just wondering if we are not starting to have a too complex interface having sub-sub-sub-menu of a context menu is not really KISS.

> Compiling a controller will clear all the consoles, that seems a bit strange. I believe no console should be cleared in that case.

👍 

> I am wondering if we should add another Webots filter for Compilation.

👍

omichel:
> Yes that might indeed be a good solution, I am just wondering if we are not starting to have a too complex interface having sub-sub-sub-menu of a context menu is not really KISS.

If we don't want too deep sub-menus, we could have Webots and Controller(s) at the same level and use a separator to separate them:
```
Level
   |----->All Webots
   |-----> Info
   |-----> Warning
   |-----> Error
   | ----------------- (separator line)
   |------> All Controller(s)
   |------> stdout
   |------> stderr
```

In that case, to be consistent, we should have the same organization in the `Filter` menu.

DavidMansolino:
> > Yes that might indeed be a good solution, I am just wondering if we are not starting to have a too complex interface having sub-sub-sub-menu of a context menu is not really KISS.
> 
> If we don't want too deep sub-menus, we could have Webots and Controller(s) at the same level and use a separator to separate them:
> 
> ```
> Level
>    |-----> All Webots
>    |-----> Info
>    |-----> Warning
>    |-----> Error
>    |------------- (separator line)
>    |-----> All Controller(s)
>    |-----> stdout
>    |-----> stderr
> ```
> 
> In that case, to be consistent, we should have the same organization in the `Filter` menu.

Good idea, I prefer this way (even If it complexity a bit the implementation, but to me it looks simpler for the user).

omichel:
:eyes:

omichel:
It seems the order of the console tabs is still not preserved after quitting Webots and restarting the same simulation...

DavidMansolino:
> It seems the order of the console tabs is still not preserved after quitting Webots and restarting the same simulation...

That's strange, I tested this and it is working for me, this was supposed to be fixed in ac98f6fb5d65a741bf74ee0bc087614fc542af3a

Do you have a specific procedure to reproduce this?

src/webots/core/WbLog.hpp
```diff
@@ -33,16 +33,35 @@ class WbLog : public QObject {
...
 
   static WbLog *instance();
 
   // Webots messages, e.g. parse error
   // the 'message' argument should not be '\n' terminated
-  static void debug(const QString &message, bool popup = false);
-  static void info(const QString &message, bool popup = false);
-  static void warning(const QString &message, bool popup = false);
-  static void error(const QString &message, bool popup = false);
+  static void debug(const QString &message, bool popup = false, const QString &name = QString());
```
omichel, line 55: Why passing a `QString` for the `name`? Passing a `WbLog::Filter` type would be simpler, no?
If for some reason a `QString` is really needed, we could add helper methods like:
```c++
   static void debug(const QString &message, bool popup = false, Filter filter);
```

src/webots/vrml/WbNode.hpp
```diff
@@ -126,9 +126,9 @@ class WbNode : public QObject {
 
   // error reporting
   // field name will be extracted from the first single quoted text in the message
-  void warn(const QString &message) const;  // show standard warning message formatted for this node
-  void info(const QString &message) const;  // show standard info message formatted for this node
-  QString usefulName() const;               // user friendy node name for error messages
+  void warn(const QString &message, bool parsingMessage = true) const;  // show standard warning message formatted for this node
+  void info(const QString &message, bool parsingMessage = true) const;  // show standard info message formatted for this node
```
omichel, line 130: It seems dangerous to use a default value to true for the parsing case. Maybe next time we will write a warning or an info for a node, we will likely forget to set this flag to false and it will appear as a parsing message. I would reverse this and set the default value to false which seems safer to me...

DavidMansolino:
> It seems the order of the console tabs is still not preserved after quitting Webots and restarting the same simulation...

This should now be fixed in all cases.

omichel:
> This should now be fixed in all cases.

Yes, that works! Thank you.

src/webots/vrml/WbNode.hpp
DavidMansolino: I don't see why the other way would be safer, it will just reverse the problem:
Next time we will write a warning or an info for a node we will likely forget to set this flag to true and it will appear as an normal message and not as a parsing message.

To me the default argument in that case should reflect the most used case, which in that case is quite clear (just a few tens for non-parsing messages VS ~500 for the parsing case).

If we really want to be safe then we should just make this argument mandatory.

src/webots/core/WbLog.hpp
DavidMansolino: Because having the information about the type of filter is not enough, for example if the type is `WbLog::CONTROLLERS`, we have to know which controller it is too.

DavidMansolino:
✔️ : travis_release

omichel:
🥳

PR #1616, opened 2020-05-03
fix dir to create tar and update qt download link

LeoJhonSong:
- It seems that we could not download qt from qt.org anymore, I suggest change it to https://download.qt.io/official_releases/qt/
- since change dir to `$WEBOTS_HOME/lib/webots` at [line 115](https://github.com/cyberbotics/webots/blob/08af20daa37028f1d076ba123b1d2b5fa6822d6a/src/install_scripts/qt_linux_installer.sh#L115), we should use `cd ../..` to change to `$WEBOTS_HOME`, or we could not find the target files when creating a tar.

### DIFF
--- a/src/install_scripts/qt_linux_installer.sh
+++ b/src/install_scripts/qt_linux_installer.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# install the latest Qt from https://download.qt.io/official_releases/qt/
+# follow the instructions from https://github.com/cyberbotics/webots/wiki/Qt-compilation#linux to download and compile Qt before executing this script.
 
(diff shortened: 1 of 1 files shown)


src/install_scripts/qt_linux_installer.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# install the latest Qt from qt.org
+# install the latest Qt from https://download.qt.io/official_releases/qt/
```
omichel, line 3:
```suggestion
# follow the instructions from https://github.com/cyberbotics/webots/wiki/Qt-compilation#linux to download and compile Qt before executing this script.
```

Issue #453, opened 2019-05-15
--minimize option not working on Ubuntu


fabienrohrer:
Launching webots with the `--minimize` option is not working on Ubuntu.

<img width="1680" alt="Capture d’écran 2019-05-15 à 11 39 36" src="https://user-images.githubusercontent.com/866788/57765646-2c677e80-7706-11e9-8043-4888f883e432.png">

DavidMansolino:
I can't reproduce the problem on Ubuntu 16.04 on master.

DavidMansolino:
For me (on Ubuntu 16.04) it is working fine on revision and develop too.

stefaniapedrazzi:
Calling `showNormal()` followed by `setWindowState(Qt::WindowMinimized)` seems to work on my Ubuntu 18.04 and the application is not displayed on the screen.

stefaniapedrazzi:
I could reproduce the same issue based on a Qt example and I reported it:
https://bugreports.qt.io/browse/QTBUG-76354

stefaniapedrazzi:
I just checked and this Qt bug has been fixed in Qt 5.14.2.
The current version used on Linux is 5.13.1. When updating to the latest Qt version we can get rid of the hotfix introduced in  #563.

PR #1617, opened 2020-05-04
Update to Qt 5.14.2 on Linux

stefaniapedrazzi:
Update Qt dependency to 5.14.2 on Linux.
The new Qt version fixes a bug with `showMinimized` (see #453).

Note that Windows is also using 5.14.2, so probably it would also be good to update the Qt dependency on macOS.


stefaniapedrazzi:
The code is compiling fine after updating to Qt 5.14.2 (this version was already used on Windows).
But Webots is crashing on quit after running a world with HTML robot windows.

omichel:
I confirm Webots is not crashing on Windows after running a simulation with a HTML robot window with Qt-5.14.2.

stefaniapedrazzi:
:heavy_check_mark: travis_unit

It would be better to also test on mac if deleting the robot window QWebEnginePage is causing issues.

DavidMansolino:
I will test this evening on the mac.

stefaniapedrazzi:
@DavidMansolino thank you!

To test you can simply run the test suite or run for example the `custom_html_robot_window.wbt` simulation and then quit Webots.

stefaniapedrazzi:
I target the develop branch because this is not a hotfix (everything already works correctly) and we will release soon R2020b.
But yes, if the path is set correctly on Webots start, it should not break the compatibility unless users uses the distributed Qt libraries for their controller/plugins.

PR #1610, opened 2020-04-30
Add Transformation Tree Getter to Controller API

lukicdarkoo:
**Description**
Add controller API support to read TF tree of the robot.

**Related Issues**
Closes #1603

**Tasks**
  - [x] Create `tf_tree_print()` to debug the results
  - [ ] Read transformation tree in `WbRobot.cpp`
    - [x] Find and send relevant primitives (joints and links) in the robot tree
    - [x] Rotation and translation in respect to the parent node
    - [x] Store relevant HingeJointParameters 
    - [x] Add support for different joint implementations. Find solid endpoints and continue parsing the tree.
      - [x] WbSliderJoint
      - [x] WbHingeJoint2
      - [x] BallJoint
      - [x] WbHingeJoint
    - [x] Compress the tf (we don't care about link<->link, only device<->joint)
    - [ ] Verify if there are relevant and unhandled nodes (if it works for e-puck doesn't mean it will work for the other robots)
  - [x] Add API for controller
    - [x] Parse the transformation tree
    - [x] Expose getter function
    - [x] Clean the tree
    - [x] Add Python support to be able to test it with `webots_ros2`
  - [ ] Add support for languages other than C
  - [ ] Add documentation
  - [ ] Add unit tests
  - [x] Shift origin of joint in case the parent is node is considered irrelevant
  - [ ] Test
    - [x] WbHingeJoint
    - [ ] WbHingeJoint2
    - [ ] BallJoint
    - [ ] WbSliderJoint

**Additional context**

<details><summary>Controller for the tree printing and visualization (.dot)</summary>

```c
#include <stdio.h>
#include <webots/robot.h>

#define TIME_STEP 64

static void tf_make_dot_rec(FILE *f, WbTransformNodeObject *root) {
  // Generate .dot file. You can visualize it later as:
  // $ dot -Tpng /tmp/robot.dot > /tmp/robot.png && xdg-open /tmp/robot.png
  for (int i = 0; i < root->n_children; i++) {
    fprintf(f, "%d -> %d;\n", root->id, root->children[i]->id);
    if (root->children[i]->type == WB_TF_NODE_LINK)
      fprintf(f, "%d [shape=box];\n", root->children[i]->id);
    tf_make_dot_rec(f, root->children[i]);
  }
}

static void tf_make_dot(WbTransformNodeObject *root) {
  FILE *f;
  f = fopen("/tmp/robot.dot", "w");
  fprintf(f, "digraph graphname {\n");
[...]

Issue #1603, opened 2020-04-29
Missing Device Fields Required for Automatic ROSification


lukicdarkoo:
**Is your feature request related to a problem? Please describe.**
In `webots_ros2` we would like to create automatic ROSification of the given robot (related PR https://github.com/cyberbotics/webots_ros2/pull/78). This means, that e.g.:
- All camera devices would be discovered, corresponding ROS2 topics of type `Image` and `CameraInfo` would be created, the camera would be enabled/disabled depending on the number of subscribers and transformation (`TransformStamped` messages with `tf2_ros`) of the camera in respect to the robot would be also published.
- All distance sensor device would be discovered, messages of type `Range` would be published, corresponding transforms as well.
- Similar applies for IMU, light sensors, LEDs and other devices.

However, we encountered a few obstacles:
- Cannot get a transformation from the robot (or world) to the device so we can publish `TransformStamped` messages.
- Cannot access to lookup table field of a distance sensor, light sensor or any other sensor, so we can get the actual value that we can publish as `Range`, `Illuminance` or similar.
- Cannot determine if InertialUnit, Gyro and Accelerometer are the part of the same IMU. This could be done by if knew that all the devices have the same transformation with respect to the robot. This assumption would not cover all the cases, but it would be configurable. 

All required fields (lookup table and transformations) are accessible if the robot is a supervisor. 

**Describe the solution you'd like**
Allow the robot to read the following fields:
- A **lookup table** of DistanceSensor, LightSensor and other sensors with a lookup table.
- **Transformation** of a device with respect to the robot/world origin.

This may not be exhaustive list of fields. Maybe we should finish the PR (https://github.com/cyberbotics/webots_ros2/pull/78) first assuming that we have access to all fields and then create an exhaustive list of fields that should be added to Webots.

**Describe alternatives you've considered**
Converting the robot to a supervisor robot was consider as an alternative. But, giving the role of supervisor to all robots that need to be ROSified destroys the purpose of the supervisor.

DavidMansolino:
This would imply to add these new functions (here in the DistanceSensor case):
```C
const double * wb_distance_sensor_get_translation(WbDeviceTag tag);  // return the translation of the device in the robot coordinate frame.
const double * wb_distance_sensor_get_rotation(WbDeviceTag tag);  // return the rotation of the device in the robot coordinate frame.
const double * wb_distance_sensor_get_lookup_table(WbDeviceTag tag);
[...]

PR #1632, opened 2020-05-05
Add Lookup Table Getter to Controller API

lukicdarkoo:
**Description**
The objective is to allow the user to read values from the lookup table from the controller API.

**Related Issues**
Related to issue #1603. Also, this PR is split from #1610.

**Tasks**
  - [x] DistanceSensor (done in 7299c34)
    - [x] Getter for lookup table size and values
    - [x] Add support for languages other than C
    - [x] Add tests
    - [x] Add documentation - https://cyberbotics.com/doc/reference/distancesensor?version=feature-lookup-table
  - [x] Accelerometer
    - [x] Getter for lookup table size and values
    - [x] Add support for languages other than C
    - [x] Add tests
    - [x] Add documentation - https://cyberbotics.com/doc/reference/accelerometer?version=feature-lookup-table
  - [x] Compass
    - [x] Getter for lookup table size and values
    - [x] Add support for languages other than C
    - [x] Add tests
    - [x] Add documentation - https://cyberbotics.com/doc/reference/compass?version=feature-lookup-table
  - [x] Gyro
    - [x] Getter for lookup table size and values
    - [x] Add support for languages other than C
    - [x] Add tests
    - [x] Add documentation - https://cyberbotics.com/doc/reference/gyro?version=feature-lookup-table
  - [x] InertialUnit
    - [x] Getter for lookup table size and values
    - [x] Add support for languages other than C
    - [x] Add tests
    - [x] Add documentation - https://cyberbotics.com/doc/reference/inertialunit?version=feature-lookup-table
  - [x] LightSensor
    - [x] Getter for lookup table size and values
    - [x] Add support for languages other than C
    - [x] Add tests
    - [x] Add documentation - https://cyberbotics.com/doc/reference/lightsensor?version=feature-lookup-table
  - [x] TouchSensor
    - [x] Getter for lookup table size and values
    - [x] Add support for languages other than C
    - [x] Add tests
    - [x] Add documentation - https://cyberbotics.com/doc/reference/touchsensor?version=feature-lookup-table
  - [x] Test the interface for languages other than C (exhaustive tests are performed only on C API, for other languages only "representative" functions are tested, e.g. obtaining array result from `wb_distance_sensor_get_lookup_table`)
    - [x] C++ is considered to be working fine if Python API works
    - [x] Python
    - [x] ROS (compiled `webots_ros` and patched to support `get_float_array.srv`)
    - [x] Java
    - [x] Matlab (tested on R2019b)
  - [x] Ignore `wb_*_get_lookup_table_size()` in:
    - [x] Matlab, 
    - [x] Python,
    - [x] Java,
    - [x] ROS and
    - [x] documentation.

**Additional context**
Verification controller:
```c
#include <stdio.h>
#include <webots/distance_sensor.h>
#include <webots/robot.h>

#define TIME_STEP 64

void print_lookup_table(WbDeviceTag tag) {
  int size = wb_distance_sensor_get_lookup_table_size(tag);
  const double* lookup = wb_distance_sensor_get_lookup_table(tag);
  printf("Number of items in lookup table %d\n", size);
  for (int i = 0; i < size; i++)
    printf("[#%d] %f %f %f\n", i, lookup[i*3], lookup[i*3+1], lookup[i*3+2]);
}

int main(int argc, char *argv[]) {
  wb_robot_init();
  WbDeviceTag ds = wb_robot_get_device("ps7");
  print_lookup_table(ds);
  wb_robot_cleanup();
  return 0;
}
```


DavidMansolino:
👀

tests/api/worlds/touch_sensor_force.wbt
```diff
@@ -25,6 +25,10 @@ Robot {
       }
     }
     TouchSensor {
+      lookupTable [
+        -1000 -1000 0
+        1000 1000 0
+      ]
```
DavidMansolino, line 31: It would be nice to test also one case where the lookup-table is the default one (not specified) and make sure it returns NULL.

docs/reference/accelerometer.md
```diff
@@ -99,6 +107,8 @@ public class Accelerometer extends Device {
   public void disable();
   public int getSamplingPeriod();
   public double[] getValues();
+  public int getLookupTableSize();
```
omichel, line 110: I am not sure these functions are really useful in Java, Python and MATLAB, as we can retrieve the array size easily:
**Java**
```java
double[] arr = getValues();
int size = arr.length / 3;
```
**Python**
```python
  arr = self.getLookupTable()
  size = len(arr) / 3
```
**MATLAB**
```matlab
  arr = wb_accelerometer_get_lookup_table();
  size = length(arr) / 3;
```
So, I would suggest to remove them.

lib/controller/matlab/wb_accelerometer_get_lookup_table.m
```diff
@@ -0,0 +1,9 @@
+function wb_accelerometer_get_lookup_table(tag)
+% Usage: wb_accelerometer_get_lookup_table(tag)
+% Matlab API for Webots
+% Online documentation is available <a href="https://www.cyberbotics.com/doc/reference/accelerometer">here</a>
+
+size = calllib('libController', 'wb_accelerometer_get_lookup_table_size', tag);
+pointer = calllib('libController', 'wb_accelerometer_get_lookup_table', tag);
+setdatatype(pointer, 'doublePtr', 3 * size, 1);
+result = get(pointer, 'Value');
```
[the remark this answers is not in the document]
lukicdarkoo: Manually, I even created it by hand. Oh, should I execute `mgenerate.py` instead?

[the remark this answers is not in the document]
DavidMansolino: Yes exactly you have to manually run mgenerate.py to generate these files, and in some case the result is not what we expect (mainly when handling pointer) and the file have to be edited manually, in that case the corresponding line should be commented in mgenerate.py (e.g. https://github.com/cyberbotics/webots/pull/1632/files#diff-9f2c0306477f52fc4ea3d24a7996f698R63)

resources/languages/java/controller.i
```diff
@@ -153,8 +153,14 @@ namespace webots {
...
 //----------------------------------------------------------------------------------------------
 
+%typemap(out) const double * {
+  $result = SWIG_JavaArrayOutDouble(jenv, (double *) $1, arg1->getLookupTableSize()*3);
+}
+
 %include <webots/Accelerometer.hpp>
 
+%clear const double *;
+
```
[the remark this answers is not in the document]
lukicdarkoo: I will check. When I tested the API in Python it was printing only a few values, so I needed to define the list length explicitly. It makes sense because the C++ function returns only a pointer and there is no way to determine the length of the array (maybe something like https://stackoverflow.com/a/21185118/1983050).

I will do a more comprehensive verification.

docs/reference/accelerometer.md
lukicdarkoo: This is something I wasn't sure about, whether to keep it consistent or use features of the other languages. The same applies to C++, we could use `vector<double>`, but we decided to be consistent. However, for Java, Python and MATLAB it makes more sense to leave out `getLookupTableSize()`, I will delete them.

omichel: I agree, it is less consistent than other languages, but I believe we should leverage on the capabilities of these languages to simplify our API. Users will have no problem figure out how to use these array length functions.
C++ is a special case as we do not use the vector class in our API so far. So introducing it only for these functions would be a big disruption for a little thing.

lib/controller/matlab/wb_accelerometer_get_lookup_table.m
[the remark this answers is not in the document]
lukicdarkoo: It is even written that I have to run `mgenerate.py` :facepalm: 

image
I ran `mgenerate.py` but now there are a few `ANSI_*.m` files - line endings are different. I guess somebody else missed to do the same thing :) Should I commit them as well?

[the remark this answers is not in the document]
DavidMansolino: Yes please feel free to commit them as well.

tests/api/worlds/touch_sensor_force.wbt
lukicdarkoo: Added the test for touch sensor only, let me know if it should be added for the other sensors as well.

DavidMansolino: One is sufficient enough, thank you!

resources/languages/java/controller.i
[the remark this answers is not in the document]
lukicdarkoo: I just added `double test[3]` property to the accelerometer and it indeed returns a list of 3 values in Python. However, if I declare it as a `double test[1]` or `double test[5]` it still returns list of 3 values. From some reason 3 is a magical number for arrays, but I couldn't find why.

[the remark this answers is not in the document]
DavidMansolino: I am not sure to understand, where did you add this `double test[3]`?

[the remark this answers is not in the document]
lukicdarkoo: Got the same behaviour for Java as well.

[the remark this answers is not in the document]
lukicdarkoo: > I am not sure to understand, where did you add this `double test[3]`?

I added the field (`double test[3]`), just to test the behaviour of C++/Java/Python interface.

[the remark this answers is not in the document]
DavidMansolino: If you use `[X]`, this is mapped here:
https://github.com/cyberbotics/webots/blob/10b96ec42c8642cd53f7b56ee23c4d3dd7b749e1/resources/languages/java/controller.i#L71-L83

But in the simplest pointer case, it mapped to a generic `[]` here: https://github.com/cyberbotics/webots/blob/10b96ec42c8642cd53f7b56ee23c4d3dd7b749e1/resources/languages/java/controller.i#L84

DavidMansolino:
👀

docs/reference/compass.md
```diff
@@ -178,6 +188,11 @@ double get_bearing_in_degrees() {
 }
 ```
 
+The `wb_compass_get_lookup_table_size` function returns a number of rows in the lookup table.
```
DavidMansolino, line 191:
```suggestion
The `wb_compass_get_lookup_table_size` function returns the number of rows in the lookup table.
```

docs/reference/accelerometer.md
```diff
@@ -149,6 +159,11 @@ Note that the gravity can be specified in the `gravity` field in the [WorldInfo]
 To obtain the acceleration due to motion alone, this offset must be subtracted.
 The device's output will be zero during free fall when no offset is substracted.
 
+The `wb_accelerometer_get_lookup_table_size` function returns the size of the lookup table.
+
+The `wb_accelerometer_get_lookup_table` function returns lookup table fields of the lookup table.
+This function returns a matrix containing exactly N * 3 values (N represents the number of mapped values optained with the `wb_accelerometer_get_lookup_table_size`function) that shall be interpreted as a N x 3 table.
```
DavidMansolino, line 165:
```suggestion
This function returns a matrix containing exactly N * 3 values (N represents the number of mapped values optained with the `wb_accelerometer_get_lookup_table_size` function) that shall be interpreted as a N x 3 table.
```

docs/reference/accelerometer.md
```diff
@@ -149,6 +159,11 @@ Note that the gravity can be specified in the `gravity` field in the [WorldInfo]
 To obtain the acceleration due to motion alone, this offset must be subtracted.
 The device's output will be zero during free fall when no offset is substracted.
 
+The `wb_accelerometer_get_lookup_table_size` function returns the size of the lookup table.
+
+The `wb_accelerometer_get_lookup_table` function returns lookup table fields of the lookup table.
```
DavidMansolino, line 164:
```suggestion
The `wb_accelerometer_get_lookup_table` function returns the values of the lookup table.
```

docs/reference/accelerometer.md
```diff
@@ -149,6 +159,11 @@ Note that the gravity can be specified in the `gravity` field in the [WorldInfo]
 To obtain the acceleration due to motion alone, this offset must be subtracted.
 The device's output will be zero during free fall when no offset is substracted.
 
+The `wb_accelerometer_get_lookup_table_size` function returns the size of the lookup table.
```
DavidMansolino, line 162:
```suggestion
The `wb_accelerometer_get_lookup_table_size` function returns returns the number of rows in the lookup table.
```

docs/reference/accelerometer.md
```diff
@@ -112,6 +120,7 @@ wb_accelerometer_enable(tag, sampling_period)
 wb_accelerometer_disable(tag)
 period = wb_accelerometer_get_sampling_period(tag)
 [x y z] = wb_accelerometer_get_values(tag)
+lookup_table = wb_accelerometer_get_lookup_table(tag)
```
DavidMansolino, line 123: We should somehow show that this is an array (see previous line).

projects/default/controllers/ros/RosAccelerometer.hpp
```diff
@@ -30,11 +33,15 @@ class RosAccelerometer : public RosSensor {
   void rosEnable(int samplingPeriod) override { mAccelerometer->enable(samplingPeriod); }
   void rosDisable() override { cleanup(); }
   int rosSamplingPeriod() override { return mAccelerometer->getSamplingPeriod(); }
+  bool getLookupTableSize(webots_ros::get_int::Request &req, webots_ros::get_int::Response &res);
```
DavidMansolino, line 36:
```suggestion
```

projects/default/controllers/ros/RosCompass.hpp
```diff
@@ -18,6 +18,9 @@
 #include <webots/Compass.hpp>
 #include "RosSensor.hpp"
 
+#include <webots_ros/get_float_array.h>
+#include <webots_ros/get_int.h>
```
DavidMansolino, line 22:
```suggestion
```

projects/default/controllers/ros/RosAccelerometer.hpp
```diff
@@ -18,6 +18,9 @@
 #include <webots/Accelerometer.hpp>
 #include "RosSensor.hpp"
 
+#include <webots_ros/get_float_array.h>
+#include <webots_ros/get_int.h>
```
DavidMansolino, line 22:
```suggestion
```

projects/default/controllers/ros/RosGyro.hpp
```diff
@@ -18,6 +18,9 @@
 #include <webots/Gyro.hpp>
 #include "RosSensor.hpp"
 
+#include <webots_ros/get_float_array.h>
+#include <webots_ros/get_int.h>
```
DavidMansolino, line 22:
```suggestion
```

projects/default/controllers/ros/RosInertialUnit.hpp
```diff
@@ -18,6 +18,9 @@
 #include <webots/InertialUnit.hpp>
 #include "RosSensor.hpp"
 
+#include <webots_ros/get_float_array.h>
+#include <webots_ros/get_int.h>
```
DavidMansolino, line 22:
```suggestion
```

projects/default/controllers/ros/RosLightSensor.hpp
```diff
@@ -18,6 +18,9 @@
 #include <webots/LightSensor.hpp>
 #include "RosSensor.hpp"
 
+#include <webots_ros/get_float_array.h>
+#include <webots_ros/get_int.h>
```
DavidMansolino, line 22:
```suggestion
```

projects/default/controllers/ros/RosTouchSensor.cpp
```diff
@@ -19,8 +19,11 @@
 
 RosTouchSensor::RosTouchSensor(TouchSensor *touchSensor, Ros *ros) : RosSensor(touchSensor->getName(), touchSensor, ros) {
   mTouchSensor = touchSensor;
-  mTypeServer = RosDevice::rosAdvertiseService((ros->name()) + '/' + RosDevice::fixedDeviceName() + "/get_type",
-                                               &RosTouchSensor::getTypeCallback);
+
+  RosDevice::rosAdvertiseService((ros->name()) + '/' + RosDevice::fixedDeviceName() + "/get_type",
```
DavidMansolino, line 23: Is there a reason for modifying this line?

docs/reference/compass.md
```diff
@@ -178,6 +192,11 @@ double get_bearing_in_degrees() {
 }
 ```
 
+The `wb_compass_get_lookup_table_size` function returns the size of the lookup table.
```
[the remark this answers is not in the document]
DavidMansolino: Please apply this fix to all the device pages.

docs/reference/compass.md
```diff
@@ -178,6 +192,11 @@ double get_bearing_in_degrees() {
...
+
+The `wb_compass_get_lookup_table` function returns lookup table fields of the lookup table.
```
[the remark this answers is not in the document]
DavidMansolino: Please apply this fix to all the device pages.

docs/reference/compass.md
```diff
@@ -178,6 +192,11 @@ double get_bearing_in_degrees() {
...
+This function returns a matrix containing exactly N * 3 values (N represents number of mapped values obtained with function `wb_compass_get_lookup_table_size`) that shall be interpreted as a N x 3 table.
```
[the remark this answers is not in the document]
DavidMansolino: Please apply this fix to all the device pages.

projects/default/controllers/ros/RosTouchSensor.cpp
lukicdarkoo: No, it is my mistake

docs/reference/accelerometer.md
lukicdarkoo: I am not sure if I understand. Something like this?
```matlab
lt = wb_accelerometer_get_lookup_table(tag)
lookup_table = array2table([ lt(1:3:end); lt(2:3:end); lt(3:3:end) ]', 'VariableNames', {'Input', 'Output', 'Std'})
```

DavidMansolino: Oh no I was thinking about something much simpler, such as:
`[value0 value1 ...] = wb_accelerometer_get_lookup_table(tag)`

lukicdarkoo: It seems that the example above doesn't work, with:
```
val = wb_accelerometer_get_values(acc)
[x y z] = wb_accelerometer_get_values(acc)
```
I get:
```
Error using wb_accelerometer_get_values
Too many output arguments.
Error in matlab (line 45)
[x y z] = wb_accelerometer_get_values(acc)
Error in launcher (line 161)
  eval(WEBOTS_CONTROLLER_NAME);
val =
    0.0007    0.0078    9.8506
```
So, `val = wb_accelerometer_get_values(acc)` returns correct result, but another line fails.

DavidMansolino: Hum..
In that case it is indeed better to keep it simple, maybe we can just write:
```
lookup_table_array = wb_accelerometer_get_lookup_table(tag)
```
Just to emphasis that this is an array.

lukicdarkoo: Ok, I will do that. But now I intrigued by this:
```
[x y z] = wb_accelerometer_get_values(acc)
```
It seems it doesn't work and it is written in many places. Should I do something about that?

DavidMansolino: Yes if you want to fix it gladly (but please do this in another PR so that it doesn't mix with the changes specific to this PR).
Or if you prefer not to do this right now feel free to open an issue about it and you (or someone else) will fix it later.

omichel: Yes, we should probably fix it by writing something like:
```matlab
x_y_z_array = wb_accelerometer_get_values(acc)
```

DavidMansolino:
👀

Issue #1622, opened 2020-05-04
Warnings when compiling vehicle and manager Python libraries on macOS


DavidMansolino:
**Describe the Bug**
The vehicle and Darwin-op Python wrapper of the libraries is raising some warning when compiled for Python 3.8.

Vehicle:
```
# make python python3.8
In file included from vehicle38.cpp:178:
In file included from /Library/Frameworks/Python.framework/Versions/3.8/include/python3.8/Python.h:85:
In file included from /Library/Frameworks/Python.framework/Versions/3.8/include/python3.8/pytime.h:6:
In file included from /Library/Frameworks/Python.framework/Versions/3.8/include/python3.8/object.h:746:
/Library/Frameworks/Python.framework/Versions/3.8/include/python3.8/cpython/object.h:177:16: warning: 'tp_print' is deprecated [-Wdeprecated-declarations]
typedef struct _typeobject {
               ^
vehicle38.cpp:1767:23: note: in implicit copy assignment operator for '_typeobject' first required here
    swigpyobject_type = tmp;
                      ^
/Library/Frameworks/Python.framework/Versions/3.8/include/python3.8/cpython/object.h:260:5: note: 'tp_print' has been explicitly marked deprecated here
    Py_DEPRECATED(3.8) int (*tp_print)(PyObject *, FILE *, int);
    ^
/Library/Frameworks/Python.framework/Versions/3.8/include/python3.8/pyport.h:515:54: note: expanded from macro 'Py_DEPRECATED'
#define Py_DEPRECATED(VERSION_UNUSED) __attribute__((__deprecated__))
                                                     ^
In file included from vehicle38.cpp:178:
In file included from /Library/Frameworks/Python.framework/Versions/3.8/include/python3.8/Python.h:85:
In file included from /Library/Frameworks/Python.framework/Versions/3.8/include/python3.8/pytime.h:6:
In file included from /Library/Frameworks/Python.framework/Versions/3.8/include/python3.8/object.h:746:
/Library/Frameworks/Python.framework/Versions/3.8/include/python3.8/cpython/object.h:177:16: warning: 'tp_print' is deprecated [-Wdeprecated-declarations]
typedef struct _typeobject {
               ^
/Library/Frameworks/Python.framework/Versions/3.8/include/python3.8/cpython/object.h:260:5: note: 'tp_print' has been explicitly marked deprecated here
    Py_DEPRECATED(3.8) int (*tp_print)(PyObject *, FILE *, int);
    ^
[...]

PR #1633, opened 2020-05-05
Fix macOS Python3.8 compilation warnings.

DavidMansolino:
**Description**
This PR fixes the compilation of the vehicle and Darwin-op Python libraries with Python 3.8.

The warnings were not raised when compiling the libcontroller Python3.8 library because there 
`isystem` was already used to include the Python includes.

**Related Issues**
This pull-request fixes issue #1622


DavidMansolino:
✔️ : travis_release_only

PR #1639, opened 2020-05-06
Fixed Webots crash

omichel:
Because this value was not initialized, Webots was receiving random values here, causing random crashes of Webots (observed on Windows).
However, we should also fix Webots as it should not crash when the controller is sending bad values.

### DIFF
--- a/src/lib/Controller/api/supervisor.c
+++ b/src/lib/Controller/api/supervisor.c
@@ -15,6 +15,7 @@
  */
 
 #include <assert.h>
+#include <float.h>
 #include <math.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -992,17 +993,31 @@ static bool check_field(WbFieldRef f, const char *func, WbFieldType type, bool c
   return true;
 }
 
-static bool checkVector(const char *function, const double values[], int n) {
+static bool check_float(const char *function, double value) {
+  if (isnan(value)) {
+    fprintf(stderr, "Error: %s() called with a NaN value.\n", function);
+    return false;
+  }
+  if (value > FLT_MAX) {
+    fprintf(stderr, "Error: %s() called with a value greater FLX_MAX: %g > %g.\n", function, value, FLT_MAX);
+    return false;
+  }
+  if (value < -FLT_MAX) {
+    fprintf(stderr, "Error: %s() called with a value smaller than -FLX_MAX): %g < %g.\n", function, value, -FLT_MAX);
+    return false;
+  }
+  return true;
+}
+
+static bool check_vector(const char *function, const double values[], int n) {
   if (!values) {
     fprintf(stderr, "Error: %s() called with NULL argument.\n", function);
     return false;
   }
   int i;
   for (i = 0; i < n; i++) {
-    if (isnan(values[i])) {
-      fprintf(stderr, "Error: %s() called with a NaN value.\n", function);
+    if (!check_float(function, values[i]))
       return false;
-    }
   }
   return true;  // ok
 }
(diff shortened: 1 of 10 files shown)


omichel:
Replacing the `0` value with `3.87206e+228` crashes Webots systematically.

omichel:
I would propose to set a maximum double absolute value to object coordinates to avoid this kind of problem: Webots should clamp the coordinate values within [-1e-20 and 1e20] or so, which represents a space way bigger than our solar system, so it should be sufficient for any robotics application and should prevent further ODE crashes.

omichel:
Instead of the 1e+20 arbitrary value, would suggest to use `FLT_MAX` which is 3.402823466e+38.
This can be somehow justified by the fact that our 3D rendering engine uses float and not double, so it would be unable to handle bigger values. FYI `DBL_MAX` value is 1.7976931348623158e+308.

src/lib/Controller/api/supervisor.c
```diff
@@ -1062,16 +1081,16 @@ void wb_supervisor_set_label(int id, const char *text, double x, double y, doubl
 }
 
 void wb_supervisor_export_image(const char *filename, int quality) {
-  if (!robot_check_supervisor("wb_supervisor_export_image"))
+  if (!robot_check_supervisor(__FUNCTION__))
     return;
 
   if (!filename || !filename[0]) {
-    fprintf(stderr, "Error: wb_supervisor_export_image() called with NULL or empty 'filename' argument.\n");
+    fprintf(stderr, "Error: %s() called with NULL or empty 'filename' argument.\n", __FUNCTION__);
```
DavidMansolino, line 1088: Be careful you are doing double-work here. This is already fixed on develop.

omichel: Oops... Too late...

src/webots/maths/WbVector2.hpp
```diff
@@ -129,6 +130,18 @@ class WbVector2 {
...
+  void clamp() {  // clamp to FLT_MAX
+    if (mX > FLT_MAX)
+      mX = FLT_MAX;
+    else if (mX < -FLT_MAX)
+      mX = -FLT_MAX;
+    if (mY > FLT_MAX)
+      mY = FLT_MAX;
+    else if (mY < -FLT_MAX)
+      mY = -FLT_MAX;
+  }
```
DavidMansolino, line 143:
```suggestion
  void clamp(double min = -FLT_MAX, double max = FLT_MAX) {
    if (mX > max)
      mX = max;
    else if (mX < min)
      mX = min;
    if (mY > max)
      mY = max;
    else if (mY < min)
      mY = min;
  }
```

Since we are introducing a clamp format it doesn't cost much to make it generic.

src/webots/maths/WbVector3.hpp
```diff
@@ -150,6 +151,21 @@ class WbVector3 {
...
+      mX = -FLT_MAX;
+    if (mY > FLT_MAX)
+      mY = FLT_MAX;
+    else if (mY < -FLT_MAX)
+      mY = -FLT_MAX;
+    if (mZ > FLT_MAX)
+      mZ = FLT_MAX;
+    else if (mZ < -FLT_MAX)
+      mZ = -FLT_MAX;
+  }
```
DavidMansolino, line 167:
```suggestion
  void clamp(double min = -FLT_MAX, double max = FLT_MAX) {
    if (mX > max)
      mX = max;
    else if (mX < min)
      mX = min;
    if (mY > max)
      mY = max;
    else if (mY < min)
      mY = min;
    if (mZ > max)
      mZ = max;
    else if (mZ < min)
      mZ = min;
  }
```

src/lib/Controller/api/supervisor.c
```diff
@@ -1024,16 +1039,20 @@ void wb_supervisor_set_label(int id, const char *text, double x, double y, doubl
   unsigned int color_and_transparency = (unsigned int)color;
   color_and_transparency += (unsigned int)(transparency * 0xff) << 24;
 
-  if (!robot_check_supervisor("wb_supervisor_set_label"))
+  if (!check_float(__FUNCTION__, x) || !check_float(__FUNCTION__, y) || !check_float(__FUNCTION__, size) ||
+      !check_float(__FUNCTION__, transparency))
```
DavidMansolino, line 1043: These checks don't make sense since the value should be in the range 0 to 1.

Issue #1637, opened 2020-05-06
rostopic echo /......./...../recognition_objects doesnt work error. webots


Mathiebhan:
**Describe the Bug**
Trying to get the topic to work so i can get object position/orientation using object recognition. Which the topic 
/<device_name>/recognition_objects | topic | webots_ros::RecognitionObject
doesnt work. 
Maybe you guys have a solution for it :)

**Screenshots**
I launch my ros node. Then i try to echo the ros topic which gives error as seen in red below in the picture. 
![error](https://user-images.githubusercontent.com/15376018/81188856-2d985800-8fb6-11ea-8672-acc34263295a.PNG)
![Code](https://user-images.githubusercontent.com/15376018/81188863-2f621b80-8fb6-11ea-88e7-c921759c0437.PNG)

**System**
 - Operating System: Tried on Linux Ubuntu 18.04 ros Melodic and ros kinetic 16.04] same issue.

DavidMansolino:
I can indeed reproduce this issue, it seems the md5 sum is wrong on the Webots side.

DavidMansolino:
This has been fixed in #1641, the patch will be available in the next release of Webots.
In the meantime, you can use the nightly builds from tomorrow available here: https://github.com/cyberbotics/webots/releases

PR #1641, opened 2020-05-08
Fix ROS Camera Recognition

DavidMansolino:
**Description**
The camera recognition topic is not working as expected.

**Related Issues**
This pull-request fixes issue #1637


DavidMansolino:
The test if failing due to #1645 but the ros test is passing.

PR #1653, opened 2020-05-11
Add a 'Physics Plugins' Console filter.

DavidMansolino:
**Description**
This PR adds a new 'Physics Plugins' filter for the console.

Capture du 2020-05-11 11-42-29


DavidMansolino:
I will add the use of the new console features in #1559 instead of here.

Issue #1606, opened 2020-04-29
The spot web component has broken pistons


omichel:
Moving the piston motors seems to break the [Spot robot](https://cyberbotics.com/doc/guide/spot) in the user guide.

DavidMansolino:
That's not coming from the robot model as I can confirm the rotation of the piston is correct in Webots.
And it seems the spot is not the only robot concerned, moving the ``wheel1_joint`` motor of the robotino gives also very strange result: https://cyberbotics.com/doc/guide/robotino3

DavidMansolino:
Same with the new P-Rob3, some solids are rotating around the wrong axis: https://cyberbotics.com/doc/guide/p-rob3?version=feature-prob3

Issue #1365, opened 2020-02-14
Implement JointLink


omichel:
In order to be more scalable than #1359 and include non-motorized joints, the following design proposes joint links allowing to couple motorized and non-motorized joints.

We should implement the `JointLink` node as:

```
JointLink {
  field SFString endPointName ""    # name of the Solid end point of the linked joint
  field SFFloat multiplier     1    # gear/pulley ratio (may be negative to reverse motion)
  field SFFloat gain        0.95    # energy gain (should be < 1 for purely mechanical system)
  field SFInt32{1, 2, 3} jointId 1  # joint identifier
}
```

And we should add link fields to all joint types:

```
HingeJoint {
  field MFNode link [ ]
}

Hinge2Joint {
  field MFNode link  [ ]
  field MFNode link2 [ ]
}

BallJoint {
  field MFNode link  [ ]
  field MFNode link2 [ ]
  field MFNode link3 [ ]
}

SliderJoint {
  field MFNode link [ ]
}
```

Example usage:

```
Hinge2Joint {
  link2 [
    JointLink {
      endPointName "arm"
      jointId 1
    }
  ]
}

Hinge2Joint {
  endPoint Solid {
    name "arm"
  }
}
```

In the above example, the second axis of the first hinge2 joint is coupled to the first axis of the second hinge2 joint.

DavidMansolino:
There is a small issue about the example:
```
Hinge2Joint {
  link2 [
    JointLink {
      endPointName "arm"
      jointId 1
    }
  ]
}

Hinge2Joint {
  endPoint Solid {
    name "arm"
  }
}
```

DavidMansolino:
Once implemented, this should be used here: https://github.com/cyberbotics/webots/blob/403fa190883e988ae50ba5970bfcca0e344b24a2/projects/robots/boston_dynamics/spot/protos/SpotLeftLeg.proto#L58

DavidMansolino:
And here:
https://github.com/cyberbotics/webots/blob/017d6f51745a82235e1eefaa407a4c835ab2cffe/projects/robots/fp_robotics/p-rob3/protos/P-Grip.proto#L283

tlemarec:
My team and I would really appreciate this feature. Any news on the progress ?

omichel:
[...]

PR #1657, opened 2020-05-12
P-Rob3 Model

DavidMansolino:
**Description**
This PR adds a model of the P-Rob 3 robotic arm.

**Tasks**
Remaining tasks:
  - [x] Bounding objects.
  - [x] Physics.
  - [x] PROTOization.
  - [x] Gripper
    - [x] Add specific fingers.
    - [x] Move the gripper out of the PROTO.
    - [x] Create a specific gripper PROTO.
    - [x] Icon.
    - [x] doc update.
  - [x] Sample world.
  - [x] Sample controller.
  - [x] Texture integration.
  - [x] Icon.
  - [x] Documentation.
  - [x] Changelog.
  - [x] Remaining bug / questions:
    - [x] Distance sensors.
      - [x] Calibration.
      - [x] Noise => +- 0.2%
    - [x] Motor torque limit too low.
    - [x] Finger limit.
    - [x] Finger torque limit => gear ratio 100:1
    - [x] Speed/acceleration expressed before or after redactor?
    - [x] Add button.
      - [x] Textures.
  - [x] Final regeneration of the robot component.

**Documentation**
https://cyberbotics.com/doc/guide/p-rob3?version=feature-prob3

**Screenshots**

prob3
prob3_1


DavidMansolino:
**Status**: the model is ready, remains the sample simulation (world + controller)

DavidMansolino:
:heavy_check_mark:: travis_release

DavidMansolino:
> I believe the default rotation should be 0 and along the Z axis.

:+1:
We should probably create a doc page to explain what is the recommended orientation for each type of robot.

> The 3D component in the doc is broken as soon as you move the second axis (and others), as with other robots...

This is unfortunately already reported here: https://github.com/cyberbotics/webots/issues/1606#issuecomment-628011610

> Why did you call a finger "7" and the other "7 left". Why not "7 right" and "7 left"?

The reason behind is that the actual name of the joint is `7` for both (which is not yet possible in Webots, but will soon once #1365 is implemented). So I tried to have it already very close to what it will be in the end (so that the only thing to change in the controller will be to remove the use of `7 left` but not touching the use of `7`. But maybe instead of `7 left`, something like `7-linked` would be more clean?

omichel:
Thank you for the explanations. Then, I believe that we should adopt some formalism for linked joint names. "left" is useful to indicate which finger it is, but "linked" is also useful, so I would suggest something like "[link] 7 (left)". But maybe since linked motors are not supposed to be controlled directly (it shouldn't be possible to get a handle to a linked device from a controller), we could consider leaving the name field empty and put rather in the DEF name some useful information (like LEFT_FINGER) for the developer looking into the model.

DavidMansolino:
From what I understood from #1365, once the `JointLink` will be implemented, these `linked` motors will completely disappear, so this formalism will become useless, this is just for the transition period (which I expect to be very short as I am planning after this PR (when I will have time) to followup on #1365).

omichel:
Oh, yes, you are right. So, the name is not important as it will disappear soon.

DavidMansolino:
> About the world name, I would maybe rename it to p-rob3.wbt to be more consistent with the proto name.

:+1: 

>The mesh seems to display some strange effects like this one:
(but that's not the only one)
I guess this is due to bad triangles in the mesh?

Yes, this is due to triangles for which is crease angle is not able to smooth the edge (this is particularly visible for rough part), I fixed the most visible ones but some less visible are remaining.
Unfortunately, the remaining ones are quite difficult to fix mainly because they are not visible in blender because of the different smoothing algorithm.

DavidMansolino:
I confirm this is not a texture mapping issue.
Without texture mapping:
p-rob3
And without texture:
p-rob3_1

PR #1661, opened 2020-05-14
Fix packaging

DavidMansolino:
**Description**
Fix Webots packaging broken in #1420


DavidMansolino:
✔️: travis_relase_only

PR #1663, opened 2020-05-14
Fix cppcheck 2.0 errors

omichel:
The source tests failed last night due to an upgrade of cppcheck from version 1.9 to version 2.0 which detects more problems.
This PR fixes the source code of this repository to pass the cppcheck-2.0 tests.

### DIFF
--- a/projects/default/controllers/ros/RosSupervisor.cpp
+++ b/projects/default/controllers/ros/RosSupervisor.cpp
@@ -449,7 +449,7 @@ bool RosSupervisor::virtualRealityHeadsetIsUsedCallback(webots_ros::get_bool::Re
   return true;
 }
 
-bool RosSupervisor::nodeGetIdCallback(webots_ros::node_get_id::Request &req, webots_ros::node_get_id::Response &res) {
+bool RosSupervisor::nodeGetIdCallback(const webots_ros::node_get_id::Request &req, webots_ros::node_get_id::Response &res) {
   assert(this);
   if (!req.node)
     return false;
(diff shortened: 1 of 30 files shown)


omichel:
✔️ appveyor [source test](https://ci.appveyor.com/project/omichel/webots/builds/32868383) passed.

DavidMansolino:
👀

projects/default/controllers/ros/RosSupervisor.cpp
```diff
@@ -449,6 +449,7 @@ bool RosSupervisor::virtualRealityHeadsetIsUsedCallback(webots_ros::get_bool::Re
   return true;
 }
 
+// cppcheck-suppress constParameter
```
DavidMansolino, line 452: It is strange that the warning appears only to those few functions and not all the others. Do you have an explanation for this?

resources/projects/libraries/qt_utils/motion_editor/Motion.cpp
```diff
@@ -307,12 +307,8 @@ void Motion::newPoseAt(int index) {
...
 
     if (!hasFixedStep()) {
-      Pose *previousPose = NULL, *nextPose = NULL;
-      if (index > 0)
-        previousPose = mPoses.at(index - 1);
-      if (index < mPoses.count())
-        nextPose = mPoses.at(index);
-
+      Pose *previousPose = index > 0 ? mPoses.at(index - 1) : NULL;
+      Pose *nextPose = index < mPoses.count() ? mPoses.at(index) : NULL;
```
DavidMansolino, line 311:
```suggestion
      const Pose *previousPose = index > 0 ? mPoses.at(index - 1) : NULL;
      const Pose *nextPose = index < mPoses.count() ? mPoses.at(index) : NULL;
```

projects/default/controllers/ros/RosSupervisor.cpp
omichel: No idea. I guess it is unable to detect that in some case the *Request* object can be modified. Maybe this *Request* has some non-standard method modifying it?

DavidMansolino: > Maybe this Request has some non-standard method modifying it?

No idea what the ROS libraries are doing behind, but what looks strange is that it works exactly the same way as for the other callback so I was wondering why the error was raised only here. But that was just by curiosity.

omichel: It's not always the very same *Request* object (different namespace), that might explain why it behaves differently.

Issue #1677, opened 2020-05-19
Mac R2020b Broken


DavidMansolino:
**Describe the Bug**
A user reported that the R2020b is currently broken on Mac:
```
Dyld Error Message:
  Library not loaded: @rpath/lib/webots/libIrrXML.dylib
  Referenced from: /Applications/Webots.app/lib/webots/libassimp.dylib
  Reason: image not found
```

PR #1679, opened 2020-05-19
Fix MacOS Packaging

DavidMansolino:
Fixes issue #1677.

### DIFF
--- a/src/packaging/webots_distro.c
+++ b/src/packaging/webots_distro.c
@@ -836,8 +836,6 @@ static void create_file(const char *name, int m) {
       fprintf(fd, "cd \"%s/%s/lib/webots\"\n", distribution_path, bundle_name);
       fprintf(fd, "ln -s libssl.1.0.0.dylib libssl.dylib\n");
       fprintf(fd, "ln -s libcrypto.1.0.0.dylib libcrypto.dylib\n");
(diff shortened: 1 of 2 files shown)


DavidMansolino:
✔️: trvis_release_only

src/packaging/webots_distro.c
```diff
@@ -836,8 +836,6 @@ static void create_file(const char *name, int m) {
       fprintf(fd, "cd \"%s/%s/lib/webots\"\n", distribution_path, bundle_name);
       fprintf(fd, "ln -s libssl.1.0.0.dylib libssl.dylib\n");
       fprintf(fd, "ln -s libcrypto.1.0.0.dylib libcrypto.dylib\n");
-      fprintf(fd, "ln -s libssh.4.dylib libssh.dylib\n");
-      fprintf(fd, "ln -s libzip.2.dylib libzip.dylib\n");
```
omichel, line 840: Don't we need these libraries for the Darwin-OP robot window?

omichel: Oh yes, I see, they are installed locally...

DavidMansolino: Exactly, actually here this command was just creating a symlink to a file not existing 😄

PR #1689, opened 2020-05-22
Fix appveyor pacman update

omichel:
Since a couple of days, appveyor build are broken, originally due to a general update problem and now due to a [pacman update problem](https://github.com/msys2/MSYS2-packages/issues/1967). Since it is not sure the problem will be quickly fixed on the MSYS2 side, this PR aims at patching it.


omichel:
The problem seems to be fixed: https://ci.appveyor.com/project/omichel/webots-9ufow/builds/33047447

omichel:
The problem seems to be fixed: https://ci.appveyor.com/project/omichel/webots-9ufow/builds/33048488

PR #1706, opened 2020-05-27
Allow Basic URDF Export

### LINKED ISSUE #1603: Missing Device Fields Required for Automatic ROSification

lukicdarkoo:
**Description**
According to the discussion, we should rather convert the node structure to URDF and publish transforms through `robot_state_publisher`. Therefore, this PR aims to allow export of stripped URDF version for `robot_state_publisher`. Extended, but still stripped version will be allowed later for RViz visualization.

URDF + `robot_state_publisher` vs publishing TF messages directly:
- Having URDF is much more flexible:
  - can be used to publish TF messages with `robot_state_publisher`,
  - can be used to visualize in RViz2,
  - can be used with MoveIt,
  - transfer to real-robot is much faster (just copy-paste generated URDF file),
  - part of maintenance (e.g. `robot_state_publisher`) is done by ROS community and
  - it makes the implementation of TF messages easier for ROS1 (just publish joint states)
- Publishing TF messages directly is faster and simpler to use:
  - no overhead caused by `robot_state_publisher` (mostly caused by exchanging the messages),
  - no need to start another process `robot_state_publisher`.

**Related Issues**
Closes #1603. Replaces #1610

**Tasks**
- [x] Determine what part of URDF `<link />` is relevant. (A) None, only joints are relevant for transformations. `<link />` should just exist as placeholders.
- [x] Determine mappings from Webots joints to URDF `<joint />`. (A) `WbHinge` --> `<joint type="continious" />`, `WbSlider` --> `<joint type="prismatic" />` and link-to-link --> `<joint type="fixed" />`.
- [x] Create converter for joints:
  - [x] link-to-link --> `<joint type="fixed" />` (only where needed)
  - [x] `WbHinge` --> `<joint type="continious" />`
  - [x] `WbSlider` --> `<joint type="prismatic" />`
- [x] The tree is currently huge, we don't need everything, ignore irrelevant nodes:
  - [x] Children of the last node that has `name` parameter (this may be bad later for exporting bounding objects)
  - [x] Physics
  - [x] Squash all, non-device, solid-to-solid nodes under the same `<link />`
- [x] Add API support
  - [x] C
  - [x] C++
  - [x] Python
  - [x] Java
  - [x] MATLAB (not tested)
  - [x] ROS
  - [x] Other
- [x] Documentation (https://cyberbotics.com/doc/reference/robot?version=feature-urdf-export&tab-language=c#wb_robot_get_urdf)
- [x] Unit tests
- [x] Improve URDF formatting
  - [x] Avoid new lines when unnecessary
  - [x] Fix NULL artifacts in URDF file
- [x] Verify why adding a tool to UR5 crashes the simulation 
- [x] Test with a few models
  - [x] E-Puck (because it has hinge joints without limits)
  - [x] E-Puck + ROS2 
  - [x] UR5e (because it has hinge joints with limits)
  - [x] UR5e + ROS2 
  - [x] Tiago++ (because it has slider joint)
  - [x] UR5e + tool (to test if slot is handled properly)
  - [x] UR5e + tool + ROS2 

**Additional context**
Useful commands:
```
./webots --minimize > /tmp/test.xml
check_urdf /tmp/test.xml
urdf_to_graphiz /tmp/test.xml
ros2 run robot_state_publisher robot_state_publisher /tmp/test.xml
```
(`/tmp/test.xml` is URDF file, `.xml` extension is temporary to benefit from the syntax highlighters)


lukicdarkoo:
URDF output is valid (although not properly formatted):
```
$ check_urdf /tmp/test.xml
robot name is: e-puck
---------- Successfully Parsed XML ---------------
root Link: base_link has 34 child(ren)
    child(1):  group_322
        child(1):  transform_323
...
```
so we can generate URDF graphs:
```
$ urdf_to_graphiz /tmp/test.xml
Created file e-puck.gv
Created file e-puck.pdf
```

If we start `robot_state_publisher` with file:
```
$ ros2 run robot_state_publisher robot_state_publisher /tmp/test.xml
Parsing robot urdf xml string.
Link group_322 had 2 children
Link transform_323 had 1 children
...
```
and run RViz2 we get:
image
(positions of e-puck's distance sensors)

Currently, only fixed transforms work fine. It is aware of joints, but I have to fix origins, and I have to make a relation between joint and it's built-in PositionSensor.

DavidMansolino:
Well done, this looks very promising!

lukicdarkoo:
Peek 2020-06-01 21-16
It works now with joints as well!

src/lib/Controller/api/messages.h
```diff
@@ -46,6 +46,7 @@
 #define C_SUPERVISOR_MOVIE_STATUS 166
 
 // ctr <-> sim
+#define C_ROBOT_URDF 108
```
DavidMansolino, line 49: This should be moved right after `C_ROBOT_SET_JOYSTICK_FORCE_AXIS`.

src/webots/Makefile
```diff
@@ -33,7 +33,7 @@ WB_MATHS_INCLUDE         = $(QT_CORE_INCLUDE)
 WB_CORE_INCLUDE          = $(QT_CORE_INCLUDE) $(QT_NETWORK_INCLUDE) $(QT_GUI_INCLUDE) $(LUA_INCLUDE) -Iexternal/compilation_timestamp
 WB_UTIL_INCLUDE          = $(QT_CORE_INCLUDE) $(ODE_INCLUDE) -Icore
 WB_WREN_INCLUDE          = $(QT_CORE_INCLUDE) $(QT_GUI_INCLUDE) $(QT_OPENGL_INCLUDE) $(WREN_INCLUDE) $(GLU_INCLUDE) -Icore -Imaths -Ivrml -Iutil
-WB_VRML_INCLUDE          = $(QT_CORE_INCLUDE) -Icore -Imaths -Ivrml
+WB_VRML_INCLUDE          = $(QT_CORE_INCLUDE) -Icore -Imaths -Ivrml -Inodes -Iode -Inodes/utils
```
DavidMansolino, line 36: I know this is very tempting, but should be avoided as much as possible, we try to avoid introducing dependencies on high-level modules for low-level modules.
Most of the time this can be avoided using the signal-slot mechanism or inheritance.

src/webots/nodes/WbJoint.cpp
```diff
@@ -292,3 +292,34 @@ void WbJoint::updateJointAxisRepresentation() {
...
+  if (positionSensor() && positionSensor()->findSFString("name"))
+    return positionSensor()->findSFString("name")->value();
+  return WbBaseNode::urdfName();
+}
+
+void WbJoint::writeExport(WbVrmlWriter &writer) const {
+  if (writer.isUrdf() && solidEndPoint()) {
+    WbNode *parentRoot = findUrdfLinkRoot();
+    WbVector3 translation = solidEndPoint()->translationFrom(parentRoot);
+    WbVector3 rotationEuler = solidEndPoint()->rotationMatrixFrom(parentRoot).toEulerAngles();
```
DavidMansolino, line 306: This deserves a few const.

src/webots/nodes/WbJoint.cpp
```diff
@@ -292,3 +292,34 @@ void WbJoint::updateJointAxisRepresentation() {
...
+    writer << QString("    <child link=\"%1\"/>\n").arg(solidEndPoint()->urdfName());
+    writer << QString("    <axis xyz=\"%1 %2 %3\"/>\n").arg((int)axis().x()).arg((int)axis().y()).arg((int)axis().z());
+    writer << QString("    <origin xyz=\"%1 %2 %3\" rpy=\"%4 %5 %6\" />\n")
+                .arg(translation.x())
+                .arg(translation.y())
+                .arg(translation.z())
+                .arg(rotationEuler.x())
+                .arg(rotationEuler.y())
+                .arg(rotationEuler.z());
+    writer << QString("  </joint>\n");
```
DavidMansolino, line 319: Instead of manually adding spaces to indent, the `WbVrmlWriter` has these two functions:
```
void increaseIndent()
void decreaseIndent()
```

src/webots/nodes/WbRobot.cpp
```diff
@@ -54,6 +54,7 @@
 #include <QtCore/QStringList>
 #include <QtCore/QTimer>
 
+#include <iostream>
```
DavidMansolino, line 57: I assume this is for debugging?

src/webots/nodes/WbRobot.cpp
```diff
@@ -1127,6 +1132,23 @@ void WbRobot::writeAnswer(QDataStream &stream) {
...
     mSupervisorUtilities->writeAnswer(stream);
+
+  if (mNeedToWriteUrdf) {
+    stream << (short unsigned int)0;
+    stream << (unsigned char)C_ROBOT_URDF;
+
+    QString urdfContent;
+    WbVrmlWriter writer(&urdfContent, modelName() + ".urdf");
+
+    writer.writeHeader(this->findSFString("name")->value());
```
DavidMansolino, line 1143:
```suggestion
    writer.writeHeader(name());
```

src/webots/nodes/WbTransform.cpp
```diff
@@ -406,3 +406,39 @@ QStringList WbTransform::fieldsToSynchronizeWithX3D() const {
...
+
+WbVector3 WbTransform::translationFrom(WbNode *fromNode) const {
+  const WbNode *parentNode = parent();
+  const WbNode *childNode = this;
+  WbVector3 translationResult;
+
+  while (parentNode != fromNode) {
+    WbVector3 translation;
+    if (dynamic_cast<const WbTransform *>(childNode))
+      translation = static_cast<const WbTransform *>(childNode)->translation();
```
DavidMansolino, line 418:
```suggestion
    const WbTransform *transform = dynamic_cast<const WbTransform *>(childNode);
    if (transform)
      translation = transform->translation();
```

src/webots/nodes/WbTransform.cpp
```diff
@@ -406,3 +406,39 @@ QStringList WbTransform::fieldsToSynchronizeWithX3D() const {
...
+    translationResult += translation;
+
+    childNode = parentNode;
+    parentNode = parentNode->parent();
```
DavidMansolino, line 422: The `WbNodeUtilities::findUpperTransform(const WbNode *node)` function might simplify this function.

src/webots/nodes/WbTransform.cpp
```diff
@@ -406,3 +406,39 @@ QStringList WbTransform::fieldsToSynchronizeWithX3D() const {
...
+  }
+
+  return translationResult;
+}
+
+WbMatrix3 WbTransform::rotationMatrixFrom(WbNode *fromNode) const {
```
DavidMansolino, line 428: Same comment than for translation.

src/webots/vrml/WbNode.cpp
```diff
@@ -972,13 +977,45 @@ void WbNode::writeParameters(WbVrmlWriter &writer) const {
     parameter->write(writer);
 }
 
+bool WbNode::isUrdfLinkRoot() const {
+  if (findSFString("name") || dynamic_cast<WbBasicJoint *>(parent()))
+    return true;
+  return false;
+}
```
DavidMansolino, line 984: Here for example to avoid having to include WbBasicJoint you can make this function virtual and return false.
Then in WbMatter and  WbBasicJoint you make it returns true.

src/webots/vrml/WbNode.cpp
```diff
@@ -1060,9 +1097,42 @@ QStringList WbNode::listTextureFiles() const {
...
+      gUrdfNodesQueue.append(this);
+      return true;
+    }
+
+    writer << "    <visual name=\"" << urdfName() << "\">\n";
+    writer << "       <geometry>\n";
+    writer << "         <box size=\"0.01 0.01 0.01\"/>\n";
+    writer << "       </geometry>\n";
+    writer << "    </visual>\n";
+    return false;
```
DavidMansolino, line 1134: same comment about increase/decrease indent.

src/webots/vrml/WbNode.cpp
```diff
@@ -1080,34 +1150,64 @@ bool WbNode::exportNodeHeader(WbVrmlWriter &writer) const {
 }
 
 void WbNode::exportNodeFields(WbVrmlWriter &writer) const {
-  foreach (WbField *field, fields())
-    if (!field->isDeprecated() && ((field->isVrml() || writer.isProto()) && field->singleType() != WB_SF_NODE))
-      field->write(writer);
+  if (!writer.isUrdf())
+    foreach (WbField *field, fields())
+      if (!field->isDeprecated() && ((field->isVrml() || writer.isProto()) && field->singleType() != WB_SF_NODE))
+        field->write(writer);
```
DavidMansolino, line 1156: Only the last inner level doesn't require brackets: https://github.com/cyberbotics/webots/wiki/CPP-Coding-Style#cs360-single-if-while-or-for-statement-must-be-written-without-brackets-the-statement-part-should-be-put-on-a-separate-line-in-case-of-nested-block-keep-brackets

src/webots/vrml/WbNode.cpp
```diff
@@ -1080,34 +1150,64 @@ bool WbNode::exportNodeHeader(WbVrmlWriter &writer) const {
...
+    if (!field->isDeprecated() &&
+        ((field->isVrml() || writer.isProto() || writer.isUrdf()) && field->singleType() == WB_SF_NODE)) {
       const WbSFNode *const node = dynamic_cast<WbSFNode *>(field->value());
-      if (node == NULL || node->value() == NULL || node->value()->shallExport() || writer.isProto()) {
-        if (writer.isX3d())
+      if (node == NULL || node->value() == NULL || node->value()->shallExport() || writer.isProto() || writer.isUrdf()) {
+        if (writer.isX3d() || writer.isUrdf()) {
           field->value()->write(writer);
-        else
+        } else
```
DavidMansolino, line 1167: These brackets are not mandatory.

lukicdarkoo:
> As discussed yesterday, here is a first high level review (some comments are not so high at all, it's just details I say while reading the code).
> 
> Globally it looks very good, the only 'design' issue I saw is the changes to the makefile (see comment).
> 
> I also tested it and it seems to work very well. I am just wondering if we should really name the root link `base_link` in all cases (isn't this going to cause conflicts when we use more than one robot?). But this is probably the norm in ROS.

Thanks for the review! I will try to avoid modifying the Makefile. I wanted to avoid it in the beginning, but it was indeed very tempting.

Regarding the `base_link`. I have been working on the integration of `.get_urdf()` to `webots_ros2` and there `robot_state_publisher` should also be able to work with multiple robots. I will try to utilise namespaces for that. If not, we can root link name as the parameter to `.get_urdf()`.

lukicdarkoo:
It seems there is no good solution for multiple robots. The namespaces put all topics, node names etc, under the namespace, but transforms are excluded.

Before, 'tf_prefix' used to exist, but now it is deprecated:
http://wiki.ros.org/tf2/Migration#Removal_of_support_for_tf_prefix

There is an explanation here on how to solve the issue:
https://github.com/ros2/robot_state_publisher/issues/35#issuecomment-538544615
but it involves creating another node which would republish transforms to root `/tf`

In general, it seems it is not simple as adding a prefix to the transforms. However, for our examples, we can simply allow users to add a prefix to all transforms. Therefore, it would be something like  `robot.getUrdf('my_prefix')` and `robot_state_publisher` would publish proper transforms.

DavidMansolino:
> In general, it seems it is not simple as adding a prefix to the transforms. However, for our examples, we can simply allow users to add a prefix to all transforms. Therefore, it would be something like robot.getUrdf('my_prefix') and robot_state_publisher would publish proper transforms.

That seems like a good solution, adding a prefix (empty by default) sounds good.

src/webots/nodes/WbTransform.cpp
lukicdarkoo: Nice, it works!

src/webots/nodes/WbJoint.cpp
lukicdarkoo: It is a bit messy now, but I guess it is easier to extend and change indentation size

src/webots/nodes/WbRobot.cpp
lukicdarkoo: It was, forgot to delete, I discovered `QTextStream(stdout) <<` is much better for most of the cases.

src/webots/nodes/WbJoint.cpp
DavidMansolino: Yes indeed, using the increaseIndent/decreaseIndent adds more lines, but is is simpler for the long term maintenance and avoid having to count exactly the number of space for each line.

src/webots/vrml/WbNode.cpp
lukicdarkoo: I put it in `WbBaseNode`, I think it is more general this way.

lukicdarkoo:
Peek 2020-06-09 16-08
Tiago++ demo (linear motors also work)

DavidMansolino:
:eyes:

src/webots/vrml/WbNode.hpp
```diff
@@ -308,9 +310,16 @@ class WbNode : public QObject {
   virtual void exportNodeSubNodes(WbVrmlWriter &writer) const;
   virtual void exportNodeFooter(WbVrmlWriter &writer) const;
 
+  WbNode *findUrdfLinkRoot() const;
+  virtual bool isUrdfLinkRoot() const;
```
DavidMansolino, line 314:
```suggestion
  WbNode *findUrdfRootLink() const;
  virtual bool isUrdfRootLink() const;
```

tests/api/controllers/robot_urdf/robot_urdf.c
```diff
@@ -0,0 +1,38 @@
...
+
+  // Write URDF to file
+  FILE *f;
+  f = fopen("robot.urdf", "w");
+  const char *urdf = wb_robot_get_urdf("");
+  fprintf(f, "%s", urdf);
+  fclose(f);
+
+  // Save output of `check_urdf` to file
+  int urdf_check_status = system("check_urdf robot.urdf > result.txt");
```
DavidMansolino, line 24: This is dangerous as it has to be installed separately.
I would rather make this part of the test optional (only if this command is available), and add a second test that compares the result of the export with a know version (committed as an urdf file for example).

src/webots/nodes/WbBaseNode.cpp
```diff
@@ -235,28 +236,70 @@ bool WbBaseNode::isInvisibleNode() const {
...
+      writer
+        << QString(" docUrl=\'%1/doc/%2/%3\'").arg(WbStandardPaths::cyberboticsUrl()).arg(bookAndPage[0]).arg(bookAndPage[1]);
+
+    if (isUseNode() && defNode()) {  // export referred DEF node id
+      writer << " USE=\'n" + QString::number(defNode()->uniqueId()) + "\'></" + x3dName() + ">";
+      return true;
+    }
+  } else {
     return WbNode::exportNodeHeader(writer);
+  }
```
DavidMansolino, line 260: Useless brackets.

src/webots/vrml/WbNode.cpp
```diff
@@ -972,13 +974,43 @@ void WbNode::writeParameters(WbVrmlWriter &writer) const {
...
 }
 
+bool WbNode::isUrdfLinkRoot() const {
+  return findSFString("name") ? true : false;
+}
+
+WbNode *WbNode::findUrdfLinkRoot() const {
+  WbNode *parentRoot = parent();
+  while (!parentRoot->isUrdfLinkRoot())
+    parentRoot = parentRoot->parent();
```
DavidMansolino, line 984: That looks dangerous, for example, if you call `Transform::findUrdfLinkRoot` you will end up in an infinite loop.
We should add an exit condition (and return NULL in that case).

src/webots/vrml/WbNode.cpp
```diff
@@ -972,13 +974,43 @@ void WbNode::writeParameters(WbVrmlWriter &writer) const {
...
+
 void WbNode::write(WbVrmlWriter &writer) const {
   if (uniqueId() == -1) {
     if (nodeModelName() == "Plane" || nodeModelName() == "Capsule") {
       WbNodeFactory::instance()->exportAsVrml(this, writer);
       return;
     }
   }
+  if (writer.isUrdf()) {
+    if (gUrdfCurrentNode != this && nodeModelName().contains("Joint") && !gUrdfNodesQueue.contains(this)) {
```
DavidMansolino, line 996: > nodeModelName().contains("Joint")

That looks dangerous (if for example in the future we add a new join who doesn't contain the 'Joint' keyword, we might miss it here without noticing it). But I don't see any better solution :-/

src/webots/vrml/WbNode.cpp
```diff
@@ -1060,9 +1092,58 @@ QStringList WbNode::listTextureFiles() const {
   return list;
 }
 
+const QString WbNode::urdfName() const {
+  QString name;
+  if (this->findSFString("name")) {
```
DavidMansolino, line 1097:
```suggestion
  if (isUrdfLinkRoot()) {
```

src/webots/vrml/WbNode.cpp
```diff
@@ -1080,28 +1161,39 @@ bool WbNode::exportNodeHeader(WbVrmlWriter &writer) const {
...
 void WbNode::exportNodeFields(WbVrmlWriter &writer) const {
-  foreach (WbField *field, fields())
-    if (!field->isDeprecated() && ((field->isVrml() || writer.isProto()) && field->singleType() != WB_SF_NODE))
-      field->write(writer);
+  if (!writer.isUrdf()) {
+    foreach (WbField *field, fields()) {
+      if (!field->isDeprecated() && ((field->isVrml() || writer.isProto()) && field->singleType() != WB_SF_NODE))
+        field->write(writer);
+    }
+  }
```
DavidMansolino, line 1169:
```suggestion
  if (writer.isUrdf())
    return;

  foreach (WbField *field, fields()) {
    if (!field->isDeprecated() && ((field->isVrml() || writer.isProto()) && field->singleType() != WB_SF_NODE))
      field->write(writer);
  }
```

We tend to use return as soon as possible instead of encapsulating everything in a condition.

projects/default/controllers/ros/Ros.cpp
```diff
@@ -477,6 +479,12 @@ bool Ros::getModelCallback(webots_ros::get_string::Request &req, webots_ros::get
   return true;
 }
 
+bool Ros::getUrdfCallback(webots_ros::get_string::Request &req, webots_ros::get_string::Response &res) {
```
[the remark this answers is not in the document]
lukicdarkoo: From some reason, I considered prefix non-important, but we can definitely add new .srv for that

src/webots/nodes/WbBaseNode.cpp
```diff
@@ -235,28 +236,70 @@ bool WbBaseNode::isInvisibleNode() const {
...
+  return QString();
+}
+
 bool WbBaseNode::exportNodeHeader(WbVrmlWriter &writer) const {
-  if (!writer.isX3d())
+  if (writer.isX3d()) {
+    writer << "<" << x3dName() << " id=\'n" << QString::number(uniqueId()) << "\'";
+    if (isInvisibleNode())
+      writer << " render=\'false\'";
+    QStringList bookAndPage = documentationBookAndPage(WbNodeUtilities::isRobotTypeName(nodeModelName()));
```
stefaniapedrazzi, line 249: Sorry but I don't understand the changes in this function.
It seems that only this line (and the following) is changed in the whole function.
But then why did you also change the if block structure if it behaves exactly like before?

This is really subjective topic, but I find that previous version returning immediately in case of `!writer.isX3d()` is more readable than having to search for the else block.

src/webots/nodes/WbBaseNode.cpp
```diff
@@ -235,28 +236,70 @@ bool WbBaseNode::isInvisibleNode() const {
   return WbWorld::instance()->viewpoint()->getInvisibleNodes().contains(const_cast<WbBaseNode *>(this));
 }
 
+QString WbBaseNode::documentationUrl() const {
+  // TODO: Hotfix, undefined method
+  return QString();
+}
```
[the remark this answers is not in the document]
lukicdarkoo: I pulled this from `develop`, it didn't want to compile, so I have just defined it. I am not sure what to do with this since it is not related to this PR

docs/reference/robot.md
```diff
@@ -1964,6 +1964,95 @@ Users unfamiliar with the mutex concept may wish to consult a reference on multi
...
+There are certain rules that are followed to create an efficient output.
+Webots nodes are squashed into a single URDF link node whenever is possible to simplify the exported robot model.
+In case you want Webots node to be shown as a separate URDF link it is enough to define `name` field.
+URDF links inherit `name` field from Webots node except when there is no `name` field defined or there are two or more Webots nodes with the name.
+URDF joints are named after position sensor in the corresponding Webots joints.
+
+
+
+> **note** Exported URDF is not complete. 
+Currently, the generated URDF consists of a minimal number of elements to be used with [`robot_state_publisher`](http://wiki.ros.org/robot_state_publisher). [RViz](http://wiki.ros.org/rviz) will show only small boxes in places where the links are located.
```
DavidMansolino, line 2052: We should mention that visual geometries are exported in any case as boxes.

src/webots/nodes/WbBaseNode.cpp
lukicdarkoo: You are right, there was no reason for me to change this function. I think that I added introduced changes at some point related to URDF, reverted back but I wasn't aware I made unnecessary changes in comparison to `develop` branch.

tests/api/controllers/robot_urdf/robot_urdf.c
lukicdarkoo: It supposed to come with ROS installation which is installed for the ROS tests.

I understand, but it is a bit tricky as flexibility compromised. Robot model might change (which we can solve by creating a custom robot model just for this purpose) and making a slight change like changing naming convention or adding more detailed visual elements will cause the test to fail.

Ideally, we should parse the exported URDF and than verify different aspects individually, but that is a lot of work (already done by `check_urdf`).

DavidMansolino: I agree, the solution I proposed will need update each time we change the exporter, but it will at least detect cases where we change it not on purpose.

The check_urdf is indeed the ideal solution, but unfortunately, we want to have the test suite to run on all the platforms whereas ROS is not always installed (we are for example using it on Windows too), so maybe checking it check_urdf is available and use it only in this case is good (it will be for example available on Travis where we run the nightly tests).

src/webots/vrml/WbNode.cpp
lukicdarkoo: I am sorry for this and similar code style mistakes, we were often forced to use [MISRA C](https://en.wikipedia.org/wiki/MISRA_C) (e.g. function has to have only one return statement) and the old habits stick :)

DavidMansolino: Don't worry, I know it is not easy to switch from one coding style to another one.

src/webots/nodes/WbGeometry.cpp
```diff
@@ -524,6 +524,12 @@ int WbGeometry::triangleCount() const {
     return 0;
 }
 
+bool WbGeometry::exportNodeHeader(WbVrmlWriter &writer) const {
+  if (writer.isUrdf())
+    return true;
+  return WbNode::exportNodeHeader(writer);
```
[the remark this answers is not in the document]
lukicdarkoo: No reason, both have the same effect at this point and I didn't know what practice is better. Now I know, thanks :)

src/webots/nodes/WbBaseNode.hpp
```diff
@@ -121,6 +121,9 @@ public slots:
   virtual void showResizeManipulator(bool enabled) {}
 
 protected:
+  bool isUrdfLinkRoot() const override;
```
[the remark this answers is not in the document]
lukicdarkoo: Added the explanation in `WbNode.hpp` as it is the file in which the method is initially declared.

src/webots/nodes/WbGeometry.cpp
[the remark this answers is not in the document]
DavidMansolino: Ok, in that case calling the closest parent is better as it might be dangerous to skip nodes in the hierarchy. With the current code, if in the future we override `exportNodeHeader` in `WbBaseNode`, then it will not be called from `WbGeometry` and this is not very easy to detect, so to avoid future issues we prefer to call the closest parent.

src/webots/vrml/WbNode.cpp
lukicdarkoo: This is not the same, `isUrdfLinkRoot()` can return true and it could happen that the node doesn't have field `name` defined. This happens when there is `joint-link-joint` and the link doesn't have a name. In that situation, we have to consider Webots node as URDF link even though there is no name.

DavidMansolino: Oh yes, you're completely right!

src/webots/vrml/WbNode.hpp
```diff
@@ -319,6 +328,8 @@ private slots:
 private:
   WbNode &operator=(const WbNode &);  // non copyable
 
+  QString mUrdfPrefix;
```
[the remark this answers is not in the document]
lukicdarkoo: I want to prefix to all nodes, but it seems to work fine.

`urdfName()` calls `getUrdfPrefix()` which searches for the robot's node link and takes `mUrdfPrefix`.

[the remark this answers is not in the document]
DavidMansolino: You're right, I forget to remove my comment when I tested it it was working fine!

src/webots/vrml/WbNode.cpp
```diff
@@ -2046,6 +2146,17 @@ QStringList WbNode::documentationBookAndPage(bool isRobot) const {
   return mModel->documentationBookAndPage();
 }
 
+const WbNode *WbNode::findRobotRootNode() const {
+  const WbNode *tmpNode = this;
+  while (tmpNode != NULL && tmpNode->nodeModelName() != "Robot")
```
DavidMansolino, line 2151: To avoid comparing directly with the `model name` we could introduce a `isRobot` function in WbNode that returns false and override it in WbRobot.

tests/api/controllers/robot_urdf/robot_urdf.c
lukicdarkoo: Good point, I will create one more test then

projects/default/controllers/ros/Ros.cpp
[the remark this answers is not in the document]
lukicdarkoo: Added

src/webots/vrml/WbNode.cpp
lukicdarkoo: We can do the same thing as for Robot (`isRobot()`), we can add `isJoint()` method and override it in `WbBasicJoint`.

src/lib/Controller/api/robot.c
```diff
@@ -1119,3 +1136,19 @@ WbSimulationMode robot_get_simulation_mode() {
...
+
+const char *wb_robot_get_urdf(const char *prefix) {
+  robot_mutex_lock_step();
+
+  robot.need_urdf = true;
+  free(robot.urdf_prefix);
+  robot.urdf_prefix = malloc(strlen(prefix) + 1);
+  strcpy(robot.urdf_prefix, prefix);
+
+  generate_request(0, false);
```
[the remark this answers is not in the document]
lukicdarkoo: You are right!

src/webots/nodes/WbRobot.cpp
```diff
@@ -1160,6 +1172,23 @@ void WbRobot::writeAnswer(QDataStream &stream) {
...
+
+    QString urdfContent;
+    WbVrmlWriter writer(&urdfContent, modelName() + ".urdf");
+
+    writer.writeHeader(name());
+    write(writer);
+    writer.writeFooter();
+
+    const QByteArray n = writer.readAll().toUtf8();
+    stream.writeRawData(n.constData(), n.size() + 1);
```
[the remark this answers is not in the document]
lukicdarkoo: I think I simplified now. Unfortunately, `stream << urdfContent` doesn't work, it is still uses `writeRawData()`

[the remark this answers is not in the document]
DavidMansolino: Ok that seems strange because in the convert to base node we can successfully use directly the string, have you tried printing urdfContent? What does it contains?

DavidMansolino:
Sorry, I just realized that a entry in the changelog is missing.

docs/reference/changelog-r2020.md
```diff
@@ -15,6 +15,7 @@ Released on XXX.
     - Added the possibility to edit the `S/MFRotation` fields using quaternions ([#1491](https://github.com/cyberbotics/webots/pull/1491)).
     - Improved edition of `S/MFRotation` fields using the field editor spin-boxes, only the last decimal is now incremented ([#1491](https://github.com/cyberbotics/webots/pull/1491)).
     - Added a [Gyro](https://www.cyberbotics.com/doc/reference/gyro) node to e-puck PROTO ([#1484](https://github.com/cyberbotics/webots/pull/1484)).
+    - Added a method to [Robot](https://cyberbotics.com/doc/reference/robot) node which allows URDF export ([#1706](https://github.com/cyberbotics/webots/pull/1706))
```
DavidMansolino, line 18:
```suggestion
    - Added a [Gyro](gyro.md) node to e-puck PROTO ([#1484](https://github.com/cyberbotics/webots/pull/1484)).
    - Added a [`wb_robot_get_urdf`](robot.md#wb_robot_get_urdf) function to the [Robot](robot.md) node which allows URDF export ([#1706](https://github.com/cyberbotics/webots/pull/1706))
```

tests/api/controllers/robot_urdf/.gitignore
```diff
@@ -0,0 +1,3 @@
+/robot_urdf
+robot.urdf
+result.txt
```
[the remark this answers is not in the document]
DavidMansolino: That was actually not the same file ignored at the first line. I am fixing this.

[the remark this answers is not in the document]
omichel: Oops, your are right. Sorry for overlooking this.

PR #1709, opened 2020-05-28
Fix qt5.15 deprecation warnings

omichel:
Qt 5.15 landed on MSYS2, happily breaking the Windows nightly builds because of deprecation warnings.
This PR fixes these warnings and hopefully the Windows nightly builds.


omichel:
✔️ [appveyor tag build](https://ci.appveyor.com/project/omichel/webots-9ufow)

DavidMansolino:
:eyes:

src/webots/sound/WbWaveFile.cpp
```diff
@@ -211,20 +211,24 @@ void WbWaveFile::loadFromFile(int side) {
...
   static QString percentageChar = "%%";
 #endif
 
   QString outputFilename = WbStandardPaths::webotsTmpPath() + fi.baseName() + ".wav";
-  ffmpeg += " -y -i " + mFilename + " " + outputFilename + " -sample_fmt s16 -loglevel quiet";
+  QStringList arguments("-y");
+  arguments << "-i" << mFilename << outputFilename << "-sample_fmt"
+            << "s16"
+            << "-loglevel"
+            << "quiet";
```
DavidMansolino, line 227: Something like this (not tested) will allows us to make the arguments a const (which I slightly prefer, but this is personal).

```suggestion
  const QStringList arguments = QStringList() << "-y" << "-i" << mFilename << outputFilename << "-sample_fmt" << "s16"  << "-loglevel" << "quiet";
```

src/webots/control/WbControlledWorld.cpp
```diff
@@ -71,7 +71,7 @@ WbControlledWorld::WbControlledWorld(WbProtoList *protos, WbTokenizer *tokenizer
   QFile file(WbStandardPaths::webotsTmpPath() + "WEBOTS_SERVER");
   if (file.open(QIODevice::WriteOnly)) {
     QTextStream stream(&file);
-    stream << mServer->fullServerName().toUtf8() << endl;
+    stream << mServer->fullServerName().toUtf8() << '\n';
```
DavidMansolino, line 74: Note that at the other occurrence of this issue you fixed it that way:
```suggestion
    stream << mServer->fullServerName().toUtf8() << Qt::endl;;
```

stefaniapedrazzi:
We could also upgrade to Qt5.15 on Linux and Mac.
Or is there a reason to use anolder version of Qt on Mac?

src/webots/control/WbControlledWorld.cpp
omichel: I had problems with compiling that on Linux. So, I believe it's simpler to use `\n` everywhere.

DavidMansolino:
> We could also upgrade to Qt5.15 on Linux and Mac.
Or is there a reason to use an older version of Qt on Mac?

There is a rush for updating on Windows but not on the other platforms, but indeed it should be possible to update on the other platforms too.

stefaniapedrazzi:
personally I think it would be better to try to upload Qt all the platforms rather than adding special cases for Windows.

omichel:
I agree with you @stefaniapedrazzi, but the problem here is mainly to fix the Appveyor nightly builds which are broken. I propose to upgrade to Qt-5.15 on Linux and mac OS in another PR.

omichel:
I restarted the travis builds: https://travis-ci.com/github/cyberbotics/webots/builds/168538941

omichel:
It seems to compile now on Linux as well.
👍

src/webots/gui/WbStreamingServer.cpp
```diff
@@ -48,6 +48,13 @@
 #include <QtWebSockets/QWebSocket>
 #include <QtWebSockets/QWebSocketServer>
 
+#include <QtCore/QDir>
+#include <QtCore/QFileInfo>
+#include <QtCore/QProcess>
+#include <QtCore/QString>
+#include <QtCore/QStringList>
+#include <QtCore/QTextStream>
```
DavidMansolino, line 56: Are all these includes really needed?

omichel: No, my mistake... I am removing them.

PR #1710, opened 2020-05-28
Update to Qt 5.15.0 on Linux

stefaniapedrazzi:
Upgrade to Qt 5.15 also on Linux (for Windows done in #1709).


stefaniapedrazzi:
The failure is due to the fact that Qt 5.15 `libxcb.so` is now linking to the system libraries (I had to install at least 4 additional libraries on the docker image to make it work).
Even if the Qt 5.15 online documentation still mention the `qt-xcb` configuration option, it is not supported when running the configure script.

I can try to recompile the Qt 5.15 `xcb` plugin with the `qt-xcb` option by using a Makefile taken from Qt 14.2

stefaniapedrazzi:
This has already been reported on the Qt interest mailing list:
https://www.mail-archive.com/interest@qt-project.org/msg33978.html

I will update the dependencies to add the new `xcb` ones.

DavidMansolino:
> The failure is due to the fact that Qt 5.15 `libxcb.so` is now linking to the system libraries (I had to install at least 4 additional libraries on the docker image to make it work).
> Even if the Qt 5.15 online documentation still mention the `qt-xcb` configuration option, it is not supported when running the configure script.
> 
> I can try to recompile the Qt 5.15 `xcb` plugin with the `qt-xcb` option by using a Makefile taken from Qt 14.2

Well spotted!

stefaniapedrazzi:
I updated the wiki pages (but I can easily revert the changes if needed):
* Linux installation: https://github.com/cyberbotics/webots/wiki/Linux-installation
* Qt build https://github.com/cyberbotics/webots/wiki/Qt-compilation#linux
* Docker https://github.com/cyberbotics/webots/wiki/Docker

stefaniapedrazzi:
:heavy_check_mark: (travis_release)

Everything compiles and runs correctly now also on Travis. 
But I still have to test the packages.

stefaniapedrazzi:
Tarball and debian packages are working fine on Ubuntu 18.04 and Ubuntu 20.04.
I also tested the snap on Ubuntu 18.04 (Ubuntu 20.04 test still in progress).

src/webots/control/WbController.cpp
```diff
@@ -284,7 +284,7 @@ void WbController::start() {
       QFile file(runtimeFilePath);
       if (file.open(QIODevice::WriteOnly)) {
         QTextStream stream(&file);
-        stream << endl;
+        stream << Qt::endl;
```
omichel, line 287: This is likely to fail on mac OS.
```suggestion
        stream << '\n';
```

PR #1713, opened 2020-05-29
Fixed controller args with double quotes

### LINKED ISSUE #1712: Broken Controller Arguments
**Describe the Bug**
The controller arguments are broken (at least in python).

**Steps to Reproduce**
1. Open and run the pedestrian world.

**Screenshots**
The starting command line looks wrong:
```
INFO: pedestrian: Starting controller: python!-u+pedestrian.py+--trajectory="0+3,+0+-3"+--speed=2.5
```

The controllers fails:
```
[pedestrian] Usage: pedestrian.py [options]
[pedestrian] 
[pedestrian] pedestrian.py: error: no such option: -3
WARNING: 'pedestrian' controller exited with status: 2.
[pedestrian] Usage: pedestrian.py [options]
```

This has probably been introduced in #1709.

omichel:
Fixes #1712.

Note: This is a quick & dirty fix until we introduce `Robot.controllerArgs` as a `MFString` instead of a `SFString` that will provide a much better interface.

### DIFF
--- a/src/webots/control/WbController.cpp
+++ b/src/webots/control/WbController.cpp
@@ -894,7 +894,9 @@ QStringList WbController::argsList() const {
     if (args[i] == '"')
       quote = !quote;
     if (args[i] == ' ' && !quote) {
-      list << args.mid(previous, i - previous - 1).replace("\"", "");
+      const QString argument = args.mid(previous, i - previous).replace("\"", "").trimmed();
+      if (!argument.isEmpty())
+        list << args.mid(previous, i - previous).replace("\"", "");
       previous = i + 1;
     }
   }
--- a/src/webots/control/WbController.cpp
+++ b/src/webots/control/WbController.cpp
@@ -230,7 +230,7 @@ void WbController::start() {
   if (mCommand.isEmpty())  // python has wrong version or Matlab 64 not available
     return;
 
-  info(tr("Starting controller: %1").arg(mCommand + " " + mArguments.join(" ")));
+  info(tr("Starting controller: %1").arg(commandLine()));
 
 #ifdef __linux__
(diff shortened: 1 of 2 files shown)


DavidMansolino:
:eyes:

src/webots/control/WbController.cpp
```diff
@@ -887,6 +883,25 @@ const QString &WbController::args() const {
   return mRobot->controllerArgs();
 }
 
+QStringList WbController::argsList() const {
```
DavidMansolino, line 886: This function would deserve a few comments in the code.

omichel:
I believe it is correct to remove the quotes. This is actually done by the system before passing the arguments to the program.

src/webots/control/WbController.cpp
```diff
@@ -887,6 +883,25 @@ const QString &WbController::args() const {
...
+  const QString args = mRobot->controllerArgs().trimmed();
+  if (args.length() == 0)
+    return list;
+  bool quote = false;
+  int previous = 0;
+  for (int i = 0; i < args.length(); i++) {
+    if (args[i] == '"')
+      quote = !quote;
+    if (args[i] == ' ' && !quote) {
+      list << args.mid(previous, i - previous - 1).replace("\"", "");
```
DavidMansolino, line 897:
```suggestion
      const QString argument = args.mid(previous, i - previous).replace("\"", "").trimmed();
      if (!argument.isEmpty())
        list << args.mid(previous, i - previous).replace("\"", "");
```

The problem was indeed not the removed quotes, but there were actually to issues on this line:
  1. if more than one space was used between arguments, then some empty arguments were pushed in the list.
  2. the cutting of the arguments was wrong. This was actually visible in the starting sequence I send before (but I just saw it know): `... --initial-linear-velocity= --initial-angular-velocity=0. --are...` (last character of each argument is missing).

PR #1718, opened 2020-06-02
Switched Robot.controllerArgs from SFString to MFString

omichel:
Better design for `Robot.controllerArgs`.
Relates with #1713 and #1709.
- [x] remove useless `wb_robot_get_controller_arguments()` API function.
- [x] deprecated almost useless `wb_robot_get_controller_name()` API function.


projects/humans/pedestrian/worlds/pedestrian.wbt
```diff
@@ -33,13 +33,16 @@ CircleArena {
 Pedestrian {
   translation 0 1.26 3
   rotation 0 1 0 3.14159
-  controllerArgs "--trajectory=\"0 3, 0 -3\" --speed=2.5"
+  controllerArgs [
+    "--trajectory=0 3, 0 -3",
+    "--speed=2.5"
+  ]
```
DavidMansolino, line 39:
```suggestion
  controllerArgs [
    "--trajectory=0 3, 0 -3"
    "--speed=2.5"
  ]
```

projects/humans/pedestrian/worlds/pedestrian.wbt
```diff
@@ -33,13 +33,16 @@ CircleArena {
...
 }
 Pedestrian {
   translation 5 1.27 1
   rotation 0 1 0 3.14159
   name "pedestrian(1)"
-  controllerArgs "--trajectory=\"5 1, 5 -1, 1 -5, -1 -5, -5 -1, -5 1, -1 5, 1 5\""
+  controllerArgs "--trajectory=5 1, 5 -1, 1 -5, -1 -5, -5 -1, -5 1, -1 5, 1 5"
```
DavidMansolino, line 45:
```suggestion
  controllerArgs [
    "--trajectory=5 1, 5 -1, 1 -5, -1 -5, -5 -1, -5 1, -1 5, 1 5"
  ]
```

projects/humans/c3d/protos/C3dViewer.proto
```diff
@@ -28,17 +28,17 @@ PROTO C3dViewer [
...
+                              fields.playbackSpeed.value .. '" "' ..
+                              fields.leftGroundReactionForceOffset.value.x .. '" "' ..
+                              fields.leftGroundReactionForceOffset.value.y .. '" "' ..
+                              fields.leftGroundReactionForceOffset.value.z .. '" "' ..
+                              fields.rightGroundReactionForceOffset.value.x .. '" "' ..
+                              fields.rightGroundReactionForceOffset.value.y .. '" "' ..
+                              fields.rightGroundReactionForceOffset.value.z .. '" "' ..
+                              fields.bodyTransparency.value .. '" "' ..
+                              fields.bodyHeight.value .. '" "' ..
+                              fields.bodyOffset.value .. '" ]' }%
```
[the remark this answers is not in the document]
omichel: It is not as simple in `SumuInterface` because many parameters are conditional, so I believe the current implementation is difficult to improve.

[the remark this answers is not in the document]
DavidMansolino: Ok, in that case this is indeed not transposable as is for the `SumoInterface`, but it need in any case some improvements as this PROTO doesn't even load as it is currently.
I will make proposition (in another comment).

projects/default/protos/SumoInterface.proto
```diff
@@ -32,15 +32,15 @@ PROTO SumoInterface [
...
     -- tracks check
-    local arguments = ""
+    local arguments = "[ "
     if fields.gui.value ~= true then
-      arguments = arguments .. "--no-gui "
+      arguments = arguments .. "\"--no-gui\" "
     end
     if fields.useNetconvert.value ~= true then
-      arguments = arguments .. "--no-netconvert "
+      arguments = arguments .. "\"--no-netconvert\" "
```
DavidMansolino, line 40: Instead of a string we can use an array.
Something like (untested):
```
local arguments = {}
if fields.gui.value ~= true then
  table.insert(arguments, "--no-gui")
end
...
```

And then 'insert' the arguments with something like:
```
controllerArgs [
  %{ for index, value in ipairs(arguments) do }%
  %{= value .. '\n' }%
  %{ end }%
]
```

DavidMansolino:
:eyes:

omichel:
I improved backwards compatibility by automatically splitting the controllerArgs of versions < R2020b at space boundaries. That works nicely for the projects/robots/gctronic/elisa/worlds/elisa3.wbt example.

PR #1607, opened 2020-04-29
fix icon path

LeoJhonSong:
Currently the `WEBOTS_HOME` on Ubuntu1804 is `/usr/local/webots`, so the icon is at `/usr/local/webots/resources/icons/core/webots.png`.

PR #1721, opened 2020-06-03
Fix icon path

DavidMansolino:
**Description**
The fix introduced in #1607 was correct, but only for Ubuntu 18.04 and more recent, not for Ubuntu 16.04 and snap (regardless of the Ubuntu version).


DavidMansolino:
This should also fix https://github.com/cyberbotics/webots-snap/issues/11#issuecomment-638033247.

DavidMansolino:
:heavy_check_mark:: travis_release_only

PR #1727, opened 2020-06-04
Improve Test Suite Robustness

DavidMansolino:
**Description**
The PR improves the test suite to make it location independant, the 'test' folder doesn't have anymore to be located in the 'webots' folder it can be located anywhere, this way it is possible for example to test with another version of Webots than the one of its development environment.

It also add support for Github actions.

I did this while testing the Github Actions and I think it is worth integrating these changes as it just makes the test suite more robust.

### DIFF
--- a/tests/command.py
+++ b/tests/command.py
@@ -162,8 +162,8 @@ class Command(object):
             outFile.close()
             errFile.close()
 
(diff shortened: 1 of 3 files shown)


DavidMansolino:
:heavy_check_mark:: travis_unit_only

PR #1730, opened 2020-06-05
Better controller starting message and backwards compatibility

omichel:
Improve controller starting message (in case the argument contains single or double quotes).
Fixed backwards compatibility in case the controller arguments contain single or double quotes.

The backward compatibility was not working with the *projects/humans/pedestrian/worlds/pedestrian.wbt* example. After this PR, is it still not working, not because of the double quotes, but because it is a template generator field of the Pedestrian PROTO.

### DIFF
--- a/src/webots/nodes/WbRobot.cpp
+++ b/src/webots/nodes/WbRobot.cpp
@@ -197,16 +197,17 @@ void WbRobot::preFinalize() {
     QStringList arguments;
     const QString args = mControllerArgs->value()[0].trimmed();
     const int l = args.length();
-    bool doubleQuote = false;
-    bool singleQuote = false;
+    bool insideDoubleQuote = false;
+    bool insideSingleQuote = false;
     int previous = 0;
     for (int i = 0; i < l; i++) {
-      if (!singleQuote && args[i] == '"' && ((i == 0) || args[i - 1] != '\\'))
-        doubleQuote = !doubleQuote;
-      else if (!doubleQuote && args[i] == '\'' && ((i == 0) || args[i - 1] != '\\'))
-        singleQuote = !singleQuote;
-      else if (args[i] == ' ' && !(singleQuote || doubleQuote)) {
-        arguments << args.mid(previous, i - previous).remove('"').remove('\'');
+      if (!insideSingleQuote && args[i] == '"' && ((i == 0) || args[i - 1] != '\\'))
+        insideDoubleQuote = !insideDoubleQuote;
+      else if (!insideDoubleQuote && args[i] == '\'' && ((i == 0) || args[i - 1] != '\\'))
+        insideSingleQuote = !insideSingleQuote;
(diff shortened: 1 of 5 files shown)


omichel:
Sorry, I noticed a bug in this PR, fixing it...

omichel:
There is a fundamental problem here: changing `Robot.controllerArgs` in the *preFinalize* method causes a PROTO template regeneration which crashes Webots at this point. I believe we should revert the change of `Robot.controllerArgs` and display only a warning, without automatically fixing the world file.

DavidMansolino:
Maybe we can fix it only in the case it is a world (or a non-regenerative field) ?

omichel:
The automatic backwards compatibility is now ensured in all the cases, except if the `controllerArgs` field is a regenerative field of a PROTO. E.g., that will work for the *projects/robots/gctronic/elisa/worlds/elisa3.wbt* world, but not for *projects/humans/pedestrian/worlds/pedestrian.wbt*.

DavidMansolino:
:eyes:

src/webots/nodes/WbRobot.cpp
```diff
@@ -190,11 +191,35 @@ void WbRobot::preFinalize() {
...
-                   "Splitting controllerArgs at space boundaries. "
-                   "You may need to update your PROTO and/or world file(s)."));
+      mControllerArgs->value().size() == 1 && mControllerArgs->value()[0].contains(' ')) {
+    // we need to split the controllerArgs[0] at space boundaries into a list of strings
+    // taking into account quotes and double quotes to avoid splitting in the middle of a quoted (or double quoted) string
+    QStringList arguments;
+    const QString args = mControllerArgs->value()[0].trimmed();
+    const int l = args.length();
+    bool doubleQuote = false;
+    bool singleQuote = false;
```
DavidMansolino, line 201: I would add the `inside` prefix to make these two variables more self-explanatory.

src/webots/nodes/WbRobot.cpp
```diff
@@ -190,11 +191,35 @@ void WbRobot::preFinalize() {
...
+    int previous = 0;
+    for (int i = 0; i < l; i++) {
+      if (!singleQuote && args[i] == '"' && ((i == 0) || args[i - 1] != '\\'))
+        doubleQuote = !doubleQuote;
+      else if (!doubleQuote && args[i] == '\'' && ((i == 0) || args[i - 1] != '\\'))
+        singleQuote = !singleQuote;
+      else if (args[i] == ' ' && !(singleQuote || doubleQuote)) {
+        arguments << args.mid(previous, i - previous).remove('"').remove('\'');
```
DavidMansolino, line 209:
```suggestion
        if (args[i - 1] != ' ')
          arguments << args.mid(previous, i - previous).remove('"').remove('\'');
```
Otherwise, if more than one space is used between arguments it will add an empty argument to the list.

PR #1736, opened 2020-06-08
Update to Qt5.15 on macOS

DavidMansolino:
**Description**
This PR updates Qt from 5.13.1 to 5.15.0 on macOS.

wiki upgrade: https://github.com/cyberbotics/webots/wiki/Qt-compilation/_compare/4e8aa0aa8a07427143204f89a75f73e5ef96a964...234f0fe5de0873343474dc6e7054bdb1b58e5d80


DavidMansolino:
✔️: travis_release_only

src/webots/control/WbController.cpp
```diff
@@ -832,23 +832,23 @@ void WbController::copyBinaryAndDependencies(const QString &filename) {
     return;
 
   QProcess process;
-  QString cmd;
   bool success;
 
   // get current RPATH
-  cmd = QString("otool -l %1 | grep LC_RPATH -A 3 | grep path | cut -c15- | cut -d' ' -f1").arg(filename);
+  const QString cmd = QString("otool -l %1 | grep LC_RPATH -A 3 | grep path | cut -c15- | cut -d' ' -f1").arg(filename);
```
stefaniapedrazzi, line 838: Would also be possible to make this a reference instead of copying the value?
```suggestion
  const QString &cmd = QString("otool -l %1 | grep LC_RPATH -A 3 | grep path | cut -c15- | cut -d' ' -f1").arg(filename);
```

DavidMansolino: Probably yes, let me try to be sure.

DavidMansolino: It seems to work fine.

Issue #1737, opened 2020-06-08
Display Normals


DavidMansolino:
**Describe the Bug**
For debugging purpose it would be very useful to be able to display the normals in the 3D view as an optional rendering.

DavidMansolino:
Implemented in #1740

PR #1740, opened 2020-06-08
Display Normals

DavidMansolino:
**Description**
This PR adds a new optionnal rendering allowing to show the normals.

**Related Issues**
This pull-request fixes issue #1737

**Preview**

unnamed
unnamed_1


DavidMansolino:
Works on IndexedFaceSet node:
high_resolution_indexedfaceset
And Mesh node:
mesh

DavidMansolino:
I have added (not yet committed) the fact that the normal have different colors if they are 'creased' or not:
normals

normals_2

Unfortunately, this is quite slow and become almost unusable as soon as the mesh as a non-negligible number of triangles (the car takes already 3-4 seconds). This is because for each normal we should check against all the other normal if there is a similar one.
So I am not so sure it is good to integrate this?

omichel:
Why not adding a boolean flag in the computation of the normals to say if the creaseAngle was applied or not? We could use this boolean flag for the color representation and that would cost nothing.

DavidMansolino:
> Why not adding a boolean flag in the computation of the normals to say if the creaseAngle was applied or not? We could use this boolean flag for the color representation and that would cost nothing.

Yes, that was my first idea, but I did gave up on this because this will imply that for each normal we have one boolean and this in any case (even if the optional rendering is disabled) I was a bit afraid that this would increase the memory footprint in all the cases, but I will try (the current solution is anyway not usable).

DavidMansolino:
Ok, it seems to works fine this way, much more performant and the memory footprint is not visible.
The only drawback is that it is purely based on the creaseAngle, which means that if the normals of the indexedFaceSet are defined or if a Mesh node includes the normal, the already creased normals will not be detected. But this is maybe not so bad as the main point of this color difference is to see the impact of the crease angle.

DavidMansolino:
Seems to works quite well:
high_resolution_indexedfaceset
high_resolution_indexedfaceset_1

DavidMansolino:
When changing the crease angle:
ezgif-1-56efaa559f00

src/webots/nodes/utils/WbTriangleMesh.hpp
```diff
@@ -57,6 +57,7 @@ class WbTriangleMesh {
     return mVertices[coordinateIndexAt(triangle, vertex, component)];
   }
   double normalAt(int triangle, int vertex, int component) const { return mNormals[3 * indexAt(triangle, vertex) + component]; }
+  bool isNormalAtCreased(int triangle, int vertex) const { return mIsNormalCreased[indexAt(triangle, vertex)]; }
```
omichel, line 60:
```suggestion
  bool isNormalCreased(int triangle, int vertex) const { return mIsNormalCreased[indexAt(triangle, vertex)]; }
```

DavidMansolino: All these functions have the 'At' (e.g. normalAt, textureCoordinateAt, nonRecursiveTextureCoordinateAt, indexAt, ...) so why not for this one?

omichel: That sounds a little bit odd to me... If we want to keep this "At" for consistency with the other methods, then, we should probably use it as a suffix. That would make "isNormalCreasedAt" which looks more consistent with other methods, although I am not super fan of the "At" suffix which doesn't say much. Maybe we should get rid of it for all methods...

DavidMansolino: I am also in favor of removing it for all the functions :wink:

DavidMansolino:

Issue #1700, opened 2020-05-25
Disable shift+drag for PROTO with translation parameter


stefaniapedrazzi:
If a PROTO has a translation parameter, the translation handles are disabled, but the shift+drag is not disabled.

This is wrong and shift+drag interaction should be disabled as well.

This can be reproduced with a user PROTO file:
* add the `test` PROTO object in the scene
* move the `test` PROTO object with shift+drag
-> the object moves and suddenly Webots crashes

```
#VRML_SIM R2019a utf8

PROTO test [
  field  SFVec3f     translation  0 1.04 0
  field  SFRotation  rotation     1 0 0 0
  field  SFString    controller   ""
  field  SFString    name         "test"
]
{
  %{
    function chassis()
      return [[DEF CHASSIS Transform {
                translation 0 0.18 0
                children [
                  Shape {
                    appearance DEF gray PBRAppearance {
                      baseColor 0.5 0.5 0.5
                      roughness 1
                      metalness 0
                    }
                    geometry Box {
                      size 1.3 0.58 3
                    }
                  }
                ]
              }]]
    end
  }%
  Robot {
    translation IS translation
    rotation IS rotation
    controller IS controller
    supervisor TRUE
    name IS name
    children [
      DEF CHASSIS Transform {
        translation 0 0.18 0
        rotation 0 0 0 0
        children [
          Shape {
            appearance DEF gray PBRAppearance {
              baseColor 0.5 0.5 0.5
              roughness 1
              metalness 0
            }
            geometry Box {
              size 1.3 0.58 3
            }
          }
        ]
      }
    ]
    boundingObject Group {
      children [
        USE CHASSIS
      ]
    }
    physics Physics {
      density -1
      mass 1000
    }
  }
}
```

PR #1761, opened 2020-06-10
Fix disabling shift+drag

stefaniapedrazzi:
Address #1700:
shfit + drag should also be disabled if translation is a PROTO paramter.

Besides this, I found that a wrong commit was introduced here (PR https://github.com/cyberbotics/webots-dev/pull/6684): https://github.com/cyberbotics/webots-dev/pull/6684/commits/130676215fc06520be777b49afe8547c66f44b86
Probably, it was not clear why the manipulator could be NULL. But this commit breaks the mechanism caching the computation of valid translation and rotations fields. I will test if I can reproduce some errors after reverting this change.


stefaniapedrazzi:

src/webots/nodes/WbAbstractTransform.cpp
```diff
@@ -257,11 +259,25 @@ bool WbAbstractTransform::isTranslationFieldVisible() const {
...
+    mCanBeRotated =
+      !WbNodeUtilities::isTemplateRegeneratorField(mBaseNode->findField("rotation", true)) && mIsRotationFieldVisible;
     mIsRotationFieldVisibleReady = true;
   }
   return mIsRotationFieldVisible;
 }
 
+bool WbAbstractTransform::canBeTranslated() const {
+  if (!mIsTranslationFieldVisibleReady)
+    isTranslationFieldVisible();
```
omichel, line 271: This seems useless. Did you mean `return isTranslationFieldVisible();`?

stefaniapedrazzi: No it is not useless.
The value `mCanBeTranslated` is computed and cached in `isTranslationFieldVisible()`.
So, if the value of `mCanBeTranslated` has not been computed yet, we need to call `isTranslationFieldVisible()` to compute it.

omichel: I see. I was fooled by those mutable fields.

stefaniapedrazzi:
I solved the conflicts in the ChangeLog

stefaniapedrazzi:
Mutable fields could maybe not be very intuitive, but helps a lot for caching functionalities that will speed up the code.
I will now propose a change where the getter is not called, and the code will be slightly less efficient but should not affect the performance.

stefaniapedrazzi:
For removing the mutable properties of cache fields, we have to modify a lot functions and remove the constant attribute of the node argument.
From my point of view, this will be worse than using mutable fields for cached values.

omichel:
I don't say we must remove all these mutable properties. I just have the feeling that calling a getter, ignoring it's return value and expecting it modifies the object, looks to be a bad design. At minimum it deserves some comment to explain this weird behavior. At best some re-design may be helpful. See my previous comment about it.

stefaniapedrazzi:
Here is a proposed version that should be more readable.

PR #1766, opened 2020-06-11
Update compiling-controllers-in-a-terminal.md

omichel:
These instructions were very outdated.

### DIFF
--- a/docs/guide/compiling-controllers-in-a-terminal.md
+++ b/docs/guide/compiling-controllers-in-a-terminal.md
@@ -29,29 +29,28 @@ Like with the editor buttons, it is possible to build the whole project, or only
 ```sh
 $ make
 $ make clean
-$ make my_robot.class
+$ make -f Makefile_java my_robot.class
 $ make my_robot.o
 ```
 
 ### Windows
 
-On Windows you must use the MSYS terminal to compile the controllers.
-MSYS is a UNIX-like terminal that can be used to invoke MinGW commands.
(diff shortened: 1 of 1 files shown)


docs/guide/compiling-controllers-in-a-terminal.md
```diff
@@ -29,29 +29,28 @@ Like with the editor buttons, it is possible to build the whole project, or only
...
+Please follow the instructions [here](https://github.com/cyberbotics/webots/wiki/Windows-installation#msys2-development-environment-and-git) to install it.
 
-```sh
-WEBOTS_HOME=C:\Program Files\Webots
-PATH=C:\program Files\Webots\mingw\bin;C:\...
+You will also have to set the `WEBOTS_HOME` environment variable to point to the installation folder of Webots, typically `C:\Program Files\Webots` and the path to the Webots binaries in the MSYS2 `PATH` environment variable:
+
+```bash
+export WEBOTS_HOME="C:\Program Files\Webots"
+export PATH=$PATH:/C/Program\ Files/Webots/msys64/mingw64/bin:/C/Program\ Files/Webots/msys64/mingw64/bin
```
DavidMansolino, line 46: Webots/lib/controller is not required?

omichel: I don't think so. It's needed for executing the controller, but not for compiling as the Makefile rely on WEBOTS_HOME to find the libraries.

DavidMansolino: Oh yes you are right.

PR #1768, opened 2020-06-11
Fix self node in supervisor.c for a Robot proto parameter node

### LINKED ISSUE #1604: Can't change custom MFBoolField of a proto
Associated files (proto/world/controller) : 
[Bug report.zip](https://github.com/cyberbotics/webots/files/4552435/Bug.report.zip)

Can't change value of a custom MFBool field.

This field has 36 values. It's placed in the proto named `TileContainer`, which is a children of the proto `SpacecraftStructure`. 
The corresponding world is available in the archive and is named `FullMission.wbt`.

This bug has been seen on Webots R2020b, on comit  `fe5893b968e160447a52cc4619ad68a8f41f8022` compiled from the git.

**Steps to Reproduce**
1. Add the different proto and controller file to your project path
2. Open Webot and open `FullMission.wbt`
3. Make sure there is only one proto in the scene tree of type `SpacecraftStructure`, which has only one proto child of type `TileContainer`.
Make sure the proto child of type `TileContainer` has one supervisor named `controller_21`.
4. Launch the whole simulation.

**Expected behavior**
Supervisor should display the same output.

isLocked field should have different value in the scene tree. All values should be at `True` instead of `False`. 

**Screenshots**
The field "isLocked" doesn't change. 

image

Output of the supervisor: 
image
https://pastebin.com/bCN2Q9n0

**System**
 - Operating System: Ubuntu 16.04
 - Webots: Version R2020b, commit  `fe5893b968e160447a52cc4619ad68a8f41f8022`
 - Graphics Card: Nvidia gtx1050 4go
 - Processor: Intel(R) Core(TM) i5-6500 CPU

**Additional context**
Webots runned on Docker.
Docker version 19.03.5, build 633a0ea838

stefaniapedrazzi:
Fix #1604:
in case of PROTO and nested robots, the node passed during the supervisor `C_CONFIGURE` message is a PROTO parameter instance and not the node visible in the scene tree.
But the `wb_supervisor_node_get_from_def` function (and `wb_supervisor_node_get_self`) are supposed to return by default the visible nodes and not the instances, so that changes to the node are applied correctly from the Supervisor API in the exact same way they are applied from the scene tree.

### DIFF
--- a/src/webots/nodes/utils/WbSupervisorUtilities.cpp
+++ b/src/webots/nodes/utils/WbSupervisorUtilities.cpp
@@ -1585,12 +1585,15 @@ void WbSupervisorUtilities::writeAnswer(QDataStream &stream) {
 }
 
 void WbSupervisorUtilities::writeConfigure(QDataStream &stream) {
+  WbNode *selfNode = mRobot;
+  while (selfNode->protoParameterNode())
+    selfNode = selfNode->protoParameterNode();
   stream << (short unsigned int)0;
   stream << (unsigned char)C_CONFIGURE;
-  stream << (int)mRobot->uniqueId();
-  QByteArray s = mRobot->modelName().toUtf8();
+  stream << (int)selfNode->uniqueId();
+  const QByteArray &s = selfNode->modelName().toUtf8();
   stream.writeRawData(s.constData(), s.size() + 1);
-  QByteArray ba = mRobot->defName().toUtf8();
+  const QByteArray &ba = selfNode->defName().toUtf8();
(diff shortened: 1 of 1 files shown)


stefaniapedrazzi:
I checked but I don't see any use case where it should be needed to retrieve the instantiated node instead of the PROTO parameter node from the Supervisor API.
So it seems correct to send directly the Robot node visible in the scene tree in the `C_CONFIGURE` message.

stefaniapedrazzi:

DavidMansolino:
👀

PR #1770, opened 2020-06-11
Fix Qt 5.15 deprecation warnings reintroduced during merge

stefaniapedrazzi:
Readd the deprecation warnings fixes that were mistakenly removed in #1754.


src/webots/gui/WbView3D.cpp
```diff
@@ -2209,7 +2209,7 @@ void WbView3D::wheelEvent(QWheelEvent *event) {
 
 #ifndef __APPLE__  // bug in qt on Mac: -> QWheelEvent->orientation() is wrong when SHIFT + MOUSE_WHEEL_VERTICAL_SCROLL
   // Some mouse wheels can be scrolled horizontally
-  if (event->orientation() != Qt::Vertical)
+  if (event->angleDelta().x() != 0)
```
stefaniapedrazzi, line 2212: This is exactly the same as on the master branch.
But if we still have this described bug on mac, then it won't be possible to perform SHIFT + MOUSE_WHEEL_VERTICAL_SCROLL and move objects.

But this should be tested and fixed on the master branch if needed.

PR #1775, opened 2020-06-12
Fix urdf test

DavidMansolino:
**Description**
Fix various issues with the URDF test.


lukicdarkoo:
I didn't handle the return code from `system()` properly:
```
Bits 15-8 = Exit code.
Bit     7 = 1 if a core dump was produced.
Bits  6-0 = Signal number that killed the process.
```

Therefore, `(urdf_check_status >> 8) == 127` means that the command doesn't exist. I can edit it so `liburdfdom-tools` is optional.

DavidMansolino:

Issue #1779, opened 2020-06-12
Typo in PannelWithTubes proto name


omichel:
[PannelWithTubes](https://github.com/cyberbotics/webots/blob/master/projects/objects/panels/protos/PannelWithTubes.proto) proto contains a typo. It should be spelled `PanelWithTubes` instead.

gokulp01:
@omichel I've fixed it in #1781 . Let  me know if there are any issues :)

PR #1781, opened 2020-06-14
Fixed Typo in PannelWithTubes

gokulp01:
**Description**
Fixed the typo in PannelWithTubes to PanelWithTubes

**Related Issues**
This pull-request fixes issue #1779 

**Tasks**
  - [x] Fix the typo throughout the software stack


DavidMansolino:
Thank you for your fix!
Just one minor detail, to avoid breaking the backward compatibility, we usually create a backward compatibility PROTO, here is an example:
https://github.com/cyberbotics/webots/blob/master/projects/appearances/protos/ChequeredParquetry.proto
The main point are:
  - Add the 'hidden' tag:
https://github.com/cyberbotics/webots/blob/8d130e5a728ff65dc4f2e04d29762c1f08c693dc/projects/appearances/protos/ChequeredParquetry.proto#L4
  - Add a warning:
  %{ io.stderr:write("Deprecated 'ChequeredParquetry' PROTO, please use the 'Parquetry' PROTO instead and select the right type.\n") }%
  - It does not copy the content of the new PROTO but just use it.
  - It is not documented (As we don't we to encourage users to use it).

gokulp01:
@DavidMansolino just confirming: in the warning, I just copy the same message that is there in ChequeredParquetry? Or do I give a different warning?

DavidMansolino:
Almost copy-pasting yes, however, you need to update the PROTO names in the message.
I checked what you did and I think there is a small misunderstanding, you should not add the `hidden` tag and warning message in the new PROTO but in the old one. Let me now if this is not clear.

gokulp01:
> Almost copy-pasting yes, however, you need to update the PROTO names in the message.

Oh sorry my bad. I'll make the change

> I checked what you did and I think there is a small misunderstanding, you should not add the `hidden` tag and warning message in the new PROTO but in the old one. Let me know if this is not clear.
Just to be clear, the old PROTO is the Panel.proto right.

DavidMansolino:
Thank you,
The old PROTO is `PannelWithTubes`, and the new one is `PanelWithTubes` (the filename should be changed too).

gokulp01:
@DavidMansolino 
From what I understand:
- There is no PanelWithTubes PROTO. So I need to create one, and the contents are almost the same as the PannelWithTubes PROTO, with the applied changes
- Add the hidden tag and the warning message to the old PROTO. 

Am I on the right path?

DavidMansolino:
> There is no PanelWithTubes PROTO. 
> So I need to create one, 

Yes,

> and the contents are almost the same as the PannelWithTubes PROTO, with the applied changes

Not exactly, the idea is that the old PROTO will use the new one (it is just a kind of wrapper), e.g.:
```
# tags: hidden
# A gray panel with tubes.

PROTO PannelWithTubes [
  field SFVec3f    translation 0 0 0
  field SFRotation rotation    0 1 0 0
  field SFString   name        "panel with tubes"
  field SFNode     appearance  BrushedAluminium { textureTransform TextureTransform { scale 3 3 } }
]
{
  %{ io.stderr:write("Deprecated 'PannelWithTubes' PROTO, please use the 'PanelWithTubes' PROTO instead and select the right type.\n") }%
  PanelWithTubes {
    translation IS translation
    rotation IS rotation
    name IS name
    appearance IS appearance
  }
}
```

> Add the hidden tag and the warning message to the old PROTO.

Yes.

DavidMansolino:
@gokulp01 may I let you rename the icon (I don't have write access)?
https://github.com/gokulp01/webots/blob/master/projects/objects/panels/protos/icons/PannelWithTubes.png

gokulp01:
Thank you and sorry for all the hassle :)

DavidMansolino:
> Thank you and sorry for all the hassle :)

Don't worry :wink:

PR #1782, opened 2020-06-14
Update call to ``inverse_kinematics``

alxy:
The library only allows to pass 3D vectors as arguments now, the orientations is handled by the second argument
Reference: https://ikpy.readthedocs.io/en/latest/chain.html#ikpy.chain.Chain.inverse_kinematics

**Description**
Without this change, the corresponding controller throws the following warning
```
[inverse_kinematics] Draw a circle on the paper sheet...
[inverse_kinematics] Traceback (most recent call last):
[inverse_kinematics]   File "inverse_kinematics.py", line 100, in <module>
[inverse_kinematics]     [0, 0, 0, 1]
[inverse_kinematics]   File "F:\Programme\Anaconda3\lib\site-packages\ikpy\chain.py", line 137, in inverse_kinematics
[inverse_kinematics]     frame_target[:3, 3] = target_position
[inverse_kinematics] ValueError: cannot copy sequence with size 4 to array axis with dimension 3
WARNING: 'inverse_kinematics' controller exited with status: 1.
```

**Related Issues**
https://github.com/Phylliade/ikpy/issues/85

### DIFF
--- a/projects/robots/abb/irb/controllers/inverse_kinematics/inverse_kinematics.py
+++ b/projects/robots/abb/irb/controllers/inverse_kinematics/inverse_kinematics.py
@@ -80,7 +80,7 @@ for motorName in ['A motor', 'B motor', 'C motor', 'D motor', 'E motor', 'F moto
 
 # Get the arm and target nodes.
 target = supervisor.getFromDef('TARGET')
-arm = supervisor.getFromDef('ARM')
+arm = supervisor.getSelf()
 
 # Loop 1: Draw a circle on the paper sheet.
 print('Draw a circle on the paper sheet...')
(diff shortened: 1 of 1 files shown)


projects/robots/abb/irb/controllers/inverse_kinematics/inverse_kinematics.py
```diff
@@ -80,7 +80,7 @@
 
 # Get the arm and target nodes.
 target = supervisor.getFromDef('TARGET')
-arm = supervisor.getSelf()
+arm = supervisor.getFromDef('ARM')
```
DavidMansolino, line 83: It is safer to keep the `getSelf` as in any case in the world the `ARM` DEF is not defined.

alxy: Done!

alxy:
Maybe it even makes sense to ship an example controller like this one with a `requirements.txt` file to enforce a specific version constraint on the dependencies, so anyone can easily reproduce the example by installing the dependencies from this file.

DavidMansolino:
Good idea, if you want, feel free to open a new PR to add a requirements.txt file in the controller folder.

PR #1785, opened 2020-06-15
Removed useless paths added to the controllers PATH variable on Windows

omichel:
Following #1765, these two paths are now redundant.


omichel:
Done in #1765.

PR #1790, opened 2020-06-17
Add missign Robotis-op2 supervisor field.

DavidMansolino:
**Description**
Add missing Robotis-op 2 and 3 supervisor field.


docs/guide/robotis-op2.md
```diff
@@ -48,6 +48,7 @@ RobotisOp2 {
   MFString   controllerArgs  []
   SFString   customData      ""
   SFBool     supervisor      FALSE
+  SFBool     supervisor      FALSE
```
omichel, line 51: That seems to be duplicate...
```suggestion
```

PR #1802, opened 2020-06-18
Fix Camera Reset

### LINKED ISSUE #1778: does not work robotis-op2 camera after simulation reset
**Describe the Bug**

**Steps to Reproduce**
1. ' generate a controller and a supervisor for op2. '
2. 'use your new simulation reset command in supervisor'
3. run for your controller for one step then see reset and the first position of robot.
4. Check the camera image. It is one before reset. Please see it. 
5. I can see new images in screen. However I cant receive no more change in the camera array. 
by using camera.getImageArray()).

**Expected behavior**
After reset, it is expected to change the images taken by using camera.getImageArray())

**Screenshots**

**System**
 - Operating System: [e.g., Windows 10, Linux Ubuntu 18.04, macOS Mojave]
 - Graphics Card: [e.g., NVIDIA GeForce RTX 2080 11 GB, AMD Radeon RX 580 8GB, etc.]

**Additional context**

DavidMansolino:
**Description**
Fix #1778.

### DIFF
--- a/src/lib/Controller/api/abstract_camera.c
+++ b/src/lib/Controller/api/abstract_camera.c
@@ -138,13 +138,15 @@ void abstract_camera_toggle_remote(WbDevice *d, WbRequest *r) {
 }
 
 bool abstract_camera_request_image(AbstractCamera *ac, const char *functionName) {
-  double current_simulation_time = wb_robot_get_time();
+  const double current_simulation_time = wb_robot_get_time();
+  const double previous_image_update_time = ac->image_update_time;  // in case of reset time can go backward
   if (ac->image_update_time >= current_simulation_time)
     return true;
 
   ac->image_requested = true;
   wb_robot_flush_unlocked();
-  if (ac->image_update_time != current_simulation_time && robot_is_quitting() == 0) {
+  if (ac->image_update_time != current_simulation_time && previous_image_update_time <= ac->image_update_time &&
+      robot_is_quitting() == 0) {
     fprintf(stderr, "Warning: %s: image could not be retrieved.\n", functionName);
     return false;
   }
--- a/src/lib/Controller/api/abstract_camera.c
+++ b/src/lib/Controller/api/abstract_camera.c
@@ -140,7 +140,7 @@ void abstract_camera_toggle_remote(WbDevice *d, WbRequest *r) {
 bool abstract_camera_request_image(AbstractCamera *ac, const char *functionName) {
   const double current_simulation_time = wb_robot_get_time();
   const double previous_image_update_time = ac->image_update_time;  // in case of reset, time can go backward
-  if (ac->image_update_time >= current_simulation_time)
+  if (previous_image_update_time >= current_simulation_time)
     return true;
 
   ac->image_requested = true;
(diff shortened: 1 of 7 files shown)


DavidMansolino:

omichel:
:eyes:

src/webots/nodes/WbRobot.cpp
```diff
@@ -981,6 +983,8 @@ void WbRobot::dispatchAnswer(QDataStream &stream, bool includeDevices) {
       }
     }
   } else {
+    if (mNeedToForceTime)
+      writeAnswer(stream);
```
omichel, line 987: Are you sure this is needed as `writeAnswer()` is called anyhow on line 993?

DavidMansolino: Yes, this is actually the main point of the fix, in the specific reset case we need to make sure that the robot messages are sent before the device ones to update the time.
If we want to avoid calling it a second time, we could add a kind of `if (!mNeedToForceTime)` for the second call but IMO this complexifies the code uselessly (as the second call to writeAnswer will strictly do nothing as all the messages are already sent).

omichel: OK, and why do we need to send the Robot answer after the device answers in the normal case? I am afraid the reset case will change this and it may have some side effect. If there is no reason for sending the robot answer after, we should send it before in all cases.

DavidMansolino: I did not put it by default to the beginning for this purpose. When doing the reset, since the time is reset to 0 it does not cause any problem. But for the other steps, I am afraid of non-evident side effects of increasing the time before sending the device messages.

omichel: I see. Thank you for the explanation. I will try to investigate the possible side effects of moving the robot answer before the devices answer, as it seems cleaner (we do that for the configure message).

src/lib/Controller/api/abstract_camera.c
```diff
@@ -138,13 +138,15 @@ void abstract_camera_toggle_remote(WbDevice *d, WbRequest *r) {
...
-  double current_simulation_time = wb_robot_get_time();
+  const double current_simulation_time = wb_robot_get_time();
+  const double previous_image_update_time = ac->image_update_time;  // in case of reset, time can go backward
   if (ac->image_update_time >= current_simulation_time)
     return true;
 
   ac->image_requested = true;
   wb_robot_flush_unlocked();
-  if (ac->image_update_time != current_simulation_time && robot_is_quitting() == 0) {
+  if (ac->image_update_time != current_simulation_time && previous_image_update_time <= ac->image_update_time &&
```
omichel, line 148: The new condition seems to be always true, isn't it?

DavidMansolino: No, because the `wb_robot_flush_unlocked` function will increase the `ac->image_update_time` time, therefore `ac->image_update_time == current_simulation_time`, and about `previous_image_update_time <= ac->image_update_time`, this is not true in case of reset, because `ac->image_update_time == 0.0`.

omichel: OK, I see. Thanks for the explanation.

src/lib/Controller/api/abstract_camera.c
```diff
@@ -138,13 +138,15 @@ void abstract_camera_toggle_remote(WbDevice *d, WbRequest *r) {
 }
 
 bool abstract_camera_request_image(AbstractCamera *ac, const char *functionName) {
-  double current_simulation_time = wb_robot_get_time();
+  const double current_simulation_time = wb_robot_get_time();
+  const double previous_image_update_time = ac->image_update_time;  // in case of reset, time can go backward
   if (ac->image_update_time >= current_simulation_time)
```
omichel, line 143:
```suggestion
  if (previous_image_update_time >= current_simulation_time)
```

PR #1803, opened 2020-06-18
Update system-requirements.md

nishantkr18:
**Description**
Should the latest LTS OS version for ubuntu be updated, in system requirements?

**Documentation**
https://cyberbotics.com/doc/guide/system-requirements?version=master


DavidMansolino:
Our build environment is not yet full yet compatible with Ubuntu 20.04, but it will be for the next release and we will merge you PR at this point.

DavidMansolino:
Outedated

PR #1806, opened 2020-06-19
Fix URDF Link Naming

lukicdarkoo:
Before, URDF link names were in form `[name]_[unique_id]` and now this is avoided when possible leaving only `[name]` as the URDF link name.


DavidMansolino:
Shouldn't this file be updated?
 - https://github.com/cyberbotics/webots/blob/develop/tests/api/controllers/robot_urdf/reference_robot.urdf

lukicdarkoo:
> Shouldn't this file be updated?
> 
> * https://github.com/cyberbotics/webots/blob/develop/tests/api/controllers/robot_urdf/reference_robot.urdf

Yes! I will do it in a minute

DavidMansolino:
:heavy_check_mark:: travis_unit (https://travis-ci.com/github/cyberbotics/webots/builds/172174330)

PR #1812, opened 2020-06-19
Updated world files

omichel:
This PR contains world files that were converted using the `convert_to_nue.py` script.


omichel:
Not the best way to proceed.

PR #1817, opened 2020-06-23
Create test_suite.yml

DavidMansolino:
**Description**
The goal of this PR is to run the test suite on Github actions, the advantage is that we can easily compile on a clean environment and then use the output of the compilation in a new clean environment (the test does not run in the development environment but in the released environment).

Additionally, the following test will also run:
  - [ ] tests/manual_tests/test_worlds_warnings.py
  - [ ] src/packaging/update_all_worlds.py

### DIFF
--- a/.github/workflows/test_suite.yml
+++ b/.github/workflows/test_suite.yml
@@ -27,7 +27,7 @@ jobs:
     needs: build
     strategy:
       matrix:
-        test: [tests/test_suite.py, tests/manual_tests/test_worlds_warnings.py]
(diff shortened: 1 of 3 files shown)


DavidMansolino:
For now, the update_worlds and test_world_warnings tests are commented on because it will require to fix #1830.
But I think it is still worth merging it already as it tests the complete compilation and test suite (on a clean environment and not the development on as currently on Travis).

For now, the test is run on each commit of a PR as soon as the PR is not in draft mode. This is will probably generate too much test and saturate it, but let's try it (it is in any case not blocking for now) before introducing a mechanism allowing to trigger the tests.

PR #1822, opened 2020-06-23
Fix Mesh Node Crash

DavidMansolino:
**Description**

Fix crash of the Mesh node with some meshes (private mesh sent to support by a user).

These sometimes large buffers can overflow the stack, allocating the memory on the heap instead fix the crash.


DavidMansolino:
:heavy_check_mark:: travis_unit_only

PR #1826, opened 2020-06-24
Hotfix infra red distance sensors

DavidMansolino:
**Description**
In the untextured case, the color (red component) was not taken into account when computing the distance of infra-red distance sensors.


DavidMansolino:
:heavy_check_mark:: travis_unit (https://travis-ci.com/github/cyberbotics/webots/builds/172806186)

PR #1833, opened 2020-06-24
Fix parent methods in qobject

### LINKED ISSUE #1829: Cleanup parent()/setParent() methods for classes derived from QObject
Following-up investigations about #1824, a number of classes, including WbNode and WbField inherit from QObject and implement parent() and setParent() methods with a different signature from the one of QObject. This is highly error prone and should be fixed by renaming these methods. Moreover, WbField.cpp calls the setParent() method of QObject which is likely wrong...

omichel:
Fixed #1829, including the removal of apparently useless calls to QObject::setParent() in WbField.cpp.
✔️ passed the test suite locally on Windows.

### DIFF
--- a/src/webots/vrml/WbNode.cpp
+++ b/src/webots/vrml/WbNode.cpp
@@ -820,7 +820,7 @@ int WbNode::findSubFieldIndex(const WbField *const searched) const {
   return -1;
 }
 
-WbField *WbNode::findSubField(int index, WbNode *&parent) const {
+WbField *WbNode::findSubField(int index, WbNode *&parentNode) const {
   int count = 0;
   QList<WbNode *> list(subNodes(true));
   list.prepend(const_cast<WbNode *>(this));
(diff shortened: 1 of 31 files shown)


DavidMansolino:
:eyes:

src/webots/vrml/WbNode.cpp
```diff
@@ -827,7 +827,7 @@ WbField *WbNode::findSubField(int index, WbNode *&parentNode) const {
   foreach (WbNode *const node, list) {
     foreach (WbField *const field, node->mFields) {
       if (count == index) {
-        parentNode = node;
+        parent = node;
```
DavidMansolino, line 830:
```suggestion
        parentNode = node;
```
This change is wrong.

omichel: Well spotted! But in that case, I would prefer to rename the parameter name as the `parentNode` parameter name shadows the `parentNode()` method.

omichel:
For consistency, I also changed `WbNode::setGlobalParent()` to `WbNode::setGlobalParentNode()`, but there was much more occurrences...

PR #1834, opened 2020-06-25
Fixed rendering of tracks and muscles in robot cameras

### LINKED ISSUE #1808: pen_box test is failing on Windows
These two tests are failing on Windows with the following error message:
```
FAILURE with pen_box: The number of pixels painted after the first step should be lower than 20 not 1224.
```
```
FAILURE with pen_mesh_scaled: The number of pixels painted after the first step should be lower than 20 not 58
```

omichel:
Fixes #1808.

The rendering of muscles and tracks was not correct if the main rendering was disabled which happens when the main window is minimized. The callback functions need to be called also before robot camera rendering, not just before main scene rendering.


omichel:
I will retarget it to the master branch.

PR #1847, opened 2020-06-29
Fix warning for is datatype mismatch in procedural protos

### LINKED ISSUE #1843: Broken Backward compatibility with ControllerArgs
**Describe the Bug**
In some cases (e.g. procedural PROTOs) the change of the `controllerArgs` fields from SFString to MFString broke the backward compatibility.

For example, with such PROTO:
```
#VRML_SIM R2020a utf8
# tags: static

PROTO MyCar [
  field SFVec3f      translation              0 0.4445 0
  field SFRotation   rotation                 0 1 0 0
  field SFColor      color                    0.92 0.94 0.94
  field SFString     controller               "void"
  field SFString     controllerArgs           ""
]
{
  Car {
    translation IS translation
    rotation IS rotation
    controller IS controller
    controllerArgs IS controllerArgs
    recognitionColors [
      %{= fields.color.value.r }% %{= fields.color.value.g }% %{= fields.color.value.b }%
    ]
  }
}
```

2 controllerArgs fields are generated in the scene-tree (without any warnings):
aaa

None of them is usable (not passed to the controller).

omichel:
Fixes #1843.

This PR:
- removes some useless piece of code: we never need to write the IS keyword in a `.wbt` file.
- fixes missing warning about IS datatype mismatch for procedural protos.

It doesn't automatically fixes the backwards compatibility problem related to the controllerArgs datatype change (which should be nevertheless relatively rare).

### DIFF
--- a/projects/objects/road/protos/LaneSeparation.proto
+++ b/projects/objects/road/protos/LaneSeparation.proto
@@ -281,7 +281,9 @@ PROTO LaneSeparation [
         %{ end }%
         bottom IS bottom
         roadBorderHeight IS roadBorderHeight
-        roadBorderWidth IS roadBorderWidth
+        roadBorderWidth [
+          %{= borderWidth }%
+        ]
         appearance IS appearance
         pavementAppearance IS pavementAppearance
         roadBoundingObject IS roadBoundingObject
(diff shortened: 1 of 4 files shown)


stefaniapedrazzi:
:eyes:

stefaniapedrazzi:
I'm running the `test_worlds_warnings.py` manual test to check if verifying the aliasing for template PROTO node is causing some issues in the distributed worlds.
It will take some time.

For the moment I just got some correct errors with the `projects/objects/road/protos/LaneSeparation.proto`:
```
ERROR: '/home/stefi/webots/projects/objects/road/protos/LaneSeparation.proto':23:37: error: Type mismatch between 'roadBorderWidth' PROTO parameter and field 'roadBorderWidth'.
```

omichel:
Thank you for the review.
I believe I fixed these two protos.

PR #1851, opened 2020-06-30
Fix Texture Filtering Field auto-changed.

### LINKED ISSUE #1830: Avoid Changing Fields Depending on user Hardware
**Describe the Bug**
Currently, the filtering field of the ImageTextures node is changed depending on the user hardware capability:
https://github.com/cyberbotics/webots/blob/f3d714bf66351abe9df8aff92282e037ea6430c7/src/webots/nodes/WbImageTexture.cpp#L223-L236

This is bad because:
  - It is performed independently for each ImageTexture node.
  - It is bad to save hardware-dependent values in the world file.

Instead, we should find a more generic way to handle this, for example by adding a new field in the opengl preferences and test this only at the first launch of Webots.

DavidMansolino:
Fix :
  - #1830
  - limit filtering to 5 as documented.

### DIFF
--- a/src/webots/gui/WbView3D.cpp
+++ b/src/webots/gui/WbView3D.cpp
@@ -1212,7 +1212,6 @@ void WbView3D::checkRendererCapabilities() {
   bool disableAntiAliasing = false;
   bool disableGTAO = false;
   int reduceTextureQuality = 0;
-  int maxTextureFiltering = 1;
 
   // 2. determine what has to be reduced
   if (!mWrenRenderingContext->isNvidiaRenderer() && !mWrenRenderingContext->isAmdRenderer() &&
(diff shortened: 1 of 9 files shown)


src/webots/gui/WbView3D.cpp
```diff
@@ -1212,6 +1212,7 @@ void WbView3D::checkRendererCapabilities() {
   bool disableAntiAliasing = false;
   bool disableGTAO = false;
   int reduceTextureQuality = 0;
+  int maxTextureFiltering = 1;
```
omichel, line 1215: I would move this line to line 1262.

src/webots/gui/WbView3D.cpp
```diff
@@ -1299,6 +1305,12 @@ void WbView3D::checkRendererCapabilities() {
     WbPreferences::instance()->setValue("OpenGL/textureQuality", 2 - reduceTextureQuality);
   }
 
+  if (maxTextureFiltering < WbPreferences::instance()->value("OpenGL/textureFiltering").toInt()) {
+    message += "\n - ";
+    message += tr("Texture maximum filtering has been reduced.");
```
omichel, line 1310:
```suggestion
    message += tr("Texture maximum filtering has been reduced due to GPU limitations.");
```

PR #1856, opened 2020-06-30
Disable Screen Specific Perspective Save

DavidMansolino:
**Description**
The goal of this PR is instead of providing a way to disable all the wbproj update (with the current `WEBOTS_DISABLE_SAVE_PERSPECTIVE_ON_CLOSE`), disable only those who are computer (more precisely screen) specific.

This should help us maintaining the wbproj up to date (they were updated a few weeks ago and already most of them outdated), it will also allow us to test from the test suite that they are up to date.

@cyberbotics/maintainers a few questions remains:
  - should this replace `WEBOTS_DISABLE_SAVE_PERSPECTIVE_ON_CLOSE`.
  - what can be a good name for this new environment variable?
  - the goal yet is to disable only the window perspective update, all the rest (e.g. files open in the text editor) will be saved. Do you see something else important to disable?

### DIFF
--- a/src/install_scripts/bash_profile.mac
+++ b/src/install_scripts/bash_profile.mac
@@ -8,9 +8,9 @@
 # Edit these WEBOTS environment variables when needed #
 #######################################################
 
-export WEBOTS_DISABLE_SAVE_PERSPECTIVE_ON_CLOSE=1  # If defined, Webots will not save perspective changes when closed.
-export WEBOTS_ALLOW_MODIFY_INSTALLATION=1          # If defined, you are allowed to modify files in the Webots home using Webots.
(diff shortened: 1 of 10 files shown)


omichel:
That's a good idea! We could use a variable named `WEBOTS_DISABLE_SAVE_SCREEN_PERSPECTIVE_ON_CLOSE` and we could get rid of `WEBOTS_DISABLE_SAVE_PERSPECTIVE_ON_CLOSE` which would become almost useless.

DavidMansolino:
I also changed the use of QHash for a QMap for the perspectives to make sure the order is respected (and that we don't get a different order each time we save the perspectives), this should also be slightly faster as QMap is faster than QHash for a low number of items ~20 (https://woboq.com/blog/qmap_qhash_benchmark.html).

DavidMansolino:
Result of the update is visible here:  #1861

PR #1858, opened 2020-07-01
Fix memory management in Python RangeFinder.rangeImageGetDepth

### LINKED ISSUE #1783: Python rangeImageGetDepth Function Crashes
**Describe the Bug**
The following controller makes the controller crash (can be tested with the 'range_ffinder.wbt' world):
```Python
from controller import Robot

robot = Robot()

range_finder = robot.getRangeFinder('range-finder')
range_finder.enable(timestep);

    image = range_finder.getRangeImage()
    print(range_finder.rangeImageGetDepth(image, range_finder.getWidth(), 0, 0))
```

The crash, unfortunately, doesn't give any useful information (even in Python):
```
INFO: my_controller: Starting controller: python3.7 -u my_controller.py
WARNING: my_controller: The process crashed some time after starting successfully.
WARNING: 'my_controller' controller crashed
```

Tested both with Python 2.7 and 3.7

stefaniapedrazzi:
Fix #1783.

When returning the pointer to an internal PyObject, it is required to increment the reference count of the object so that the Python runtime is notified that the object is used somewhere else and the memory is not reused when the internal image list is updated (deleted and recreated).

I applied the same fix also to the deprecated `RangeFinder.rangeImageGetValue` and improved the warning text that were referring to Camera instead of RangeFinder.


stefaniapedrazzi:
I don't think there is a memory leak.
In all the examples I found only `Py_INCREF` is called. It should be the garbage collector of Python that decreases automatically the reference count when the object gets out of scope.

PR #1863, opened 2020-07-01
Fixed outline offset for boxes

### LINKED ISSUE #1777: Use of WorldInfo.lineScale for Box gives odd results
With a `WorldInfo.lineScale` set to 0.5, the corner of the `RectangleArena` is displayed like this:
e-puck_matlab
Which looks odd. This is due to the fact that the wall is long and the wire frame box size is computed from the different sizes of the box (X, Y, Z) multiplied by the line scale. Instead, we should multiply by the average or minimum size, e.g., `wireframebox.x = box.x + min(box.x, box.y, box.z) * factor * lineScale` instead of `wireframebox.x = box.x + box.x * factor * lineScale`.

omichel:
Fixes #1777.


stefaniapedrazzi:
:eyes:

omichel:
Thank you for the review. I reduced the scale to 1. 👍

Issue #1867, opened 2020-07-02
Improve detection of PROTO template regeneration fields


stefaniapedrazzi:
Are there other ways to access a parameter from a Lua statement other than using the `fields.` prefix?

Otherwise we could maybe improve the search by including the `fields.` prefix in the regular expression, something like
```
QRegExp(QString("(^|[^a-zA-Z0-9_])fields\.%1($|[^a-zA-Z0-9_])").arg(QRegExp::escape(model->name())))))
```

Here is the current regular expression used to extract if a PROTO field triggers template regeneration:
https://github.com/cyberbotics/webots/blob/0fe9e56f266a8855e9ab74e811f128f083307e5b/src/webots/vrml/WbProtoModel.cpp#L204

DavidMansolino:
No, I don't see any other way to get the value.
But we can have something like this `fields.wheelFrontRight.value.fields.tireRadius.value` with multiple time the `fields.` keyword, but this should not cause problem.

DavidMansolino:
Oh, just thought that one tricky way could be to copy the dictionnary:
```
local newDict = fields
local parameterValue = fields.myField.value
``
But none of our examples are using this, and I don't see any reason for doing this.

omichel:
So, we should probably search for `fields` instead of `fields.` to be on the safe side.

DavidMansolino:
I am not sure to understand how this will help to determine if the field is regenerative or not?

omichel:
Sorry, forget about my comment. I was confused. The only way to handle this rare case would be to search for this kind of statement (like `newDict = fields`, `newDict2=fields`) and add all the `newDict.` (or whatever is found), `newDict2.`, etc. in the search pattern in addition to `fields.`. But I am not sure it is worth adding that complexity to our code for such a rare use case.

Issue #1853, opened 2020-06-30
Crash When Changing ImageTexture.filtering field


DavidMansolino:
**Describe the Bug**
Changing the `filtering` field to a negative value in the following ImageTexture node:
```
Floor {
  appearance PBRAppearance {
    baseColorMap ImageTexture {
      url [
        "textures/checkered_marble.jpg"
      ]
    }
  }
}
```

Makes Webots crash with the following stack (on an up to date develop branch):
```
0x000055d06e90ab78 in WbTreeItem::parent (this=0x1000000010000)
    at scene_tree/WbTreeItem.hpp:42
42	  WbTreeItem *parent() const { return mParent; }
(gdb) where
#0  0x000055d06e90ab78 in WbTreeItem::parent() const (this=0x1000000010000)
    at scene_tree/WbTreeItem.hpp:42
#1  0x000055d06e90ecdf in WbSceneTreeModel::itemToIndex(WbTreeItem const*) const (this=0x55d091a32c10, item=0x7f3db400b2c0)
    at scene_tree/WbSceneTreeModel.cpp:210
#2  0x000055d06e909323 in WbSceneTree::applyNodeRegeneration(WbNode*)
    (this=0x55d07056d810, node=0x55d09269da50)
    at scene_tree/WbSceneTree.cpp:1311
```

stefaniapedrazzi:
I tried to reproduce this issue on the develop branch and I get a crash but with a different stack.
```
Thread 1 "webots-bin" received signal SIGSEGV, Segmentation fault.
0x00007ffff305740d in __dynamic_cast () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
(gdb) bt
#0  0x00007ffff305740d in __dynamic_cast () at /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x0000555555a78f55 in WbPbrAppearance::roughnessMap() const (this=0x55556158eb40) at nodes/WbPbrAppearance.cpp:307
#2  0x0000555555a77842 in WbPbrAppearance::preFinalize() (this=0x55556158eb40) at nodes/WbPbrAppearance.cpp:87
#3  0x0000555555aedeb0 in WbShape::preFinalize() (this=0x7fffd800bb80) at nodes/WbShape.cpp:69
#4  0x00005555559c48aa in WbGroup::preFinalize() (this=0x555561581e40) at nodes/WbGroup.cpp:67
[...]

PR #1868, opened 2020-07-02
Fix various crashes related to PROTO template regenerator fields

### LINKED ISSUE #1854: Converting PBRAppearance to Base Node Make Webtos Crash
**Describe the Bug**
When converting for example the `floorAppearance` field of a RectangleArena node to base node, Webots crashes with the following stack:

```
#0  0x00007f4a2b769ee2 in __dynamic_cast ()
    at /lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x000056380980706e in WbNodeOperations::importNode(WbNode*, WbField*, int, QString const&, QString const&, bool, int*, bool)
    (this=0x56380c0887c0, parentNode=0x5638117b8020, field=0x5638115062e0, itemIndex=0, filename=..., nodeString=..., avoidIntersections=false, importedNodesNumber=0x0, fromSupervisor=false) at nodes/utils/WbNodeOperations.cpp:121
#2  0x000056380988420c in WbSceneTree::convertProtoToBaseNode(bool)
    (this=0x56380b9b61b0, rootOnly=false) at scene_tree/WbSceneTree.cpp:727
#3  0x0000563809883d9b in WbSceneTree::convertToBaseNode()
    (this=0x56380b9b61b0) at scene_tree/WbSceneTree.cpp:691
```

stefaniapedrazzi:
Fix #1854:
if the PROTO node that is going to be converted is contained in a PROTO template regenerator field, the parent node and parent field pointers will become invalid just after deleting the previous node and will cause memory issues when inserting the new one.
To overcome this issue, the template regeneration triggered by the delete should be blocked (same approach also used for example in the `WbSceneTree::reset` method). So that then PROTO will only be regenerated once after the new node is imported.


stefaniapedrazzi:
I also fixed another crash when deleting a descendant node of a template PROTO parameter cause by a wrong computation of the currently selected node.

To reproduce the issue (before b7f80bb):
1. select the Parquetry appearance node
2. delete the node without changing the selection
3. add a new appearance PROTO node without changing the selection
-> webots crashes

```
Container {
  slot [
    Shape {
      appearance Parquetry {
        type "chequered"
      }
      geometry Box {
        size 0.1 0.1 0.1
      }
    }
  ]
}
```
Container.proto:
```
#VRML_SIM R2020b utf8
# license: Copyright Cyberbotics Ltd. Licensed for use only with Webots.
# tags: static
# Cotnainer.

PROTO Container [
  field MFNode slot []
]
{
  %{
    io.stderr:write("'slot' template field.\n")
  }%
  Solid {
    children IS slot
  }
}
```

Stack:
```
0x0000555555ae258a in WbTreeItem::parent (this=0x4) at scene_tree/WbTreeItem.hpp:42
42	  WbTreeItem *parent() const { return mParent; }
(gdb) bt
#0  0x0000555555ae258a in WbTreeItem::parent (this=0x4) at scene_tree/WbTreeItem.hpp:42
#1  0x0000555555ae641b in WbSceneTreeModel::itemToIndex (this=0x555562688e40, item=0x555563b51af0) at scene_tree/WbSceneTreeModel.cpp:210
#2  0x0000555555ae0e23 in WbSceneTree::applyNodeRegeneration (this=0x5555569c2230, node=0x555560fad080) at scene_tree/WbSceneTree.cpp:1320
```

DavidMansolino:
The stack looks very similar to this one: #1853
Maybe your fix closes this issue too :smiley:

stefaniapedrazzi:
Issue #1853 should be fixed in commit eaf7f47:
if some fields are not valid, a regenerated node could be triggered for regeneration again in the `finalize` step.
This will make many pointers invalid not just after but also during the `finalize` step.
Blocking the regeneration during the finalize step and activating it again just after (and skipping the post node regeneration instructions) prevent accessing invalid pointer and notifying the creation of the last generated node only.
Additionally to restore correctly the scene tree selection, the `preNodeRegeneration` signal has to be skipped when regenerating the same node for the second time.

stefaniapedrazzi:
It just remain to fix issues with resize and rescale handles that are in some cases still displayed for template regenerator fields and causes some crashes.

Then I will try to add some tests at least for #1853.

stefaniapedrazzi:
I added a test simulation with two different checks for crashes, but currently only the first one is executed due to #1885.
Once #1885 is fixed, the second test will work as well.

Then , I decided to fix issues with resize and rescale handles in a different PR as it is more complicated than expected.

DavidMansolino:
:eyes:

tests/protos/controllers/modify_proto_template_field/.gitignore
```diff
@@ -0,0 +1 @@
+modify_proto_template_field
```
DavidMansolino, line 1:
```suggestion
/modify_proto_template_field
```

Just to be coherent with the other gitignore.

tests/protos/protos/TestRectangleArena.proto
```diff
@@ -0,0 +1,109 @@
+#VRML_SIM R2020b utf8
+# license: Copyright Cyberbotics Ltd. Licensed for use only with Webots.
+# license url: https://cyberbotics.com/webots_assets_license
+# tags: static
+# A configurable rectangle arena composed of a floor surrounded by walls.
+
+PROTO TestRectangleArena [
```
DavidMansolino, line 7: I am not sure to understand why we need to copy this PROTO here?

stefaniapedrazzi: When we will apply the modification described in #1867, the `floorAppearance` field of the default RectangleArena Proto won't be any longer a template regeneration field and thus this test will be useless.
So in this case it is better to use a custom project PROTO to be sure that also in the future the correct case will be tested. Note that this test would also change if the depracation message will be removed. 

But I will update this PROTO so that even after #1867 it will remain a regeneration field (currently it is not the case).

stefaniapedrazzi:
I updated the tests, but now #1895 should be merged before the test passes successfully.

omichel:
:eyes:

tests/protos/protos/TestRectangleArena.proto
```diff
@@ -0,0 +1,107 @@
...
+  field SFFloat    wallThickness   0.01                            # Defines the thickness of the walls.
+  field SFFloat    wallHeight      0.1                             # Defines the height of the walls.
+  field SFNode     wallAppearance  BrushedAluminium {}             # Defines the appearance of the wall.
+
+  # Deprecated in R2019a
+  hiddenField MFString   floorTextureUrl "textures/checkered_parquetry.jpg"
+  hiddenField MFString   wallTextureUrl  "textures/metal.jpg"
+  # Deprecated in R2019a-rev2
+  hiddenField SFColor    wallColor 0.8 0.8 0.8
+  hiddenField SFVec2f    wallTileSize 0.5 0.5
```
omichel, line 24: Can't we remove these deprecated fields? It seems they are not useful in this test.

stefaniapedrazzi: Yes, I will remove them

tests/manual_tests/worlds/modify_proto_template_field.wbt
```diff
@@ -0,0 +1,23 @@
+#VRML_SIM R2020b utf8
+WorldInfo {
+  info [
+    "Test procedure:"
+    "- convert to base the Floor > appearance Parquetry PROTO node"
```
omichel, line 5:
```suggestion
    "- convert to base the TestFloor > appearance Parquetry PROTO node"
```

stefaniapedrazzi:
I simplified as much as possible the PROTO models. Worlds and controller are already as simple as possible to test the functionality.
Some of the crashes I got when starting working on this issue were not reproducible with simple version of the PROTO models, but I checked and the tested cases only works correctly after the recent fixed to PROTO regeneration.

PR #1879, opened 2020-07-07
Added correct joint limits

Simon-Steinmann:
**Description**
The joint limits in the existing files dont represent the actual range of motion. The changes of limits are according to the urdf definitions of robotiq.
https://github.com/ros-industrial/robotiq/tree/kinetic-devel/robotiq_3f_gripper_visualization/cfg

I could provide a simple example controller, performing a gripping motion as can be seen in this link:
https://drive.google.com/file/d/1GZdvlzdHxuJpFMCqYNmrPTi5M4oW6Sr4/view?usp=sharing


Simon-Steinmann:
I created a demo controller, that allows pinch gripping, grip width adjustment, and torque-feedback wrap grasping.
demo:
https://drive.google.com/file/d/13hY_Ttp_FDaW3gF6Mtc-Wq13i1ccEp5a/view?usp=sharing
torque feedback demo:
https://drive.google.com/file/d/17GpVFXbi8F3jN27GxZOx2o9D6U1d2ruV/view?usp=sharing

The controller only interacts with the gripper. I can share it, but I have no idea where I would upload it.

DavidMansolino:
:eyes:

projects/devices/robotiq/protos/Robotiq3fGripper.proto
```diff
@@ -1,11 +1,11 @@
-#VRML_SIM R2020a utf8
+#VRML_SIM R2020b utf8
 # license: Apache License 2.0
 # license url: http://www.apache.org/licenses/LICENSE-2.0
 # documentation url: https://www.cyberbotics.com/doc/guide/universal_robots
 # A 3-finger adaptive robot gripper from ROBOTIQ.
 
 PROTO Robotiq3fGripper [
-  field   SFVec3f    translation    0 0 0
+  field   SFVec3f    translation    0 0.04 0
```
DavidMansolino, line 8: Why changing the default translation?

projects/devices/robotiq/protos/Robotiq3fGripper.proto
```diff
@@ -1,11 +1,11 @@
-#VRML_SIM R2020a utf8
+#VRML_SIM R2020b utf8
```
DavidMansolino, line 1:
```suggestion
#VRML_SIM R2020a utf8
```

Please do not update the header when targeting the master branch.

projects/devices/robotiq/protos/Robotiq3fGripper.proto
```diff
@@ -76,16 +76,17 @@ PROTO Robotiq3fGripper [
               device [
                 RotationalMotor {
                   name "finger_1_joint_1"
-                  maxVelocity 100
-                  maxPosition 3.1416
-                  maxTorque 100
+                  maxVelocity 100.0
+                  maxPosition 1.1723
```
DavidMansolino, line 80: Can you explain the difference with https://github.com/ros-industrial/robotiq/blob/kinetic-devel/robotiq_3f_gripper_visualization/cfg/robotiq-3f-gripper_articulated.urdf#L109 ?

projects/devices/robotiq/protos/Robotiq3fGripper.proto
```diff
@@ -95,9 +96,9 @@ PROTO Robotiq3fGripper [
                     device [
                       RotationalMotor {
                         name "finger_1_joint_2"
-                        maxVelocity 100
-                        maxPosition 3.1416
-                        maxTorque 100
+                        maxVelocity 100.0
+                        maxPosition 1.5708
```
DavidMansolino, line 100: Same here: https://github.com/ros-industrial/robotiq/blob/kinetic-devel/robotiq_3f_gripper_visualization/cfg/robotiq-3f-gripper_articulated.urdf#L116

projects/devices/robotiq/protos/Robotiq3fGripper.proto
```diff
@@ -115,16 +116,17 @@ PROTO Robotiq3fGripper [
                           device [
                             RotationalMotor {
                               name "finger_1_joint_3"
-                              maxVelocity 100
-                              maxPosition 3.1416
-                              maxTorque 100
+                              maxVelocity 100.0
+                              minPosition -1.1694
```
DavidMansolino, line 120: And here.

projects/devices/robotiq/protos/Robotiq3fGripper.proto
DavidMansolino: Ok I see, this comes from https://github.com/ros-industrial/robotiq/blob/66961ec6b6c9c493f13410d16ce32cea1648babf/robotiq_3f_gripper_visualization/cfg/robotiq-3f-gripper_articulated_macro.xacro#L49 right?

projects/devices/robotiq/protos/Robotiq3fGripper.proto
DavidMansolino: Come from here right?
https://github.com/ros-industrial/robotiq/blob/kinetic-devel/robotiq_3f_gripper_visualization/cfg/robotiq-3f-gripper_finger_articulated_macro.xacro#L120

projects/devices/robotiq/protos/Robotiq3fGripper.proto
DavidMansolino: But this one doesn't match:
https://github.com/ros-industrial/robotiq/blob/66961ec6b6c9c493f13410d16ce32cea1648babf/robotiq_3f_gripper_visualization/cfg/robotiq-3f-gripper_finger_articulated_macro.xacro#L127

projects/devices/robotiq/protos/Robotiq3fGripper.proto
```diff
@@ -353,16 +356,17 @@ PROTO Robotiq3fGripper [
                           device [
                             RotationalMotor {
                               name "finger_2_joint_3"
-                              maxVelocity 100
-                              maxPosition 3.1416
-                              maxTorque 100
+                              maxVelocity 100.0
+                              minPosition -1.1694
```
DavidMansolino, line 360: Doesn't match too:
https://github.com/ros-industrial/robotiq/blob/66961ec6b6c9c493f13410d16ce32cea1648babf/robotiq_3f_gripper_visualization/cfg/robotiq-3f-gripper_finger_articulated_macro.xacro#L127

projects/devices/robotiq/protos/Robotiq3fGripper.proto
```diff
@@ -498,16 +503,17 @@ PROTO Robotiq3fGripper [
                         device [
                           RotationalMotor {
                             name "finger_middle_joint_3"
-                            maxVelocity 100
-                            maxPosition 3.1416
-                            maxTorque 100
+                            maxVelocity 100.0
+                            minPosition -1.1694
```
DavidMansolino, line 507: Doesn't match too:
https://github.com/ros-industrial/robotiq/blob/66961ec6b6c9c493f13410d16ce32cea1648babf/robotiq_3f_gripper_visualization/cfg/robotiq-3f-gripper_finger_articulated_macro.xacro#L127

projects/devices/robotiq/protos/Robotiq3fGripper.proto
Simon-Steinmann: The finger limits are specifically from here: https://github.com/ros-industrial/robotiq/blob/66961ec6b6c9c493f13410d16ce32cea1648babf/robotiq_3f_gripper_visualization/cfg/robotiq-3f-gripper_finger_articulated_macro.xacro
Since I changed it so either max or min position is 0, one value is subtracted from the other. Here  lower="0.0495" upper="1.2218" turns to  lower="0.0" upper="1.1723"

projects/devices/robotiq/protos/Robotiq3fGripper.proto
Simon-Steinmann: yes

projects/devices/robotiq/protos/Robotiq3fGripper.proto
Simon-Steinmann: it's lower - upper
`lower="-1.2217" upper="-0.0523" ` --> `lower="-1.1694" upper="0"   `
together with  line 129, 
`rotation 0 0 1 -0.0523`
this makes it correct

projects/devices/robotiq/protos/Robotiq3fGripper.proto
Simon-Steinmann: Same reasoning as above, lower - upper and rotating endpoint by lower amount

projects/devices/robotiq/protos/Robotiq3fGripper.proto
Simon-Steinmann: Same reasoning as above, lower - upper and rotating endpoint by lower amount

projects/devices/robotiq/protos/Robotiq3fGripper.proto
Simon-Steinmann: To correctly attach the gripper. The base link origin is in the middle of the gripper:
image

projects/devices/robotiq/protos/Robotiq3fGripper.proto
Simon-Steinmann: Was on accident, will look out for it in the future

projects/devices/robotiq/protos/Robotiq3fGripper.proto
DavidMansolino: Ok, then the idea would be to keep the original limit and instead of putting the rotation to set the `jointparameter.position` field. I will give it a try.

DavidMansolino:
@Simon-Steinmann  I have slightly improved your PR, let me know if my changes are ok for you.

DavidMansolino:
I am merging this PR as we have no news.

PR #1884, opened 2020-07-08
Add URDF `base_footprint` link

lukicdarkoo:
**Description**
Turtlebot3 and Husky have `base_footprint` link as a parent:
https://github.com/husky/husky/blob/4647d303365b17ccb455f4345bd14320f49f1134/husky_description/urdf/husky.urdf.xacro#L95
https://github.com/ROBOTIS-GIT/turtlebot3/blob/41e43d4a10aada68b8a7d05675550550091586d6/turtlebot3_description/urdf/turtlebot3_burger.urdf#L47
Probably, it is similar to the other robots as well.

This is important because some packages depend on `base_footprint` link, e.g. `turtlebot3_navigation2`:
https://github.com/ROBOTIS-GIT/turtlebot3/blob/41e43d4a10aada68b8a7d05675550550091586d6/turtlebot3_navigation2/param/burger.yaml#L9

Usually, this can be changed by overriding configuration or by adding static transform in a launch file. However, this may be a cleaner solution and closer to what the community has been doing.

**Related Issues**
https://github.com/cyberbotics/webots_ros2/pull/110

**Tasks**
  - [x] Add `base_footprint` link
  - [x] Update tests

### DIFF
--- a/src/webots/nodes/WbRobot.cpp
+++ b/src/webots/nodes/WbRobot.cpp
@@ -1435,30 +1435,31 @@ const QString WbRobot::urdfName() const {
 }
 
 void WbRobot::writeExport(WbVrmlWriter &writer) const {
-    writer.increaseIndent();
-    writer.indent();
-    writer << QString("<link name=\"%1\"></link>\n").arg(getUrdfPrefix() + "base_footprint");
-    writer.indent();
-    writer << QString("<joint name=\"%1\" type=\"fixed\">\n").arg(getUrdfPrefix() + "base_joint");
-    writer.increaseIndent();
-    writer.indent();
-    writer << QString("<parent link=\"%1\"/>\n").arg(getUrdfPrefix() + "base_link");
-    writer.indent();
-    writer << QString("<child link=\"%1\"/>\n").arg(getUrdfPrefix() + "base_footprint");
-    writer.indent();
-    writer << QString("<axis xyz=\"0 0 0\"/>\n");
-    writer.indent();
-    writer << QString("<origin xyz=\"0 0 0\" rpy=\"0 0 0\"/>\n");
-    writer.decreaseIndent();
-    writer.indent();
-    writer << QString("</joint>\n");
+  writer.increaseIndent();
+  writer.indent();
+  writer << QString("<link name=\"%1\"></link>\n").arg(getUrdfPrefix() + "base_footprint");
 
-    writer.indent();
-    writer << QString("<link name=\"%1\">\n").arg(getUrdfPrefix() + "base_link");
-    exportNodeSubNodes(writer);
-    writer.indent();
-    writer << QString("</link>\n");
-    writer.decreaseIndent();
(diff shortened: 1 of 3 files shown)


DavidMansolino:
I looked a bit if this 'artificial' was a standard and I found a reference in the ROS REP:
https://www.ros.org/reps/rep-0120.html#base-footprint
> The base_footprint is the representation of the robot position on the floor. The floor is usually the level where the supporting leg rests, i.e. z = min(l_sole_z, r_sole_z) where l_sole_z and r_sole_z are the left and right sole height respecitvely. The translation component of the frame should be the barycenter of the feet projections on the floor. With respect to the odom frame, the roll and pitch angles should be zero and the yaw angle should correspond to the base_link yaw angle.

It seems that the current implementation does not respect the fact that the position should be the barycenter of the robot. Did you found some urdf files where the location is not [0,0,0]? I also checked a few URDF files, and this link is present in most of them but not 100%.

lukicdarkoo:
> Did you found some urdf files where the location is not [0,0,0]?

Both, [Turtlebot3](https://github.com/ROBOTIS-GIT/turtlebot3/blob/41e43d4a10aada68b8a7d05675550550091586d6/turtlebot3_description/urdf/turtlebot3_burger.urdf#L52) and [Husky](https://github.com/husky/husky/blob/4647d303365b17ccb455f4345bd14320f49f1134/husky_description/urdf/husky.urdf.xacro#L98), have a slightly translated `base_footprint` in respect to `base_link`. 

> It seems that the current implementation does not respect the fact that the position should be the barycenter of the robot.

Can we read the centre of mass of the robot and shift the `base_footprint` accordingly?

> I also checked a few URDF files, and this link is present in most of them but not 100%.

I guess not everybody respects this. I picked Turtlebot3 and Husky because they supposed to be a good reference for ROS2 mobile robots.

The only problem I see currently (without `base_footprint`) is incompatibility with `turtlebot3_navigation2` and probably similar packages. This can be solved without changes in Webots, but I am not sure what is the best solution.

DavidMansolino:
> Can we read the centre of mass of the robot and shift the base_footprint accordingly?

That looks indeed like a good tradeoff.

> The only problem I see currently (without base_footprint) is incompatibility with turtlebot3_navigation2 and probably similar packages. This can be solved without changes in Webots, but I am not sure what is the best solution.

Yes, I was also hesitating as this is indeed very simple to fix on the webots_ros side.
Maybe we can put this on hold for know and discuss about this during the Monday meeting?

lukicdarkoo:
> Maybe we can put this on hold for know and discuss about this during the Monday meeting?

Monday sounds good, we may come with better solutions by then or at the meeting.

lukicdarkoo:
As decided, we will add `base_footprint` link in PROTOs when necessary. The produced URDF will be the same, but `base_footprint` will not be added when unnecessary. Therefore, I am closing this PR.

PR #1888, opened 2020-07-08
Improve MultiSense S21 internal names description

DavidMansolino:
Improve MultiSense S21 internal names description

https://cyberbotics.com/doc/guide/camera-sensors?version=DavidMansolino-patch-1#multisense-s21


docs/guide/camera-sensors.md
```diff
@@ -29,7 +29,7 @@ The left and right RGB images of the stereoscopic camera can be retrieved indepe
 Alternatively, a RGB [Camera](../reference/camera.md) and a [RangeFinder](../reference/rangefinder.md) device can be enabled at the center of the `MultiSense S21` device.
 These two "meta" devices are present to mimic the cloud of 3D colored points generated by the real device.
 
-The internal device names used to retrieve the devices from the controller can be computed from the `MultiSenseS21.name` field using [the following table](#internal-devices-names):
+The internal device names used to retrieve the devices from the controller can be computed from the `MultiSenseS21.name` field using [the following table](#internal-devices-names) (the `MultiSenseS21.name` part should be replaced by the actual value of the `name` field):
```
stefaniapedrazzi, line 32: Thank you!

Maybe we could also add an example
```suggestion
The internal device names used to retrieve the devices from the controller can be computed from the `MultiSenseS21.name` field using [the following table](#internal-devices-names), where the `MultiSenseS21.name` part should be replaced by the actual value of the `name` field.
For example if `MultiSenseS21.name` is set to "MultiSense S21" the inertial unit device name is "MultiSense S21  inertial unit".
```

DavidMansolino: Good idea.

omichel: There is a double space here between "S21" and "inertia unit". Is that correct?

stefaniapedrazzi: No, there should be only one space

PR #1893, opened 2020-07-09
Fix Github Actions

DavidMansolino:
Without the package update, the installation of `libpulse0` is failing.


DavidMansolino:
Installation step succedded: https://github.com/cyberbotics/webots/runs/853427619?check_suite_focus=true#step:5:1

PR #1895, opened 2020-07-09
Fix node list in supervisor controller after PROTO regeneration

### LINKED ISSUE #1885: References to PROTO regenerated nodes not updated in Supervisor API
After regenerating a PROTO node, the reference in the Supervisor API doesn't seem to be updated and `wb_supervisor_node_get_from_def` still returns the previous (before PROTO regeneration) node reference.

World file:
```
#VRML_SIM R2020b utf8
WorldInfo {
}
Viewpoint {
}
RectangleArena {
  floorAppearance PBRAppearance {
    baseColorMap DEF TEST_NODE ImageTexture {
      url [
        "textures/parquetry/chequered_parquetry_base_color.jpg"
      ]
    }
  }
}
Robot {
  controller "modify_proto_template_field"
  supervisor TRUE
}
```

C Controller:
```
#include <stdio.h>
#include <webots/robot.h>
#include <webots/supervisor.h>

#define TIME_STEP 32

  wb_robot_init();
  WbNodeRef node = wb_supervisor_node_get_from_def("TEST_NODE");
  printf("Node before: %p \n", node);
  WbFieldRef filteringField = wb_supervisor_node_get_field(node, "filtering");
  
  // If the url field reference is requested before regeneration,
  // then it is cached and returned after regeneration even if invalid
  //WbFieldRef urlFieldBefore = wb_supervisor_node_get_field(node, "url");
  //printf("Url field before: %p \n", urlFieldBefore);
  

  wb_robot_step(TIME_STEP);
  
  // trigger RectangleArena regeneration
  wb_supervisor_field_set_sf_int32(filteringField, 2);
  
  wb_robot_step(TIME_STEP);
  
  node = wb_supervisor_node_get_from_def("TEST_NODE");
  printf("Node after: %p \n", node);
  // Invalid node reference returned and url field is not found 
  WbFieldRef urlField = wb_supervisor_node_get_field(node, "url");
  printf("Url field after: %p \n", urlField);

  wb_robot_cleanup();
  return 0;
}
```

stefaniapedrazzi:
Fix #1885.

Nodes deleted during PROTO regeneration where not propagated to the controller that was still returning invalid references.
Instead of looping through all the PROTO subnodes, it seems more efficient to emit the `WbNodeOperations::nodeDeleted` signal directly from the `WbBaseNode` destructor.

A second issues was caused by notifying the deletion of regenerated PROTO nodes to the supervisor controller in two steps:
1) add node to the `WbSupervisorUtilities::mNodesDeletedSinceLastStep`
2) propagate the `WbSupervisorUtilities::mNodesDeletedSinceLastStep` to the controller
At step 1), the node reference was still valid, but the same was not guaranteed at step 2).
So it is better to compute all the requires node attributes at step 1 and store them directly in the list, so that even if the node is invalid at step 2) it doesn't cause any issue.


stefaniapedrazzi:
The correct run of controllers in case of regenerated PROTO nodes is based on the assumption that a node unique ID doesn't change after regeneration.
This is true for the regenerated node itself, but not for PROTO parameter nodes (for example node in a `slot` parameter).

stefaniapedrazzi:
Replaced by #1922.

Issue #1896, opened 2020-07-09
Assertion Failure in 'conveyor_belt' Controller


DavidMansolino:
**Describe the Bug**
The `conveyor_belt` controller is raising this assertion failure on the develop branch:
```
WARNING: conveyor_belt: The process crashed some time after starting successfully.
conveyor_belt: conveyor_belt.c:26: main: Assertion `argc == 3' failed.
WARNING: 'conveyor_belt' controller crashed
```

PR #1897, opened 2020-07-10
Fix not updated PROTO.

DavidMansolino:
**Description**
The backward compatibility feature was not working on these PROTO files because the header version was preventing the backward compatibility mechanism.

**Related Issues**
This pull-request fixes issue #1896

### DIFF
--- a/projects/objects/factory/conveyors/protos/ConveyorBelt.proto
+++ b/projects/objects/factory/conveyors/protos/ConveyorBelt.proto
@@ -37,7 +37,10 @@ PROTO ConveyorBelt [
     rotation IS rotation
     name IS name
     controller "conveyor_belt"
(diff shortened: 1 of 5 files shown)


stefaniapedrazzi:
:eyes:

DavidMansolino:
The test suite is less problematic because the headers are not updated and therefore the backward compatibility mechanism works fine in those cases.

stefaniapedrazzi:
> The test suite is less problematic because the headers are not updated and therefore the backward compatibility mechanism works fine in those cases.

Yes, I agree.
But it would be better if the test format is up-to-date and we are not using old nodes or fields format.

DavidMansolino:
I completely agree, there is a lot of 'old' stuff in the test suite (old field format, old appearance, etc.) and this is very bad because it does not represent/test the current state of Webots :disappointed:

PR #1900, opened 2020-07-10
Fix Perspective Assertion Failure

DavidMansolino:
**Description**

When the `WEBOTS_DISABLE_SAVE_SCREEN_PERSPECTIVE_ON_CLOSE` was set and the wbproj file was not existing an assertion was raised because the main and view window state were not defined.


stefaniapedrazzi:
:eyes:

Issue #1605, opened 2020-04-29
Outdated libfontconfig version


joecksma:
**Describe the Bug**
The release (webots-R2020a-rev1-x86-64.tar.bz2) ships with an ancient version of libfontconfig, which makes it incompatible with the fontconfig files (/etc/fomts). See error message below. This causes the start of Webots to take about 1 minute. 
Please update libfontconfig to a more recent version.

**Steps to Reproduce**
1. Download latest release *.tar.bz2 
2. extract it
3. start webots (./webots/webots)

**Expected behavior**
Lauch without errors

**System**
 - Operating System: Arch Linux 
 - Graphics Card: NVIDIA GeForce GTX 980
 - fontconfig Version: 2:2.13.1+12+g5f5ec56-1

**Additional context**
Error:
`Fontconfig warning: line 5: unknown element "its:rules"
Fontconfig warning: line 6: unknown element "its:translateRule"
Fontconfig error: line 6: invalid attribute 'translate'
Fontconfig error: line 6: invalid attribute 'selector'
Fontconfig error: line 7: invalid attribute 'xmlns:its'
Fontconfig error: line 7: invalid attribute 'version'
Fontconfig warning: line 9: unknown element "description"
Fontconfig warning: "/etc/fonts/conf.d/10-hinting-slight.conf", line 4: unknown element "its:rules"
Fontconfig warning: "/etc/fonts/conf.d/10-hinting-slight.conf", line 5: unknown element "its:translateRule"
Fontconfig error: "/etc/fonts/conf.d/10-hinting-slight.conf", line 5: invalid attribute 'translate'
Fontconfig error: "/etc/fonts/conf.d/10-hinting-slight.conf", line 5: invalid attribute 'selector'
Fontconfig error: "/etc/fonts/conf.d/10-hinting-slight.conf", line 6: invalid attribute 'xmlns:its'
Fontconfig error: "/etc/fonts/conf.d/10-hinting-slight.conf", line 6: invalid attribute 'version'
Fontconfig warning: "/etc/fonts/conf.d/10-hinting-slight.conf", line 8: unknown element "description"
Fontconfig warning: "/etc/fonts/conf.d/10-scale-bitmap-fonts.conf", line 4: unknown element "its:rules"
Fontconfig warning: "/etc/fonts/conf.d/10-scale-bitmap-fonts.conf", line 5: unknown element "its:translateRule"
[...]

PR #1901, opened 2020-07-10
Update test_suite.yml

DavidMansolino:
**Description**
This PR attempts to use github Action to create the Ubuntu 20.04 release (needed because of libfontconfig:  #1605).

**Tasks**
  - [x] Udpate Workflow.
  - [x] Update webots_distro
  - [x] Fix various compilation issues


DavidMansolino:
Some remaining issues with the 18.04 package: 
https://github.com/cyberbotics/webots/runs/868238199#step:6:4059

DavidMansolino:
Remaining issue: for some reason, the test step is using GLIBC 2.27 but the compilation step includes GLIBC 2.31 which is causing compilation issues for the test suite.

DavidMansolino:
Status:
Now the compilation is done both on Ubuntu 18.04 and 20.04 and the packages are uploaded in different action artifacts: https://github.com/cyberbotics/webots/actions/runs/171713119
Then the test suite is performed on Ubuntu 18.04 and 20.04: https://github.com/cyberbotics/webots/runs/878990546
And the 'update all world' test is done on Ubuntu 18.04 only (because this test is extremely slow and there is no point to run it on different version of Ubuntu).
For know, I kept Ubuntu 16.04 apart because I can't figure out how to run Webots headless in the Github action ubuntu 16.04 image (and I am not sure we really want to invest time for this).

The next step will be to create a new step that uploads the package on Github during the nightly tests and on tag creation.

PR #1916, opened 2020-07-14
Fix supervisor field update.

### LINKED ISSUE #1915: Supervisor Field Need Reset to be Effective
**Describe the Bug**
When the `supervisor` field is set to true, the world needs to be saved and reverted to be effective.
Otherwise the libcontroller gives errors when using the supervisor API.

DavidMansolino:
**Description**
Attempt to fix  #1915


tests/api/controllers/supervisor_field/supervisor_field.c
```diff
@@ -384,6 +384,22 @@ int main(int argc, char **argv) {
...
   wb_robot_step(TIME_STEP);
 
+  // supervisor field update
+  self = wb_supervisor_node_get_self();
+  supervisor_field = wb_supervisor_node_get_field(self, "supervisor");
+  ts_assert_boolean_equal(wb_supervisor_field_get_sf_bool(supervisor_field), "'supervisor' field should be true");
+  ts_assert_boolean_equal(wb_robot_get_supervisor(), "'wb_robot_get_supervisor' field should return true");
+  wb_supervisor_field_set_sf_bool(supervisor_field, false);
+  wb_robot_step(TIME_STEP);
+  wb_robot_step(TIME_STEP);
```
omichel, line 394: Do we really need two time steps here?

DavidMansolino: Yes, this is required because since the field is updated after the end of the step we need a second step to get the update in the controller, but this is bad, I have an idea to improve this.

DavidMansolino:
> I believe we could now get rid of the supervisor value in the configure message. That would simplify things.

That's actually the contrary, this will complexify things, as the `writeAnswer` function is called to answer a controller request, which means that the `supervisor` flag will be false at the first step of the controller. We will, therefore, need to include a mechanism to send the information before the first step (mechanism duplicated with the configure message).

omichel:
>>    I believe we could now get rid of the supervisor value in the configure message. That would simplify things.

> That's actually the contrary, this will complexify things, as the writeAnswer function is called to answer a controller request, which means that the supervisor flag will be false at the first step of the controller. We will, therefore, need to include a mechanism to send the information before the first step (mechanism duplicated with the configure message).

OK, I see. Thank you for the explanation.

tests/api/controllers/supervisor_field/supervisor_field.c
DavidMansolino: This is now fixed, the supervisor state is now up to date after one single step.

PR #1918, opened 2020-07-14
Update test worlds

DavidMansolino:
**Description**
Regenerate all the test world.


tests/api/worlds/camera_color.wbt
```diff
@@ -40,15 +40,17 @@ Robot {
...
       }
     }
     TestSuiteEmitter {
     }
   ]
   controller "camera_checker"
-  controllerArgs "camera_color"
+  controllerArgs [
+    "camera_color"
+  ]
```
DavidMansolino, line 53: Just for the note, we discussed this case on Monday, and it seems Webtos is saving these MFString fields on a new line.

omichel:
I am running the test suite on Windows to check if it still passes...

DavidMansolino:
thank you for the test

PR #1922, opened 2020-07-15
Hotfix unique id after regeneration

### LINKED ISSUE #1885: References to PROTO regenerated nodes not updated in Supervisor API

stefaniapedrazzi:
Fix #1885.

Unique ids of PROTO parameter nodes are not correctly restored after procedural PROTO regeneration.
This causes issues in the controller library where it is assumed that the ids remains the same and it is no longer possible to access the nodes.
(No notification is received in the libController that nodes have been deleted during regeneration and thus it is also not possible to retrieve the new nodes with the new unique ids).

I fixed the unique ids for the PROTO parameters but not for internal PROTO nodes (the ones retrieved by `wb_supervisor_node_get_from_proto_def`). This case is much more complicated and it is even not ensured that the node structure remain the same.

A second issues was caused by notifying the deletion of regenerated PROTO nodes to the supervisor controller in two steps:
1. add node to the WbSupervisorUtilities::mNodesDeletedSinceLastStep
2. propagate the WbSupervisorUtilities::mNodesDeletedSinceLastStep to the controller

At step 1), the node reference was still valid, but the same was not guaranteed at step 2).
Checking if deleted nodes are references by other nodes fixes the issues and prevents to use invalid pointers.


stefaniapedrazzi:
It is not possible to guarantee the validity of internal PROTO fields and nodes after PROTO regeneration on the libController side.
Even detecting exactly which node and field became invalid could be complicated.
Instead, I introduced a message that will invalidate all the internal PROTO fields and nodes after PROTO regeneration without checking if they were descendant of the regenerated node or not.

The libController is now able to detect these invalidated references and returns an error so that the user can retrieve the new reference.

omichel:
:eyes:

src/lib/Controller/api/supervisor.c
```diff
@@ -235,6 +238,46 @@ static const char *extract_node_def(const char *def_name_expression) {
...
+  while (field) {
+    if (field->is_proto_internal) {
+      if (previous_field)
+        previous_field->next = field->next;
+      else
+        field_list = field->next;
+      WbFieldStruct *current_field = field;
+      field = field->next;
+      // clean the field
+      printf("remove_internal field %p\n", current_field);
```
omichel, line 270: We should probably remove this debug statement.
```suggestion
```

stefaniapedrazzi:
Thank you for the review.

I just added a new condition to avoid that an internal proto node could be returned instead of a visible DEF node if sharing the same DEF name.

PR #1927, opened 2020-07-16
Fix video recording with asynchronous controllers.

### LINKED ISSUE #1407: Wrong Video Speed with Asynchronous Controllers
**Describe the Bug**
When a controller synchronization field is set to `FALSE` the speed of the video recorded by Webots is wrong.

**Steps to Reproduce**
1. Open the city.wbt world
2. Set the synchronization field of the vehicle to `FALSE`.
3. Set the world time step to 1.
4. Start the movie recording.
5. Run the simulation for 20 seconds (and make sure simulation runs slower than 1.0x).
6. After 20 seconds stop video recording and check the length of the recorded video, it is far longer than 20 seconds.

**Expected behavior**
The synchronization field of the robot should not have any impact on the video recording.

DavidMansolino:
**Description**
Fix #1407.

Using a counter in the `WbView3D::refresh` to take a video frame is not a good idea as several events can trigger this slot, even in synchornous mode (controller, supervisor world edition, mouse, etc.).

### DIFF
--- a/src/webots/gui/WbVideoRecorder.cpp
+++ b/src/webots/gui/WbVideoRecorder.cpp
@@ -168,18 +168,18 @@ bool WbVideoRecorder::setMainWindowFullScreen(bool fullScreen) {
 }
 
 void WbVideoRecorder::estimateMovieInfo(double basicTimeStep) {
-  int roundedBasicTimeStep = round(basicTimeStep);
-  double refresh = EXPECTED_FRAME_STEP / (double)roundedBasicTimeStep;
-  int floorRefresh = floor(refresh);
-  int ceilRefresh = ceil(refresh);
-  int frameStep0 = floorRefresh * roundedBasicTimeStep;
-  int frameStep1 = ceilRefresh * roundedBasicTimeStep;
+  const int roundedBasicTimeStep = round(basicTimeStep);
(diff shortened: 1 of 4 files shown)


omichel:
Maybe we should add a line in the change log about it...

src/webots/gui/WbVideoRecorder.cpp
```diff
@@ -177,10 +177,10 @@ void WbVideoRecorder::estimateMovieInfo(double basicTimeStep) {
 
   if (frameStep0 == 0 || abs(frameStep0 - EXPECTED_FRAME_STEP) > abs(frameStep1 - EXPECTED_FRAME_STEP)) {
     mMovieFPS = 1000.0 / frameStep1;
-    cDisplayRefresh = ceilRefresh * mVideoAcceleration;
```
stefaniapedrazzi, line 180: is the video acceleration used somewhere else than for displaying the label?
is this really correct?

DavidMansolino: Good point, I tested and it was working, but now I am not so sure, I will retry.

stefaniapedrazzi: In any case, in the current state the `mVideoAcceleration` variable is useless and could be removed.

PR #1928, opened 2020-07-16
Fix Webots<=>Controller Protocol

DavidMansolino:
**Description**
There are some issues in the protocol to communicate between Webots and the libController.
This explains the nightly failure of the ROS test.


DavidMansolino:
:heavy_check_mark:: travis_unit_release

src/lib/Controller/api/remote_control.c
```diff
@@ -375,11 +375,6 @@ static void handleRobotMessage(WbRequest *r, unsigned char c) {
     case C_ROBOT_SET_BATTERY_SAMPLING_PERIOD:
       CALL_INTERFACE_FUNCTION(wbr_robot_battery_set_sampling_period, request_read_uint16(r));
       break;
-    case C_ROBOT_SET_KEYBOARD_SAMPLING_PERIOD: {
```
stefaniapedrazzi, line 378: This message is used in WbRobot and set in `messages.h`:
https://github.com/cyberbotics/webots/blob/36205efe6601985efee1c891b45cad68e965967d/src/webots/nodes/WbRobot.cpp#L864
https://github.com/cyberbotics/webots/blob/36205efe6601985efee1c891b45cad68e965967d/src/lib/Controller/api/messages.h#L53

is it correct to remove it?

DavidMansolino: Yes, because the only time that this message is added in a request is here:
https://github.com/cyberbotics/webots/blob/048f65e8f7d7663aecbf3c1ce065ef191abd2c54/src/lib/Controller/api/keyboard.c#L51
This means that it is only a message that goes from libController to Webots, there is, therefore, no need to check if the request coming from Webots in the libController contains this key.

stefaniapedrazzi: Ok, thank you for the explanation.
So then this also applies to `C_ROBOT_SET_BATTERY_SAMPLING_PERIOD` (and maybe others).

DavidMansolino: Well spotted, I just checked all the one-way messages and found one more useless case.

stefaniapedrazzi:
:eyes:

PR #1933, opened 2020-07-17
Fixed Background VRML Export.

### LINKED ISSUE #1866: VRML Export-Import Fail with Background Node
**Describe the Bug**
When exporting a node to VRML and then re-importing this node, the importation of the background node fails with this warning:
```
WARNING: Background: Cannot load texture '/home/david/webots/projects/samples/devices/worlds/textures/cubic/mountains_right.hdr': Unsupported image format.
```

DavidMansolino:
**Description**
Fix #1866.

### DIFF
--- a/src/webots/nodes/WbBackground.cpp
+++ b/src/webots/nodes/WbBackground.cpp
@@ -467,7 +467,7 @@ void WbBackground::exportNodeFields(WbVrmlWriter &writer) const {
     }
   } else if (writer.isVrml()) {
     for (int i = 0; i < 6; ++i) {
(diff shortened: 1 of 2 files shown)


src/webots/nodes/WbBackground.cpp
```diff
@@ -469,7 +469,7 @@ void WbBackground::exportNodeFields(WbVrmlWriter &writer) const {
     for (int i = 0; i < 6; ++i) {
       if (!irradianceFileNames[i].isEmpty()) {
```
omichel, line 470:
```suggestion
      if (!backgroundFileNames[i].isEmpty()) {
```

PR #1934, opened 2020-07-17
Fixed libController assertion failures in release mode

### LINKED ISSUE #1891: Assertion Failure on Windows Release
**Describe the Bug**
It is not the first time (#1881) that a user is reporting an assertion failure with a release version of Webots on Windows.
In release mode, no assertion failure should be raised.

This is not good, not only because we raise assertions for the end-user, but also because it means that the assertions are evaluated even in release mode (which impacts the performance).

omichel:
Fixes #1891.

### DIFF
--- a/src/lib/Controller/api/robot_private.h
+++ b/src/lib/Controller/api/robot_private.h
@@ -23,15 +23,15 @@
 #include "request.h"
 #include "webots/nodes.h"
 
-#ifndef NDEBUG
+#ifdef NDEBUG
+#define ROBOT_ASSERT(condition) \
+  {}
+#else
 #define ROBOT_ASSERT(condition)                                                     \
   {                                                                                 \
     if (!(condition))                                                               \
       robot_abort("%s:%d: assertion failed: %s\n", __FILE__, __LINE__, #condition); \
   }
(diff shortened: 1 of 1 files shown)


src/lib/Controller/api/robot_private.h
```diff
@@ -23,11 +23,16 @@
...
+#ifndef NDEBUG
 #define ROBOT_ASSERT(condition)                                                     \
   {                                                                                 \
     if (!(condition))                                                               \
       robot_abort("%s:%d: assertion failed: %s\n", __FILE__, __LINE__, #condition); \
   }
+#else
+#define ROBOT_ASSERT(condition) \
+  {}
+#endif
```
DavidMansolino, line 35:
```suggestion
#ifdef NDEBUG
#define ROBOT_ASSERT(condition) \
  {}
#else
#define ROBOT_ASSERT(condition)                                                     \
  {                                                                                 \
    if (!(condition))                                                               \
      robot_abort("%s:%d: assertion failed: %s\n", __FILE__, __LINE__, #condition); \
  }
#endif
```

Just to avoid double negation.

PR #1961, opened 2020-07-22
Update installation-procedure.md

omichel:
Fixed installation procedure on Linux servers.

### DIFF
--- a/docs/guide/installation-procedure.md
+++ b/docs/guide/installation-procedure.md
@@ -150,6 +150,7 @@ Webots requires some graphical features that are usually not available by defaul
 
 - `xserver-xorg-core`
 - `libpulse0`
+- `libxkbcommon-x11-dev`
 
 These packages are automatically installed when using the Debian package, but in case of the tarball package the user has to manually install them.
 


DavidMansolino:
Just for the note, we can probably take inspiration from this list (github action for a Ubuntu-server container):
https://github.com/cyberbotics/webots/blob/develop/.github/workflows/test_suite.yml#L89

omichel:
I am not sure we want to write such a list in the documentation and maintain it up-to-date...
Wouldn't it be better to refer to this test_suite.yml file instead or have this list of dependencies listed in a separate file and included by test_suite.xml?

DavidMansolino:
I like the idea of having a list of dependencies listed in a separate file and included by test_suite.xml!

omichel:
I believe we should add a couple of bash scripts in `webots/src/install_scripts` that will install the necessary Debian packages.
E.g.: `linux_server_development_installer.sh` which would in turn call `linux_server_runtime_installer.sh`.
The first one would be installing all the packages needed for a development environment while the second one would be installing only the packages needed at run-time by Webots.
Theses scripts should be used by our CI files (including `test_suite.yml`), so that they tested on a daily basis.
We should refer to them in the development Wiki and documentation instead of trying (and probably failing) to maintain a list of packages in the documentation.

DavidMansolino:
@omichel I see you unassigned yourself from this issue/PR, would you like me to take it?

omichel:
Yes, please, as I am not sure what's the best way to implement it.

DavidMansolino:
Replaced by #2008

PR #1991, opened 2020-07-24
Notify user if stream mode is not valid

### LINKED ISSUE #1981: Stream mode is not verified
**Describe the Bug**

**Steps to Reproduce**
Execute the command:
```
webots --stream="mode=adasds;port=1233;sdadsa=asdasd"
```

**Expected behavior**
```
ERROR: Streaming server: unknown mode 'adasds'
ERROR: Streaming server: unknown option 'sdadsa=asdasd'
INFO: Streaming server listening on port 1233.
```
but there is no warning about streaming mode:
```
ERROR: Streaming server: unknown mode 'adasds'
```

**System**
 - Graphics Card: Intel UHD Graphics 620

lukicdarkoo:
**Description**
Fixes #1981, by providing an error in Webots console.

**Related Issues**
Closes #1981

### DIFF
--- a/src/webots/gui/WbGuiApplication.cpp
+++ b/src/webots/gui/WbGuiApplication.cpp
@@ -122,7 +122,7 @@ void WbGuiApplication::restart() {
 #endif
 }
 
-void WbGuiApplication::parseStreamArguments(const QString &serverArguments) {
+void WbGuiApplication::parseStreamArguments(const QString &streamArguments) {
   bool monitorActivity = false;
   bool disableTextStreams = false;
   bool ssl = false;
@@ -133,7 +133,7 @@ void WbGuiApplication::parseStreamArguments(const QString &serverArguments) {
 #ifdef __APPLE__
   const QStringList &options = argument.split(';', QString::SkipEmptyParts);
 #else  //  Qt >= 5.15
-  const QStringList &options = serverArguments.split(';', Qt::SkipEmptyParts);
+  const QStringList &options = streamArguments.split(';', Qt::SkipEmptyParts);
 #endif
   foreach (QString option, options) {
     option = option.trimmed();
--- a/src/webots/gui/WbGuiApplication.cpp
+++ b/src/webots/gui/WbGuiApplication.cpp
@@ -122,7 +122,7 @@ void WbGuiApplication::restart() {
 #endif
 }
 
-void WbGuiApplication::parseStreamArguments(const QString &serverArgument) {
+void WbGuiApplication::parseStreamArguments(const QString &serverArguments) {
   bool monitorActivity = false;
   bool disableTextStreams = false;
   bool ssl = false;
--- a/src/webots/gui/WbGuiApplication.cpp
+++ b/src/webots/gui/WbGuiApplication.cpp
@@ -183,12 +183,12 @@ void WbGuiApplication::parseStreamArguments(const QString &serverArguments) {
     return;
   }
   if (mode == "mjpeg") {
-    mStreamingServer = new WbMultimediaStreamingServer();
-    mStreamingServer->start(port, monitorActivity, disableTextStreams, ssl, controllerEdit);
+    mStreamingServer = new WbMultimediaStreamingServer(monitorActivity, disableTextStreams, ssl, controllerEdit);
+    mStreamingServer->start(port);
     return;
   }
-  mStreamingServer = new WbX3dStreamingServer();
-  mStreamingServer->start(port, monitorActivity, disableTextStreams, ssl, controllerEdit);
+  mStreamingServer = new WbX3dStreamingServer(monitorActivity, disableTextStreams, ssl, controllerEdit);
+  mStreamingServer->start(port);
   WbWorld::enableX3DStreaming();
 }
 
--- a/src/webots/gui/WbGuiApplication.cpp
+++ b/src/webots/gui/WbGuiApplication.cpp
@@ -164,11 +164,11 @@ void WbGuiApplication::parseStreamArguments(const QString &serverArgument) {
         }
       } else if (key == "mode") {
         if (value != "x3d" && value != "mjpeg") {
-          mode = value;
           cout << tr("webots: invalid option: '%1' in --stream").arg(serverArgument).toUtf8().constData() << endl;
           cout << tr("webots: stream mode can only be x3d or mjpeg").toUtf8().constData() << endl;
           mTask = FAILURE;
-        }
+        } else
+          mode = "mjpeg";
       } else {
         cout << tr("webots: unknown option: '%1' in --stream").arg(serverArgument).toUtf8().constData() << endl;
         mTask = FAILURE;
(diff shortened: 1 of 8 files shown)


lukicdarkoo:
I simply followed the implementation for the invalid port (if the port is invalid it will fallback to the default one). But, I agree with you that both approaches you suggested are better. Any is fine for me.

lukicdarkoo:
@stefaniapedrazzi  Let me know if it is better now :)

stefaniapedrazzi:
Thank you! The error messages are fine.

But it seems more clean to keep all the streaming parameter parsing in the `WbStreamingServer` class instead of the `WbGuiApplication` and then having to pass them again to `WbStreamingServer` and declaring a setting struct.

Given that the mode has to be extracted before instantiating the WbStreamingServer instance, it was better from my point of view to avoid a double check.

lukicdarkoo:
@stefaniapedrazzi I wasn't sure whether making a struct is a good idea, but I found it is used in a few places in Webots for the very similar problem. It is good because it is general, you can build the configuration on the way, not only from command line arguments.

----

There are two problems with your proposal:
1. Streaming arguments have to be parsed in `WbStreamingServer` and `WbGuiApplication` as `mode` has to be extracted before `WbStreamingServer` initialization (as you mentioned). Partially parsing (using `serverArgument.contains("mode=mjpeg")`) in `WbGuiApplication` misses the point.
2. Error messages and status code have to be passed back from `WbStreamingServer` to `WbGuiApplication`. We can simply print the errors in `WbStreamingServer`, but it looks hacky to me.

Here is my idea how to solve it if you agree:
1. Add a method to `WbGuiApplication` that converts the stream (and sub-parameters in general) parameters to a map. Then the map would be passed to the `WbStreamingServer` while taking `mode` to initialize correct `WbStreamingServer`.
2. Error messages from `WbStreamingServer` will be passed back to the `WbGuiApplication` by throwing an error in `WbStreamingServer`.

stefaniapedrazzi:
I agree that it is not the best to check the mode parameter in `WbGuiApplication` and the other parameters in `WbStreamingServer`.
On the other hand it is not so good to have to pollute the `WbGuiApplication` class with code specific to the `WbStreamingServer`, but it's ok and easy solution.
It is also bad to add a dependency from `WbStreamingServer` to `WbMultimediaStreamingServer` and `WbX3dStreamingServer` (even if just in the cpp), given that these classes already depend on `WbStreamingServer. So the instantiation of the `WbStreamingServer` hould remain in `WbGuiApplication` and should not be moved in `WbStreamingServer`.

So at the end: the current version is almost fine for me.
But it would be better if you could:
- move all the streaming server related code in a new function
- remove the struct and pass the parameters directly as arguments of the WbStreamingServer instance.
- try to improve the code so that
    ```
    cout << tr("Try 'webots --help' for more information.").toUtf8().constData() << endl;
    ```
    is executed and written in the new code only once?

src/webots/gui/WbGuiApplication.cpp
```diff
@@ -122,6 +122,75 @@ void WbGuiApplication::restart() {
...
+          cout << tr("webots: invalid option: '%1' in --stream").arg(serverArgument).toUtf8().constData() << endl;
+          cout << tr("webots: stream port has to be integer").toUtf8().constData() << endl;
+          mTask = FAILURE;
+        }
+      } else if (key == "mode") {
+        if (value != "x3d" && value != "mjpeg") {
+          cout << tr("webots: invalid option: '%1' in --stream").arg(serverArgument).toUtf8().constData() << endl;
+          cout << tr("webots: stream mode can only be x3d or mjpeg").toUtf8().constData() << endl;
+          mTask = FAILURE;
+        } else
```
stefaniapedrazzi, line 170: Shouldn't be?

```suggestion
        } else if (value == "mjpeg")
```

src/webots/gui/WbGuiApplication.cpp
```diff
@@ -122,6 +122,75 @@ void WbGuiApplication::restart() {
...
+      } else {
+        cout << tr("webots: unknown option: '%1' in --stream").arg(serverArgument).toUtf8().constData() << endl;
+        mTask = FAILURE;
+      }
+    } else {
+      cout << tr("webots: unknown option: '%1' in --stream").arg(serverArgument).toUtf8().constData() << endl;
+      mTask = FAILURE;
+    }
+  }
+  if (mTask != FAILURE) {
```
stefaniapedrazzi, line 181: Just very minor detail: this could be simplified by first checking if `mTask == FAILURE` and returning in the `if` block.

src/webots/gui/WbGuiApplication.hpp
```diff
@@ -59,6 +59,7 @@ class WbGuiApplication : public QApplication {
   WbStreamingServer *mStreamingServer;
 
   void parseArguments();
+  void parseStreamArguments(const QString &);
```
stefaniapedrazzi, line 62: Usually we always also add the argument name:
```suggestion
  void parseStreamArguments(const QString &serverArguments);
```

not sure why this is not the case on line 66.

src/webots/gui/WbGuiApplication.cpp
```diff
@@ -122,6 +122,75 @@ void WbGuiApplication::restart() {
 #endif
 }
 
+void WbGuiApplication::parseStreamArguments(const QString &serverArgument) {
```
stefaniapedrazzi, line 125: argument or arguments?
Should be consistent in the function name and argument name.

```suggestion
void WbGuiApplication::parseStreamArguments(const QString &serverArguments) {
```

src/webots/gui/WbGuiApplication.cpp
```diff
@@ -122,6 +122,75 @@ void WbGuiApplication::restart() {
...
+      }
+    } else {
+      cout << tr("webots: unknown option: '%1' in --stream").arg(serverArgument).toUtf8().constData() << endl;
+      mTask = FAILURE;
+    }
+  }
+  if (mTask != FAILURE) {
+    if (mode == "mjpeg") {
+      mStreamingServer = new WbMultimediaStreamingServer();
+      mStreamingServer->start(port, monitorActivity, disableTextStreams, ssl, controllerEdit);
```
stefaniapedrazzi, line 184: This function seems useless, the options could be set directly in the constructor, and then the already existing `start(port)` function could be used instead (making it public).

lukicdarkoo: In that case, if I am not mistaken, those parameters have to be passed through `WbX3dStreamingServer` and `WbMultimediaStreamingServer` which makes them aware of `monitorActivity`, `disableTextStreams`, `ssl` and `controllerEdit` (accessibility of these fields should be upgraded to protected).

stefaniapedrazzi: No, they can simply passed directly to the WbStreamingServer constructor, like this

```
WbMultimediaStreamingServer::WbMultimediaStreamingServer(bool monitorActivity, bool disableTextStreams, bool ssl, bool controllerEdit) :
  WbStreamingServer(monitorActivity, disableTextStreams, ssl, controllerEdit),
  mImageWidth(-1),
  mImageHeight(-1),
  ...
```

lukicdarkoo: Ok, I see, I wanted to avoid adding new arguments in `WbMultimediaStreamingServer` and `WbX3dStreamingServer` as well.

stefaniapedrazzi:
:eyes:

src/webots/gui/WbGuiApplication.cpp
```diff
@@ -122,6 +122,76 @@ void WbGuiApplication::restart() {
...
+void WbGuiApplication::parseStreamArguments(const QString &streamArguments) {
+  bool monitorActivity = false;
+  bool disableTextStreams = false;
+  bool ssl = false;
+  bool controllerEdit = false;
+  int port = 1234;
+  QString mode = "x3d";
+
+#ifdef __APPLE__
+  const QStringList &options = argument.split(';', QString::SkipEmptyParts);
```
stefaniapedrazzi, line 134:
```suggestion
  const QStringList &options = streamArguments.split(';', QString::SkipEmptyParts);
```

PR #2000, opened 2020-07-27
Check for ikpy version in inverse_kinematic controller

stefaniapedrazzi:
Print a warning if the ikpy available version is too old and not compatible with the provided controller.
Following  #1782.

### DIFF
--- a/projects/robots/abb/irb/controllers/inverse_kinematics/inverse_kinematics.py
+++ b/projects/robots/abb/irb/controllers/inverse_kinematics/inverse_kinematics.py
@@ -28,7 +28,7 @@ from controller import Supervisor
 
 if ikpy.__version__[0] < '3':
     sys.exit('The "ikpy" Python module version is too old. '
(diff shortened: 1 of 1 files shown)


projects/robots/abb/irb/controllers/inverse_kinematics/inverse_kinematics.py
```diff
@@ -14,17 +14,22 @@
...
-    import sys
     sys.exit('The "ikpy" Python module is not installed. '
              'To run this sample, please upgrade "pip" and install ikpy with this command: "pip install ikpy"')
 
 import math
 from controller import Supervisor
 
+if ikpy.__version__[0] < '3':
+    sys.exit('The "ikpy" Python module version is too old. '
+             'Please upgrade "ikpy" Python module to version "3.0" or newer with this command: "pip install ikpy"')
```
omichel, line 31: Shouldn't it be:
```suggestion
             'Please upgrade "ikpy" Python module to version "3.0" or newer with this command: "pip install --upgrade ikpy"')
```

PR #1947, opened 2020-07-20
Fixed accelerometer measurements

omichel:
Fixes #1693.
There was actually two problems:

1. If an `Accelerometer` has a parent node without a `Physics` node, it cannot determine its velocity from ODE needed to compute the acceleration and will return a wrong value. This is now fixed by displaying a warning in this case.

2. The computation of the velocity was not taking into account the complete offset of the accelerometer in case the parent node of the accelerometer was merged with an upper solid (solid merger). The computation of velocity is now based on the absolute point position instead of the wrong relative point position to work around this problem.

The code snipped provide in #1693 now returns correct values (after fixing the missing Physics node).

PR #2004, opened 2020-07-27
Fixed accelerometer warning and return value

omichel:
Follow-up of #1947.


stefaniapedrazzi:
:eyes:

stefaniapedrazzi:
Thank you!

I tested (also the `test_suite`) and it works fine now.
On simulation reset, the warning is not printed again but this seems ok.
I would just wait for the tests result before approving.

omichel:
I believe all the necessary tests passed (except for update_all_worlds.py which takes very very long).

PR #2014, opened 2020-07-27
Fixed Python 3.8 vehicle library which was broken on Windows

### LINKED ISSUE #2010: highway_overtake.wbt is failing on Windows with Python 3.8
It seems Python 3.8 cannot import some module (could be related with #1311).
```
INFO: sumo_supervisor: Starting controller: python.exe -u sumo_supervisor.py --no-gui --no-netconvert --max-vehicles=100 --port=8873 --seed=1 --step=200 --radius=250 --maximum-lateral-speed=2.5 --maximum-angular-speed=3 --lane-change-delay=4
INFO: radar_target_tracker: Starting controller: D:\msys64\home\Olivier\webots\projects\vehicles\controllers\radar_target_tracker\radar_target_tracker.exe
INFO: highway_overtake: Starting controller: python.exe -u highway_overtake.py
  File "highway_overtake.py", line 22, in <module>
    from vehicle import Driver
  File "D:\msys64\home\Olivier\webots\lib\controller\python38\vehicle.py", line 21, in <module>
    import _vehicle
ImportError: DLL load failed while importing _vehicle: The specified module could not be found.
WARNING: 'highway_overtake' controller exited with status: 1.
```

omichel:
Fixes #2010.

### DIFF
--- a/projects/default/libraries/vehicle/python/vehicle.i
+++ b/projects/default/libraries/vehicle/python/vehicle.i
@@ -1,5 +1,20 @@
 %module vehicle
 
+%pythonbegin %{
+import sys
+import os
+if os.name == 'nt' and sys.version_info >= (3, 8):  # we need to explicitly list the folders containing the DLLs
+    webots_home = os.environ['WEBOTS_HOME']
+    os.add_dll_directory(os.path.join(webots_home, 'lib', 'controller'))
+    # MSYS2_HOME should be set by Webots or ~/.bash_profile
+    # if not set, we are in the case of an extern controller and a regularly installed version of Webots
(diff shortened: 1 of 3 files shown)


omichel:
The Darwin-op managers python wrapper is also broken with Python 3.8 on Windows (i.e., humanoid_marathon is broken). However, the fix is not that straightforward...

DavidMansolino:
> The Darwin-op managers python wrapper is also broken with Python 3.8 on Windows (i.e., humanoid_marathon is broken). However, the fix is not that straightforward...

Note that the runtime file has to be updated: https://github.com/cyberbotics/webots/blob/develop/projects/samples/robotbenchmark/humanoid_marathon/controllers/marathon/runtime.ini#L4

omichel:
The trick for the Darwin-op libraries was to add the path to the `robotis-op2.dll` which is a dependency of `manager.dll`.

Issue #2007, opened 2020-07-27
Issues rotating object with the handles in the 3d view


stefaniapedrazzi:
Rotating an object seems to work only for specific viewpoint positions.
For example:
1. open the compass.wbt world
2. move the viewpoint to the blue box
3. change view to "Top view"
4. rotate the box with the handle -> everything works fine
5. slightly move the viewpoint -> it becomes almost impossible to rotate the box with the handles

Here are some Viewpoint values for which it doesn't work on my system:
```
Viewpoint {
  orientation 0.9998142420400471 -0.016477618076035542 -0.009998475705145874 5.1922241916445895
  position 0.4396693546155517 0.3189667472426704 0.25910779490914826
}
```

stefaniapedrazzi:
I can reproduce the same issue on R2020a-rev1, but not on R2019a rev1 (now I'm testing the other versions in between)

PR #693, opened 2019-07-03
Fix source tests on Windows

omichel:
The header test was broken (parsing files inside .git folder).
The cppcheck test was broken (on Windows, it relies on a newer version of cppcheck which detects more problems)

PR #2019, opened 2020-07-27
Fixed wrong conversion from WbVector3 to WbVector4

stefaniapedrazzi:
Fix issue #2007 introduced in R2019b revision 1 (#693): when converting from WbVector3 to WbVector4 the forth value was scaled too.


omichel:
Well spotted!

PR #2023, opened 2020-07-28
Add macOS to GitHub Action Worflow

DavidMansolino:
**Description**
Add macOS to the compilation and test workflow.


DavidMansolino:
Test suite is not yet functional because of #2110, but this workflow is already testing compilation and installation (in two different stages), which is already useful.

Note that I started by adding macOS in the same file as the linux tests, but it was a mess, so I finally preferred to have a separate file for macOS, this implies some small duplications, but this is way more cleaner and will be way more simple to maintain.

omichel (approved):
Thank you.
Maybe we can check if the `LIBGL_ALWAYS_SOFTWARE` environment variable is really needed on macOS.

DavidMansolino:
> Maybe we can check if the LIBGL_ALWAYS_SOFTWARE environment variable is really needed on macOS.

Good point, that looks indeed to be MESA specific: https://docs.mesa3d.org/envvars.html

omichel (approved):
Thank you.

PR #2029, opened 2020-07-29
Regenerate object-appearance doc + thumbnails

DavidMansolino:
**Description**
ditto


DavidMansolino:
The test are Failing because I the new blog post, the thumbnail where set manually without including the original files, I will fix this.

docs/guide/object-floors.md
```diff
@@ -157,6 +157,8 @@ UnevenTerrain {
   SFBool     flatCenter     FALSE
   SFBool     flatBounds     FALSE
   SFInt32    perlinNOctaves 3
+  deprecatedField MFString   texture        "textures/sand.jpg"
+  deprecatedField SFVec2f    textureScale   1.0 1.0
```
omichel, line 161: I am not sure we should document the deprecated fields...
```suggestion
```

DavidMansolino: Good point, I will improve the script to avoid this.

DavidMansolino: This is now done in cc4cfddbd34415909867a40d3222fae702fdb888.

PR #2050, opened 2020-07-31
Fix Windows Extern Controller Doc

### LINKED ISSUE #2047: Trouble with external controller and environmental variables in Windows with new Webots version R2020b
**Describe the Bug**
After installing the new version R2020b (from R2020a), I get the following error when running my external controller:

> Traceback (most recent call last):
>   File "robot_controller.py", line 3, in <module>
>     from controller_base import WebotSupervisor
>   File "D:\Robert\Documents\Currence\Webots\WorldGit\emulation2.0\controllers\robot_controller\controller_base.py", line 4, in <module>
>     from controller import Supervisor
>   File "C:\Program Files\Webots\lib\controller\python37\controller.py", line 31, in <module>
>     import _controller
> ImportError: DLL load failed: ImportError: DLL load failed: The specified module could not be found.

I have [set the environmental variables](https://cyberbotics.com/doc/guide/running-extern-robot-controllers?tab-os=windows&tab-language=python#environment-variables), tried to create a new pycharm project, as well as running my controller from the terminal (outside pycharm).

I got the same error when I started using Webots (R2020a) a few months ago, but setting the variables fixed it.

**System**

DavidMansolino:
Fix #2047

### DIFF
--- a/docs/guide/running-extern-robot-controllers.md
+++ b/docs/guide/running-extern-robot-controllers.md
@@ -28,10 +28,10 @@ Generic Webots environment variables needed for all the controller languages:
 %tab-component "os"
 
 %tab "Windows"
-| Environment Variable     | Typical Value                                                               |
-|--------------------------|-----------------------------------------------------------------------------|
-| WEBOTS\_HOME             | `C:\Program Files\Webots`                                                   |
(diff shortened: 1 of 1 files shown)


DavidMansolino:
> This is however not needed for C and MATLAB controllers.

You're completely right, but maybe it is better to keep it like this (just to avoid many cases that can confuse the user)?

Issue #1906, opened 2020-07-13
Fix "controllerArgs" tyep in test suite worlds 


stefaniapedrazzi:
The `Robot.controllerArgs` field has been changed from SFString to MFString.
All the occurrences have been updated but the test worlds.

In the test worlds, the "controllerArgs" is usually used to passed the controller name, but this is probably no longer necessary.
So we could take the opportunity to check if the "controllerArgs" are still needed and remove it.

omichel:
Note that putting the brackets around a MFString list of a single element is optional. Therefore all the test worlds are syntactically correct with respect to `Robot.controllerArgs`.

PR #2058, opened 2020-08-03
Update test_suite worlds

stefaniapedrazzi:
Complete  #1906.

These changes have been applied to the worlds:
* [x] fix light attenuation warning when the light conditions were not affecting the simulation
* [x] remove unneeded `controllerArgs`
* [x] fix solid name errors
* [x] convert most of the DifferentialWheels nodes to HingeJoint + Motor (just one remains in `physics/dynamic_light_sensor_rays.wbt`)
* [x] update a couple of worlds to use the ENU coordinate system

### DIFF
--- a/tests/api/controllers/camera_checker/camera_checker.c
+++ b/tests/api/controllers/camera_checker/camera_checker.c
@@ -76,21 +76,21 @@ int main(int argc, char **argv) {
     samples_positions[3][1] = height - height / 8;
 
     // brown
-    samples_expected_colors[0][0] = 123;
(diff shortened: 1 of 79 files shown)


stefaniapedrazzi:
I changed the `camera_color` and `camera_color_motion` so that they use the `PBRAppearance` node. Other tests are already using them (some of the pen and distance_server_infra-red tests).

When removing the DifferentialWheels node from the physics tests about dynamic sensor measurements I also updated them so that they now use the PBRAppearance nodes and the ENU coordinate system.

PR #2062, opened 2020-08-03
Improve Inverse Kinematic Sample

DavidMansolino:
**Description**
This PR improves the inverse kinematic sample by automatically parsing the robot kinematic using the URDF definition instead of hard-coding the kinematic.

This way it will be very simple to use it with other robots.

It also improves the URDF exporter to add joint limit: http://wiki.ros.org/urdf/XML/joint#Elements

### DIFF
--- a/projects/robots/abb/irb/controllers/inverse_kinematics/inverse_kinematics.py
+++ b/projects/robots/abb/irb/controllers/inverse_kinematics/inverse_kinematics.py
@@ -19,7 +19,6 @@ import tempfile
 try:
     import ikpy
     from ikpy.chain import Chain
-    from ikpy.link import OriginLink, URDFLink
 except ImportError:
     sys.exit('The "ikpy" Python module is not installed. '
              'To run this sample, please upgrade "pip" and install ikpy with this command: "pip install ikpy"')
--- a/projects/robots/abb/irb/controllers/inverse_kinematics/inverse_kinematics.py
+++ b/projects/robots/abb/irb/controllers/inverse_kinematics/inverse_kinematics.py
@@ -15,6 +15,7 @@
 """Demonstration of inverse kinematics using the "ikpy" Python module."""
 
 import sys
+import tempfile
 try:
     import ikpy
     from ikpy.chain import Chain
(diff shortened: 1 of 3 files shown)


projects/robots/abb/irb/controllers/inverse_kinematics/inverse_kinematics.py
```diff
@@ -30,58 +30,24 @@
...
-
 # Initialize the Webots Supervisor.
 supervisor = Supervisor()
 timeStep = int(4 * supervisor.getBasicTimeStep())
 
+# Create the arm chain from the URDF
+filename = None
+with tempfile.NamedTemporaryFile(suffix='.urdf', delete=False) as file:
+    filename = file.name
+    file.write(supervisor.getUrdf())
```
omichel, line 41: I am getting an error here with Python 3.8:
```
  File "inverse_kinematics.py", line 41, in <module>
    file.write(supervisor.getUrdf())
  File "C:\Program Files\Python38\lib\tempfile.py", line 474, in func_wrapper
    return func(*args, **kwargs)
TypeError: a bytes-like object is required, not 'str'
WARNING: 'inverse_kinematics' controller exited with status: 1.
```
I could fix it with:
```suggestion
    file.write(supervisor.getUrdf().encode('utf-8'))
```

nico1as:
Hello, Thanks for the IK sample for webots, it is very helpful and work very well as long as we don't change the original orientation of the robot. 
For example, if I rotate a bit the robot to the left, save the webot world and reload the simulation: The IK will always have an offset relatively to the target.
I suppose that we should adjust the first joint angle in regards to the webots robot rotation value, but don't figure out how doing that. 
Any idea?
Thanks very much

lukicdarkoo:
@nico1as This is a very simple example and it doesn't consider rotation. See here:
https://github.com/cyberbotics/webots/blob/77d97ed63b130ce4e4bff7c24986a1520f4bf22a/projects/robots/abb/irb/controllers/inverse_kinematics/inverse_kinematics.py#L88

We will be happy to review you contribution

nico1as:
I figure out how to fix the arm orientation issue when we change the robot rotation.

```
    from scipy.spatial.transform import Rotation as R

    [...]

    # we need to adjust orientation
    r = R.from_matrix(np.array(arm.getOrientation()).reshape(3, 3))
    # get angle from the z axe. We assume that robot initial z angle is 0.
    angle_adjust = r.as_euler('zyx', degrees=False)[0]

    # we change the x y z repair according to the robot rotation
    xp = x * math.cos(angle_adjust) + y * math.sin(angle_adjust)
    yp = -x * math.sin(angle_adjust) + y * math.cos(angle_adjust)

    # Call "ikpy" to compute the inverse kinematics of the arm.
    ikResults = armChain.inverse_kinematics([xp, yp, z])

    for i in range(len(motors)):
        motors[i].setPosition(ikResults[i + 1])
```

PR #2072, opened 2020-08-04
Fix SliderJoint URDF export.

DavidMansolino:
The slider-joint export to URDF was buggy because:
  - It was crashing when the joint has no motor.
  - It was not including the joint limit.


lukicdarkoo:
Sorry for making a commit directly here (0beeab49493dfaf1ef49) :/

In the example that I used for testing, the translations were wrong. There was a bug in translations definitely, but I am not sure whether it solves your problem as well. Let me know!

DavidMansolino:
> Sorry for making a commit directly here ([0beeab4](https://github.com/cyberbotics/webots/commit/0beeab49493dfaf1ef49795dcc4f2862edfec420)) :/
> 
> In the example that I used for testing, the translations were wrong. There was a bug in translations definitely, but I am not sure whether it solves your problem as well. Let me know!

Thank you for the fix, it seems indeed completely correct! But unfortunately it still doesn't fix the axis orientation of my use case.

Issue #2066, opened 2020-08-04
Use Python from Brew on macOS


DavidMansolino:
**Describe the Bug**
On MacOs it seems much more common to use the python version coming from Brew (e.g. GitHub action, travis, etc, all of them are using brew, and all the tutorials for installing python3 are mentioning brew).

We should probably use the brew version to compile the Webots Python 3.x API (possibly in addition to the standard Python3.x).

PR #2079, opened 2020-08-05
Add Python Brew API Compilation.

DavidMansolino:
**Description**
On MacOs it seems much more common to use the python version coming from Brew (e.g. GitHub action, travis, etc, all of them are using brew, and all the tutorials for installing python3 are mentioning brew).

We should probably use the brew version to compile the Webots Python 3.x API (possibly in addition to the standard Python3.x).

**Related Issues**
This pull-request fixes issue #2066

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -254,6 +254,8 @@ jobs:
         homebrew:
           packages:
             - python
+            - python@3.7
+            - python@3.8
             - cmake
             - swig
             - wget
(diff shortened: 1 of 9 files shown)


DavidMansolino:
To install python3.7 and 3.8 with brew:
```
brew install python@3.7 python@3.8
```

DavidMansolino:
I am currently trying ROS2 on Mac and it uses Python from homebrew too.

PR #2082, opened 2020-08-06
Improve update of USE nodes

### LINKED ISSUE #159: Asssertion Failure When Changing TextureTransform Scale
**Describe the Bug**
An assertion is raised when changing some `scale` fields of texture transform nodes.

**Steps to Reproduce**
Change the `scale` field of the `TextureTransform` node in the following world:
```
#VRML_SIM R2019a utf8
WorldInfo {
}
Viewpoint {
  orientation -0.3152880148513402 -0.8294815172282154 -0.461035660516493 1.9912523755950893
  position -2.4364618763913173 8.613048234119175 -0.3041357321642886
}
TexturedBackground {
}
TexturedBackgroundLight {
}
Solid {
  translation 1.6000000000000008 2.479999999999998 5.399053564414761e-08
  rotation -0.5773502747262953 -0.57735027749463 0.5773502553479516 2.094395119003205
  children [
    DEF SHAPE Shape {
      appearance ThreadMetalPlate {
        textureTransform TextureTransform {
        }
      }
      geometry Box {
        size 0.042 3.238 3.234
      }
    }
  ]
  boundingObject USE SHAPE
}
```

**System**
 - Operating System: Linux Ubuntu 16.04, debug mode
 - Graphics Card: NVIDIA GeForce GTX 980

**Additional context**
```
#0  0x00007ffff2859428 in __GI_raise (sig=sig@entry=6)
    at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007ffff285b02a in __GI_abort () at abort.c:89
#2  0x00007ffff2851bd7 in __assert_fail_base (fmt=<optimized out>, 
    assertion=assertion@entry=0xc0bd4b "0", 
    file=file@entry=0xc0bc9b "vrml/WbNode.cpp", line=line@entry=796, 
    function=function@entry=0xc0c140 <WbNode::findSubFieldIndex(WbField const*) const::__PRETTY_FUNCTION__> "int WbNode::findSubFieldIndex(const WbField*) const") at assert.c:92
#3  0x00007ffff2851c82 in __GI___assert_fail (assertion=0xc0bd4b "0", 
    file=0xc0bc9b "vrml/WbNode.cpp", line=796, 
    function=0xc0c140 <WbNode::findSubFieldIndex(WbField const*) const::__PRETTY_FUNCTION__> "int WbNode::findSubFieldIndex(const WbField*) const")
    at assert.c:101
#4  0x000000000090479b in WbNode::findSubFieldIndex (this=0x6ecad80, 
    searched=0x6edc9f0) at vrml/WbNode.cpp:796
#5  0x00000000009043b1 in WbNode::notifyFieldChanged (this=0x1661cc0)
    at vrml/WbNode.cpp:752
#6  0x000000000091047f in QtPrivate::FunctorCall<QtPrivate::IndexesList<>, QtPrivate::List<>, void, void (WbNode::*)()>::call(void (WbNode::*)(), WbNode*, void**) (
    f=(void (WbNode::*)(WbNode * const)) 0x9042e8 <WbNode::notifyFieldChanged()>, o=0x1661cc0, arg=0x7fffffffc9c0)
```

stefaniapedrazzi:
Fix #159:
the assert was wrong and in case of PROTO it could happen that the field was not found when searching in fields instead of parameters list.

I changed the field search so that now it searches in the parameters instead of fields for PROTO nodes given that PROTO parameter changes are propagated to the internal fields but not the other way.

### DIFF
--- a/src/webots/vrml/WbNode.cpp
+++ b/src/webots/vrml/WbNode.cpp
@@ -777,7 +777,10 @@ void WbNode::notifyFieldChanged() {
         }
 
         setGlobalParentNode(NULL);
-      }
+      } else
+        // index could be invalid for fields in sub-PROTO scope
+        // but in this case the USE node will be updated by the parameter change notification
+        break;
     }
 
     // climb up the family tree but stop if child insertion is not completed
@@ -798,7 +801,7 @@ void WbNode::notifyParameterChanged() {
 
 int WbNode::findSubFieldIndex(const WbField *const searched) const {
   int count = 0;
-  QList<WbNode *> list(subNodes(true));
+  QList<WbNode *> list(subNodes(true, false, false));
   list.prepend(const_cast<WbNode *>(this));
   foreach (WbNode *const node, list) {
     foreach (WbField *const field, node->mFields) {
@@ -807,13 +810,12 @@ int WbNode::findSubFieldIndex(const WbField *const searched) const {
       ++count;
     }
   }
-  assert(0);
   return -1;
 }
 
 WbField *WbNode::findSubField(int index, WbNode *&parent) const {
   int count = 0;
-  QList<WbNode *> list(subNodes(true));
+  QList<WbNode *> list(subNodes(true, false, false));
   list.prepend(const_cast<WbNode *>(this));
   foreach (WbNode *const node, list) {
     foreach (WbField *const field, node->mFields) {


stefaniapedrazzi:
Not sure it make sense to add this particular test because it was simply a wrong assertion raised in debug mode. But in release mode everything was working fine and the texture transform was correctly updated.

DavidMansolino:
Oh yes you're right, since it is only an assertion it is probably not so important to test.

PR #2107, opened 2020-08-10
Fix DEF not found on load for default PROTO parameter values

### LINKED ISSUE #1231: DEF not found if default argument of PROTO field.
**Describe the Bug**
If a PROTO has a DEF in a default field value:
```
PROTO MyProto [
  field SFNode geometry DEF BODY Cylinder { }
]
{
  Solid {
    boundingObject IS geometry
    physics Physics {
    }
  }
}
```

When using this PROTO in a world it is possible to USE this DEF, but when saving the world and reverting the DEF is not found:
```
#VRML_SIM R2020a utf8
WorldInfo {
}
Viewpoint {
  orientation -0.4431147445295423 -0.882635722087249 -0.15685504542742262 0.7999430253825172
  position -0.45003995793645357 0.4725437716052594 0.6616148217362576
}
TexturedBackground {
}
TexturedBackgroundLight {
}
Robot {
  children [
    MyProto {
    }
  ]
  boundingObject USE BODY
}
RectangleArena {
}
```

Screenshot from 2020-01-03 11-03-11

It is working fine if the DEF is not the default value, e.g. with this world it works fine:
```
#VRML_SIM R2020a utf8
WorldInfo {
}
Viewpoint {
  orientation -0.4431147445295423 -0.882635722087249 -0.15685504542742262 0.7999430253825172
  position -0.45003995793645357 0.4725437716052594 0.6616148217362576
}
TexturedBackground {
}
TexturedBackgroundLight {
}
Robot {
  children [
    MyProto {
      geometry DEF BODY2 Cylinder {
      }
    }
  ]
  boundingObject USE BODY2
}
RectangleArena {
}
```

**Expected behavior**

We should decide if we want to allow to use DEF in default field argument and fix this or deny the it's use.

stefaniapedrazzi:
Fix #1231: 

* [x] add DEF nodes from default PROTO parameter values to the WbNodeReader DEF map if not overridden
* [x] print a warning about wrong field order if any default PROTO parameter value contains a DEF node
* [x] add a test in the test suite
* [x] update ChangeLog

### DIFF
--- a/src/webots/vrml/WbNode.cpp
+++ b/src/webots/vrml/WbNode.cpp
@@ -1442,8 +1442,23 @@ WbNode *WbNode::createProtoInstance(WbProtoModel *proto, WbTokenizer *tokenizer,
 
   // 2. create the parameters list from the model (default values)
   QVector<WbField *> parameters;
-  foreach (WbFieldModel *model, protoFieldModels)
-    parameters.append(new WbField(model, NULL));
+  QVector<QMap<QString, WbNode *>> parametersDefMap;
+  bool hasDefaultDefNodes = false;
+  foreach (WbFieldModel *model, protoFieldModels) {
+    WbField *defaultParameter = new WbField(model, NULL);
+    parameters.append(defaultParameter);
+
+    parametersDefMap.append(QMap<QString, WbNode *>());
+    if (tokenizer && WbNodeReader::current()) {
+      // extract DEF nodes defined in default PROTO parameter
+      QList<WbNode *> defNodes = subNodes(defaultParameter, true, false, false);
+      foreach (WbNode *node, defNodes) {
+        if (!node->defName().isEmpty())
+          parametersDefMap.last().insert(node->defName(), node);
+      }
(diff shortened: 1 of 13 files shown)


src/webots/vrml/WbNode.cpp
```diff
@@ -1442,8 +1442,23 @@ WbNode *WbNode::createProtoInstance(WbProtoModel *proto, WbTokenizer *tokenizer,
 
   // 2. create the parameters list from the model (default values)
   QVector<WbField *> parameters;
-  foreach (WbFieldModel *model, protoFieldModels)
-    parameters.append(new WbField(model, NULL));
+  QVector<QMap<QString, WbNode *>> parametersDefMap;
+  bool hasDefaultDefNodes = false;
+  foreach (WbFieldModel *model, protoFieldModels) {
```
DavidMansolino, line 1447:
```suggestion
  foreach (const WbFieldModel *model, protoFieldModels) {
```

We saw in the past that Qt `foreach` performance is very bad with non-const items.

PR #2113, opened 2020-08-12
Feature Export URDF Geometry

DavidMansolino:
**Description**
While trying to debug inverse kinematic for a project I needed to visualize the robot in URDF to check the joint motion.

The exportation is still not perfect (which is kind of wanted) in the sense that only the bounding object is exported as visual and collision shapes, and only basic geometries are exported.

Webots:
wb
URDF:
urdf

With the ur5e (the missing part is because 'capsule' is not supported in URDF):
image


DavidMansolino:
> I was wondering whether it could be possible to add support for Capsule (which we use sometimes in bounding objects) and export them as one cylinder and two spheres in URDF (as I assume the capsule primitive is not natively supported in URDF).

This is now implemented:
image

DavidMansolino:
Cylinder orientation is now fixed :tada: 

image

DavidMansolino:
(sorry, this PR is not et ready)

DavidMansolino:
It seems there is a rounding issue somewhere as I am having such kind of rotation:
```
rpy="3.1415923000206876 0 0"
``

File: src/webots/maths/WbRotation.cpp [hunk 1]
```diff
@@ -45,27 +45,45 @@ void WbRotation::fromQuaternion(const WbQuaternion &q) {
...
+      mZ = M(0, 2) / (2 * mX);
+      mAngle = theta;
+    } else if (M(1, 1) > M(0, 0) && M(1, 1) > M(2, 2)) {
+      mY = sqrt(M(1, 1) - M(0, 0) - M(2, 2) + 0.5);
+      mX = M(0, 1) / (2 * mY);
+      mZ = M(1, 2) / (2 * mY);
+      mAngle = theta;
+    } else {
+      mZ = sqrt(M(2, 2) - M(0, 0) - M(1, 1) + 0.5);
+      mX = M(0, 1) / (2 * mZ);
```
DavidMansolino, line 82: If I am correct this should be instead:
```suggestion
      mX = M(0, 2) / (2 * mZ);
```

@lukicdarkoo can you confirm?

lukicdarkoo: Yes, well noticed! According to the reference document, it should be `M(0, 2)`

DavidMansolino: Thank you for the confirmation :-)

DavidMansolino:
I checked the new formula to convert from matrix to rotation, it looks basically the same but with more edge cases handled.

lukicdarkoo:
> I checked the new formula to convert from matrix to rotation, it looks basically the same but with more edge cases handled.

That makes sense. The cylinders in TIAGo++ seems to be falling into those edge cases. The previous implementation of `fromMatrix3()` handled only one or two edge cases (that's why I suspected it is the source of the error).

Issue #882, opened 2019-09-12
Removing Camera with External Window Open Crashes Webots


chilaire:
**Describe the Bug**
When a Robots has a camera, if you double-clic ont the camera displayer, it opens in another windows. Afterward, if you delete the camera node, webots.exe stops working.

**Steps to Reproduce**
1. start telemax.wbt for example
2. double-clic on one of the camera displayer to open it in a new windows
3. delete the camera or the robots 
4.  you got an error "Webots.exe .exe has stopped working"

(it occurs even if you previously closed the displayer windows, set the robot's controller to void, disabled the camera...)

**Expected behavior**
Deleting a camera even while it's working doesn't create any error or warning if you don't open the displayer in a new windows first.

**System**
 - Operating System: Windows 7 pro
 - Graphics Card: NVIDIA GeForce GTX 1060 6GB

PR #2119, opened 2020-08-12
Fix issues with extern rendering device windows

### LINKED ISSUE #302: Broken External Windows
**Describe the Bug**
The camera external windows are not working as expected if they are open the camera is enabled.
On master the window remains black when the camera is enabled, on develop Webots crashes when the camera is enabled.

**Steps to Reproduce**
1. Open the 'samples/devices/camera' simulation in pause mode.
2. Open the camera overlay in an external window.
3. On develop: enable the ambient occlusion on the camera with the `Camera.ambientOcclusionRadius` field.
4. Start the simulation

**Expected behavior**
The camera image should be visible even if the external window is open before the simulation starts.

**Screenshots**
master:
Capture du 2019-03-22 11-23-25

**System**
 - Operating System: Linux Ubuntu 16.04
 - Graphics Card: NVIDIA GeForce GTX 970

**Stack on develop**

```
#0  0x00007fffe6e69956 in ?? ()
   from /usr/lib/nvidia-384/libnvidia-glcore.so.384.130
#1  0x00007fffe6e7b91c in ?? ()
   from /usr/lib/nvidia-384/libnvidia-glcore.so.384.130
#2  0x00007fffe6a3a14e in ?? ()
   from /usr/lib/nvidia-384/libnvidia-glcore.so.384.130
#3  0x0000000000bb1e14 in wren::StaticMesh::render (this=0xfe43210, 
    drawingMode=4) at StaticMesh.cpp:1565
#4  0x0000000000ba356a in wren::Renderable::setupAndRender (this=0xfe6a220, 
    program=0xfe42df0) at Renderable.cpp:217
#5  0x0000000000ba2996 in wren::Renderable::render (this=0xfe6a220, 
    program=0x0) at Renderable.cpp:80
#6  0x0000000000bc22e1 in wren::Scene::renderToViewport (this=0x60fe4f0, 
    culling=true) at Scene.cpp:367
#7  0x0000000000bc19f9 in wren::Scene::renderToViewports (this=0x60fe4f0, 
    viewports=std::vector of length 1, capacity 1 = {...}, culling=true)
    at Scene.cpp:266
#8  0x0000000000bc15e4 in wren::Scene::render (this=0x60fe4f0, culling=true)
    at Scene.cpp:223
#9  0x0000000000bc467a in wr_scene_render (scene=0x60fe4f0, material_name=0x0, 
    culling=true) at Scene.cpp:862
#10 0x0000000000a7a222 in WbWrenWindow::renderNow (this=0x5c426d0, 
    culling=true) at gui/WbWrenWindow.cpp:229
```

### LINKED ISSUE #623: Wrong rendering after opening external overlay window
After loading many worlds, at some time the rendering becomes much darker than what it should be.
For example starting Webots and running the Guided Tour, at some point I got this appearance for the `receiver_noise.wbt` simulation:

receiver_noise

Then, I quitted and restarted Webots and the rendering was correct:

receiver_noise_1

I could reproduce this behaviour a couple of times (at some point the also the hokuyo.wbt Display overlays were not shown any longer on the scene).

Probably my NVIDIA drivers are not up-to-date, but it would be strange if this is the cause of the issue because it works at the beginning.

System: Windows 10
Graphics card: NVIDIA GeForce GT 640M with 2GB VRAM
OpenGL version: 3.3.0 NVIDIA 418.81
Ambient Occlusion: Medium
Texture Quality: High

stefaniapedrazzi:
Tasks:
  - [x] fix #623 and #882:
    when deleting the rendering device window the OpenGL context was not the correct one
  - [x] fix  #302:
    if the rendering device window is open before the device is setup, the initialization of the textures, VAO and VBO is skipped and the image is not displayed
  - [x] fix the code so that it really uses VAO (currently just declared but not used)
  - [x] update change log
  - [x]  update test_suite mirror test and add regression tests for #882 and #302


omichel:
:eyes:

src/webots/gui/WbRenderingDeviceWindow.cpp
```diff
@@ -135,10 +136,22 @@ WbRenderingDeviceWindow::WbRenderingDeviceWindow(WbRenderingDevice *device) :
...
+
+  const bool success = mContext->makeCurrent(this);
+  assert(success);
+  if (!success)
+    return;
+
   QOpenGLFunctions_3_3_Core *f = mContext->versionFunctions<QOpenGLFunctions_3_3_Core>();
-  f->glDeleteBuffers(2, (GLuint *)&mVboId);
   f->glDeleteVertexArrays(1, &mVaoId);
-  delete mContext;
```
omichel, line 141: I believe there is a memory leak somewhere because the `mContext` object allocated line 279 seems to be never deleted.

src/webots/gui/WbRenderingDeviceWindow.cpp
```diff
@@ -135,10 +136,22 @@ WbRenderingDeviceWindow::WbRenderingDeviceWindow(WbRenderingDevice *device) :
...
+  f->glDeleteBuffers(2, (GLuint *)&mVboId);
+  mInitialized = false;
+  mContext->doneCurrent();
```
omichel, line 154: It seems safe to add it here:
```suggestion
  mContext->doneCurrent();
  delete mContext;
```

src/webots/gui/WbRenderingDeviceWindow.cpp
stefaniapedrazzi: No, there is no memory leak.
`mContext` is a `QObject` created with `this` as parent and Qt will automatically delete it when the parent is deleted.
So the Qt best practice (also applied in the rest of the Webots code) is to not delete it manually.

The same approach is also used in the OpenGL Window example where there is the new instruction but the delete is implicit in the destructor:
https://doc.qt.io/qt-5/qtgui-openglwindow-example.html

omichel: Yes, you are right.

src/webots/gui/WbRenderingDeviceWindow.cpp
```diff
@@ -135,10 +136,22 @@ WbRenderingDeviceWindow::WbRenderingDeviceWindow(WbRenderingDevice *device) :
...
+  assert(success);
+  if (!success)
+    return;
+
   QOpenGLFunctions_3_3_Core *f = mContext->versionFunctions<QOpenGLFunctions_3_3_Core>();
-  f->glDeleteBuffers(2, (GLuint *)&mVboId);
   f->glDeleteVertexArrays(1, &mVaoId);
-  delete mContext;
+  f->glDeleteBuffers(2, (GLuint *)&mVboId);
+  mInitialized = false;
```
DavidMansolino, line 153: Is this really useful?

stefaniapedrazzi: No, given that it is in the destructor it can be removed.

DavidMansolino (approved):
I tested on Linux and it is working perfectly, thank you!

stefaniapedrazzi:
Thank you for all the tests

PR #2122, opened 2020-08-13
Add minimum python version for test_worlds_warnings.py

stefaniapedrazzi:
The current version of `test_worlds_warnings.py` only works with python >= 3.7.
There is no need to make it compatible with previous versions, but it is better to have a nice message explaining why it doesn't work.

### DIFF
--- a/tests/manual_tests/test_worlds_warnings.py
+++ b/tests/manual_tests/test_worlds_warnings.py
@@ -23,9 +23,7 @@ from threading import Timer
 from subprocess import Popen, PIPE
 
 if sys.version_info[0] != 3 or sys.version_info[1] < 7:
-    print("This script requires Python version 3.7")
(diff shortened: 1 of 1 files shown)


stefaniapedrazzi:
Yes, I just fixed the pep8 warnings.

tests/manual_tests/test_worlds_warnings.py
```diff
@@ -22,6 +22,10 @@
 from threading import Timer
 from subprocess import Popen, PIPE
 
+if sys.version_info[0] != 3 or sys.version_info[1] < 7:
+    print("This script requires Python version 3.7")
+    sys.exit(1)
+
```
DavidMansolino, line 28:
```suggestion
    sys.exit('This script requires Python version 3.7')
```

Issue #2121, opened 2020-08-13
Sensor callbacks in languages/ros/webots_ros/complete_test are not working


stefaniapedrazzi:
It seems that the sensors callbacks in the ROS `complete_test` are never called.
I checked the camera recognition one in particular and even if some objects are  detected (after modifying the world) callback messages are not printed in the logs and no topic subscribers are detected.

PR #2125, opened 2020-08-14
Fix ROS complete_test

DavidMansolino:
This pull-request fixes issue #2121

### DIFF
--- a/projects/languages/ros/webots_ros/src/complete_test.cpp
+++ b/projects/languages/ros/webots_ros/src/complete_test.cpp
@@ -133,17 +133,16 @@ using namespace std;
 
 static int model_count;
 static vector<string> model_list;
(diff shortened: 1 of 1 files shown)


DavidMansolino:

projects/languages/ros/webots_ros/src/complete_test.cpp
```diff
@@ -143,13 +143,16 @@ double GyroValues[3] = {0, 0, 0};
 double inertialUnitValues[4] = {0, 0, 0, 0};
 double touchSensorValues[3] = {0, 0, 0};
 
+bool callbackCalled = false;
```
omichel, line 146: We should probably declare this variable as `static` and all the other ones above as well.

```suggestion
static bool callbackCalled = false;
```

PR #2130, opened 2020-08-17
Improve conversion.

DavidMansolino:
**Description**
This PR improves the convert to support NUE to ENU conversion.

### DIFF
--- a/scripts/converter/convert_to_enu.py
+++ b/scripts/converter/convert_to_enu.py
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Convert world file from R2020b from the NUE to the ENU coordinate system."""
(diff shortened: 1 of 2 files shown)


DavidMansolino:
Not yet perfect, but already allowed me to convert all the world of https://github.com/cyberbotics/webots_ros2/pull/123

scripts/converter/convert_to_enu.py
```diff
@@ -14,20 +14,20 @@
...
 import sys
 
 from transforms3d import quaternions
 
 from webots_parser import WebotsParser
-from converted_protos import converted_protos
 
 
-def rotation(value, r):
+def rotation(value, r, invert=False):
```
omichel, line 27: I would remove this `invert` argument, as it is never used with the default value.

PR #2133, opened 2020-08-19
Add comments to setup_viewer.js

stefaniapedrazzi:
Improve documentation of `setup_viewer.js` and explicitly unset the timeout so that the simulation is not automatically stopped when changing the broadcast value to false.


omichel (approved):
Thank you.

PR #2137, opened 2020-08-20
meter per radian => radian per secod

alxy:
This was probably a typo


DavidMansolino (approved):
Thank you for this correction.

PR #2138, opened 2020-08-20
Fix camera image update time check for asynchronous controller

### LINKED ISSUE #1926: Rendering Devices not working in Asynchronous Mode
**Describe the Bug**
There are some issues with rendering devices in asynchronous mode.

**Steps to Reproduce**
1. Open the camera.wbt world.
2. Set the 'synchronization' field of the robot to 'FALSE'.
3. Run the simulation in fast mode.
4. The controller displays some `Warning: wb_camera_get_image: image could not be retrieved.` warnings in the console and then crashes.

stefaniapedrazzi:
**Description**
Fix #1926: fix image update time in case of asynchronous controller.


src/Controller/api/robot_private.h
```diff
@@ -43,6 +43,7 @@ int robot_get_number_of_devices();
 WbDeviceTag robot_get_device_tag(WbDevice *);
 WbDevice *robot_get_robot_device();
 double wb_robot_get_time();
+bool wb_robot_get_synchronization();
```
omichel, line 46: I believe we should not add public API functions into private include files like this `robot_private.h` file, as it is redundant with the public include files. Instead, we should `#include <webots/robot.h>` from the C file. The same probably applies to `wb_robot_get_time()` and maybe other functions.

projects/samples/devices/controllers/camera/camera.c
```diff
@@ -101,8 +101,10 @@ int main() {
      * The robot turns and analyse the camera image in order
      * to find a new blob
      */
-    else {  // pause_counter == 0
-
+    else if (!image) {  // image may be NULL is Robot.synchronization is FALSE
```
DavidMansolino, line 104:
```suggestion
    else if (!image) {  // image may be NULL if Robot.synchronization is FALSE
```

DavidMansolino:
:eyes:

DavidMansolino (changes requested):
Just one minor typo.

DavidMansolino (approved):
Thank you both!

PR #2142, opened 2020-08-25
Fix unsupported PbrAppearance in IndexedLineSet and PointSet

### LINKED ISSUE #1794: Broken IndexedLineSet Appearance
**Describe the Bug**
The appearance of the indexedLineSet node is not working, it is always black:
image
Here is how it should looks like (image of the doc):
geometric_primitives

omichel:
Fixes #1794.


DavidMansolino:
:eyes:

DavidMansolino (changes requested):
When I add a PbrAppearance I don't have any warning, it appears only once I save and revert the simulation.

The `point_set` test may be updated (it uses an appearance and we recommend to not define appearances for PointSet nodes).

Ideally, we should improve the insertion rules, it makes no sense to allow the user to insert a PbrAppearance in the `appearance` field if a PointSet or IndexedLineSet is already present in the `geometry` field.

omichel:
`Appearance` actually works pretty well for PointSet. I will update the documentation accordingly and leave the test unchanged.

omichel:
I believe I fixed all the problem reported by @DavidMansolino.

docs/reference/pointset.md
```diff
@@ -10,7 +10,9 @@ PointSet {
 ### Description
 
 The [PointSet](#pointset) node represents a set of 3D points specified in the `coord` field.
-[PointSet](#pointset) nodes are not lit, are not texture-mapped and they do not cast or receive shadows.
+[PointSet](#pointset) nodes don't support [PBRAppearance](pbrappearance.md).
+Instead, you should an [Appearance](appearance.md) node in the [Shape](shape.md) containing the [PointSet](#pointset) geometry or leave the `appearance` field to NULL.
```
DavidMansolino, line 14:
```suggestion
Instead, you should use an [Appearance](appearance.md) node in the [Shape](shape.md) containing the [PointSet](#pointset) geometry or leave the `appearance` field to NULL.
```

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -304,9 +306,19 @@ namespace {
...
         return true;
-      else {
+      else if (nodeName == "PBRAppearance") {
+        const WbShape *const shape = dynamic_cast<const WbShape *const>(node);
+        if (!shape)
+          return true;
+        const WbGeometry *const geometry = shape->geometry();
+        if (!geometry)
+          return true;
+        if (dynamic_cast<const WbIndexedLineSet *const>(geometry) || dynamic_cast<const WbPointSet *const>(geometry))
```
DavidMansolino, line 318:
```suggestion
        if (geometry->nodeType() == WB_NODE_INDEXED_FACE_SET || geometry->nodeType() == WB_NODE_POINT_SET)
```

src/webots/nodes/WbIndexedLineSet.cpp
```diff
@@ -110,6 +110,13 @@ bool WbIndexedLineSet::sanitizeFields() {
     return false;
   }
 
+  const WbShape *const shape = dynamic_cast<WbShape *>(parentNode());
+  if (shape && shape->pbrAppearance()) {
+    parsingWarn(tr(
+      "The 'IndexedLineSet' node doesn't support 'PBRAppearance' in its 'appearance' field, please use 'Appearance' instead."));
+    return false;
+  }
```
DavidMansolino, line 118: This test is now pointless right?

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -304,9 +306,19 @@ namespace {
...
+          return false;
```
DavidMansolino, line 319:
```suggestion
          errorMessage = QObject::tr("The '%1' node doesn't support 'PBRAppearance' in its 'appearance' field, please use 'Appearance' instead.").arg(geometry->nodeModelName());
          return false;
```

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -482,6 +494,11 @@ namespace {
         }
 
       } else if (fieldName == "geometry") {
+        if (nodeName == "IndexedLineSet" || nodeName == "PointSet") {
+          const WbShape *const shape = dynamic_cast<const WbShape *const>(node);
+          if (shape && shape->pbrAppearance())
+            return false;
```
DavidMansolino, line 500:
```suggestion
          if (shape && shape->pbrAppearance()) {
            errorMessage = QObject::tr("Can't insert '%1' node in 'geometry' field of 'Shape' node if the 'appearance' field contains a 'PBRAppearance' node, please use 'Appearance' instead.").arg(nodeName);
            return false;
          }
```

DavidMansolino (changes requested):

src/webots/nodes/WbIndexedLineSet.cpp
omichel: No, it is still useful when loading a world file with a PBRAppearance for an IndexedLineSet.

DavidMansolino: In that case this warning is not raised, because the PBRAppearance node is never instantiated because it does not respect the insertion rules:
```
WARNING: DEF INDEXED_LINE_SET Transform > Shape: Skipped unexpected PBRAppearance node in 'appearance' field of Shape node.
``

**Comment by omichel** [comment]
OK, I addressed all the issues you reported.

**Comment by DavidMansolino** [comment]
:eyes:

File: src/webots/nodes/utils/WbNodeUtilities.cpp [hunk 6]
```diff
@@ -304,9 +306,24 @@ namespace {
...
+        const WbShape *const shape = dynamic_cast<const WbShape *const>(node);
+        if (!shape)
+          return true;
+        const WbGeometry *const geometry = shape->geometry();
+        if (!geometry)
+          return true;
+        if (geometry->nodeType() == WB_NODE_INDEXED_LINE_SET || geometry->nodeType() == WB_NODE_POINT_SET) {
+          errorMessage =
+            QObject::tr(
+              "The '%1' node doesn't support 'PBRAppearance' in its 'appearance' field, please use 'Appearance' instead.")
```
DavidMansolino, line 321:
```suggestion
              "The '%1' node doesn't support 'PBRAppearance' in the 'appearance' field of its parent node, please use 'Appearance' instead.")
```

DavidMansolino (changes requested):

DavidMansolino (approved):
Thank you

PR #2144, opened 2020-08-25
Added missing wbproj files

omichel:
These files are generated every time the test suite is run, adding garbage in the `git status`.


tests/rendering/worlds/.normals.wbproj
```diff
@@ -0,0 +1,9 @@
+Webots Project File version R2020b
```
DavidMansolino, line 1: Maybe a gitignore in the '/rendering/worlds/' directory like for the other category (https://github.com/cyberbotics/webots/blob/590a26b182f46189d3ccc96e89fecdb872c43060/tests/protos/worlds/.gitignore) is cleaner?

tests/api/worlds/.robot_window_html.wbproj
```diff
@@ -0,0 +1,10 @@
+Webots Project File version R2020b
```
DavidMansolino, line 1: Do we know why this particular one is un-ignored (https://github.com/cyberbotics/webots/blob/master/tests/api/worlds/.gitignore#L2) ?

omichel: No idea. I believe we should remove it from the .gitignore.

tests/api/worlds/.robot_window_html.wbproj
```diff
@@ -0,0 +1,10 @@
...
+perspectives: 000000ff00000000fd000000030000000000000320000003d0fc0100000002fb0000001a0044006f00630075006d0065006e0074006100740069006f006e0000000000ffffffff0000005400fffffffb0000001e00480074006d006c0052006f0062006f007400570069006e0064006f00770100000000000003200000005400ffffff0000000100000124000002f2fc0200000001fb0000001400540065007800740045006400690074006f00720100000015000002f20000003e00ffffff00000003000006da000000d8fc0100000001fb0000001a0043006f006e0073006f006c00650041006c006c0041006c006c0100000326000006da0000005400ffffff000005b0000002f200000001000000020000000100000008fc00000000
```
DavidMansolino, line 2: This is now ignored by the gitignore, so it should be removed from the repository.

tests/rendering/worlds/.normals.wbproj
DavidMansolino, line 1: Same for these files.

DavidMansolino (changes requested):

tests/api/worlds/.robot_window_html.wbproj
omichel: No, it's finally useful to have the perspective for the robot window, so I re-added the exception to the `.gitignore` and we should keep it here.

DavidMansolino (approved):
Thank you

PR #2145, opened 2020-08-25
Use motor name as urdf-joint name

Simon-Steinmann:
Converting a urdf to .proto, we use the jointname as the motorname. Converting .proto to .urdf should do the same in reverse, not take the sensor name.


lukicdarkoo:
This will break `JointStatePublisher`:
https://github.com/cyberbotics/webots_ros2/blob/master/webots_ros2_core/webots_ros2_core/joint_state_publisher.py

`JointStatePublisher` has to be improved to retrive corresponding `PositionSensor` (with `wb_motor_get_position_sensor()`).

Simon-Steinmann:
I created a PR for ROS2 adressing this issue https://github.com/cyberbotics/webots_ros2/pull/131

src/webots/nodes/WbJoint.cpp
```diff
@@ -296,8 +296,8 @@ void WbJoint::updateJointAxisRepresentation() {
 }
 
 const QString WbJoint::urdfName() const {
-  if (positionSensor() && positionSensor()->findSFString("name"))
-    return getUrdfPrefix() + positionSensor()->findSFString("name")->value();
+  if (motor() && motor()->findSFString("name"))
+    return getUrdfPrefix() + motor()->motor("name")->value();
```
DavidMansolino, line 300:
```suggestion
  if (motor() && motor()->findSFString("name"))
    return getUrdfPrefix() + motor()->motor("name")->value();
  else if (positionSensor() && positionSensor()->findSFString("name"))
    return getUrdfPrefix() + positionSensor()->findSFString("name")->value();
```

Just to handle the case of a non-actuated joint.

DavidMansolino:
This file probably needs to be updated: https://github.com/cyberbotics/webots/blob/master/tests/api/controllers/robot_urdf/reference_robot.urdf

DavidMansolino (approved):
Looks all good now.

omichel (approved):
Thank you.

PR #2146, opened 2020-08-26
Cleanup ODE MT.

DavidMansolino:
**Description**
Most of the debugging code of the multi-threading part of ODE was not maintained at all (was probably only development debug prints) and was not compiling anymore.


omichel (approved):
Thank you. Good cleanup.

PR #2148, opened 2020-08-26
Fix URDF test on Windows

omichel:
Due to the fact that mathematical functions like sin() and cos() do not produce the exact same result when compiled on Linux or Windows, it turns out the output of the URDF produced by Webots is not exactly the same, because it is exported with the maximum floating point precision (double).

This was causing the robot_urdf test to fail.

This PR fix this problem by exporting the URDF using FLOAT_MAX precision instead of DOUBLE_MAX.


tests/api/controllers/robot_urdf/robot_urdf.c
```diff
@@ -10,20 +10,22 @@
...
+  int count = 0;
   do {
+    count++;
     line_status_1 = fgets(line_1, MAX_LINE_LENGTH, file_1);
     line_status_2 = fgets(line_2, MAX_LINE_LENGTH, file_2);
-    if (strcmp(line_1, line_2))
-      return false;
+    if (strcmp(line_1, line_2)) {
+      return count;
+    }
```
DavidMansolino, line 25:
```suggestion
    if (strcmp(line_1, line_2))
      return count;
```

DavidMansolino (approved):
Thank you.

PR #2150, opened 2020-08-26
OSM_Importer  upgrade lib urllib for python3

HefnySco:
I updated 
`import urllib`

to 

`import urllib.parse
import urllib.request`

to work on python3


DavidMansolino (commented):
This looks good, thank you!

However, since this breaks compatibility with python2.7, the python3 shebang should be added on the first line of the files:
```
#!/usr/bin/env python3
```

And it should be mentioned in the changelog: https://github.com/cyberbotics/webots/blob/master/docs/reference/changelog-r2020.md

DavidMansolino (approved):
Thank you, that looks all good now.

PR #2162, opened 2020-08-31
Hide Indicator Motor in Vehicle Robot-Window

DavidMansolino:
**Description**
This motor is an internal motor and should not be displayed in the robot-window.


omichel (approved):
Thank you.

PR #2163, opened 2020-08-31
Fix Empty Forest

DavidMansolino:
**Description**
If the forest does not contain any tree (e.g. the forest file is not found), the PROTO was raising two warnings, the first one about forest without trees (which is fine) and a second one about an IndexedFaceSet without coordinates.


omichel (approved):
Thank you.

PR #2166, opened 2020-08-31
Fix ros controller Makefile

stefaniapedrazzi:
Fix python command in `ros` controller Makefile in case `PYTHON_COMMAND` was set.


DavidMansolino (approved):
Thank you

PR #2172, opened 2020-08-31
Enforce check of ros controller in package

DavidMansolino:
**Description**
Currently the ros controller is missing from the snap package because ROS is not installed in the snap environment, this PR does not fix the issue, but makes sure that the ros controller is available in all the Linux packages.


omichel (approved):
That seems good.

PR #2174, opened 2020-08-31
Fix Ctrl-click in documentation pages

### LINKED ISSUE #2160: Documentation link handling doesn't respect "ctrl-click" to open in new tab
Most user guide and reference pages. For example: https://www.cyberbotics.com/doc/reference/inertialunit

Clicking on links to other pages, for example using the left nav or on an inline link for another node, while holding <kbd>Ctrl</kbd> does not open in a new tab. Normally, the browser would handle this, but the docs pages override the click handler. The manual handling of the link then doesn't check the original event to see if it should be opening in a new tab. Relevant event handler:

https://github.com/cyberbotics/webots/blob/0ed61bd0390dc98de84f64b3a1692069893a61cc/docs/js/viewer.js#L264

**Describe how you would fix it**

I don't know the original motivation for the JS link handling, so this might not solve the original problem, but I assume the event's default action would work fine:

```js
if (event.ctrlKey) {
    return;
}
```

omichel:
Fixes #2160.
Test it here: https://www.cyberbotics.com/doc/reference/inertialunit?version=fix-ctrl-click-in-doc

### DIFF
--- a/docs/js/viewer.js
+++ b/docs/js/viewer.js
@@ -261,6 +261,8 @@ function addDynamicLoadEvent(el) {
     return;
   el.addEventListener('click',
     function(event) {
+      if (event.ctrlKey)
+        return;
       aClick(event.target);
       event.preventDefault();
     },


DavidMansolino (approved):
Thank you.

PR #2176, opened 2020-08-31
Fix importer_demo.sh on Linux

### LINKED ISSUE #2173: Python command in OSM importer demo
By default 'python' command doesn't seem to be defined on Ubuntu 20.04.
Alias doesn't work on bash scripts.
It would be good to use the PYTHON_COMMAND variable in the `importer_demo.sh` script so that the python command can be easily set by the user without modifying the file.

For example similarly to 
https://github.com/cyberbotics/webots/blob/b7ebc72b20f68693fd1339ef30bbf5817874fe60/projects/default/controllers/ros/Makefile#L38-L41

stefaniapedrazzi:
Fix #2173: the command `python` is not available by default on Ubuntu 20.04 (and Ubuntu 18.04).
Given that the script is not compatible with python 2.7, the used command can be changed to `python3`.

### DIFF
--- a/projects/vehicles/importer_demo.sh
+++ b/projects/vehicles/importer_demo.sh
@@ -17,7 +17,7 @@ usage() {
   echo
   echo "Assumptions:"
   echo "- The OSM file is located in the resources directory. e.g. ./resources/osm_files/$(ARG1).osm"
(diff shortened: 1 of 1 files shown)


DavidMansolino (approved):
Thank you.

PR #2185, opened 2020-09-01
Fix UnevenTerrain Size

DavidMansolino:
**Description**
Fix the generated size of the UnevenTerrain PROTO.


DavidMansolino:
textured_shapes world with the fix:
image

omichel (approved):

PR #2191, opened 2020-09-01
Avoid duplicating ros files

### LINKED ISSUE #2189: Avoid duplicating ros files
**Describe the Bug**
Currently, most of the ros files are duplicated:
  - https://github.com/cyberbotics/webots/tree/master/projects/languages/ros/webots_ros
  - https://github.com/cyberbotics/webots_ros

As those files don't change so often, it would be nice to use cyberbotics/webots_ros as a submodule instead.

DavidMansolino:
**Description**
Fix  #2189


omichel (approved):
Very nice cleanup!

docs/guide/tutorial-8-using-ros.md
```diff
@@ -119,13 +119,3 @@ You can find in the [Reference Manual](../reference/nodes-and-api-functions.md)
 If you are running the latest version of Webots, the easiest way to setup the `webots_ros` package is to install the `ros-$ROS_DISTRO-webots-ros` package directly from the package manager as described above.
 But if you are running an older Webots version, some functionality might not be fully supported by the latest `ros-$ROS_DISTRO-webots-ros` package.
 In this case, it is possible to install an older `webots_ros` package version from sources following the instructions on the [ROS wiki page](http://wiki.ros.org/webots_ros#From_Sources) and selecting the desired branch in the [`webots_ros` GitHub repository](https://github.com/cyberbotics/webots_ros).
-
-Alternatively, the `webots_ros` package can also be setup taking the resources directly from the Webots installation directory:
```
stefaniapedrazzi, line 123: These instructions are very useful and were added just few weeks ago because they were missing. 
If they are no longer up to date, they should be updated not removed.

stefaniapedrazzi: Note that the problem was version mismatch between the Webots_ros released packages and the Webots repo status.

DavidMansolino: The problem is that these files are not released anymore, instead the user should get them directly from the webots_ros repo (previous paragraph).
But if you want that we can better 'match' Webots version to 'webots_ros' version, we can create a release for each tag (https://github.com/cyberbotics/webots_ros/releases) and describe in the release to which Webots version it corresponds. Would that fix the problem?

omichel: A simple 1:1 match rule would be nice indeed, for example R2020b-rev2 could match version 2020.1.2 and R2021a could match version 2021.0.0, etc.

stefaniapedrazzi: Only partially.
It won't fix the problem of master and develop Webots branches nor synchronized with the Webots_ros package until they are released. 

Here it would just be fine to describe how to get the exact Webots._roa package matching the the current Webots sources.
But then maybe there is a bigger synchronization issue between the two repositories.

DavidMansolino: > A simple 1:1 match rule would be nice indeed, for example R2020b-rev2 could match version 2020.1.2 and R2021a could match version 2021.0.0, etc.

I was more thinking about adding the version in the description of the github release, having such high numbers in the ros release is not standard at all for ros packages (actually no single package are using 2 digit number for the major version), I fear that using these numbers would make the package appears as some outsider not well aligned with the ros packaging spirit.

> Only partially.
It won't fix the problem of master and develop Webots branches nor synchronized with the Webots_ros package until they are released.

My proposal is for users only, for developers we should use the version pointed by the submodule in our current branch (but I agree we can document this somewhere, but not sure the docs is the best location as it is only for developers).

omichel: > I was more thinking about adding the version in the description of the github release, having such high numbers in the ros release is not standard at all for ros packages (actually no single package are using 2 digit number for the major version), I fear that using these numbers would make the package appears as some outsider not well aligned with the ros packaging spirit.

I see. Then, something less exotic, but still matching the major/minor/revision could work, like 2.1.2 &rarr; R2020b-rev2 and 3.0.0 &rarr; R2021a.

DavidMansolino: Yes, this sounds good.

PR #2192, opened 2020-09-01
Cleanup simplify splash screen code

omichel:
Small cleanup (to be merged after the release).

### DIFF
--- a/src/webots/gui/WbGuiApplication.cpp
+++ b/src/webots/gui/WbGuiApplication.cpp
@@ -387,11 +387,6 @@ bool WbGuiApplication::setup() {
     const QDir screenshotLocation = QDir("images:splash_images/", "*.jpg");
     const QString webotsLogoName("webots.png");
     mSplash = new WbSplashScreen(screenshotLocation.entryList(), webotsLogoName);
-    mSplash->setWindowFlags(Qt::SplashScreen);
-#ifdef _WIN32
-    mSplash->setWindowModality(Qt::ApplicationModal);
-#endif
-
     if (WbPreferences::instance()->value("MainWindow/maximized", false).toBool()) {
       // we need to center the splash screen on the same window as the mainWindow,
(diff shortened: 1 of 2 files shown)


stefaniapedrazzi (approved):
It works fine.

Just a detail that I noticed when testing, the Webots window is shown quite immediately above the splash screen on Linux and thus the splash screen is not very visible.

omichel:
> Just a detail that I noticed when testing, the Webots window is shown quite immediately above the splash screen on Linux and thus the splash screen is not very visible.

Yes I know, this was the case before this PR, and I believe we could fix it only on Windows, not on Linux / macOS.

DavidMansolino:
> > Just a detail that I noticed when testing, the Webots window is shown quite immediately above the splash screen on Linux and thus the splash screen is not very visible.
> 
> Yes I know, this was the case before this PR, and I believe we could fix it only on Windows, not on Linux / macOS.

It disappears when the loading dialog appears right? This is not the desired behavior?

stefaniapedrazzi:
> > > Just a detail that I noticed when testing, the Webots window is shown quite immediately above the splash screen on Linux and thus the splash screen is not very visible.
> > 
> > 
> > Yes I know, this was the case before this PR, and I believe we could fix it only on Windows, not on Linux / macOS.
> 
> It disappears when the loading dialog appears right? This is not the desired behavior?

The slash screen remains open but the Webots window is displayed on top of it.
So instead of seeing the splash screen, I see only the black Webots window.

Screenshot from 2020-09-02 13-46-10

DavidMansolino:
Oh yes you right, I never saw this before, because when the window is in 'maximized' mode it works fine:
image

omichel:
Note that this PR didn't change the behavior, it merely simplify the code.

It works better on Windows where the splash screen remains over the main Webots windows until the world is loaded.
It's not the case for Linux or mac OS, but I believe we failed at fixing it on these platforms.

PR #2193, opened 2020-09-02
Update Webots version on master

stefaniapedrazzi:
Tasks:
* Update version of next release
* Fix `new_version.sh` script (needed after moving in a different directory)
     * use global paths instead of relative ones
     * remove update of `doc.php` that doesn't exist any longer
     * fix duplicated update of `resources/projects/worlds/empty.wbt`

### DIFF
--- a/scripts/new_version/new_version.sh
+++ b/scripts/new_version/new_version.sh
@@ -7,6 +7,11 @@ if [ "$(uname)" == "Darwin" ]; then
   exit 2
 fi
 
+if [[ -z "${WEBOTS_HOME}" ]]; then
+  echo "WEBOTS_HOME is not defined."
+  exit 1
+fi
+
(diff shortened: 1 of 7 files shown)


scripts/new_version/new_version.sh
```diff
@@ -37,38 +42,40 @@ new_version_without_revision=$3
 new_version_year=${new_version:1:4}
 new_version_letter=${new_version:5:1}
 
+CURRENT_DIR=$WEBOTS_HOME/scripts/new_version
```
DavidMansolino, line 45:
```suggestion
CURRENT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```

This way it will work even if we move it.

DavidMansolino (approved):
Thank you

PR #2196, opened 2020-09-02
Notify when the overlay foreground texture is updated

### LINKED ISSUE #2179: Pop-out camera preview black after world reload
**Describe the Bug**

The pop-out camera preview works initially, but becomes irreversibly blank upon project reload.

black_camera_preview

I have also occasionally seen the broken buffer wrapping partially shown previously but it doesn't have a consistent repro.

**Steps to Reproduce**

This is via the following demo project, and DOES NOT seem to appear on the stock camera examples:

[demo.zip](https://github.com/cyberbotics/webots/files/5153599/demo.zip)

Repro steps:

- Open the provided world
- Observe that the camera preview is displayed
- Double-click the camera preview
- Note that the camera preview is still displayed
- Reload the project (without closing Webots)
- The camera view is now black.

**Expected behavior**

To see the live camera preview like is shown within the original window.

**Screenshots**

Included above.

**System**
Pop! OS 20.04.

```
$ webots --version
Webots version: R2020b
$ webots --sysinfo
System: Linux 5.4.0-7634-generic x86_64
Processor: Intel(R) Core(TM) i7-4930K CPU @ 3.40GHz
Number of cores: 12
OpenGL renderer: GeForce GTX 780 Ti/PCIe/SSE2
OpenGL version: 4.6.0 NVIDIA 440.100
```

**Additional context**

Follow-up to separate issue after discussion in #2157. 

I don't have a stock Ubuntu environment handy (neither the will to install natively nor an OpenGL-compatible VM). Minus the window manager tweaks, my pop! environment should be equivalent to an Ubuntu 20.04 system with the proprietary NVIDIA drivers enabled. That being said, naturally, it's possible it's something nuanced about my environment.

stefaniapedrazzi:
Fix #2179: changes to the overlay foreground texture are not notified to the `WbRenderingDeviceWindow`.

`WbRenderingDevice::foregroundTextureIdUpdated()` is defined but never used.
But if the external camera window has to be shown at load (or reload), the foreground texture is reinitialized after the camera window was setup and thus the window still links to an invalid texture id.
The foreground texture is only used in case camera recognition.


DavidMansolino:
:eyes:

DavidMansolino (commented):
It's much better, now the camera image is correctly displayed even after the revert, but it seems there is still an inssue with the foreground image.

First time openning the window:


After revert:


I checked the Display too and in that case, it seems to work fine with the regular-background texture.

stefaniapedrazzi:
Thank you, I will check

stefaniapedrazzi:
The foreground texture and overlay are recreated many times during the configuration.
But if the camera external window is open, during last time the foreground texture is created the WREN context is not active.
This will delay the generation of the foreground texture name (or ID) and thus the robot window receives a null texture id and disables the foreground image.

DavidMansolino (approved):
It works now perfectly, thank you!

PR #2198, opened 2020-09-03
Merge master into develop

DavidMansolino:
**Description**
No conflicts


stefaniapedrazzi (approved):
Is not better to also run the test suite?

DavidMansolino:
> Is not better to also run the test suite?

Good idea, there is indeed a problem because of some 'conflicts' in the ros controller.

PR #2199, opened 2020-09-03
Feature segmentation camera

DavidMansolino:
**Description**
I am opening this PR as a placeholder to save/share this demo.

This is just a draft demo of this feature, currently, it overrides the regular RGB camera, and some design questions are not tackled (i.e. we should probably disable noise, motion blur, etc. on this kind of camera, or is it good to use the recognition colors to distinguish between objects, etc.).

However, it is already stable and functional:
ezgif-6-e8941a91ead0


stefaniapedrazzi:
There are three different types of segmentation:
* _semantic segmentation_: each pixel of the image is assigned to a class but the classification methos doesn't care about the individual instances
* _instance segmentation_: each individual instance of the same class has to be highlighted with a different color in the resulting image. Only instances of a given class (or model) should be detected.
* _panoptic segmentation_: is a mix of the previous two. We need to classify all the pixels in the image as belonging to a class label, yet also identify what instance of that class they belong to.

A complete segmentation output would include a PNG image and the annotation data where the bounding boxes of the detected object with their color are listed.

**Proposed Design**

The proposal of David is good and I would continue in the same direction.
The segmented image and the existing camera recognition functionality are closely related because often the image annotations produced by the Recognition functionality are needed to process the segmented image.
So the output of the two functionalities should be consistent and synchronized.
For this reason, it seems good to use the `Solid.recognitionColors` field to produce the segmented image.
About disabling noise, motion blur, etc. we should use the same rule as for the recognition functionality to decide if they need to be disabled or not.

***Step 1 - Integrate the basic functionality***

* Add segmentation functionality to the `Recognition` and `Camera` node and render the segmented image to a new texture
* Add new field in `Recognition` node to enable/disable the generation of the segmented image, for example:
    `SFBool Recognition.imageSegmentation`
* Add new Camera API functions
    * `wb_camera_recognition_get_segmented_image`
    * `wb_camera_recognition_save_segmented_image`
    * `wb_camera_recognition_set_image_segmentation`: the sampling period should be the same as the recognition functionality, so using `enable/disable` could be confusing
* Show segmented image over the original image (semi-transparent) in the overlay

***Step 2 - Improve segmentation/recognition functionality***

Based on the `Solid.recognitionColors` it is possible to apply all the segmentation methods, but it could be annoying to set the recognition color for each single object in the scene.
So, once the segmentation is fully integrated we should analyze if additional functionalities would be needed to make it easier to use.
For example automatically compute colors for instances of the same model, or specify a subset of models that will be segmented (and ignore all the others).
Any improvement would affect both the recognition and the segmentation functionality, so it would be fine if needed to add new fields in the `Recognition` node.


What do you think?

DavidMansolino:
That sounds very good to me.
I think it is a good idea to match recognition and segmentation, just be aware that one is based on the bounding object (recognition, but maybe this is a bad idea and can be changed) whereas the other is based on the visual part (segmentation) how it is currently implemented.

omichel:
That seems good to me as well. I would simplify the terminology by using "segmentation" instead of "image segmentation", like `SFBool Recognition.segmentation` and `wb_camera_recognition_set_segmentation`. Also we could use multiple color values in `Solid.recognitionColors` to enable panoptic segmentation.

stefaniapedrazzi:
> Also we could use multiple color values in Solid.recognitionColors to enable panoptic segmentation.

Yes, we could think about using for example the first color for the class and the second for the instance.
But this is part of step 2 and I would suggest to analyze it later.
If you have to set the recognitionColors for all the objects (to detect instances) in the scene manually, then we should evaluate if it really makes a different to use multiple colors.
Note that for classes we could also use the `Solid.model` field.

stefaniapedrazzi:
`copyContentsToMemory` works correctly on my development system but not in the test environment.
I'm inspecting the issue in a VM that seems to have the same issue.

stefaniapedrazzi:
For this first implementation `Recognition.maxObjects` and `Recognition.occlusion` properties don't apply to the segmentation image.

omichel:
:eyes: note: this branch has unresolved conflicts, but I can nevertheless start my review.

stefaniapedrazzi:
Oops, strange that I didn't see the conflicts when writing my last comment..

src/webots/nodes/WbCamera.cpp
```diff
@@ -214,6 +231,20 @@ void WbCamera::initializeSharedMemory() {
...
+void WbCamera::initializeSegmentationSharedMemory() {
+  cCameraNumber++;
+  delete mSegmentationShm;
+  mSegmentationShm = initializeSharedMemory();
+  mHasSegmentationSharedMemoryChanged = true;
+  if (mSegmentationShm) {
+    unsigned char *data = (unsigned char *)mSegmentationShm->data();
+    // initialize the shared memory with a black image
+    int *im = reinterpret_cast<int *>(data);
+    for (int i = 0; i < width() * height(); i++)
```
omichel, line 243: For performance, it would be better to cache `width() * height()`
```suggestion
    const size = width() * height();
    for (int i = 0; i < size; i++)
```

omichel (changes requested):
I have some doubt about the API, but it's probably better to chat about it first to clarify things...

src/webots/nodes/WbCamera.cpp
DavidMansolino: Isn't GCC doing this kind of 'simple' optimization this for us?

omichel:
Following our audio discussion, I believe there is a problem with the current design: the `Camera.recognition.segmentation` value should not be changeable from the Camera API, because we would like to be able to provide for example a benchmark with a camera device with the recognition capability, but not the segmentation capability and it should not be possible to change this from the camera API so that users cannot cheat, the segmentation feature being simply not implemented in the proposed camera. This is also more consistent with the device API where devices can't change their own fields set in the scene tree. As a consequence, I would propose the following changes:
1. remove the `wb_camera_recognition_set_segmentation` function.
2. add `wb_camera_recognition_enable_segmentation` and `wb_camera_recognition_disable_segmentation` functions which would enable or disable segmentation, but not change the `Camera.recognition.segmentation` field.
3. change `wb_camera_recognition_is_segmentation_enabled` to `wb_camera_recognition_has_segmentation` (somehow similar to `wb_camera_has_recognition`) that would simply return the value of the segmentation field or false if the Recognition node doesn't exist.

src/webots/nodes/WbCamera.cpp
stefaniapedrazzi: I think that we never measured the performance difference and in most of the cases we didn't store the size in a different variable (for example this is just a copy of what is already done in other classes for example WbLidar, etc.).

But I'm fine with this change.

omichel: I believe gcc cannot guess what the `WbRenderingDevice.width()` method does and whether it will always return the same value, so it will repeatedly call this method (and the `height()` method as well), multiply their return values and compare with `i` at each loop step, which will certainly have a significant impact on the performance of such a simple loop. Depending on the image size, it can be quite big... I believe we should be aware of this when writing new code and probably, we should fix existing code when we spot the problem (e.g., in WbLidar, etc.).

stefaniapedrazzi:
I updated the Recognition segmentation API and addressed the comment about the loop performance.

src/wren/Overlay.cpp
```diff
@@ -59,31 +60,12 @@ namespace wren {
...
-  }
-
-  void Overlay::setForegroundTexture(Texture *texture) {
-    mForegroundTexture = texture;
-
-    if (mForegroundTexture)
-      mParams.mActiveFlags.z = 1.0f;
-    else
-      mParams.mActiveFlags.z = 0.0f;
+      mParams.mActiveFlags.x = mParams.mActiveFlags.x & ~(1u << role);
```
omichel, line 68: It looks like it should be:
```suggestion
      mParams.mActiveFlags.y = mParams.mActiveFlags.y & ~(1u << role);
```
Isn't it?

omichel (changes requested):
It looks very good to me.
What I still miss is:
1. A description of what is a segmentation ground truth image (e.g., for the newbees).
2. A `projects/sample/device/worlds/camera_segmentation.wbt` simulation showing these segmentation images.

Also, maybe @DavidMansolino could also review this PR, as the original author, I guess he will have more critic eyes than me on the implementation details.

src/wren/Overlay.cpp
stefaniapedrazzi: No, the code is correct.
The `mParams.mActiveFlags.y` contains if the border is active or not.
All the textures flags are stored in `mParams.mActiveFlags.x` value.

docs/reference/camera.md
```diff
@@ -1188,6 +1224,25 @@ The `wb_camera_recognition_get_sampling_period` function returns the period give
...
+
+If a [Recognition](recognition.md) node is present in the `recognition` field and the [Recognition](recognition.md).`segmentation` field is set, the camera generates a segmentation ground truth image corresponding to the camera image.
+The segmented image is generated at the same sampling period as the recognition objects.
+For the segmentation to work it is necessary to enable the recognition, but it is not necessary to enable the camera.
+
+The `wb_camera_recognition_has_segmentation` function can be used to check if the [Recognition](recognition.md).`segmentation` field is set to TRUE.
+If the [Recognition](recognition.md) node is not defined, the function returns FALSE.
+
+The `wb_camera_recognition_enable_segmentation` and `wb_camera_recognition_disable_segmentation` functions toggle the generation of the segmented image.
+Note that the generation of the segmented image can only be enabled if the recognition functionality is enabled (see [`wb_camera_has_recognition`](camera.md#wb_camera_has_recognition) and [`wb_camera_recognition_enable`](camera.md#wb_camera_recognition_enable)).
```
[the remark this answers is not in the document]
stefaniapedrazzi: This is not true.
The user has to explicitly enable the segmentation using the `wb_camera_recognition_enable_segmentation` after enabling the recognition.

To me it seems better to let the user have to enable explicitly the segmentation feature rather than force him to explicitly disable it when he doesn't want it.

DavidMansolino:
> Also, maybe @DavidMansolino could also review this PR, as the original author, I guess he will have more critic eyes than me on the implementation details.

Yes sure, as soon as you are done with the review, I will have a check.

docs/reference/camera.md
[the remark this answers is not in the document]
omichel: I agree it is better this way, so we should probably state it clearly in the documentation as well, as I was unsure of the behavior from the description, like:
```
Even if the [Recognition](recognition.md).`segmentation` field is TRUE, the generation of segmented images is not automatically enabled when calling the [`wb_camera_recognition_enable`](camera.md#wb_camera_recognition_enable) function, so it is necessary to call the `wb_camera_recognition_enable_segmentation` function to enable the generation of segmented images.
```

stefaniapedrazzi:
I addressed all the comments and checked the behavior so that it matches the documentation:
- the camera segmentation image generation has to be explicitly enabled from the controller
- the camera segmentation image can be generated also if the camera is disabled
- the camera segmentation image and the recognized objects information are displayed in the overlay on if the camera image is enabled

docs/reference/recognition.md
```diff
@@ -16,6 +17,18 @@ The [Recognition](#recognition) node provides a [Camera](camera.md) device with
 When a [Camera](camera.md) device has a [Recognition](#recognition) node in its `recognition` field, it is able to recognize which objects are present in the camera image.
 Only [Solids](solid.md) whose `recognitionColors` field is not empty can be recognized by the camera.
 
+Additionally, the [Recognition](#recognition) also provides the `segmentation` functionalities to generate segmentation ground truth images displaying the recognized objects.
```
omichel, line 20:
```suggestion
Additionally, the [Recognition](#recognition) also provides the `segmentation` functionality to generate segmentation ground truth images displaying the recognized objects.
```

docs/reference/recognition.md
```diff
@@ -16,6 +17,18 @@ The [Recognition](#recognition) node provides a [Camera](camera.md) device with
...
+In the segmentation image, each pixel will be colored using the `recognitionColors` first item of the corresponding object rendered from the [Camera](camera.md) device.
```
omichel, line 21: That seems clearer to me this way:
```suggestion
In the segmentation image, each pixel will be colored using the first item of the `recognitionColors` of the corresponding object rendered from the [Camera](camera.md) device.
```

omichel (changes requested):
When running the example, it is strange that the main camera image is changed when enabling segmentation (like the main camera image is blended with the segmentation image) and the segmented image is not actually displayed. It would be nice to see it in a Display device if not possible to display it otherwise.

src/webots/wren/WbWrenCamera.cpp
```diff
@@ -425,13 +427,23 @@ void WbWrenCamera::render() {
...
   if (mIsSpherical)
     applySphericalPostProcessingEffect();
-
+  else if (mUpdateTextureFormatEffect) {
+    WrPostProcessingEffectPass *updateTextureFormatPass =
+      wr_post_processing_effect_get_pass(mUpdateTextureFormatEffect, "PassThrough");
+    wr_post_processing_effect_pass_set_input_texture(updateTextureFormatPass, 0,
+                                                     WR_TEXTURE(wr_frame_buffer_get_output_texture(mResultFrameBuffer, 0)));
+    wr_post_processing_effect_apply(mUpdateTextureFormatEffect);
+  }
```
DavidMansolino, line 446: That's the only part I am not sure to understand.
We now need a post-processing effect in any case, this is to make sure the texture format is correct in every case?

DavidMansolino (commented):
It works perfectly, and the code is very clean! I especially like the way it is displayed in the overlay!

Just a few comments:
  - The recognitionColor of the `BiscuitBox (PROTO)` should probably be fixed/improved (currently black).
  - It might be good (and probably quite simple) to set the background color of the 'segmented image' to be the one of the `Background.skyColor` field (imagine you want to see the segmented image of dark objects, it might be useful to set the background to a lighter color).
  - It would be good to be able to enable recognition and segmentation from the default robot-window with additional checkboxes (but this might be a future feature, maybe doesn't need to be integrated into this PR, just an idea, I can open an issue about this if you want).

src/webots/wren/WbWrenCamera.cpp
stefaniapedrazzi: The format of the "rendering" texture is different than the "output" texture (passed to the controller).
For all type of previously existing cameras (color, radar and lidar) there are post processing effect(s) that, among other things, save the "output" texture in the correct format.
The segmentation type is the only one that doesn't have post-processing effects.

The "rendering" texture is also used internally by Webots for example for the overlays etc.
So using the "rendering" texture only or the "output" texture only will create conflicts with the format of the other textures.

resources/wren/shaders/overlay.frag
```diff
@@ -19,21 +20,25 @@ out vec4 fragColor;
...
 
 layout(std140) uniform Overlay {
   vec4 positionAndSize;
   vec4 defaultSize;  // x,y: size, z: render default size instead of actual overlay
   vec4 borderColor;
   vec4 backgroundColor;
-  vec4 activeFlags;   // x: bg texture, y: main texture, z: fg texture, w: border
   vec4 textureFlags;  // x: flip vertically, y: additional texture count, z: maxRange (depth textures only),
                       // w: overlay transparency
+  uvec2 activeFlags;  // x: textures, y: border
```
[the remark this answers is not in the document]
stefaniapedrazzi: It not used in the `overlay.frag` but it is used in the `overlay.vert`.

src/webots/wren/WbWrenCamera.cpp
DavidMansolino: Ok, that makes perfectly sense, thanks a lot for the explanation!

stefaniapedrazzi:
Thank you for the review and the suggestions.

>Just a few comments:
    * The recognitionColor of the BiscuitBox (PROTO) should probably be fixed/improved (currently black).

Yes, for the correctness of the PROTO model (and probably in the master branch).
But for the segmentation example it is not essential. It could also be good to show that if the recognitionColor is set to black the object is not visible in the segmentation image.

>    * It might be good (and probably quite simple) to set the background color of the 'segmented image' to be the one of the Background.skyColor field (imagine you want to see the segmented image of dark objects, it might be useful to set the background to a lighter color).

Yes, it is a good idea to let the user choose the background color.
But I wouldn't use the `Background.skyColor` because it will limit the possibilities if a non-textured background is used in the world.
Given that in any case the segmentation functionality needs to be enabled from the controller. I personally would rather add a new API function to set the background color (default color will remain black).

>    * It would be good to be able to enable recognition and segmentation from the default robot-window with additional checkboxes (but this might be a future feature, maybe doesn't need to be integrated into this PR, just an idea, I can open an issue about this if you want).

Yes, I agree. I didn't add it in this PR because it is currently not implemented for the recognition too.
I will open a new issue for this.

DavidMansolino:
> Thank you for the review and the suggestions.
> 
> > Just a few comments:
> > * The recognitionColor of the BiscuitBox (PROTO) should probably be fixed/improved (currently black).
> 
> Yes, for the correctness of the PROTO model (and probably in the master branch).
> But for the segmentation example it is not essential. It could also be good to show that if the recognitionColor is set to black the object is not visible in the segmentation image.
> 
> > * It might be good (and probably quite simple) to set the background color of the 'segmented image' to be the one of the Background.skyColor field (imagine you want to see the segmented image of dark objects, it might be useful to set the background to a lighter color).
> 
> Yes, it is a good idea to let the user choose the background color.
> But I wouldn't use the `Background.skyColor` because it will limit the possibilities if a non-textured background is used in the world.
> Given that in any case the segmentation functionality needs to be enabled from the controller. I personally would rather add a new API function to set the background color (default color will remain black).
> 
> > * It would be good to be able to enable recognition and segmentation from the default robot-window with additional checkboxes (but this might be a future feature, maybe doesn't need to be integrated into this PR, just an idea, I can open an issue about this if you want).
> 
> Yes, I agree. I didn't add it in this PR because it is currently not implemented for the recognition too.
> I will open a new issue for this.

That sounds all good to me!

stefaniapedrazzi:
I addressed all the comments and  synchronized with `develop`.

But macOS test is failing during the installation of the Webots compilation dependencies.
https://github.com/cyberbotics/webots/pull/2199/checks?check_run_id=1320970862#step:3:1230

stefaniapedrazzi:
macOS failure is not related to this PR (I got the same error on another PR targeting the master branch).

DavidMansolino (commented):
Approved :+1:  (can't approve since I am the original author)

I confirm failure is unrelated.

omichel (approved):
All good for me.
Thank you!

PR #2200, opened 2020-09-03
Fix Webots version in github workflow

stefaniapedrazzi:
Complete #2193:
Fix version in `.github/workflows/test_suite_mac.yml` and inspect if it not possible to automatically retrieve the version.

### DIFF
--- a/.github/workflows/test_suite_mac.yml
+++ b/.github/workflows/test_suite_mac.yml
@@ -68,7 +68,7 @@ jobs:
         path: artifact
     - name: Extract Webots
       run: |
-        hdiutil mount artifact/webots-R2020b-rev2.dmg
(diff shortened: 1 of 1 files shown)


.github/workflows/test_suite_mac.yml
```diff
@@ -68,15 +68,15 @@ jobs:
         path: artifact
     - name: Extract Webots
       run: |
-        hdiutil mount artifact/webots-R2020b-rev1.dmg
+        hdiutil mount artifact/webots-R2020b-rev2.dmg
```
DavidMansolino, line 71:
```suggestion
        hdiutil mount artifact/webots*.dmg
```

FYI, I quickly checked on the mac and this should work :-)

DavidMansolino (approved):
Thank you

PR #2201, opened 2020-09-03
Improve Export Functionnality.

DavidMansolino:
**Description**
There was no check for the extension entered by the user when exporting a node.

### DIFF
--- a/src/webots/scene_tree/WbSceneTree.cpp
+++ b/src/webots/scene_tree/WbSceneTree.cpp
@@ -1476,16 +1476,38 @@ void WbSceneTree::exportObject() {
     topLevelWidget, tr("Export Webots Object"),
     WbProject::computeBestPathForSaveAs(WbPreferences::instance()->value("Directories/objects").toString() + "/" +
                                         mSelectedItem->node()->modelName() + ".wbo"),
-    tr("Files (*.wbo *.WBO)"));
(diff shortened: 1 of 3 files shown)


src/webots/vrml/WbNode.hpp
```diff
@@ -278,6 +278,7 @@ class WbNode : public QObject {
   // void printDebugNodeStructure(int level = 0);
   // void printDebugNodeFields(int level, bool printParameters);
   const WbNode *findRobotRootNode() const;
+  virtual const bool isRobot() const { return false; };
```
omichel, line 281: Shouldn't we make it public also in the WbRobot class?

omichel (approved):
Thank you.

src/webots/scene_tree/WbSceneTree.cpp
```diff
@@ -1476,16 +1476,38 @@ void WbSceneTree::exportObject() {
...
+
+  if (fileName.endsWith("urdf", Qt::CaseInsensitive) && !mSelectedItem->node()->isRobot()) {
+    WbLog::error(tr("URDF exportation available only for Robot nodes."));
+    return;
+  }
 
   QFile file(fileName);
-  if (!file.open(QIODevice::WriteOnly))
+  if (!file.open(QIODevice::WriteOnly)) {
+    WbLog::error(tr("File not writable."));
```
stefaniapedrazzi, line 1508: Maybe it would be good to also print the path of the file that cannot be written.
Like it is done here:
https://github.com/cyberbotics/webots/blob/66116d3fcd3dc42aa739ff609a0430a73b530f4a/src/webots/gui/WbVideoRecorder.cpp#L553

PR #2203, opened 2020-09-04
Improve northDirection Field Warning

DavidMansolino:
**Description**
Is some cases the north direction (used for GPS for example) was changed without explicitely warning the user.


omichel (dismissed):
It's better this way.
Thank you.

DavidMansolino:
I just fixed an issue causing the warning to be raised when the field was not defined in the world file.

omichel (approved):

PR #2204, opened 2020-09-04
Add EUN to NUE converter.

DavidMansolino:
**Description**
Old world before R2020b are often using north direction set to '0 0 1', this is unfortunately not compatible with any of the coordinate system of Webots (ENU or NUE), this PR adds a script that transforms the world to NUE.

### DIFF
--- a/scripts/converter/convert_eun_to_nue.py
+++ b/scripts/converter/convert_eun_to_nue.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python3
+
+# Copyright 1996-2020 Cyberbotics Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Convert R2020b world file from the EUN (northDirection = '0 0 1') to the NUE coordinate system."""
+
+import math
+import sys
+
+from transforms3d import quaternions
+
+from webots_parser import WebotsParser
+
+
+def rotation(value, r):
+    q0 = quaternions.axangle2quat([float(value[0]), float(value[1]), float(value[2])], float(value[3]))
+    q1 = quaternions.axangle2quat([r[0], r[1], r[2]], r[3])
+    qr = quaternions.qmult(q1, q0)
+    v, theta = quaternions.quat2axangle(qr)
+    return [WebotsParser.str(v[0]), WebotsParser.str(v[1]), WebotsParser.str(v[2]), WebotsParser.str(theta)]
+
+
+def convert_to_nue(filename):
+    world = WebotsParser()
+    world.load(filename)
+
+    for node in world.content['root']:
+        if node['name'] == 'WorldInfo':
+            for field in node['fields']:
+                if field['name'] == 'northDirection':
+                    assert field['value'] == ['0', '0', '1']
+                    # remove the 'northDirection 0 0 1'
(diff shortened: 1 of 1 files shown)


DavidMansolino:
After using the conversion script, the difference in the GPS measurement in my test case is much smaller (but there is still a very small difference that I can't yet explain):

| Webots Version | GPS Values |
| --- | --- |
| R2020a | [36.7472412750660, -77.75337766408911, 73.92595425897193] |
| R2020b (without conversion) | [36.7508864539240, -77.76107238252905, 73.92595425897193] |
| R2020b (after conversion) | [36.7472412747659, -77.75337766431713, 73.92595547242688] |

DavidMansolino:
The absolute position of the GPS is exactly the same (in the tree), but the position set to the lat-long conversion function is not exctaly the same.

omichel (approved):
Thank you.

HefnySco:
@DavidMansolino Many thanks for this useful tool. I tried to convert my worlds and it worked well. However Inertia Sensor reports strange values "roll": -1.569329,"pitch": 1.570961,"yaw": 1.570041} while it should return 0,0,0.
could you advise what I am missing here?

DavidMansolino:
Did you first opened and saved the world in Webots, before starting the script?

bondolo (commented):
I had a question on the scripts

scripts/converter/convert_eun_to_nue.py
```diff
@@ -0,0 +1,68 @@
...
+    world = WebotsParser()
+    world.load(filename)
+
+    for node in world.content['root']:
+        if node['name'] == 'WorldInfo':
+            for field in node['fields']:
+                if field['name'] == 'northDirection':
+                    assert field['value'] == ['0', '0', '1']
+                    # remove the 'northDirection 0 0 1'
+                    del node['fields'][node['fields'].index(field)]
```
bondolo, line 45: Would you not need to add a 'coordinate System "NUE"' node since the default is ENU?

DavidMansolino: Yes, this could indeed help, however, if you open a world from R2020a in Webots it will automatically set the coordinate system field to NUE. But that's anyway a good idea to add this filed here if not present.

bondolo: Thank you, it was not clear that the default coordinate system varied based on the webots version in the first line of the world file. I had understood we needed to convert everything to ENU for Webots R2020b+

DavidMansolino: Yes, this mechanism is indeed hidden in Webots, it is there to assure backward compatibility is not broken. In any case, I implemented your suggestion here: #2217

HefnySco:
Hi @DavidMansolino 

> Did you first opened and saved the world in Webots, before starting the script?

Sorry for taking time to answer. I need to make full test.
The converted model as is gave me bad readings of inertia control and all other controls. So I had to rotate all sensors  [0 1 0 1.5708] so that inertia gives me pitch and roll correctly -I had to make also minor changes to code- to change axis.
Again I have not succeed to make my quadcopter fly. I revisited acc & gyro readings and made sure that it is as expected by the autopilot software, still no good result :( sorry.

HefnySco:
Another thing that the converted world made issue for lighting in the robot.


dark_after_convertion
light_original _world

Please let me know if I should post these issues in another thread or as an issue rather than this pr.

DavidMansolino:
> Hi @DavidMansolino
> 
> > Did you first opened and saved the world in Webots, before starting the script?
> 
> Sorry for taking time to answer. I need to make full test.
> The converted model as is gave me bad readings of inertia control and all other controls. So I had to rotate all sensors [0 1 0 1.5708] so that inertia gives me pitch and roll correctly -I had to make also minor changes to code- to change axis.
> Again I have not succeed to make my quadcopter fly. I revisited acc & gyro readings and made sure that it is as expected by the autopilot software, still no good result :( sorry.

That makes indeed sense for the inertial unit.
About the light, there is unfortunately currently no way to rotate the background (which is casting the lights), therefore the whole scene is rotated but not the lighting setup.

DavidMansolino:
I just opened an issue do discuss about improving the support for this: #2220

PR #2207, opened 2020-09-04
Improve Linux Runtime Dependencies Script

DavidMansolino:
**Description**
The `lsb-release` package is not always available and is required both by the script and by some of the Webots include Makefiles.


omichel (approved):
Thank you.

Issue #2208, opened 2020-09-06
Memory leak in controller process when rangefinder is enabled and robot window is visible


WasabiFan:
**Describe the Bug**

There appears to be a memory leak triggered in the controller's process (not the Webots host process) when the following conditions are met:
- There is an enabled rangefinder
- The "Robot Window" on the left of the Webots application is open

My PC has frozen a few times in the past day or two, and it's taken me a while to isolate the problem. Earlier today I caught my Python controller using just under 6GB of RAM. This combination of circumstances _seems_ strange to me (the controller process being affected by the "robot window") but I suppose there could be e.g. a broken reference counter involved somewhere. 

The image preview in the robot window doesn't have to be visible, and both C++ and Python controllers appear to have the same issue. It also seems like, once the "robot window" has been opened, closing it does not avoid the issue.

I have not been able to reproduce this with other similar sensors, e.g. the camera, but that may just be user error.

**Steps to Reproduce**
1. Load the `devices/worlds/rande_finder.wbt` sample world
2. In System Monitor (or equivalent), observe that there is a `range_finder` process using ~1MB or less of RAM, constant.
3. In Webots, right-click the `Robot` node and choose "Show Robot Window"
4. Observe the memory usage of `range_finder` begin to increase (this appears unbounded)

**Expected behavior**

Constant memory usage and my PC doesn't crash.

**Screenshots**

State after on the order of 10mins:

![image](https://user-images.githubusercontent.com/3310349/92322322-63d9f880-efe5-11ea-9b0b-49caa845f5b9.png)

30-sec gif illustrating the behavior:

![memleak](https://user-images.githubusercontent.com/3310349/92322432-03978680-efe6-11ea-8fd5-b2f855d505e9.gif)

**System**
Pop! OS 20.04

```
System: Linux 5.4.0-7634-generic x86_64
Processor: Intel(R) Core(TM) i7-4930K CPU @ 3.40GHz
Number of cores: 12
OpenGL vendor: NVIDIA Corporation
OpenGL renderer: GeForce GTX 780 Ti/PCIe/SSE2
OpenGL version: 4.6.0 NVIDIA 440.100
```

```
Webots version: R2020b revision 1
```

**Additional context**

N/A

DavidMansolino:
Thank you for these precise bug report, I can indeed reproduce this with range-finders and lidars, we will fix this for the next release of Webots.

PR #2210, opened 2020-09-07
Fix Memory Leak in Robot-Window

DavidMansolino:
This pull-request fixes issue  #2208

### DIFF
--- a/docs/reference/changelog-r2020.md
+++ b/docs/reference/changelog-r2020.md
@@ -1,5 +1,11 @@
 # Webots R2020 Change Log
 
+## Webots R2020b Revision 2
+Released on XXX YYY, 2020.
+
+  - Bug fixes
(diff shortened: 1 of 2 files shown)


omichel (commented):
Thank you!
I would call the `free` at the end of `append_rescaled_image_to_buffer` function as this is closer to where the `malloc` is performed (and would factorize a couple of lines of code).

DavidMansolino:
> Thank you!
> I would call the `free` at the end of `append_rescaled_image_to_buffer` function as this is closer to where the `malloc` is performed (and would factorize a couple of lines of code).

Note that the malloc is not performed in `append_rescaled_image_to_buffer` but in `g_image_downscale` but that's fine for me, but to be clear I would then rename the function to something like `append_rescaled_image_to_buffer_and_free` (because if the image is already RGB, it might looks strange that after calling this function `img->data` is NULL).

omichel (approved):
Thank you.

PR #2213, opened 2020-09-07
Cleanup Unused 'mAnchor' WbJoint Member

DavidMansolino:
**Description**
This member is never assigned and is highly confusing as one can think it does represent the actual joint anchor.


omichel (approved):
Well spotted!

PR #2215, opened 2020-09-07
Fix noise generation in GLSL

### LINKED ISSUE #2139: Noise becomes fixed patterned.
Set noise field of Camera node to any positive value, for example 0.2, and start simulation. At the beginning image noise works fine, which is random noise in 3 channels: 
freeze_noise0

But minutes later(simulation time), the image noise becomes fixed patterned, and is only black-and-white:
freeze_noise1

Lidar noise suffers from the same issue.

This issue exists on both 2020a-rev1 and 2020b-rev1-nightly-build. I tested with ubuntu 16.04 and ROS kinetic.

stefaniapedrazzi:
Fix #2139: fix generation of pseudo random values in GLSL shaders that didn't properly work on RTX Nvidia GPUs.


DavidMansolino:
👀

DavidMansolino (approved):
Thank you, I tested on the Mac-mini with the intel GPU and it works fine too.

PR #2217, opened 2020-09-08
Force 'NUE' in convert_eun_to_nue script

DavidMansolino:
Follow-up of https://github.com/cyberbotics/webots/pull/2204#pullrequestreview-483720264

### DIFF
--- a/scripts/converter/convert_eun_to_nue.py
+++ b/scripts/converter/convert_eun_to_nue.py
@@ -45,7 +45,7 @@ def convert_to_nue(filename):
                     # remove the 'northDirection 0 0 1'
                     del node['fields'][node['fields'].index(field)]
                 if field['name'] == 'coordinateSystem':
-                    field['value'] == 'NUE'
+                    field['value'] = 'NUE'
                     is_nue = True
             if not is_nue:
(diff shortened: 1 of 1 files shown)


scripts/converter/convert_eun_to_nue.py
```diff
@@ -38,11 +38,19 @@ def convert_to_nue(filename):
...
     for node in world.content['root']:
         if node['name'] == 'WorldInfo':
+            is_nue = False
             for field in node['fields']:
                 if field['name'] == 'northDirection':
                     assert field['value'] == ['0', '0', '1']
                     # remove the 'northDirection 0 0 1'
                     del node['fields'][node['fields'].index(field)]
+                if field['name'] == 'coordinateSystem':
+                    field['value'] == 'NUE'
```
stefaniapedrazzi, line 48: Is this a comparison or an assignment?
Maybe it should be
```suggestion
                    field['value'] = 'NUE'
```

DavidMansolino: 😨 you're right well spotted, I am surprised it worked when I tested it.

stefaniapedrazzi (approved):

Issue #2220, opened 2020-09-09
EUN Worlds


DavidMansolino:
**Describe the Bug**
Many users were using the East-Up-North convention before (mainly users using the OSM importer), i.e. setting the nordDirection to '0 0 1'.

It is currently not so easy to convert the world, a script is available (https://github.com/cyberbotics/webots/blob/master/scripts/converter/convert_eun_to_nue.py), but unfortunately, the conversion is not complete in the sense that it can't rotate the background and some sensors rely on the north direction.

It would probably be good (and not so complicated) to introduce a new 'EUN' option in the coordinateSystem field of the WorldInfo node.

This should be added in:
  - The rotation of the background: https://github.com/cyberbotics/webots/blob/master/src/webots/nodes/WbBackground.cpp#L65
  - Update of the TexturedBackgroundLight: https://github.com/cyberbotics/webots/blob/master/projects/objects/backgrounds/protos/TexturedBackgroundLight.proto#L105
  - Handle this new coordinate system in sensors using it: Compass, Gps and InertialUnit

PR #2228, opened 2020-09-11
EUN Coordinate System

DavidMansolino:
This PR implements:  #2220

In addition, it brings several cleanups from #1643 and generalizes the coordinate system implementation (it will be extremely simple in the future to add more coordinate system).

### DIFF
--- a/src/webots/nodes/WbBackground.cpp
+++ b/src/webots/nodes/WbBackground.cpp
@@ -66,7 +66,7 @@ static int gCoordinateSystemSwap(int i) {
   static const int enu_swap[] = {5, 4, 0, 1, 3, 2};
   if (WbWorld::instance()->worldInfo()->coordinateSystem() == "ENU")
     return enu_swap[i];
-  else
+  else  // "NUE" or "EUN"
     return i;
 }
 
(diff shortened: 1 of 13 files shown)


src/webots/nodes/WbBackground.cpp
```diff
@@ -66,15 +66,15 @@ static int gCoordinateSystemSwap(int i) {
   static const int enu_swap[] = {5, 4, 0, 1, 3, 2};
   if (WbWorld::instance()->worldInfo()->coordinateSystem() == "ENU")
     return enu_swap[i];
-  else
+  else  // "NUE" or "EUN"
```
DavidMansolino, line 69: I am not sure in the EUN case it makes sense to rotate the background.
In the ENU case it clearly makes sense since the sky should be up, but here the sky is already up, and in the background, there is no notion of north and east but rather front and right. Furthermore, if we want to keep 100% compatibility with < R2020b worlds with northDirection set to '0 0 1' we should not rotate it. 

@cyberbotics/maintainers what do you think?

omichel (approved):
Thank you.

PR #2231, opened 2020-09-11
Run Test in the Github Action Nightlies

DavidMansolino:
It is good to make sure that these tests are run at least once per day (unlike on Travis, the nightly build will be released even if the test fails like on AppVeyor).


omichel (approved):
Thank you.

PR #2233, opened 2020-09-11
Improve docker doc

DavidMansolino:
**Description**
It is probably good to use the 'headless' keyword here as it is often mentioned by users.


omichel (approved):
Indeed.

Issue #1898, opened 2020-07-10
Access to devices imported during simulation


tsampazk:
**Is your feature request related to a problem? Please describe.**
A previously exported device node .wbo that is imported into the simulation on runtime cannot be accessed with the ```getDevice()``` methods through a Robot controller.

This makes dynamic Robot creation impossible in a straightforward way, using the import* methods.

**Describe the solution you'd like**
To somehow be possible to access previously imported devices during runtime.

**Describe alternatives you've considered**
1. When building a dynamic robot, import (or preload) it with all the devices and nodes you might need and use the creation algorithm to delete nodes instead of importing, but this seems cumbersome and backwards.
2. Access the devices some other way, e.g. through their respective node refs, which seems to also not be possible.

Thank you!

tsampazk:
A workaround i am working on is to dynamically create the Robot node with all the appropriate devices etc. as a VRML string by using a custom parser and loading it altogether using ```importMFNodeFromString``` method, from another supervisor. Then the robot controller will have access to devices normally because it was initialized with them from the start.

omichel:
Yes, that seems to be an excellent approach. Let us know whether it works as expected.

stefaniapedrazzi:
The issue here is that the devices are all configured at the very first start of the controller program.
So restarting the controller or (re-)creating the robot node both should work for you.

The workaround you described is perfectly fine.
But just in case another option would also be to just restart the controller of the robot after importing all the devices [wb_supervisor_node_restart_controller](https://www.cyberbotics.com/doc/reference/supervisor#wb_supervisor_node_restart_controller)).

There is a third option that is to generate the Robot using procedural PROTO node with flags to choose the devices but depending on the application could be more complex than needed.

tsampazk:
> But just in case another option would also be to just restart the controller of the robot after importing all the devices [wb_supervisor_node_restart_controller](https://www.cyberbotics.com/doc/reference/supervisor#wb_supervisor_node_restart_controller)).

This workaround indeed works great and is probably way simpler to use than parsing VRML strings etc. 

Here is a barebones example:

```
from controller import Supervisor

robot = Supervisor()
timestep = int(robot.getBasicTimeStep())

stringNode = """Solid {
  children [
    LightSensor {
    }
  ]
}"""

ls = robot.getLightSensor("light sensor")

if ls is None:
    robot.getSelf().getField("children").importMFNodeFromString(0, stringNode)
    robot.getSelf().restartController()
else:
[...]

PR #2237, opened 2020-09-14
Add device during robot controller run

stefaniapedrazzi:
**Description**
Address issue #1898.
Previously devices added after the robot controller was started where not accessible from the robot controller.
This PR adds a new message to notify the libController that new devices have been added and configures them so that they can 
be used in a robot controller.

**Tasks**
  - [x] Add new `C_ROBOT_NEW_DEVICE` (simulation -> libController) message to configure devices independently from `C_CONFIGURE` action
  - [x] In  `WbRobot` extend method that assigns the device tag and notify the libController when new devices have been added
  - [x] Update documentation if needed


DavidMansolino:
This probably deserves a test :wink:

stefaniapedrazzi:
I just added a test

DavidMansolino:
:eyes:

src/Controller/api/robot.c
```diff
@@ -132,6 +132,17 @@ static void init_remote_control_library() {
   }
 }
 
+static void init_devices(WbRequest *r, int firstTag) {
```
DavidMansolino, line 135:
```suggestion
static void init_devices_from_tag(WbRequest *r, int firstTag) {
```

This is just an idea.

tests/api/controllers/supervisor_import_robot_node/supervisor_import_robot_node.c
```diff
@@ -40,6 +41,20 @@ int main(int argc, char **argv) {
...
+  wb_supervisor_field_import_mf_node_from_string(self_children, 0, "Camera { name \"imported camera\"}");
+
+  WbDeviceTag camera = wb_robot_get_device("imported camera");
+  ts_assert_int_not_equal(camera, 0, "Camera imported during controller execution not found.");
+  wb_camera_enable(camera, TIME_STEP);
+
+  wb_robot_step(TIME_STEP);
+
+  const unsigned char *image = wb_camera_get_image(camera);
+  ts_assert_boolean_not_equal(image == NULL, "Camera image is NULL.");
```
DavidMansolino, line 56:
```suggestion
  ts_assert_pointer_not_null(image, "Camera image is NULL.");
```

src/webots/nodes/WbRobot.cpp
```diff
@@ -727,6 +734,22 @@ void WbRobot::keyReleased(const QString &text, int key) {
   emit keyboardChanged();
 }
 
+void WbRobot::writeDeviceConfigure(QList<WbDevice *> devices, QDataStream &stream) const {
+  QListIterator<WbDevice *> it(devices);
+  while (it.hasNext()) {
+    WbDevice *device = it.next();
```
DavidMansolino, line 740:
```suggestion
    const WbDevice *device = it.next();
```

(I know this was already like this)

src/webots/nodes/WbRobot.cpp
```diff
@@ -727,6 +734,22 @@ void WbRobot::keyReleased(const QString &text, int key) {
...
+    stream << (short int)device->deviceNodeType();
+    QByteArray n(device->deviceName().toUtf8());
+    stream.writeRawData(n.constData(), n.size() + 1);
+    WbSolidDevice *solidDevice = dynamic_cast<WbSolidDevice *>(device);
```
DavidMansolino, line 744:
```suggestion
    const WbSolidDevice *solidDevice = dynamic_cast<WbSolidDevice *>(device);
```

DavidMansolino (changes requested):
Note that the robot window is not updated when inserting new devices (we probably don't want to support this, but maybe we can document this somewhere).

If I had a camera node at runtime controller crashes.

tests/api/controllers/supervisor_import_robot_node/supervisor_import_robot_node.c
stefaniapedrazzi: Your suggestion throws errors on compilation:
```
supervisor_import_robot_node.c:56:30: warning: passing argument 1 of ‘ts_assert_pointer_not_null’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
```

We could add a new function taking constant pointer into account, but it works perfectly fine with the `ts_assert_boolean_*` existing function so for me it would be useless.

DavidMansolino: Ah ok you right, in that case, lets keep the `ts_assert_boolean_*` indeed.

stefaniapedrazzi:
> Note that the robot window is not updated when inserting new devices (we probably don't want to support this, but maybe we can document this somewhere).
> 
> If I had a camera node at runtime controller crashes.

The controller crashes only if the robot window is open, is it correct?

DavidMansolino:
Probably yes, let me retry to be sure.

DavidMansolino:
> Probably yes, let me retry to be sure.

I confirm it is only when the robot-window is open.

stefaniapedrazzi:
I fixed the crash of the default robot window.
I will now check if there is a simple way to update the list of devices once the robot window is already open.

stefaniapedrazzi:
I could easily improve the default robot window so that the new devices are automatically displayed (during the next controller step).

DavidMansolino:
:eyes:

DavidMansolino (approved):
It now works like a charm!
Thank you, it is very nice to see this long-term bug finally resolved in such a clean way :-)

PR #2250, opened 2020-09-18
Allow upload of C3D files from the robot window

omichel:
This PR adds the capability of uploading C3D files through the robot window, including from the web.
It also fixes:
- a bug in `webots.js`: the previous robot windows were not closed upon reload.
- the `server.sh` script, so that it runs on Windows as well.


stefaniapedrazzi:
:eyes:

stefaniapedrazzi (approved):

projects/humans/c3d/controllers/c3d_viewer/c3d_viewer.py
```diff
@@ -245,46 +267,51 @@ def getPointsList(reader, name):
...
-            bodyTransparency = float(value[1])
-            bodyTransparencyField.setSFFloat(bodyTransparency)
+            c3dfile.bodyTransparency = float(value[1])
+            c3dfile.bodyTransparencyField.setSFFloat(c3dfile.bodyTransparency)
         elif action == 'speed':
             playbackSpeed = float(value[1])
             offsetTime = supervisor.getTime()
             totalFrameCoutner = 0
+        elif action == 'c3dfile':
+            with open('default.c3d', 'wb') as file:
```
DavidMansolino, line 297: If the user changed the file in the tree before, this will have no effect.

projects/humans/c3d/controllers/c3d_viewer/c3d_viewer.py
```diff
@@ -245,46 +267,51 @@ def getPointsList(reader, name):
...
+                file.write(base64.b64decode(value[1]))
```
DavidMansolino, line 298: Will this work in the released package?
We probably don't want to change files in the installation directory.

projects/humans/c3d/controllers/c3d_viewer/c3d_viewer.py
```diff
@@ -44,33 +45,195 @@
...
-    """Get a group of points and extract it's labels as a list of strings."""
-    if name not in reader.groups['POINT'].params:
-        return None
-    list = reader.groups['POINT'].get_string(name)
-    elementSize = reader.groups['POINT'].get(name).dimensions[0]
-    newlist = [list[i:i + elementSize] for i in range(0, len(list), elementSize)]
-    for i in range(len(newlist)):
-        newlist[i] = newlist[i].strip()
-    return newlist
+class c3dFile:
```
DavidMansolino, line 48: It is a very good idea to create a class for this, it is now much cleaner, but the name of the class looks odd to me as at first glance it me looks like it is handling C3D files, but it is actually doing much more than that (C3D file reading is only a minor part of the class, it is also handling markers, robot-window communication, etc.)

omichel: Yes, this class does a lot of things in relationship with a C3D file, so I believe that *c3dFile* is a good name for this class. I would like to avoid *c3dManager* or the like which sounds less precise to me. Do you have a better idea?

DavidMansolino: Maybe simply `C3dViewer` as it is the name of the controller/file.
But that's not very important, I was just breafly confused when first reading the name of the class (maybe just a brief class description comment would be enough).

DavidMansolino: > I would like to avoid c3dManager

I completely agree, in 99% of the cases, such kind of 'manager' keyword are completely useless.

omichel: OK, I added some comment to better describe what this class does. `c3dFile` sounds better to me than `c3dViewer` as it basically loads and initializes a C3D file, but doesn't actually display it...

PR #2252, opened 2020-09-20
docs: fix acceleration units for rotational motors

WasabiFan:
Omitting detailed description because I think this is fairly self-explanatory.


DavidMansolino (approved):
Thank you!

PR #2255, opened 2020-09-21
Fix forgeUrl in viewer.js

stefaniapedrazzi:
The `isFirstArgument` value was not set correctly and the URL could have the form ``localhost?book=reference?page=index`` instead of ``localhost?book=reference&page=index``


omichel (commented):
I believe this should be fixed in a much more simple way.
Because the `book` argument is set in any case, we don't need any more the `isFirstArgument` variable after this point and can safely append other parameters with `&param=value`.

omichel (approved):
Nice clean-up.
Thank you.

PR #2260, opened 2020-09-22
Fixed startup of extern controllers with WEBOTS_PID set

omichel:
Fixed synchronization bug when starting an extern controller while specifying the `WEBOTS_PID` environment variable.

### DIFF
--- a/src/Controller/api/system.c
+++ b/src/Controller/api/system.c
@@ -177,6 +177,7 @@ const char *wbu_system_webots_tmp_path() {
       struct dirent *entry;
       char folder_start[32];
       sprintf(folder_start, "webots-%d", webots_pid);
+      webots_pid = 0;
       while ((entry = readdir(dir))) {
         if (strncmp(entry->d_name, folder_start, strlen(folder_start)) == 0) {
           struct stat s;
@@ -185,7 +186,7 @@ const char *wbu_system_webots_tmp_path() {
             continue;
           if (!S_ISDIR(s.st_mode))
             continue;
-          if (strlen(entry->d_name) > 64)
+          if (strlen(entry->d_name) > 70)
             continue;
           sscanf(entry->d_name, "webots-%d%63s", &webots_pid, random_part);
           break;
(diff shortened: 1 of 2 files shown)


Simon-Steinmann (approved):
Tested the fix, and it solves the issue. Robot can be initialized directly after starting a webots process

stefaniapedrazzi:
Could you please add more details about the bug?
This will both help for the review and for history logs.

omichel:
@Simon-Steinmann wrote (on Discord):

> I have an issue. When setting a webots PID variable, webots can't find it. The issue is, that it tries to do this:
> Cannot open file: /tmp/webots-20244/WEBOTS_SERVER (retrying in 1 second)
>
> while the tmp files generated, have an extra string:
>
> (I know it's not the same PID, i closed the process already. but with the correct PID, the issue is there as well)

src/Controller/api/system.c
```diff
@@ -185,7 +186,7 @@ const char *wbu_system_webots_tmp_path() {
             continue;
           if (!S_ISDIR(s.st_mode))
             continue;
-          if (strlen(entry->d_name) > 64)
+          if (strlen(entry->d_name) > 70)
```
stefaniapedrazzi, line 189: Does the same fix also apply to line 165?

src/Controller/api/system.c
```diff
@@ -177,6 +177,7 @@ const char *wbu_system_webots_tmp_path() {
       struct dirent *entry;
       char folder_start[32];
       sprintf(folder_start, "webots-%d", webots_pid);
+      webots_pid = 0;
```
stefaniapedrazzi, line 180: it would probably good to add comment here, for example (not sure if this is really correct):
```suggestion
      // reset webots_pid containing the webots script process id to before retrieving the webots binary id
      webots_pid = 0;
```

stefaniapedrazzi (commented):
Thank you for adding a description of the bug.
The fix is fine.
I just have a couple of comments/questions.

stefaniapedrazzi (approved):
Thank you

PR #2265, opened 2020-09-23
Fixed Ctrl key keyboard interface on macOS

omichel:
A user reported a bug on macOS: when pressing the <kbd>Ctrl</kbd> key in the 3D view together with another key, it was not taken into account by the keyboard API. This PR fixes that bug.

### DIFF
--- a/docs/reference/changelog-r2020.md
+++ b/docs/reference/changelog-r2020.md
@@ -4,6 +4,7 @@
 Released on XXX YYY, 2020.
 
   - Bug fixes
+    - Fixed handling of <kbd>Ctrl</kbd> key on mac OS from the keyboard API ([#2265](https://github.com/cyberbotics/webots/pull/2265)).
     - Fixed synchronization bug in the start up of extern controllers when the `WEBOTS_PID` environment variable was defined ([#2260](https://github.com/cyberbotics/webots/pull/2260)).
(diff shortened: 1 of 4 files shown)


omichel:
There were actually several problems:
1. The <kbd>Ctrl</kbd> and <kbd>Meta</kbd> keys are [swapped](https://doc.qt.io/qt-5/macos-issues.html#special-keys) on mac OS.
2. The `event->text()` value appeared to be an empty string in some cases, like <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>1</kbd> on mac OS.
3. The latest keyboard change was appended to the list of changes whereas it seems better to prepend it as only the first value is read by the controller.

stefaniapedrazzi:
:eyes:

docs/reference/changelog-r2020.md
```diff
@@ -4,6 +4,7 @@
 Released on XXX YYY, 2020.
 
   - Bug fixes
+    - Fixed handling of <kbd>Ctrl</kbd> key on mac OS from the keyboard API ([#2265](https://github.com/cyberbotics/webots/pull/2265)).
```
stefaniapedrazzi, line 7: minor suggestions to be more consistent with the rest of the changelog entries
```suggestion
    - Fixed handling of <kbd>Ctrl</kbd> key on macOS from the [Keyboard](keyboard.md) API ([#2265](https://github.com/cyberbotics/webots/pull/2265)).
```

stefaniapedrazzi (approved):
The fix is fine.
I would just wait the result of the tests before merging.

PR #2266, opened 2020-09-24
Compute Animation step update after Solid coordinates have been updated

### LINKED ISSUE #2234: Mismatch between local simulation and recorded html animation
**Describe the Bug**
Recorded html5 animation is different from the simulation. Object position lags behind moving gripper.

**Steps to Reproduce**
I created a html5 recording of a simulation but the animation is different from the local simulation. 

**Expected behavior**
In local simulation, everything looks normal. However, in the animation replay, when the gripper moves objects, the object's position lags behind the gripper's position. 

**Screenshots**

**System**
 - Graphics Card: NVIDIA GeForce RTX 2080 11 GB

**Additional context**
I do have a lot of objects in the environment, that might be the issue since I had no problem with a different environment with less objects. I'd like to know if there is an easy fix (possibly from the html end) for this particular environment. 

Thanks

stefaniapedrazzi:
**Description**
Fixes #2234: the animation recording system publishes the changes too early, before the Solid nodes translation and rotation field values are correctly updated.

This issue alone won't create so big position/rotation mismatches if it was not for the `WbSimulationCluster::step` that updates the Solid coordinates earlier (in the middle of the physics step) in order to update the DistanceSensor rays.
This should also be inspected, because the translation values computed when updating the rays are different from the ones at the end of the step and could lead to wrong detection.
Note that these intermediate coordinate values are not simply "one step old" values.

**Tasks**
* [x] store Animation step update after the Solid coordinates have been updated
* [x] check that all the other methods related to the `physicsStepEnd` signal still work even if called after updating the coordinates
* [x] check Solid coordinates values when updating of ODE rays

### DIFF
--- a/docs/reference/changelog-r2020.md
+++ b/docs/reference/changelog-r2020.md
@@ -9,6 +9,8 @@ Released on XXX YYY, 2020.
     - Fixed [Lidar](lidar.md) and [RangeFinder](rangefinder.md) memory leak when the robot-window is opened ([#2210](https://github.com/cyberbotics/webots/pull/2210)).
     - Fixed noise generation for [Camera](camera.md), [Lidar](lidar.md) and [RangeFinder](rangefinder.md) producing fixed patterns on some GPUs (like the NVIDIA GeForce RTX series)([#2215](https://github.com/cyberbotics/webots/pull/2215)).
     - Fixed re-initialization of external camera window if recognition is enabled ([#2196](https://github.com/cyberbotics/webots/pull/2196)).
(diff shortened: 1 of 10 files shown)


stefaniapedrazzi:
The problem with the update of the Transform coordinates to setup rays is that they were computed top-down (from current node to the root) instead of bottom-up as required given that a node position/orientation depends on the parent position/rotations.
Fixing this, the values are just one step old as expected.

omichel (approved):
Thank you.

PR #2268, opened 2020-09-25
Fixed movie recording when WorldInfo.basicTimeStep was greater than 40

omichel:
This bug was reported by a Discord user:

> I'm trying to record video and I'm having  an issue. The video acceleration won't let me change it from inf. I then get an error when it tries to save. Unable to parse option value "inf" as video rate.

This actually happens if the `WorldInfo.basicTimeStep` value is greater than 40.


stefaniapedrazzi:
:eyes:

stefaniapedrazzi (approved):

PR #2271, opened 2020-09-28
Enhancement Docker installation instructions

### LINKED ISSUE #2254: Ubuntu 16.04: issues with nodejs during compilation
When compiling Webots `resources/web/wwi` on Ubuntu 16.04 docker image following the instruction on the wiki page (https://github.com/cyberbotics/webots/wiki/Docker#create-your-own-image) the compilation fails because of nodejs issues.

On Ubuntu 18.04 docker image it works correctly.

We should fix it or explicitly mention that only Ubuntu 18.04 and Ubuntu 16.04 are supported.

stefaniapedrazzi:
**Description**
Fix #2254: instructions to compile Webots on Docker container doesn't work on Ubuntu 16.04

**Tasks**
  - [x] Fix instructions in wiki page: https://github.com/cyberbotics/webots/wiki/Docker
  - [x] Adjust installation scripts so that they work on a Docker container
  - [x] Fix instructions to run the streaming server in a Docker container:
    * https://www.cyberbotics.com/doc/guide/web-simulation?version=enhancement-docker-installation#using-docker
    * https://github.com/cyberbotics/webots/wiki/Docker#run-streaming-server-in-docker-container

### DIFF
--- a/docs/guide/web-simulation.md
+++ b/docs/guide/web-simulation.md
@@ -311,18 +311,70 @@ In this case, you can open automatically a user session when the computer is swi
 If you have a headless system, i.e., a system without any physical monitors attached, then with the NVIDIA graphics card you could fake a monitor in the X session.
 This solution basically consists in adding a screen configuration to the X server configuration file by copying the Extended Display Identification Data (EDID) of a temporary attached monitor.
 
-#### Startup Procedure
+#### Using Docker
 
-The startup procedure is the following:
+The simulation server can also be run in a Docker container.
+In this section we provide a sample Docker setup that could be used to run the simulation on localhost.
 
-1. Start all the simulation servers (`simulation_server.py`).
-2. Start the session server (`session_server.py`).
+You can use the following `Dockerfile` to build your Docker image.
+```
+FROM cyberbotics/webots:R2021a-ubuntu18.04
 
-This procedure should be automated in a startup script, so that the servers are restarted after a reboot of the machine.
+RUN apt update
+RUN apt install -y nano psmisc firejail python3-pip
+RUN pip3 install websocket-client tornado nvidia-ml-py psutil requests distro
+RUN pip3 install -U pynvml --user
+ENV AUDIODEV=null
+ENV DEBIAN_FRONTEND=noninteractive
+ENV DISPLAY=:99
+ENV LIBGL_ALWAYS_SOFTWARE=true
+ENV WEBOTS_HOME=/usr/local/webots
+ENV QTWEBENGINE_DISABLE_SANDBOX=1
+RUN echo 'alias python=python3' >> ~/.bashrc
+COPY server/config /usr/local/webots/resources/web/server/config
+COPY server/key /usr/local/webots/resources/web/server/key
+COPY server.sh /usr/local/server.sh
+RUN chmod 654 /usr/local/server.sh
+CMD ["/usr/local/server.sh", ""]
+```
 
-This folder also contains a `server.sh` utility script to automatically start and stop `session_server.py` and `simulation_server.py` with a given configuration file.
+To correctly setup and automatically run the simulation server, you should provide the following files:
+* `server/config`: a folder containing the simulation and session configuration files.
+    The current example `server.sh` the local configuration by default, i.e. you should provide the `server/config/session/local.json` and `server/config/simulation/local.json` files or specify the configuration files to be used.
+    For example, `server/config/session/localhost.js`
+    ```
+    {
+      "port": 1999,
+      "server": "localhost",
+      "simulationServers": [
+        "localhost:2000"
+      ]
+    }
+    ```
+    and `server/config/simulation/localhost.js`
+    ```
+    {
+      "port": 2000
+    }
+    ```
+* `server/key`: a folder containing your website host keys needed for validation (see [Session server](#session-server) section).
+* `server.sh`: a script that configures the virtual screen and starts the simulation and session servers.
+    ```bash
+    #!/bin/sh
+    Xvfb :99 -screen 0 1024x768x16 &
+    cd $WEBOTS_HOME/resources/web/server
+    ./server.sh start
+    while true; do sleep 1000; done
+    ```
 
-Please make sure that the `WEBOTS_HOME` variable is set before running the simulation and session server scripts.
+Then, you can open in a terminal the directory containing the Dockerfile to build and run the Docker container:
+```bash
+docker build -t webots-simulation-server .
+sudo docker run -p 1999-2100:1999-2100 -it webots-simulation-server
+```
+
+This example runs the simulation server on localhost.
+If you want to publish the simulation server on the web, then you may need to setup a web server, such as the Apache web server.
 
 ### Website Host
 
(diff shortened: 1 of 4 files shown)


stefaniapedrazzi:
Some of the libraries that were previously installed when following the instructions in the wiki/Docker page has been removed (and are not included in any of the linux installation scripts):
* libjpeg8-dev
* libgd-dev
* libtiff5-dev
* libssh-dev

stefaniapedrazzi:
Instructions on the User Guide can then be updated in the `develop` branch to use the streaming viewer page to start the simulation.

docs/guide/web-simulation.md
```diff
@@ -324,6 +324,71 @@ This folder also contains a `server.sh` utility script to automatically start an
...
+COPY server/key /usr/local/webots/resources/web/server/key
+COPY server.sh /usr/local/server.sh
+RUN chmod 654 /usr/local/server.sh
+CMD ["/usr/local/server.sh", ""]
+```
+
+To correctly setup and automatically run the simulation server, you should provide the following files:
+* `server/config`: a folder containing the simulation and session configuration files.
+    The current example `server.sh` the local configuration by default, i.e. you should provide the `server/config/session/local.json` and `server/config/simulation/local.json` files or specify the configuration files to be used.
+    For example, `server/config/session/localhost.js`
```
omichel, line 357:
```suggestion
    For example, `server/config/session/localhost.json`
```

docs/guide/web-simulation.md
```diff
@@ -324,6 +324,71 @@ This folder also contains a `server.sh` utility script to automatically start an
...
+    ```
+    {
+      "port": 1999,
+      "server": "localhost",
+      "simulationServers": [
+        "localhost:2000"
+      ]
+    }
+    ```
+    and `server/config/simulation/localhost.js`
```
omichel, line 367:
```suggestion
    and `server/config/simulation/localhost.json`
```

docs/guide/web-simulation.md
```diff
@@ -324,6 +324,71 @@ This folder also contains a `server.sh` utility script to automatically start an
...
+FROM cyberbotics/webots:R2021a-ubuntu18.04
+
+RUN apt update
+RUN apt install -y nano psmisc firejail python3-pip
+RUN pip3 install websocket-client tornado nvidia-ml-py psutil requests distro
+RUN pip3 install -U pynvml --user
+ENV AUDIODEV=null
+ENV DEBIAN_FRONTEND=noninteractive
+ENV DISPLAY=:99
+ENV LIBGL_ALWAYS_SOFTWARE=true
```
omichel, line 343: Will that disable 3D acceleration?

docs/guide/web-simulation.md
```diff
@@ -324,6 +324,71 @@ This folder also contains a `server.sh` utility script to automatically start an
...
+    }
+    ```
+* `server/key`: a folder containing your website host keys needed for validation (see [Session server](#session-server) section).
+* `server.sh`: a script that configures the virtual screen and starts the simulation and session servers.
+    ```bash
+    #!/bin/sh
+    Xvfb :99 -screen 0 1024x768x16 &
+    cd $WEBOTS_HOME/resources/web/server
+    ./server.sh start
+    while true; do sleep 1000; done
```
omichel, line 380: That seems odd... What is the purpose of that? Not to kill the server.sh script by exiting? Couldn't we do something like this instead?
```suggestion
    nohup ./server.sh start
```

omichel (changes requested):

docs/guide/web-simulation.md
stefaniapedrazzi: I will check.
I copied this from the already existing documentation.

docs/guide/web-simulation.md
stefaniapedrazzi: The purpose is to keep the Docker container running otherwise it will return and stop immediately after starting the server script. 
Not sure that nohup will work. 

There is another alternatively that is to add a run option to run directly the container in background.
But it has the drawback that requires more steps to stop it so I personally prefer this one for testing (in any case it is not a system that could be used as-is in production). 

I will improve it on Monday.

stefaniapedrazzi: Docker needs a foreground running process to continue running, otherwise it stops immediately.
That is why `nohup` doesn't work and even running the container in background doesn't make any difference.
We could avoid this loop by using a custom `server.sh` script that doesn't run the session and/or simulation server in background.

Checking again the documentation of Docker, it seems that is designed to run a single server the container, so another possibility is to run one container for the session server and one for the simulation server.

omichel: It looks like it is overkill to run one Docker just for the session server.

docs/guide/web-simulation.md
stefaniapedrazzi: Yes, this disables hardware acceleration.
But I checked and it is no longer needed.

omichel (approved):
Thank you.

docs/guide/web-simulation.md
```diff
@@ -324,6 +324,68 @@ This folder also contains a `server.sh` utility script to automatically start an
...
+FROM cyberbotics/webots:R2021a-ubuntu18.04
+
+RUN apt update
+RUN apt install -y firejail python3-pip
+RUN pip3 install tornado pynvml psutil requests distro
+ENV AUDIODEV=null
+ENV DEBIAN_FRONTEND=noninteractive
+ENV DISPLAY=:99
+ENV WEBOTS_HOME=/usr/local/webots
+ENV QTWEBENGINE_DISABLE_SANDBOX=1
```
DavidMansolino, line 343: Isn't this variable inherited from the parent container image?

docs/guide/web-simulation.md
```diff
@@ -324,6 +324,68 @@ This folder also contains a `server.sh` utility script to automatically start an
...
 Please make sure that the `WEBOTS_HOME` variable is set before running the simulation and session server scripts.
 
+#### Using Docker
+
+The simulation server can also be run in a Docker container.
+In this section we provide a sample Docker setup that could be used to run the simulation on localhost.
+
+You can use the following `Dockerfile` to build your Docker image.
+```
+FROM cyberbotics/webots:R2021a-ubuntu18.04
```
DavidMansolino, line 334: This image is not available yet as Webots R2021a is not yet released.
To avoid having to maintain this we can use instead `FROM cyberbotics/webots:latest`

docs/guide/web-simulation.md
stefaniapedrazzi: Probably yes. I will check and remove all the unneeded variables redefinition.

docs/guide/web-simulation.md
stefaniapedrazzi: Which is the Ubuntu version of `cyberbotics/webots:latest`?

DavidMansolino: Oh good point, it is 20.04, then we have to use `R2020b-rev1-ubuntu18.04` and maintain it :-/

DavidMansolino: Or even better:
```suggestion
FROM cyberbotics/webots:{{ webots.version.package }}-ubuntu18.04
```

If it works in code parts.

Issue #2269, opened 2020-09-25
API for fast simulation data retrieval


Simon-Steinmann:
I did some deeper investigation and found out an issue, when trying to retrieve simulation data through the supervisor api.  I ran a c-profler on my webots controller, and I noticed that getting position data from object in the scene took very long. Here is a list of the times per call:

node.getPosition() = 23 851 ns
node.getOrientation() = 21 896 ns
Field.getSFVec3f() = 22 563 ns

If we compare that to a sensor:

PositionSensor.getValue() = 450 ns

This really becomes and issue, when needing to extract data, such as object position and orientation, or anything else from the environment. As an example, I have a reinforcement learning environment it Webots, where I have to retrieve the position and orientation of 25 objects. That is 50 calls. This takes more than 1ms, which is longer than the simulation takes for each step. 

**What is needed:**
A fast method, to retrieve all needed data.

**Ideas for API addition:**
Olivier wrote on discord:
> To speed-up things, we could implement the same mechanism as with sensors where we subscribe to a position/orientation service and get faster batched results.

I like this idea, however, perhaps it could be expanded. Perhaps useers could create a custom data service, and fill them with any Field of the scene tree. This could look something like this:
```python
timestep = supervisor.getBasicTimeStep()

field_1 = supervisor.getFromDef('box_1').getField('translation')
field_2 = supervisor.getFromDef('box_1').getField('rotation')
...
field_n = supervisor.getFromDef('box_n').getField('rotation')

dataService = supervisor.dataService.create([field_1, field_2, ..., field_n])
dataService.enable(samplingPeriod = timestep)

data = dataService.getValue()
```

Perhaps this could be even expanded to set fields in the simulation as well. Just like motors or other devices. Being able to use a single command to send a vector of commands, could be very beneficial and speed things up.

omichel:
Thank you for this nice proposal.

I have some concerns about returning a non-heterogeneous data structure (which may be difficult to handle depending on the programming language such as C/C++/Java and MATLAB). So I believe that *dataService.getValue()* should simply return an single dimension array of floating point values (double type in C) where X, Y, Z or alpha values would be added according to the order specified in the constructor.

Also from a naming perspective, *dataService* sounds too generic to me. I would replace it with *tracker*.
And why not using a normal constructor instead of a *create* method, like:
```python
tracker = supervisor.tracker([field_1, field_2, field_3])
```
[...]

PR #2279, opened 2020-09-30
Optimized supervisor field tracking

Simon-Steinmann:
**Description**
The goal is to implement a supervisor api, which allows to enable tracking for fields, so that their values can be retrieved much faster. Similar to how sensors work.

**Related Issues**
https://github.com/cyberbotics/webots/issues/2269


Simon-Steinmann:
Okay, so far I did the following:
I took the enable function from here 
https://github.com/cyberbotics/webots/blob/master/src/Controller/api/position_sensor.c#L118
and implemented it in the supervisor.c here:
https://github.com/cyberbotics/webots/pull/2279/files#diff-2b519526ed3669863e9d441f3bbd79b8R2252

I have two questions:
1. Do I get and use a pointer, like the position sensor in [line:125](https://github.com/cyberbotics/webots/blob/master/src/Controller/api/position_sensor.c#L125), or do I do it differently. 

2. Which of these typedef structs do I add the enable bool and sampling period? Is it [line:56](https://github.com/cyberbotics/webots/blob/790420da05cdb0d91b2110545946cfcd176842f2/src/Controller/api/supervisor.c#L56)

omichel:
It seems that your code doesn't yet compile (as the `tag` symbol doesn't seem to be defined).
You should not use a position sensor (this is device that has nothing to do with the supervisor).
Instead, you should probably create a new type of field request in https://github.com/cyberbotics/webots/blob/develop/src/Controller/api/supervisor.c#L32 that you may call `TRACK`. You should also add the refresh rate value in the WbFieldRequest structure (which will be 0 if disabled). Then, you should handle it (like other types of requests are handled to communicate it to Webots).

Justin-Fisher:
Webots API already offers a reasonably good interface for controllers to access field values.  The only problem is that the current under-the-hood implementation for accessing these values is too slow.  It seems a shame to clutter the API with a second form of accessing field values (via subscription).  Instead, **I would encourage you to just improve the under-the-hood implementation so that it goes faster with the current controller interface.  In particular, I would encourage you to keep track of which fields the controller had to send slow requests through the pipeline to get, and then _automatically_ "subscribe" to those fields for subsequent timesteps**, automatically making a copy of that field value be available on the controller side of the pipeline.  

**"Automatic subscription" would have multiple benefits.**  It would **automatically speed up existing code** without need for anyone to go in and add any subscribe commands.  It would also **keep things simpler for coders** rather than making them learn two different ways of accessing fields and the relative advantages of each.  This simpler interface simpler will also **leave Webots with more freedom as it goes forwards to potentially re-implement how field access works** without having to retain backwards-compatibility with a second form of field access.

Here are a few questions that would arise if you went the automated route that I'm suggesting:

1. **How many slow "through the pipe" requests does it take for you to automatically decide to "subscribe"?**  I'd be inclined to say 2, to allow that a controller might want to scan lots of aspects of the world once, but anything it takes an interest in a second time seems like something to give it quick access to.  Another plausible answer is 1, but don't automatically subscribe to anything until after the controller has had a step or two to initialize itself.  

2. **Will you ever automatically "unsubscribe" a field that it turns out the controller doesn't seem to use?**  My inclination is to say that, relative to all the other things that consume processor time and memory, the costs of "oversubscription" would be negligible, so I wouldn't worry about unsubscribing at this point.  
 
3. **Should you offer an API for controllers to pre-declare that they want to subscribe/unsubscribe to certain fields?**  Meh...  I'd say focus on getting automatic detection to work reasonably well.  After that, there probably will be very little incremental benefit of subscribing a timestep or two quicker, or of unsubscribing to save a few C copy operations.  So I'd say probably don't clutter the controller interface with such commands for now, or probably ever.

omichel:
@Justin-Fisher: this is indeed very interesting. However, to speed-up things, we have to implement this one shot communication system based on subscriptions that will be used behind the scene in any case. Then, we could decide to use some form of automatic subscription (as you propose) that wouldn't need any new API function or to implement some new API functions as initially envisionned. I believe that new API functions should be implemented only if they provide a significantly better performance or a better usability. But this can be decided later on, when we have the fast implementation working.

Simon-Steinmann:
I really like this idea too!

Simon-Steinmann:
Currently I have no time for working on this. I would pass it back onto the cyberbotics team. I still think it is an essential feature and should be implemented soon, but I simply don't have the knowledge and skills to do this in a timely manner. The branch can be deleted, as nothing significant was done. But perhaps the discussions so far are insightful. 

For testing I will always be available.

lukicdarkoo:
I would like to resurrect this PR, ideally making it available in R2021b.

I afraid of using the `Automatic subscription` feature as we cannot know how often the control loop is running (it might be significantly slower than the physics step frequency). If you agree I will continue based on the API started by Simon.

lukicdarkoo:
**TODO**:

- Field tracking
  - [x] Basic implementation
    - [x] C++
    - [x] Python
    - [x] Java
    - [ ] Matlab
    - [ ] ROS
  - [ ] Documentation
  - [ ] Tests
  - [ ] ROS tests
  - [ ] Controller.def + WbLanguage.cpp
- Field tracking (`getPosition` and `getOrientation` will not be covered, only `getPose`)
  - [ ] Basic implementation
    - [ ] C++
    - [ ] Python
    - [ ] Java
    - [ ] Matlab
    - [ ] ROS
  - [ ] Documentation
  - [ ] Tests
  - [ ] ROS tests
  - [ ] Controller.def + WbLanguage.cpp
- [ ] Changelog    


**Notes**:

```
#VRML_SIM R2021b utf8
WorldInfo {
  coordinateSystem "NUE"
}
Viewpoint {
  orientation 0.0626809120566472 0.9749704283213647 0.2133161202595843 3.699618222299445
  position -1.2991653627436153 1.7760140946848124 -2.6061943463293913
}
TexturedBackground {
}
TexturedBackgroundLight {
}
Floor {
}
DEF BOX WoodenBox {
}
Robot {
  controller "fieldtracking"
  supervisor TRUE
}
```

```py

robot = Supervisor()

field = robot.getFromDef('BOX').getField('translation')
field2 = robot.getFromDef('BOX').getField('rotation')
# field2.enableTracking(timestep)
# field.enableTracking(timestep)
robot.step(timestep)

    field.getSFVec3f()
    field2.getSFRotation()
```

@omichel, I finished a very basic implementation (possible bugs), so you can test RoboCop performance improvements. In my tests, with a single field performance improvements is around 2x (I assume more fields == higher performance improvements).

lukicdarkoo:
What do we do with multi fields, should we support those? If we choose to support multi fields, we need to have something like `wb_supervisor_field_enable_mf_tracking` and `wb_supervisor_field_enable_mf_tracking`. I cannot think of a use-case in which we have to track a multi-field, @Simon-Steinmann @omichel?

omichel:
> What do we do with multi fields, should we support those? If we choose to support multi fields, we need to have something like `wb_supervisor_field_enable_sf_tracking` and `wb_supervisor_field_enable_mf_tracking`. I cannot think of a use-case in which we have to track a multi-field, @Simon-Steinmann @omichel?

I believe there are not many use cases for `wb_supervisor_field_enable_mf_tracking`, but I would keep the possibility to implement it later if needed, so I would suggest to rename `wb_supervisor_field_enable_tracking` to `wb_supervisor_field_enable_sf_tracking` to make it scalable.

omichel:
I just tested and the previous `Node.getField` or `Node.getProtoField` functions appear to be broken.

omichel:
I tested it and it seems to work great. However, for the robocup referee application, it seems the bottleneck is more in the retrieval of contact points, so, it would be nice to extend this PR to tracking contact points as well.

omichel:
Also, I believe we should improve the `getPose` documentation, e.g., give an example on how to retrieve the X, Y, Z coordinates from the return value. I assumed:
```python
pose = node.getPose()
x = pose[3]
y = pose[7]
z = pose[11]
```
Is that correct? Or is it `x = pose[3] / pose[15]`, etc.

omichel:
Also, I believe that [target_coordinates.c](https://github.com/cyberbotics/webots/blob/develop/projects/robots/neuronics/ipr/controllers/target_coordinates/target_coordinates.c) example should be properly referenced by the documentation page and updated to use the new `wb_supervisor_node_get_pose` function instead of `wb_supervisor_node_get_position` and `wb_supervisor_node_get_orientation`.

docs/reference/changelog-r2021.md
```diff
@@ -13,6 +13,7 @@ Released on June, Xth, 2021.
     - Changed the rendering engine of the streaming viewer and of the animations for WREN ([#2769](https://github.com/cyberbotics/webots/pull/2769)).
     - Added the `redColorSensitivity` field to [DistanceSensor](distancesensor.md) that allows tuning (or even disabling) of the red color sensitivity for an infra-red distance sensor ([#3077](https://github.com/cyberbotics/webots/pull/3077)).
   - Enhancements
+    - Added pose and field tracking functions to improve performances of getting pose and field data ([#2279](https://github.com/cyberbotics/webots/pull/2279)).
```
omichel, line 16: Performance is always singular in English.
```suggestion
    - Added pose and field tracking functions to improve the performance of getting pose and field data ([#2279](https://github.com/cyberbotics/webots/pull/2279)).
```

docs/reference/supervisor.md
```diff
@@ -881,6 +895,12 @@ The `wb_supervisor_node_get_pose` function returns an array of 16 values.
 The array shall be interpreted as a 4 x 4 [transformation matrix](https://en.wikipedia.org/wiki/Transformation_matrix) that represents an absolute transform of the node.
 The function returns the relative pose of the node with respect to the node specified in `from_node`. If `from_node` is null, it returns the absolute pose of the node in the global coordinate system.
 
+The `wb_supervisor_node_enable_pose_tracking` function forces Webots to stream poses to the controller.
+It improves performances as the controller by default uses a request-response pattern to get pose data.
```
omichel, line 899:
```suggestion
It improves the performance as the controller by default uses a request-response pattern to get pose data.
```

docs/reference/supervisor.md
```diff
@@ -3045,6 +3079,14 @@ The `wb_supervisor_field_get_count` function returns the number of items of a mu
 If a single field (SF) or NULL is passed as an argument to this function, it returns -1.
 Hence, this function can also be used to test if a field is MF (like `WB_MF_INT32`) or SF (like `WB_SF_BOOL`).
 
+The `wb_supervisor_field_enable_sf_tracking` function forces Webots to stream field data to the controller.
+It improves performances as the controller by default uses a request-response pattern to get data from the field.
```
omichel, line 3083:
```suggestion
It improves the performance as the controller by default uses a request-response pattern to get data from the field.
```

lib/controller/matlab/wb_supervisor_field_disable_sf_tracking.m
```diff
@@ -0,0 +1,6 @@
+function result = wb_supervisor_field_disable_sf_tracking(field)
+% Usage: wb_supervisor_field_disable_sf_tracking(field)
+% Matlab API for Webots
+% Online documentation is available <a href="https://www.cyberbotics.com/doc/reference/supervisor">here</a>
+
+result = calllib('libController', 'wb_supervisor_field_disable_sf_tracking', field);
```
omichel, line 6: Since these matlab API functions should be automatically generated by the new `mgenerate.py` script, they should not be committed here, but ignored (`.gitignore`). See https://github.com/cyberbotics/webots/pull/3114. I believe the script has an option to update the local `.gitignore` file.

projects/robots/neuronics/ipr/controllers/target_coordinates/target_coordinates.c
```diff
@@ -70,29 +70,35 @@ int main() {
...
-    const double *m3x3 = wb_supervisor_node_get_orientation(gripper);
+    // transformation matrices are composable as is
+    // however, we extract rotation and translation for demonstration purposes
+    double gripper_rotation[9] = {gripper_pose[0], gripper_pose[1], gripper_pose[2], gripper_pose[4], gripper_pose[5],
+                                  gripper_pose[6], gripper_pose[8], gripper_pose[9], gripper_pose[10]};
+    double gripper_translation[3] = {gripper_pose[3], gripper_pose[7], gripper_pose[11]};
+
+    // get current target pose in world coordinates
+    const double *target_pose = wb_supervisor_node_get_pose(target, NULL);
+    double target_traslation[3] = {target_pose[3], target_pose[7], target_pose[11]};
```
omichel, line 84:
```suggestion
    double target_translation[3] = {target_pose[3], target_pose[7], target_pose[11]};
```

omichel (changes requested):
I also tested it and it looks good to me.

lib/controller/matlab/wb_supervisor_field_disable_sf_tracking.m
lukicdarkoo: Maybe we should exclude all files in the folder except the ones which need custom implementation:
```
*.m
!wb_camera_get_image.m
```

omichel: @BenjaminDeleze: do you think it would be better to do so?

docs/reference/supervisor.md
```diff
@@ -880,6 +894,23 @@ Where *p* is a point whose coordinates are given with respect to the local coord
...
+Where `M[i]` is the result of the `wb_supervisor_node_get_pose` function at index `i`.
+The matrix is composed of a rotation matrix `R` and a translation vector `T`.
+
+The `wb_supervisor_node_enable_pose_tracking` function forces Webots to stream poses to the controller.
+It improves the performance as the controller by default uses a request-response pattern to get pose data.
+The `sampling_period` argument determines how often the pose data should be sent to the controller.
+
+The `wb_supervisor_node_disable_pose_tracking` function disables pose data tracking.
 
 The "[WEBOTS\_HOME/projects/robots/neuronics/ipr/worlds/ipr\_cube.wbt]({{ url.github_tree }}/projects/robots/neuronics/ipr/worlds/ipr_cube.wbt)" simulation shows how to use these functions to achieve this.
```
omichel, line 915: Because the code of this example is not easy to find, I would also link to `{{ url.github_tree }}/projects/robots/neuronics/ipr/controllers/target_coordinates/target_coordinates.c` explicitly.

omichel (changes requested):
Thank you.
Just one minor improvement of the doc could be nice.

lib/controller/matlab/wb_supervisor_field_disable_sf_tracking.m
BenjaminDeleze: It would definitely reduce the size of the .gitignore (from ~500 lines to ~100) but I am not sure that it is possible to automate the maintenance of such a .gitignore

omichel (approved):
Thanks.

Issue #2236, opened 2020-09-12
UR10e collision bug


Simon-Steinmann:
There is a collision bug, when using self collision. The objects do not touch where it shows the collision points. I created a world showing the example. 
[ur10e_collisionbugwbt.zip](https://github.com/cyberbotics/webots/files/5212731/ur10e_collisionbugwbt.zip)

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/52213164/92996535-a8492500-f50c-11ea-929b-689a409ad5ca.gif)

DavidMansolino:
There is indeed something wrong going out with this simulation, I can reproduce the issue too:

![ur10e_collisionbugwbt](https://user-images.githubusercontent.com/2461619/93050565-61ba0e80-f663-11ea-93b4-0acdbccc19ab.png)

DavidMansolino:
I tried simplifying the bounding objects to keep only the two 'colliding' cylinders (to make sure it was not a collision with a 'hidden' bounding object), but the problem is still here:
![ur10e_collisionbugwbt_3](https://user-images.githubusercontent.com/2461619/93052211-f9b8f780-f665-11ea-9431-83d0cd59bf3e.png)

westinsykes:
As a note, real UR arms treat that pose similarly to a collision in order to prevent people's fingers from getting pinched.

omichel:
Fixed in #2282.

PR #2282, opened 2020-09-30
Fix cylinder-to-cylinder collision detection

lukicdarkoo:
**Description**
It seems there is a bug in the physics engine which makes Webots doesn't properly detect collisions between two cylinders. 

**Related Issues**
This pull-request fixes issue #2236

**Additional context**
In tests, there seems to be a similar problem with cylinder-to-capsule collision detection.

**Tasks**
- [x] Try to sync `libccd`  (collision detection library)
- [x] Check for side-effects (regression) - **I am not sure how to properly test this, I did some basic tests and ran the test suit**
- [x] Add to changelog

<details>
<summary>Notes</summary>

Command:
```bash
rm src/ode/libode.so lib/webots/libode.so; make -j16
```

</details>


lukicdarkoo:
Peek 2020-10-02 10-47
It seems to be working now!

omichel (approved):
Thank you.
I assume that in addition to fixing this bug, we reduced the diff with the current version of ODE, which is good in case we decide to synchronize back with a more recent version of ODE.

Issue #2245, opened 2020-09-16
New mode: realtime without rendering


PymZoR:
**Is your feature request related to a problem? Please describe.**
Running webots headless in a CI environment with xvfb-run triggers some performances issues. As rendering is not required, It would be great to completely disable it (as in `fast` mode).

**Describe the solution you'd like**
A mode without rending (like `fast`) in 1x speed (like `realtime`)

**Describe alternatives you've considered**
Digging the reasons behind the performances drop while using xvfb-run

**Additional context**
Run context :
- Host on CentOS / intel CPU / nvidia GPU
- Gitlab CI runner inside a virtual machine
- Job executed in an ubuntu 20.04 based docker container (inspired by https://github.com/cyberbotics/webots-docker/blob/master/Dockerfile)

omichel:
I believe we should take the opportunity of this feature request to split the `mode` option in two:
Currently `--mode` option can take the following values: `pause`, `realtime`, `run` and `fast`.
My proposal would be to introduce a `--disable3dview` option and remove the `fast` option.
Therefore, to get the previous `--mode=fast` behavior, we would have to combine `--mode=run --disable3dview`.
And the new requested behavior would be obtained with `--mode=realtime --disable3dview`.
Similarly on the GUI, we should remove the fast button and replace it with a checkbox entitled "3D View" that would be checked by default.
What do you think about it?

DavidMansolino:
That looks like a very good idea to me!
Simple and flexible.

lukicdarkoo:
Implemented in #2286

PR #2286, opened 2020-10-01
Add `--no-rendering` flag

lukicdarkoo:
**Description**
This PR adds `--no-rendering` which hides 3D view.

**Related Issues**
Implements #2245

**Tasks**
  - [x] Basic implementation
  - [x] Full implementation
    - [x] Configuration should be read/written to file
    - [x] Configuration should be accessible through `--no-rendering` flag
    - [x] Buttons in play menu
    - [x] Menu bar
      - [x] Simulation > Show/Hide
      - [x] View > Scene Interactions > Disable 3D View
    - [x] <kbd>CTRL</kbd> + <kbd>4</kbd>
    - [x] Supervisor API 
    - [x] Movie recording shortcuts
  - [x] Add GUI icon
    - [x] Improve icon resolution and shape 
  - [x] Implement backward compatibility strategy
    - [x] Command line
    - [ ] Supervisor (this will fail, open for discussion)
    - [x] From file
  - [x] Add to changelog
  - [x] Change documentation

**New behavior**

- Case A
  - before `webots --mode=run`
  - now `webots --mode=fast`
- Case B
  - before `webots --mode=fast`
  - now `webots --mode=fast --no-rendering`
- Case C
  - before you want to use mode `realtime` without 3D view (you cannot)
  - now `webots --mode=realtime --no-rendering`
- Case D
  - before <kbd>CTRL</kbd> + <kbd>4</kbd> to start fast mode
  - now <kbd>CTRL</kbd> + <kbd>3</kbd> to enable new fast mode and <kbd>CTRL</kbd> + <kbd>4</kbd> to toggle 3D view
- Case E
  - before `wb_supervisor_simulation_get_mode()` returns `WB_SUPERVISOR_SIMULATION_MODE_RUN` or `WB_SUPERVISOR_SIMULATION_MODE_FAST`
  - now `wb_supervisor_simulation_get_mode()` returns only `WB_SUPERVISOR_SIMULATION_MODE_FAST`

**Backward incompatible**
- `webots --mode=run`
  - deprecation warning + falling back to `fast` mode
- `WB_SUPERVISOR_SIMULATION_MODE_RUN` from Supervisoir API is deleted


lukicdarkoo:
@cyberbotics/maintainers I would like to hear a high-level review before I make it ready for the actual review. Everything should work except backward compatibility for certain cases.

src/webots/gui/WbGuiApplication.cpp
```diff
@@ -482,13 +490,20 @@ WbSimulationState::Mode WbGuiApplication::startupModeFromPreferences() const {
...
   return WbSimulationState::PAUSE;
 }
 
+bool WbGuiApplication::show3DViewFromPreferences() const {
+  WbPreferences *const prefs = WbPreferences::instance();
+  const QString show3DView(prefs->value("General/show3DView").toString());
+
+  if (show3DView == "false")
+    return false;
+  return true;
```
stefaniapedrazzi, line 504: It seems this could be simplified
```suggestion
  return show3DView != "false"
```

src/webots/nodes/WbConnector.cpp
```diff
@@ -920,7 +920,7 @@ void WbConnector::hasMoved() {
 // look recursively through Solid and notify each Connector
 void WbConnector::solidHasMoved(WbSolid *solid) {
   // when the simulation is running we don's allow changes
-  if (WbSimulationState::instance()->isRunning() || WbSimulationState::instance()->isFast())
+  if (WbSimulationState::instance()->isFast() || !WbSimulationState::instance()->is3dViewShown())
```
stefaniapedrazzi, line 923: The following code doesn't seem to depend on the rendering
```suggestion
  if (WbSimulationState::instance()->isFast()))
```

src/webots/gui/WbView3D.cpp
```diff
@@ -1064,7 +1064,7 @@ void WbView3D::setWorld(WbSimulationWorld *w) {
 
   // Creates the fast mode overlay
   if (!mDisabled3dViewOverlay) {
-    mDisabled3dViewOverlay = new WbWrenFullScreenOverlay("Disabled 3D view", 128, true);
+    mDisabled3dViewOverlay = new WbWrenFullScreenOverlay("Hidden 3D view", 128, true);
```
stefaniapedrazzi, line 1067: This is more consistent with the rest of the UI
```suggestion
    mDisabled3dViewOverlay = new WbWrenFullScreenOverlay("Hidden 3D View", 128, true);
```

stefaniapedrazzi (commented):
The changes seems globally fine to me.

Note that for the icons, the one representing the hidden 3d view will be the first non black&white icon.

omichel:
I have nothing to add to the comments of @stefaniapedrazzi.
I agree we should try to have a black & white icon as the red slash may look a bit aggressive.

src/webots/core/WbAction.hpp
```diff
@@ -72,7 +72,7 @@ namespace WbAction {
     LOCK_VIEWPOINT,
     DISABLE_OBJECT_MOVE,
     DISABLE_FORCE_AND_TORQUE,
-    DISABLE_FAST_MODE,
+    DISABLE_3D_VIEW,
```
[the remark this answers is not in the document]
lukicdarkoo: I wanted to preserve as many features as possible that are related to the FAST mode. With `DISABLE_3D_VIEW` we can forbid users to turn on the 3D view and impact load on a server.

That was my reasoning, but I will be happy to delete it :)

src/webots/gui/WbSimulationView.cpp
```diff
@@ -894,49 +897,61 @@ void WbSimulationView::updatePlayButtons() {
...
       break;
 
     default:  // PAUSE
-      actions << realtime << run << fast;
+      actions << realtime << fast;
       break;
   }
 
+  if (WbSimulationState::instance()->is3dViewShown()) {
+    QIcon icon = QIcon();
```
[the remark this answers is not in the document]
lukicdarkoo: I created `toggle3DView()` method in `WbActionManager`. Let me know if you meant that

resources/translations/wb_de.ts
```diff
@@ -984,11 +984,15 @@ This can be caused by duplicate vertices in your mesh. Try to open your model in
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>Run the simulation. (%1+3)</source>
+        <source>Run the simulation as fast as possible. (%1+3)</source>
```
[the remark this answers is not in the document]
lukicdarkoo: I will note it for the future. For this PR, I guess it is fine to leave as is?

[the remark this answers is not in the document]
DavidMansolino: It is indeed fine (I guess :wink:), but we usually try to avoid it as it generates useless diff and may cause conflicts between master/develop.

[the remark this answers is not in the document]
stefaniapedrazzi: Yes, it is fine to keep them this time.

stefaniapedrazzi:
:eyes:

src/webots/app/WbPerspective.cpp
```diff
@@ -389,8 +389,6 @@ QString WbPerspective::getActionName(WbAction::WbActionKind action) {
       return "objectMoveDisabled";
     case WbAction::DISABLE_FORCE_AND_TORQUE:
       return "forceAndTorqueDisabled";
-    case WbAction::DISABLE_FAST_MODE:
```
stefaniapedrazzi, line 392: The `DISABLE_3D_VIEW` status should be added in the perspective file (as it was the case for the `DISABLE_FAST_MODE` option.

src/webots/gui/WbGuiApplication.cpp
```diff
@@ -482,13 +489,17 @@ WbSimulationState::Mode WbGuiApplication::startupModeFromPreferences() const {
...
-    return WbSimulationState::RUN;
   if (startupMode == "Fast")
     return WbSimulationState::FAST;
   return WbSimulationState::PAUSE;
 }
 
+bool WbGuiApplication::show3DViewFromPreferences() const {
+  WbPreferences *const prefs = WbPreferences::instance();
+  const QString show3DView(prefs->value("General/show3DView").toString());
+  return show3DView != "false";
```
stefaniapedrazzi, line 500: This could be probably simplified:
```suggestion
  return prefs->value("General/show3DView").toString() != "false";
```

src/webots/gui/WbMainWindow.cpp
```diff
@@ -1375,9 +1372,6 @@ void WbMainWindow::updateAfterWorldLoading(bool reloading, bool firstLoad) {
   WbActionManager::instance()
     ->action(WbAction::DISABLE_FORCE_AND_TORQUE)
     ->setChecked(perspective->isUserInteractionDisabled(WbAction::DISABLE_FORCE_AND_TORQUE));
-  WbActionManager::instance()
-    ->action(WbAction::DISABLE_FAST_MODE)
-    ->setChecked(perspective->isUserInteractionDisabled(WbAction::DISABLE_FAST_MODE));
```
stefaniapedrazzi, line 1380: This should be replaced by the `DISABLE_3D_VIEW` code

src/webots/gui/WbSimulationView.cpp
```diff
@@ -555,7 +559,7 @@ void WbSimulationView::startVideoCapture(const QString &fileName, int codec, int
 
 void WbSimulationView::stopVideoCapture(bool canceled) {
   WbVideoRecorder::instance()->stopRecording(canceled);
-  restoreFastModeIfNecessary();
+  restore3dViewIfNecessary();
```
stefaniapedrazzi, line 562: Currently there are for example this function `restore3dViewIfNecessary` and another one `is3DViewShown`.
We should be consistent in the variable and function name and always use `3D` or `3d`.
it would probably be better to always use `3D`.

src/webots/gui/WbSimulationView.cpp
```diff
@@ -582,7 +586,7 @@ void WbSimulationView::stopMovie() {
 }
 
 void WbSimulationView::makeMovie() {
-  if (WbSimulationState::instance()->isFast()) {
+  if (!WbSimulationState::instance()->is3DViewShown()) {
     WbLog::warning(tr("Impossible to record a movie while running the simulation in 'Fast' mode."), true);
```
stefaniapedrazzi, line 590: This warning message has to be updated too

src/webots/gui/WbSimulationView.cpp
```diff
@@ -610,20 +614,20 @@ void WbSimulationView::makeMovie() {
...
 }
 
-void WbSimulationView::restoreFastModeIfNecessary() {
-  if (mNeedRestoreFastMode) {
+void WbSimulationView::restore3dViewIfNecessary() {
+  if (mNeedToHide3dView) {
     mView3D->showFastModeOverlay();
-    WbSimulationState::instance()->setMode(WbSimulationState::FAST);
-    mNeedRestoreFastMode = false;
+    WbSimulationState::instance()->show3DView(true);
```
stefaniapedrazzi, line 629: Shouldn't this code disable 3D rendering instead?
```suggestion
    WbSimulationState::instance()->show3DView(false);
```

src/webots/gui/WbSimulationView.cpp
```diff
@@ -746,24 +750,27 @@ void WbSimulationView::realTime() {
...
-  if (WbSimulationState::instance()->mode() == WbSimulationState::FAST)
-    renderABlackScreen();
+void WbSimulationView::toggle3DView() {
+  if (WbSimulationState::instance()->is3DViewShown())
+    WbSimulationState::instance()->show3DView(false);
   else
+    WbSimulationState::instance()->show3DView(true);
+}
+
+void WbSimulationView::updateFastModeOverlay() {
```
stefaniapedrazzi, line 769: This function should probably be renames as the overlay is no longer a `Fast mode` overlay but a hiddeen 3D view overlay.

src/webots/user_commands/WbActionManager.cpp
```diff
@@ -1091,6 +1085,26 @@ void WbActionManager::enableTextEditActions(bool enabled) {
...
 
+void WbActionManager::toggle3DView() {
+  QAction *toggle3DView = action(WbAction::TOGGLE_3D_VIEW);
+
+  if (WbSimulationState::instance()->is3DViewShown()) {
+    QIcon icon = QIcon();
+    icon.addFile("enabledIcons:show_3d_view.png", QSize(), QIcon::Normal);
+    toggle3DView->setIcon(icon);
+    toggle3DView->setChecked(true);
+    toggle3DView->setStatusTip(tr("Hide 3D view to gain better performance. (%1+4)").arg(mapControlKey()));
```
stefaniapedrazzi, line 1096:
```suggestion
    toggle3DView->setStatusTip(tr("Hide 3D View to gain better performance. (%1+4)").arg(mapControlKey()));
```

src/webots/user_commands/WbActionManager.cpp
```diff
@@ -1091,6 +1085,26 @@ void WbActionManager::enableTextEditActions(bool enabled) {
...
+    toggle3DView->setToolTip(tr("Hide 3D View. (%1+4)").arg(mapControlKey()));
+  } else {
+    QIcon icon = QIcon();
+    icon.addFile("enabledIcons:hide_3d_view.png", QSize(), QIcon::Normal);
+    toggle3DView->setIcon(icon);
+    toggle3DView->setChecked(false);
+    toggle3DView->setStatusTip(tr("Show 3D view to see the simulation. (%1+4)").arg(mapControlKey()));
```
stefaniapedrazzi, line 1103:
```suggestion
    toggle3DView->setStatusTip(tr("Show 3D View to see the simulation. (%1+4)").arg(mapControlKey()));
```

src/webots/gui/WbSimulationView.cpp
```diff
@@ -894,43 +901,36 @@ void WbSimulationView::updatePlayButtons() {
...
+      actions << realtime << pause;
       break;
 
     default:  // PAUSE
-      actions << realtime << run << fast;
+      actions << realtime << fast;
       break;
   }
 
+  manager->toggle3DView();
```
stefaniapedrazzi, line 926: Are there many instructions from this `updatePlayButtons` that are needed in case the 3D rendering is toggled?
Because at first sight it seems more efficient to split the two methods and emit a different signal from the WbSimulationState instance if the 3D rendering  is toggled but the mode is not changed.
I checked just the first slots connected to the `WbSimulationState::modeChanged` signal but it seems that in most cases only the play mode or only the 3D rendering status is needed but not both so two different signals could be better.

stefaniapedrazzi (changes requested):
It works fine but there are still some issues to address.

There are some issues in the simulation menu:

simulation_menu

* the `3D View` checkbox should go at the end of the list
* The `Run` action should be renamed `Fast`
By the way I wouldn't name the action `3D View` because there is already a `3D View` item in the `View` menu. It could be better to name it `3D rendering` or similar.
In this case the code function and variable names should be adapted accordingly.

src/webots/core/WbAction.hpp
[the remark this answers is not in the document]
stefaniapedrazzi: Yes, it is fine to keep it for disable turning on the 3D rendering.
Note that the implementation of this functionality is not complete as this value doesn't seem to be included in the perspective file.

src/webots/gui/WbSimulationView.cpp
lukicdarkoo: You right, I missed changing this one from `3d` to `3D`

src/webots/app/WbPerspective.cpp
lukicdarkoo: This seems not to work correctly (initial state of `DISABLE_FAST_MODE` is not applied) in Webots 2020b Rev 1, maybe we should address that.

See: https://github.com/cyberbotics/webots/pull/2286/commits/ca8f681c6555aeb43cddb42e14796a8100b00644

src/webots/gui/WbSimulationView.cpp
lukicdarkoo: Agree. I added the signal, it works fine, but I only afraid that I forgot to connect it to something.

src/webots/app/WbPerspective.cpp
stefaniapedrazzi: Yes, this should also be addressed on the master branch then.
But we should in any case write and read the new `DISABLE_3D_VIEW` options from the perspective file.
So it seems that the new option should be added here.

src/webots/gui/WbSimulationView.cpp
stefaniapedrazzi: Once the new signal is added I can double check that everything is correctly connected.

lukicdarkoo:
> My comment about the `3D View` menu item has not been addressed.
> I really think that we should re-evaluate the term `3DView` (`disable3dview`, `show3DView`, `hide3DView`) because to me it is now a little bit confusing. For me the 3D View is the Qt widget where the rendering is shown (also see the Tools menu item that already contains a `3D View` item).
> 
> What about using the term `3D rendering` or similar instead of `3D view`?
> and `enable/disable` instead of `show/hide`?

Sorry, you wrote "there is already a `3D View` item in the `View` menu", I didn't see it there, so I thought it was a mistake :)

We can call it `3D rendering` if it is more accurate. I wouldn't use disable/enable instead of hide/show though because it could be mistaken from items from `Scene Interactions`. Maybe Turn on/Turn off?

omichel:
Why not putting this menu item in the *View* menu, just after *Plain Rendering* and *Wireframe Rendering* and call it *Disable Rendering* or something like that?

stefaniapedrazzi:
Using turn on/off rendering is fine for me.
And also to move it in the `View` menu.
If we already use disable for the scene interactions maybe we could simply use `No rendering` for the menu item text and `--no-rendering` or similar for the starting option.

omichel:
Also, it seems to me that this is a bit redundant with the *Tools* menu which also allows the user to hide the 3D view and hence disable de-facto the 3D rendering...

omichel:
I am fine with the *No Rendering* option in the *View* menu.

lukicdarkoo:
So we:
- delete 3D View from the Tools menu, 
- add No Rendering to View menu,
- delete 3D View from Simulation menu,
- delete 3D View icon (???),
- delete <kbd>CTRL</kbd> + <kbd>4</kbd> shortcut (???),
- rename `--disable3dview` to `--no-rendering`

stefaniapedrazzi:
>     * delete 3D View from the Tools menu,
No we should not delete the 3D View from the Tools menu.
This action does something different: it hides the Qt widget containing the rendering but doesn't stop the rendering.
This is part of the layout management where the user can choose which widget to display and you cannot achieve the same with the new "no rendering" action (where the black screen is still displayed).

>     * delete 3D View icon (???),
Do you mean the new one on the toolbar between the play buttons and the speedometer?
This one we want to keep it.
> 
>     * delete CTRL + 4 shortcut (???),
The "No rendering" action still needs to have a shortcut.
Maybe we should change the existing ones for the `Plain rendering` and `Wireframe rendering`

lukicdarkoo:
What about `Scene Rendering` instead of `Rendering` or `No Rendering`? That way it would be more consistent with `Plain Rendering` and `Wireframe Rendering`.

omichel:
In the **View** menu, we currently have a section with a toggle selection between **Plain Rendering F11** and **Wireless Rendering F12**. Adding an item named **Scene Rendering** in the same section or outside this section would be confusing to me. I believe we should add something outside of the *plain/wireframe* toggle section to disable/enable rendering, which should be consistent with the command line option. So far, my best candidates are:

- **Disable Rendering**  `--disable-rendering` (which doesn't pose any problem to me with respect to other disable options)
- **No Rendering** `--no-rendering`

lukicdarkoo:
@omichel and I have just had a discussion and here are some conclusions:
- `View > Rendering` should be added (`Rendering` would be the first option) instead of the current `3D View`
  - if the rendering option is not checked, it disables other options in the menu (`Plain Rendering`, `Wireframe Rendering`, `Optional Rendering`...)
- `--no-rendering` instead of `--disable3dview`
- In separated PR, `Tools > 3D View` should automatically uncheck `View > Rendering`

@stefaniapedrazzi, it is open for suggestions

stefaniapedrazzi:
Yes, this design is good.

lukicdarkoo:
> if the rendering option is not checked, it disables other options in the menu (Plain Rendering, Wireframe Rendering, Optional Rendering...)

I am reconsidering this behavior. A user now can change e.g. from `Plain Rendering` to `Wireframe Rendering` while the rendering is turned off and the `Wireframe Rendering` will be visible once rendering is turned on again. This may be the desired behavior?

DavidMansolino:
> A user now can change e.g. from Plain Rendering to Wireframe Rendering while the rendering is turned off and the Wireframe Rendering will be visible once rendering is turned on again. This may be the desired behavior?

To me, that makes indeed more sense. Changing these options while rendering is disabled will have no immediate impact, but still, it might be useful (and less confusing for the user) and will make the code simpler/cleaner.

omichel:
Yes, it is OK for me as well. I don't have a strong opinion on that.

src/webots/gui/WbSimulationView.cpp
```diff
@@ -128,16 +128,16 @@ WbSimulationView::WbSimulationView(QWidget *parent, const QString &toolBarAlign)
   WbSimulationState *state = WbSimulationState::instance();
 
   //  show a black screen is fast mode is selected
```
stefaniapedrazzi, line 130:
```suggestion
  //  show a black screen if rendering is turned off
```

src/webots/gui/WbSimulationView.cpp
```diff
@@ -232,10 +232,17 @@ QToolBar *WbSimulationView::createToolBar() {
   mToolBar->widgetForAction(action)->setObjectName("invisibleButton");
   mToolBar->widgetForAction(action)->setVisible("false");
 
+  WbActionManager::instance()->updateRenderingButton();
+  connect(WbSimulationState::instance(), &WbSimulationState::renderingStateChanged, this, &WbSimulationView::updateRendering);
+
```
stefaniapedrazzi, line 237: I would move these lines just after adding the `WbAction::RENDERING` to the toolbar

src/webots/gui/WbSimulationView.cpp
```diff
@@ -582,8 +589,8 @@ void WbSimulationView::stopMovie() {
 }
 
 void WbSimulationView::makeMovie() {
-  if (WbSimulationState::instance()->isFast()) {
-    WbLog::warning(tr("Impossible to record a movie while running the simulation in 'Fast' mode."), true);
+  if (!WbSimulationState::instance()->isRendering()) {
+    WbLog::warning(tr("Impossible to record a movie while rendering is hidden."), true);
```
stefaniapedrazzi, line 593:
```suggestion
    WbLog::warning(tr("Impossible to record a movie while rendering is turned off."), true);
```

src/webots/gui/WbSimulationView.cpp
```diff
@@ -607,23 +614,23 @@ void WbSimulationView::makeMovie() {
...
+void WbSimulationView::showRenderingIfNecessary() {
   // remove "Fast Mode" overlay if necessary
-  if (WbSimulationState::instance()->isFast()) {
-    WbSimulationState::instance()->setMode(WbSimulationState::RUN);
-    mView3D->hideFastModeOverlay();
-    mNeedRestoreFastMode = true;
+  if (!WbSimulationState::instance()->isRendering()) {
+    WbSimulationState::instance()->setRendering(true);
+    mView3D->hideBlackRenderingOverlay();
+    mNeedToHideBlackRenderingOverlay = true;
```
stefaniapedrazzi, line 625: This variable name doesn't seem correct.
```suggestion
    mNeedToRestoreBlackRenderingOverlay = true;
```

src/webots/gui/WbSimulationView.cpp
```diff
@@ -555,7 +562,7 @@ void WbSimulationView::startVideoCapture(const QString &fileName, int codec, int
 
 void WbSimulationView::stopVideoCapture(bool canceled) {
   WbVideoRecorder::instance()->stopRecording(canceled);
-  restoreFastModeIfNecessary();
+  restoreRenderingIfNecessary();
```
stefaniapedrazzi, line 565:
```suggestion
  restoreRenderingIfNecessary();
```
This function name is wrong.
If a video is recorded but the rendering is turned off, the rendering is turned while recording the video and the turned off again when the video recording is over.
So something like this would be a better function name
```suggestion
  restoreNoRenderingIfNecessary();
```

src/webots/gui/WbSimulationView.cpp
```diff
@@ -746,24 +753,27 @@ void WbSimulationView::realTime() {
...
 
-void WbSimulationView::updateFastModeOverlay() {
-  if (WbSimulationState::instance()->mode() == WbSimulationState::FAST)
-    renderABlackScreen();
+void WbSimulationView::toggleRendering() {
+  if (WbSimulationState::instance()->isRendering())
+    WbSimulationState::instance()->setRendering(false);
   else
+    WbSimulationState::instance()->setRendering(true);
+}
```
stefaniapedrazzi, line 770: This could be simplified
```suggestion
void WbSimulationView::toggleRendering() {
  WbSimulationState::instance()->setRendering(!WbSimulationState::instance()->isRendering());
}
```

stefaniapedrazzi (changes requested):
There are still some issues with text and variable/functions names.
I will now double check the new signal introduced in `WbSimulationState` and test the functionality.

src/webots/gui/WbView3D.cpp
```diff
@@ -329,7 +329,7 @@ void WbView3D::refresh() {
   mPhysicsRefresh = true;
   if (sim->isPaused())
     renderLater();
```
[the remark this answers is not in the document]
stefaniapedrazzi: Sorry, this is wrong as the black overlay needs to the rendered too.
Maybe we could change the first function check from 
```
  if (!mWorld) {
    // render black screen
    renderLater();
    return;
  }
```
to
```
  if (!mWorld || !WbSimulationState::instance()->isRendering) {
    // render black screen
    renderLater();
    return;
  }
```

stefaniapedrazzi:
Connection of `modeChanged` and `renderingStateChanged` have to be updated.
If you want I can commit the new connections tomorrow.

src/webots/gui/WbView3D.cpp
```diff
@@ -1063,14 +1063,14 @@ void WbView3D::setWorld(WbSimulationWorld *w) {
   mPicker = new WbWrenPicker();
 
   // Creates the fast mode overlay
-  if (!mFastModeOverlay) {
-    mFastModeOverlay = new WbWrenFullScreenOverlay("Fast mode", 128, true);
-    mFastModeOverlay->attachToViewport(wr_scene_get_viewport(wr_scene_get_instance()));
+  if (!mDisabledRenderingOverlay) {
+    mDisabledRenderingOverlay = new WbWrenFullScreenOverlay("Rendering is Hidden", 128, true);
```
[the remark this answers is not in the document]
lukicdarkoo: `Rendering is turned off` is too long, I prefer `No Rendering`

lukicdarkoo:
> Connections to `modeChanged` and `renderingStateChanged` signals have to be updated.
> If you want I can commit the new connections tomorrow.

As you wish. But what is wrong with the current implementation?

stefaniapedrazzi:
You did not update the slots that are connected to the `modeChanged` but they actually need the new `renderingStateChanged` (or both).

src/webots/gui/WbSimulationView.cpp
```diff
@@ -746,24 +753,24 @@ void WbSimulationView::realTime() {
...
-
 void WbSimulationView::fast() {
   WbSimulationState::instance()->setMode(WbSimulationState::FAST);
 }
 
-void WbSimulationView::disableFastMode(bool disabled) {
-  WbActionManager::instance()->action(WbAction::FAST)->setEnabled(!disabled);
-  mView3D->setUserInteractionDisabled(WbAction::DISABLE_FAST_MODE, disabled);
+void WbSimulationView::disableRendering(bool disabled) {
+  WbActionManager::instance()->action(WbAction::RENDERING)->setEnabled(!disabled);
```
stefaniapedrazzi, line 761: Do you think we should also turn off the rendering when the DISABLE_RENDERING option is checked?

stefaniapedrazzi (commented):
Thank you for the changes!
I checked the the functionality and everything works fine.

Could you please add an entry in the ChangeLog?

Then, the `SIMULATION_MODE_RUN` should be removed from the C++ (and Java and Python) APIs, i.e. at least at
* include/controller/cpp/webots/Supervisor.h
* docs/reference/supervisior.md (still remains in the specific language tabs)

I think we should later also add a function to disable the rendering from the controller, but this could be done in a different PR.

src/webots/gui/WbSimulationView.cpp
lukicdarkoo: Previously, it was just locking the `FAST` state, but I agree that turning the rendering off makes more sense

lukicdarkoo: Implemented. I made more changes than expected as the rendering state has to be restored from the current session and from the `.wbproj` file. So let me know if it is good.

stefaniapedrazzi: Thank you! It works fine

stefaniapedrazzi:
Just to not forget, given that the appearance of the toolbar changed we should also update all the screenshot of the Webots UI included in the documentation. But this can be done in a separate PR (before releasing Webots R2021a).

src/webots/gui/WbSimulationView.cpp
lukicdarkoo: Also, the changelog was updated before (the last item):
https://github.com/cyberbotics/webots/blob/a76b3502fd24dd5b619174874ba3c9246f32d7ae/docs/reference/changelog-r2021.md

lukicdarkoo: Does it look good now, can we merge?

stefaniapedrazzi:
:eyes:

docs/reference/changelog-r2021.md
```diff
@@ -8,3 +8,4 @@ Released on December XXth, 2020.
     - Devices added during simulation run are now available in the robot controller ([#2237](https://github.com/cyberbotics/webots/pull/2237)).
     - Added the `wb_connector_is_locked` function to returns the current *isLocked* state of a [Connector](connector.md) ([#2087](https://github.com/cyberbotics/webots/pull/2087)).
     - Added the possibility to use an EUN (East-Up-North) coordinate system in the `coordinateSystem` field of the [WorldInfo](worldinfo.md) node ([#2228](https://github.com/cyberbotics/webots/pull/2228)).
+    - Added the possibility to disable rendering in the `realtime` mode ([#2286](https://github.com/cyberbotics/webots/pull/2286))
```
stefaniapedrazzi, line 11: We should also add in bold (because it breaks the compatibility) that the `run` mode has been removed.

lukicdarkoo: Added as a separated point

stefaniapedrazzi (dismissed):
Thank you!

omichel:
It looks all good to me as well. However, I believe the white bar on the disabled rendering icon looks strange. I would use some slightly dark red for the bar instead. I know this would make an icon with colors, but that's only the disabled version of the original icon (like the animation or movie recording with have a blinking red dot). Maybe we could use the same red for the cross of the disabled sound icon, so that it will make the whole thing look more consistent?

omichel:
I changed a little bit the red color to make it more red and more consistent with the rest of the GUI (I took the same red as the red text in the console in dark mode).

omichel (dismissed):
Thank you.

lukicdarkoo:
@omichel I added fix for macOS:
https://github.com/actions/virtual-environments/issues/1811#issuecomment-708480190

DavidMansolino:
58dab33990c484daa6130733c4ce6938d4bd2cfe should be merged on master too.

omichel (approved):
Thank you.

PR #2289, opened 2020-10-02
Document the limits of WorldInfo.CFM

omichel:
Following a Discord discussion:

> yes Webots doesn't accept a CFM value negative or equal to 0 and the brackets in the documentation are wrong. But when changing it from the Webots scene tree I don't get any crash, just a warning in the console.

### DIFF
--- a/docs/reference/worldinfo.md
+++ b/docs/reference/worldinfo.md
@@ -5,7 +5,7 @@ WorldInfo {
   SFString title                          ""         # any string
   MFString info                           [ ]        # any string
   SFFloat  gravity                        9.81       # [0, inf)
-  SFFloat  CFM                            0.00001    # ]0, inf)
+  SFFloat  CFM                            0.00001    # (0, inf)
   SFFloat  ERP                            0.2        # [0, 1]
   SFString physics                        ""         # any string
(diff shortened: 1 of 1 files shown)


docs/reference/worldinfo.md
```diff
@@ -5,7 +5,7 @@ WorldInfo {
   SFString title                          ""         # any string
   MFString info                           [ ]        # any string
   SFFloat  gravity                        9.81       # [0, inf)
-  SFFloat  CFM                            0.00001    # [0, inf)
+  SFFloat  CFM                            0.00001    # ]0, inf)
```
lukicdarkoo, line 8: I am not familiar with this notation, shouldn't it be `(` instead of `]`?

omichel: Yes, you are perfectly right:
- https://socratic.org/questions/how-do-you-know-when-to-use-brackets-or-parenthesis-in-finding-domain-or-range
- https://hannahtaylormath.weebly.com/domain-and-range.html

I used the closing bracket in my math courses, but that is maybe something not very standard. I will change it.

lukicdarkoo (approved):
Thanks!

PR #2296, opened 2020-10-05
Optimize point cloud getter in Python

### LINKED ISSUE #2283: Point cloud performance improvement
**Describe the Bug**
Method `lidar.getPointCloud()` (Python) is a few times slower than `wb_get_point_cloud()` (C).

See conversation:
https://github.com/cyberbotics/webots_ros2/pull/148#issuecomment-702001390

**Steps to Reproduce**
1. Run the following world in Webots: https://github.com/cyberbotics/webots_ros2/blob/b4d79ee9259ead364dbc7ee50ad7115e39ba5cb6/webots_ros2_universal_robot/worlds/universal_robot_lidar.wbt
2. Use the Python controller (see **Additional context**) and check the real-time factor.
2. Use the C controller (see **Additional context**) and check the real-time factor.

**Expected behavior**
Method `lidar.getPointCloud()` (Python) should be nearly the speed of `wb_get_point_cloud()` (C).

**Additional context**
<details>
<summary>Example C controller</summary>

```c
#include <stdio.h>
#include <webots/robot.h>
#include <webots/lidar.h>


#define TIME_STEP 32


  wb_robot_init();
  int device = wb_robot_get_device("Velodyne VLP-16");
  wb_lidar_enable(device, TIME_STEP);
  wb_lidar_enable_point_cloud(device);
  
    WbLidarPoint* points = wb_lidar_get_point_cloud(device);
  };

  wb_robot_cleanup();

  return 0;
}
```

</details>

<details>
<summary>Example Python controller</summary>

```Python
from controller import Robot


robot = Robot()
timestep = 32
lidar = robot.getLidar('Velodyne VLP-16')
lidar.enable(timestep)
lidar.enablePointCloud()

    vals = lidar.getPointCloud()
    sum_val = sum(val.x for val in vals)
    print(sum_val)
```

</details>

lukicdarkoo:
**Description**
This PR improves the performances of Python's `getPointCloud()` by 4-5 times.

**Related Issues**
Fixes #2283 

**Tasks**
  - [ ] Add to changelog
  - [ ] Verify side effects (memory ownership)
  - [ ] Hide `_getPointCloud()` from user

### DIFF
--- a/resources/languages/python/controller.i
+++ b/resources/languages/python/controller.i
@@ -85,6 +85,7 @@ using namespace std;
 
 //handling std::string
 %include "std_string.i"
+%include "carrays.i"
 
 %rename ("__internalGetLookupTableSize") getLookupTableSize;
 
@@ -522,8 +523,10 @@ class AnsiCodes(object):
     $result = Py_None;
 }
 
-%ignore webots::Lidar::getPointCloud();
+%rename (_getPointCloud) getPointCloud() const;
 %ignore webots::Lidar::getLayerPointCloud();
+
+%array_class(WbLidarPoint, WbLidarPointArray);
 
 %extend webots::Lidar {
 
@@ -539,10 +542,8 @@ class AnsiCodes(object):
 
   %pythoncode %{
   def getPointCloud(self):
-     ret = []
-     for i in range(self.getNumberOfPoints()):
-       ret.append(self.getPoint(i))
-     return ret
+    points = self._getPointCloud()
+    return WbLidarPointArray.frompointer(points)
 
   def getLayerPointCloud(self, layer):
      ret = []


lukicdarkoo:
This unfortunately doesn't return the Python list. Any kind of conversion to list (that I can find) would require memory copying and iterating through the array - performance drop. Maybe we can add it as another method, e.g. `getPointCloudFast()`

DavidMansolino:
> This unfortunately doesn't return the Python list. Any kind of conversion to list (that I can find) would require memory copying and iterating through the array - performance drop. Maybe we can add it as another method, e.g. `getPointCloudFast()`


We have something similar for the camera API: https://www.cyberbotics.com/doc/reference/camera?tab-language=python#wb_camera_get_image
`getImage` returns the 'raw pointer' and `getImageArray` returns a Python list, we can probably do something similar and reuse the terminology.

lukicdarkoo:
@DavidMansolino In that case `getPointCloudArray()` would return list of objects, while `getPointCloud()` would return C-like array. By introducing the change we are breaking the API (`getPointCloud()` returns C-like array instead of list). It is good to make it consistent, but it is quite a big change.

lukicdarkoo:
Here is the new API proposal:
- Images:
  - `getImageRaw()` returns raw data
  - `getImageList()` returns list of lists
  - `getImage()` returns raw data + deprecation
  - `getImageArray()` returns list of lists + deprecation
- Point cloud:
  - `getPointCloud()` returns list + deprecation
  - `getPointCloudRaw()` returns raw data
  - `getPointCloudList()` returns list

But still, I don't like the fact that `getPointCloud()` and `getImage()` will be deprecated (which makes Python API inconsistent with C++) and I am not sure if there is any other sensor that we should also consider.

DavidMansolino:
About the other functions I see those ones:
  - `Camera.getRecognitionObjects` returns currently a list.
  - `Lidar.getLayerPointCloud` returns currently a list.
  - `Lidar.getRangeImageArray` returns currently a list.
  - `RangeFinder.getRangeImageArray` returns currently a list.
  - `RangeFinder.getImage` returns currently returns raw data.

> But still, I don't like the fact that getPointCloud() and getImage() will be deprecated (which makes Python API inconsistent with C++) and I am not sure if there is any other sensor that we should also consider.

I agree, I like this C++/Python matching, but on the other side I don't see any other clean way to do this (the alternative, would  be to keep `Camera.getImage` instead of `Camera.getImageRaw`, but then we have again the problem with lidar).

omichel:
After a second thought, I don't like too much `getImageList` as it sounds like this function will return a list of images.

omichel:
What about removing the *array* methods and keep only the methods returning raw data. Then we could have local python methods that could convert this raw data into different kind of list or arrays? That would make the API very consistent with C/C++ and we would provide some extra utility functions for Python ease-of-use. I am wondering if performance-wise, it is equivalent to get raw data and convert it in Python list using a python utility method or to get already data converted to Python list in swig (as we do currently). If the performance is the same, it is probably the cleanest option.

lukicdarkoo:
> What about removing the _array_ methods and keep only the methods returning raw data? Then we could have local python methods that could convert this raw data into different kind of list or arrays? That would make the API very consistent with C/C++ and we would provide some extra utility functions for Python ease-of-use. I am wondering if performance-wise, it is equivalent to get raw data and convert it in Python list using a python utility method or to get already data converted to Python list in swig (as we do currently). If the performance is the same, it is probably the cleanest option.

In the case of the image it should be slower (but I am not sure how much) as the list is created in C:
https://github.com/cyberbotics/webots/blob/6bb8d98232275e19bef3bfad52f89a355bca830f/resources/languages/python/controller.i#L285
In the case of the point cloud it should be the same as it is already implemented in Python:
https://github.com/cyberbotics/webots/blob/6bb8d98232275e19bef3bfad52f89a355bca830f/resources/languages/python/controller.i#L541

omichel:
Could you try to evaluate how much slower it is for the camera? This will help us to decide whether or not we want only one or several API functions to retrieve the same data in different formats.
By the way, instead of several API functions, we could have a single function which takes an optional format argument.

lukicdarkoo:
It seems that creating a list in SWIG is not slow, but wrapping objects and memory copying. For example:

```c++
  PyObject* getPointCloud() const {
    const WbLidarPoint* rawPoints = $self->getPointCloud();
    const int size = $self->getNumberOfPoints();

    PyObject *points = PyList_New(size);
    for (int i = 0; i < size; i++) {
      PyObject *v = SWIG_NewPointerObj(SWIG_as_voidptr(&rawPoints[0]), $descriptor(WbLidarPoint *), SWIG_POINTER_NOSHADOW);
      PyList_SetItem(points, i, v);
    }
    return points;
  }
```

is much faster than:

```c++
  PyObject* getPointCloud() const {
    const WbLidarPoint* rawPoints = $self->getPointCloud();
    const int size = $self->getNumberOfPoints();

    PyObject *points = PyList_New(size);
    for (int i = 0; i < size; i++) {
      PyObject *v = SWIG_NewPointerObj(SWIG_as_voidptr(&rawPoints[0]), $descriptor(WbLidarPoint *), 0);
      PyList_SetItem(points, i, v);
    }
    return points;
  }
```

Both return a Python list of objects. Unfortunately, the first implementation returns opaque objects. In any case, the second implementation is around 50-60% faster than the implementation currently used in R2020b Rev 1. This means we can create lists 50-60% faster in SWIG.

PR #2303, opened 2020-10-07
Workaround AppVeyor broken server

omichel:
Another attempt at working around broken MSYS2 main server.


DavidMansolino:
:eyes:

DavidMansolino:
Shouldn't we apply the patch to https://github.com/cyberbotics/webots/blob/master/.appveyor.yml too, it seems it is having issues installing cppcheck: https://ci.appveyor.com/project/omichel/webots#L18

DavidMansolino (approved):
Seems to works fine now thank you.

DavidMansolino:
Unfortunately it is still failing: https://ci.appveyor.com/project/omichel/webots-master?fullLog=true#L144
But the good news is that the MSYS2 bug is supposed to be fixed now: https://github.com/msys2/MSYS2-packages/issues/2171#issuecomment-705210531

Issue #2317, opened 2020-10-12
Python 3.9 support


omichel:
Python 3.9.0 was just released. Webots should support it.

PR #2318, opened 2020-10-12
Add python3.9 support

DavidMansolino:
**Description**
Implement #2317 for Linux.

### DIFF
--- a/projects/default/libraries/vehicle/Makefile
+++ b/projects/default/libraries/vehicle/Makefile
@@ -39,6 +39,8 @@ endif
 	+PYTHON_COMMAND=python3.7 make -s -C python $(MAKECMDGOALS)
 	@echo "# make" $@ python3.8;
 	+PYTHON_COMMAND=python3.8 make -s -C python $(MAKECMDGOALS)
+	@echo "# make" $@ python3.9;
+	+PYTHON_COMMAND=python3.9 make -s -C python $(MAKECMDGOALS)
 endif
 ifeq ($(OSTYPE),windows)
 	@echo "# make" $@ python3.8;
--- a/projects/default/libraries/vehicle/Makefile
+++ b/projects/default/libraries/vehicle/Makefile
@@ -57,6 +57,8 @@ ifeq ($(OSTYPE),darwin)
 	+PYTHON_COMMAND=python3.7 make -s -C python $(MAKECMDGOALS)
 	@echo "# make" $@ python3.8;
 	+PYTHON_COMMAND=python3.8 make -s -C python $(MAKECMDGOALS)
+	@echo "# make" $@ python3.9;
+	+PYTHON_COMMAND=python3.9 make -s -C python $(MAKECMDGOALS)
 endif
 
 
--- a/projects/default/libraries/vehicle/Makefile
+++ b/projects/default/libraries/vehicle/Makefile
@@ -43,8 +43,10 @@ endif
 	+PYTHON_COMMAND=python3.9 make -s -C python $(MAKECMDGOALS)
 endif
 ifeq ($(OSTYPE),windows)
+	@echo "# make" $@ python3.9;
+	+PATH="$(PYTHON39_HOME):$(PATH)" make -s -C python $(MAKECMDGOALS)
 	@echo "# make" $@ python3.8;
-	+make -s -C python $(MAKECMDGOALS)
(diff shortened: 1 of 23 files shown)


DavidMansolino:
Current issues:
  - **Linux**: all good.
  - **Windows**:
    - Python 3.9 is not available on Appveyor, but we can install it manually.
    - Some pip package used in the test are not yet available on Windows/Python 3.9, but that's fine we can keep the source test on Python 3.8 for now.
  - **MacOS**: For some reason, on Github Actions I am not allowed to run Python 3.9 after (manually) installing it, it gives a `/Library/Frameworks/Python.framework/Versions/3.9/bin/python3.9: Permission denied` even so I do a `chmod 777` on it.

DavidMansolino:
> MacOS: For some reason, on Github Actions I am not allowed to run Python 3.9 after (manually) installing it, it gives a /Library/Frameworks/Python.framework/Versions/3.9/bin/python3.9: Permission denied even so I do a chmod 777 on it.

Same problem on Travis :-/

projects/robots/robotis/darwin-op/libraries/Makefile
```diff
@@ -54,10 +54,14 @@ endif
 	+PYTHON_COMMAND=python3.7 make -s -C python $(MAKECMDGOALS)
 	@echo "# make" $@ python3.8;
 	+PYTHON_COMMAND=python3.8 make -s -C python $(MAKECMDGOALS)
+	@echo "# make" $@ python3.9;
+	+PYTHON_COMMAND=python3.9 make -s -C python $(MAKECMDGOALS)
 endif
 ifeq ($(OSTYPE),windows)
+	@echo "# make" $@ python3.9;
+	+PATH="$(PYTHON39_HOME):$(PATH)" make -s -C python $(MAKECMDGOALS)
```
omichel, line 62: Can't we simply have this?
```suggestion
	+make -s -C python $(MAKECMDGOALS)
```
As `PYTHON39_HOME` should already be in the `PATH`, isn't it?

resources/languages/Makefile
```diff
@@ -34,10 +34,14 @@ endif
 	+PYTHON_COMMAND=python3.7 make -s -C python $@
 	@echo "# make" $@ python3.8;
 	+PYTHON_COMMAND=python3.8 make -s -C python $@
+	@echo "# make" $@ python3.9;
+	+PYTHON_COMMAND=python3.9 make -s -C python $@
 endif
 ifeq ($(OSTYPE),windows)
+	@echo "# make" $@ python3.9;
+	+PATH="$(PYTHON39_HOME):$(PATH)" make -s -C python $@
```
omichel, line 42: Same comment here.

projects/default/libraries/vehicle/Makefile
```diff
@@ -39,10 +39,14 @@ endif
 	+PYTHON_COMMAND=python3.7 make -s -C python $(MAKECMDGOALS)
 	@echo "# make" $@ python3.8;
 	+PYTHON_COMMAND=python3.8 make -s -C python $(MAKECMDGOALS)
+	@echo "# make" $@ python3.9;
+	+PYTHON_COMMAND=python3.9 make -s -C python $(MAKECMDGOALS)
 endif
 ifeq ($(OSTYPE),windows)
+	@echo "# make" $@ python3.9;
+	+PATH="$(PYTHON39_HOME):$(PATH)" make -s -C python $(MAKECMDGOALS)
```
omichel, line 47: Same comment here (see below).

omichel (changes requested):
Thank you.
It looks very good, just a minor detail.

projects/robots/robotis/darwin-op/libraries/Makefile
DavidMansolino: Yes, sure we can as in theory `scripts/install/msys.sh` is already setting it.
But I thought it was more homogeneous like this (less special cases means less sources of errors) and more handy/safer (I already encountered several times on Windows the case where I changed manually my PATH to point to python 2.7 and then the compilation failed because of this).
But I don't have any strong opinion on this, so if you prefer that I remove it it's fine too for me.

omichel: I see. It's true, that it is more robust like that, and it is often useful to change the PATH environment variable to make some tests, so let's keep it like this.

omichel (approved):
Thank you.

PR #2321, opened 2020-10-12
Get point cloud as a buffer

### LINKED ISSUE #2283: Point cloud performance improvement

lukicdarkoo:
**Description**
Since I couldn't find a solution to improve the performance of `getPointCloud()` without changing the API (see #2296) this introduces a new argument `dtype` which defines data type.

See conversation:
https://github.com/cyberbotics/webots_ros2/pull/148#issuecomment-702001390

**Related Issues**
Closes #2283

**Tasks**
 - [x] Add to changelog
 - [x] Add documentation

**API Proposal**

`getPointCloud([dtype])`
- `dtype` can be ['list', 'buffer']

**Additional context**
- If `dtype` == 'list'
  - Creating a list is 50-60% faster than the previous implementation
- If `dtype` == 'buffer'
  - Performances are almost the same as for C API
  - Data can be passed directly to [`PointCloud2`](http://docs.ros.org/melodic/api/sensor_msgs/html/msg/PointCloud2.html) (see [example](https://gist.github.com/pgorczak/5c717baa44479fa064eb8d33ea4587e0))
  - It can be converted to numpy array using `data = np.frombuffer(data, dtype=np.uint8)` without overhead
- This change doesn't break the API
- It can scale to other parts of the API without breaking it
- We can continue adding new data types

----

**Personal notes:**

- Constructor for `array.array` (`newarrayobject`):
https://github.com/python/cpython/blob/e51dd9dad6590bf3a940723fbbaaf4f64a3c9228/Modules/arraymodule.c#L570

- Creates `array.array` object by calling `array_array_frombytes`:
https://github.com/python/cpython/blob/e51dd9dad6590bf3a940723fbbaaf4f64a3c9228/Modules/arraymodule.c#L2663

- Bytes to `array.array` is only doing `memcpy()` (and some checks):
https://github.com/python/cpython/blob/e51dd9dad6590bf3a940723fbbaaf4f64a3c9228/Modules/arraymodule.c#L1615

- Access to memory without copying (only get):
```Python
mv = memoryview(data)
mv[0:100]
```

- Pointer to the memory of `array.array`:
```C
op->ob_item
```

- The internal buffer cannot be alternated through [`PyObject_GetBuffer`](https://docs.python.org/3/c-api/buffer.html#c.PyObject_GetBuffer) as `view->buf` is assigned on the call:
https://github.com/python/cpython/blob/e51dd9dad6590bf3a940723fbbaaf4f64a3c9228/Modules/arraymodule.c#L2523

-


lukicdarkoo:
I found a way to pass `bytearray` to ROS2 without mimicking `array.array`:
https://github.com/cyberbotics/webots_ros2/blob/4a02a97659ab84c8bd01e72eb1ac9d7ed7e4e6b0/webots_ros2_core/webots_ros2_core/devices/lidar_device.py#L114

It is actually a straightforward solution, but I thought it would not work as the data type is not the same.

resources/languages/python/controller.i
```diff
@@ -557,17 +557,40 @@ class AnsiCodes(object):
     return point[index];
   }
 
+  PyObject* __getPointCloudBuffer() const {
+    const char * points = (const char *)$self->getPointCloud();
```
omichel, line 561:
```suggestion
    const char *points = (const char *)$self->getPointCloud();
```

resources/languages/python/controller.i
```diff
@@ -557,17 +557,40 @@ class AnsiCodes(object):
...
+    const int size = $self->getNumberOfPoints() * sizeof(WbLidarPoint);
+    return PyBytes_FromStringAndSize(points, size);
+  }
+
+  PyObject* __getPointCloudList() const {
+    const WbLidarPoint* rawPoints = $self->getPointCloud();
```
omichel, line 567:
```suggestion
    const WbLidarPoint *rawPoints = $self->getPointCloud();
```

resources/languages/python/controller.i
```diff
@@ -557,17 +557,40 @@ class AnsiCodes(object):
     return point[index];
   }
 
+  PyObject* __getPointCloudBuffer() const {
```
omichel, line 560:
```suggestion
  PyObject *__getPointCloudBuffer() const {
```

omichel (changes requested):
That solution looks great and simple!
I only have a doubt about the `dtype` naming (I saw this is used in numpy, but usually, we prefer a more standard naming scheme). I would suggest `data_type` instead.

lukicdarkoo:
> That solution looks great and simple!
> I only have a doubt about the `dtype` naming (I saw this is used in numpy, but usually, we prefer a more standard naming scheme). I would suggest `data_type` instead.

Yes, `dtype` comes from numpy, but `data_type` is prettier, I will change it.

omichel (approved):
Thank you.

Issue #2323, opened 2020-10-12
Error after adding battery on DJI Mavic2Pro drone


kosnet2:
- I am gettting this warning after adding a battery on the existing dji Mavic2Pro drone with predefined values

```
  field MFFloat    battery         [50, 100, 30]
  field SFFloat    cpuConsumption  10
```
`WARNING: Mavic2Pro (PROTO) > DEF FRONT_RIGHT_HP_HELIX Propeller > RotationalMotor: Feedback is available for motorized joints only`

- If I don't add the battery the controller is giving this error:

`WARNING: Mavic2Pro (PROTO): 'wb_robot_battery_sensor_enable' called while the 'battery' field is empty.`

- If go with the first solution but leave the battery fields empty (ie)

```
  field MFFloat    battery         []
  field SFFloat    cpuConsumption  10
```
and then add the battery values through the webots hierarchy editor, Webots **crashes**

Does anyone have a clue what am I doing wrong ?
My controller is in python

DavidMansolino:
Which version of Webots are you using?
We fixed several issues related to battery recently (which sounds like what you are mentioning):
> Fixed crash when calling the wb_robot_battery_sensor_enable() function if the battery field of the Robot is empty (#1801).
>Fixed annoying warning messages when calling the wb_robot_battery_sensor_enable() function if the Robot contains a Track node (#1801).

But it looks indeed that the 'annoying warning messages' are still present if the Robot contains a Propeller node. We will fix this soon. In the meantime, you can use the first option and ignore the warning messages.

kosnet2:
@DavidMansolino Thank you for your feedback :)

I am currently using Webots R2020b revision 1 and I have installed the latest update.

Is there any way to switch off the warnings ?

DavidMansolino:
To 'hide' them you hve to right click on the Webots console and select all the filters except 'Others'.
Note also that I just fixed the issue in  #2324. The fix will be included in the nightly build of R2020b-rev2 from tomorrow: https://github.com/cyberbotics/webots/releases

PR #2324, opened 2020-10-13
Fix battery with propeller

DavidMansolino:
**Description**
Fix error message when using a battery with propeller: #2323

### DIFF
--- a/src/webots/nodes/WbPropeller.cpp
+++ b/src/webots/nodes/WbPropeller.cpp
@@ -218,26 +218,23 @@ void WbPropeller::prePhysicsStep(double ms) {
     const double V = dCalcVectorDot3(vp, mNormalizedAxis.ptr());
 
     const WbVector2 &tcs = mTorqueConstants->value();
-    double torque = tcs.x() * velocity * absoluteVelocity - tcs.y() * absoluteVelocity * V;
+    mCurrentTorque = tcs.x() * velocity * absoluteVelocity - tcs.y() * absoluteVelocity * V;
     const double mt = m->maxForceOrTorque();
-    if (fabs(torque) > mt)
-      torque = torque > 0.0 ? mt : -mt;
+    if (fabs(mCurrentTorque) > mt)
+      mCurrentTorque = mCurrentTorque > 0.0 ? mt : -mt;
 
     const WbVector2 &fcs = mThrustConstants->value();
-    const double thrust = fcs.x() * velocity * absoluteVelocity - fcs.y() * absoluteVelocity * V;
+    mCurrentThrust = fcs.x() * velocity * absoluteVelocity - fcs.y() * absoluteVelocity * V;
 
     // Applies thrust and torque
     const WbMatrix3 &m3 = ut->rotationMatrix();
     const WbVector3 &axis = m3 * mNormalizedAxis;
-    const WbVector3 &thrustVector = thrust * axis;
-    const WbVector3 &torqueVector = -torque * axis;
+    const WbVector3 &thrustVector = mCurrentThrust * axis;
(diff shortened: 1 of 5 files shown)


src/webots/nodes/WbPropeller.cpp
```diff
@@ -230,6 +236,9 @@ void WbPropeller::prePhysicsStep(double ms) {
     dBodyAddForceAtPos(b, thrustVector.x(), thrustVector.y(), thrustVector.z(), cot.x(), cot.y(), cot.z());
     dBodyAddTorque(b, torqueVector.x(), torqueVector.y(), torqueVector.z());
 
+    mCurrentThrust = thrust;
+    mCurrentTorque = torque;
+
```
omichel, line 241: I believe the `thrust` and `torque` variables are not needed any more and we can directly compute the value in the ` mCurrentThrust` and `mCurrentTorque` member fields.

src/webots/nodes/WbRotationalMotor.cpp
```diff
@@ -84,6 +85,10 @@ void WbRotationalMotor::turnOffMotor() {
 }
 
 double WbRotationalMotor::computeFeedback() const {
+  const WbPropeller *propeller = dynamic_cast<WbPropeller *>(parentNode());
+  if (propeller)
+    return fabs(propeller->currentThrust()) + fabs(propeller->currentTorque());
```
omichel, line 90: Is that correct? Can we simply add a force to a torque? That seems odd... In the case of a rotational motor, the feedback is supposed to be a torque. So the thrust should be somehow converted to the equivalent torque.

omichel (changes requested):

src/webots/nodes/WbRotationalMotor.cpp
DavidMansolino: This is indeed an approximation, but we anyway claim in the documentation that the consumption of motors is a rough estimation. We define it as `force * consumptionFactor` for motors generating a force (LinearMotor) or as `torque * consumptionFactor` for motors generating a torque (RotationalMotor). In the case of Propeller, it is special, it is a RotationalMotor, but it is generating both a torque and a force (thrust), so the idea here is to take both into account for the consumption, for converting the force (thrust) to a moment we are here missing information (we need a kind of 'Lever Arm Length', we could introduce a new field to the propeller, but this looks overkilled to me, especially because we are then using an extremely approximated model to convert torque to consumption).
Otherwise, another solution would be to do like we did for Tracks (simply disabling consumption for motor in Propellers).

omichel: I am not sure we need a 'lever arm length' as it could probably be computed (or estimated) from the torque and thrust constants. But I am not sure about the formula... So, let's proceed with this approximation, although it looks wrong... Maybe some comment in the code would be helpful, so that we know it's wrong.

omichel: Intuitively, I would say it should be something like torque feedback = currentCurrentTorque + currrentForce * (q1/t1 + q2/t2).

DavidMansolino: This formula looks indeed intuitively rights. But there are some cases which don't make sense e.g.:
Imagine a 'perfect' propeller (produce thrust without resistant torque), then q1 and q2 and currentTorque = 0, but currentForce != 0, this means we return a feedback of 0, but some thrust is produced.
And the problem is even worse if t1 or t2 is 0 (which is the case with the default fields value).

omichel: OK, so let write a comment in there until we find a better formula.

omichel (approved):
Thank you.

PR #2326, opened 2020-10-13
Fix Motor Target Position when Changing Joint Parameter Position

DavidMansolino:
**Description**
Previously, when changing the 'position' field of a joint parameter, a save and revert was required to update the motor target position, otherwise when starting the simulation the motor was pushing the joint to the 0 position again.

### DIFF
--- a/tests/api/controllers/motor/motor.c
+++ b/tests/api/controllers/motor/motor.c
@@ -81,7 +81,7 @@ int main(int argc, char **argv) {
 
   const double target_position = wb_motor_get_target_position(motor);
   ts_assert_double_equal(target_position, max_position,
-                         "The target position value measured by the motor should be %g and not %g", M_PI_2, target_position);
+                         "The target position value measured by the motor should be %g and not %g", max_position, target_position);
   int i;
   for (i = 0; i < NUMBER_OF_STEPS; ++i)
     wb_robot_step(TIME_STEP);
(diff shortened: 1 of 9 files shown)


DavidMansolino:
Failures of the mac-sources test are not related to this PR.

tests/api/controllers/motor/motor.c
```diff
@@ -75,13 +75,13 @@ int main(int argc, char **argv) {
...
-  wb_motor_set_position(motor, M_PI_2);
+  wb_motor_set_position(motor, 100);
 
   wb_robot_step(TIME_STEP);
 
   const double target_position = wb_motor_get_target_position(motor);
-  ts_assert_double_equal(target_position, M_PI_2, "The target position value measured by the motor should be %g and not %g",
-                         M_PI_2, target_position);
+  ts_assert_double_equal(target_position, max_position,
+                         "The target position value measured by the motor should be %g and not %g", M_PI_2, target_position);
```
omichel, line 84: Something looks wrong in this assert, shouldn't it be:
```suggestion
  ts_assert_double_equal(target_position, max_position,
                         "The target position value measured by the motor should be %g and not %g", max_position, target_position);
```

omichel (changes requested):

omichel (dismissed):
Thank you.

omichel (approved):

PR #2328, opened 2020-10-14
Fix lua dependency download

DavidMansolino:
**Description**
Currently, the main build process on Linux is broken because the lua website switched from http to https only.


omichel (approved):
Thank you.

stefaniapedrazzi:
Same issue also on Windows

PR #2329, opened 2020-10-14
Fix installing lua dependency on Linux

stefaniapedrazzi:
Today on my Ubuntu 18.04 I had issues installing the lua dependency.
But using HTTPS instead of HTTP it works fine.


DavidMansolino:
Just fixed this in #2328 ;-)

PR #2330, opened 2020-10-14
Fixed Webots Windows launcher to escape properly spaces in arguments

omichel:
A user reported on Discord that he was unable to open a world file that includes spaces in the path when starting Webots from a Windows DOS console. He tried several things, including adding double quotes around the world path, but that didn't work. I found that adding 3 double quotes around the world path worked, but is quite odd:
```dos
C:\Users\Olivier>"C:\Program Files\Webots\msys64\mingw64\bin\webots.exe" """C:\Program Files\Webots\projects\samples\devices\worlds\camera.wbt"""
```
 This PR fixes the problem and makes that adding a single double quote around the path is sufficient, as it should be:
```dos
C:\Users\Olivier>"C:\Program Files\Webots\msys64\mingw64\bin\webots.exe" "C:\Program Files\Webots\projects\samples\devices\worlds\camera.wbt"
```


DavidMansolino (approved):
Thank you!

PR #2333, opened 2020-10-14
Update tests_sources.yml

omichel:
This will avoid starting the test source action when we modify only the documentation in a PR.


DavidMansolino (approved):
Thank you

PR #2327, opened 2020-10-14
Fix enabling Help item in context menu for Recognition node

stefaniapedrazzi:
Currently the `Help..` item is disabled in the context menu when a Recognition node is selected.

All the other nodes without WREN objects have the same issue: ContactProperties, Focus, Zoom, etc.

PR #2335, opened 2020-10-14
Fix cppcheck issues

DavidMansolino:
**Description**
  - [x] Fix new cppcheck issues on macOS.
  - [x] Add a new `test sources` label to trigger sources test on all platforms
  - [x] Fix python3.7 brew build on macOS.

### DIFF
--- a/src/webots/sound/WbPicoTextToSpeech.cpp
+++ b/src/webots/sound/WbPicoTextToSpeech.cpp
@@ -349,9 +349,11 @@ qint16 *WbPicoTextToSpeech::generateBufferFromText(const QString &text, int *siz
         bufferSize += 1024 * 1024;  // 1 MB
         qint16 *previousBuffer = buffer;
         buffer = (qint16 *)realloc(buffer, bufferSize);
+        // cppcheck-suppress nullPointerArithmeticRedundantCheck
         if (!buffer)  // re-allocation failed, this is required otherwise CppCheck raises an error
           free(previousBuffer);
       }
+      // cppcheck-suppress nullPointerArithmeticRedundantCheck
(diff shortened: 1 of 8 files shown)


DavidMansolino:
I am taking advantage of this PR to introduce a new 'test sources' label to run the test on all the platforms without switching the PR to ready for review.

omichel (approved):
Thank you.

src/webots/sound/WbPicoTextToSpeech.cpp
```diff
@@ -351,6 +351,10 @@ qint16 *WbPicoTextToSpeech::generateBufferFromText(const QString &text, int *siz
         buffer = (qint16 *)realloc(buffer, bufferSize);
         if (!buffer)  // re-allocation failed, this is required otherwise CppCheck raises an error
           free(previousBuffer);
+        else {
+          gError = QString("Cannot re-allocate buffer.");
+          return NULL;
+        }
```
stefaniapedrazzi, line 357: This code seems wrong (and it is still failing at cppcheck on macOS https://github.com/cyberbotics/webots/pull/2327/checks?check_run_id=1259146248 ).

If `buffer` is not NULL, then the allocation didn't fail so we should neither return nor print an error.

stefaniapedrazzi: I fixed this in  #2347

DavidMansolino: You're completely right, fixing this right now.
Strangely, the cppcheck did't fails my previous PRs today: https://github.com/cyberbotics/webots/runs/1257659737#step:4:45

PR #2337, opened 2020-10-14
Merge master into develop

DavidMansolino:
**Description**
No conflicts.


omichel (approved):

omichel (changes requested):
Well, it seems we need to fix some new cppcheck issues in the code of the develop branch.

DavidMansolino:
It's actually not develop, but the other branch I merged just before merging the cppcheck on :disappointed:

DavidMansolino:
I can't understand why `motor2()` and `motor3()` are raising the `Possible null pointer dereference` error but `motor()` is not :confused:

omichel:
Maybe because in the base class (WbJoint.hpp) they return NULL and cppcheck is not smart enough to understand they are overridden?

DavidMansolino:
That can be a good explanation indeed.

omichel:
Anyhow, I believe it's not bad to cache the motor pointer to avoid calling the motor() method (which is not a simple getter) three times:
```C++
  WbMotor *m = motor();
  if (m && !m->isConfigureDone())
    m->setTargetPosition(position);
```
That should also fix the cppcheck errors.

DavidMansolino:
> Anyhow, I believe it's not bad to cache the motor pointer to avoid calling the motor() method (which is not a simple getter) three times:
> 
> ```c++
>   WbMotor *m = motor();
>   if (m && !m->isConfigureDone())
>     m->setTargetPosition(position);
> ```
> 
> That should also fix the cppcheck errors for motor2 and motor3.

Yes, that's exactly what I am doing in #2340

omichel (approved):
Thank you.

PR #2341, opened 2020-10-15
Improve device getters

### LINKED ISSUE #2338: Python's Robot.getDevice() crashes
The Python API includes a definition for `Robot.getDevice` despite this being undocumented.  The Webots docs encourage you to use more specific functions like `Robot.getGPS` instead, to ensure that the device you get will be a member of the appropriate subclass.  However, when I look at the actual source code for Webots `controller.py`, the function `__getOrCreateDevice` automatically detects each device's type and generates a device of the appropriate subclass, and this code seems to be used by each of the specific functions like getGPS, so I don't think the more specific call should really be necessary.  This is evidenced by the fact that Python's generic `getDeviceByIndex` works fine without needing different versions for the different types of devices.  So there's no reason `getDevice` couldn't work too.

Unfortunately, the existing Python `def getDevice(tag)` function suffers several problems.  It is defined as a `Robot` class method, but is missing the initial `self` argument that is required for such methods in Python, so it always crashes.  ( `getDevice( x )` has one too many arguments, and `getDevice()` treats the robot itself as the `tag` argument, which of course doesn't work.)  Furthermore, the given argument is called `tag` but Webots docs discuss the relevant arguments to various Python device-getting functions as being "names" or "indexes", not "tags" (which seem to be an internally used label for something that behaves much like an index).  The code itself presumes that the given `tag` must be a number, and will crash otherwise when it compares `tag` to the length of its list of devices.  This code could handle `robot.getDevice(1)` but not `robot.getDevice("gps")`.  This is odd when the analogous `get_device` functions in C and Matlab take a _string_ rather than an integer as their argument, and there already is another Python function `getDeviceByIndex` (with analogs in C and Matlab) which does look up devices by index.

**Proposed bugfixes:**  (1) Python's `Robot.getDevice` should include the required `self` parameter, becoming `getDevice(self, name_or_index)` so that it won't always crash. (2) Python's `getDevice` should accept string arguments, to parallel the corresponding functions in C and MatLab, and to distinguish it from `getDeviceByIndex`.  (3) Python does not require that an argument be of a pre-specified type, so `getDevice` could still accept integer index arguments too, which strikes me as useful, and better than making coders try to remember whether you called the indexed version "FromIndex" or "ByIndex" (the source code uses both of these, at different levels!).  (4) Documentation for Python's version of `Robot.getDevice` should be added matching whatever functionality it ends up having.  (5) The documentation encouraging using specific getX functions rather than this more generic `getDevice()` should be removed.  (In the source code, the only other thing the specific getX functions do is check that the device with that name really is a Webots device of the specified type, so e.g. `robot.getMotor("gps")` will return `None` if the device named "gps" isn't a motor.)

**Suggested Code.**  Note, the first line adds the missing `self` argument required in Python class methods, and makes clear that the given argument can be a device name or index number.  The `if` condition detects string inputs and treats them in exactly the same way names are treated when passed to specific commands like `getGPS`, whereas the `else` condition instead treats the parameter as an index number and treats it in exactly the same way as does `getDeviceByIndex` which is documented to work and has less redundancy than the original code with what is done by the final line.  The final line then calls a function that gets/creates the relevant device, doing so in a way that ensures that a device of the relevant subclass will be created.   

```python
    def getDevice(self, name_or_index):
      """This returns this robot's device with the given name if the given argument is a string,
      or with the given index number otherwise.  Returns None if there is no such device."""       
      if isinstance(name_or_index,str):
        tag = self.__internalGetDeviceTagFromName(name_or_index)
      else:
        tag = self.__internalGetDeviceTagFromIndex(name_or_index)
      return self.__getOrCreateDevice(tag)
```

omichel:
Resolves #2338.

Remains to be done:
- [ ] documentation: https://cyberbotics.com/doc/reference/robot?version=feature-improve-device-getters
- [ ] update all the python examples to use the new `getDevice()` method.

### DIFF
--- a/include/controller/cpp/webots/Robot.hpp
+++ b/include/controller/cpp/webots/Robot.hpp
@@ -82,6 +82,7 @@ namespace webots {
     double getBasicTimeStep() const;
     int getNumberOfDevices() const;
     Device *getDeviceByIndex(int index);
+    Device *getDevice(const std::string &name);
     int getType() const;
 
     virtual void batterySensorEnable(int samplingPeriod);
@@ -127,7 +128,7 @@ namespace webots {
     void setData(const std::string &data);
 
     // internal functions
-    static Device *getDevice(int tag);
+    static Device *getDeviceFromTag(int tag);
     static int getDeviceTypeFromTag(int tag);
     static std::string getDeviceNameFromTag(int tag);
     static int getDeviceTagFromIndex(int index);
--- a/projects/languages/python/controllers/driver/driver.py
+++ b/projects/languages/python/controllers/driver/driver.py
@@ -29,7 +29,7 @@ class Driver (Supervisor):
 
     def __init__(self):
         super(Driver, self).__init__()
-        self.emitter = self.getEmitter('emitter')
+        self.emitter = self.getDevice('emitter')
         robot = self.getFromDef('ROBOT1')
         self.translationField = robot.getField('translation')
         self.keyboard.enable(Driver.timeStep)
--- a/projects/languages/python/controllers/slave/slave.py
+++ b/projects/languages/python/controllers/slave/slave.py
@@ -43,18 +43,18 @@ class Slave (Robot):
     def __init__(self):
         super(Slave, self).__init__()
         self.mode = self.Mode.AVOIDOBSTACLES
-        self.camera = self.getCamera('camera')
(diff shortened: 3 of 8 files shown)


omichel:
Replaced by #2510.

PR #2347, opened 2020-10-15
Fix invalid buffer condition

stefaniapedrazzi:
Fix bug checking if reallocation failed introduced in #2335:

### DIFF
--- a/src/webots/sound/WbPicoTextToSpeech.cpp
+++ b/src/webots/sound/WbPicoTextToSpeech.cpp
@@ -349,7 +349,7 @@ qint16 *WbPicoTextToSpeech::generateBufferFromText(const QString &text, int *siz
         bufferSize += 1024 * 1024;  // 1 MB
         qint16 *previousBuffer = buffer;
         buffer = (qint16 *)realloc(buffer, bufferSize);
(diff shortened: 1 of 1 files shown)


src/webots/sound/WbPicoTextToSpeech.cpp
```diff
@@ -349,9 +349,8 @@ qint16 *WbPicoTextToSpeech::generateBufferFromText(const QString &text, int *siz
         bufferSize += 1024 * 1024;  // 1 MB
         qint16 *previousBuffer = buffer;
         buffer = (qint16 *)realloc(buffer, bufferSize);
-        if (!buffer)  // re-allocation failed, this is required otherwise CppCheck raises an error
+        if (!buffer) { // re-allocation failed, this is required otherwise CppCheck raises an error
```
DavidMansolino, line 352:
```suggestion
        if (!buffer) {  // re-allocation failed, this is required otherwise CppCheck raises an error
```
😉

stefaniapedrazzi: Thank you!

DavidMansolino (approved):
Thank you!

DavidMansolino (approved):

PR #2350, opened 2020-10-16
Replace ros test

DavidMansolino:
**Description**

This PR removes the ros test on travis (as it is often making the nightly fail) and add it to Github Actions (which doesn't block the nightly if failed and use a more recent version of ROS1).

It also introduced a new tag `test ros` which allows to trigger the ros test only (or in addition to other tests).

### DIFF
--- a/tests/ros.sh
+++ b/tests/ros.sh
@@ -2,6 +2,7 @@
 
 echo @@@ Compile ros webots_ros package
 source /opt/ros/$ROS_DISTRO/setup.bash
+# BASEDIR might differ from WEBOTS_HOME if we are not using Webots compiled from this repository directory (e.g. in CI).
 BASEDIR=$(dirname $(realpath $0))/..
 [ -d $BASEDIR/webots_catkin_ws ] && rm -r $BASEDIR/webots_catkin_ws
 mkdir -p $BASEDIR/webots_catkin_ws/src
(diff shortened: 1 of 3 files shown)


tests/ros.sh
```diff
@@ -2,13 +2,14 @@
 
 echo @@@ Compile ros webots_ros package
 source /opt/ros/$ROS_DISTRO/setup.bash
-[ -d $WEBOTS_HOME/webots_catkin_ws ] && rm -r $WEBOTS_HOME/webots_catkin_ws
-mkdir -p $WEBOTS_HOME/webots_catkin_ws/src
-cd $WEBOTS_HOME/webots_catkin_ws/src
+BASEDIR=$(dirname $(realpath $0))/..
```
omichel, line 5: We should maybe explain in a comment why BASEDIR is defined and why we don't use WEBOTS_HOME instead.

DavidMansolino, line 5:
```suggestion
# BASEDIR might differ from WEBOTS if we are not using Webots compiled from this repository directory (e.f. in CI).
BASEDIR=$(dirname $(realpath $0))/..
```

@omichel would this be clear enough?

omichel: Yes, thank you.

omichel (approved):
Thank you.

PR #2351, opened 2020-10-16
Add missing supervisor fields

DavidMansolino:
**Description**
The supervisor field was missing in these two PROTO files.


omichel (approved):
Thank you.

PR #2362, opened 2020-10-20
Fixed Python 3.9 support on Windows

omichel:
It was not possible to launch a python controller with Python 3.9.
This PR fixes the problem.


DavidMansolino (approved):
Thank you.

PR #2364, opened 2020-10-20
Fixed TexturedParallelepiped proto

### LINKED ISSUE #2361: Textured Parallelipiped checkboxes don't match cube mapping
Create a textured parallelipiped, select cube mapping, select the Webots-suggested cube_mapping.jpg as the texture to map.  The parallelipiped appears reasonably with the expected faces adjoining.  Now change the setting for _right_ to FALSE.

**Expected behavior:**  The blue side labelled RIGHT should be the side that disappears.  
**What actually happens:** the side labeled FRONT disappears instead.  (Similar problems affect several of the other checkboxes. 
 E.g. the side I wanted to remove was labelled BACK, but to remove it I actually had to set _left_ to FALSE.)

I'm not sure what to suggest to fix this mismatch, as any change you make may break backwards compatibility for some existing uses.  If you do change it, I guess I'd suggest changing which faces the checkboxes remove, rather than changing how the texture maps to the vertices, as it'd be annoying to have to use different cubemaps for these from what you use for textured boxes, and if the change does break backwards compatibility for someone, it's generally much easier to change a couple of boolean checkboxes than it is to make new texture maps.  Or maybe you should just clearly document the mismatch and live with it?

image

omichel:
Fixes #2361.

### DIFF
--- a/docs/reference/changelog-r2020.md
+++ b/docs/reference/changelog-r2020.md
@@ -6,7 +6,7 @@ Released on XXX YYY, 2020.
   - Enhancements
     - Added support for Python 3.9 ([#2318](https://github.com/cyberbotics/webots/pull/2079)).
   - Bug fixes
-    - Fixed cube, compact and flat texture mappings of [TexturedParallelepiped](https://cyberbotics.com/doc/guide/object-geometries#texturedparallelepiped) proto ([#2364](https://github.com/cyberbotics/webots/pull/2364)).
+    - **Fixed cube, compact and flat texture mappings of [TexturedParallelepiped](https://cyberbotics.com/doc/guide/object-geometries#texturedparallelepiped) proto** ([#2364](https://github.com/cyberbotics/webots/pull/2364)).
     - Fixed supervisor label color change which was not working if the text remained unchanged ([#2357](https://github.com/cyberbotics/webots/pull/2357)).
     - Fixed recording of movies which was broken when the [WorldInfo](worldinfo.md).`basicTimeStep` was greater than 40 ([#2268](https://github.com/cyberbotics/webots/pull/2268)).
     - Fixed handling of <kbd>Ctrl</kbd> key on macOS from the [Keyboard](keyboard.md) API ([#2265](https://github.com/cyberbotics/webots/pull/2265)).
--- a/docs/reference/changelog-r2020.md
+++ b/docs/reference/changelog-r2020.md
@@ -6,6 +6,7 @@ Released on XXX YYY, 2020.
   - Enhancements
     - Added support for Python 3.9 ([#2318](https://github.com/cyberbotics/webots/pull/2079)).
   - Bug fixes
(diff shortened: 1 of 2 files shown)


DavidMansolino:
:eyes:

DavidMansolino (changes requested):
It works now as expected, but as the user said, this clearly breaks the compatibility.
Before:
tmp2-2

Now: 
tmp2

This should therefore at least be written in bold in the changelog (I would even have targeted develop).

docs/reference/changelog-r2020.md
```diff
@@ -6,6 +6,7 @@ Released on XXX YYY, 2020.
   - Enhancements
     - Added support for Python 3.9 ([#2318](https://github.com/cyberbotics/webots/pull/2079)).
   - Bug fixes
+    - **Fixed cube, compact and flat texture mappings of [TexturedParallelepiped](https://cyberbotics.com/doc/guide/object-geometries#texturedparallelepiped) proto** ([#2364](https://github.com/cyberbotics/webots/pull/2364)).
```
DavidMansolino, line 9:
```suggestion
    - **Fixed cube, compact and flat texture mappings of [TexturedParallelepiped](object-geometries.md#texturedparallelepiped) proto** ([#2364](https://github.com/cyberbotics/webots/pull/2364)).
```

omichel:
I updated the change log with a bold sentence. However, I believe it is clearly a bug fix and as many bug fixes, it indeed may break the backwards compatibility. However, this primitive doesn't seem to be very much used: none of our PROTO or world file use it except for the `textured_proto_shape.wbt` sample world. So, the risk of breaking user simulations is low and in case it would happen, the fix is very simple. So, I would leave this bug fix on the master branch.

docs/reference/changelog-r2020.md
```diff
@@ -6,6 +6,7 @@ Released on XXX YYY, 2020.
   - Enhancements
     - Added support for Python 3.9 ([#2318](https://github.com/cyberbotics/webots/pull/2079)).
   - Bug fixes
+    - **Fixed cube, compact and flat texture mappings of [TexturedParallelepiped](object-geometries.md#texturedparallelepiped) proto** ([#2364](https://github.com/cyberbotics/webots/pull/2364)).
```
DavidMansolino, line 9:
```suggestion
    - **Fixed cube, compact and flat texture mappings of [TexturedParallelepiped](../guide/object-geometries.md#texturedparallelepiped) proto** ([#2364](https://github.com/cyberbotics/webots/pull/2364)).
```
Sorry for this.

DavidMansolino:
> However, this primitive doesn't seem to be very much used: none of our PROTO or world file use it except for the textured_proto_shape.wbt

I am not sure about this, it is used directly in 3 PROTO files (and even more indirectly), and indirectly in a lot of worlds.

But that's fine, most of the case the texture is face independent.

DavidMansolino (approved):
Thank you.

omichel:
Sorry, the meant the problematic texture mappings (cube, compact and flat) seems to be never used in our PROTO files.

PR #2366, opened 2020-10-21
Fix field operations order in Webots

### LINKED ISSUE #2287: Supervisor alterations to extrusion spines can crash webots
You can use a supervisor (a) to change existing spine points in an extrusion proto, or (b) to insert an additional spine point at the end of the list, but doing both during the same timestep causes Webots to crash to desktop.

I get this bug with the following simple Python supervisor controller and a simple 2-point extrusion (with its scale reduced to [0.4 0.4] to make it appear more noodle-like).

```
robot=Supervisor() 

robot.step(timestep) # so hitting play-one-frame from beginning will pause at next robot.step when it's uncommented

spine = robot.getFromDef("TESTX").getField("spine")
spine.setMFVec3f( 0 , [0,0,0] )
spine.setMFVec3f( 1 , [0,0.5,0.5] )

# robot.step(timestep) # uncommenting this will fix the crash

spine.insertMFVec3f( -1 , [0,1,0] )
```

Uncommenting the later call to robot.step fixes the crash, as does commenting out either the setMF* commands above it or the insertMF* command below it, which suggests that the problem has something to with too many changes happening to the extrusion in the very same timestep, since it seems to handle these changes fine in different time steps.  Sometimes it crashes too quickly too see this, but sometimes it is briefly visually apparent that the extrusion has successfully updated itself to the new 3 point shape before Webots closes, so it looks like all of these *MFVec3f commands are executing and the proto is (at least mostly) finishing its procedural update in light of all these changes, so I'm not sure what after that should cause a crash.  Maybe the setMF* and insertMF* commands each try to trigger procedural updates to the proto and the first somehow blocks the second from proceeding?

Here's the simple extrusion I was using.  My world had more stuff in it too, but I don't think that should matter.

```
Shape {
  appearance MetalStainlessSteelCable {
  }
  geometry DEF TESTX Extrusion {
    scale [
      0.04 0.04
    ]
    splineSubdivision 0
  }
}
```
**Expected behavior**
Not to crash to desktop -- perhaps an error message if there was something it didn't like, but it works fine when these ops are spread over multiple timesteps, and it even briefly shows up fine with the single timestep before crashing, so I can't imagine what error it should have to complain about.  (Note that the two altered MF entries are indices 0 and 1, so there's no apparent reason why changes to those should clash with inserting a third entry after them at index 2.)

**System**
 - Graphics Card: NVIDIA something fancyish

stefaniapedrazzi:
Fix #2287: fix field operations order not respected in Webots.
The reason of the crash was that the INSERT operation was applied immediately and before the SET operations that were queued. As a consequence, when applied, the SET operation requests were referring to an old field pointer.

### DIFF
--- a/docs/reference/changelog-r2020.md
+++ b/docs/reference/changelog-r2020.md
@@ -7,6 +7,7 @@ Released on XXX YYY, 2020.
     - Added support for Python 3.9 ([#2318](https://github.com/cyberbotics/webots/pull/2079)).
   - Bug fixes
     - **Fixed cube, compact and flat texture mappings of [TexturedParallelepiped](../guide/object-geometries.md#texturedparallelepiped) proto** ([#2364](https://github.com/cyberbotics/webots/pull/2364)).
+    - Fixed crash in the Supervisor API occurring when [setting](supervisor.md#wb_supervisor_field_set_mf_bool) an item of a multiple field just before [inserting](supervisor.md#wb_supervisor_field_insert_mf_bool) or [removing](supervisor.md#wb_supervisor_field_remove_mf) an item in the same field during the same controller step ([#2366](https://github.com/cyberbotics/webots/pull/2366)).
     - Fixed supervisor label color change which was not working if the text remained unchanged ([#2357](https://github.com/cyberbotics/webots/pull/2357)).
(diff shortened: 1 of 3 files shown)


omichel (approved):
Thank you.

Issue #2292, opened 2020-10-02
insertMF causes reversion of other values


Justin-Fisher:
If you use field.setMFColor to set an entry in an LED's MF color field, the new value is immediately apparent.  But if you then use field.setMFColor(-1, newcolor) to append a new entry at the end of the MF field, this unexpectedly causes the recently set entry to to revert back to its former value.

**Steps to Reproduce**
Set up an LED with DEFname TESTLED with gradual FALSE, and the color field having the following three entries: 0 0 0, 1 0 0, 0 1 0 (i.e. black, red, green).

Run the following Python controller:
```python
from controller import Supervisor
robot = Supervisor()

colorfield = robot.getFromDef("TESTLED").getField("color")
colorfield.setMFColor(2,[0,0,1]) # overwrite green with blue
print("Before appending, color[2] =", colorfield.getMFColor(2) ) # prints blue [0, 0, 1] as expected
colorfield.insertMFColor(-1,[1,1,0]) # append yellow
print("After appending, color[2] =", colorfield.getMFColor(2) ) # prints green [0,1,0] for no good reason!
```
**Expected behavior**
The final line should still print blue [0, 0, 1] since this entry was already blue two lines up, and there's no reason that the intervening line, which simply appended yellow [1,1,0] to the end of the list, should change blue back to green.  Yellow and blue are supposed to make green in paint -- not in neighboring list elements!

It's possible that this bug is related to crash #2287 which also involved unwanted behavior when setMF and insertMF are used in the same timestep, though fortunately with color LED's we're just getting an unpredicted outcome, not the outright crash that arose in that case.

**System**
 - Operating System: Windows 10
 - Graphics Card: [e.g., NVIDIA something fancyish

Justin-Fisher:
I guess maybe the surprising part here should actually be that the first print prints blue [0, 0, 1].  Webots docs say that setMF statements will be queued and executed at the end of the time step, which fits with getting a not-yet-updated reading from the second print statement.  But it doesn't fit with the first print statement showing an updated value already.

At any rate, it looks like the expected blue value does stick in the scene tree afterwards, so the only blips have to do with the surprisingly inconsistent getMF values.  (I had initially stumbled across this in a case where I had first appended the green to the list even earlier in the timestep.  In that case, the value that stuck in the scene tree was green, which seemed wrong, but it might be partly explained if the various queued set and insert effects didn't pop off in quite the expected order or something).

stefaniapedrazzi:
[...]

PR #2375, opened 2020-10-22
Fix Supervisor API field operations

stefaniapedrazzi:
Address #2292.

The `wb_supervisor_field_get_*` functions return inconsistent values during the same controller step if a previous `wb_supervisor_field_set_*` instructions is called:
* it returns the correct updated value before the set instruction(s) are sent to Webots
* it returns the old value after the set instruction(s) are sent to Webots and are no longer stored on the controller side

It is well [documented](https://www.cyberbotics.com/doc/reference/supervisor#wb_supervisor_field_set_sf_bool) that values could be wrong, but it is very unclear for users why sometimes they are correct and sometimes not.

In this PR I tried to fix this behavior by keeping as long as possible the set operation list on the controller side and send it to Webots only if another field operation is requested ensuring that the set operations are applied before the other field operation.

Improved documentation:
https://www.cyberbotics.com/doc/reference/supervisor?version=fix-supervisor-field-operations#wb_supervisor_field_set_sf_bool


stefaniapedrazzi:
I still have to add an entry in the ChangeLog.

omichel (changes requested):

stefaniapedrazzi:
There are issues retrieving packages in the appveyor test, but it doesn't seem related to this PR.
https://ci.appveyor.com/project/omichel/webots/builds/35952083

omichel (approved):
Thank you.
(yes, it seems some appveyor servers are currently down)

Issue #2365, opened 2020-10-20
This texture doesn't play nice with infrared sensors


Justin-Fisher:
When I use the following texture as the basemap for a texturedbox (set to compact mapping), infrared sensors cannot detect the difference between high-red colors (red/yellow) versus low-red colors (blue/black) even though it has no problem doing so on many other quite similar textures.  In fact, when I simply switch with textured box has which texture as its basemap, that switches which textured box the infrared sensor messes up on, which tells me that it must be something about this texture.  But I have no idea what!

(I'm not sure if it matters, but I've been trying this on textured boxes with some faces removed.)

![upper_block](https://user-images.githubusercontent.com/19636109/96641318-29ed5900-12ea-11eb-8b63-a016b01bb85f.jpg)

**Expected behavior**
Pointing an infrared sensor with default settings at this should produce 1000 when aimed at blue or black, as it does for lots of other textures.  Instead it produces basically the same reading as it produces when aimed at red or yellow (for me that's usually 300-something, depending upon just how close the sensor is to the surface).

**System**
 - Operating System: Windows 10
 - Graphics Card: NVIDIA something fancyish

DavidMansolino:
I have made some tests with this texture and the texturedbox, to me it looks all correct.

Note the reflection factor used for infra-red distance-sensors:
> f = 0.2 + 0.8 * red_level * (1 - 0.5 * roughness) * (1 - 0.5 * occlusion)

In our case with the texturedbox the roughness is always 1 and the occlusion is always 0, so we can simplify like this: 
> f = 0.2 + 0.4 * red_level

Now I have tried with a distance-sensor with the default lookup table and at a distance of 0.1 meter of the box.

On red:
> red_level = 1
> f = 0.2 + 0.4 * red_level = 0.6                
> used distance = 0.1  / 0.6 = 0.1666..

The distance-sensor returns me 166.666... in the simulation, this looks correct.

On blue:
> red_level = 0
> f = 0.2 + 0.4 * red_level = 0.2
> used distance = 0.1  / 0.2 = 0.5

The distance-sensor returns me 500 in the simulation, this looks correct too.

If you have different values can you please share at which distance and pointing where exactly on the texture?

Justin-Fisher:
[...]

PR #2378, opened 2020-10-22
Fix Distance-Sensors Reading with Cached Textures

DavidMansolino:
**Description**
Fix Distance-Sensors Reading with Cached Textures

**Related Issues**
This pull-request fixes issue #2365


omichel:
The test suite is failing with distance sensor readings on textures...

DavidMansolino:
> The test suite is failing with distance sensor readings on textures...

The test was indeed previously affected by the bug, the `WbImageTexture::pickColor` was fallbacking to (1.0, 1.0, 1.0), it is now working correctly and the value returned by the distance-sensors therefore changed.
I updated the test accordingly.

DavidMansolino:
The test is now passing.

omichel (approved):
Thank you.

PR #2381, opened 2020-10-23
Update modeling.md

omichel:
Almost a copy/past of one of the very useful answers of @DavidMansolino on Discord.


DavidMansolino (approved):
Thank you, good idea to include it in the doc.

PR #2383, opened 2020-10-23
change scene_render to correspond to prototype

BenjaminDeleze:
**Description**
default.cpp (a demo program using wren) was not compiling because it misses one argument (the culling)

I added this last argument


omichel (approved):
Thank you!

Issue #2386, opened 2020-10-25
Movie recording is cancelled if simulation is reset before movie starts


PeterJCLaw:
**Describe the Bug**

If I reset the simulation before starting the movie capture then the capture is always cancelled.

**Steps to Reproduce**

I'm running `webots` with `webots --batch worlds/Arena.wbt`, running without `--batch` results in the message being a dialog box rather than a log message, but has the same effect.

Controller:
``` python
from controller import Supervisor
supervisor = Supervisor()

# I also pause the simulation here, though that doesn't appear to affect the bug

supervisor.simulationReset()

# I also restart all the other (non-supervisor) controllers at this point too,
# though that doesn't appear to affect the bug

supervisor.movieStartRecording(
    'that.mp4', width=640, height=480, quality=100,
    codec=0, acceleration=1, caption=False,
)
# Short step length for easy reproduce
supervisor.step(320)
supervisor.movieStopRecording()
supervisor.simulationSetMode(Supervisor.SIMULATION_MODE_PAUSE)
```

This results in a log line of `INFO: Video creation canceled.`.
(aside: there's a typo in "canceled")

Commenting out the `simulationReset` call successfully captures the video.

**Expected behavior**

I would like this to successfully record the video.

**System**
 - Operating System: Ubuntu 18.04.5 LTS

**Additional context**

My goal here is to be able to capture a video of the whole simulation, from the very start of all the (non-supervisor) controllers, hence the desire to pause then reset the simulation before starting the movie.
(I've also recently found the `--mode=pause` argument, which may provide an alternative for my case, though I think the above is still a bug)

PeterJCLaw:
> I've also recently found the `--mode=pause` argument, which may provide an alternative for my case

Turns out that this doesn't help me because of course when starting paused the supervisor controller doesn't get started upfront.

omichel:
That is clearly a bug.
However, if you add a call to `supervisor.step(32)` just after the `supervisor.simulationReset()`, it fixes the problem.

omichel:
BTW, "canceled" is correct in American English, which is what we use in Webots.

stefaniapedrazzi:
The `Supervisor.simulationReset()` method is applied only at the end of the step (see [documentation](https://www.cyberbotics.com/doc/reference/supervisor?tab-language=python#wb_supervisor_simulation_reset)) to guarantee that the simulation works correctly even if multiple Robot and Supervisor controllers are running.
[...]

PR #2406, opened 2020-10-28
Fix movie recording behavior on simulation reset from Supervisor controller

stefaniapedrazzi:
Address #2386: cancel supervisor movie recording on reset only if controllers are restarted (i.e. reset from the UI).


stefaniapedrazzi:
Additionally it is possible to start simultaneously different movie recordings from more than one controller.
This happens in particular after a controller is restarted, because the movie status is sent only when changes and not when the controller is initialized, i.e. if a controller starts when the movie recording is already running it is notified about the movie status.
But the new movie configuration currently overwrites the previous ones without any warnings.

I would adjust the code so that the movie status is always updated when a controller starts and force the user to stop the previous movie recording from the supervisor in order to be able start a new one.

DavidMansolino:
👀

DavidMansolino (approved):
Thank you, it works now perfectly.

PR #2407, opened 2020-10-28
Fix Github Action macOS build

DavidMansolino:
**Description**
This link with python 3.8 from brew is causing issues when installing python 3.9 from brew (https://github.com/cyberbotics/webots/runs/1321185852#step:3:331)

### DIFF
--- a/.github/workflows/test_suite_mac.yml
+++ b/.github/workflows/test_suite_mac.yml
@@ -38,7 +38,7 @@ jobs:
         brew update
         brew upgrade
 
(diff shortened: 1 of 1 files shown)


DavidMansolino:
It seems the bug is affecting Travis too: https://travis-ci.com/github/cyberbotics/webots/jobs/417037757#L6583

DavidMansolino:
Travis part will be fixed in another PR (#2414).

omichel (approved):
Very nice.

PR #2408, opened 2020-10-28
Attempt to enable source test on Windows Github Action

DavidMansolino:
**Description**
This is just a simple test.


DavidMansolino:
With this PR we will be able to remove the per commit Appveyor test (which is often problematic because failing on first commit) and keep only the nightly for now.

I have used the Chocolatey package manager for this workflow, for the workflow compiling Webots we will for sure need to use MSYS, but for this one, using Chocolatey has several advantages such as it is already available (no need to download and install) on Github action, and it supports installing specific version of packages (which is safer in CI to avoid that an update out of our control suddenly breaks the CI).

omichel (dismissed):
Thank you.
We should now remove the corresponding appveyor test.

omichel:
Shall we also remove `.appveyor.yml`, `test/sources/generate_diff.py` and `test/sources/parent_branch.sh`?

DavidMansolino:
> Shall we also remove `.appveyor.yml`, `test/sources/generate_diff.py` and `test/sources/parent_branch.sh`?

Yes sure, well spotted!

DavidMansolino:
> We should now remove the corresponding appveyor test.

I will remove them as soon as this PR is merged and then master is merged into develop.

omichel (approved):
Thank you.

PR #2423, opened 2020-11-02
Modify shaders to have float in the form 1.0 and not just 1

BenjaminDeleze:
**Description**
WebGL does not accept shaders where a float has the form 1 or 0 (instead of 1.0 and 0.0). 

I changed the float in some shaders in order for them to be compatible with WebGL


omichel:
That seems good and safe. I am just running the test suite to ensure this doesn't break it (but it would be surprising).

omichel (approved):
The test suite passed successfully.
Thank you!

PR #2424, opened 2020-11-02
Get quaternion from `InertialUnit`

lukicdarkoo:
This PR adds support to the `InertialUnit` node that allows retrieval of quaternions. This allows us to avoid gimbal-lock and to be closer to the real sensor implementation. It also fixes orientation in ROS.

**Physical Sensor Example**
I used an MPU6050 sensor (driver implementation) as an example:
- quaternions are directly returned from the sensor: https://github.com/ElectronicCats/mpu6050/blob/845ee3978b6554967e510d8dd5f74c2c764908e3/src/MPU6050_6Axis_MotionApps20.h#L438
- roll/pitch/yaw are computed from the quaternions: https://github.com/ElectronicCats/mpu6050/blob/845ee3978b6554967e510d8dd5f74c2c764908e3/src/MPU6050_6Axis_MotionApps20.h#L547
- gimbal-lock issue: https://github.com/ElectronicCats/mpu6050/blob/845ee3978b6554967e510d8dd5f74c2c764908e3/examples/MPU6050_DMP6_ESPWiFi/MPU6050_DMP6_ESPWiFi.ino#L103

**Non-related fixes**
- PointStamped is published by GPS instead of Point. This also fixes ROS tests.

**Documentation**
https://cyberbotics.com/doc/reference/inertialunit?version=feature-inertial-unit-quaternion

**ROS Objective**
The ROS layer should be changed, so RVIz and Webots orientation match if the coordinate system is ENU:
image

**Related Issues**
https://github.com/cyberbotics/webots_ros/issues/37
https://github.com/cyberbotics/webots_ros/pull/38/

**Tasks**
  - [x] Basic implementation
  - [x] Decide on how the lookup table should be implemented
  - [x] Add getter for `noise` parameter
  - [x] Python/Java/Matlab
  - [x] Documentation
  - [x] Document gimbal-lock case
  - [x] Changelog


tests/api/controllers/inertial_unit/inertial_unit.c
```diff
@@ -26,13 +26,6 @@ int main(int argc, char **argv) {
     {0, 0, 0}, {0, 0, -1.5708}, {0, 0, 1.5708}, {1.5708, 0, 0}, {-1.5708, 0, 0}, {0, 0.7854, 0}, {0, -0.7854, 0},
   };
 
-  int lookup_table_size = wb_inertial_unit_get_lookup_table_size(inertial_unit);
-  ts_assert_double_equal(lookup_table_size, 2, "Lookup table size returned is wrong (%d instead of 2)", lookup_table_size);
-  const double *lookup_table = wb_inertial_unit_get_lookup_table(inertial_unit);
-  ts_assert_double_equal(lookup_table[3], 1000, "Lookup table (index 3) returned is wrong (%lf instead of 1000)",
-                         lookup_table[3]);
-  ts_assert_double_equal(lookup_table[5], 0, "Lookup table (index 5) returned is wrong (%lf instead of 0)", lookup_table[5]);
-
```
omichel, line 35: Instead of that, we should have added a test that verify that the noise reading is correct (we would also need to set a very low noise in the world file to avoid testing against 0).

omichel (changes requested):
Thank you. It looks very good to me. A few minor change requests.

src/webots/maths/WbAxisAngle.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+//
+// Description: Axis-angle representation
+//
+
+class WbAxisAngle {
+public:
+  WbAxisAngle(double x, double y, double z, double angle) : mAxis(WbVector3(x, y, z)), mAngle(angle){};
+
+  WbVector3 &axis() { return mAxis; };
```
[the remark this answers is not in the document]
lukicdarkoo: I don't think can add `const` here because `mAxis` is returned as a reference. That means we can alternate `mAxis`, effectively changing `WbAxisAngle`?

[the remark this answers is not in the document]
omichel: OK, I was not sure about it. So, it's fine this way. Other wise we may return a const WbVector3:
`const WbVector3 &axis() const { return mAxis; };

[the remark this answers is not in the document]
lukicdarkoo: I am modifying the `axis()` output here:
https://github.com/cyberbotics/webots/pull/2424/files#diff-f711b25b459281533b960b951ada0d487fdcd2200884b1e6f495da1209fe7ccbR149

The only solution would be to wrap the methods, `setX`, `setY`, `setZ` and `normalize`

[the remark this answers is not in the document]
omichel: OK, then keep it as it is. It's fine.

omichel (approved):
Thank you!

docs/reference/changelog-r2021.md
```diff
@@ -13,6 +13,7 @@ Released on December XXth, 2020.
     - Added a new [`wb_supervisor_node_get_contact_point_node`](supervisor.md#wb_supervisor_node_get_contact_point_node) [Supervisor](supervisor.md) API function to get the node reference associated to a given contact point ([#2228](https://github.com/cyberbotics/webots/pull/2228)).
     - Added two new functions to the [Camera](camera.md) API called [`wb_camera_get_exposure`](camera.md#wb_camera_get_exposure) and [`wb_camera_set_exposure`](camera.md#wb_camera_set_exposure) to retrieve and change the [Camera](camera.md) exposure ([#2363](https://github.com/cyberbotics/webots/pull/2363)).
     - Added a new functionality in the [Recognition](recognition.md) node and [Camera](camera.md) API for generating segmented ground truth images ([#2199](https://github.com/cyberbotics/webots/pull/2199)).
+    - **Added a new [`wb_inertial_unit_get_quaternion`](inertialunit.md#wb_inertial_unit_get_quaternion) [InertialUnit](inertialunit.md) API function to get the orientation measurement represented as a quaternion ([#2424](https://github.com/cyberbotics/webots/pull/2424)).**
```
stefaniapedrazzi, line 16: Why this entry is written in bold?
Does it break compatibility with previous version? if not it should not be in bold to be consistent with the rest of the ChangeLog.

omichel: Yes, and we should also mention here in bold that a function was removed and another one was added (lookup table and noise).

resources/nodes/InertialUnit.wrl
```diff
@@ -21,7 +21,7 @@ InertialUnit {
   field     SFFloat    radarCrossSection   0.0              # radar cross section of this solid
   field     MFColor    recognitionColors   []               # colors returned for this Solid by Cameras with a Recognition node
   #fields specific to the InertialUnit node:
-  field     MFVec3f    lookupTable     []          # empty: no interpolation
```
DavidMansolino, line 24: Usually, the field is actually kept in the '*.wrl' file (at the end) with the 'deprecatedField' keyword instead of 'field' so that a deprecation warning is automatically generated if needed. See for example: https://github.com/cyberbotics/webots/blob/master/resources/nodes/Camera.wrl#L46

docs/reference/changelog-r2021.md
DavidMansolino: It seems this comment has still not been solved, and in addition, the removal of the ``lookupTable`` field should be mentioned too.

omichel:
I am taking care of fixing the remaining issues here.

PR #2425, opened 2020-11-02
Fix padding of vec3 for std140

BenjaminDeleze:
**Description**
Wren uses the size of the struct Lights to initialize the corresponding uniform buffer.
In the shaders, we redeclare this struct with the std140 layer.

However, std140 always rounds up to size of vec4 (16 bytes) but we have a vec3 (12 bytes) in the struct. (https://www.khronos.org/opengl/wiki/Interface_Block_(GLSL)#Memory_layout)

So the sizeof is short of 4 bytes compared to the size expected by GLSL.


omichel (approved):
Nice fix, thank you.
Please wait until the test suite passes successfully before merging this PR.

omichel (approved):
Test suite passed.

PR #2429, opened 2020-11-04
Improve Python guide

### LINKED ISSUE #2428: Running python controller with Anaconda environment
Hi,

I'm new to using webots (R2020 b revision 1)but I'm trying to use it with a Mac (10.15.7) with python installed in an Anaconda environment.  I've installed python3.8 as described in the documentation (via Python website).  So I have the Python dynlib in /Library/Frameworks/Python.framework/Versions/3.8 (which seems necessary and isn't included in anaconda).  However, I'm not sure how I can use libraries I've installed in my conda environment to run my controller.  Currently I get an error (see below).  Help would be appreciated.

https://cyberbotics.com/doc/guide/using-python?tab-os=macos#macos-installation

Current documentation does not show how to install when using an anaconda version of python.  If Anaconda is not supported then this should also be pointed out.

Currently when trying to use python from within a conda environment I get an error:

```
/Users/dwicke/anaconda3/envs/webots/bin/python /Users/dwicke/Documents/webots/my_first_simulation/controllers/epuck_go_forward/epuck_go_forward.py

Process finished with exit code 139 (interrupted by signal 11: SIGSEGV)
```

**Describe how you would fix it**
Add documentation if I new how to do so.

From my attempts I've been unsuccessful at solving this.  If anyone knows how, I am very interested in learning how.  Let me know if you need additional info.

Thanks,
Drew

omichel:
Fixes #2428.

### DIFF
--- a/docs/guide/using-python.md
+++ b/docs/guide/using-python.md
@@ -20,7 +20,7 @@ Webots starts Python using the standard `python` command line.
 As a consequence, it executes the first `python` binary found in the current `PATH`.
 If you want to use a different version of Python, please install it if needed and configure your environment so that it becomes the default `python` version when called from the command line in a terminal.
 Alternatively, you can change the default Python command from the Webots Preferences in the General tab.
-If you set it for example to `python3.8` instead of `python`, this version of python will be used by default, if available from the command line.
+If you set it for example to `python3.8` instead of `python`, this version of Python will be used by default, if available from the command line.
 It is also possible to set a different version of Python for each robot controller by editing the `[python]` section of the `runtime.ini` file in each robot controller directory and setting the `COMMAND` value to `python3`, `python3.8` or `python2.7`, etc.
 If specified in the `runtime.ini` file of a controller, this Python command will be executed instead of the default one to launch this controller.
 On Linux and macOS, it is also possible to override this value by setting a standard Python shebang header line in your main python controller file, for example:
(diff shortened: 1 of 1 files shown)


docs/guide/using-python.md
```diff
@@ -20,13 +20,13 @@ Webots starts Python using the standard `python` command line.
 As a consequence, it executes the first `python` binary found in the current `PATH`.
 If you want to use a different version of Python, please install it if needed and configure your environment so that it becomes the default `python` version when called from the command line in a terminal.
 Alternatively, you can change the default Python command from the Webots Preferences in the General tab.
-If you set it for example to `python3.7` instead of `python`, this version of python will be used by default, if available from the command line.
-It is also possible to set a different version of Python for each robot controller by editing the `[python]` section of the `runtime.ini` file in each robot controller directory and setting the `COMMAND` value to `python3`, `python3.7` or `python2.7`, etc.
+If you set it for example to `python3.8` instead of `python`, this version of python will be used by default, if available from the command line.
```
stefaniapedrazzi, line 23:
```suggestion
If you set it for example to `python3.8` instead of `python`, this version of Python will be used by default, if available from the command line.
```

stefaniapedrazzi (changes requested):
Thank you

stefaniapedrazzi (approved):

PR #2422, opened 2020-10-30
Commercial to Open-source

lukicdarkoo:
I am not sure about the capitalization, but it seems fine:
https://english.stackexchange.com/questions/157178/capitals-or-not-for-open-source
https://en.wikipedia.org/wiki/Open-source_software

PR #2430, opened 2020-11-04
Fix CI `tests_sources.yml` workflow

lukicdarkoo:
Addresses the following issue:
https://github.com/cyberbotics/webots/pull/2422#issuecomment-719914230

### DIFF
--- a/.github/workflows/tests_doc.yml
+++ b/.github/workflows/tests_doc.yml
@@ -10,7 +10,7 @@ defaults:
     shell: bash
 
 jobs:
-  workflow_skipper:
+  workflow-skipper:
     runs-on: ubuntu-latest
     outputs:
       should_skip: ${{ steps.skip_check.outputs.should_skip }}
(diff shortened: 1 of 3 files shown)


lukicdarkoo:
@omichel It can put the `test-doc` job as mandatory as it should be skipped automatically now.

omichel:
I just set the `test-doc` as Required. However, it's not started...

omichel (approved):
Very good. Thank you.

.github/workflows/tests_doc.yml
```diff
@@ -13,18 +10,33 @@ defaults:
...
     steps:
     - uses: actions/checkout@v2
+      if: needs.job-skipper.outputs.should_skip != 'true'
     - name: Set up Python 3.9
       uses: actions/setup-python@v2
+      if: needs.job-skipper.outputs.should_skip != 'true'
       with:
         python-version: 3.9
     - name: Tests
+      if: needs.job-skipper.outputs.should_skip != 'true'
```
DavidMansolino, line 39: Just for the note, it should be possible to move this condition at the job level (i.e. just after the ``needs: job-skipper`` line) to avoid duplicating it and starting this job at all during the run.

.github/workflows/tests_sources.yml
```diff
@@ -13,14 +11,26 @@ defaults:
...
+        with:
+          github_token: "${{ secrets.GITHUB_TOKEN }}"
+          paths_ignore: '["docs/**"]'
+
   cleanup-runs:
     runs-on: ubuntu-latest
     steps:
     - uses: rokroskar/workflow-run-cleanup-action@master
       env:
         GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
```
DavidMansolino, line 30: Is this job still useful? Isn't ``job-skipper`` canceling previous workflows too?

.github/workflows/tests_doc.yml
lukicdarkoo: Nice tip, I didn't know about that

.github/workflows/tests_sources.yml
lukicdarkoo: I am not sure about that, we should test it, but it is possible:
> Skip concurrent workflow-runs: If the same workflow is unnecessarily triggered twice, then one of the workflow-runs will be skipped. For example, this can happen when you push a tag right after pushing a commit.

.github/workflows/tests_doc.yml
DavidMansolino: Basically, you can put the 'if' key almost everywhere, but not all the conditions are working everywhere (for example, test based on 'matrix.os' doesn't work at the job level, but only at the step one).

.github/workflows/tests_sources.yml
DavidMansolino: The 'Cancel outdated workflow-runs' feature might also allow to remove the cleanup job: https://github.com/fkirc/skip-duplicate-actions#cancel-outdated-workflow-runs

Issue #2400, opened 2020-10-28
Enable recognition and segmentation from the default robot window


stefaniapedrazzi:
Add checkboxes to enable/disable the recognition and segmentation from the default robot-window.

PR #2431, opened 2020-11-04
Add recognition and segmentation functionalities to default robot window

stefaniapedrazzi:
Address #2400: add checkboxed in the Camera tab of the default robot window to enabled/disable recognition and segmentation functionalities.
These checkboxes are displayed only if the `Recognition` node is available, respectively `Recognition.segmentation` field is set to TRUE.

Screenshot from 2020-11-04 13-03-40

Tasks:
* [x] add Camera API function `wb_camera_recognition_is_segmentation_enabled` to check if the segmentation has been enabled from the controller or not: this is needed to update the checkbox in the robot window
* [x] add checkboxes and related code to show the recognition and segmentation status and update it when changed from the  controller
* [x] fix internal status of segmentation enabled/disabled in the libController that was not matching Webots after disabling the camera recognition functionality

The `wb_camera_recognition_is_segmentation_enabled` C API function is needed internally.
We can decide if we want to make it public, documented and available on all the APIs or if we want to keep it private and move it in a `camera_private.h` header file.


stefaniapedrazzi:
:+1:

omichel:
If I disable segmentation from the checkbox in the robot window, I am getting lots of errors in the Webots console:
```
Error: wb_camera_recognition_get_segmentation_image(): segmentation is disabled! Please use: wb_camera_recognition_enable_segmentation().
Error: wb_display_image_new(): 'data' argument is NULL.
Error: wb_display_image_paste(): invalid WbImageRef argument.
Error: wb_display_image_delete(): invalid WbImageRef.
```

stefaniapedrazzi:
Yes, these errors are normal because it warns that you are trying to retrieve a value for a disabled device.
You will get the same errors in the console for any device that is used in a controller if you disable it from the robot window.

stefaniapedrazzi:
I can add a check and remove the Display errors but the `wb_camera_recognition_get_segmentation_image` one will remain.

omichel:
Can't we use `wb_camera_recognition_is_segmentation_enabled` to avoid getting the segmentation image if disabled and hence get rid of all warnings?

stefaniapedrazzi:
Yes, if we decide to document it and make it available on all the APIs.

omichel:
Oh, I see, it's not documented... Then, let's just take care of the display warning.

omichel:
It looks bad to me that these functions (like `wb_camera_recognition_is_segmentation_enabled`) used in `default_robot_window.c` are not public and hence not available for use in a user robot window (which could be inspired from the default robot window). If you agree, I will open an issue about it.

stefaniapedrazzi:
New tasks:
* [ ] document the `wb_camera_recognition_is_segmentation_enabled` and implement it in the other programming languages APIs
* [ ] check that segmentation image is generated even if camera is disabled
* [ ] open an issue to display results of recognition and/or segmentation in the robot window

stefaniapedrazzi:
I could finally fix the rendering issue due to the use of uninitialized GLuint variables.

omichel:
Is that normal that the display is transparent, showing sometimes some very light shadows of the segmented image?
camera_segmentation

stefaniapedrazzi:
> Is that normal that the display is transparent, showing sometimes some very light shadows of the segmented image?

No. Meanwhile I fixed the Display format (#2452) and I didn't update this PR.
I will fix it.

omichel:
It works nicely now, but it would be better to have a better default location for the display overlay as it is hidden behind the camera overlay and we have to move the camera overlay to see it.

stefaniapedrazzi:
In the project file, it is already specified that the Display overlay should be located at the bottom left corner (and camera at the top left corner):
https://github.com/cyberbotics/webots/blob/073197cb44759aff409a2d504c77e1d7ab6fc88f/projects/samples/devices/worlds/.camera_segmentation.wbproj#L12

Do you have some local modifications at the WBPROJ file?
Otherwise there are maybe issues saving/restoring the perspectives.

omichel (approved):
Yes, it seems my `.wbproj` file was not up-to-date. Sorry.
Now everything looks fine.
Thank you.

PR #2433, opened 2020-11-04
Fix the type of GL_R16f, GLRGB16f and GL_RGBA16f in GlFormatParams

BenjaminDeleze:
**Description**
The type corresponding to WR_TEXTURE_INTERNAL_FORMAT_R16F, WR_TEXTURE_INTERNAL_FORMAT_RGB16F and WR_TEXTURE_INTERNAL_FORMAT_RGBA16F is currently GL_UNSIGNED_BYTE.

However, according to the documentation of opengl (https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glTexImage2D.xhtml) and webgl (https://www.khronos.org/registry/webgl/specs/latest/2.0/#TEXTURE_TYPES_FORMATS_FROM_DOM_ELEMENTS_TABLE) it should be GL_HALF_FLOAT (or GL_FLOAT).

Webgl does not work if the type is set to GL_UNSIGNED_BYTE.


omichel (approved):
Thank you!

PR #2437, opened 2020-11-05
Minor simplification for Cylinders and Capsules

omichel:
That may lead to a minor performance improvement (not measured).


lukicdarkoo:
It looks like there are fewer instructions now. Do you want me to benchmark it?

omichel:
No need to benchmark it, I don't expect any significant performance benefit.

lukicdarkoo (approved):
Thanks

PR #2439, opened 2020-11-06
Improve stability and performance of the IK example

lukicdarkoo:
It should be around 6 times faster now. We could run it even faster, but we need to modify the `ikpy` library.

---

The new CI skipping mechanism didn't work properly for the first commit in a PR (the first commit is always skipped). Adding `concurrent_skipping: 'false'` should fix it.

### DIFF
--- a/.github/workflows/tests_doc.yml
+++ b/.github/workflows/tests_doc.yml
@@ -20,6 +20,7 @@ jobs:
         with:
           github_token: "${{ secrets.GITHUB_TOKEN }}"
           paths: '["docs/**"]'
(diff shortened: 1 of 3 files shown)


lukicdarkoo:
Ready for review

omichel (approved):
It works great now, although it is not always very fast, as the squared distance threshold condition is often entered. But it is still better than before. Thank you.

lukicdarkoo:
> It works great now, although it is not always very fast, as the squared distance threshold condition is often entered. But it is still better than before. Thank you.

For small distances, it is faster. But indeed, for big movements, it fallbacks to the old mechanism.

PR #2442, opened 2020-11-09
Improve startup performance of external controller

lukicdarkoo:
The purpose of the PR is to:
- increase waiting time for Webots' pipe,
- request more frequently (e.g. in `webots_ros2` we were usually waiting for 4 ticks which makes 2s late detection on average), and
- gracefully exit the controller.

----

I tested with `webots_ros2` and a simple controller (independent of `webots_ros2`) and it works fine.
I am not aware of side-effects, so let me know if I should correct something.


lukicdarkoo:
Regarding external controller and <KBD>CTRL</KBD> + <KBD>C</KBD> (previous implementation):
- with C and Ubuntu the process stops as expected,
- with Python and Ubuntu the `sleep` instruction is skipped but it stays in the loop (you have to kill it multiple times to exit),
- with Python and `cmd.exe` nothing happens, the process is completely blocked.

Therefore, we had inconsistent behavior across different programming languages and operating systems. To get consistent behavior I guess we have to override default `SIGINT`, but I am not sure whether we want to do that.

Maybe we should investigate it, but as for now, I overrode `SIGINT` only for part of the code which establishes communication with Webots.

omichel:
It would be nice to have a consistent behavior across platforms (at least between Windows MSYS2, Linux and macOS) and languages (C/C++ and Python). I just tested and currently (develop branch), <kbd>Ctrl</kbd>-<kbd>C</kbd> works to break an extern controller from `cmd.exe`.

lukicdarkoo:
So Python makes problems. 

P.S.: I use this command to test Python controllers:
```bash
python -c "from controller import Supervisor; Supervisor()"
```

omichel:
Maybe we should use a try/catch construct in Python, see: https://stackoverflow.com/questions/1364173/stopping-python-using-ctrlc or simply give up and document that Python controllers should be stopped with <kbd>Ctrl</kbd>-<kbd>Break</kbd>.

omichel:
Or use `win32api.SetConsoleCtrlHandler` on Windows (see the [StackOverflow discussion thread](https://stackoverflow.com/questions/1364173/stopping-python-using-ctrlc)).

lukicdarkoo:
> Maybe we should use a try/catch construct in Python
Our problem is that the controller stucks in C and `SIGINT` is not properly passed there.

> simply give up and document that Python controllers should be stopped
Unfortunatelly, that doesn't solve problem for Linux. I have <kbd>ctrl</kbd> + <kbd>z</kbd> and then `kill -9 %1`

04e40f0 works fine on Linux, not sure about Windows

lukicdarkoo:
> Therefore, we had inconsistent behavior across different programming languages and operating systems. To get a consistent behavior I guess we have to override default `SIGINT`, but I am not sure whether we want to do that.

One not here, we can handle the `SIGINT` signal and then we propagate it:
https://stackoverflow.com/a/6015911/1983050
That way, users will still be able to write a custom handler for the `SIGINT` signal.

lukicdarkoo:
@omichel Let me know whether the implemented approach is acceptable (I will test on Windows). It catches the SIGINT signal, but it propagates the signal further.

omichel:
Yes, that looks good to me.

omichel (approved):
Thank you.

Issue #2441, opened 2020-11-09
Display Python API: support Camera.getImage return value


stefaniapedrazzi:
Passing the `Camera.getImage` return value to the `Display.imageNew` function is not working on Python and the controller crashes throwing this error:
```
munmap_chunk(): invalid pointer
```

This is probably because `Camera.getImage` returns a string and `Display.imageNew` doesn't support this format.
But Webots should at least throw a meaningful error or accept the output of the `Camera.getImage` function given that this is a common usage example.

Additionally, you should add an example in the Display documentation page showing how to correctly display a Camera output in Python (and other programming languages).

stefaniapedrazzi:
I quickly checked the image format returned by the `wb_camera_get_image` function because I couldn't get enough information from the documentation:
* _C/C++ and Python (Camera.getImage)_: **BGRA**
* _Java_: `Camera.getImage` has format **ARGB** but the Java `ImageRef` applies a conversion from big endian to little endian so that the **BGRA** format has to be passed to the `Display.imageNew` function to display it correctly
    https://github.com/cyberbotics/webots/blob/master/resources/languages/java/controller.i#L327-L339
* _Python_ `Camera.getImageArray`: **RGB**
    https://github.com/cyberbotics/webots/blob/master/resources/languages/python/controller.i#L299-L300

I will also check if I can find why different formats have been chosen for Java and Python `Camera.getImageArray`.

Then, I will check that all the information are written in the documentation and add the example about how to show a Camera image in a Display.

omichel:
I believe BGRA format was chosen to improve performance, as it used to be faster than RGBA on old graphics hardware (a long time ago). This is not the case any more and we kept this format for backwards compatibility reasons. The RGB and RGBA format are however more standard (when passed to image libraries such as OpenCV). With the introduction of `wren.js`, we may have to drop the BGRA format and use RGBA natively instead as BGRA is not supported by WebGL. So, it could be that we will have to change Webots to rely on RGBA format only so that we could use the very same shaders in Webots and in `webots.js`.

stefaniapedrazzi:
Ok. Anyway there are many consistency issues and worst of all the format is not clearly documented.

The difference in Java doesn't seem to rely on performance but only at the little vs. big endianness and the API should be consistent: we cannot output the **ARGB** format and asking the user to input it specifying that is **BGRA**.

Also in Python I don't understand why the two available functions return data using different format.
[...]

PR #2443, opened 2020-11-09
Fix Python Display.imageNew and improve documentation

stefaniapedrazzi:
Address #2441: improve documentation of the image format returned by the Camera API and how to display a Camera image in a Display.

Tasks:
* [x] specify image format in Camera section
* [x] add examples in Display section
* [x] ~add a warning in the Python API if the passed image has string format~
    _edit:_ fix free of invalid pointer in the Python API if the passed image has string format

### DIFF
--- a/docs/reference/display.md
+++ b/docs/reference/display.md
@@ -748,7 +748,7 @@ int main() {
 
   int time_step = wb_robot_get_basic_time_step();
   WbDeviceTag camera = wb_robot_get_device("camera");
-  wb_camera_enable(time_step);
+  wb_camera_enable(camera, time_step);
   const int width = wb_camera_get_width(camera);
   const int height = wb_camera_get_height(camera);
   WbDeviceTag display = wb_robot_get_device("display");
@@ -787,7 +787,7 @@ int main() {
   Display *display = robot->getDisplay("display");
 
   while (robot->step(timeStep) != -1) {
-    const unsinged char *data = camera->getImage();
+    const unsigned char *data = camera->getImage();
     if (data) {
       ImageRef *ir = display->imageNew(width, height, data, Display::ARGB);
       display->imagePaste(ir, 0, 0, false);
@@ -845,7 +845,7 @@ public class ShowCameraInDisplay {
 
     while (robot.step(32) != -1) {
       int[] data = camera.getImage();
-      if (data) {
+      if (data != null) {
         ImageRef ir = display.imageNew(width, height, data, Display.ARGB);
         display.imagePaste(ir, 0, 0, false);
         display.imageDelete(ir);
@@ -877,4 +877,4 @@ end
 %tab-end
 %end
 
-> **Note** [Java]: The `Display.imageNew` function can display the image returned by the `Camera.getImage` function directly if the pixel format argument is set to ARGB.
+> **Note**: The `wb_display_image_new` function can display the image returned by the [`wb_camera_get_image`](camera.md#wb_camera_get_image) function directly if the pixel format argument is set to ARGB.
--- a/docs/reference/display.md
+++ b/docs/reference/display.md
@@ -815,9 +815,9 @@ height = camera.getHeight()
 display = robot.getDisplay('display');
 
 while robot.step(32) != -1:
-    data = camera.getImageArray()
+    data = camera.getImage()
     if data:
-        ir = display.imageNew(data, Display.RGB, width, height)
+        ir = display.imageNew(data, Display.ARGB, width, height)
         display.imagePaste(ir, 0, 0, False)
         display.imageDelete(ir)
 ```
--- a/docs/reference/display.md
+++ b/docs/reference/display.md
@@ -733,4 +733,148 @@ The `wb_display_image_delete` function releases the memory used by a clipboard i
 After this call the value of `ir` becomes invalid and should not be used any more.
 Using this function is recommended after a clipboard image is not needed any more.
 
+The following controller snippet shows how to copy a [Camera](camera.md) image to the main display image.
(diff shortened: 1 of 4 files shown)


stefaniapedrazzi:
I could finally identify the error causing the crash when passing a string image data to the Python `Camera.imageNew` function.
The problem is that the string is freed but it should not (`malloc` is only executed in case of a list type):
https://github.com/cyberbotics/webots/blob/fb5cd3df739e4773aa677d9445eebd2ab849493f/resources/languages/python/controller.i#L414-L416

I'm checking but I'm not sure if there is a way to recognize if the `const void *` argument has been created from a list (with `malloc`) or from a string (no `free` required).

docs/reference/display.md
```diff
@@ -733,4 +733,148 @@ The `wb_display_image_delete` function releases the memory used by a clipboard i
...
+  WbDeviceTag camera = wb_robot_get_device("camera");
+  wb_camera_enable(camera, time_step);
+  const int width = wb_camera_get_width(camera);
+  const int height = wb_camera_get_height(camera);
+  WbDeviceTag display = wb_robot_get_device("display");
+
+  while(wb_robot_step(time_step) != -1) {
+    const unsigned char *data = wb_camera_get_image(camera);
+    if (data) {
+      WbImageRef ir = wb_display_image_new(display, width, height, data, WB_IMAGE_ARGB);
```
stefaniapedrazzi, line 759: The `wb_camera_get_image` returns data in BGRA format.
But to work correctly I have to set the `wb_display_image_new` format to ARGB.
This doesn't seem very intuitive for users.

stefaniapedrazzi: The inversion comes from this line:
https://github.com/cyberbotics/webots/blob/fb5cd3df739e4773aa677d9445eebd2ab849493f/src/Controller/api/display.c#L751

For `wb_camera_get_image` input data in BGRA format the image pixel uint32_t is `0xbbggrraa`.
But it seems that internally it is stored in little-endian because accessing the image bytes you get inversed values and inversed bytes are sent to Webots ([here](https://github.com/cyberbotics/webots/blob/master/src/Controller/api/request.c#L113-L118)).
```
printf("pixel %02x %02x %02x %02x\n", ((unsigned char *)(i->image))[0], ((unsigned char *)(i->image))[1], ((unsigned char *)(i->image))[2], ((unsigned char *)(i->image))[3]);
# result for pixel data B=152(0x98) G=109(0x6d) R=90(0x5a) A=255(0xff)
# "pixel ff 5a 6d 98"
```

I fixed it locally and run the test suite and everything was working correctly.
But I would fix this on the `develop` branch to prevent issues with user controllers.

docs/reference/display.md
```diff
@@ -55,6 +55,7 @@ For example, in order to draw two red lines, the `wb_display_set_color` contextu
 The display image is shown by default on top of the 3D window with a cyan border, see [this figure](#display-overlay-image).
 The user can move this display image at the desired position using the mouse drag and drop and resize it by clicking on the icon at the bottom right corner.
 Additionally a close button is available on the top right corner to hide the image.
+If the mouse cursor is over the overlay image and the simulation is paused, the RGBA value of the selected pixel is displayed in the status bar at the Webots window bottom.
```
omichel, line 58:
```suggestion
If the mouse cursor is over the overlay image and the simulation is paused, the RGBA value of the selected pixel is displayed in the status bar at the bottom of the Webots window.
```

docs/reference/display.md
```diff
@@ -733,4 +734,148 @@ The `wb_display_image_delete` function releases the memory used by a clipboard i
 After this call the value of `ir` becomes invalid and should not be used any more.
 Using this function is recommended after a clipboard image is not needed any more.
 
-> **Note** [Java]: The `Display.imageNew` function can display the image returned by the `Camera.getImage` function directly if the pixel format argument is set to ARGB.
+The following controller snippet shows how to copy a [Camera](camera.md) image to the main display image.
+This is particularly useful if you need to process the [Camera](camera.md) image before copying it to the main display image, otherwise please use the [`wb_display_attach_camera`](#wb_display_attach_camera) function that provides better performance.
```
omichel, line 738:
```suggestion
This is particularly useful if you need to modify the [Camera](camera.md) image before copying it to the main display image, otherwise you should use the [`wb_display_attach_camera`](#wb_display_attach_camera) function that provides a better performance.
```

docs/reference/display.md
```diff
@@ -733,4 +734,148 @@ The `wb_display_image_delete` function releases the memory used by a clipboard i
...
+%tab "C"
+```c
+#include <webots/robot.h>
+#include <webots/camera.h>
+#include <webots/display.h>
+
+int main() {
+  wb_robot_init();
+
+  int time_step = wb_robot_get_basic_time_step();
```
omichel, line 750:
```suggestion
  const int time_step = wb_robot_get_basic_time_step();
```

docs/reference/display.md
```diff
@@ -733,4 +734,148 @@ The `wb_display_image_delete` function releases the memory used by a clipboard i
...
+```cpp
+#include <webots/Robot.hpp>
+#include <webots/Camera.hpp>
+#include <webots/Display.hpp>
+
+using namespace webots;
+
+int main() {
+  Robot *robot = new Robot();
+  int timeStep = robot->getBasicTimeStep();
```
omichel, line 782:
```suggestion
  const int timeStep = robot->getBasicTimeStep();
```

omichel (changes requested):
It looks good to me. I just have a few minor suggestions.
We shall now fix the develop branch with respect to the ARGB format.

omichel (approved):
Thank you.

PR #2444, opened 2020-11-10
Update jointparameters.md

omichel:
The range for `position`, `minStop` and `maxStop` was wrong in case of `SliderJoint`.


lukicdarkoo (approved):
Thanks

PR #2451, opened 2020-11-16
Enhancement Linux installation scripts

stefaniapedrazzi:
Address issues with `npm` installation mentioned at https://github.com/cyberbotics/webots-snap/pull/22
  - [x] install latest `Node.js` version directly from the [nodesource repo](https://github.com/nodesource/distributions#installation-instructions):
    default version available through apt is still depending on libssl1.0 that it is not compatible with ROS
  - [x] move installation of Python 3.5. that is only used on Ubuntu 16.04
  - [x] add missing dependencies:
      - `zip` and `unzip` for compiling robotbenchmark and e-puck xc16
      - `dirmngr` to to add the key associated with the ROS install.

### DIFF
--- a/scripts/install/linux_optional_compilation_dependencies.sh
+++ b/scripts/install/linux_optional_compilation_dependencies.sh
@@ -10,8 +10,8 @@ apt install --yes software-properties-common
 add-apt-repository -y ppa:deadsnakes/ppa
 apt update
 apt install --yes lsb-release curl python3.6-dev python3.7-dev python3.8-dev python3.9-dev dirmngr
-curl -sL https://deb.nodesource.com/setup_15.x | sudo -E bash -
-sudo apt-get install -y nodejs
+curl -sL https://deb.nodesource.com/setup_15.x | bash -
+apt-get install --yes nodejs
 
 UBUNTU_VERSION=$(lsb_release -rs)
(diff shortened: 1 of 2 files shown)


stefaniapedrazzi:
I tested the creation of the snap package based on this branch and it could be created successfully.

scripts/install/linux_optional_compilation_dependencies.sh
```diff
@@ -9,21 +9,19 @@ fi
 apt install --yes software-properties-common
 add-apt-repository -y ppa:deadsnakes/ppa
 apt update
-apt install --yes lsb-release wget python3.5-dev python3.6-dev python3.7-dev python3.8-dev python3.9-dev
+apt install --yes lsb-release curl python3.6-dev python3.7-dev python3.8-dev python3.9-dev dirmngr
+curl -sL https://deb.nodesource.com/setup_15.x | sudo -E bash -
```
omichel, line 13: Is the sudo really needed here (two occurrences)? It seems the script need to be run as sudoer anyhow because of the apt install commands.

stefaniapedrazzi: Yes, you are right. We can get rid of the sudo.

omichel (approved):
Thank you.

PR #2452, opened 2020-11-16
Fix Display image format when loaded from data

stefaniapedrazzi:
Follow-up https://github.com/cyberbotics/webots/pull/2443#discussion_r524073387:
pixel data format is inverted when building the `uint32_t` from the separate R, G, B, A values.

* [ ] fix pixel `uint32_t` format
* [ ] update Display documentation


omichel (approved):
Thank you.

PR #2463, opened 2020-11-17
remove unused function

BenjaminDeleze:
Remove the ```wr_texture_cubemap_set_wrap_r``` function from texture_cubemap.h as there is no corresponding function in texture_cubemap.cpp or texture_cubemap.hpp


omichel (approved):
Thank you.

PR #2466, opened 2020-11-18
Add xvfb to requirements.

brettle:
**Description**
The build of the snap fails on a stock bionic system because xvfb is not installed.

### DIFF
--- a/scripts/packaging/Makefile
+++ b/scripts/packaging/Makefile
@@ -51,7 +51,7 @@ endif
 endif
 
 ifneq ($(SNAPCRAFT_PROJECT_NAME),)
-ifeq ($(TRAVIS),)
+ifeq ($(DISPLAY),)
 export DISPLAY = :99
 XVFB = xvfb
 endif
--- a/scripts/packaging/Makefile
+++ b/scripts/packaging/Makefile
@@ -106,7 +106,6 @@ iscc_formatter.exe: iscc_formatter.c
 xvfb:
 	+@echo "# starting Xvfb: X Virtual Frame Buffer"
 	+@Xvfb :99 -screen 0 1024x768x16 &
-endif
 
 clean:
 	+@rm -rf $(WEBOTS_DISTRO_EXE) webots.deb webots.snap webots.iss webots.mac files.txt iscc_formatter.exe
--- a/scripts/packaging/Makefile
+++ b/scripts/packaging/Makefile
@@ -104,7 +104,6 @@ iscc_formatter.exe: iscc_formatter.c
 	+@$(CC) -Wall -o $@ iscc_formatter.c
 
 xvfb:
-ifndef DISPLAY
 	+@echo "# starting Xvfb: X Virtual Frame Buffer"
 	+@Xvfb :99 -screen 0 1024x768x16 &
 endif
--- a/scripts/packaging/Makefile
+++ b/scripts/packaging/Makefile
@@ -104,8 +104,10 @@ iscc_formatter.exe: iscc_formatter.c
 	+@$(CC) -Wall -o $@ iscc_formatter.c
 
 xvfb:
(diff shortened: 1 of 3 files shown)


stefaniapedrazzi:
Thank you!

But in this case `xvfb` is only needed to create the snap and for running Webots on native systems, so we don't want to add it as a runtime dependency of Webots.
This should be added in the [webots-snap](https://github.com/cyberbotics/webots-snap) repo.

brettle:
Since `make install` fails on a stock headless bionic system, I'd argue this is a bug in this repo not `webots-snap`. I agree it's a build dependency, not a runtime dependency though.

stefaniapedrazzi:
Here are the instruction for example for installing Webots on Docker:
https://github.com/cyberbotics/webots/wiki/Docker#get-and-compile-webots
where it is specified that `xvfb` has to be installed.
But this is only required for running Webots on headless systems.
It is a very particular case and we don't want to install xvfb on other systems using the default installation scripts.

We can evaluate if it makes sense to add an additional script for headless system, but it will contain only `xvfb` for the moment, so I'm not sure it makes much sense.

stefaniapedrazzi:
I can add it in the Linux installation page:
https://github.com/cyberbotics/webots/wiki/Linux-installation

brettle:
If it's added in the compilation dependencies script, it won't be included in the deb package, right? It will just be installed for people trying to build from source. The addition would simplify the build documentation. I'm really not seeing the downside.

stefaniapedrazzi:
The downside is that it will be installed on the computer of all the users and developers that are compiling Webots and have a monitor and won't need the `xvfb` package on their system.

brettle:
It's only 2.2MB which is a drop in the bucket compared to all of the other deps that need to be installed to build from source. Weigh that against the time and frustration of developers who need it but don't have it installed. For example, it has taken you multiple attempts to get [the new automated snap build](https://github.com/cyberbotics/webots-snap/pull/23) to work because of this issue. That would not have occurred if this had been fixed.

stefaniapedrazzi:
Yes, in https://github.com/cyberbotics/webots-snap/pull/23 I'm experimenting but it is clear from the error message that `xvfb` is causing issues.

I got your point and I'm taking seriously into consideration your use case. But we should also consider all the other use cases. For example to run Webots in headless system it is not enough to install `xvfb` but it also needs to be setup before being able to start Webots.

So we think that the best option is, in the runtime dependencies script, to check if the system is headless, for example checking if the `DISPLAY` environment variable is empty, and if yes install `xvfb`.

A second step, as much importart as this one, would be to automatically setup and start `xvfb` and the other variables if the DISPLAY variable is empty (headless mode) so that Webots can be started out of the box.

brettle:
Sounds reasonable. I've made the suggested changes to the scripts and packaging Makefile. I'm assuming that getting webots to start out of the box on a headless server will be addressed separately.

scripts/packaging/Makefile
```diff
@@ -104,8 +104,10 @@ iscc_formatter.exe: iscc_formatter.c
 	+@$(CC) -Wall -o $@ iscc_formatter.c
 
 xvfb:
+ifndef DISPLAY
 	+@echo "# starting Xvfb: X Virtual Frame Buffer"
 	+@Xvfb :99 -screen 0 1024x768x16 &
+endif
```
stefaniapedrazzi, line 110: See previous comment
```suggestion
```

stefaniapedrazzi (changes requested):
There are some syntax errors making the tests crashing and a slightly different approach should be used in the packaging Makefile to not break the automatic generation of snap packages.

stefaniapedrazzi (approved):
Thank you!

I restarted the tests that were failing due to issues with in some dependency and now the Linux tests passed successfully.

PR #2472, opened 2020-11-25
replace rgb16f by rgba16f in order to be compatible with webgl

BenjaminDeleze:
**Description**
In WebGL2, the RGB16F format is not color-renderable whereas the RGBA16F format is (https://www.khronos.org/registry/webgl/extensions/EXT_color_buffer_half_float/). I propose to replace RGB16F by RGBA16F is these two functions that I need in WebGl.


omichel (approved):
That looks good. Thank you.

PR #2485, opened 2020-11-30
Fix artifacts upload issues

stefaniapedrazzi:
Here some suggestions to try to reduce the number of artifacts uploaded from GitHub actions:
- [x] on test suite / ros / world update do not upload debian packages
- [x] delete artifact at the end of the workflow: keep them only if label "test distribution" is set
- [x] limit artifacts life time to 14 days (in repo settings)
- [ ] cleanup already existing artifacts (2 months old)


stefaniapedrazzi:
I'm first using this script to purge the artifacts as the available actions are all failing due to ``API rate limit exceeded``:
https://gist.github.com/qwe321qwe321qwe321/efae4569576006624c34f23b2dd76a58

DavidMansolino:
FYI, I never saw any mention of artefacts limit, but the problem might be due to the single size of the artefact (see https://github.com/actions/upload-artifact/issues/147).

stefaniapedrazzi:
Yes, the size of the artifacts is also a limitations.
We should try to limit as much as possible the number of uploaded packages to just the needed ones.

Even if no specific total size is specified for the artifacts, many users reported issues with it.
Given that we don't need it, it seems in any case a good practice to keep them only as long as needed.

stefaniapedrazzi:
Deleting already existing artifacts doesn't seem to work as expected from scripts or actions due to high request rate.
But this can be done from the UI or from shell scripts.
So some of the old artifacts have been removed but not all of them that will expire automatically 90 days after the creation.

I would start merging these improvements and then evaluate if further improvements are needed.

omichel (approved):
Thank you.

Issue #2342, opened 2020-10-15
Move the object oriented API sources to a better folder


omichel:
Currently the source code for the object oriented APIs (C++, Python, Java and MATLAB) is located in [webots/resources/languages/](https://github.com/cyberbotics/webots/tree/master/resources/languages) which seems odd to me.
I believe this is inherited from the old time when Webots was not open source and we wanted to distribute the source code of these APIs, but we didn't want to distribute anything from the [src](https://github.com/cyberbotics/webots/tree/master/src) folder.
I think it would be at a better place in [webots/src/languages](https://github.com/cyberbotics/webots/tree/master/src/languages). @cyberbotics/maintainers: do you agree with that?

stefaniapedrazzi:
why not creating different folders for each languages in `webots/src/Controller` (`webots/src/Controller/c/`, (`webots/src/Controller/cpp/`, etc) ?
This would be a more intuitive structure for me.

In any case we have languages files in many different locations:
the Matlab API generation code is in the `webots/lib` folder (and ROS API in the `webots/projects/default`).

omichel:
Fixed in #2494.

PR #2494, opened 2020-12-02
Fix folder hierarchy

omichel:
Addresses #2342.
I took the opportunity of this PR to:
- [x] make the print out nicer during the compilation of the C, C++, Java and Python libraries
- [x] use the build folder for the compilation of the C++ library (and hence keep the debug and release objects like we do with the C library and webots)
- [x] fix the Java language Makefile which was wrongly recompiling the Java file each time it was called.
- [x] fix the MATLAB script mgenerate.py to generate UNIX line ending even when run on Windows.
- [x] [update the wiki](https://github.com/cyberbotics/webots/wiki/Adding-New-Node-and-API-Function/_compare/f85684ea6b7d3e582449ed0ad9085df25a0bc84f) and also [that](https://github.com/cyberbotics/webots/wiki/Adding-New-Node-and-API-Function/_compare/f85684ea6b7d3e582449ed0ad9085df25a0bc84f...7194433547822555ed76ddd74a6e07c928192e21).

I didn't move the `ros` folder in the new `src/controller` folder because `ros` is a controller, not a library. Therefore, it's better to keep its source code along with the resulting binary, like in any other controller folder.


stefaniapedrazzi (changes requested):
`scripts/packaging/files_core.txt` still contains references to the `resources/languages` folder:
https://github.com/cyberbotics/webots/blob/7b5e82a32b069a5bfd3ab6943b773ff39c938a08/scripts/packaging/files_core.txt#L229-L241

src/controller/cpp/Makefile
```diff
@@ -102,12 +115,14 @@ $(LIBCONTROLLER):
 -include $(DEPENDENCIES)
 
 # cpp compilation rule
-%.o:%.cpp
+$(OBJDIR)/%.o:%.cpp
+	@echo "# compiling "$<
```
stefaniapedrazzi, line 119: It seems that on linux $(OBJDIR) has an additional slash (`build//Accelerometer.o`), here is the log:
```
# compiling Accelerometer.cpp
g++ -c -fpic -Wall -I/home/stefi/develop/webots/include/controller/c -I/home/stefi/develop/webots/include/controller/cpp -o build//Accelerometer.o Accelerometer.cpp
```

But it works correctly.

stefaniapedrazzi (changes requested):
Than you!

I have some issues with `src/controller/c/Makefile` on linux.
If I run the Makefile on the `src/controller/` everything works fine. But if I then type `make clean` and then `make` in the `src/controller/c` folder, nothing is compiled.
Do you get the same issue?

omichel:
>  If I run the Makefile on the src/controller/ everything works fine. But if I then type make clean and then make in the src/controller/c folder, nothing is compiled.
Do you get the same issue?

Yes, you have to type "make release". I will fix the default target to release...

src/controller/cpp/Makefile
omichel: Again, you should write "make release" to avoid this problem...

stefaniapedrazzi: It is fine now that `make` without target just prints the message and doesn't compile.
Otherwise, it `make` was triggering the compilation, all the path should have be correctly also without target.

stefaniapedrazzi:
Changes are fine for me now but the test on macOS failed.

lukicdarkoo (approved):
The macOS fix looks good, thanks

lukicdarkoo:
The macOS fix should also go to the `master` branch

omichel:
Sure, I am taking care of that now. Thanks.

PR #2500, opened 2020-12-04
extends the exported fields for the stream viewer

BenjaminDeleze:
**Description**
Added export for the ambientIntensity variable of Ligth, filtering of ImageTexture, far and ambientOcclusionRadius of Viewpoint.

The far field is called "far" and not "zFar" because the zFar field is already utilized by the current streaming viewer and it expects another default value (2000 instead of 0)


omichel (approved):
Thank you.

Issue #2455, opened 2020-11-16
The `Group` node causing instabilities in the simulation (Webots crashes)


lukicdarkoo:
**Describe the Bug**
The simulation bellow crashes in R2020b rev1 and R2021a, while it works fine in R2020a rev1. Changing the `rotation` parameter, or deleting `boundingObject USE ToolShape2`, or moving the `Floor` node out of the `Group` node makes the simulation work fine. Therefore, I am not sure what is causing the issue, but I suspect it is the `Group` node as the user managed to make the full simulation (this is just a minimal version) work when removed the `Group` node.

It may be related to #2420

**Steps to Reproduce**
<details>
<summary>Relavant .wbt file</summary>

```
#VRML_SIM R2021a utf8
WorldInfo {
  coordinateSystem "NUE"
}
Viewpoint {
}
TexturedBackground {
}
TexturedBackgroundLight {
}
DEF ENVIRONMENT Group {
  children [
    Floor {
    }
  ]
}
DEF ROBOT3 UR10e {
  rotation 1 2.1577e-08 1.89076e-06 -1.5707953071795862
  name "robot_3"
  toolSlot [
    Solid {
      children [
        SliderJoint {
          jointParameters JointParameters {
          }
          device [
            LinearMotor {
              name "linear_motor1"
            }
          ]
          endPoint Solid {
            children [
              DEF ToolShape2 Shape {
                appearance PBRAppearance {
                }
                geometry Box {
                  size 0.1 0.1 0.1
                }
              }
            ]
            name "solid (3)"
            boundingObject USE ToolShape2
            physics Physics {
            }
          }
        }
        SliderJoint {
          jointParameters JointParameters {
          }
          device [
            LinearMotor {
              name "linear_motor2"
            }
          ]
          endPoint Solid {
          }
        }
        DEF ToolShape1 Shape {
          appearance PBRAppearance {
          }
          geometry Box {
            size 0.1 0.1 0.1
          }
        }
      ]
      boundingObject USE ToolShape1
      physics Physics {
        density -1
        mass 0.2
      }
    }
  ]
}
```
</details>
1. Save and run the code from the collapsable above  

<br />
<br />

**Expected behavior**
Should not crash 

**System**
 - Operating System: Windows 10
 - Graphics Card: NVIDIA GeForce RTX 2060
 - Webots version: R2020b rev1 and R2021a

omichel:
Fixed in #2505.

PR #2505, opened 2020-12-07
Call the physics plugin callback `webots_physics_collide` on Group node

### LINKED ISSUE #2420: Collision callback in physics plugin is not called if `boundingObject` is a `Group` node
**Describe the Bug**
Callback function `int webots_physics_collide(dGeomID g1, dGeomID g2)` in physics plugin is not called if `boundingObject` is a `Group` node.

**Steps to Reproduce**
1. Download [physics_callback_bug.zip](https://github.com/cyberbotics/webots/files/5465323/physics_callback_bug.zip)
2. Run `default.wbt`

**Expected behavior**
In the world, there are two same robots, but one has a `Group` node as a bounding object while the other only a `Shape` node. Therefore, the callback should be called for both of them which is not the case.

**System**
 - Operating System:  Windows 10
 - Graphics Card: NVIDIA GeForce RTX 2060

lukicdarkoo:
**Description**
Callback function `webots_physics_collide` used not to be called if `boundingObject` is a `Group` node. This PR fixes the behavior.

**Related Issues**
Fixes #2420
It also seems to be fixing #2455 (it is not crashing anymore)

**Tasks**
  - [x] Fix
  - [x] Check regressions
  - [x] Add changelog


omichel (approved):
Thank you.
I slightly improved the sentence in the change log to make it clearer.

PR #2508, opened 2020-12-07
Fix device list update in C++ API after device import

### LINKED ISSUE #2504: Device imported during controller run not correctly initialized in C++ API
A user reported a controller crash after importing two new devices in his C++ controller.
Both devices are correctly imported in Webots and initialized in the C controller, but the C++ API returns a NULL pointer.

World:
```
#VRML_SIM R2021a utf8
WorldInfo {
  coordinateSystem "NUE"
}
Viewpoint {
  orientation -0.25856043799738354 0.9596144768798006 0.1108447367505912 0.840281555844615
  position 26.89322046333004 9.636926848902812 18.708127242398263
}
TexturedBackground {
}
TexturedBackgroundLight {
}
Robot {
  controller "base_controller"
  supervisor TRUE
}
```

Controller:
```
#include <webots/Field.hpp>
#include <webots/Supervisor.hpp>
#include <webots/PositionSensor.hpp>

int main() {
    webots::Supervisor *supervisor = new webots::Supervisor();

    webots::Field* children_field = supervisor->getSelf()->getField("children");

    // Import of the first position sensor work
    std::string first_sensor_node = "HingeJoint { device [ PositionSensor { name \"first_position_sensor\" } ] }";
    children_field->importMFNodeFromString(0, first_sensor_node);
    webots::PositionSensor* first_pos_sensor = supervisor->getPositionSensor("first_position_sensor");
    first_pos_sensor->enable(50);
    // I can enable and access to first sensor value without problems

    // Import of the second position sensor
    std::string second_sensor_node = "HingeJoint { device [ PositionSensor { name \"second_position_sensor\" } ] }";
    children_field->importMFNodeFromString(0, second_sensor_node);
    webots::PositionSensor* second_pos_sensor = supervisor->getPositionSensor("second_position_sensor");

    if (second_pos_sensor == nullptr) {
        std::cout << "pos sensor nullptr" << std::endl;
    }

    // This line make the controller crash because of Segmentation fault
    second_pos_sensor->enable(50);

    while (supervisor->step(10) != -1) {
    }

    return EXIT_SUCCESS;
}
```

stefaniapedrazzi:
Fix #2504: fix the conditions that prevent to update the device list in C++, Java and Python APIs  after importing a new device during the controller run.


stefaniapedrazzi:
This feature was introduced in R2021a so no entry in the changelog is needed.
But I will add a test testing a C++ controller.

src/controller/cpp/Robot.cpp
```diff
@@ -449,15 +449,14 @@ Device *Robot::getOrCreateDevice(int tag) {
   if (tag == 0)
     return NULL;
 
-  int size = (int)deviceList.size();
-  if (size > 0) {
-    if (size <= tag)
-      return NULL;
+  const int count = wb_robot_get_number_of_devices();
+  const int size = (int)deviceList.size();
+  if (size == count && size > 0 && tag < size)
```
omichel, line 454: If we remove a device and add a new one during the same step, the count will remain the same, however, this lists may differ, isn't it? So, we should also re-initialize the list. It seems to me this test may not be sufficient to determine if the device list changed.

tests/other_api/controllers/cpp_import_device/cpp_import_device.cpp
```diff
@@ -0,0 +1,32 @@
...
+  int timeStep = (int)robot->getBasicTimeStep();
+  Field *childrenField = robot->getSelf()->getField("children");
+
+  // Import the first PositionSensor
+  childrenField->importMFNodeFromString(0, "DistanceSensor { name \"ds1\" }");
+  DistanceSensor *ds1 = robot->getDistanceSensor("ds1");
+  ts.assertPointerNotNull(ds1, "DistanceSensor 2 not found after import.");
+  ds1->enable(timeStep);
+
+  robot->step(timeStep);
```
omichel, line 23: Is this step needed here? If not, we should remove it. Otherwise we should explain why it is needed.

tests/lib/TestUtils.hpp
```diff
@@ -0,0 +1,93 @@
...
+    // write the result to the console
+    std::cout << "OK: " << mTestName << std::endl;
+    notifyControllerStatus(false);
+
+    delete mRobot;
+    exit(EXIT_SUCCESS);
+  }
+
+  // please use rather the assertion tests that this function
+  void sendErrorAndExit(const std::string &errorMessage) {
```
omichel, line 66: So why not making this method private instead of writing such a comment?

tests/lib/TestUtils.hpp
```diff
@@ -0,0 +1,93 @@
...
+    // remove the full path
+    std::size_t index = testName.find_last_of("/\\");
+    mTestName = testName.substr(index + 1);
+    // keep the basename (remove the .exe extension)
+    index = mTestName.find_last_of(".");
+    mTestName = mTestName.substr(0, index);
+    notifyControllerStatus(true);
+  }
+
+  void notifyControllerStatus(bool running) {
```
omichel, line 37: This one should also probably be private as well.

omichel (changes requested):

tests/other_api/controllers/cpp_import_device/cpp_import_device.cpp
stefaniapedrazzi: It is not needed. I added just because it is much easier to test if we apply one action per step.
But it can be removed without any problem.

tests/lib/TestUtils.hpp
stefaniapedrazzi: Because I copied it as-is from the C library.
I will make it private.

src/controller/cpp/Robot.cpp
stefaniapedrazzi: Yes.
We need to add an additional function in the C API that returns if the device list has been changed or not.

tests/other_api/controllers/cpp_import_device/cpp_import_device.cpp
omichel: I believe it's better to remove it, as it seems useless, e.g., a user would probably not call step in there. To me, it looks like a hack to work around a bug 😉.

src/controller/cpp/Robot.cpp
stefaniapedrazzi: I checked again, and with the current implementation it is enough to check if the count changed because devices are never removed from the C controller list so the size of the devices can only increase.

The return value of `robot_get_number_of_devices` is always the maximum tag value + 1.
Even if you remove a device it the count won't change and given that tags are not reused, when you add a new device the count will for sure increase.

tests/other_api/controllers/cpp_import_device/cpp_import_device.cpp
stefaniapedrazzi: I removed it.
But the bug was not affected in any way by the step.
It showed independently from the call of the step function.

src/controller/cpp/Robot.cpp
omichel: OK, that sounds good.

omichel: I would be nice nevertheless to write a comment about it in the code to make it clear.

omichel (approved):
Thank you.

PR #2509, opened 2020-12-07
Make Lidar return INF if depth is out-of-range

### LINKED ISSUE #1985: Multiple issues with LiDAR
**Describe the Bug**
HokuyoUtm30lx sensor shows wrong readings.

**Steps to Reproduce**
1. Run `webots /snap/webots/x1/usr/share/webots/projects/samples/devices/worlds/hokuyo.wbt`
2. Check the screenshots for incorrect behaviours

**Screenshots**
Default
image
Default, no changes in the world.

`minRange`
image
If we assume that the blue circle is `minRange` something is not correct according to the display in the up-left corner.

`noise` 
image
Noise is set to 0.005 meaning that the standard deviation is 15cm. Distances shown on the display, that represent the distance to the green box, are sometimes huge.
This is most likely related to #1784 

**System**
-  Graphics Card: Intel UHD Graphics 620

lukicdarkoo:
**Description**
Previously, the LiDAR would return MAX_RANGE if the depth is out-of-range. The behavior is hard to handle in robotic applications as we don't know whether the depth is equal to MAX_RANGE (consider the noise as well) or the depth is out-of-range.

**Analysis of the similar systems**

LDS-01
It seems it doesn't return measurements at all indices
https://github.com/ROBOTIS-GIT/hls_lfcd_lds_driver/blob/master/src/hlds_laser_publisher.cpp#L110
This probably depends on the measurement quality or range. I have this LiDAR in Lausanne so that I can verify - documentation is poor.

XV-11
It has very detailed information about each measurement
https://github.com/memristor/avr-lidar/blob/master/src/xv11.c#L99
https://github.com/rohbotics/xv_11_laser_driver/blob/kinetic-devel/src/xv11_laser.cpp#L93

More expensive LiDARs like Hokuyo URG also give information if an object is too close or too far:
https://www.roscomponents.com/en/index.php?controller=attachment&id_attachment=49

Gazebo returns infinity if the object above or below the range.

**Related Issues**
Fixes #1985

**Tasks**
  - [x] Implement fix
  - [x] Add to changelog
  - [x] Add to documentation
  - [x] Verify whether the noise is added to INF

**Result**
Note: infinity values are not visualized.

Peek 2020-12-08 20-36
**before**: besides improper handling of invalid values, notice there are some artifacts as well

Peek 2020-12-08 20-50
**now**


docs/reference/lidar.md
```diff
@@ -153,9 +153,11 @@ A typically good value for this field is to set it just big enough so that the s
 More information about the frustum is provided in the [frustum](camera.md#frustum) section of the [Camera](camera.md) node.
 
 - The `minRange` field defines the minimum range of the lidar, objects closer to the lidar than the minimum range are not detected (but still occlude other objects).
+If the range value is smaller the `minRange` value then infinity is returned.
```
stefaniapedrazzi, line 156:
```suggestion
If the range value is smaller than the `minRange` value then infinity is returned.
```

docs/reference/lidar.md
```diff
@@ -153,9 +153,11 @@ A typically good value for this field is to set it just big enough so that the s
...
 
 - The `maxRange` field defines the distance between the lidar and the far clipping plane of the OpenGL view frustum.
 This field defines the maximum range that the lidar can achieve and so the maximum possible value of the range image (in meter).
+If the range value is bigger the `maxRange` value then infinity is returned.
```
stefaniapedrazzi, line 160:
```suggestion
If the range value is bigger than the `maxRange` value then infinity is returned.
```

docs/reference/rangefinder.md
```diff
@@ -56,9 +56,11 @@ A typically good value for this field is to set it just big enough so that the s
 More information about the frustum is provided in the [frustum](camera.md#frustum) section of the [Camera](camera.md) node.
 
 - The `minRange` field defines the minimum range of the range-finder (objects closer to the range-finder than the minimum range are not detected (but still occlude other objects).
+If the depth is smaller the `minRange` value then infinity is returned.
```
stefaniapedrazzi, line 59:
```suggestion
If the depth is smaller than the `minRange` value then infinity is returned.
```

docs/reference/rangefinder.md
```diff
@@ -56,9 +56,11 @@ A typically good value for this field is to set it just big enough so that the s
...
 
 - The `maxRange` defines the distance between the range-finder and the far clipping plane of the OpenGL view frustum.
 This field defines the maximum range that a range-finder can achieve and so the maximum possible value of the range image (in meter).
+If the depth is bigger the `maxRange` value then infinity is returned.
```
stefaniapedrazzi, line 63:
```suggestion
If the depth is bigger than the `maxRange` value then infinity is returned.
```

projects/samples/devices/controllers/hokuyo/hokuyo.c
```diff
@@ -68,9 +65,11 @@ static void display(WbDeviceTag d,   // display
 
   int i;
   for (i = 0; i < ns; i++) {
-    float f = MIN(CLAMP_MAX, v[i]);
-    float alpha = -fov2 + fov * i / ns - M_PI_2;
-    wb_display_draw_line(d, dw2, dh2, dw2 + f * cos(alpha) * dw2 / CLAMP_MAX, dh2 + f * sin(alpha) * dh2 / CLAMP_MAX);
+    const float f = v[i];
+    if (f != INFINITY) {
+      float alpha = -fov2 + fov * i / ns - M_PI_2;
```
stefaniapedrazzi, line 70: Maybe `alpha` could also be set const
```suggestion
      const float alpha = -fov2 + fov * i / ns - M_PI_2;
```

tests/api/controllers/lidar_point_cloud/lidar_point_cloud.c
```diff
@@ -71,8 +71,12 @@ int main(int argc, char **argv) {
     float x = point_cloud[i].x;
     float y = point_cloud[i].y;
     float z = point_cloud[i].z;
-    float d = sqrt(x * x + y * y + z * z);
-    ts_assert_double_in_delta(image_lidar[i], d, 0.05, "Point cloud and range image does not correspond.");
+    if (!isinf(x) && !isinf(y) && !isinf(z)) {
+      float d = sqrt(x * x + y * y + z * z);
```
stefaniapedrazzi, line 75: If you are modifying this part of the code, maybe all these float variables could be set constant
```suggestion
      const float d = sqrt(x * x + y * y + z * z);
```

stefaniapedrazzi (changes requested):
This PR is fine for me.
There are just some typos in the documentation, and maybe some more `const` atrributes that could be added  in the code.

tests/api/controllers/lidar_point_cloud/lidar_point_cloud.c
lukicdarkoo: I noticed a lot of places where `const` is missing, but I wanted to keep the diff small. Although, maybe it is better use the opportunity and improve the code.

stefaniapedrazzi: Yes, keeping the diff quite small is good and I would not change all the files.
I just noticed this block because it was highlighted in the diff.
This is not strictly related with this PR, so it's fine for me even if you prefer to keep it like this.

stefaniapedrazzi (approved):
Thank you!

PR #2482, opened 2020-11-30
Fix updating device list after importing a joint descendant node

stefaniapedrazzi:
Fix #2476: import even in joint is not propagated to the top robot node and the list of devices is not updated.

PR #2514, opened 2020-12-07
Fix propagating import information to joint with slot parent

stefaniapedrazzi:
Fix check for joint's parent node introduced in #2482.
Previously the test_suite was failing if Webots was compiled in debug mode.


omichel (approved):
Thank you.

PR #2515, opened 2020-12-07
Moved R2020b-rev2 changelog into R2021a

omichel:
Since R2020b-rev2 is not going to be released, we need to move its change log into the one R2021a.


lukicdarkoo:
It seems there is a conflict

lukicdarkoo (approved):
Thanks

PR #2516, opened 2020-12-07
mesh proto, default Transform in toolSlot

Simon-Steinmann:
Changes to the protos files for:
ur3e, ur5e, ur10e, irb4600-40, P-rob3, Puma560

Changes:
- Split mesh data into seperate proto files
- Turn toolSlots into Solids with a name (needed for urdf conversion)
- Add a Transfrom node as the default in the exposed MFNode Field

for irb4600:
Give the Solids (links) names. They were nameless before.


omichel:
What is the benefit of creating so many proto files?

Simon-Steinmann:
All meshes are turned into proto files. This makes the main proto much smaller and actually editable. A multi-MB proto file is too much for the integrated IDE to properly handle.

Simon-Steinmann:
[new_protos_test.zip](https://github.com/cyberbotics/webots/files/5653213/new_protos_test.zip)
This sample project includes the robots in question. Load the world and see how the "view PROTO source" behaves. Now rename the protos folder (like "protos1") and reload the world. Now do the same thing. You should notice a HUGE difference when viewing and editing the PROTO source

omichel:
I see. However, this PR is difficult to review because the split induced so many changes that the other changes are difficult to observe. And you did a single commit that includes all the changes...
Did you use a script to split the protos?

Simon-Steinmann:
I used the proto splitter script for the splitting, However I also manually edited the things stated above. The issue is, that the universal robots and the P-rob3 have lots of individual meshes. Is there a way I can make this easier to review?

omichel:
I see. So, I would suggest that once your splitter script is merged, we apply it a separate PR, merge it and then merge the develop branch into this PR, so that we don't have any more this 150 file changes and we can review it more efficiently.

Simon-Steinmann:
the script is merged. Are you making the new PR? Are we just splitting these robots or more?

omichel:
I will let you create this new PR where you will only apply the script to split the robots, then we will merge it into develop. Then the other modifications that you did here will become more readable after we merge the develop branch into this PR.

omichel:
I am closing this PR as we are going to address the problem in a different way.

PR #2517, opened 2020-12-07
irb4600 multifile PROTO

Simon-Steinmann:


omichel:
We like the idea of splitting these big proto files into multiple files, and especially putting the meshes into big files that we will never open as a text file. However, we believe there is a better way to achieve this than creating sub-proto files.
Instead of using PROTO nodes for meshes, I believe we should use Webots [Mesh](https://cyberbotics.com/doc/reference/mesh) nodes.

Instead of:
```
  geometry Irb4600-40Mesh1 {
  }
```

we should have:

```
  geometry Mesh {
    url "Irb4600-40_meshes/base.dae"
  }
```

Where `base.dae` would be the collada file corresponding to that mesh. Of course, we can use any other format than collada.
The advantage of this method is many fold:

1. The 3D mesh can be easily edited with any 3D editor (like Blender) as saved as is and it will immediately take effect into Webots.
2. The load time in Webots should be faster (no multiple proto instantiations).
3. The `urdf2webots` converter could easily generate such proto by re-using the mesh files provided in the URDF model by simply copying them instead of converting them. This should ease the maintenance of URDF models.

Also, we believe the 3D file names should be carefully chosen to represent the 3D object the contain (instead of meaningless numbers).

Simon-Steinmann:
That makes sense. I just followed existing convention, but especially being able to edit the meshes is very important. I'll have a crack at it.  Simply using the trimesh data will be the simplest, but collada also allows for appearance data. Do you have existing functionality in mind?

omichel:
The appearance data is currently ignored by the Webots Mesh node. So, instead of Collada, we can use any other format that doesn't include appearance and is quite standard, like OBJ. Also OBJ is particularly simple and easy to convert from the mesh data we have, so it might be a good candidate.

Simon-Steinmann:
Yeah I was thinking of .obj too. I'll tinker around a bit.

omichel:
I believe it would be ideal to improve your proto_splitter script to generate OBJ files instead of geometry PROTO files. Let me know if I can help you doing this.

Simon-Steinmann:
From another Project I have a proto2xml converter that should make the conversion to .obj easy. The issue with a per-line conversion is, that the name of a solid is usually at the end of the proto. But it should be doable. I'll take a crack at it later. Didn't have much time until now.

Simon-Steinmann:
Hmm actually I could use some help. Turning the coords and indices into the obj file. A few lines of python would suffice. I dont fully understand the relationship between indices and faces.

omichel:
The tricky thing with OBJ format compared to IndexedFaceSet is that the index starts at 1 instead of 0.

omichel:
The VRML97 IndexedFaceSet:
```vrml97
IndexedFaceSet {
  coord Coordinate {
    point [
      0.0 1.0 0.0
      1.0 1.0 0.0
      1.0 0.0 0.0
      0.0 0.0 0.0
    ]
  }
  coordIndex [
    0 1 2 -1
    2 3 0 -1
  ]
}
```

should translate into the OBJ file:

```obj
v 0.0 1.0 0.0
v 1.0 1.0 0.0
v 1.0 0.0 0.0
v 0.0 0.0 0.0
f 1 2 3
f 3 4 1
```

Simon-Steinmann:
The index in obj. file is the line number right?

omichel:
Well, that's true in this case, but if you have empty lines or comment lines in the vertex list, it is not true anymore. The index is the number of the vertex. The first vertex is 1, the second is 2, etc.

omichel:
I am closing this PR as we are going the problem in a different way.

PR #2518, opened 2020-12-07
Fixed aseba compilation warnings

### LINKED ISSUE #2507: Remove outdated aseba dependency
The aseba submodule is causing tests to fail as it seems broken.
This is anyhow an old and outdated version of this library which is of little interest as no user seems to be using it (no feedback about it).
So, I would propose to remove it.

omichel:
Fixes #2507.

It was indeed pretty easy to update the aseba repositories (dashel and aseba) to the current master version, and recompile everything with no error and no warning any more.

So, I changed my mind and believe now that we can afford to keep this aseba dependency for another while.


projects/robots/mobsya/thymio/controllers/thymio2_aseba/Thymio2AsebaHub.cpp
```diff
@@ -76,8 +76,8 @@ void Thymio2AsebaHub::connectionCreated(Dashel::Stream *stream) {
 }
 
 void Thymio2AsebaHub::incomingData(Dashel::Stream *stream) {
-  uint16 temp;
-  uint16 len;
+  unsigned short int temp;
```
lukicdarkoo, line 79: Was it possible to include `#include <stdint.h>`?

lukicdarkoo:
Maybe we should mention in the changelog that the Aseba version is upgraded to v1.3.3

projects/robots/mobsya/thymio/controllers/thymio2_aseba/Thymio2AsebaHub.cpp
omichel: Yes, but `stdint` uses `uint16_t` and not `uint16`, so I would anyhow have to rename all those types... Also `short int` was used at other places in the code, so, it was more homogeneous like that.

lukicdarkoo (approved):
I capitalized "aseba" as here:
https://cyberbotics.com/doc/guide/thymio2#connect-aseba-to-the-thymio-ii-model

Thank you

PR #2525, opened 2020-12-08
Fixed compilation of OpenCV sample controller on Windows

omichel:
OpenCV libraries on Windows MSYS2 are now more similar to ones on the Linux/mac OS as the version is not included any more in the shared library name. This actually simplifies the Makefile 😄.


lukicdarkoo:
Looks good, should it be added to the changelog?

omichel:
I don't think so, as it used to work before and this doesn't bring any new feature.

lukicdarkoo (approved):
Thanks

PR #2526, opened 2020-12-08
Clear console on make clean

### LINKED ISSUE #2511: Clear console on make clean
When cleaning a controller compilation pressing the "Clean" button in the Webots text editor toolbar, the console is no longer cleared. This probably stopped working when introducing the multiple consoles.
But it is much more handy if the console is automatically cleared so that only the latest compilation information are displayed and I would reintroduce it.

omichel:
Fixes #2511.


stefaniapedrazzi (approved):
Thank you!

PR #2478, opened 2020-11-26
Publish macOS package from GitHub Action

lukicdarkoo:
Since Travis is not used anymore we need to publish macOS packages from GitHub Actions

PR #2527, opened 2020-12-08
Create Windows nightly builds from GitHub Action

omichel:
This will allow us to get rid of AppVeyor dependency.
I took advantage of this PR to remove the now useless `.travis.yml` file.


lukicdarkoo:
If the 702ead9 test doesn't work, you can add the `test_suite` label and do something like this:
https://github.com/cyberbotics/webots/pull/2478/commits/96981f133ab3672a2ffec02d045526c78dde9111

omichel:
That seems to work. The Window package was uploaded [here](https://github.com/cyberbotics/webots/releases/tag/untagged-fc1fbd1795fe0bf7a424).

omichel:
👍🏻 This PR is now ready for review

.github/workflows/test_suite_windows.yml
```diff
@@ -57,12 +57,15 @@ jobs:
...
         export WEBOTS_HOME=$GITHUB_WORKSPACE
         make distrib -j4
+    - name: Create/Update GitHub release
+      if: ${{ (github.event_name == 'push' || github.event_name == 'schedule') }}
+      run: |
+        python -m pip install requests PyGithub
+        scripts/packaging/publish_release.py --key=${{ secrets.GITHUB_TOKEN }} --repo=${{ github.repository }} --branch=${{ github.ref }} --commit=${{ github.sha }} --tag=${{ github.ref }}
     - uses: actions/upload-artifact@v2
-      if: ${{ contains(github.event.pull_request.labels.*.name, 'test distribution') }}
+      if: ${{ contains(github.event.pull_request.labels.*.name, 'test suite') || contains(github.event.pull_request.labels.*.name, 'test distribution') }}
```
stefaniapedrazzi, line 68: Why do you upload packages in case of the `test_suite` label if there is no test suite workflow for Windows?

Artifacts are only needed to pass the built packages to other jobs (like `test_suite`) but Windows.
Note that for macOS and Linux, the artifacts are deleted at the end of the workflow in case of `test_suite` label but this is currently not the case here in this version of the Windows workflow if you upload them here.

omichel: Thank you for the review.
Let me know if it looks better now.

stefaniapedrazzi: Thank you.
It's fine now even if the `delete-artifacts` job will never have artifacts to remove.

stefaniapedrazzi (approved):

.github/workflows/test_suite_windows.yml
omichel: It will if we use the "test distribution" label, isn't it?

omichel:
I just removed the AppVeyor app from the GitHub settings of webots.

.github/workflows/test_suite_windows.yml
stefaniapedrazzi: No, because the artifacts are uploaded only if the "test distribution" label is set and deleted only if the "test distribution" label is not set.
So if the "test distribution" label is set, artifacts are not deleted.

omichel: Oh yes... But that is ready for when we need it (when are going to introduce the nightly unit tests on Windows).

Issue #2281, opened 2020-09-30
Camera Image Noise


anilozdemir:
**Describe the Bug**
The simulation is adopted from Highway_overtake environment. I removed all the other cars, SUMO, etc. There is only a simple car with a front-facing camera, lots of trees around and many buildings (added from Webots objects database). The issue is the overlayed camera image includes plenty of noise (black shapes, ground texture rendered in wrong places). 

**Expected behavior**
The camera image should render the environment as is. 

**Screenshots**
![issue](https://user-images.githubusercontent.com/12694324/94703777-2001b700-0337-11eb-8fdf-4297dfdbb106.png)
![issue-2](https://user-images.githubusercontent.com/12694324/94704050-6820d980-0337-11eb-8f07-29558c5790ca.png)

**System**
 - Operating System: Ubuntu 20.04 (with i3 WM)
 - Graphics Card: two NVIDIA TITAN RTX

```
webots --sysinfo
System: Linux 5.4.0-45-generic x86_64
Processor: AMD Ryzen Threadripper 3970X 32-Core Processor
Number of cores: 64
OpenGL vendor: NVIDIA Corporation
OpenGL renderer: TITAN RTX/PCIe/SSE2
OpenGL version: 4.6.0 NVIDIA 440.100
```

**Additional context**
Added more tree to `.forest` file.

stefaniapedrazzi:
Could you please also share your project files producing the issue?

Do you have the same bug with the default `highway_overtake.wbt` world included in the Webots library?
To enable the camera device you should simply uncomment this line in the controller file:
https://github.com/cyberbotics/webots/blob/b34529dbdf110018eeb30257853b62d2bc970d88/projects/samples/robotbenchmark/highway_driving/controllers/highway_driving/highway_driving.py#L37

anilozdemir:
Please find the files in the repo: https://github.com/anilozdemir/Webots-Highway-Modification

In fact I do not have the issue when I run `highway_overtake.wbt` world. 

PS: I am also controlling the car using supervisor, by simply setting the translation coordinates. I don't know if the controller would have an impact on camera images.

stefaniapedrazzi:
The number of objects and the supervisor are not the cause of the issue.

But the are issues when merging the textures to produce the spherical camera image.
I'm investigating the problem.

stefaniapedrazzi:
I can also reproduce the issue on Ubuntu 16.04 with:
[...]

PR #2535, opened 2020-12-09
Improve highway grass shape

stefaniapedrazzi:
Disable shadows for the grass that could cause z-fighting issues on some particular camera settings (see #2281).


omichel (commented):
Shouldn't we also change the camera near and slightly move the camera as suggested in #2281?

stefaniapedrazzi:
No. The suggestions in #2281 are related to the custom robot used by the user in his custom simulation.
The default camera setup provided within the Webots samples doesn't cause any issue.

omichel (approved):
OK, thank you.

PR #2537, opened 2020-12-09
Fix pointer in Matlab API

### LINKED ISSUE #2316: Matlab Supervisor API functions return pointers
**Describe the Bug**
Some of the Matlab API functions of the Supervisor return non-native Matlab array, e.g.:
https://github.com/cyberbotics/webots/blob/master/lib/controller/matlab/wb_supervisor_node_get_velocity.m#L6

Instead, their returned value should be converted to an array, e.g.:
https://github.com/cyberbotics/webots/blob/master/lib/controller/matlab/wb_supervisor_field_get_sf_vec3f.m#L6-L8

stefaniapedrazzi:
Fix #2316

### DIFF
--- a/lib/controller/matlab/wb_inertial_unit_get_lookup_table.m
+++ b/lib/controller/matlab/wb_inertial_unit_get_lookup_table.m
@@ -1,0 +1,9 @@
+function wb_inertial_unit_get_lookup_table(tag, sampling_period)
+% Usage: wb_inertial_unit_get_lookup_table(tag, sampling_period)
+% Matlab API for Webots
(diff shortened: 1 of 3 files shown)


omichel (approved):
Thank you.
I assume this is the only method which has this problem, did you check them all?

omichel (changes requested):
We should probably comment out the corresponding line in the `mgenerate.py` script.

stefaniapedrazzi:
Yes, I checked all the methods and all the others are correct.

omichel (approved):
Thank you.

PR #2544, opened 2020-12-10
Fix sick example after range finder behavior change (infinity return value)

omichel:
The example was displaying many warnings in red in the console.
This was due to the return value of the lidar which can now return infinity and was not properly handled.
This PR fixes it.


lukicdarkoo (approved):
Thank you!

PR #2547, opened 2020-12-10
Fix controller external path for snap packages

lukicdarkoo:
It is wrong, it should be like here:
https://cyberbotics.com/doc/guide/running-extern-robot-controllers?tab-os=linux#environment-variables


omichel (approved):
Thank you.

PR #2551, opened 2020-12-10
Reduce timestep in the `ure` world to avoid warnings

lukicdarkoo:
Running the `ure.wbt` shows the following error:
```
WARNING: The current physics step could not be computed correctly. Your world may be too complex. If this problem persists, try simplifying your bounding object(s), reducing the number of joints, or reducing WorldInfo.basicTimeStep.
```


omichel (approved):
Thank you.

PR #2554, opened 2020-12-11
Fixed detection of Visual Studio on Windows

omichel:
There was a problem with the detection of Visual Studio 10 on Windows.
If it was not installed, the Makefile was setting its path to a non-existing folder instead of ignoring it.
This was breaking the compilation process in case Visual Studio 10 is not installed.
This PR fixes the problem.


lukicdarkoo (approved):
Thanks

Issue #2553, opened 2020-12-11
Shaders compilation problem on Mac


lukicdarkoo:
**Steps to Reproduce**
1. Open any world
```
ERROR: Shader compilation failed! Filename: pass_through.vert
ERROR: 0:1: '' :  version '300' is not supported
ERROR: 0:1: '' : syntax error: #version
ERROR: 0:2: '' :  #version required and missing.
ERROR: 0:3: '0' : syntax error: integers in layouts require GLSL 140 or later

ERROR: Shader compilation failed! Filename: line_set.vert
ERROR: 0:1: '' :  version '300' is not supported
ERROR: 0:1: '' : syntax error: #version
ERROR: 0:2: '' :  #version required and missing.
ERROR: 0:3: '0' : syntax error: integers in layouts require GLSL 140 or later

ERROR: Shader compilation failed! Filename: encode_depth.vert
ERROR: 0:1: '' :  version '300' is not supported
ERROR: 0:1: '' : syntax error: #version
ERROR: 0:2: '' :  #version required and missing.
ERROR: 0:3: '0' : syntax error: integers in layouts require GLSL 140 or later
```

**System**
 - Operating System: macOS Big Sur
 - Graphics Card: IntelHD

PR #2555, opened 2020-12-11
reverts the headers to 330 core for compatibility with Mac

BenjaminDeleze:
**Description**
It seems that Mac OS doesn't support the "300 ES" headers for the shaders. We will revert the headers to the previous core version

**Related Issues**
This pull-request fixes issue #2553


omichel (approved):
Thank you!

PR #2558, opened 2020-12-11
Fix nightly builds on master

omichel:
Fixes the creation of a non-draft prerelease for nightly build.

### DIFF
--- a/scripts/packaging/publish_release.py
+++ b/scripts/packaging/publish_release.py
@@ -48,18 +48,19 @@ if now.weekday() >= 5:
 warningMessage = '\nIt might be unstable, for a stable version of Webots, please use the [latest official release]' \
                  '(https://github.com/cyberbotics/webots/releases/latest).'
 if options.tag and not options.tag.startswith('refs/heads/'):
-    tag = options.tag.replace('refs/tags/', '')
-    title = tag
-    message = 'This is a nightly build of Webots from the "%s" tag.%s' % (tag, warningMessage)
-    if tag.startswith('nightly_'):
+    tagName = options.tag.replace('refs/tags/', '')
+    title = tagName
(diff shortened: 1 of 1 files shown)


scripts/packaging/publish_release.py
```diff
@@ -120,9 +120,7 @@
                         except requests.exceptions.ConnectionError:
                             remainingTrials -= 1
                             print('Release upload failed (remaining trials: %d)' % remainingTrials)
-                    if (releaseExists and
-                            not (options.tag and not options.tag.startswith('refs/heads/')) and
-                            not releaseCommentModified and
+                    if (releaseExists and tag.startswith('nightly_') and not releaseCommentModified and
```
lukicdarkoo, line 123:
```suggestion
                    if (releaseExists and not release.draft and not releaseCommentModified and
```
I am not completely sure whether it is going to work. But I believe it is better to define a logic (whether it is a draft or not) in the single place.

scripts/packaging/publish_release.py
```diff
@@ -85,7 +85,7 @@
 
 if not releaseExists:
     print('Creating release "%s" with tag "%s" on commit "%s"' % (title, tag, options.commit))
-    draft = True if options.tag else False
+    draft = False if tag.startswith('nightly_') else True
```
lukicdarkoo, line 88: `tag` is a scoped variable, defined in if condition before. I am surprised it used to work.

Maybe we should define `tag` as a public variable, just `tag = None` somewhere in the beginning.

lukicdarkoo (changes requested):
`tag` seems not to be declared as a public variable.

omichel:
It seems to be slightly better like this, but the diff is bigger, so hopefully, it doesn't break anything...

omichel:
When merged I will port it to the develop branch as well.

lukicdarkoo (approved):
Thank you

PR #2566, opened 2020-12-14
Fix Webots freeze on NaN value in hokuyo.wbt

### LINKED ISSUE #2563: hokuyo.wbt makes Webots freezing on Ubuntu 18.04 VM
In my Ubuntu 18.04 VM, the `hokuyo.wbt` blocks at 32ms and Webots is completely freezed and I have to manually kill it.
I tested with the snap package but I will now check with other packages too.

stefaniapedrazzi:
Fix #2563: in the VM some lidar values are NaN causing Webots to freeze when handling the `wb_display_draw_line` function.
However, on my native Ubuntu 18.04, it never occurs that lidar values are NaN.

### DIFF
--- a/projects/samples/devices/controllers/hokuyo/hokuyo.c
+++ b/projects/samples/devices/controllers/hokuyo/hokuyo.c
@@ -66,7 +66,7 @@ static void display(WbDeviceTag d,   // display
   int i;
   for (i = 0; i < ns; i++) {
     const float f = v[i];
-    if (f != INFINITY) {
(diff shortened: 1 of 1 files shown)


lukicdarkoo (approved):
OpenGL doesn't fully support the notion of infinity values. Therefore, I had to introduce a trick - max float that represents infinity. It is possible that the Mesa driver handles max float values differently.

Thanks!

omichel (approved):

Issue #1638, opened 2020-05-06
Better URL support in wbt files


omichel:
# Problem with the current design

It is currently pretty complicated to understand which textures, protos, robot controllers, sounds or plugin are used by looking only at the content of a *wbt* file. This is because there are many search paths with [complex priority rules](https://www.cyberbotics.com/doc/reference/imagetexture#search-rule-of-the-texture-path). This defeats the principle of Uniform Resource Locator (URL).

This also yields to problems when exporting to X3D format for example, where it is difficult to refer to the correct resources used by the X3D file (see, [textures name clash](#1619)).

# Proposed solution

We should take inspiration from the original VRML97 design principles where resources are located very precisely thanks to a better URL implementation.

1. The URL fields should be implemented as real URLs, either absolute or relative to the *wbt*/*proto* file:
  - `ImageTexture.url`
  - `Mesh.url`
  - `Background.*Url`
  - `Camera.noiseMaskUrl`
  - `ContactProperties.bumpSound`
  - `ContactProperties.rollSound`
  - `ContactProperties.slideSound`
  - `Skin.model`
  - `Robot.window [code]`
  - `Robot.controller [code]`
  - `Robot.remoteControl [code]`
  - `WorldInfo.physics [code]`
Examples:
  - `url "textures/image.jpg"`
  - `url "https://webots.cloud/R2021a/textures/webots_billboard.jpg"`
  - `url "../textures/protos/textures/card.jpg"`

2. Each PROTO used in a *wbt*/*proto* file should be declared at the beginning of the file using the simplified **EXTERNPROTO** syntax:
- `EXTERNPROTO "../protos/MyProto.proto"`
- `EXTERNPROTO "https://webots.cloud/R2021a/robots/boston_dynamics/spot/protos/Spot.proto"`
- `EXTERNPROTO "https://mywebsite.com/myfolder/MyBot.proto"`
All the resources (images and subsequent **EXTERNPROTO** will be downloaded as cascading URLs).
Icon URLs will be forged automatically by Webots to download the corresponding PROTO icons.

3. For URLs referring to code (marked `[code]` in the above list), one should consider they refer to the folder containing the code to be used by Webots or to the file to be used by Webots. Examples:
- `controller "../controllers/my_controller/"` will launch `my_controller[.exe]` located inside the `my_controller` folder.
- `controller "../controllers/my_controller/my_controller.py"` will launch `my_controller.py` with the default Python interpreter.
[...]

PR #2591, opened 2020-12-21
Assets download & caching

omichel:
This PR implements support for assets download and caching.
It addresses partly #1638.
- [x] Download of `ImageTexture.url`.
- [x] Cache of downloaded asserts.
- [x] `Background.*Url`
- [x] `Camera.noiseMaskUrl`
- [x] `Mesh.url`
- [x] `ContactProperties.*Sound`
- [x] `RotationalMotor.sound`
- [x] `LinearMotor.sound`
- [x] Support for assets download from PROTO nodes.
- [x] Example world.
- [x] Support changing the URL from the scene tree.
- [x] Support changing the URL from the supervisor.
- [x] Clean-up.
- [x] Tests.
- [x] Documentation.
- [x] Change log.

`WbApplication.mWorldLoadTimer`, which was causing the world loading dialog appear only after a few seconds, was removed for the following reasons:
1. it was causing problems with the asset download and would have required a redesign of the code to work properly.
2. it was not very useful as most of the worlds are anyhow displaying this dialog.


src/webots/nodes/WbImageTexture.cpp
```diff
@@ -383,8 +430,12 @@ void WbImageTexture::pickColor(WbRgb &pickedColor, const WbVector2 &uv) {
   // printf("pickedColor (u=%f, v=%f): (r=%f g=%f b=%f)\n", u, v, pickedColor.red(), pickedColor.green(), pickedColor.blue());
 }
 
-QString WbImageTexture::path() {
-  return WbUrl::computePath(this, "url", mUrl, 0);
+const QString WbImageTexture::path(bool warning) const {
+  if (mUrl->size() == 0)
+    return "";
+  if (mUrl->item(0).startsWith("https://"))
```
lukicdarkoo, line 436: `http://` should be included as well?

lukicdarkoo (approved):
Nice! I briefly checked and it works fine on Ubuntu

omichel:
I applied your suggestions and re-passed the tests.

projects/samples/howto/url/controllers/url/Makefile
```diff
@@ -0,0 +1,25 @@
+# Copyright 1996-2021 Cyberbotics Ltd.
```
lukicdarkoo, line 1: This file should be included to `projects/samples/howto/Makefile` and the description is missing in `samples-howto.md`.

lukicdarkoo (changes requested):
`hall.wbt` is all black, but I am not sure whether it is related to this PR or something else.

omichel:
Yes, it seems the problem with `hall.wbt` is caused by this PR. I am investigating it.

omichel:
It turns out the `hall.wbt` world has no light at all, so it seems normal that it should appear black.
However, it doesn't appear black on the `develop` branch, and I don't understand why...

DavidMansolino:
> However, it doesn't appear black on the develop branch, and I don't understand why...

It is because the background is producing some light. This is a commun technic used to optimize performance to use only the background to illuminate the scene (and additionally it it looks more realistic in indoors environment where the source of the light is extremely diffuse).

The fact that it doesn't work in this branch probably indicates that the `*IrradianceUrl` images of the background are not loaded.

omichel:
> It is because the background is producing some light. This is a commun technic used to optimize performance to use only the background to illuminate the scene (and additionally it it looks more realistic in indoors environment where the source of the light is extremely diffuse).
>
> The fact that it doesn't work in this branch probably indicates that the *IrradianceUrl images of the background are not loaded.

That makes sense. Thank you!

omichel:
Now the `hall.wbt` is rendered properly (thanks to @DavidMansolino!)

lukicdarkoo (dismissed):
Looks good to me. However, it would be good that @stefaniapedrazzi check it again.

src/webots/nodes/WbBackground.cpp
```diff
@@ -287,199 +389,248 @@ void WbBackground::applyColourToWren(const WbRgb &color) {
...
+bool WbBackground::loadTexture(int i) {
+  if (mTexture[i])
+    return true;
+  QString url;
+  QIODevice *device;
+
+  if (mDownloader[i]) {
+    url = mUrlFields[i]->item(0);
+    if (!mDownloader[i]->error().isEmpty()) {
+      warn(tr("Cannot not retrieve '%1': %2").arg(url).arg(mDownloader[i]->error()));
```
stefaniapedrazzi, line 401:
```suggestion
      warn(tr("Cannot retrieve '%1': %2").arg(url).arg(mDownloader[i]->error()));
```

src/webots/nodes/WbBackground.cpp
```diff
@@ -287,199 +389,248 @@ void WbBackground::applyColourToWren(const WbRgb &color) {
...
+      return false;
+    url = WbUrl::computePath(this, QString("%1Url").arg(gDirections[i]), mUrlFields[i]->item(0), false);
+    if (url == WbUrl::missingTexture()) {
+      warn(tr("Texture not found: '%1'").arg(mUrlFields[i]->item(0)));
+      return false;
+    }
+    device = new QFile(url);
+    if (!device->open(QIODevice::ReadOnly)) {
+      warn(tr("Cannot open texture file: '%1'").arg(url));
+      return false;
```
stefaniapedrazzi, line 419: Is it not needed to delete `device` before returning?
Same for the other return cases later in this function.

src/webots/nodes/WbBackground.cpp
```diff
@@ -287,199 +389,248 @@ void WbBackground::applyColourToWren(const WbRgb &color) {
...
+      WbUrl::computePath(this, QString("%1IrradianceUrl").arg(gDirections[i]), mIrradianceUrlFields[j]->item(0), false);
+    if (url.isEmpty()) {
+      warn(tr("%1IrradianceUrl not found: '%2'").arg(gDirections[i], mUrlFields[i]->item(0)));
+      return false;
+    }
+    device = new QFile(url);
+    shouldDelete = true;
+    if (!device->open(QIODevice::ReadOnly)) {
+      warn(tr("Cannot open HDR texture file: '%1'").arg(mIrradianceUrlFields[j]->item(0)));
+      return false;
```
stefaniapedrazzi, line 509: Is it not needed to delete `device` before returning?

stefaniapedrazzi (changes requested):
Globally everything is fine, but I found a couple of issues:
* as described more in detail in one of the comments, deleting the `ImageTexture` url item (`PBRAppearance` node) doesn't trigger the object appearance change in the 3D view
* `Background`:
    * also in this case deleting one `*Url` value doesn't trigger the scene update (I have to save and revert the world to see the black background)
    * the console message warning about the incomplete cubemap when one of the `*Url` field is empty is no longer printed in the console. I would keep it as it helps users understanding why the background is still black after defining some of the fields

src/webots/app/WbApplication.cpp
```diff
@@ -199,7 +198,7 @@ void WbApplication::linkLibraries(QString projectLibrariesPath) {
 }
 
 void WbApplication::setWorldLoadingProgress(const int progress) {
-  if (!mWorldLoadingProgressDialogCreated && mWorldLoadTimer && (mWorldLoadTimer->elapsed() / 1000) > 2) {
```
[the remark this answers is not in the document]
omichel: I just explained in the PR description.

omichel:
Thank you @stefaniapedrazzi for the in depth review. I believe I fixed all the problems you reported.

stefaniapedrazzi (changes requested):
Thank you! Yes, all my previous comments were fixed.

There is still an issue with the irradiance cubemap.
If the cubemap is complete but the irradiance cubemap is not complete,  Webots crashes with the following log:
```
webots-bin: TextureCubeMap.cpp:73: virtual void wren::TextureCubeMap::prepareGl(): Assertion `mDatas[i]' failed.
```

I can reproduce it by converting to base nodes the `TexturedBackround` PROTO, deleting one of the `*IrradianceUrl` item, saving and reverting the world.
Maybe there are also some update issues, because otherwise it should crash immediately when the irradiance url is delete. Also, previously the following message was printed in the console, but now nothing is printed when deleting the irradiance url
```
WARNING: Background: Incomplete irradiance cubemap
```

omichel:
I cannot reproduce this crash. I tried both on Windows and Linux. Which world did you open to make it crash?

stefaniapedrazzi:
I tested different worlds and I can reproduce it with all of them.
For example the `battery.wbt` world.

It is an assertion failure so I think it is not reproducible if Webots is compiled in release mode.

omichel:
Oh, I see, Wren should be compiled in debug mode...

omichel:
OK, I fixed the crash.

stefaniapedrazzi:
I confirm the crash is gone and now the message is correctly printed in the console.

Could you just check one last behavior?
If I load a modified `battery.wbt` world where the `frontIrradianceUrl` is empty, and I copy the `bottomIrradianceUrl` to the `frontIrradianceUrl`, the rectangle arena walls become darker. But if I remove again the `frontIrradianceUrl` value they doesn't return hell. It seems like the update is not applied.

omichel:
> If I load a modified battery.wbt world where the frontIrradianceUrl is empty, and I copy the bottomIrradianceUrl to the frontIrradianceUrl, the rectangle arena walls become darker. But if I remove again the frontIrradianceUrl value they doesn't return hell. It seems like the update is not applied.

Well spotted! I believe I fixed this one as well.

stefaniapedrazzi:
Thank you!
Now the update is applied correctly and I can clearly see the difference when re-adding an irradiance url (and I get the same after save and revert). Just one last note, the behavior in this PR is now different from before and if the irradiance cubemap is not complete, also the background "baseColor" cubemap is not displayed.

omichel:
> Just one last note, the behavior in this PR is now different from before and if the irradiance cubemap is not complete, also the background "baseColor" cubemap is not displayed.

Yes, that is true, but I believe it's better that way, as we don't want to encourage incomplete maps. Users should fix all the warnings if they want the background to be displayed.

stefaniapedrazzi (approved):
If the behavior is wanted then everything is fine for me.

DavidMansolino:
> > Just one last note, the behavior in this PR is now different from before and if the irradiance cubemap is not complete, also the background "baseColor" cubemap is not displayed.
> 
> Yes, that is true, but I believe it's better that way, as we don't want to encourage incomplete maps. Users should fix all the warnings if they want the background to be displayed.

Note that adding the constraint to always having a complete irradiance map to display the background is probably fine, but adding the opposite constraint (complete background (color) map for the irradiance to work) is IMO a regression as it is often useful to have the reflection of the background without actually seeing it (e.g. when you are inside a building, by the way, several of the world files use this technic currently).

In any case, the ``TexturedBackground`` PROTO currently doesn't respect the new constraint introduced in this PR (https://github.com/cyberbotics/webots/blob/develop/projects/objects/backgrounds/protos/TexturedBackground.proto#L36-L37)

omichel:
> Note that adding the constraint to always having a complete irradiance map to display the background is probably fine, but adding the opposite constraint (complete background (color) map for the irradiance to work) is IMO a regression as it is often useful to have the reflection of the background without actually seeing it (e.g. when you are inside a building, by the way, several of the world files use this technic currently).
>
> In any case, the TexturedBackground PROTO currently doesn't respect the new constraint introduced in this PR (https://github.com/cyberbotics/webots/blob/develop/projects/objects/backgrounds/protos/TexturedBackground.proto#L36-L37)

You are right, it seems the [hall.wbt](https://github.com/cyberbotics/webots/blob/develop/projects/samples/environments/factory/worlds/hall.wbt) world is broken again for this reason. I could fix it previously as it was reported broken by @lukicdarkoo, but it seems I broke my own fix afterward 😒... I will further investigate and fix this.

omichel:
That should be fixed in #2695: the color map should be either complete or empty (e.g., 6 or 0 textures loaded) so that the irradiance map is loaded.

Issue #2539, opened 2020-12-10
Simulation reset not working with Stewart platform


omichel:
When running the [stewart_platform.wbt](https://github.com/cyberbotics/webots/blob/master/projects/samples/demos/worlds/stewart_platform.wbt) simulation, if you press the reset button the simulation is not reset properly:
![stewart_platform](https://user-images.githubusercontent.com/1264964/101739685-2e531880-3ac8-11eb-9197-2d34d4bfa96f.png)

omichel:
Fixed in #2593.

PR #2593, opened 2020-12-22
removal of the forced reset for solids with ballJoint parents that causes incorrect reset

ad-daniel:
**Description**
The reset function of WbSolid explicitly treats solids having a ballJoint parent differently from all other types of joint, suggesting that in this particular case the parent doesn't take care of it, hence it forces a manual reset of these types of Solid to their initial values.

In practice, that doesn't appear to be the case, as the reset propagates through the chain just fine:

> BallJoint::reset() ---> WbHinge2Joint::reset() ---> WbJoint::reset() ---> WbBasicJoint::reset() ---> endpoint reset (in this case, the _upper_surface_ solid)

It also explains why in issue #2539 resetting twice the steward_platform.wbt works:

- **first reset:** _upper_surface_ solid forced to its initial position, then when the rest of the chain updates it messes up (the pistons reset fine, only _upper_surface_ isn't where it should be).
- **second reset:** since the rest of chain is now in the correct position, the second forced reset of the _upper_surface_ solid puts it where it should be

**Fix**
By removing this forced manual update for the specific case of a Solid having a BallJoint parent, the simulation reverts to the initial position correctly.

PS: the manual reset itself shouldn't be removed. It takes care of resetting free-floating objects (in this case the flying boxes) and resetting the root of the robot, from which the rest propagates.

**Tests**
visual validation of the fix in the following worlds:
- all files in samples/demos

The addition of a world having a highly jointed robot to use as testbench for the reset mechanism might be warranted

**Related Issues**
This pull-request fixes issue #2539

**Screenshots**
Before reset:
messed_up
After reset:
after_reset

**Additional context**
Add any other context about the pull-request here.

### DIFF
--- a/docs/reference/changelog-r2021.md
+++ b/docs/reference/changelog-r2021.md
@@ -5,6 +5,7 @@ Released on XX Xth, 2021.
 
   - Bug fixes
     - Fix Display external window not updated when the attached camera image changes ([#2589](https://github.com/cyberbotics/webots/pull/2589)).
+    - Fix reset of simulations including [BallJoint](balljoint.md) nodes like the [Stewart Platform](https://github.com/cyberbotics/webots/blob/master/projects/samples/demos/worlds/stewart_platform.wbt) ([#2593](https://github.com/cyberbotics/webots/pull/2593)).
 
 ## Webots R2021a
 Released on December 15th, 2020.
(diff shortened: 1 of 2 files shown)


omichel (approved):
Thank you for this very nice contribution!

omichel:
After you accept my invitation to the Committers team, you should be able to merge this PR.

Issue #1170, opened 2019-12-10
Port vehicle robot window to HTML


stefaniapedrazzi:
Qt robot windows are deprecated, we should update the vehicle robot window and port it to HTML.
Related to #893.

PR #2602, opened 2020-12-23
Add HTML vehicle robot window

stefaniapedrazzi:
Addresses #1170: port vehicle robot window from deprecated native Qt to HTML

Tasks:
* [x] implement vehicle window functionalities on top of the generic window
* [x] refactor the files and try to avoid duplicating code as much as possible
* [x] check performance of the simulation when robot window is active
* [x] open the `window` field for the vehicle PROTO models
* [x] update documentation: https://www.cyberbotics.com/doc/automobile/overview-tab
* [x] move update timeout from individual plots to the window


stefaniapedrazzi:
It would be good to have a first review.
If everything is fine, then I will update the documentation.

* I'm not sure if it is better to keep the `automobile` name or switch to `vehicle`.

* Then, currently I copied some sources (CSS, C, JS) from the generic robot window (as already done in other robot windows). Probably it would be better to link to the original ones, but this would mean to link to resources outside the current project and to restructure a little bit the generic robot window sources to make them reusable.
I already separated the generic and vehicle-specific code in this plugin so it is not a big work.

* Further improvement on the performance will be achieved by improving the generic robot window directly.

omichel (commented):
I tested a little bit on Windows, but the `boomer.wbt` example gives many errors in the console:
image
I also noticed that the layout is not perfect as the figure in green in the middle (0 m) is too close to the axis of the vehicle.

Also if I disable the camera in `city.wbt`, it crashes the controller:
image
I understand why it happens, but it's not very nice and we should probably fix it in some way.

projects/vehicles/plugins/robot_windows/automobile/automobile.c
```diff
@@ -0,0 +1,397 @@
...
+  char *tokens = strdup(command);
+  char *token = NULL;
+
+  enum DriverInfo info = NONE;
+  WbDeviceTag tag = 0;
+  bool robot = false;
+
+  while ((token = string_utils_strsep(&tokens, ":"))) {
+    if (info == NONE && tag == 0 && !robot) {  // first token = device or robot name
+      char *name0 = string_utils_replace(token, "\\:", ":");
```
lukicdarkoo, line 142: `const`?

lukicdarkoo (commented):
It works very well (except for a few errors Olivier mentioned)!

It takes ages to open the `PositionSensor` and `RotationalMotor` tabs. Is there any way to speed it up?

.github/workflows/test_suite_windows.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
   build:
     needs: cleanup-runs
-    if: ${{ github.event_name == 'push' || github.event_name == 'schedule' || contains(github.event.pull_request.labels.*.name, 'test distribution') }}
+    if: ${{ github.event_name == 'push' || github.event_name == 'schedule' || contains(github.event.pull_request.labels.*.name, 'test distribution') || contains(github.event.pull_request.labels.*.name, 'test suite') }}
```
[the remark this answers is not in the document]
stefaniapedrazzi: Yes, it is unrelated but I need it to test the compilation on Windows that otherwise it is not triggered.
Once finalizing this PR I will remove it.

stefaniapedrazzi:
I agree with all your comments but most of them are issues of the generic robot window and I think it should be fixed directly in the generic robot window code (and ported here only when we decide if and how to share the code).

I will now inspect the two issues about the error in `boomer` and the position of the wheelbase label that are currently the only ones that are cause by new code of this PR and not by the generic robot window.

stefaniapedrazzi:
>  I also noticed that the layout is not perfect as the figure in green in the middle (0 m) is too close to the axis of the vehicle.

I don't have this issue on my system and the label is shown correctly.
I will test on another machine, it could be that the rotated text is handled differently by the browsers.
Did you test on Windows?

label_issue

projects/vehicles/plugins/robot_windows/automobile/automobile.c
stefaniapedrazzi: I will address this first in #2655

projects/vehicles/plugins/robot_windows/automobile/plot_widget.js
```diff
@@ -0,0 +1,150 @@
+// @param container: DOM element in which to create the plot.
```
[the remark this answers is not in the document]
stefaniapedrazzi: Yes, exactly. And maintaining the same file three times it not good.

projects/vehicles/plugins/robot_windows/automobile/plot_widget.js
```diff
@@ -0,0 +1,150 @@
...
+  if (!this.initialized)
+    return;
+  if (show) {
+    this.canvas.style.visibility = 'visible';
+    this.xLabel.style.visibility = 'visible';
+    this.xMinLabel.style.visibility = 'visible';
+    this.xMaxLabel.style.visibility = 'visible';
+    this.yLabel.style.visibility = 'visible';
+    this.yMinLabel.style.visibility = 'visible';
+    this.yMaxLabel.style.visibility = 'visible';
```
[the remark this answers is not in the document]
stefaniapedrazzi: I will address this first in #2655

projects/vehicles/plugins/robot_windows/automobile/plot_widget.js
```diff
@@ -0,0 +1,150 @@
...
+  if (!this.initialized)
+    return;
+
+  this.xMinLabel.textContent = roundLabel(this.xRange['min']);
+  this.xMaxLabel.textContent = roundLabel(this.xRange['max']);
+  this.yMinLabel.textContent = roundLabel(this.yRange['min']);
+  this.yMaxLabel.textContent = roundLabel(this.yRange['max']);
+
+  // downscale the current image at the center.
+  var srcWidth = this.canvasWidth;
```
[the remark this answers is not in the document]
stefaniapedrazzi: I will address this first in #2655

stefaniapedrazzi:
> It takes ages to open the PositionSensor and RotationalMotor tabs. Is there any way to speed it up?

Yes, this is one of the issues of the generic robot window and should be fixed directly for the generic robot window and not only for the vehicle robot window.

omichel:
> I don't have this issue on my system and the label is shown correctly.
> I will test on another machine, it could be that the rotated text is handled differently by the browsers on Windows.

Yes, I tested on Windows. It seems to be a behavior discrepancy between QtWebKit and QtWebEngine, but there is probably a workaround.

projects/vehicles/plugins/robot_windows/automobile/automobile.c
stefaniapedrazzi: I checked but this variable cannot be turned to const because the `free` function would discard the `const` qualifier.

stefaniapedrazzi:
This PR is now ready for review.

omichel (changes requested):
It works nicely and looks very good.
However, when trying to drive the boomer tractor with the keyboard, the steering angle was displayed correctly but the speed was always 0.00 km/h regardless the actual speed of the tractor.

stefaniapedrazzi:
> However, when trying to drive the boomer tractor with the keyboard, the steering angle was displayed correctly but the speed was always 0.00 km/h regardless the actual speed of the tractor.

Yes, this is exactly the same behavior as with the previous Qt-based robot window.
The speed is 0 km/h because the tractor speed is not set using the Driver library but directly setting the wheel's motor velocity.

stefaniapedrazzi:
But I will check if there is a good way to improve it.

stefaniapedrazzi:
The vehicle robot window is not designed to show the speed of a vehicle that is not set with the Driver library (all the wheels could have different speed but only one label is available, etc.) and if users are interested in the single motor speed they can look at the Motor tab.
However, I agree that the previous representation was not clear and showing speed 0 seems a bug.
So I added a new control mode value to the Driver library to be able to detect if the speed was never set using the Driver library and disable the speed representation in the vehicle robot window hoping that this makes it easier to understand.

omichel:
Thank you for this change. However, it is still displaying the following misleading information:
- for each wheel: Speed: 0.0 rad/s
- Target Speed: 0.000 km/h

I also noticed that in the Steering tab, the graph is often jumping (clear and reset) at the beginning, making it difficult to watch.

stefaniapedrazzi:
Ok, I missed these additional speed.
I will also disabled them and try to improve the steering graph update.

resources/projects/plugins/robot_windows/generic/timeplot_widget.js
```diff
@@ -242,8 +242,19 @@ TimeplotWidget.prototype.updateRange = function() {
 TimeplotWidget.prototype.updateGridConstants = function() {
   let delta = this.yRange['max'] - this.yRange['min'];
   delta *= 0.999; // in order to decrease the order of magnitude if delta is a perfect divider of the increment
-  const orderOfMagnitude = Math.floor(Math.log(delta) / Math.LN10);
-  this.verticalGridSteps = Math.pow(10, orderOfMagnitude);
+  const minStep = delta / 10; // initial step size
+  const orderOfMagnitude = Math.floor(Math.log(minStep) / Math.LN10);
+  const magnitude = Math.pow(10, orderOfMagnitude);
+  // calculate most significant digit of the new step size
+  var residual = Math.round(minStep / magnitude + 0.5);
```
omichel, line 249:
```suggestion
  let residual = Math.round(minStep / magnitude + 0.5);
```

projects/vehicles/plugins/robot_windows/automobile/automobile.js
```diff
@@ -0,0 +1,237 @@
...
+
+  // Overview tab
+  overviewWidget = new OverviewWidget(document.getElementById('overview-section'));
+
+  addTab('Speed');
+  appendNewElement('Speed-layout', '<h2><span id="target-speed-label" style="color:red"></span></h2>');
+  addDriverInfo({name: 'Speed', htmlName: 'Speed'}, '[km/h]', -20.0, 20.0);
+
+  addTab('Steering');
+  addDriverInfo({name: 'Steering'}, '[rad]', -0.1, 0.1);
```
omichel, line 122: This is not sufficient for the boomer and probably for most vehicles.
```suggestion
  addDriverInfo({name: 'Steering'}, '[rad]', -1, 1);
```

omichel (changes requested):
Thank you. I just have a couple of comments.

projects/vehicles/plugins/robot_windows/automobile/automobile.js
stefaniapedrazzi: In the `city.wbt` example the steering ranges from [-0.15, 0.1]. So a range [-1, 1] is extremely big in most of the real cases.

Personally I really prefer (and it is what all the graphs in the robot window are doing) to set a smaller initial range and let the graph resize automatically. Otherwise for most of the vehicles there will be a big range where only 10% of it is used and a very poor resolution. So we will produce bad and hardly readable graphs.

omichel: I see... Can't we initialize these min and max value with the actual min and max position of the corresponding motors as we do in the default motor control widget?

omichel: The problem with the HTML window and resizing is that it erases everything before resizing (unlike the Qt robot window) which makes it hard to view and understand what's going on...

stefaniapedrazzi: The min and max motor positions of the BMW X5 wheels are [-1, 1].
So if we should use these value, this range will be too big for a normal run of the city simulation.

omichel: To me this is less annoying than having the graph resetting (and clearing all data) every time you increase or decrease the position outside of the current range.

stefaniapedrazzi: The problem is that if the graph is reset, data will be lost just for some frames. But if you set a bad range, then this will remains for the whole simulation.

I think that the real solution is to fix the graph update (in the generic robot winow) when the range changes and not hard coding the range based on the worst cases.

Just to information, this is the resulting graph for `city.wbt` when setting the range to [-1, 1] and it is very poor in comparison to what we can achieve automatically resizing the graph.
city_steering

omichel: I agree that the best solution is to fix the graph update.
For most users, the broken graph update looks really buggy whereas the small amplitude of the graph within the min/max range is more understandable even if it is not ideal.

stefaniapedrazzi:
When comparing the performance of this new HTML robot window with the old Qt robot window I found out that it was running slower than before.
So I improved the update rate of the plots (previously updated each time new data was received),  removed the individual timeplot update timeouts and added a single global update timeout at the window level.
Now the performance is comparable: depending on the visible/enabled plot the simulation sometimes runs slightly slower and sometimes slightly faster.

omichel (approved):
It now looks very good to me (and fast).
Thank you.

Issue #2600, opened 2020-12-23
The BMW model is missing rear blinkers


lukicdarkoo:
`setIndicator` doesn't affect the rear blinkers even though it should be the case.

PR #2615, opened 2021-01-02
Enhancement of vehicle rear lights for mercedes, lincoln, citroen, bmw and rangerover

ad-daniel:
**Description**

- the existing mesh, where possible, was split/adjusted to make use of the [VehicleLights](https://www.cyberbotics.com/doc/automobile/car#vehiclelights) primitive so that the rear car lights react accordingly to the different Driver calls. No new meshes were added, just reworked the existing ones.
- the documentation for VehicleLights ([car.md](https://github.com/cyberbotics/webots/blob/master/docs/automobile/car.md)) was updated according to the values present in the [proto](https://github.com/cyberbotics/webots/blob/master/projects/vehicles/protos/abstract/VehicleLights.proto) (default color values were wrong) 

PS: I've attempted to keep the general location of the lights correct with their real counterpart, but sometimes was tricky not knowing the year of the cars

**Test world**
[rear_indicators.zip](https://github.com/cyberbotics/webots/files/5760636/rear_indicators.zip)

**Related Issues**
This pull-request solves issue #2600

**When `wbu_driver_set_cruising_speed(-10)`**
reverse
notes: no change

**When calling `wbu_driver_set_antifog_lights`**
antifog
notes: the citroen c-zero rear fog light is white because this car is the only one with front fog lights working, and distinction between front and rear colors can't be made at the moment

**When calling `wbu_driver_set_dipped_beams`**
dipped_beam
notes: most light meshes were in this category (basically everything lights up), so they've been separated. The mercedes ones were just wrong. The citroen one was split to serve multiple purposes (like in reality)

**When calling `wbu_driver_set_brake_intensity(1.0)`**
braking
notes: despite having emissivity like blinkers, it doesn't currently show it much (compare with hazard lights below), not sure why.

**When calling `wbu_driver_set_hazard_flashers` (or the left/right indicators)**
indicators
notes: for the BMW (4th car) I took the mesh from the body of the car, which as far as I can tell it's the way it looks in older BMW x5 cars (which from the shape it appears to be), if it's a more recent model, then it's technically the wrong spot, assuming it matters.


omichel (changes requested):
Thank you for this great contribution.
We would simply need an entry in the change log in order to be able to merge this PR.

ad-daniel:
Thank you, I've added the entry. Also, should've mentioned it before, the reason for the disparity of additions/deletions is because the MercedesBenz model contrary to I believe all others has the coordinates/indexes on a single line (and there's lots) whereas where I modified things I've formatted it on multiple ones. Should I fix it and make it consistent with the rest of this proto? Or should the entire model be changed to be consistent with the others?

omichel:
We recently decided that it would be better to have points and indexes on a single line to improve readability of PROTO files. So, only recent or recently modified PROTO files are like that. However, this is not very important as we now plan to move the meshes outside PROTOs files using the Mesh primitive and probably the OBJ format for storing the meshes (coordinates, texture coordinates and normals).

omichel:
About the problem with the missing bloom effect of the brake lights, it may be due to the fact that blinkers appearance rely on a base color texture whereas brake lights rely on a plain base color (which could be a rendering bug). So, I would suggest to leave it like that and maybe open a new issue about a possible rendering bug for bloom with plain base color PBR materials.

docs/reference/changelog-r2021.md
```diff
@@ -3,6 +3,8 @@
 ## Webots R2021a Revision 1
 Released on XX Xth, 2021.
 
+  - Enhancements
+    - Rework of car meshes to have more realistic rear lights for MercedesBenz, Lincoln, Citroen, BMW and Range Rover models ([#2615](https://github.com/cyberbotics/webots/pull/2615)).
```
omichel, line 7:
```suggestion
    - Rework of car meshes to have more realistic rear lights for Mercedes Benz, Lincoln, Citroen, BMW and Range Rover models ([#2615](https://github.com/cyberbotics/webots/pull/2615)).
```

omichel (changes requested):
Just a minor change.

omichel (approved):
Thank you!

PR #2617, opened 2021-01-03
Fix URe ROS example

lukicdarkoo:
The example didn't work in Noetic. With this PR it works with Noetic and it should work fine with previous distributions.

Fixed issues:
- Import error:
  ```log
  ImportError: cannot import name `JointStatePublisher` from `joint_state_publisher` 
  ```
- Python3 issue:
  ```log
  genpy.message.SerializationError: <class 'struct.error'>: 'required argument is not an integer' when writing 'clock: 
    secs: 0
    nsecs: 608000000'
  ```

In addition, users can now use `JointStatePublisher` and `TrajectoryFollower` in other ROS packages.


omichel (approved):
Thank you.

andyleeeeeeeeee:
that works on ubuntu 20.04 ros-noetic, thanks for the effort

PR #2624, opened 2021-01-05
Fix MATLAB path

### LINKED ISSUE #2608: Using Matlab in Webots
hi guys

I have a problem with using Matlab in Webots.
When I try to open Webots sample projects for this purpose like nao_matlab or e_puck, it gives the following error. 

Untitled

This is while I added MATLAB to  PATH environment variable and also looked at Webots' User guide (Matlab section) and did other things like installing MATLAB MinGW-w64 C/C++ Compiler, but I still get this error.

could you please help me to solve this problem ?

lukicdarkoo:
Fixes #2608. We assumed MATLAB is installed in the path that contains `matlab`, this simple fix avoids it.


omichel:
I would add a line in the change log about this fix.

omichel (approved):

PR #2627, opened 2021-01-05
fix compilation error

ad-daniel:
Compilation fails without the header, since it's missing I presume the debug line slipped through.


omichel (approved):
Thank you.

PR #2629, opened 2021-01-06
Fix Cylinder unwrapped texture mapping

### LINKED ISSUE #2628: Wrong painting on bottom face of Cylinder
Painting with the Pen device on a Cylinder works fine on the top and side faces of the cylinder but not on the bottom face.
The issue is probably due to the texture mapping used for the painted texture (different from the one used for the appearance of thy Cylinder).

Steps to reproduce:
1. open the `pen_cylinder.wbt` test
2. Add a Transform node with rotation `[0 0 1 3.1415]` between the `BOARD` Solid and the Shape node.
3. Run the simulation -> the test fails because the dot on the bottom face is not painted

stefaniapedrazzi:
Fix #2628: texture coordinates for the bottom face of Cylinder where wrong.

Note: in the changelog I changed "Fix" to "Fixed" (form that we always used in the change log) for the existing entries.


omichel (approved):
Thank you.

Issue #358, opened 2019-04-11
Issues applying a force to the e-puck from the scene tree


stefaniapedrazzi:
**Describe the Bug**
Depending on where you click the apply force functionality doesn't work: it seems there are issues picking the object due to transparency.

**Steps to Reproduce**
1. Go to e-puck.wbt
2. Start the simulation
3. Apply a force by clicking on the a wheel -> it works correctly
4. Apply a force by clicking on the **transparent** external case -> it doesn't work

**Expected behavior**
It should be possible to apply a force to a Solid with physics by clicking on any part of it.

ad-daniel:
It isn't technically a bug, the e-puck's "body/chassis" is actually a huge LED.

     DEF EPUCK_BODY_LED LED {
      rotation 0 1 0 4.712399693899575
      children [
        Shape {
          appearance PBRAppearance {
            baseColor 0.5 0.5 0.5
            transparency 0.4
            roughness 0.5
            metalness 0
            emissiveIntensity 0.2
          }
          geometry IndexedFaceSet {

            [ ... POINTS & COORDS ...  ]

            creaseAngle 0.5
          }
        }
      ]
      name "led8"
      color [
        0 1 0
      ]
    }

Being a SolidDevice it doesn't have physics enabled by default hence why it can't be pulled from there. The e-puck docs mention that its body can light up so that might explain the choice of modeling it like this.

Enabling the physics in the body is kinda awkward though because the boundingObject and mass/inertia of the entire e-puck is enforced at the [Robot level](https://github.com/cyberbotics/webots/blob/3766a52ca250cc6ba41a777182a4aeb61d2d3642/projects/robots/gctronic/e-puck/protos/E-puck.proto#L1144-L1177). So it requires giving this LED a negligible mass and a dummy boundingObject  like this:

      boundingObject Transform {
        translation 0.0 0.03 0.0
        children [
          Box {
            size 0.001 0.001 0.001
          }
        ]
      }
      %{ if kinematic == false then }%
      physics Physics {
        density -1
        mass 0.0001
      }
      %{ end }%

Whether that's desirable I don't know.

elvir585:
pon, 28. pro 2020. 13:42 ad-daniel <notifications@github.com> je napisao:

stefaniapedrazzi:
We don't want to change the `e-puck` robot model.

[...]

PR #2635, opened 2021-01-07
Rework of external force/torque logic

ad-daniel:
**Related Issues**
This pull-request relates to issue #358

**Description**
As suggested in issue #358, if an external force is applied to a node without physics (typically a sensor which by default don't), and a node ancestor with physics exists, then it should be possible to apply the external force anyway.

The proposed solution goes up the chain until it finds a dynamic solid and does fix issue #358.

Currently, **it searches upwards, not sideways,** so if there's a Transform or a Group with multiple children (some with and some without physics), applying a force on a children wouldn't actually move the entire Group. (Like the stacked box/cylinder on the video)

The search can be extended sideways too, but in practice I don't really know if it's desirable because:
- If the Group/Transform is part of a Robot, then by the Physics rules (i.e _if any of the children nodes have physics, then the topmost node (the Robot) also needs to_), so searching upwards will always lead to a solid with physics to be found.
- If it's not part of a Robot, it means it's part of the environment and likely was _intended_ to have parts with physics and parts without, would it make sense for everything to be affected by the force? (assuming there's an ancestor with physics, otherwise can't be moved anyway)

Which begs the question if it's even necessary to search at all since we can go straight to the Robot node and test that one. 
The only reason for actually climbing the hierarchy that I can imagine is if it's desirable to apply the force as close to the cursor as possible. In which case an even more deep search should be done, in all directions, and shouldn't just involve node vicinity (with transforms it could very well not be physically near, despite being close node-wise). In short, reverting to the Robot node might be the more 'reasonable' option.

Basically, I don't know if the proposed solution is sufficient, or even overkill.

**Test world**
[force_logic.zip](https://github.com/cyberbotics/webots/files/5780206/force_logic.zip)

**Video**

https://user-images.githubusercontent.com/44834743/103865457-87609d00-50c4-11eb-82cb-9e3aa13fed37.mp4

Red have physics, green don't. The Stack behind is a Transform composed by a Box + Cylinder

https://user-images.githubusercontent.com/44834743/103865488-92b3c880-50c4-11eb-83e9-ffab10ea300b.mp4

### DIFF
--- a/src/webots/gui/WbView3D.cpp
+++ b/src/webots/gui/WbView3D.cpp
@@ -1945,8 +1945,10 @@ void WbView3D::mouseMoveEvent(QMouseEvent *event) {
       return;
 
     WbNode *node = dynamic_cast<WbNode *>(mPickedMatter);
-    WbSolid *selectedSolid = NULL;
-    while (node) {
+    if (!node)
+      return;
+    WbSolid *selectedSolid;
+    while (1) {
       selectedSolid = dynamic_cast<WbSolid *>(node);
       if (selectedSolid && selectedSolid->bodyMerger() != NULL)
         break;
@@ -1955,8 +1957,6 @@ void WbView3D::mouseMoveEvent(QMouseEvent *event) {
       if (!node || node->level() < 1)  // abort the search at the top of this node chain
         return;
     }
-    if (!selectedSolid)
-      return;
 
     Qt::MouseButtons buttons = event->buttons();
     bool forceButtonPressed = buttons == Qt::LeftButton;
--- a/src/webots/gui/WbView3D.cpp
+++ b/src/webots/gui/WbView3D.cpp
@@ -1943,9 +1943,21 @@ void WbView3D::mouseMoveEvent(QMouseEvent *event) {
     if (mDisabledUserInteractionsMap.value(WbAction::DISABLE_FORCE_AND_TORQUE, false))
       // user interaction disabled
       return;
-    WbSolid *const selectedSolid = dynamic_cast<WbSolid *>(mPickedMatter);
-    if (!selectedSolid || selectedSolid->bodyMerger() == NULL)
+
+    WbNode *node = dynamic_cast<WbNode *>(mPickedMatter);
+    WbSolid *selectedSolid = NULL;
+    while (node) {
+      selectedSolid = dynamic_cast<WbSolid *>(node);
+      if (selectedSolid && selectedSolid->bodyMerger() != NULL)
+        break;
+
+      node = node->parentNode();
+      if (!node || node->level() < 1)  // abort the search at the top of this node chain
+        return;
+    }
+    if (!selectedSolid)
       return;
+
     Qt::MouseButtons buttons = event->buttons();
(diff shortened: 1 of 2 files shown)


omichel:
I agree with you it makes sense to search upwards and not sideways.
Also we should not go straight up to the top Solid node (you mentioned the Robot node, but we should actually be more generic) as the force should be applied as close as possible to the cursor. This is especially useful if a robot or object has multiple joints (legs) and we want to apply the force/torque at the joint we click and not on the upper parts. So, I believe your approach so far seems good.
I will be happy to review this PR more in depth once its out of draft.

ad-daniel:
I've reworded the text a few times, as I kept changing my mind, so might be quite different now. Basically what I'm wondering is if node-vicinity suffices, with transforms involved they could be physically nowhere near despite being close node-wise, the force ending up affecting something else regardless.

Thinking of a spider-like robot, I suppose then a downwards search should be done as well, picking among the two directions the one with less jumps?

omichel:
I understand your point, but we should actually consider that a Solid node with no Physics (or a Group/Transform node) is somehow a part of the closest upper Solid node with Physics. The "real" object with which we want to physically interact with is actually a Solid node, including all its non-physics enabled children (which we consider to be part of the same mechanical body, e.g., in the same solid merger).

A downward search seems odd, as a Solid node without Physics **cannot** have children with Physics.

ad-daniel:
I was thinking of something like a spider-leg where the segments are made out of LEDs like the e-pucks body mentioned above, but I see, unless the Physics in the SolidDevice are active it wouldn't be possible to chain them like that anyway if the end-point is supposed to interact with the ground.

In which case the proposed solution should suffice.

src/webots/gui/WbView3D.cpp
```diff
@@ -1943,9 +1943,21 @@ void WbView3D::mouseMoveEvent(QMouseEvent *event) {
...
+      selectedSolid = dynamic_cast<WbSolid *>(node);
+      if (selectedSolid && selectedSolid->bodyMerger() != NULL)
+        break;
+
+      node = node->parentNode();
+      if (!node || node->level() < 1)  // abort the search at the top of this node chain
+        return;
+    }
+    if (!selectedSolid)
       return;
```
omichel, line 1959: I believe the logic is correct, however, the code could be slightly simplified:

```suggestion
    if (!node)
      return;
    WbSolid *selectedSolid;
    while (1) {
      selectedSolid = dynamic_cast<WbSolid *>(node);
      if (selectedSolid && selectedSolid->bodyMerger() != NULL)
        break;

      node = node->parentNode();
      if (!node || node->level() < 1)  // abort the search at the top of this node chain
        return;
    }
```

omichel (changes requested):

src/webots/gui/WbView3D.cpp
ad-daniel: Indeed, thank you.

omichel (changes requested):
Thank you.
We would also need an entry in the change log about this bug fix.

omichel (approved):
Thank you!

PR #2640, opened 2021-01-11
Mac keys - Update the-3d-window.md

correll:
There is no "Alt" key on a Mac. Pressing the control key allows to change from rotating the scene to translating the camera. This is not really necessary, as translation can also be achieved by pressing the pad with two fingers.

I'm able to zoom by pinching, but I'm not able to zoom out. None of the other Mac keys (option, command) help with this...

### DIFF
--- a/docs/guide/the-3d-window.md
+++ b/docs/guide/the-3d-window.md
@@ -32,8 +32,7 @@ If you click on the background, the camera will rotate around its own position.
 Dragging the mouse horizontally will rotate the camera around the viewing axis.
 Alternatively, the mouse wheel alone can also be used for zooming.
 
-> **Note**: If you are a Mac user with a single button mouse, hold the Alt key and press the mouse button to translate the camera according to the mouse motion.
-Hold the control key (Ctrl) down and press the mouse button to zoom / rotate the camera.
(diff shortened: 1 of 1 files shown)


omichel:
It seems the <kbd>alt</kbd> key on European Mac keyboards corresponds to the <kbd>option</kbd> key on US Mac keyboards. See some explanation [here](https://osxdaily.com/2016/05/02/where-option-key-apple-keyboards/). I believe this shortcut is useful on desktop Macs where you don't have a trackpad, so I would keep it, but add the "option" name in addition to "alt".

I was not aware about the trackpad capabilities, but after playing with it, I found out that you can perform all operations:
1. Rotate viewpoint: click with one finger and drag.
2. Translate viewpoint: click with two fingers and drag.
3. Zoom: slide two fingers (without clicking) downwards to zoom in and upwards to zoom out.

_Note_: pinching doesn't seems to work well for me. It seems that instead of pinching, sliding two fingers in parallel, up and down, does the trick for zooming in and out.

Could you try this and confirm that it works also for you?
After your confirmation, we will update the documentation accordingly.

correll:
I had the some findings, in particular sliding up and down with two fingers does the zoom trick. It took me quite some time. Maybe you can add keyboard commands like + and - too.

Best,
Nikolaus

omichel:
I fixed this in #2642.
Let me know if it looks better now.

PR #2639, opened 2021-01-11
Improve robot window documentation

lukicdarkoo:
Fixes #1055

**Documentation**
- https://cyberbotics.com/doc/guide/samples-howto?version=feature-robot-window-doc#custom_robot_window_simple-wbt
- https://cyberbotics.com/doc/guide/controller-plugin?version=feature-robot-window-doc#robot-window

**TODO**
- [x] Add a very basic robot window
- [x] Refactor the `projects/samples/howto` directory, so each example is in a dedicated folder
  - [x] Change the documentation accordingly
- [x] Delete extra Makefiles (e.g. `projects/samples/howto/plugins/physics/Makefile`) to minimize the noise
- [x] Fix the `custom_html_robot_window` example (it wasn't working)
- [x] Create a tutorial
- [x] Add to changelog
- [x] Verify regressions

PR #2649, opened 2021-01-13
Add branch placeholder to documentation

lukicdarkoo:
This PR adds a branch placeholder for documentation. See the original discussion here:
https://github.com/cyberbotics/webots/pull/2639#discussion_r556399502

Page example:
https://cyberbotics.com/doc/guide/controller-plugin?version=feature-doc-branch-placeholder


omichel:
It would be nice if we could reuse the same mechanism and/or syntax as with the Webots version, see https://github.com/cyberbotics/webots/blob/master/docs/js/viewer.js#L574-L584

docs/js/viewer.js
```diff
@@ -139,9 +139,11 @@ function redirectUrls(node) {
       continue;
     else if (href.startsWith('#'))
       addDynamicAnchorEvent(a); // on firefox, the second click on the anchor is not dealt cleanly
-    else if (href.startsWith('http')) // open external links in a new window
+    else if (href.startsWith('http')) {
+      // open external links in a new window
       a.setAttribute('target', '_blank');
-    else if (href.endsWith('.md') || href.indexOf('.md#') > -1) {
+      a.setAttribute('href', href.replaceAll('{version}', localSetup.branch));
```
stefaniapedrazzi, line 145: Note that other placeholders user the form `{{ <variable_name> }}` (e.g., `{{ webots.version.full }}`).
Won't be possible to use the same format? That would be more consistent and easier to use.

Maybe we could also replace the whole HTTPS URL prefix with the variable (`https://github.com/cyberbotics/webots/tree/<branch>`) given that it doesn't seem we need for anything else and this will simplify the MD files.

lukicdarkoo: I am on it. @omichel mentioned to me about the plugins, I completely overlooked it :/

omichel (approved):
Thank you.
It looks better like that.

stefaniapedrazzi:
I still think that, given that we are on it, we could further simplify the MD files and make them more readable by replacing ``https://github.com/cyberbotics/webots/tree/{{ setup.branch }}`` with a single variable (for example ``{{ github_webots_home }} `` or similar) and hard code the ``https://github.com/cyberbotics/webots/tree/`` part in `docs/js/showdown-extensions.js`

lukicdarkoo:
> by replacing `https://github.com/cyberbotics/webots/tree/{{ setup.branch }}` with a single variable (for example `{{ github_webots_home }} `

This could be useful. We also can make a collection of premade URLs, e.g.:
```js
urls: {
  github_webots_home: `https://github.com/cyberbotics/webots/tree/${setup.branch}`
}
``` 

> the `https://github.com/cyberbotics/webots/tree/` part in `docs/js/showdown-extensions.js`
I don't like this, e.g. I may want to reference the `master` branch specifically, and then I have to fight the extension. I would rather link paths automatically in that case. Like, if there is a string in the format `WEBOTS_HOME/some/path` it should be converted to a link.

omichel:
This branch seems to have conflicts now.

omichel (approved):
Thank you.
It looks good to me.

lukicdarkoo:
Since the implementation is done after Stefania's guidelines I believe it is safe to merge.

Issue #2493, opened 2020-12-02
Robot windows slows performance after closed / reset


Simon-Steinmann:
**Describe the Bug**
Opening the robot window (double click on robot) slows the performance of Webots. Closing or resetting the simulation does not return to the same performance level as before. Only a world reload does.

**Steps to Reproduce**
1. Open a world with a robot
2. Run simulation in fast mode and take a note of the realtime factor (RF)
3. open the robot window (double click on robot) - RF should decrease
4. Close robot window - RF should go up but not as high as in step 2
5. Reset simulation (back-arrow or ctrl-shift-T) - RF still not as high as in step 2
6. Reload world. RF should now be the same as in step 2

**Expected behavior**
Closing the robot window should turn off all background tracking / monitoring and there should be no impact of performance anymore

**Feature Suggestions:**
- Give the user the option to adjust the polling rate of the robot window (in a multiple of timesteps). This could allow for less performance impact
- Give the user the option to enable / disable what they want to monitor / controll. An "disable all" and "enable all" option would be great too.

stefaniapedrazzi:
Performance should be improved in #2655

PR #2655, opened 2021-01-14
Enhancement generic robot window

stefaniapedrazzi:
Address #2493.

Improvements:
* [x] do not update if the robot window is hidden/closed
    -> this fixes the simulation performance drop when the robot window is closed (even after simulation reset)
* [x] add enable/disable all button in the device tabs to enable/disable all the devices of a given type
* [x] add a `Settings` tab to set the robot window refresh rate and disable all the devices (of any type)
* [x] check if it is possible to speed up the tab initialization
* [x] try to improve the motor slider visibility -> moved to the left  if motor plot is not completely visible
* [x] evaluate disabling plot updates but not the device
* [x] improve the JS and C code
* [x] create library from C code to reuse it in other projects
* [x] reuse JS sources liking to the raw GitHub repo content URL. Webots will convert it automatically to the local absolute path if the Webots version matches.

device_buttons
settings_tab


Simon-Steinmann:
THANK YOU! This is fantastic. I only have one more suggestion: I think the slider for motor controls should be Either 
- on the left of the graph
or
- the graph should autoadjust its width, so the slider is always visible

Currently there is the issue, that the sliders cannot be controlled, if the Robot window is not wide enough. On a single monitor with IDE, Simulation, SceneTree and RobotWindow open, there is simply too little space.
image

lukicdarkoo:
> THANK YOU! This is fantastic. I only have one more suggestion: I think the slider for motor controls should be Either
> 
> * on the left of the graph
>   or
> * the graph should autoadjust its width, so the slider is always visible
> 
> Currently there is the issue, that the sliders cannot be controlled, if the Robot window is not wide enough. On a single monitor with IDE, Simulation, SceneTree and RobotWindow open, there is simply too little space.

Also, the left menu takes quite some space. Maybe, the left menu could be replaced by a simple drop-down menu and stick it to the top (`<select>` and  `<option>`)?

stefaniapedrazzi:
> Also, the left menu takes quite some space. Maybe, the left menu could be replaced by a simple drop-down menu and stick it to the top (<select> and <option>, instead of title)?

Note that it is already possible to close the left menu using the `x` button.

Simon-Steinmann:
Perhaps instead of an `x` a minimize or side triangle could be used. Might be more intuitive

stefaniapedrazzi:
For reusing the JS code of the generic robot window (or any JS code from the Webots package in general) we could maybe use this procedure:
* in the HTML file add the URL of the `https://raw.githubusercontent.com/cyberbotics/webots` (or another CDN) path of the Webots folder instead of the relative one
* in the WbRobotWindow class when loading it, we could replace the URLs referring to `https://raw.githubusercontent.com/cyberbotics/webots` with the local absolute path on the Webots installation

This way it should work both locally or online.
The only issue would be to link to the correct Webots tag in the develop and master branch because during developing time the tag doesn't exist.

omichel:
Yes, that's a good plan. Not sure how to address the problem of the URL with a tag or a reference to the current develop/master branches... Maybe we should commit the tag in the code (even though it doesn't exist) and have Webots to check somehow whether we are on a development environment and if yes, use the current branch instead of the tag.

stefaniapedrazzi:
> Maybe, the left menu could be replaced by a simple drop-down menu and stick it to the top

I think that it is important that the menu and all the devices tabs are always visible and users doesn't have to manually open the menu to select it. This will make it easier to use especially for new users.
Given that we already have a button to close the menu, advanced users can close the menu if they need more space for the plots.

> Perhaps instead of an x a minimize or side triangle could be used. Might be more intuitive

I checked various websites and templates, and the current symbols (the menu sign (three horizontal line) for opening the menu and the `x` for closing) are one of the most used standard. So I don't think that changing the symbol will make a big difference.
But I will make the `x` symbol bigger so that it will be more visible.

stefaniapedrazzi:
As reported by @lukicdarkoo in #2602, if there are many sensors of the same type, opening the corresponding tab takes lot of time.
This is because the values are added to the plot data even if another tab is selected and when they become visible it all the values needs to be drawn on the canvas.

A simple performance improvement would be to save the sensor data only if the plot is visible and discard them otherwise. This would mean that every time you select a device type you won't see anything at the beginning.
However we could set this behavior as the default but, if needed, add an option in the new settings tab where the user can specify if the plot data is updated when the plot itself is hidden.

stefaniapedrazzi:
I added a toggle button to select if the device checkbox enables/disables the plot update only or also the device.
When the mouse is over the toggle button, a sort of tooltip is displayed with a description.

robot_window_control_device

For the motor slider, I added resize event listener that moves the slider to the left of the plot in case the plot it is not completely visible. We can test it, otherwise the users still have the options to close the menu and undock the robot window in order to make the content more visible with a small robot window.

I addressed all the improvement tasks, but I will need to make some more tests (tested only on Linux for the moment).

resources/projects/libraries/robot_window_utils/generic_robot_window.c
```diff
@@ -0,0 +1,301 @@
...
+    return WB_NODE_ROTATIONAL_MOTOR;
+  if (strcmp(typeString, "Speaker") == 0)
+    return WB_NODE_SPEAKER;
+  if (strcmp(typeString, "TouchSensor") == 0)
+    return WB_NODE_TOUCH_SENSOR;
+  return WB_NODE_NO_NODE;
+}
+
+static void enable_device(WbDeviceTag tag, bool enable) {
+  WbNodeType type = wb_device_get_node_type(tag);
```
omichel, line 106:
```suggestion
  const WbNodeType type = wb_device_get_node_type(tag);
```

include/controller/c/webots/robot_window_utils/string_utils.h
```diff
@@ -0,0 +1,35 @@
...
+
+#ifndef STRING_UTILS_H
+#define STRING_UTILS_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+char *string_utils_strsep(char **stringp, const char *delim);
+char *string_utils_replace(char *orig, char *rep, char *with);
```
omichel, line 29: Those are generic string manipulation functions, not specific to the robot window. So, I would rename them with a `wbu_` prefix and would put them into the `utils` folder under the `string.h` filename.

include/controller/c/webots/robot_window_utils/generic_robot_window.h
```diff
@@ -0,0 +1,47 @@
...
+bool handle_generic_robot_window_messages(const char *message);
+void init_robot_window();
+void update_robot_window();
+bool robot_window_is_hidden();
+double robot_window_refresh_rate();
+bool robot_window_needs_update();
+
+// Utility functions to parse robot window message
+char *string_utils_strsep(char **stringp, const char *delim);
+char *string_utils_replace(char *orig, char *rep, char *with);
```
omichel, line 41: That seems duplicated from the `string_utils.h` include file.

include/controller/c/webots/robot_window_utils/generic_robot_window.h
```diff
@@ -0,0 +1,47 @@
...
+#endif
+
+void parse_device_command(char *token, char *tokens);
+bool parse_device_control_command(char *token, char *tokens);
+bool handle_generic_robot_window_messages(const char *message);
+void init_robot_window();
+void update_robot_window();
+bool robot_window_is_hidden();
+double robot_window_refresh_rate();
+bool robot_window_needs_update();
```
omichel, line 37: I would rename these functions with a `wbu_generic_robot_window_` prefix, like in `utils/default_robot_window.h` and move this include file to the `utils` folder.

resources/projects/libraries/robot_window_utils/string_utils.c
```diff
@@ -1,4 +1,4 @@
-#include "string_utils.h"
```
omichel, line 1: I would rename this file `string.c`.

src/webots/gui/WbRobotWindow.cpp
```diff
@@ -169,7 +171,19 @@ void WbRobotWindow::show() {
 }
 
 QString WbRobotWindow::formatUrl(const QString &urlString) {
-  const QUrl &url(urlString);
+  static QString repoUrl;
+  if (repoUrl.isEmpty()) {
+    // forge the GitHub URL to the current version of Webots
+    const QString webotsVersion = WbApplicationInfo::version().toString().replace(" revision ", "-rev");
+    repoUrl = "https://raw.githubusercontent.com/cyberbotics/webots/" + webotsVersion + "/";
```
omichel, line 178: Unfortunately, this will not work until we create the version tag: it won't work for the night builds for example.

omichel (commented):
Here is first code review.
I am now checking the behavior.

src/webots/gui/WbRobotWindow.cpp
stefaniapedrazzi: This link is automatically converted by Webots to the local file if the path found matches the current version.
So it works also before the tag is creates and even in the nightly builds.

omichel:
The spin-box doesn't look very nice on Windows:
spinbox
Maybe it is simpler to remove this spin-box and keep only a normal text input field?

stefaniapedrazzi:
>  Maybe it is simpler to remove this spin-box and keep only a normal text input field?

Yes. I will remove it so that it is simpler to maintain it on the different browsers.

omichel:
I found this checkbox not very intuitive to understand.
control_devices
Maybe if the label was saying something like "override controller" instead of "Control devices", it would easier to understand, isn't it? Also I would rephrase the legend with "If enabled, the devices will be enabled/disabled from their checkbox"

stefaniapedrazzi:
I think that these new utility functions specific for the generic robot window and derived should not be included in the controller library, as is the case for the `default_robot_window.h` functions.
But they should rather be included in a separate library as the `vehicle` one.

That's why I don't think they should be moved in the `include/controller/c/utils` folder but they should have their own folder (previously "robot_window_utils") whose name matches the library name.

About the string utilities I'm not sure if it will really make sense to add them to the libController too or we should include them only in the `robot_window_utils` library.

include/controller/c/webots/robot_window_utils/string_utils.h
stefaniapedrazzi: I added the `wbu_` prefix but I would not move them in the `utils` folder given that they should be distributed as a separate libraries and not included in the controller library.

stefaniapedrazzi:
I addressed all the comments but the one about moving the files in the `utils` folder.

omichel (changes requested):
I would move the `string.h` include file into the `utils` folder as it is not specific to robot windows and could be used also from controller programs.
Regarding the `generic_robot_window.h` include file, I would suggest the following clean-up / simplification:
- rename `robot_window_utils` to `robot_window`.
- rename `generic_robot_window.h` to `generic.h`
- move `utils/default_robot_window.h` to `robot_window/default.h`

stefaniapedrazzi:
I will rename the files and move `string.h` in the libController, but we cannot move `utils/default_robot_window.h` to `robot_window/default.h` because this has to be part of the libController.

stefaniapedrazzi:
I inspected and here are the changes required to move the `default_robot_window` functions to an external library (like `robot_window`):
* turn public the `robot_get_simulation_mode` function (currently only available in the `Supervisor` API)
* turn public or reimplement the `base64_encode` and `g_image` functions to handle images

docs/reference/robot-window-plugin.md
```diff
@@ -0,0 +1,156 @@
...
+    "height": <double>
+  }```
+* [TouchSensor](touchsensor.md):
+    ```
+  {
+    "sensorType": <"bumper"|"force"|"force-3d">
+  }```
+
+The `wbu_default_robot_window_update` function sends to the HTML robot window an update JSON messages, prefixed with the "update" label, containing the current devices measurements.
+Here is the sample structure of the configure message:
```
omichel, line 108:
```suggestion
Here is the sample structure of the update message:
```

docs/reference/robot-window-plugin.md
```diff
@@ -0,0 +1,156 @@
...
+      "azimuth": <double>
+    ]
+  }```
+* [RangeFinder](rangefinder.md):
+    ```
+  {
+    "image": "data:image/jpg;base64," + <data>
+ }```
+
+The `wbu_default_robot_window_set_images_max_size` function sets the maximum image size used to send rendering devices' image from the controller to the robot window.
```
omichel, line 155:
```suggestion
The `wbu_default_robot_window_set_images_max_size` function sets the maximum size for the images of rendering devices sent from the controller to the robot window.
```

docs/reference/robot-window-plugin.md
```diff
@@ -0,0 +1,156 @@
...
+If the size of an image is bigger than the maximum size, then the image is scaled before appending it to the update message.
```
omichel, line 156:
```suggestion
If the size of an image is bigger than the maximum size, then the image is scaled down before it is appended to the update message.
```

omichel (changes requested):
It looks very good to me.
But I have a few suggestions.

resources/projects/libraries/generic_robot_window/generic.c
```diff
@@ -0,0 +1,278 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <webots/plugins/robot_window/default.h>
+#include <webots/plugins/robot_window/generic_robot_window/generic.h>
```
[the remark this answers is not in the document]
stefaniapedrazzi: `generic_robot_window` is a separate library. 
I think that this extra layer is very important to make it more clear that it is a different library than the libController. Saving header files from different libraries in the same folder seems a bad practice.

docs/reference/c-string-api.md
```diff
@@ -0,0 +1,25 @@
...
+char *wbu_string_strsep(char **stringp, const char *delim);
+char *wbu_string_replace(char *value, char *before, char *after);
+```
+
+##### Description
+
+*parse and generate string messages*
+
+The `wbu_string_strsep` function locates, in the string referenced by the `stringp` pointer, the first occurrence of any character in the string `delim` and returns the substring found on the left of the delimiter.
+Additionally, the `stringp` pointer is updated so that it stores the location of the character after the delimiter character.
```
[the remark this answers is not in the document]
stefaniapedrazzi: Personally I found that the description [here](https://www.freebsd.org/cgi/man.cgi?strsep(3)) is much more clear than the  you [linked](https://man7.org/linux/man-pages/man3/strsep.3.html).
So I would keep the first sentence and then copy the "official" one we will link to.

stefaniapedrazzi:
I applied all the suggestions for the documentation.

resources/projects/libraries/generic_robot_window/generic.c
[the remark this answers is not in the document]
omichel: It refers to this robot window: `$WEBOTS_HOME/resources/projects/plugins/robot_windows/generic/`, so I would at least rename the folder so that the include file is `$WEBOTS_HOME/include/controllers/c/webots/plugins/robot_window/generic/generic.h` which seems more consistent, isn't it?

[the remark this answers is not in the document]
stefaniapedrazzi: No, it refers to `$WEBOTS_HOME/resources/projects/libraries/generic_robot_window/`.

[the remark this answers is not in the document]
omichel: OK, I see, it doesn't refer to a robot window, but to a library for building robot windows...
So, let's keep it like that.

omichel (dismissed):
Thank you.

omichel (approved):

Issue #2661, opened 2021-01-18
Improve performance of Python RangeFinder getRangeImage


WasabiFan:
**Is your feature request related to a problem? Please describe.**

The `getRangeImage` and `getRangeImageArray` APIs both return Python `list`s. For my 640x480 image, the former takes ~10ms as measured by `time.time()` (i.e., real-world time) and the latter takes ~20ms. That time is, presumably, spent allocating a huge list (or list of lists). For comparison, the C `wb_range_finder_get_range_image` is functionally instantaneous, since it just returns a pointer to the existing buffer.

In my application, I don't need a `list`. A list is a silly data structure for an image 🙂If I'm shooting for a 60Hz controller, and want to run real-time, I have a 16ms tick budget. Spending 10ms retrieving the data and then another 10ms loading it into numpy is, in itself, more than 16ms.

**Describe the solution you'd like**

I imagine the solution is to provide a new API in the Python interface.

One option would be to return a `bytes` object like the Camera `getRangeImage` does. It may be a bit strange to expose the raw float layout in a language that otherwise wouldn't, but either way I'd be sufficiently happy with it since I could stuff it into `numpy`. Unfortunately, this would still require at least one copy to get it into a `bytes` and then another to load it into numpy for my use-case.

Another option would be to provide an interface returning a type which inherently shares the underlying buffer; the `bytearray` interface added for point clouds in #2321 would be great. It still suffers from the same slight weirdness as above, though.

A third option would be to directly expose `numpy` arrays as first-class data structures. It _feels_ like that may be worth doing for cameras/range-finders/etc. My _guess_ is that most users are loading the data into `numpy` either way if they're writing in Python, although I haven't done any research there.

**Describe alternatives you've considered**

Writing my controller in C or C++ could be done, but that's a silly reason to migrate. I could also potentially write just a shim for this portion in C and then invoke it via FFI. That's additional toolchain overhead which I'd rather avoid.

WasabiFan:
The above comparison was not exactly accurate: the original experimentation with the C API above was with a lower resolution. With the same 640x480 setting, in my testing the C API takes 1-3ms, typically, to grab a reading, with sporadic spikes to 20ms+. These spikes are enough to significantly increase the mean time to take a measurement. I also see these spikes in the Python API.

[...]

PR #2664, opened 2021-01-19
Expose new API for accessing RangeFinder's raw float buffer in Python

WasabiFan:
**Description**

This PR introduces a new `data_type` keyword parameter on the `getRangeImage` method. The default value, `list`, preserves its existing behavior. The new alternative, `buffer`, returns a `bytes` object with a copy of the machine-dependent `float` buffer.

The behavior of `getRangeImageArray` is unchanged. I opted to add the new functionality to `getRangeImage` rather than `getRangeImageArray` because a byte buffer more closely resembles a 1D list than 2D. 🤷 

**Related Issues**

Fixes issue #2661. There's a full description in that issue.

API is modelled after #2321.

**Tasks**

I don't know what tasks I'm supposed to put here, but... I added the option and updated the documentation.

I have _not_ introduced a changelog entry.

**Documentation**

This MR is sourced from my fork for politeness; I believe I could push the branch to the upstream repo and change the source if that's what is preferred.

**Additional context**

<details>
<summary>Not-very-scientific test controller</summary>

```python
from controller import Robot

import time
import numpy as np
import cv2

robot = Robot()

rf = robot.getDevice("range-finder")
rf.enable(int(robot.getBasicTimeStep()))

i = 0
    if i % 2 == 0:
        start = time.time()
        image = rf.getRangeImage(data_type="list")
        image_np = np.array(image).reshape(rf.getHeight(), rf.getWidth())
        end = time.time()
        #cv2.imwrite("list.png", np.array(image_np * 100, dtype=np.uint8))
        print(f"List time: {(end-start)*1000:.2f} ms")
    else:
        start = time.time()
        image = rf.getRangeImage(data_type="buffer")
        image_np = np.frombuffer(image, dtype=np.float32).reshape(rf.getHeight(), rf.getWidth())
        end = time.time()
        #cv2.imwrite("buffer.png", np.array(image_np * 100, dtype=np.uint8))
        print(f"Buffer time: {(end-start)*1000:.2f} ms")
    i += 1

```
</details>

On my machine, with a 640x480 image (formatted as `median` / `mean`):

-|list|buffer
---|---|---
function call only | 8ms / 8.7ms | 5ms / 5.5ms
function call plus numpy array | 22ms / 24.5ms | 7ms / 7.6ms

I'm not 100% satisfied with this performance (7ms is still rather high for what should, ideally, do effectively zero work) but I figure it's worth opening the PR for discussion.

### DIFF
--- a/src/controller/python/controller.i
+++ b/src/controller/python/controller.i
@@ -753,19 +753,29 @@ class AnsiCodes(object):
 //  RangeFinder
 //----------------------------------------------------------------------------------------------
 
-%typemap(out) float * {
-  int width = arg1->getWidth();
-  int height = arg1->getHeight();
-  int len = width * height;
-  if ($1) {
-    $result = PyList_New(len);
-    for (int x = 0; x < len; ++x)
-      PyList_SetItem($result, x, PyFloat_FromDouble($1[x]));
-  } else
-    $result = Py_None;
-}
-
 %extend webots::RangeFinder {
+
+  PyObject *__getRangeImageList() {
+    const float *im = $self->getRangeImage();
+    int width = $self->getWidth();
+    int height = $self->getHeight();
+    int len = width * height;
+    if (im) {
+      PyObject *ret = PyList_New(len);
+      for (int x = 0; x < len; ++x)
+        PyList_SetItem(ret, x, PyFloat_FromDouble(im[x]));
+      return ret;
+    } else {
+      return Py_None;
+    }
+  }
+
+  PyObject *__getRangeImageBuffer() {
+    const float *im = $self->getRangeImage();
+    const char *im_bytes = (const char *)im;
+    const int size = $self->getWidth() * $self->getHeight() * sizeof(float);
+    return PyBytes_FromStringAndSize(im_bytes, size);
+  }
 
   PyObject *getRangeImageArray() {
     const float *im = $self->getRangeImage();
@@ -785,6 +795,17 @@ class AnsiCodes(object):
     }
     return ret;
   }
+
+  %pythoncode %{
+    def getRangeImage(self, data_type='list'):
+      if data_type == 'list':
+        return self.__getRangeImageList()
+      elif data_type == 'buffer':
+        return self.__getRangeImageBuffer()
+      else:
+        print("Error: `data_type` cannot be `{}`! Supported values are 'list' and 'buffer'.".format(data_type), file=sys.stderr)
+        return None
+  %}
 
   static PyObject *rangeImageGetValue(PyObject *im, double minRange, double maxRange, int width, int x, int y) {
     if (!PyList_Check(im)) {
--- a/src/controller/python/controller.i
+++ b/src/controller/python/controller.i
@@ -757,9 +757,9 @@ class AnsiCodes(object):
 
   PyObject *__getRangeImageList() {
     const float *im = $self->getRangeImage();
-    int width = $self->getWidth();
-    int height = $self->getHeight();
-    int len = width * height;
+    const int width = $self->getWidth();
+    const int height = $self->getHeight();
+    const int len = width * height;
     if (im) {
       PyObject *ret = PyList_New(len);
       for (int x = 0; x < len; ++x)
--- a/docs/reference/rangefinder.md
+++ b/docs/reference/rangefinder.md
@@ -483,7 +483,7 @@ namespace webots {
 from controller import RangeFinder
 
 class RangeFinder (Device):
-    def getRangeImage(self):
+    def getRangeImage(self, data_type='list'):
     def getRangeImageArray(self):
     @staticmethod
     def rangeImageGetDepth(image, width, x, y):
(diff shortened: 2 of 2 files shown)


lukicdarkoo:
Thank you for the nice PR!

I have just briefly tested the performance of a C controller, and the C controller is indeed faster (around 1.5ms, so around 3.5ms difference). The only bottleneck I see is `PyBytes_FromStringAndSize`, but I assume it shouldn't be that slow. Do you have any idea?

If there is no obvious way to solve it we can merge it and investigate further in another PR.

lukicdarkoo:
Here is the reason:
> Return a new bytes object with a copy of the string v as value and length len on success, and NULL on failure. If v is NULL, the contents of the bytes object are uninitialized.

https://docs.python.org/3/c-api/bytes.html#c.PyBytes_FromStringAndSize

The only way to avoid it and I am aware of is Python Buffer Protocol. However, maybe there is something in SWIG to should simplify the implementation.

WasabiFan:
@lukicdarkoo take a look at my findings posted in #2661 — there's some performance info there regarding the function you mentioned.

lukicdarkoo:
@WasabiFan Sorry, I missed it :disappointed: 

> I implemented both the bytes and bytearray variants and, evidently

Do you have the `bytearray` implementation somewhere so I can check it?

If we make the `bytearray` version work faster than `bytes` then we can keep both of them. Even though Python doesn't own data in `bytearray` there could be use-cases in which `bytearray` is preferable.

WasabiFan:
I didn't keep a copy, but the version I tested was a direct copy of the `__getRangeImageBuffer` function which is in my change set here, but with `PyByteArray_FromStringAndSize` instead of `PyBytes_FromStringAndSize`. I'd certainly appreciate any thoughts on what I may have missed.

lukicdarkoo:
It seems that `PyByteArray_FromStringAndSize` doesn't share a buffer:
https://github.com/python/cpython/blob/e8e66eab941b983b6e85cd0d57cd45838880c568/Objects/bytearrayobject.c#L131-L139

I wanted to add a custom buffer type that exposes `setpointer` method:
https://gist.github.com/lukicdarkoo/40684dbffe234796376dc001d2fc4cfb
but I gave up in the end because it looked overengineered.

A possibly better solution would be to export `numpy` objects directly:
https://numpy.org/devdocs/reference/swig.interface-file.html

WasabiFan:
> It seems that PyByteArray_FromStringAndSize doesn't share a buffer:

Interesting, and surprising!

> I wanted to add a custom buffer type that exposes setpointer method:

That seems like a very sensible solution... I agree with your assessment though. Integrating something like that into the controller library would just add more surface area for strange maintenance edge-cases. I would be a bit surprised if there aren't existing libraries which expose a wrapper like this, but I'm not seeing anything from a cursory search.

> A possibly better solution would be to export numpy objects directly:

Agreed in principle. Is this something that the project as a whole is open to? In particular:
- NumPy would now be a build-time dependency
- Either only this one function would be able to return a NumPy array, or it would ideally be retrofit into e.g. `Lidar` and `Camera`

lukicdarkoo:
> Integrating something like that into the controller library would just add more surface area for strange maintenance edge-cases. I would be a bit surprised if there aren't existing libraries which expose a wrapper like this, but I'm not seeing anything from a cursory search.

Exactly!

> Agreed in principle. Is this something that the project as a whole is open to? In particular:
> 
> * NumPy would now be a build-time dependency
> * Either only this one function would be able to return a NumPy array, or it would ideally be retrofit into e.g. `Lidar` and `Camera`

If I am not mistaken, it would be just a matter of including `numpy.i`. `numpy.i` is built on top of Python Buffer Protocol, so it should not be a big issue.

The idea behind `getRangeImage(data_type='list')` is that we can easily add an arbitrary number of data types without breaking backward compatibility. Therefore, for getting numpy buffer we could use `getRangeImage(data_type='numpy')`.

@omichel We can merge this one, and open another PR for potential numpy integration?

omichel:
> @omichel We can merge this one, and open another PR for potential numpy integration?

Sure. Thank you for the investigation and great contribution in the works!

src/controller/python/controller.i
```diff
@@ -753,20 +753,30 @@ class AnsiCodes(object):
...
-    $result = Py_None;
-}
-
 %extend webots::RangeFinder {
 
+  PyObject *__getRangeImageList() {
+    const float *im = $self->getRangeImage();
+    int width = $self->getWidth();
+    int height = $self->getHeight();
+    int len = width * height;
```
omichel, line 762:
```suggestion
    const int width = $self->getWidth();
    const int height = $self->getHeight();
    const int len = width * height;
```

omichel (changes requested):

WasabiFan:
Thanks for the review and comments; I'll take a look tomorrow!

> If I am not mistaken, it would be just a matter of including numpy.i. numpy.i is built on top of Python Buffer Protocol, so it should not be a big issue.

I am just on my phone at the moment so can't grab the full details, but earlier I looked at the definition of that SWIG file and as far as I could tell it expanded to use functions from [here](https://numpy.org/devdocs/reference/c-api/array.html). I'm not familiar with SWIG but as I understand it it's purely a preprocessor step, so if it uses those functions I think we'll need to be able to link against numpy.

omichel (approved):
Thank you.

lukicdarkoo:
@WasabiFan Feel free to click the merge the `Squash and merge` button :)

Issue #2660, opened 2021-01-17
Lidar Point Cloud does not work


it14149:
**Describe the Bug**
The Lidar Cloud Point does not properly work during the simulation. It will start and stop showing the lidar points at "random" intervals. This happens in both the lidar.wbt sample and any new simulation I've tried.

**Steps to Reproduce**
1. Open the lidar.wbt sample from Help --> Webots Guided Tour --> Devices Tour --> lidar.wbt
2. Run the simulation

**Expected behavior**
The Lidar Cloud Point should not stop working. It should constantly show the lidar points.

**Video**
https://user-images.githubusercontent.com/12613442/104852089-5d4b7e00-5901-11eb-85e7-838dd85b9a8d.mp4

**System**
 - Operating System: Windows 10 Pro
 - Graphics Card: NVIDIA GeForce GTX 1050

**Additional context**
In the video provided it seems that the cloudpoint works at certain places (in the center of the map and momentarily near the bottom green object which implies it is not random.

omichel:
I believe we can close this issue as it was fixed in #2666.

PR #2666, opened 2021-01-19
Fix lidar point cloud visualisation issue

ad-daniel:
**Related Issues**
This pull-request addresses issue #2660

**Description**
After changing lidar out-of-range behavior in PR #2509, vertexes can have INF components, in which case they shouldn't be added to the mesh for rendering, as it wouldn't make much sense.

PS: As to why having a single vertex that contains an INF component in an otherwise valid cloud of points prevents all others from being drawn wasn't investigated in this PR. It's normal or should be addressed in a different issue? 

**Video**

https://user-images.githubusercontent.com/44834743/105106992-58c8c600-5ab7-11eb-813b-91c1baff1497.mp4

https://user-images.githubusercontent.com/44834743/105107008-5ebea700-5ab7-11eb-8c50-511c2afb3fbb.mp4

**Test**
Validated only on ubuntu 20.04.

### DIFF
--- a/docs/reference/changelog-r2021.md
+++ b/docs/reference/changelog-r2021.md
@@ -15,6 +15,7 @@ Released on XX Xth, 2021.
     - Fixed a strategy used to find a MATLAB executable in the `PATH` environment variable ([#2624](https://github.com/cyberbotics/webots/pull/2624)).
     - Fixed external force/torque logic such that the closest dynamic Solid ancestor is picked if the selected one lacks it ([#2635](https://github.com/cyberbotics/webots/pull/2635)).
     - Fixed the [robot window example](../guide/samples-howto.md#custom_robot_window-wbt) ([#2639](https://github.com/cyberbotics/webots/pull/2639)).
+    - Fixed visual bug where the [Lidar](lidar.md) point cloud disappears when out-of-range points are present ([#2666](https://github.com/cyberbotics/webots/pull/2666)).
   - Cleanup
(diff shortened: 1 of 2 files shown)


omichel:
Thank you for this new great contribution.
However, it seems that `cppcheck` is not happy with it, from the [CI tests](https://github.com/cyberbotics/webots/pull/2666/checks?check_run_id=1731562639):
```
src/webots/nodes/WbLidar.cpp:444:45: error: The address of local variable 'x' is accessed at non-zero index. [objectIndex]
        if (isinf(vertex[0]) || isinf(vertex[1]) || isinf(vertex[2]))
                                            ^
src/webots/nodes/WbLidar.cpp:443:31: note: Address of variable taken here.
        const float *vertex = &pointArray()[k * resolution + l].x;
                              ^
src/webots/nodes/WbLidar.cpp:444:45: note: The address of local variable 'x' is accessed at non-zero index.
        if (isinf(vertex[0]) || isinf(vertex[1]) || isinf(vertex[2]))
```

ad-daniel:
yes, oddly enough cppcheck on linux didn't give any error, only macos does. In practice in this case the modification doesn't really bring much as in the very next line `wr_dynamic_mesh_add_vertex` assumes the data is contiguous anyway. Whatever the case, it shouldn't give an error now.

omichel (changes requested):
Thank you.
Unfortunately, we don't have the same versions of ccpcheck on the different CI machines and the one of the mac may differ from the one on Linux.
Another acceptable way of fixing it would have been to add a special comment to disable the cppcheck warning, which you be OK as well.
Anyway, I believe we still need a line in the change log (as this PR somehow fixes a regression).

omichel (approved):
Thank you!

PR #2668, opened 2021-01-20
proto2mesh

omichel:
**Description**
This script aims at converting all the PROTO files distributed in Webots to replace the IndexedFaceSet nodes with Mesh nodes referring to OBJ files.

The benefits are manifold:
- The loading time of the PROTO files should be greatly improved.
- The readability of the PROTO file will be improved (way smaller files).
- The OBJ files could be edited separately (in Blender or other 3D editor) without having to re-export to VRML97 and copy/paste the IndexedFaceSet in the PROTO file.
- The same OBJ files could be used by different PROTO files.

**Preliminary results:** loading 10 UR3e robots using OBJ files take 12 seconds on my computer while it takes 27 seconds to load 10 original UR3e robots using IndexedFaceSet. I suppose the benefit comes from the fact that the creaseAngle doesn't need to be computed for OBJ files (which is a pretty intensive computation).

This PR includes an example of converted PROTO file: the RubberDuck.proto.
It was also successfully tested against the UR3e robot PROTO file.
Once accepted we should open new PRs to convert more complex PROTO files, including robots.


Simon-Steinmann:
image
The converted spot robot has some issues. This came up during conversion:
```Bash
  Processing mesh BOTTOM
    Removed 11 duplicate vertex coordinate indices
    Removed 11 duplicate vertex coordinates
    Removed 907 duplicate texture coordinate indices
    Removed 316 duplicate texture coordinates
    Computing normals from creaseAngle
.\proto2mesh.py:171: RuntimeWarning: invalid value encountered in true_divide
  normalized = n / np.sqrt(np.sum(n**2))
```

Simon-Steinmann:
MultiThreaded works on linux with no issues. I think Windows has some issues with terminating processes.

omichel:
To clean-up things, I would propose that all the resources (textures, meshes, sounds, etc.) of a PROTO file would be placed into a folder named after this PROTO file:

```
RubberDuck.proto
RubberDuck/beak.obj
RubberDuck/body.obj
RubberDuck/eye.obj
RubberDuck/eye.jpg
```

Simon-Steinmann:
meshes and textures should perhaps be separated. We also have to make sure, no issues arise, when BoundingObject has a mesh. These should either be placed in a seperate folder, or  have a suffix in the name. This might not be that relevant for the existing proto files, but looking forward, this should probably added. As this will probably be the base for a new version of the urdf2webots tool

My suggestion:
```
protos/myRobot/myRobot.proto
protos/myRobot/meshes/Mesh1.obj
protos/myRobot/meshes/Mesh1_col.obj
protos/myRobot/textures/eye.jpg
```

omichel:
I would not define a hard rule to separate meshes and textures as in simple cases (like the RubberDuck), it is simpler to have them altogether whereas when there is a large number of meshes and textures, we might need to create one texture folder per mesh or one texture folder for all the meshes. Also, in the case of the URxe robots, textures are shared across robots, but not meshes are not. Therefore the textures should be at a upper level in a common folder.

I tested this script again with the UR3e robot and although it takes long to execute, the result is good and many duplicate 3D data is discarded:
```
python proto2mesh.py --input=../../projects/robots/universal_robots/protos/UR3e.proto
Proto 2 mesh proto converter by Simon Steinmann & Olivier Michel
  Processing mesh screw
    Removed 1 duplicate vertex coordinates
    Removed 781 duplicate texture coordinate indices
    Removed 268 duplicate texture coordinates
    Computing normals from creaseAngle
  Processing mesh mesh1
    Computing normals from creaseAngle
  Processing mesh mesh2
    Computing normals from creaseAngle
  Processing mesh mesh3
    Computing normals from creaseAngle
  Processing mesh mesh4
    Computing normals from creaseAngle
  Processing mesh mesh5
    Computing normals from creaseAngle
  Processing mesh mesh6
    Removed 2 duplicate vertex coordinates
    Removed 47868 duplicate texture coordinate indices
    Removed 9764 duplicate texture coordinates
    Computing normals from creaseAngle
  Processing mesh mesh7
    Computing normals from creaseAngle
  Processing mesh mesh8
    Computing normals from creaseAngle
  Processing mesh mesh9
    Removed 1 duplicate vertex coordinate indices
    Removed 1 duplicate vertex coordinates
    Computing normals from creaseAngle
  Processing mesh mesh10
    Computing normals from creaseAngle
  Processing mesh mesh11
    Removed 1 duplicate vertex coordinates
    Removed 37158 duplicate texture coordinate indices
    Removed 7598 duplicate texture coordinates
    Computing normals from creaseAngle
  Processing mesh mesh12
    Computing normals from creaseAngle
  Processing mesh mesh13
    Computing normals from creaseAngle
  Processing mesh mesh14
    Computing normals from creaseAngle
  Processing mesh mesh15
    Computing normals from creaseAngle
  Processing mesh mesh16
    Removed 15239 duplicate texture coordinate indices
    Removed 3250 duplicate texture coordinates
    Computing normals from creaseAngle
  Processing mesh mesh17
    Computing normals from creaseAngle
  Processing mesh mesh18
    Computing normals from creaseAngle
  Processing mesh mesh19
    Computing normals from creaseAngle
  Processing mesh mesh20
    Computing normals from creaseAngle
  Processing mesh mesh21
    Computing normals from creaseAngle
  Processing mesh mesh22
    Computing normals from creaseAngle
  Processing mesh mesh23
    Computing normals from creaseAngle
  Processing mesh mesh24
    Removed 12868 duplicate texture coordinate indices
    Removed 2654 duplicate texture coordinates
    Computing normals from creaseAngle
  Processing mesh mesh25
    Computing normals from creaseAngle
  Processing mesh mesh26
    Computing normals from creaseAngle
  Processing mesh mesh27
    Computing normals from creaseAngle
  Processing mesh mesh28
    Computing normals from creaseAngle
  Processing mesh mesh29
    Removed 19645 duplicate texture coordinate indices
    Removed 4040 duplicate texture coordinates
    Computing normals from creaseAngle
  Processing mesh mesh30
    Computing normals from creaseAngle
  Processing mesh mesh31
    Computing normals from creaseAngle
  Processing mesh mesh32
    Computing normals from creaseAngle
  Processing mesh mesh33
    Removed 17747 duplicate texture coordinate indices
    Removed 3674 duplicate texture coordinates
    Computing normals from creaseAngle
  Processing mesh mesh34
    Removed 2 duplicate vertex coordinate indices
    Removed 2 duplicate vertex coordinates
    Computing normals from creaseAngle
Done
```

I know the script doesn't work quite well with the Spot robot, but I believe this problem should be addressed in a different PR (the one targeting the Spot).

So, I would suggest that we merge this PR now and fix the script as we meet problems in subsequent PRs.

Simon-Steinmann:
Sounds good to me. I think adhering to stricter protos formatting rules for the official models is the better route. That will make future work also easier.

stefaniapedrazzi (changes requested):
I tested with a couple of models and it works fine for plain meshes.

But it fails on PROTO models using lua for defining IndexedFaceSet (for example the `TinkerbotAxle.proto`).
If possible it would be better to detect it and abort the current mesh replacement instead of aborting the whole file (and folder) conversion.

Here are a couple of other comments:
* I added comment only on one script file but the same comments also apply to the other file.
* It would be better to add execution right to the new python script files.
    The existing `proto2multi.py` already in the folder has execution bit set.
* An entry should also be added on the ChangeLog at least mentioning the changed to the  RubberDuck model but also mentioning the script.
* Sometimes the alignment of the geometry field in the newly create file is not correct, but this is not a big issue.

scripts/proto_splitter/proto2mesh.py
```diff
@@ -0,0 +1,453 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Author Simon Steinmann (https://github.com/Simon-Steinmann)
+#
+
+
+# Automatic PROTO to multi-file PROTO trimesh extractor
```
stefaniapedrazzi, line 21: it would be good to add here what is the difference between this and the `proto2mesh_multiThreaded.py` script and when a user should use one or the other script.

omichel: I will actually remove this script from this PR as @Simon-Steinmann suggested he will actually merge it with the main script in a separate PR.

omichel:
Thank you for the review. I believe I addressed all the points you mentioned.
However, I was not able to reproduce the geometry align problem you mentioned.

omichel:
I could finally reproduce and fix the geometry alignment problem.

DavidMansolino:
Very nice idea and PR 🎉 

> Preliminary results: loading 10 UR3e robots using OBJ files take 12 seconds on my computer while it takes 27 seconds to load 10 original UR3e robots using IndexedFaceSet. I suppose the benefit comes from the fact that the creaseAngle doesn't need to be computed for OBJ files (which is a pretty intensive computation).

In that specific case, I think the caching mechanism is responsible also (and probably more than the creaseAngle) for the speed-up, as for IndexedFaceSet the hash is based on the complete mesh so the whole IndexedFaceSet has to be parsed whereas with the Mesh node, only the path of the mesh file has to be parsed.

Note also that you might want to add a copyright as a comment header to avoid other people using these meshes outside of Webots without Cyberbotics to agree (meshes are one of the most important assets of the PROTOs IMO).

omichel:
> In that specific case, I think the caching mechanism is responsible also (and probably more than the creaseAngle) for the speed-up, as for IndexedFaceSet the hash is based on the complete mesh so the whole IndexedFaceSet has to be parsed whereas with the Mesh node, only the path of the mesh file has to be parsed.

Yes, good point. I didn't make any measurement to see where is the gain.

> Note also that you might want to add a copyright as a comment header to avoid other people using these meshes outside of Webots without Cyberbotics to agree (meshes are one of the most important assets of the PROTOs IMO).

Yes,I believe we will have to check case-by-case what should be the license for these OBJ files. It turns out that many of them were imported from elsewhere, in which case we should inherit their license as well.

docs/reference/changelog-r2021.md
```diff
@@ -4,6 +4,8 @@
 Released on XX Xth, 2021.
 
   - Enhancements
+    - Added a script to convert PROTO files to use [Mesh](mesh.md) nodes instead of [IndexedFaceSet](indexedfaceset.md) nodes to speed-up loading times, improve PROTO readability and maintenance ([#2668](https://github.com/cyberbotics/webots/pull/2668)).
+    - Converted the several PROTO files to use [Mesh](mesh.md) nodes ([#2668](https://github.com/cyberbotics/webots/pull/2668)).
```
stefaniapedrazzi, line 8:
```suggestion
    - Converted several PROTO files to use [Mesh](mesh.md) nodes ([#2668](https://github.com/cyberbotics/webots/pull/2668)).
```

stefaniapedrazzi (changes requested):
Thank you. it remains just a minor typo in the changelog to fix.

What about the problem with PROTO files containing lua code?
Currently this script can only be applied as-is to a small subset of PROTO files and it will fail most of the times when called on a folder.

omichel:
> What about the problem with PROTO files containing lua code?
> Currently this script can only be applied as-is to a small subset of PROTO files and it will fail most of the times when called on a folder.

Yes, this is a limitation of this script. However, we don't aim at having a bullet proof script for converting all PROTO files in a raw (which would be difficult to achieve). The goal is to convert one-by-one most of the existing PROTO file, and then the script will become useless as we will avoid to create PROTO with large IndexedFaceSet nodes in the future. So, this script is just a temporary helper, which doesn't have to be perfect and should help us converting the PROTO files efficiently.

stefaniapedrazzi (approved):
Ok. It was just to mention that most of the files that I tried to convert failed.

omichel:
> Ok. It was just to mention that most of the files that I tried to convert failed.

Yes, the plan is to improve the script when we will face problems, or to work around problems manually, if it's quicker...

Simon-Steinmann:
With my new multithreaded version, only a few files throw errors. Only one due to LUA in the indexedFaceSet.

PR #2670, opened 2021-01-21
Fixed wrong indent

omichel:
Minor layout issue.
https://cyberbotics.com/doc/guide/the-user-interface?version=doc-layout-indent
I took advantage of this PR to improve the documentation of normals display.


lukicdarkoo (approved):
Maybe we should fix it as it may hide the other problems (the tests fails before other tests are executed)

PR #2673, opened 2021-01-22
Fix #2663 memory use on Display node

### LINKED ISSUE #2663: Increasing memory consumption over iterative reset simulation
I'm currently developing an reinforcement learning algorithm with Webots R2020b and recently has happened that the memory used by the webots-bin process reached out 5GB and more in some cases.
The scenario includes three robots Nodes and one custom proto Node, it is and ongoing project and still need some changes, I create and special scenario for replication called memory_check.wbt at in [the following repository](https://github.com/angel-ayala/gym-webots-fire)

**Steps to Reproduce**
1. Load the memory_check.wbt world.
2. Check the initial memory used by the webots-bin process
3. Run the simulation at fast as possible (CTL+4)
4. Iterative reset the simulation about 10 or 20 times
5. Check the memory increasing while the simulation is being reset.

As the problem it is continuously executed, it seems that the iterative reset simulation process does not leave enough time for the garbage collector algorithm to clear the memory or something like that, but after several times, as epochs needed in a RL problem the memory consumed by the simulator leaves with no memory to use by the algorithm freezing out and as consequently  ends the training process.

**System**
Different configuration system were used but.
 - Operating System: Manjaro (Arch based Linux distro), Linux kernel 5.10 x86_64
 - Graphics Card: AMD Radeon R5 M330 5GB.
 And
 - Operating System: Manjaro (Arch based Linux distro), Linux kernel 5.4 x86_64
 - Graphics Card: NVIDIA GeForce RTX 2060 6 GB.

Same results

angel-ayala:
Display's image loaded in memory was not properly de-allocated, when the simulation is reset iteratively, the Display node increased the memory use.

### DIFF
--- a/docs/reference/changelog-r2021.md
+++ b/docs/reference/changelog-r2021.md
@@ -7,6 +7,7 @@ Released on XX Xth, 2021.
     - Don't display warnings for recent Intel and AMD graphics cards ([#2623](https://github.com/cyberbotics/webots/pull/2623)).
     - Rework of car meshes to have more realistic rear lights for Mercedes Benz, Lincoln, Citroen, BMW and Range Rover models ([#2615](https://github.com/cyberbotics/webots/pull/2615)).
   - Bug fixes
+    - Fixed memory leak in [Display](display.md) image ([#2663](https://github.com/cyberbotics/webots/issues/2663)).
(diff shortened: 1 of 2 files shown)


omichel (approved):
Thank you!

Issue #2658, opened 2021-01-15
Simulation reset causes non-reproducible simulations


lukicdarkoo:
**Describe the Bug**
After the simulation reset, the behavior of the robots is not exactly the same. There are very tiny deviations, but enough to consider the simulations non-reproducible.

**Steps to Reproduce**
1. Download [reproducibility.zip](https://github.com/cyberbotics/webots/files/5841388/reproducibility.zip)
2. Start the world (let it run for 2-3 seconds)
3. Reset the simulation (let it run for 2-3 seconds)
4. There should be two files in the controller's directory, one from step 2 and the other one from step 3. Compare them. Only the first row should be the same.

With PROTOs (this may be a better example):
1. Download [reproducability_proto.zip](https://github.com/cyberbotics/webots/files/5856675/reproducability_proto.zip)
2. `cd controllers/position_tracker`
3. `webots ../../worlds/reproducability_proto.wbt`
4. Put it in the fast mode and reset the simulation like 20 times
5. `python3 check.py` to check for deviations

ad-daniel:
I can't seem to replicate it, maybe I'm misunderstanding

- if you mean the box behaves differently after each call of `node.restartController()` in the supervisor then that's normal as only the controller is reset, not the physics. Calling `node.resetPhysics()` after `node.restartController()` gets the job done
![resetPhysics](https://user-images.githubusercontent.com/44834743/104950656-5c7a2100-59c1-11eb-9fcc-9ca6798462d6.png)


- if instead you mean that manually resetting the world multiple times after letting it run a bit generates different CSVs, then I can't replicate on Ubuntu 20.04 (compiling from the latest master branch). PS: here I use your ZIP as is, without resetting the node physics
![manualReset](https://user-images.githubusercontent.com/44834743/104950665-62700200-59c1-11eb-8411-951823769c02.png)
can't see much but here's the difference between the later CSVs and the first one
![manualResetDiff](https://user-images.githubusercontent.com/44834743/104950672-6734b600-59c1-11eb-900d-17cc424d9596.png)

lukicdarkoo:
Try to restart the world manually using:
![image](https://user-images.githubusercontent.com/2135826/104964588-f9968300-59dc-11eb-9d5e-9bdb17e44241.png)

Then, compare two generated files using simple diff (instead of plotting).

lukicdarkoo:
[...]

PR #2684, opened 2021-01-29
fix reproductibility issue when resetting

ad-daniel:
This is gonna be a little long winded but it's kinda necessary.

**Related issue**
This fix addresses issue #2658

**Test world**
Same as #2658, just printing to console instead of to file for simplicity. Results are the same in this case.

**Description**
To summarize, the situation is :
- ROBOT_A (is a supervisor)
  - has a controller `position_tracker`
  - this controller spawns ROBOT_B and tracks it
- ROBOT_B (is a supervisor)
  - has a controller  `dummy_controller`
  - this controller applies a force to itself

After looking into it the lack of reproducibility boils down to these 2 situations that should be equal but aren't. 

**Situation (1)**
1. load scenario
2. step, step, step, step (call this output RESULTS_1)
3. reset from GUI
4. step, step, step, step (call this output RESULTS_2)
5. reset from GUI
6. step, step, step, step (call this output RESULTS_3)

**conclusion:** no issues, RESULTS_1 = RESULTS_2 = RESULTS_3

**Situation (2)**

1. load scenario
2. run simulation (call this output RESULTS_1)
3. reset from GUI
4. (simulations runs again by itself) (call this output RESULTS_2)
5. reset from GUI
6. (simulations runs again by itself) (call this output RESULTS_3)

**conclusion:** RESULTS_2 = RESULTS_3 but RESULTS_1 is not equal to either

Looking closer at the differences, it's oddly small and doesn't really diverge
```
RESULT_1
step   : x                  y                   z
0.0320 : 0.0000000000000000 -0.0001024000000000 0.0000000000000000
0.0640 : 0.0000474074074074 -0.0003072000000000 0.0000000000000000
0.0960 : 0.0001422222222222 -0.0006144000000000 0.0000000000000000
0.1280 : 0.0002844444444444 -0.0010240000000000 0.0000000000000000
RESULT_2
0.0320 : 0.0000000000000000 -0.0001024000000000 0.0000000000000000
0.0640 : 0.0000000000000000 -0.0003072000000000 0.0000000000000000 <-- shenanigans (spoiler: caused by gravity)
0.0960 : 0.0000474074074074 -0.0006144000000000 0.0000000000000000
0.1280 : 0.0001422222222222 -0.0010240000000000 0.0000000000000000
```
When checking the data, the **x** of RESULT_2 is just like RESULT_1 but shifted by one timestep and the **y** is unchanged.
This remains true for much longer captures, and any consecutive reset will look exactly like RESULT_2 so it is 
reproducible from a physics standpoint but not from a "logic" one, basically the first run and the successive ones don't do the same "thing".

This is situation (1) code-wise, through some console prints inside function `WbControlledWorld::step` :
```
world is loaded
<< USER INPUT: STEP <<<
WbControlledWorld::step (SimulationState [isPaused=0][isStep=1]) <-- sequence of step() calls that don't reach  
WbControlledWorld::step (SimulationState [isPaused=1][isStep=0])     the end  of the function, controllers are 
WbControlledWorld::step (SimulationState [isPaused=1][isStep=0])     spawned etc
WbControlledWorld::step (SimulationState [isPaused=1][isStep=0]) 
WbControlledWorld::step (SimulationState [isPaused=1][isStep=0]) <-- notice that we're paused, as we're 
WbControlledWorld::step (SimulationState [isPaused=1][isStep=0])     doing it step by step
WbControlledWorld::step (SimulationState [isPaused=1][isStep=0])
WbControlledWorld::step (SimulationState [isPaused=1][isStep=0])
WbControlledWorld::step (SimulationState [isPaused=1][isStep=0])
-- controllers at the END of the function WbControlledWorld::step
- mNewControllers:
- mWaitingControllers:
- mControllers:                                                  <-- when we finally reach the end of the 
    position_tracker: isRunning? YES                                 function, both controllers are in 
    dummy_controller: isRunning? YES                                 mControllers and both are running
CALLING WbSimulationWorld::step()                                <-- doing the simulation step
done
<< USER INPUT: STEP <<<
WbControlledWorld::step (SimulationState [isPaused=0][isStep=1])
-- controllers at the END of the function WbControlledWorld::step
- mNewControllers:
- mWaitingControllers:
- mControllers:
    position_tracker: isRunning? YES
    dummy_controller: isRunning? YES
CALLING WbSimulationWorld::step()                                <-- same, all is fine
done
<< USER INPUT: RESET FROM GUI <<<                                <-- manual reset
<< USER INPUT: STEP <<<
WbControlledWorld::step (SimulationState [isPaused=0][isStep=1]) 
WbControlledWorld::step (SimulationState [isPaused=1][isStep=0])  
WbControlledWorld::step (SimulationState [isPaused=1][isStep=0])  
WbControlledWorld::step (SimulationState [isPaused=1][isStep=0])
WbControlledWorld::step (SimulationState [isPaused=1][isStep=0])
WbControlledWorld::step (SimulationState [isPaused=1][isStep=0])
WbControlledWorld::step (SimulationState [isPaused=1][isStep=0])
WbControlledWorld::step (SimulationState [isPaused=1][isStep=0])
WbControlledWorld::step (SimulationState [isPaused=1][isStep=0])
-- controllers at the END of the function WbControlledWorld::step
- mNewControllers:
- mWaitingControllers:
- mControllers:
    position_tracker: isRunning? YES
    dummy_controller: isRunning? YES
CALLING WbSimulationWorld::step()                                 <-- again by the time we reach the end,  
done                                                                  the controllers are up and running
<< USER INPUT: STEP <<<
WbControlledWorld::step (SimulationState [isPaused=0][isStep=1])
-- controllers at the END of the function WbControlledWorld::step
- mNewControllers:
- mWaitingControllers:
- mControllers:
    position_tracker: isRunning? YES
    dummy_controller: isRunning? YES
CALLING WbSimulationWorld::step()
done

```

and indeed the console output is
```
INFO: position_tracker: Starting controller: python3 -u position_tracker.py
INFO: dummy_controller: Starting controller: python3 -u dummy_controller.py
0.0320 : 0.0000000000000000 -0.0001024000000000 0.0000000000000000
0.0640 : 0.0000474074074074 -0.0003072000000000 0.0000000000000000
INFO: position_tracker: Starting controller: python3 -u position_tracker.py
INFO: dummy_controller: Starting controller: python3 -u dummy_controller.py
0.0320 : 0.0000000000000000 -0.0001024000000000 0.0000000000000000
0.0640 : 0.0000474074074074 -0.0003072000000000 0.0000000000000000
```
showing nothing anomalous.
Doing the same thing for situation (2) yields (the world was reloaded before doing this test):

```
world is loaded
<< USER INPUT: REALTIME <<<
WbControlledWorld::step (SimulationState [isPaused=0][isStep=0]) <-- sequence of step() calls that don't reach 
WbControlledWorld::step (SimulationState [isPaused=0][isStep=0])     the end of the function
WbControlledWorld::step (SimulationState [isPaused=0][isStep=0])
WbControlledWorld::step (SimulationState [isPaused=0][isStep=0]) <-- notice that now we're not paused anymore
WbControlledWorld::step (SimulationState [isPaused=0][isStep=0])
WbControlledWorld::step (SimulationState [isPaused=0][isStep=0])
WbControlledWorld::step (SimulationState [isPaused=0][isStep=0])
WbControlledWorld::step (SimulationState [isPaused=0][isStep=0])
WbControlledWorld::step (SimulationState [isPaused=0][isStep=0])
-- controllers at the END of the function WbControlledWorld::step
- mNewControllers:
- mWaitingControllers:
- mControllers:
    position_tracker: isRunning? YES                             <-- the world was reloaded, so the starting 
    dummy_controller: isRunning? YES                                 procedure is the same as the previous   
CALLING WbSimulationWorld::step()                                    case, hence, all is good so far 
done

[ ... more steps like this ... ]                                 <-- everything fine, RESULT_1 won't change

<< USER INPUT: RESET FROM GUI <<<                                <-- manual reset
WbControlledWorld::step (SimulationState [isPaused=0][isStep=0]) 
WbControlledWorld::step (SimulationState [isPaused=0][isStep=0])
WbControlledWorld::step (SimulationState [isPaused=0][isStep=0])
WbControlledWorld::step (SimulationState [isPaused=0][isStep=0])
WbControlledWorld::step (SimulationState [isPaused=0][isStep=0])
WbControlledWorld::step (SimulationState [isPaused=0][isStep=0])
-- controllers at the END of the function WbControlledWorld::step
- mNewControllers:
    dummy_controller: isRunning? NO                              <-- effect of the reset
- mWaitingControllers:
- mControllers:
    position_tracker: isRunning? YES
CALLING WbSimulationWorld::step()                                <-- here's the culprit, a simulation step is
done                                                                 taken despite one of the controllers not
WbControlledWorld::step (SimulationState [isPaused=0][isStep=0])     yet active
WbControlledWorld::step (SimulationState [isPaused=0][isStep=0])
WbControlledWorld::step (SimulationState [isPaused=0][isStep=0])
WbControlledWorld::step (SimulationState [isPaused=0][isStep=0])
-- controllers at the END of the function WbControlledWorld::step
- mNewControllers:
- mWaitingControllers:
- mControllers:
    position_tracker: isRunning? YES                             <-- the other timesteps are fine, but since 
    dummy_controller: isRunning? YES                                 we've made a simulation step we shouldn't, 
CALLING WbSimulationWorld::step()                                    it won't produce the same result
done

[ ... more steps like this ... ]                                 <-- running
```

and the console output is
```
INFO: position_tracker: Starting controller: python3 -u position_tracker.py
INFO: dummy_controller: Starting controller: python3 -u dummy_controller.py
0.0320 : 0.0000000000000000 -0.0001024000000000 0.0000000000000000
0.0640 : 0.0000474074074074 -0.0003072000000000 0.0000000000000000
0.0960 : 0.0001422222222222 -0.0006144000000000 0.0000000000000000
0.1280 : 0.0002844444444444 -0.0010240000000000 0.0000000000000000
0.1600 : 0.0004740740740741 -0.0015360000000000 0.0000000000000000
INFO: position_tracker: Starting controller: python3 -u position_tracker.py
INFO: dummy_controller: Starting controller: python3 -u dummy_controller.py
0.0320 : 0.0000000000000000 -0.0001024000000000 0.0000000000000000
0.0640 : 0.0000000000000000 -0.0003072000000000 0.0000000000000000 <-- parasitic step, gravity affects box
0.0960 : 0.0000474074074074 -0.0006144000000000 0.0000000000000000 <-- from this point dummy_controller is    
0.1280 : 0.0001422222222222 -0.0010240000000000 0.0000000000000000     up and running and it applies a 
0.1600 : 0.0002844444444444 -0.0015360000000000 0.0000000000000000     force to itself
```

The culprit is therefore a different resetting behavior for the two cases, and the reason is that either [L429-L433](https://github.com/cyberbotics/webots/blob/592090f4e1967575e9f0bd5e984f9c43caa6ec1b/src/webots/control/WbControlledWorld.cpp#L429-L433) or [L445-L450](https://github.com/cyberbotics/webots/blob/592090f4e1967575e9f0bd5e984f9c43caa6ec1b/src/webots/control/WbControlledWorld.cpp#L445-L450) (in this specific case it's this latter that triggers) the `dummy_controller` is added in one case to the mNewControllers list and in the other to the mWaitingControllers. According to the headers:
```
QList<WbController *> mWaitingControllers;  // controllers inserted in previous step and waiting to be started in current step
QList<WbController *> mNewControllers;      // controllers inserted in current step mode and waiting next step to start
```
So what happens is that in situation (1), after the reset the `dummy_controller` is added and activated in the same step call hence by the time we reach the end of the function it's ready to go. In situation (2),  `dummy_controller`  is moved from mNewControllers to mWaitingController only AFTER a `WbSimulationWorld::step()` is executed ([change occurs here](https://github.com/cyberbotics/webots/blob/592090f4e1967575e9f0bd5e984f9c43caa6ec1b/src/webots/control/WbControlledWorld.cpp#L361)).
Which explains the results obtained, as in this additional simulation step, nothing but gravity is affecting the box which results in the **y** dip and only in the successive step the controller kicks in, but that's enough for the data to mismatch.

**Proposed fix:**
execute the line `WbSimulationWorld::step()` only if `mNewControllers` is empty.

It gets the job done in this specific case and test_suite is all green, but I don't know if it messes up with the controller logic more in general. The starting procedure it's quite convoluted (between waiting for external controllers, and some being spawned by startControllers and others by other means) so although this seems like an appropriate fix I don't know if it might be overkill to prevent it every time since the problem so far appears only to arise in the specific situation of a reset. Whatever the case, at least it should clarify the origin of the problem.

### DIFF
--- a/src/webots/control/WbControlledWorld.cpp
+++ b/src/webots/control/WbControlledWorld.cpp
@@ -355,8 +355,10 @@ void WbControlledWorld::step() {
     }
   }
 
-  mIsExecutingStep = true;
-  WbSimulationWorld::step();
+  if (mNewControllers.isEmpty()) {
+    mIsExecutingStep = true;
+    WbSimulationWorld::step();
+  }
 
   waitForRobotWindowIfNeededAndCompleteStep();
 }


ad-daniel:
I noticed only now that you (@lukicdarkoo) added an additional test file, when I run the checker it does find differences even with this fix which doesn't surprise as the two scenarios are completely different (there's just one controller in this one) so either way doesn't invalidate what is said above. 

I'll need to check more in detail why that one produces differences but after giving it a quick look in this second test case the differences are of the order of 10^(-16) as compared to the other one where it was much more noticeable. Considering in that second scenario there are 11 identical solids completely superposed all having physics enabled, I suspect it might have something to do with the tolerances ODE uses or how far it tolerates solid penetration. I can try to give a look at this case 2nd case but if all of the differences are in that order of magnitude they are at the limit of my computer. Still, I only checked a couple from the list printed by the checker, so there might be bigger ones.

lukicdarkoo:
Fantastic PR! Thanks!

> I'll need to check more in detail why that one produces differences but after giving it a quick look in this second test case the differences are of the order of 10^(-16) as compared to the other one where it was much more noticeable.

Try replacing the Lua code with the generated code. The error should disappear and you may get a better clue what is happening.

lukicdarkoo (approved):
> The starting procedure it's quite convoluted (between waiting for external controllers, and some being spawned by startControllers and others by other means) so although this seems like an appropriate fix I don't know if it might be overkill to prevent it every time since the problem so far appears only to arise in the specific situation of a reset.

The integration tests are passing, I did a few tests with different controller configuration, so I think it is safe to merge.

It is not overkill. From my point of view, it is important that all controllers start at the same time, especially if that affects reproducibility. Thank you very much for this!

PR #2687, opened 2021-02-01
Fix X3D USE node export

### LINKED ISSUE #2685: USE id not exported correcty when DEF is in a PROTO
**Describe the Bug**
When you have a PROTO with a customizable field than can be set as DEF (like the appearance field in the solidPipe PROTO) and you add another node (outside the PROTO) with a USE toward this DEF, the DEF id inside the USE node is not correctly exported in x3d.

This bug seems to apply to appearance but also shape, geometries, etc

**Steps to Reproduce**
1. Open webots with "./webots --stream"
2. Open "floating_geometries.wbt"
3. Open the streaming viewer
4. See that the pool's floor is missing its appearance and  ```X3dLoader: no matching DEF node "n10" node.``` appears in the browser's console.

**Expected behavior**
It seems like two nodes are created for the appearance of the pool, one when parsing the PROTO file (id = 10) and one to actually display the pool.

The pool's floor USE referenced the id 10 whereas it should reference the id of the other node as it is the second one that is actually exported and not the one with id 10.

**Screenshots**
Screenshot from 2021-01-29 11-28-26

Simplified version of floating_geometries.wbt in webots

Screenshot from 2021-01-29 11-29-00

The rendering in the current web viewer

**Additional context**
This bug is not fixed by the new version of the web viewer as it seems directly related to webots and how it loads worlds.

stefaniapedrazzi:
Fix #2685:
PROTO parameter nodes are not exported to X3D.
So, when exporting the DEF node id linked by a USE node, we should write the id of the first finalized (and exported) instance instead of the PROTO parameter node id.

### DIFF
--- a/docs/reference/changelog-r2021.md
+++ b/docs/reference/changelog-r2021.md
@@ -1,10 +1,25 @@
 # Webots R2021 Change Log
 
-## Webots R2020a Revision 1
-Released on XXX XX, 2021.
+## Webots R2021a Revision 1
+Released on XX Xth, 2021.
 
+  - Enhancements
+    - Don't display warnings for recent Intel and AMD graphics cards ([#2623](https://github.com/cyberbotics/webots/pull/2623)).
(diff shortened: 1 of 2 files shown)


BenjaminDeleze (approved):
Thank you !

omichel (approved):

PR #2695, opened 2021-02-02
Fixed Background irradiance when no colormap is set

omichel:
Fixes bug discussed in #2591.

### DIFF
--- a/src/webots/nodes/WbBackground.cpp
+++ b/src/webots/nodes/WbBackground.cpp
@@ -379,7 +379,7 @@ void WbBackground::updateCubemap() {
         destroySkyBox();
         applyColorToWren(skyColor());
         emit WbWrenRenderingContext::instance()->backgroundColorChanged();
-      } else if (hasCompleteBackground)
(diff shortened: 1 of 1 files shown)


DavidMansolino (approved):
Thank you!

Note that this PR doesn't solve the issue with the now useless 'reflections' field of the 'TexturedBackground' PROTO:
https://github.com/cyberbotics/webots/blob/develop/projects/objects/backgrounds/protos/TexturedBackground.proto#L37

omichel:
> Note that this PR doesn't solve the issue with the now useless 'reflections' field of the 'TexturedBackground' PROTO:
https://github.com/cyberbotics/webots/blob/develop/projects/objects/backgrounds/protos/TexturedBackground.proto#L37

Thank you. Yes, I will investigate this now...

omichel:
The `reflections` field is never used (never set to `FALSE`) in the proto and world examples we distribute. So, I would suggest to simply remove it from the `TexturedBackground` proto.

PR #2696, opened 2021-02-02
Converted UR3e robot to use Mesh instead of IndexedFaceSet

omichel:
Loading a world file with 10 UR3e robots now takes 15 seconds on this branch on my PC whereas the version on the master branch takes 32 seconds.

__Note__: I will create similar PRs for the UR5e and UR10e robots once this PR is merged.


lukicdarkoo (approved):
I can confirm it loads much faster.

Thank you

PR #2697, opened 2021-02-02
Removed now useless reflections parameter of the TexturedBackground proto

omichel:
Following-up discussion in #2695.

__Note__: this `reflections` parameter is never used in the examples we distribute.


DavidMansolino (approved):
Thank you.

PR #2698, opened 2021-02-02
Proto2mesh multithreaded

Simon-Steinmann:
Adding more parsable arguments and multithreading


omichel:
The CI tests are failing (line too long, see log).

omichel:
Is the PR ready for review now or do you want to commit your fast mesh optimization code in there?

Simon-Steinmann:
The PR should be pretty much ready now. So far my tests are all successful, However, I only inspected the generated .obj files, not the whole protos loaded into Webots. We should probably think of a way to verify and benchmark a conversion. Compare load times and see if the robot looks and functions the same.

scripts/proto_splitter/proto2mesh.py
```diff
@@ -159,27 +183,35 @@ def apply_crease_angle(self):
...
             if k == 0:
-                sys.exit('Wrong face: ' + str(face[0]) + ', ' + str(face[1]) + ', ' + str(face[2]) + ', -1\n' +
-                         str(p0) + str(p1) + str(p2))
+                if self.verbose:
+                    print('Wrong face: ' + str(face[0]) + ', ' + str(face[1]) + ', ' + str(face[2]) + ', -1\n' +
+                          str(p0) + str(p1) + str(p2) + '\n' + str(n), flush=True)
+                self.coordIndex.remove(face)
+                continue
+                # raise Exception('Wrong face: ' + str(face[0]) + ', ' + str(face[1]) + ', ' + str(face[2]) + ', -1\n' +
+                #          str(p0) + str(p1) + str(p2) + '\n' + str(n))
```
omichel, line 204: Clean-up:
```suggestion
```

scripts/proto_splitter/proto2mesh.py
```diff
@@ -236,30 +263,56 @@ def write_obj(self, file):
...
         i = ln.index('[')
         data = ' '.join(ln[i:])
         while ']' not in line:
-            line = self.f.readline()
+            line = self.f.readline().split('#')[0]
+            if '{' in line:
+                self.shapeLevel += 1
+            self.lineNumber += 1
+            if '%' in line:
+                raise Exception("ERROR: LUA skript for mesh data is not supported.")
```
omichel, line 297: German word 😆:
```suggestion
                raise Exception("ERROR: LUA script for mesh data is not supported.")
```

scripts/proto_splitter/proto2mesh.py
```diff
@@ -236,30 +263,56 @@ def write_obj(self, file):
...
             ln = line.split()
             data += line
         data = ' '.join(data.split())
         data = data.replace('[', '').replace(']', '')
+        if defName is not None:  # store coord and coordIndex data for DEF to assign to USE later
+            self.meshDEFcache[defName] = data
+        # print(defName, len(data))
```
omichel, line 304: cleanup:
```suggestion
```

scripts/proto_splitter/proto2mesh.py
```diff
@@ -449,18 +551,80 @@ def write_obj(self, meshes):
...
+        default=False,
+        action='store_true',
+        help='If set, enables multicore processing for directories with several PROTO files. May not work on Windows.',
+    )
+    optParser.add_option(
+        '--verbose', '-v',
+        dest='verbose',
+        default=False,
+        action='store_true',
+        help='If set, detailed output of mesh conversion is shown in console',
```
omichel, line 572: consistency with other help lines:
```suggestion
        help='If set, detailed output of mesh conversion is shown in console.',
```

scripts/proto_splitter/proto2mesh.py
```diff
@@ -449,18 +551,80 @@ def write_obj(self, meshes):
...
-            print('Done')
-        elif os.path.isdir(inPath):
-            inPath = os.path.abspath(inPath)
-            p2m.convert_all(inPath)
-            print('Done')
-        else:
-            sys.exit('Error: --input has to be a .proto file or directory!')
+    outPath = options.outPath
+    verbose = options.verbose
+    cpv = options.cpv
```
omichel, line 585: We should try to avoid this kind of abbreviation in variable and function names, prefer `checkProtoValidity` instead.

scripts/proto_splitter/proto2mesh.py
```diff
@@ -449,18 +551,80 @@ def write_obj(self, meshes):
...
+    mt = options.mt
+    tStart = time.time()
+    p2m = proto2mesh()
+    if cpv:  # if we check protos files for validity
```
omichel, line 589: ... This will save us from this kind of comment.

scripts/proto_splitter/proto2mesh.py
```diff
@@ -449,18 +551,80 @@ def write_obj(self, meshes):
...
-        elif os.path.isdir(inPath):
-            inPath = os.path.abspath(inPath)
-            p2m.convert_all(inPath)
-            print('Done')
-        else:
-            sys.exit('Error: --input has to be a .proto file or directory!')
+    outPath = options.outPath
+    verbose = options.verbose
+    cpv = options.cpv
+    mt = options.mt
```
omichel, line 586: I would rename `mt` to `multithreading` (see explanation below).

omichel (changes requested):
Thank you for this very good PR!
I only have a few minor suggestions.
Also, I believe we should probably remove the `--input=` option and replace it with a mandatory argument that comes without this `--input=` prefix, like:
```bash
python proto2mesh.py ../../projects/robots --verbose
```

omichel (approved):
Thank you.
It looks all very good to me now.

omichel (approved):
Thank you.

PR #2700, opened 2021-02-03
Fix commit activity badge

omichel:
This badge was displaying "invalid".


omichel:
Actually it works now on the master branch. That was probably a temporary problem...

PR #2703, opened 2021-02-03
Update the list of exported images when streaming to x3d

BenjaminDeleze:
**Description**
Currently the list of textures that webots can serve to the streaming-viewer is defined only when the world is exported as x3d.

The problem is that if we add appearances with new textures to the simulation while it is streaming, the new textures will not be added to the list and therefore not served to the streaming-viewer.

I solved this issue by adding the new textures to the existing list when the server exports new nodes to the streaming-viewer.


BenjaminDeleze:
Need to be merged in master not in develop

PR #2704, opened 2021-02-03
update textures while streaming

BenjaminDeleze:
**Description**
Currently the list of textures that webots can serve to the streaming-viewer is defined only when the world is exported as x3d.

The problem is that if we add appearances with new textures to the simulation while it is streaming, the new textures will not be added to the list and therefore not served to the streaming-viewer.

I solved this issue by adding the new textures to the existing list when the server exports new nodes to the streaming-viewer.


stefaniapedrazzi:
We should not mix a Qt upgrade PR with other fixes.
So if it is needed to upgrade Qt on mac you should do it in a different branch and open a dedicated PR.
This will also make it easier to test your fix because you won't have all the errors related to the Qt upgrade.

omichel (approved):

PR #2705, opened 2021-02-03
Converted UR5e to Meshes

omichel:
Similar to #2696, but for UR5e.


stefaniapedrazzi (changes requested):
There are errors loading the robot.
The path to the meshes folder in UR5e.proto is wrong: `UR5e/meshes/meshes/<name>.obj` instead of `UR5e/meshes/<name>.obj`

omichel:
Oops... Well spotted. It should be fixed now (I tested it).

stefaniapedrazzi (approved):

PR #2708, opened 2021-02-03
Fixed discord updater

omichel:
This PR aims at fixing a problem spotted in #2702: the update of the discord pages should target the released branch in order to be immediately displayed on cyberbotics.com web site.

### DIFF
--- a/.github/workflows/discord_updater.yml
+++ b/.github/workflows/discord_updater.yml
@@ -11,6 +11,7 @@ jobs:
     - uses: actions/checkout@v2
       with:
         fetch-depth: 0
+        ref: released 
     - name: Set up Python 3.8
       uses: actions/setup-python@v2
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/discord_updater.yml
```diff
@@ -26,6 +26,7 @@ jobs:
       run: |
         git config --global user.name ${GITHUB_ACTOR}
         git config --global user.email ${GITHUB_ACTOR}@github.com
+        git checkout released
```
DavidMansolino, line 29: It sounds safer to do this in the checkout part:
````
- uses: actions/checkout@v2
      with:
        fetch-depth: 0
        ref: develop
````
Otherwise, you apply the script on the master branch and then only switch branch (just to make sure to avoid any possible conflicts).

omichel: Thank you! I guess you meant `ref: released` instead of `ref: develop`. I did that change.

DavidMansolino: You're welcome, you're right, it is indeed `ref: released`.

omichel:
👍🏻

stefaniapedrazzi (approved):

PR #2710, opened 2021-02-04
Improved Darwin-OP proto to use Mesh nodes

omichel:
Yet another IndexedFaceSet to Mesh converted PROTO file.


omichel:
This doesn't seems to improve the performance.
On my PC, loading a world with 20 Darwin-OP robots takes 14 seconds on this branch whereas it takes only 8 seconds on the master branch.
So I am wondering whether this is a good idea to merge this PR, probably not... 😒

stefaniapedrazzi:
I was just checking the loading improvement in this PR too..
And I also cannot see a big difference (even when regenerating the PROTO for example by adding the jersey).
If there is no significant performance improvement , I agree that it is probably better to keep the original single file.

PR #2712, opened 2021-02-05
delete step message

BenjaminDeleze:
**Description**
Webots is sending text messages along update messages (containing time, new position,...). These text messages can be: "pause", "real-time", "fast" and "step". The three first are used to change the appearance of the buttons in the toolbar of the streaming-viewer. But the "step" message is not used. In fact it is not even handle in the streaming-viewer, resulting in ```WebSocket error: Unknown message received "step"```.

As we don't use it, I removed it from webots.


omichel (approved):
Thank you.

PR #2713, opened 2021-02-05
regenerate x3d world at loading

BenjaminDeleze:
**Description**
Sometimes, the x3d world in not regenerated by webots when reloading the world even so the world has changed.

Minimal example:
-launch webots with ```--stream```
-create a simple cube
-save the world
-open and connect the streaming-viewer, you should see the cube
-disconnect the streaming-viewer.
-delete the cube
-save and reload the simulation.
-connect the streaming-viewer, you should still see the cube in the streaming-viewer even so it is no more in webots.

To solve this, I removed the conditions such that now the x3dworld is regenerated every time we reload the world.


omichel (approved):
That looks good to me.

stefaniapedrazzi:
Did you measure the performance drop when resetting the simulation from the web interface before removing this optimization rather than fixing it?

BenjaminDeleze:
It think it will not affect the performance, because this function is not called when resetting the simulation (from the web or from webots). It is only called when webots loads or reloads a world.
So the only situation where it can change the performance is if we reload the current world.

stefaniapedrazzi:
Ok, then it is fine. Thank you.
I just wanted to be sure that we understand the code before this kind of changes that could produce regressions.

src/webots/gui/WbX3dStreamingServer.cpp
```diff
@@ -142,19 +142,7 @@ void WbX3dStreamingServer::sendUpdatePackageToClients() {
 
 bool WbX3dStreamingServer::prepareWorld() {
   try {
-    bool regenerationRequired = mX3dWorldReferenceFile != WbWorld::instance()->fileName() || mX3dWorldGenerationTime != 0.0;
```
DavidMansolino, line 145: `mX3dWorldReferenceFile` is probably useless now.

BenjaminDeleze: You are right

PR #2715, opened 2021-02-05
Fixed URDF test which was broken by new UR5e model

omichel:
The node ids changed in this test world after the UR5e model changed.

### DIFF
--- a/tests/api/controllers/robot_urdf/prob3_reference.urdf
+++ b/tests/api/controllers/robot_urdf/prob3_reference.urdf
@@ -248,7 +248,7 @@
       </geometry>
     </collision>
   </link>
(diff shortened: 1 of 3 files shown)


lukicdarkoo:
Maybe we should delete these tests. We already have a test that verifies that the URDF file valid. This test is useful, but the maintenance of it is cumbersome.

BenjaminDeleze (commented):
Thanks

omichel:
I will fix it in a separate PR. Meanwhile, I believe we can accept this one so that @BenjaminDeleze will be able to progress with #2712.

BenjaminDeleze (approved):

PR #2716, opened 2021-02-05
Fix urdf test

omichel:
The URDF test appeared to be broken. This PR fixes it.


omichel:
From the test result, it seems that `check_urdf` is not installed on the CI machine which explains the failure.

omichel:
`check_urdf` actually comes from `apt install liburdfdom-tools`, but it seems that it is missing on the CI machine.

scripts/install/linux_test_dependencies.sh
```diff
@@ -0,0 +1,28 @@
...
+UBUNTU_VERSION=$(lsb_release -rs)
+if [[ $UBUNTU_VERSION == "16.04" ]]; then
+       export ROS_DISTRO=kinetic
+elif [[ $UBUNTU_VERSION == "18.04" ]]; then
+       export ROS_DISTRO=melodic
+elif [[ $UBUNTU_VERSION == "20.04" ]]; then
+       export ROS_DISTRO=noetic
+else
+       echo "Unsupported Linux version."
+fi
```
DavidMansolino, line 17: Note that it is wrong to say that ROS is only a test dependency as it is actually required for the compilation of the ros controller on linux.

omichel: Yes, but it is still part of the optional compilation dependencies: the optional compilation dependencies includes the test dependencies.

omichel:
👍🏻

scripts/install/linux_test_dependencies.sh
lukicdarkoo: It indeed looks weird :) I guess the intuitive solution (to read) would be something like:
```
|-- linux_optional_compilation_dependencies.sh
    |-- linux_ros_dependency.sh
    |--  linux_test_dependencies.sh
         |-- linux_ros_dependency.sh (because of check_urdf)
         |-- check_urdf
```
`linux_ros_dependency.sh` is executed twice, but it is not a big deal.

However, this is overkill for us, so I guess we can keep it as is.

omichel: Note: that `linux_test_dependecies.sh` is not executed twice in this PR as I implemented a `--norecurse` option.

lukicdarkoo: Or simply, create `linux_ros_dependency.sh` and use it in the workflow directly + `check_urdf` installation.

omichel: > Or simply, create `linux_ros_dependency.sh` and use it in the workflow directly + `check_urdf` installation.

I don't like this solution, as dependencies should be organized in a hierarchical way so that when you need some dependency, you shouldn't bother about managing yourself the sub-dependencies of it.

lukicdarkoo (approved):
Thank you

PR #2720, opened 2021-02-08
Improve synchronization of Supervisor simulation reset

### LINKED ISSUE #1384: Emitter/Receiver communication between nodes not working after controller and simulation reset

stefaniapedrazzi:
Fixes #1384: fix reset behavior.

Apply the reset at the very end of the step on Webots side, but before sending the new simulation status to the controllers.
Before the reset was applied after sending the new simulation status to the controllers, i.e. values on the controller side were referring to the status prior to the reset for one additional step after the reset was requested.


src/webots/gui/WbMainWindow.cpp
```diff
@@ -1346,6 +1346,7 @@ void WbMainWindow::updateBeforeWorldLoading(bool reloading) {
     mDockWidgets.removeAll(console);
     delete console;
   }
+  // if (!world->isUnnamed())
```
omichel, line 1349:
```suggestion
```

omichel (changes requested):
I just spotted a minor issue.

src/webots/nodes/utils/WbWorld.cpp
```diff
@@ -79,6 +79,7 @@ WbWorld *WbWorld::instance() {
 
 WbWorld::WbWorld(WbProtoList *protos, WbTokenizer *tokenizer) :
   mWorldLoadingCanceled(false),
+  mResetRequested(false),
```
omichel, line 82: We should probably also initialize this new member variable:
```suggestion
  mResetRequested(false),
  mRestartControllers(false),
```

omichel (dismissed):

omichel (approved):

PR #1949, opened 2020-07-21
Rollback macOS Qt version.

DavidMansolino:
**Description**
The version 5.15.0 of Qt has some issues with button on pop-up windows.

**Related Issues**
This pull-request fixes issue #1944

PR #2721, opened 2021-02-08
Upgrade to Qt 5.15.2 on macOS

omichel:
This new version of Qt 5.15.2 fixes the bug reported in #1949.


BenjaminDeleze (approved):
Thank you!

PR #2732, opened 2021-02-10
Allow a robot to be added to the Transform node

lukicdarkoo:
This enables easy replication of an environment inside of a single world. It is especially useful when we want to run multiple simulations in a single Webots instance (to save RAM and VRAM).

P.S. @stefaniapedrazzi do you may know why it was disabled?


stefaniapedrazzi:
Originally `Robot` nodes were only valid as top root nodes (before supporting nested `Robot` nodes).
We never had the need to enable adding a `Robot` node as non-top node or not nested node.
But when the nested `Robot` node functionality was introduced it was decided that it was better to keep the `Robot` top node rule mainly for the following points.

Here are some food for thought:
1. We don't want that users scale robots because this could lead to issues in the physics simulation.
    A parent `Transform` node will make this easy and "tempting" for users.
2. If it is allowed to insert `Robot` into `Transform` nodes why not extending it to `Group` (and/or `Solid` or derived from `Solid`).
    We should use clear and simple rules to help users understanding them.

lukicdarkoo:
> We don't want that users scale robots

This is a very good point. Unfortunately, I am not sure how to solve this. We don't have a node that only translates/rotates children.

omichel (changes requested):
I believe we should add some documentation about this.
I could not find good information in the documentation about it.
Maybe some better information could be added here: https://cyberbotics.com/doc/reference/node-chart#insertion-rules and on the robot page.

lukicdarkoo:
Allowing the Robot node to be inserted in the other node is rather expected behavior, so I don't see the reason for adding documentation about it.

Maybe we can document rules for node insertion in another PR, but I don't see a strong need for it.

omichel:
I believe it would be good to better document the insertion rules, but you are right, it can be done in a separate PR.
By the way, did you try to change the scale of a Transform containing a Robot node? Is that possible from the scene tree? If so, are there any unwanted side effects?

lukicdarkoo:
image
Spot Micro :)

However, the hinges don't work properly. Does it make sense to disable the scale field when there is a robot inside?

omichel:
> However, the hinges don't work properly. Does it make sense to disable the scale field when there is a robot inside?

Yes, I would do so, as it doesn't make sense to allow to design scaled down/up models. This is not realistic (especially with non-proportional scaling).

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -194,6 +194,8 @@ namespace {
           return true;
         else if (WbNodeUtilities::hasARobotAncestor(node))
           return true;
+        else if (dynamic_cast<const WbGroup *>(node))
```
stefaniapedrazzi, line 197: Given that this block is executed only for `children` fields, is not possible to simplify and remove all the various checks instead of adding an additional condition that seems to be always true?

src/webots/nodes/utils/WbSupervisorUtilities.cpp
```diff
@@ -1726,6 +1726,7 @@ void WbSupervisorUtilities::writeConfigure(QDataStream &stream) {
   stream << (short unsigned int)0;
   stream << (unsigned char)C_CONFIGURE;
   stream << (int)selfNode->uniqueId();
+  stream << (int)selfNode->parentNode()->uniqueId();
```
stefaniapedrazzi, line 1729: Note that if you change the communication messages between Webots and the controller, then this has to be merged on the develop branch and not on the master branch otherwise you will break the compatibility (unless this is a bug fix).

src/webots/nodes/WbAbstractTransform.cpp
```diff
@@ -215,6 +215,11 @@ bool WbAbstractTransform::checkScale(int constraintType, bool warning) {
   if (constraintType > 0 && checkScalingPhysicsConstraints(correctedScale, constraintType, warning))
     b = true;
 
+  if (WbNodeUtilities::hasARobotDescendant(dynamic_cast<const WbNode *>(this))) {
+    correctedScale.setXyz(1, 1, 1);
```
stefaniapedrazzi, line 219: We should also print an error message explaining that the scale is automatically reset because it is not possible to scale robot nodes.

stefaniapedrazzi (changes requested):
We should add in the documentation of the `Transform.scale` field this exception with descendant robots:
https://www.cyberbotics.com/doc/reference/transform#field-summary

src/webots/nodes/utils/WbSupervisorUtilities.cpp
lukicdarkoo: I see. Initially, I wanted to change only the parser, but then I figured out it is needed to update the controller as well.

I will make a new PR, so I don't mess up the history.

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -188,12 +188,9 @@ namespace {
     if (childrenField) {
       const bool isInsertingTopLevel = node->isWorldRoot();
 
-      // Robots are no longer top-level nodes
+      // A robot cannot be a bounding object
       if (!boundingObjectCase && WbNodeUtilities::isRobotTypeName(nodeName)) {
```
omichel, line 192:
```suggestion
      if (!boundingObjectCase && WbNodeUtilities::isRobotTypeName(nodeName))
```

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -188,12 +188,9 @@ namespace {
...
-        if (WbNodeUtilities::isRobotTypeName(node->nodeModelName()) || isInsertingTopLevel)
-          return true;
-        else if (WbNodeUtilities::hasARobotAncestor(node))
-          return true;
+        return true;
       }
```
omichel, line 194:
```suggestion
```

omichel (changes requested):
Just a minor issue.

stefaniapedrazzi:
Thank you.
Note that the comment about the documentation has not been addressed:

> We should add in the documentation of the `Transform.scale` field this exception with descendant robots:
> https://www.cyberbotics.com/doc/reference/transform#field-summary

lukicdarkoo:
> Thank you.
> Note that the comment about the documentation has not been addressed:
> 
> > We should add in the documentation of the `Transform.scale` field this exception with descendant robots:
> > https://www.cyberbotics.com/doc/reference/transform#field-summary

Sorry, I have just added the description.

omichel (dismissed):

docs/reference/transform.md
```diff
@@ -54,5 +54,7 @@ For such nodes a positive `scale` field initially set to *x y z* is automaticall
 The same holds for a `Transform` placed inside a `boundingObject` and with a `Sphere` or a `Capsule` descendant.
 In the case of a `Cylinder`, *x y z* will be reset to *x z x*.
 If some value changes within one of the previous constrained scale fields, the two others are actuated using the new value and the corresponding constraint rule.
+Scaling is forbidden if a `Robot` is a descendant of the `Transform`.
+In the case the `Robot` is a descendant of the `Transform`, *x y z* will be reset to *1 1 1*, and a warning will be printed.
```
stefaniapedrazzi, line 58: i think this sentence can be simplified:
```suggestion
Scaling is forbidden if a `Robot` is a descendant of the `Transform`.
In this case, *x y z* will be reset to *1 1 1* and a warning will be printed.
```

Otherwise I will at least change
"In the case the `Robot` is a descendant of the `Transform`" to 
"In the case a `Robot` is a descendant of the `Transform`"

stefaniapedrazzi (approved):
Thank you

PR #2737, opened 2021-02-10
Postpone sending logs to consoles during the world load

### LINKED ISSUE #2671: Errors in world file cause Webots to silently default to an empty world
**Describe the Bug**

When loading a world file, if there are errors (I've tried unmatched braces and incorrect property types such as swapping a number for a string), no error is printed and the window loads an empty (new) world.

**Steps to Reproduce**
1. Create a new world.
2. Insert some stuff. I added a chair and a light.
3. Save the world file.
4. Perform a fresh load of the world file. It loads fine.
5. Edit the world file and introduce an error. Originally, I discovered this with a string `"true"` instead of boolean `TRUE` in Robot `supervisor`. One similar example:
    ```
    PointLight {
      attenuation "foo"
      location 0 0.3 0
    }
    ```
    Alternately, removing a closing brace does it, or using a node type which doesn't exist.
6. Perform a fresh load of the world.
7. Observe that the world view is now black, the scene tree has reset to default, and there are no messages in the console. The window title still reports your file name.

**Expected behavior**

An error message in the console would be nice. Perhaps even stay on the existing world, or some other intermediate state. The completely empty scene view is confusing.

**Screenshots**

image

**System**
 - Graphics Card: Machine 1: NVIDIA GTX 750 Ti, Machine 2: Intel Integrated Graphics

stefaniapedrazzi:
Fixes #2671:
logs generated during world load were sent to the consoles of the previous world and thus no longer visible once the load and the new world consoles were open.
This fix consists in queuing the logs during the world load and send them to the new console once they have been opened.

### DIFF
--- a/src/webots/app/WbApplication.hpp
+++ b/src/webots/app/WbApplication.hpp
@@ -47,8 +47,8 @@ public:
   // load a world .wbt file
   // worldName must be absolute or specified with respect to WEBOTS_HOME
   // return true on success, false otherwise
-  bool loadWorld(const QString &worldName, bool reloading);
   bool loadWorld(QString worldName, bool reloading);
+  bool isValidWorldFile(const QString &worldName);
 
   // take a sceenshot of the 3d view
   // quality must be between 0 and 100 included
--- a/src/webots/app/WbApplication.hpp
+++ b/src/webots/app/WbApplication.hpp
@@ -47,6 +47,7 @@ public:
   // load a world .wbt file
   // worldName must be absolute or specified with respect to WEBOTS_HOME
   // return true on success, false otherwise
+  bool loadWorld(const QString &worldName, bool reloading);
   bool loadWorld(QString worldName, bool reloading);
 
   // take a sceenshot of the 3d view
--- a/src/webots/app/WbApplication.hpp
+++ b/src/webots/app/WbApplication.hpp
@@ -48,7 +48,7 @@ public:
   // worldName must be absolute or specified with respect to WEBOTS_HOME
   // return true on success, false otherwise
   bool loadWorld(QString worldName, bool reloading);
-  bool isValidWorldFile(const QString &worldName);
+  bool isValidWorldFileName(const QString &worldName);
 
   // take a sceenshot of the 3d view
   // quality must be between 0 and 100 included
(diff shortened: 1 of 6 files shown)


stefaniapedrazzi:
I'm fixing the initialization error reported by tests.

src/webots/app/WbApplication.hpp
```diff
@@ -48,6 +48,7 @@ class WbApplication : public QObject {
   // worldName must be absolute or specified with respect to WEBOTS_HOME
   // return true on success, false otherwise
   bool loadWorld(QString worldName, bool reloading);
+  bool isValidWorldFile(const QString &worldName);
```
omichel, line 51: The name of this function is a bit misleading as we are not checking the validity of the content of the world file as it suggests, but rather the world file name: does it ends with `.wbt` and does it correspond to a readable file. So I would rename it:
```suggestion
  bool isValidWorldFileName(const QString &worldName);
```

src/webots/gui/WbMainWindow.cpp
```diff
@@ -1321,13 +1321,22 @@ bool WbMainWindow::proposeToSaveWorld(bool reloading) {
...
   mSimulationView->cancelSupervisorMovieRecording();
   logActiveControllersTermination();
-  return WbApplication::instance()->loadWorld(fileName, reloading);
+  WbLog::setConsoleLogsPostponed(true);
+  const bool success = WbApplication::instance()->loadWorld(fileName, reloading);
+  if (!success) {
+    WbLog::setConsoleLogsPostponed(false);
+    WbLog::showPostponedPopUpMessages();
+  }
+  // else will be reset in updateAfterWorldLoading()
```
omichel, line 1334: Do you mean the console will be cleared in updateAfterWorldLoading()?

omichel (changes requested):
It looks very good, just a couple of minor issues.

src/webots/gui/WbMainWindow.cpp
stefaniapedrazzi: No, I mean that the `WbLog::setConsoleLogsPostponed` flag will be reset and the postponed messages will be forwarded in the updateAfterWorldLoading.

omichel: In that case, I would make it clearer in the comment:
```suggestion
  // else will the WbLog::setConsoleLogsPostponed() flag be reset in updateAfterWorldLoading()
```

stefaniapedrazzi: I improved the comment:
``// else console messages will be forwarded after world load in restorePerspective()``

omichel (approved):
Thank you.

Issue #2284, opened 2020-10-01
Play-one-frame button grays out too much


Justin-Fisher:
ARGUABLE BUG A: When you hit play-one-frame and a controller crashes, the play-one-frame button grays out and remains non-functional, even while the other play buttons remain functional and work to advance the simulation.  EXPECTED BEHAVIOR:  When the simulation is capable of advancing additional frames (without the controller that crashed) then the play-one-frame button should allow it do so one frame at a time, if you want.

DEFINITE BUG B:  If you hit Reset after the preceding, the play-one-frame button stays grayed out and non-functional.  EXPECTED BEHAVIOR: when you've reset a simulation, you should be able to play it one frame at a time, regardless of whether your preceding attempt ended with your advancing-one-frame into a controller crashing.  

ARGUABLE BUG C:  If you're running a quite slow simulation and triple-click play-one-frame, the user would expect to have the simulation advance three frames and then pause, but what it actually does is unpredictable, depending entirely upon how many frames it manages to process between successive clicks, and hence whether the button happens to be grayed out when each click occurs.  EXPECTED BEHAVIOR:  Each click of the the play-one-frame button should advance a designated pausing point by one timestep, and (if the simulation is paused) should ensure that the simulation becomes unpaused.  When the simulation reaches a designated pausing point, it should then pause.  This way three clicks of play-one-frame will always play three frames, rather than being unpredictable.

**System**
 - Operating System: Windows 10
 - Graphics Card: NVIDIA something fancyish

stefaniapedrazzi:
Bug B has been fixed in #2741.

"Bug A": the step button and all the other buttons works exactly the same.
* if the crashing controller is an extern controller:
    the step button remains gray and it is not possible to continue with the simulation, but the same exact behavior occurs with the real-time and run button (they can be pressed but the simulation doesn't advance)
* if the controller is started by Webots:
    the step button is enabled after the controller crash and it is possible to continue advancing the simulation step-by step

So given that the behavior is consistent, I don't see this as a bug and I think that showing that the step functionality is not enabled will help users understanding that something is wrong with the simulation and that it cannot be advanced.

About "Bug C": this doesn't seem a critical issue.
Users that want to advance the simulation step-by-step can easily check with the speedometer how many step the simulation advanced and understand that fast continuous clicking could not be processed as expected because they could occur on a disabled button.

ygoumaz:
[...]

PR #2741, opened 2021-02-11
Fix controllers logs printed in Webots console one step too late

### LINKED ISSUE #2709: Controller logs are written one step too late in the console
Logs coming from a controller are printed in the Webots console one step later than when they are actually executed.
For comparison I ran the exact same controller as "extern" and compared the logs in the Webots console vs controller terminal.

To reproduce run this simulation step by step (one `counter` controller should be started in a terminal as extern controller).

World file:
```
#VRML_SIM R2021a utf8
WorldInfo {
  coordinateSystem "NUE"
}
Viewpoint {
}
Robot {
  controller "counter"
}
Robot {
  name "robot_extern"
  controller "<extern>"
}
```

counter.c
```
#include <webots/robot.h>
#include <stdio.h>

#define TIME_STEP 64

  wb_robot_init();

  int i = 0;
    printf("Iteration %d\n", i++);
  };

  wb_robot_cleanup();

  return 0;
}
```

stefaniapedrazzi:
Fixes #2709:
execute the instructions to flush the controller buffers once all the read requests are completed.

Additionally, I also fixed an issue with the step button that is not enabled after simulation reset (BUG B in #2284) if the reset button is clicked during the step computation, i.e. when the step button was disabled.

### DIFF
--- a/docs/guide/the-console.md
+++ b/docs/guide/the-console.md
@@ -1,6 +1,8 @@
 ## The Console
 
 The Webots console displays the logs coming from Webots (such as parsing warnings or errors when opening a world, compilation outputs, ODE warnings, etc.) and the outputs of the controllers.
+To guarantee that the log printed in the Webots console is deterministic, the output coming from the controllers is not printed immediately when receiving it but it is grouped by controller and printed at the end of the simulation step.
+This means that if your simulation contains robot A and robot B, the output of robot A will be printed in the console before the output of robot B independently if some robot B output messages were received before the ones of robot A.
 
 
 %figure "Console Window"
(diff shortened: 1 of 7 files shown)


stefaniapedrazzi:
The commit that could cause some regression is c8d2b77, i.e. removing the flush from the `WbControlledWorld::waitForRobotWindowIfNeededAndCompleteStep` function rather than simply adding an additional flush in if the simulation is paused or run step-by-step.
It seems better to keep the code as clean as possible and limit the number of functions that flushes the buffers if this doesn't affect the functionality.
I inspected how the other GUI refresh is triggered and tested many different worlds and I could not find any issue related to this change.

omichel (approved):
Thank you.

omichel (approved):

omichel:
(I fixed a conflict with the master branch in the change log)

PR #2742, opened 2021-02-11
Allow the `wb_supervisor_node_reset_physics` function to reset physics of solid descendants

lukicdarkoo:
One more feature to allow users to do reinforcement learning easier. This way, a user can reset a physics in a part of the environment.

**Tasks**
- [x] Implement recursive physics reset
- [x] Add changelog
- [x] Add docs


src/webots/nodes/utils/WbSupervisorUtilities.cpp
```diff
@@ -803,15 +803,29 @@ void WbSupervisorUtilities::handleMessage(QDataStream &stream) {
...
+      if (recursive) {
+        QList<WbNode *> descendants = QList<WbNode *>({node});
+        for (int i = 0; i < descendants.size(); ++i) {
+          WbNode *child = descendants.at(i);
+          WbSolid *solid = dynamic_cast<WbSolid *>(child);
+          if (solid)
+            solid->resetPhysics();
+          descendants.append(child->subNodes(true, false, false));
+        }
+      } else {
```
DavidMansolino, line 822: Note that the ``WbSolid::resetPhysics`` function is already somehow recursive: https://github.com/cyberbotics/webots/blob/12a51b4a36050952485972df2a4490d020e7dc8b/src/webots/nodes/WbSolid.cpp#L2362-L2363

lukicdarkoo: Yes, but it cannot be applied to non-solid objects. We want to apply this to a Transform node that has Solid children.

Maybe we should delete the `recursive` argument and make it a default behavior? Do you have a better idea?

omichel: > Maybe we should delete the recursive argument and make it a default behavior?

Yes, that seems to be a very good idea.

DavidMansolino: Oh ok, I see, in that case the docs should be probably clearer because from an external point of view, the docs let think the user that if he sets ``recursive`` to false and call on a Solid node, only the top solid node is reset.
But maybe doing as you propose (i.e. just delete the recursive argument and make it a default behavior) is the simplest (both for the code and the user understanding).


Be also careful because here you are doing a kind of 'double recursion' (once on each Node and one on each Solid)
You may also use such kind of function (to find all the 'first Solids' node) to avoid rewriting the search loop and the double recursion:
https://github.com/cyberbotics/webots/blob/12a51b4a36050952485972df2a4490d020e7dc8b/src/webots/nodes/utils/WbNodeUtilities.hpp#L75

omichel (approved):
Thank you.

PR #2745, opened 2021-02-12
export basictimestep

BenjaminDeleze:
**Description**
Export basicTimeStep which is needed in the streaming-viewer for the viewpoint to correctly follow objects


omichel (approved):
Thank you.

PR #2725, opened 2021-02-08
Fix Camera images not updated after reset

stefaniapedrazzi:
After a simulation reset (without controller restart), the camera image on the controller is not updated.

PR #2751, opened 2021-02-13
Initialize device reset callback

lukicdarkoo:
This PR fixes the `wb_supervisor_simulation_reset` function which crashes the controller in some cases. The controller example is here:
https://github.com/cyberbotics/webots/pull/2725#issuecomment-778678263


stefaniapedrazzi (approved):
Thank you!

PR #2754, opened 2021-02-16
Fix fields order in openai_gym.wbt

DavidMansolino:
This is making the nightly world update test fail: https://github.com/cyberbotics/webots/runs/1906940169?check_suite_focus=true#logs

Feel free to edit/merge this PR.

P.S: note that the ros test is also failing during installation since a while: https://github.com/cyberbotics/webots/runs/1906940068?check_suite_focus=true

### DIFF
--- a/projects/samples/howto/openai_gym/worlds/openai_gym.wbt
+++ b/projects/samples/howto/openai_gym/worlds/openai_gym.wbt
@@ -3,8 +3,8 @@ WorldInfo {
   info [
     "This example shows an integration of OpenAI Gym with Webots."
   ]
-  coordinateSystem "NUE"
   title "OpenAI Gym Environment"
+  coordinateSystem "NUE"
 }
 Viewpoint {
   orientation 0.042679810423980585 0.9980594209305432 0.04534121826728663 4.649996506160134


stefaniapedrazzi:
That's strange.
I though I fixed it in #2754 and I merged the file resulting from saving it from Webots.

DavidMansolino:
> That's strange.
> I though I fixed it in #2738 and I merged the file resulting from saving it from Webots.

If you saved it from Webots, that is indeed strange, because the location of the `coordinateSystem` field is clearly wrong (VRML definition: https://github.com/cyberbotics/webots/blob/master/resources/nodes/WorldInfo.wrl#L5-L20) and the test suite seems to save it differently from you (maybe you didn't select all the lines when committing?).

lukicdarkoo:
How did the tests pass:
https://github.com/cyberbotics/webots/runs/1872916701?check_suite_focus=true

Can we catch these errors before the nightly build?

omichel:
Only quick tests on the source code are automatically performed with each PR. Longer and in depth tests (like "world update" and "ros tests") are done at night. We should monitor them and fix them if they fail.

stefaniapedrazzi:
Tests can also be triggered by setting the corresponding label in the PR and not only on nightly builds. 
But we are rarely running the tests.

lukicdarkoo (approved):
Thank you!

PR #2755, opened 2021-02-16
Improve webots server caching 

BenjaminDeleze:
**Description**
Webots does no check to verify if a streaming-viewer client already has an image in its cache or not.

To fix that, I implemented a simple etag mecanism.
- When webots send a file, it put the hash off the image in the etag header.
- If a browser want to query again webots about a file, it will put the hash from the etag field in the "If-None-Match" header. (this is done automatically by the browser)
- When receiving a query about a file, webots will check if their is a "If-None-Match" header and compare its content with the hash of the requested file. If the two hashes are the same, webots will answer with a 304 Not modified and not transfer the file again.

I implemented it by hand because I did not find such an preexisting option in Qt.

### DIFF
--- a/src/webots/gui/WbStreamingServer.cpp
+++ b/src/webots/gui/WbStreamingServer.cpp
@@ -174,7 +174,7 @@ void WbStreamingServer::onNewTcpData() {
     const QString &requestedUrl(tokens[1].replace(QRegExp("^/"), ""));
     if (!requestedUrl.isEmpty()) {  // "/" is reserved for the websocket.
       bool hasEtag = false;
-      QString etag = "";
+      QString etag;
       for (const auto &i : tokens) {
         if (i == "If-None-Match:")
           hasEtag = true;
--- a/src/webots/gui/WbStreamingServer.cpp
+++ b/src/webots/gui/WbStreamingServer.cpp
@@ -176,9 +176,9 @@ void WbStreamingServer::onNewTcpData() {
       bool hasEtag = false;
       QString etag = "";
       for (const auto &i : tokens) {
-        if (i == "If-None-Match:") {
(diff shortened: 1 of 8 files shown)


src/webots/gui/WbStreamingServer.cpp
```diff
@@ -172,12 +172,23 @@ void WbStreamingServer::onNewTcpData() {
   QStringList tokens = QString(line).split(QRegExp("[ \r\n][ \r\n]*"));
   if (tokens[0] == "GET") {
     const QString &requestedUrl(tokens[1].replace(QRegExp("^/"), ""));
-    if (!requestedUrl.isEmpty())  // "/" is reserved for the websocket.
-      sendTcpRequestReply(requestedUrl, socket);
+    if (!requestedUrl.isEmpty()) {  // "/" is reserved for the websocket.
+      bool hasEtag = false;
+      QString etag = "";
```
omichel, line 177: By default QStrings are empty:
```suggestion
      QString etag;
```

src/webots/gui/WbStreamingServer.cpp
```diff
@@ -172,12 +172,23 @@ void WbStreamingServer::onNewTcpData() {
...
+      for (const auto &i : tokens) {
+        if (i == "If-None-Match:") {
+          hasEtag = true;
+        } else if (hasEtag) {
```
omichel, line 181: Coding style:
```suggestion
        if (i == "If-None-Match:")
          hasEtag = true;
        else if (hasEtag) {
```

omichel (commented):
That seems to be a very good implementation.
I believe you are correct, there is no better way to do it with Qt.

omichel (approved):
Thank you.

PR #2756, opened 2021-02-16
TransmissionJoint specifications

ad-daniel:
Proposition for the specifications of the `TransmissionJoint`:

`TransmissionJoint` derived from `HingeJoint`

```
TransmissionJoint {
  field SFNode  jointParameters2  NULL   # { HingeJointParameters, PROTO } (driven side), required
  field SFFloat backlash          0      # (-inf, inf)
  field SFFloat multiplier        1      # (-inf, inf), required
  field SFNode startPoint         NULL   # {Solid, Slot, PROTO}
}
```

for **classic gear**:
- the axis of jointParameters and jointParameters2 should match.
- multiplier should be negative

from ODE standpoint, nothing special, we can feed the multiplier as a ratio directly.

for **chain drive**:

- the axis of jointParameters and jointParameters2 should match.
- multiplier should be positive

from ODE standpoint, chain drive transmission requires the radii of the two wheels so in practice we impose: `radius1 = 1` and `radius2 = multiplier`.

for **bevel gears**:
- the axis of jointParameters and jointParameters2 should differ.
- ~~the multiplier should be negative.~~

ODE doesn't have a way of specifying the ratio for bevel gears, it's inferred from the provided axis.

Implementation notes: 
- the two axis must intersect, show warning otherwise. ODE accepts it anyway but the results are meaningless.
- bevel gear is consistent with how it's defined in ODE, but not ideal from a user perspective as the multiplier is inferred from the axis. So webots documentation should clarify how the multiplier is inferred under the hood and that the provided value isn't being enforced. As it stands, this type of joint wouldn't be easy to use in practice.


omichel:
The `jointParameter` field should not be mentioned here as it is inherited from the [Joint](https://cyberbotics.com/doc/reference/joint) abstract node.
Also, I believe we miss the `startPoint` as discussed yesterday.
I would get rid of the `variant` field as it can be deduced from the `jointParameter2` which should be mandatory in all cases:

- for **classic gear**:
  - the axis of jointParameters and jointParameters2 should match.
  - multiplier should be negative

- for **chain drive**:
  - the axis of jointParameters and jointParameters2 should match.
  - multiplier should be positive

- for **bevel gears**:
  - the axis of jointParameters and jointParameters2 should differ.
  - the multiplier should be negative.

Note: the radii of the gears should be defined in the endPoint and startPoint only. Their ratio should match the multiplier.

What do you think?

ad-daniel:
The implementation currently lacks the startPoint and a lot of clean up, but is otherwise mostly functional. The fundamental issue of this approach is that at the engine level when computing the physics step it uses the global CFM which in most situations is fairly high whereas if one simply wants to model the backlash, having a high CFM introduces accumulating errors. This implementation is therefore on standby for the moment, and therefore closed for now.

PR #2768, opened 2021-02-19
Fix ENU in the Inertial Unit

lukicdarkoo:
This PR fixes ENU coordinate system handling by the Inertial Unit and improves documentation.

https://cyberbotics.com/doc/reference/inertialunit?version=fix-inertial-unit

I am not sure whether the rotation order is respected (roll/pitch/yaw order is fine, but at which order the rotations are applied).


omichel (changes requested):
It looks good, but I would add a line in the change log about it.

omichel (approved):
Thank you.

Issue #2584, opened 2020-12-17
Webots physics is applied even though `webots_physics_collide` returns 1 (related to the Group node)


lukicdarkoo:
**Steps to Reproduce**
1. Download [physics_callback_bug_group.zip](https://github.com/cyberbotics/webots/files/5709880/physics_callback_bug_group.zip)
2. Let `GROUP_SOLID` fall over `SINGLE_SOLID`
3. See the `GROUP_SOLID` node bouncing off the `SINGLE_SOLID`

**Expected behavior**
The `GROUP_SOLID` should go through the `SINGLE_SOLID` node since `webots_physics_collide` returns 0.

lukicdarkoo:
Fixed in #2781

PR #2775, opened 2021-02-22
Physics-plugin-group-fix

Simon-Steinmann:
Should fix this issue: https://github.com/cyberbotics/webots/issues/2584

To my understanding, if one or both bounding objects were groups (group means space NOT geom), then the collision was executed before we checked for physics plugin. I moved that part AFTER the physics plugin check. This should fix the issue.  I cannot compile and test it at the moment though.

### DIFF
--- a/src/webots/engine/WbSimulationCluster.cpp
+++ b/src/webots/engine/WbSimulationCluster.cpp
@@ -429,6 +429,18 @@ static bool needCollisionDetection(WbSolid *solid, bool isOtherRayGeom) {
   return false;
 }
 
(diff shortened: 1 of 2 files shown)


lukicdarkoo:
Now the group node never applies the default collision handling, regardless of the return value from `webots_physics_collide` :(

lukicdarkoo:
Fixed in 776778e. @Simon-Steinmann, let me know if you see any side-effect.

omichel:
We should indeed close this PR and open a new one with the same content but aiming at the master branch.

PR #2785, opened 2021-02-26
`wb_supervisor_node_restart_controller` doesn't restart external controllers

lukicdarkoo:


docs/reference/supervisor.md
```diff
@@ -1447,6 +1447,7 @@ wb_supervisor_node_restart_controller(node)
 The `wb_supervisor_node_restart_controller` function restarts the controller of the Robot passed to it.
 If a node other than a [Robot](robot.md) is passed to this function, no change is effected, and a warning message is printed to the console.
 Note that if a robot window is specified for the [Robot](robot.md) node, the robot window will be restarted as well.
+If an external controller is used then this function will only terminate the process, it will not start the controller.
```
omichel, line 1450: The official terminology is "extern", not "external".
```suggestion
If an extern controller is used, then this function will only terminate the process, it will not start the controller.
```

omichel (changes requested):
Yes, this is good to provide this precision.
Thank you.

omichel (approved):

PR #2786, opened 2021-02-26
HingeJointWithBacklash proto to model effect of backlash

ad-daniel:
**EDIT: This information is outdated, see comments below.**

# Specifications 
For the purpose of modeling mechanical backlash in Webots, in #2756 the option to introduce ODE's transmission joint was considered but this approach suffers from some drawbacks. Most notably, as it relies on the contact between the two wheels to  transmit the torque, deviations between input and output are introduced due to the effect of CFM at the transmission level, and these errors accumulate over time which makes this approach undesirable for the current objective.

The second more promising approach is to model backlash by extending the HingeJoint class and relying on minStop maxStop to achieve the desired effect. 

```
HingeJoint {
  # current fields
  field MFNode  device            [ ]   # {RotationalMotor, PositionSensor, Brake, PROTO}
  field SFFloat position          0     # [0, inf)
  # added fields  
  field SFFloat backlash          0     # [0, inf)
  field SFFloat gearMass          0     # [0, inf)
  field SFNode  startPoint        NULL  # {Transform} it's only for visualizing the gear
}
```

# Preliminary result
The following videos show the results that can be obtained using the proposed approach and are obtained by cascading two HingeJoints. In **red** is shown the behavior of a classic HingeJoint (with no backlash) whereas in **blue** the case with a very large (0.2) backlash, just to make things more evident. Both motors are given the same command.

No backlash, both pendulums (single hinge and cascaded hinge) behave the same

https://user-images.githubusercontent.com/44834743/109315387-c69ca600-784a-11eb-9f55-1d27f9743138.mp4

Without any gravity:

https://user-images.githubusercontent.com/44834743/109310957-7707ab80-7845-11eb-9d8b-be487de20f62.mp4

With gravity:

https://user-images.githubusercontent.com/44834743/109311012-8a1a7b80-7845-11eb-8b38-72ca8889ebb6.mp4


omichel:
I have been thinking overnight about this approach and I believe I suggested the wrong way.
Modifying the `HingeJoint` node will introduce a lot of complexity in Webots due to the non-linearity of the hierarchical scene structure (the `HingeJoint` will contains two actual ODE joints with two attached bodies). This will complicate the export to VRML/X3D, web interface, web animations, etc. Also, at some point we will want to introduce backlash for `Hinge2Joint` as well (since the Darwin-OP and NAO robots, for example, have several Hinge2Joint nodes). This will introduce even more complexity.

I believe a better approach is to create a `HingeJointWithBacklash` LUA proto that will create the necessary nodes (the two `HingeJoint` nodes to implement the backlash:
```js
HingeJointWithBacklash {
  field SFNode  jointParameters   NULL
  field SFNode  endPoint          NULL
  field MFNode  device            [ ]   # {RotationalMotor, PositionSensor, Brake, PROTO}
  field SFFloat position          0     # [0, inf)
  field SFFloat backlash          0.1     # [0, inf)
  field SFFloat gearMass          0.01     # [0, inf)
  field SFNode  startPoint        NULL  # {Group, Transform, or Shape} it's only for visualizing the gear
}

Hence this would require no modification to the source code of Webots and a similar approach may be taken to implement a `Hinge2JointWithBacklash` proto. Users willing to upgrade their existing `HingeJoint` nodes to include backlash should simply rename them to `HingeJointWithBacklash` and optionally adjust the default parameters.

I am sorry for suggesting a too complex approach in the first place, but I believe this one is way better. What do you think?

**Comment by ad-daniel** [comment]
Yes, it has been on my mind as well and it does make more sense for it to be a proto. This approach not only does it increase the complexity but I don't believe it would provide any actual performance improvement since compared to the demo in the videos you still need to spawn the 2 hinges/bodies just the same as it's the interaction between the two that achieves the backlash effect, so even without considering the rise in complexity, refactoring the class wouldn't really bring much to the table performance-wise.

**Comment by ad-daniel** [comment]
This feature implements the HingeJointWithBacklash PROTO.

For a more realistic modeling, it was necessary to expose the StopERP and StopCFM parameters from ODE (in HingeJointParameters) in order to make the local collision between the 2 joints more stiff. When using the global ERP values (which is the default), it would continuously over-correct making the output axis accelerate. In a pendulum-like situation similar to the examples shown in the first post, what would happen is that after some time the blue pendulum (the one with backlash) would end-up overtaking the one without (red) which doesn't make physical sense.

I'm not entirely convinced the `outputSensor` field provides meaningful information to the user.

## Sample World 
A sample world to see the hinge at work is also provided

https://user-images.githubusercontent.com/44834743/109947986-30e59880-7cda-11eb-9285-fd13f8ec68c3.mp4

File: docs/guide/hinge-joint-with-backlash.md [hunk 1]
```diff
@@ -0,0 +1,50 @@
...
+
+### HingeJointWithBacklash Model
+
+```
+HingeJointWithBacklash [
+  SFNode       jointParameters   HingeJointParameters {} 
+  SFFloat      backlash          0.1    # (0, inf)
+  SFFloat      gearMass          0.01   # [0, inf)
+  MFNode       device            [ ]    # {RotationalMotor, PositionSensor, Brake, PROTO}
+  MFNode       outputSensor      [ ]    # {PositionSensor}
```
omichel, line 19: It doesn't make sense to have MFNode type here. Why not using SFNode instead?

docs/guide/hinge-joint-with-backlash.md
```diff
@@ -0,0 +1,50 @@
...
+Derived from [HingeJoint](../reference/hingejoint.md).
+
+### HingeJointWithBacklash Model
+
+```
+HingeJointWithBacklash [
+  SFNode       jointParameters   HingeJointParameters {} 
+  SFFloat      backlash          0.1    # (0, inf)
+  SFFloat      gearMass          0.01   # [0, inf)
+  MFNode       device            [ ]    # {RotationalMotor, PositionSensor, Brake, PROTO}
```
omichel, line 18: PROTO could be mentioned everywhere, it's better to skip it.
```suggestion
  MFNode       device            [ ]    # {RotationalMotor, PositionSensor, Brake}
```

docs/guide/hinge-joint-with-backlash.md
```diff
@@ -0,0 +1,50 @@
...
+If no motor is specified, the joint is a passive joint.
+
+- `outputSensor`: this field optionally specifies an angular [PositionSensor](../reference/positionsensor.md) fixed on the output axis of the joint.
+
+> **Note**: what this sensor returns is the current position of the output axis in the range [-backlash/2, backlash/2]. When either limit is reached, the axis will begin to move.
+
+- `startPoint`: this field optionally specifies the shape of the object attached to the axis at the input of the joint. It must be either a [Transform](../reference/transform.md), a [Group](../reference/group.md) or a [Shape](../reference/shape.md).
+This object doesn't affect the joint in any way but can be useful in order to better visualize the effect of the backlash.
+
+- `endPoint`: this field specifies which [Solid](../reference/solid.md) is subjected to the joint constraints. It must be either a [Solid](../reference/solid.md) child, or a reference to an existing [Solid](../reference/solid.md), i.e. a [SolidReference](../reference/solidreference.md).
```
omichel, line 46:
```suggestion
- `endPoint`: this field specifies which [Solid](../reference/solid.md) is subject to the joint constraints. It must be either a [Solid](../reference/solid.md) child, or a reference to an existing [Solid](../reference/solid.md), i.e. a [SolidReference](../reference/solidreference.md).
```

docs/guide/joints.md
```diff
@@ -0,0 +1,10 @@
+# Joints
+
+Webots already disposes of several nodes to model the transmission of power across bodies, such as [HingeJoint](../reference/hingejoint.md), [Hinge2Joint](../reference/hinge2joint.md), [BallJoint](../reference/balljoint.md) and [SliderJoint](../reference/sliderjoint.md).
+However, in some instances it is necessary to more accurately describe the real behavior by taking into consideration factors like, for example, the clearance among gears in a drive train. The [PROTO](../reference/proto.md) described here extend the existing nodes by allowing more accurate modeling.
```
omichel, line 4:
```suggestion
However, in some instances it is necessary to more accurately describe the real behavior by taking into consideration factors like, for example, the clearance among gears in a drive train. The [PROTO](../reference/proto.md) node described here extend the existing nodes by allowing more accurate modeling.
```

projects/joints/protos/HingeJointWithBacklash.proto
```diff
@@ -0,0 +1,82 @@
...
+# license: Copyright Cyberbotics Ltd. Licensed for use only with Webots.
+# license url: https://cyberbotics.com/webots_assets_license
+# tags: static
+# This extension of the hinge joint allows to model the effect of backlash.
+
+PROTO HingeJointWithBacklash [
+  field SFNode                                 jointParameters   HingeJointParameters {}
+  field SFFloat                                backlash          0.01   # Defines the gear clearence, [0, inf).
+  field SFFloat                                gearMass          0.01   # Defines the gear mass, [0, inf).
+  field MFNode                                 device            [ ]    # {RotationalMotor, PositionSensor, Brake, PROTO}.
```
omichel, line 11:
```suggestion
  field MFNode                                 device            [ ]    # {RotationalMotor, PositionSensor, Brake}.
```

docs/reference/changelog-r2021.md
```diff
@@ -26,6 +26,7 @@ Released on June, Xth, 2021.
       - Exposed the `robot_description` ROS parameter (activated through the `--robot-description` flag) that contains the URDF of the robot.
   - New Samples:
     - Added a simple room with a Nao robot ([#2701](https://github.com/cyberbotics/webots/pull/2701)).
+    - Added HingeJointWithBacklash proto that extends [HingeJoint](hingejoint.md) to model the effect of backlash and a corresponding sample world ([#2786](https://github.com/cyberbotics/webots/pull/2786)).
```
omichel, line 29: We should also add a line about the new fields of the HingeJointParameters (stopEPF and stopCFM).

omichel (changes requested):
This is only a partial review, but the quality of this PR looks very good overall.
I have a meeting now, so I will stop for now and get back to it later.

docs/guide/hinge-joint-with-backlash.md
ad-daniel: `device` is a MFNode in `HingeJoint` and when attributing `outputSensor` through a IS statement, it throws a parameter mismatch error if `outputSensor` isn't also `MF`. It's an intended behavior according to the [docs](https://cyberbotics.com/doc/reference/proto-definition#is-statements).

> It is also an error if the type of the field being associated does not match the type declared in the PROTO's interface. For example, it is illegal to associate an SFColor with an SFVec3f. It is also illegal to associate a SFColor with a MFColor or vice versa. Results are undefined if a field of a node in the PROTO body is associated with more than one field in the PROTO's interface.

docs/reference/changelog-r2021.md
ad-daniel: Right, I noticed I didn't update the `HingeJointParameters` docs about that addition either so I fixed it now.

docs/guide/hinge-joint-with-backlash.md
omichel: OK, I understand. Let's keep it like that. Thank you for the explanation.

omichel (approved):
It looks all good to me now. Thank you.

PR #2787, opened 2021-02-26
Drastically reduce package size

omichel:
Following up #2591, this PR introduces Webots URLs. Webots URLs can be used in a world file to specify textures, meshes, sounds, etc. in a similar way as the https URLs introduced in #2591. Webots URLs come as an alternative to relative or absolute file references and identify a unique file in the Webots package.
A Webots URL has the following form: `webots://` following by a file path relative to `WEBOTS_HOME`.
For example: [webots://projects/samples/geometries/worlds/textures/webots_box.png](https://raw.githubusercontent.com/cyberbotics/webots/R2021a/projects/samples/geometries/worlds/textures/water_flow.jpg)

When releasing a new webots package, all world and proto files will be processed and all `webots://` URLs will be replaced with their equivalent `https://raw.githubusercontent.com/cyberbotics/webots/` URL, so that all the resources files (textures, meshes, sounds, etc.) can be omitted from the distributed package. This will significantly reduce the size of the Webots packages.

Released version will use the GitHub tag to refer to resources, e.g., R2021a, R2021a-rev1, R2021b, etc.
Nightly builds cannot use the GitHub tag for nightly builds because these tags are destroyed after 3 days. So, instead, they should use the commit sha.

- [x] Implement support for `webots://` URLs in Webots (useful only for Webots developers who checked out the webots repo).
- [ ] ~Add a "Download all resources for offline use" menu item (will download all resources in the cache of Webots).~
- [x] Replace the file URLs with `webots://` in all world files.
- [x] Replace the file URLs with `webots://` in all proto files.
    The following protos were converted manually as the script failed because of LUA code:
    - SimpleBuilding
    - PublicToilet
    - SimpleTree
    The following protos were skipped as the script failed because of LUA code, but no modifications was needed:
    - DirectionalPanel
    - ProtoMismatchFieldType (tests folder)
- [x] Replace the file URLs also in the webots_ros submodule. See https://github.com/cyberbotics/webots_ros/pull/59
- [x] Replace the file URLs also in the webots_ros2 repo. See https://github.com/cyberbotics/webots_ros2/pull/233
- [x] Replace the file URLs also in the webots-project. See https://github.com/cyberbotics/webots-projects/pull/4
- [x] Change the export of .x3d files to use `webots://` urls when possible.
- [x] Adapt webotsJS to support `webots://`
- [x] Replace the `webots://` of all world and proto files in the distribution.
- [x] Remove the resources files from the distribution.
- [x] Measure the resulting size of the new webots packages: on Windows, the new package is 0.99 GB instead of 1.64 GB.

Other ideas to reduce the package size:

- on Windows, remove the mingw32 binaries.
- remove the offline documentation, replace it with online documentation only.
- remove the e-puck cross-compiler &rarr; move it to a different repo.
- use EXTERNPROTO to move proto files outside of the distribution.
- move projects (worlds, protos, controllers) outside of the distribution.


omichel:
@stefaniapedrazzi and @lukicdarkoo: can you give me some initial feedback on this PR (from the above description) as it is going to be hopefully a big revolution in the Webots package size. The next steps to reduce package size will be to use EXTERNPROTO and maybe remove the offline documentation.

stefaniapedrazzi:
When you write about changing all the resources URLs in the sample worlds, do you mean to set them so that the one from the GitHub repo are taken?

This seems good, but personally I think that we cannot implement this method without a functional and easy to use method to let Webots samples run without internet connection.
From the Webots menu it should be possible to download the resources (stored in one or multiple packages), and after this (or by setting a variable) the resources path stored locally should be used by Webots instead of the online ones.

It is very good to reduce the size of the Webots package and also to be able to refer to resources stored on some servers.
But I think that we should not loose any functionality by adding these improvements and we should split the Webots package into subpackages that can be downloaded and installed separately from the core.

omichel:
We discussed and decided to add a new menu item to Webots entitled "Download all resources to work offline" or so, that will download all the resources (images, meshes, sounds, etc.) into the Webots cache so that any world file / proto file can be used offline without having to download resources.

lukicdarkoo:
Out of curiosity, I have done a brief analysis of the Webots R2021a size:
```
10.0193 dat
10.4443 exe
12.8514 qm
13.0948 csv
13.3451 h
13.4043 wav
13.8857 58
14.4276 a
14.8439 ttf
15.0028 hdr
16.6142 wbt
20.0493 pak
41.4246 xml
42.4677 1
97.1519 so
120.123 proto
195.309 0
564.859 jpg
697.568 png
```
Size in MBs of 2.1GBs in total. The `*.0` files (are actually `*.so` files) mostly come from SUMO, ROS dependencies, and Qt.

.github/workflows/test_suite_linux.yml
```diff
@@ -112,6 +112,7 @@ jobs:
       run: |
         export LIBGL_ALWAYS_SOFTWARE=true
         export WEBOTS_DISABLE_SAVE_SCREEN_PERSPECTIVE_ON_CLOSE=true
+        xvfb-run --auto-servernum python scripts/packaging/update_worlds_and_protos.py develop
```
stefaniapedrazzi, line 115: How long does it take to run this command?
If it takes long, then it should be moved in a different step (with dedicated label).
Something like it is done for the `world-update` step.

omichel: It takes 2 seconds on my laptop, so I guess it's OK.

stefaniapedrazzi (commented):
The file `tests/api/controllers/supervisor_import_remove/vrml_import.wrl` the URL have not been updated and it is still using the form `textures/cubic/noon_sunny_empty_right.jpg`.
This could cause issues at first run depending on which test is run before.

.github/workflows/test_suite_linux.yml
stefaniapedrazzi: When testing locally, for me it fails because of missing sha.
How do you make it work locally?

omichel: You should pass `develop` instead of sha.

stefaniapedrazzi: Ok. I did once but not the second time. So it seems to work.
However here it seems wrong to pass develop instead of the sha.

omichel: Yes, I believe we should pass the tag instead.

lib/controller/matlab/wb_supervisor_node_set_joint_position.m
```diff
@@ -0,0 +1,6 @@
+function result = wb_supervisor_node_set_joint_position(noderef, position, index)
```
stefaniapedrazzi, line 1: This file should be removed because it is automatically generated.

.github/workflows/test_suite_linux.yml
omichel: Actually, this is probably the reason why the test was failing... I will re-enable it and see whether it works now.

stefaniapedrazzi: Running this script before the tests, are the tests still checking the internal functionality of Webots to convert the `webots://` URL?

lib/controller/matlab/wb_supervisor_node_set_joint_position.m
omichel: Yes, I just removed it.

.github/workflows/test_suite_linux.yml
omichel: No, and this is probably not something we want to test here because:
- There is no image to test under webots:// included in the distribution that we use for testing.
- The worlds with a `webots://` URL are not distributed to users, but used only in the development environment, so it shouldn't affect users, but only developers and if it fails, we will probably spot it immediately.

lukicdarkoo (commented):
image

The `animated_skin` example is missing textures

src/webots/gui/WbPreferencesDialog.cpp
```diff
@@ -374,35 +382,63 @@ QWidget *WbPreferencesDialog::createOpenGLTab() {
...
+  layout->addWidget(new QLabel(tr("Password:"), this), 4, 0);
   layout->addWidget(mHttpProxyPassword, 4, 1);
 
+  // Cache
+  layout = new QGridLayout(cache);
+
+  // row 0
+  mCacheSize = new WbLineEdit(this);
+  mCacheSize->setValidator(new QIntValidator(0, 65535));
+  layout->addWidget(new QLabel(tr("Set the of the cache (in MB), current size: %1MB")
```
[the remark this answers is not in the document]
stefaniapedrazzi: Note that there is also a typo and `size` in "Set the *size* of the cache" is missing.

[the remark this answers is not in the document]
omichel: Oops, yes, thanks.

src/webots/Makefile
```diff
@@ -30,7 +30,7 @@ endif
 
 WB_ODE_INCLUDE           = $(QT_CORE_INCLUDE) $(QT_GUI_INCLUDE) $(ODE_INCLUDE) $(WREN_INCLUDE) $(CONTROLLER_INCLUDE) -Iapp -Icore -Imaths -Inodes -Inodes/utils -Ivrml -Iengine
 WB_MATHS_INCLUDE         = $(QT_CORE_INCLUDE)
-WB_CORE_INCLUDE          = $(QT_CORE_INCLUDE) $(QT_NETWORK_INCLUDE) $(QT_GUI_INCLUDE) $(QT_QML_INCLUDE) $(LUA_INCLUDE) -Iqml -Iexternal/compilation_timestamp
+WB_CORE_INCLUDE          = $(QT_CORE_INCLUDE) $(QT_NETWORK_INCLUDE) $(QT_GUI_INCLUDE) $(QT_QML_INCLUDE) $(LUA_INCLUDE) -Iqml -Iexternal/compilation_timestamp -Iapp
```
stefaniapedrazzi, line 33: This is adding a circular dependency between the `core` and `app` modules.
This is bad and should be avoided.

src/webots/core/WbDownloader.hpp
```diff
@@ -16,6 +16,7 @@
 #define WB_DOWNLOADER_HPP
 
 #include <QtCore/QObject>
+#include <QtCore/QTimer>
```
stefaniapedrazzi, line 19: This include seems to be useless.

src/webots/core/WbPreferences.cpp
```diff
@@ -28,6 +28,7 @@
 #include <QtCore/QRegularExpressionMatch>
 #include <QtCore/QStandardPaths>
 
+#include <iostream>
```
stefaniapedrazzi, line 31: Is this include really needed?
Because it doesn't seem that related changes have been added.

resources/stylesheet.windows.qss
```diff
@@ -0,0 +1,7 @@
+/********************************/
+/* Windows-specific style sheet for Webots */
+/********************************/
+
+QDialog * {
+  font-size: 13px;
```
stefaniapedrazzi, line 6: Now the `font-size` is set to the same value for all the platforms, so it is no longer platform specific.
Is it really this file the better place where to define it?

src/webots/nodes/WbMesh.cpp
```diff
@@ -99,11 +111,18 @@ void WbMesh::updateTriangleMesh(bool issueWarnings) {
...
                        aiProcess_JoinIdenticalVertices | aiProcess_OptimizeGraph | aiProcess_RemoveComponent;
   if (WbUrl::isWeb(filePath)) {
-    const QByteArray data = mDownloader->device()->readAll();
-    const char *hint = filePath.mid(filePath.lastIndexOf('.') + 1).toUtf8().constData();
-    scene = importer.ReadFileFromMemory(data.constData(), data.size(), flags, hint);
-    delete mDownloader;
-    mDownloader = NULL;
+    if (mDownloader == NULL) {
+      downloadAssets();
+    }
```
stefaniapedrazzi, line 116: unneeded brackets
```suggestion
    if (mDownloader == NULL)
      downloadAssets();
```

update_worlds_and_protos.py
```diff
@@ -0,0 +1,80 @@
+import os
```
stefaniapedrazzi, line 1: Is this file used anywhere?
Because it has exactly the same name as the one added in `scripts/packaging`.
I tried to check but I found just references to the one in `scripts/packaging`.

If it used, then there are some improvements to apply:
* header `#!/usr/bin/env python` and license are missing in this file
* should be moved in the `scripts` folder (or any other folder) and not be located in the main folder

stefaniapedrazzi (changes requested):
Here are a couple of other comments about the code.
I'm still performing some tests, but globally seems to work fine.

update_worlds_and_protos.py
omichel: I believe it's not used. I removed it.

stefaniapedrazzi:
The robot web components on the documentation are not loading for me in this branch.
Not sure if it depends because some of the modified files needs to be manually updated on the server.

resources/stylesheet.windows.qss
omichel: We should probably define a common `stylesheet.qss` file to store common values, but this is probably out of the scope of this PR. I will open an issue for that.

omichel:
> The robot web components on the documentation are not loading for me in this branch.
Not sure if it depends because some of the modified files needs to be manually updated on the server.

Yes, it seems the URL is badly computed: https://cyberbotics.com/doc/guide/aibo-ers7?version=feature-webots-urls

```
Error in loading : https://raw.githubusercontent.com/cyberbotics/webots/feature-webots-urls/docs/guide/scenes/aibo-ers7/webots:/projects/default/worlds/textures/cubic/mountains_back.png
```

Instead it should load https://raw.githubusercontent.com/cyberbotics/webots/feature-webots-urls/projects/default/worlds/textures/cubic/mountains_back.png

omichel:
> 
> 
> image
> 
> The `animated_skin` example is missing textures

That should now be fixed.

stefaniapedrazzi:
It would be more user-friendly to enhance the preferences dialog so that is doesn't display just the maximum size of the cache, but also the size currently in use.
Otherwise, even after pressing the button to clear the cache there is no real feedback (other than the automatic message) that shows that the cache has been really deleted.
But for this version it is fine like this and we can improve it later.

stefaniapedrazzi:
The documentation page about preference should also be updated.
It would also be good to specify the default path where the cache is stored.
https://www.cyberbotics.com/doc/guide/preferences?version=feature-webots-urls#network

stefaniapedrazzi:
I will take care of updating the documentation.

stefaniapedrazzi:
Robot documentation is working correctly after updating the JS scripts.

So it seems that the only remaining issues are the one with the mavic (#3245) and the circular dependency between `core` and `app` modules.

omichel:
I believe I fixed both the circular dependency and propeller helix appearance.

stefaniapedrazzi:
I get some compilation errors:
```
nodes/utils/WbDownloader.cpp:22:10: fatal error: QtNetwork/QNetworkReply: No such file or directory
 #include <QtNetwork/QNetworkReply>
          ^~~~~~~~~~~~~~~~~~~~~~~~~
```

omichel:
> I get some compilation errors:
> 
> ```
> nodes/utils/WbDownloader.cpp:22:10: fatal error: QtNetwork/QNetworkReply: No such file or directory
>  #include <QtNetwork/QNetworkReply>
>           ^~~~~~~~~~~~~~~~~~~~~~~~~
> ```

That was compiling on Windows. I believe I fixed it now.

src/webots/Makefile
```diff
@@ -41,7 +41,7 @@ WB_PLUGINS_INCLUDE       = $(QT_CORE_INCLUDE) $(ODE_INCLUDE) -Icore -Imaths -Ino
 WB_ENGINE_INCLUDE        = $(QT_CORE_INCLUDE) $(QT_NETWORK_INCLUDE) $(ODE_INCLUDE) $(WREN_INCLUDE) -Icore -Imaths -Inodes -Inodes/utils -Iplugins -Isound -Iuser_commands -Iutil -Ivrml
 WB_CONTROL_INCLUDE       = $(QT_CORE_INCLUDE) $(QT_NETWORK_INCLUDE) -Icore -Iengine -Imaths -Inodes -Inodes/utils -Iode -Ivrml -Iutil
 # dependency of NODES on QT_GUI is necessary in WbImageTexture to retrieve the size of a texture before loading it in Wren
-WB_NODES_INCLUDE         = $(QT_CORE_INCLUDE) $(QT_GUI_INCLUDE) $(ODE_INCLUDE) $(CONTROLLER_INCLUDE) $(OIS_INCLUDE) $(FREETYPE_INCLUDE) $(WREN_INCLUDE) $(STB_INCLUDE) $(ASSIMP_INCLUDE) -Iapp -Icore -isystem external/siphash -Imaths -Inodes -Inodes/utils -Iode -Iplugins -Isound -Iutil -Ivrml
+WB_NODES_INCLUDE         = $(QT_CORE_INCLUDE) $(QT_NETWORK_INCLUDE) $(QT_GUI_INCLUDE) $(ODE_INCLUDE) $(CONTROLLER_INCLUDE) $(OIS_INCLUDE) $(FREETYPE_INCLUDE) $(WREN_INCLUDE) $(STB_INCLUDE) $(ASSIMP_INCLUDE) -Iapp -Icore -isystem external/siphash -Imaths -Inodes -Inodes/utils -Iode -Iplugins -Isound -Iutil -Ivrml
```
stefaniapedrazzi, line 44: I don't think that it is good to add this dependency to the Qt network from the nodes module.
I think that this code deserves a better refactoring fixing properly the relation between ``WbApplication``, ``WbNetwork`` and ``WbDownload`` classes.

stefaniapedrazzi: For example all the static functionalities in `WbDownload` seems that could be moved to the singleton `WbNetwork` class instead.
Doing this it would be easier to switch the signal/slot relationship between `WbNetwork` and `WbApplication`.

stefaniapedrazzi:
> That was compiling on Windows. I believe I fixed it now.

It is fixed.

src/webots/Makefile
omichel: Moving the `reset` and `progress` static method to `WbNetwork` would be a mess as they do not refer to `WbNetwork`, but to the `WbDownloader`. So, I don't see the point here.

Also, it would be possible to remove the dependency of `nodes/utils` on `QtNetwork`, but that would force us to create some proxy class or methods (maybe in `WbNetwork`) to access the `QtNetwork` functionality from `nodes/utils`, which is a useless complication in my point of view, given that `nodes/utils` depends on `core` which in turn depends on `QtNetwork`. So we are not introducing any new dependency there, we are just using a direct access to `QtNetwork` which is more readable that creating a proxy class or proxy methods in `core`.

If you want, we can discuss this by voice, as I am not sure to understand you point here.

stefaniapedrazzi (dismissed):
Thank you

stefaniapedrazzi:
The issue with the propeller is also fixed for me.

.github/workflows/test_suite_linux.yml
```diff
@@ -112,6 +112,7 @@ jobs:
       run: |
         export LIBGL_ALWAYS_SOFTWARE=true
         export WEBOTS_DISABLE_SAVE_SCREEN_PERSPECTIVE_ON_CLOSE=true
+        xvfb-run --auto-servernum python scripts/packaging/update_worlds_and_protos.py ${{ github.sha }}
```
lukicdarkoo, line 115: Should we do this for macOS and Windows? Also, should we target a tag for the release?

omichel: For the release, it is done in https://github.com/cyberbotics/webots/blob/0bbd6fa4b003062b76e123e29a9d945e49d374c4/scripts/packaging/Makefile#L70-L73. For the tests, they are currently not implemented for macOS and Windows.

stefaniapedrazzi (approved):

lukicdarkoo (approved):
Thank you

omichel:
🚀

PR #2794, opened 2021-03-01
Update immersionproperties.md

omichel:
There is a typo in the documentation.
The corresponding formula in the source code is here:
https://github.com/cyberbotics/webots/blob/master/src/ode/ode/src/fluid_dynamics/ode_fluid_dynamics.cpp#L277-L278.


stefaniapedrazzi (approved):
Thank you

PR #2797, opened 2021-03-01
Separate visible and hidden fields more clearly in the docs

ad-daniel:
For consistency with the rest of the docs, hidden fields in joint nodes shouldn't be mentioned. The position field is visible in JointParameters and should be mentioned only there.

### DIFF
--- a/docs/reference/joint.md
+++ b/docs/reference/joint.md
@@ -39,5 +39,5 @@ These fields, which are not visible from the Scene Tree, are used to store insid
 As a result joint positions are restored when reloading the simulation just the same way they would be if [JointParameters](jointparameters.md) nodes were used.
 
 For [HingeJoint](hingejoint.md) and [SliderJoint](sliderjoint.md) nodes containing no [JointParameters](jointparameters.md), Webots uses the hidden field named `position`.
-For a [Hinge2Joint](hingejoint.md) node, an additional hidden field named `position2` is used to store the joint position with respect the second hinge.
+For a [Hinge2Joint](hingejoint.md) node, an additional hidden field named `position2` is used to store the joint position with respect to the second axis.
 For a [BallJoint](balljoint.md) node, an additional hidden field named `position3` is used to store the joint position with respect the third axis.
--- a/docs/reference/joint.md
+++ b/docs/reference/joint.md
@@ -34,9 +34,10 @@ Alternatively, a [Slot](slot.md) node can be inserted in the `endPoint` field, b
 ### Joint's Hidden Position Fields
 
 If the `jointParameters` is set to NULL, joint positions are then not visible from the Scene Tree.
-In this case Webots keeps track of the initial positions of [Joint](#joint) nodes (except for the [BallJoint](balljoint.md)) by means of hidden position fields.
+In this case Webots keeps track of the initial positions of [Joint](#joint) nodes by means of hidden position fields.
 These fields, which are not visible from the Scene Tree, are used to store inside the world file the current joint positions when the simulation is saved.
 As a result joint positions are restored when reloading the simulation just the same way they would be if [JointParameters](jointparameters.md) nodes were used.
 
 For [HingeJoint](hingejoint.md) and [SliderJoint](sliderjoint.md) nodes containing no [JointParameters](jointparameters.md), Webots uses the hidden field named `position`.
 For a [Hinge2Joint](hingejoint.md) node, an additional hidden field named `position2` is used to store the joint position with respect the second hinge.
+For a [BallJoint](balljoint.md) node, an additional hidden field named `position3` is used to store the joint position with respect the third axis.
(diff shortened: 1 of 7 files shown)


stefaniapedrazzi:
Note that joint hidden fields are particular hidden fields because they can be saved by Webots.
These fields will appear in world files if a user will save a simulation after run.

This is exactly the same for the Solid hidden fields `linearVelocity` and `angularVelocity` that are also documented:
https://www.cyberbotics.com/doc/reference/solid#solid-fields

stefaniapedrazzi:
Also note that the [Joints Hidden Position Fields](https://www.cyberbotics.com/doc/reference/joint#joints-hidden-position-fields) is outdated and doesn't mention the BallJoint `position3` hidden field.
If addressing this issue about documentation of hidden fields, we should also update the hidden position fields section and add the BallJoint `position3` field.

stefaniapedrazzi (changes requested):
Personally I would keep the documentation of these hidden fields and update the joint hidden position fields section.

ad-daniel:
The reason for the change is that as it stands it could be misinterpreted as an actual field or as if the documentation is not up to date (as they wouldn't show up in the interface). Given what you said, then it makes sense to keep them. So I've separated them to make  it more clear as per Olivier's suggestion, let me know if it's better now.

docs/reference/joint.md
```diff
@@ -34,9 +34,10 @@ Alternatively, a [Slot](slot.md) node can be inserted in the `endPoint` field, b
 ### Joint's Hidden Position Fields
 
 If the `jointParameters` is set to NULL, joint positions are then not visible from the Scene Tree.
-In this case Webots keeps track of the initial positions of [Joint](#joint) nodes (except for the [BallJoint](balljoint.md)) by means of hidden position fields.
+In this case Webots keeps track of the initial positions of [Joint](#joint) nodes by means of hidden position fields.
 These fields, which are not visible from the Scene Tree, are used to store inside the world file the current joint positions when the simulation is saved.
 As a result joint positions are restored when reloading the simulation just the same way they would be if [JointParameters](jointparameters.md) nodes were used.
 
 For [HingeJoint](hingejoint.md) and [SliderJoint](sliderjoint.md) nodes containing no [JointParameters](jointparameters.md), Webots uses the hidden field named `position`.
 For a [Hinge2Joint](hingejoint.md) node, an additional hidden field named `position2` is used to store the joint position with respect the second hinge.
```
stefaniapedrazzi, line 42:
```suggestion
For a [Hinge2Joint](hingejoint.md) node, an additional hidden field named `position2` is used to store the joint position with respect to the second axis.
```

docs/reference/joint.md
```diff
@@ -34,9 +34,10 @@ Alternatively, a [Slot](slot.md) node can be inserted in the `endPoint` field, b
...
+For a [BallJoint](balljoint.md) node, an additional hidden field named `position3` is used to store the joint position with respect the third axis.
```
stefaniapedrazzi, line 43:
```suggestion
For a [BallJoint](balljoint.md) node, an additional hidden field named `position3` is used to store the joint position with respect to the third axis.
```

docs/reference/solid.md
```diff
@@ -30,7 +31,7 @@ The [Solid](#solid) class is the base class for collision-detected objects.
 Robots and device classes are subclasses of the [Solid](#solid) class.
 In the 3D window, [Solid](#solid) nodes can be manipulated (dragged, lifted, rotated, etc) using the mouse.
 
-### Solid Fields
+### Solid Field Summary
```
stefaniapedrazzi, line 34: For consistency with other pages I would also remove the `Solid` word

```suggestion
### Field Summary
```

docs/reference/solid.md
```diff
@@ -78,6 +79,8 @@ The colors defined in this field are returned for this object if recognized by a
 - `translationStep` and `rotationStep`: these fields specify the minimum step size that will be used by the translate and rotate handles appearing in the 3D window when selecting a top solid.
 Continuous increment is obtained by setting the step value to -1.
 
+### Hidden Solid Field Summary
```
stefaniapedrazzi, line 82: For consistency with other pages I would also remove the `Solid` word

```suggestion
### Hidden Field Summary
```

stefaniapedrazzi (changes requested):
Thank you! Yes, now it is better.
I just have a couple of comments mainly regarding the text already existing before this change.

docs/reference/solid.md
omichel, line 34: For consistency, shouldn't we remove "Solid" from here:
```suggestion
### Field Summary
```

docs/reference/solid.md
omichel, line 82: ... and here:
```suggestion
### Hidden Field Summary
```

omichel:
Oops, you can ignore my comments as @stefaniapedrazzi had the same.

stefaniapedrazzi (approved):
Thank you!

PR #2799, opened 2021-03-01
Fix Lidar.getLayerRangeImage returning full image on Python and Java

### LINKED ISSUE #2672: getLayerRangeImage() and getRangeImage() return the same list
**Describe the Bug**
Both functions, when called from the Velodyne HDL-32E and presumably other Lidars, return the full range image, which is not correct according to the documentation on getLayerRangeImage().

**Steps to Reproduce**
1. Place Lidar on top of a car
2. Get the data from both functions
3. Compare their lengths

**Expected behavior**
Get only the data ranging from a * resolution to a * (2 * resolution - 1) when calling getLayerRangeImage(a).

**Screenshots**

**System**
 - Graphics Card: Nvidia GTX 1080Ti

**Additional context**

stefaniapedrazzi:
Fix #2672:
overwrite default SWIG implementation of `Lidar.getLayerRangeImage` to return just part of the full image array.
This approach is the same also used for the `Lidar.getLayerPointCloud` function.

### DIFF
--- a/docs/reference/lidar.md
+++ b/docs/reference/lidar.md
@@ -547,7 +547,7 @@ from controller import Lidar
 
 class Lidar (Device):
     def getPointCloud(self, data_type='list'):
-    def getLayerPointCloud(self, layer):
+    def getLayerPointCloud(self, layer, data_type='list'):
     def getNumberOfPoints(self):
     # ...
 ```
@@ -609,7 +609,7 @@ The `wb_lidar_get_layer_point_cloud` function is a convenient way of getting dir
 
 The `wb_lidar_get_number_of_points` function returns the total number of points contained in the point cloud (each layer is assumed to have the same number of points associated to).
 
-> **Note** [Python]: The `getPointCloud` method has `data_type` parameter which can be `list` (default) or `buffer`.
+> **Note** [Python]: The `getPointCloud` and `getLayerPointCloud` methods have `data_type` parameter which can be `list` (default) or `buffer`.
 If `data_type` is equal to `list` then the function returns a list of points, but it is slow as it has to create a list of objects.
 If `data_type` is equal to `buffer` then the function returns `bytearray` and it is fast as there is no memory copying.
 
--- a/src/controller/python/controller.i
+++ b/src/controller/python/controller.i
@@ -91,7 +91,7 @@ using namespace std;
 // manage double arrays
 %typemap(out) const double * {
   int len = 3;
-  string test("$name");
+  const string test("$name");
   if (test == "getSFVec2f" || test == "getMFVec2f")
     len = 2;
   else if (test == "getSFRotation" || test == "getQuaternion")
@@ -105,7 +105,7 @@ using namespace std;
     PyList_SetItem($result, i, PyFloat_FromDouble($1[i]));
 }
 %typemap(out) const double *getLookupTable {
-  int len = arg1->getLookupTableSize()*3;
+  const int len = arg1->getLookupTableSize()*3;
   $result = PyList_New(len);
   for (int i = 0; i < len; ++i)
     PyList_SetItem($result, i, PyFloat_FromDouble($1[i]));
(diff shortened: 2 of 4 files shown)


lukicdarkoo:
A simpler solution would be replacing the `typemap` around line 538 with something like this:
```c
%typemap(out) float * {
  int width = arg1->getHorizontalResolution();
  int height = arg1->getNumberOfLayers();
  int len = width * height;
  string test("$name");
  if (test == "getLayerRangeImage")
    len = width;
  if ($1) {
    $result = PyList_New(len);
    for (int x = 0; x < len; ++x)
      PyList_SetItem($result, x, PyFloat_FromDouble($1[x]));
  } else
    $result = Py_None;
}
```

stefaniapedrazzi:
Yes this was just an initial step and I already planned to do some performance tests because also the `Lidar.getLayerPointCloud` doesn't seem the optimal one.

stefaniapedrazzi:
I implemented the suggestion of @lukicdarkoo and improved the performance of the Python `Lidar.getLayerPointCloud` (about x2 faster than before when returning a list) by reusing the same code as for `Lidar.getPointCloud`.

src/controller/python/controller.i
```diff
@@ -552,20 +555,16 @@ class AnsiCodes(object):
...
-    const webots::LidarPoint *point = $self->getPointCloud();
-    return point[index];
-  }
-
-  PyObject *__getPointCloudBuffer() const {
-    const char *points = (const char *)$self->getPointCloud();
-    const int size = $self->getNumberOfPoints() * sizeof(WbLidarPoint);
+  PyObject *__getPointCloudBuffer(int layer) const {
+    const char *points = layer < 0 ? (const char *)$self->getPointCloud() : (const char *)$self->getLayerPointCloud(layer);
+    const int number_of_points = layer < 0 ? $self->getNumberOfPoints() : $self->getHorizontalResolution();
```
lukicdarkoo, line 560:
```suggestion
    const int numberOfPoints = layer < 0 ? $self->getNumberOfPoints() : $self->getHorizontalResolution();
```
We are in C++ here, maybe we should stick with camel-case notation.

docs/reference/lidar.md
```diff
@@ -547,7 +547,7 @@ from controller import Lidar
 
 class Lidar (Device):
     def getPointCloud(self, data_type='list'):
-    def getLayerPointCloud(self, layer):
+    def getLayerPointCloud(self, layer, data_type='list'):
```
lukicdarkoo, line 550: This is a new parameter, maybe we should merge it in the `develop` branch

lukicdarkoo (changes requested):
Thank you, only two minor comments to discuss

docs/reference/lidar.md
stefaniapedrazzi: If it would be mandatory, then yes.
But it is optional, so it doesn't break the compatibility.

I can make a version without it in master and then add it in develop. But I'm not sure it is worth it.

lukicdarkoo: > But I'm not sure it is worth it.

I agree, too much hustle for a tiny impact :)

src/controller/python/controller.i
```diff
@@ -539,6 +539,9 @@ class AnsiCodes(object):
   int width = arg1->getHorizontalResolution();
   int height = arg1->getNumberOfLayers();
   int len = width * height;
+  string function_name("$name");
+  if (function_name == "getLayerRangeImage")
+    len = width;
```
lukicdarkoo, line 544:
```suggestion
  const int width = arg1->getHorizontalResolution();
  const int height = arg1->getNumberOfLayers();
  int len = width * height;
  string functionName("$name");
  if (functionName == "getLayerRangeImage")
    len = width;
```

The camel-case should be here as well

stefaniapedrazzi:
All the comments should be addressed now.

lukicdarkoo (approved):
Thank you!

PR #2801, opened 2021-03-02
Fix physics example

### LINKED ISSUE #2643: howto/physics simulation is broken
In the `howto/physics` simulation, the MyBot robot fails detecting obstacles and detect objects even when it should not.

The expected behavior is that the robot goes down just after the box stop intersecting the robot ray and messages in the console should only be written when the ray intersects a box.

I tested R2019a and it works.
At some point the color of the ray was also changing but in R2019a this already doesn't work we should inspect this as well and restore the behavior if possible or remove unneeded code.

lukicdarkoo:
Fixes #2643

Since #2505 the callback function is called for Group nodes (spaces) as well, so the ray was colliding with the Group node in the robot's body.


omichel (changes requested):
We should add a line in the change log for this bug fix as it was affecting the released version of Webots R2021a.

omichel (approved):
Thank you.

omichel (approved):

PR #2803, opened 2021-03-02
Fix URDF exporting of leaf nodes

### LINKED ISSUE #2503: URDF export, no fixed joint generation
**Describe the Bug**
Solids without a joint as a parent dont get turned into links. The only exception is the automatically generated "baselink".

**Steps to Reproduce**
1. add a robotic arm to the world (irb4600_40)
2. add a Solid node to the handslot and give it a name
3. export urdf
4. Link does not get created. (Giving it a physics node, mass, bounding object, inertia matrix, DEF name all did not change this)

**Expected behavior**
All Solid nodes should be turned into links. For each parent-child relationship of solids, a fixed joint should be created

lukicdarkoo:
This PR should fix #2503. URDF links were not generated for the leaf nodes.

Visual inspection to check for regressions:
image

image

Visualization command (may be useful in the future):
```
ros2 launch urdf_tutorial display.launch.py model:=/tmp/robot.urdf
```


omichel (approved):
Thank you.

PR #2804, opened 2021-03-02
Fix Pen density

### LINKED ISSUE #2353: Wrong color blending when pen.inkdensity is not 1
**Describe the Bug**
> when I set inkdensity to intermediate amounts, it seems to average the pen color with white, rather than with whatever the existing color of the surface is, so it looks bad on surfaces that aren't naturally near-white

This is reproducible in the `pen.wbt` sample.

The blending with the `WorldInfo.inkEvaporation` factor works fine, the inkdensity should work the same way.

**Screenshots**
pen

**Additional context**

This is happenning both with Appearance and PBRAppearance.

lukicdarkoo:
Fixes #2353. 
image

The colors are now injected on top of the background, not on top of the white background.

This PR also changes how paint and texture colors are mixed, so it affects infra-red sensors as well. The `WbShape::pickColor`  method is a mess, we should get pixel color directly from WREN if possible.

### DIFF
--- a/src/webots/nodes/WbShape.cpp
+++ b/src/webots/nodes/WbShape.cpp
@@ -393,7 +393,7 @@ void WbShape::pickColor(WbRgb &pickedColor, const WbRay &ray, double *roughness,
                        paintContribution * paintColor.red());
     pickedColor.setGreen((1 - paintContribution) * diffuseColor.green() * textureColor.green() +
                          paintContribution * paintColor.green());
-    pickedColor.setBlue((1 - paintContribution) * diffuseColor.blue() * textureColor.blue() +
+    pickedColor.setBlue((1.0f - paintContribution) * diffuseColor.blue() * textureColor.blue() +
                         paintContribution * paintColor.blue());
   }
 }
--- a/src/webots/nodes/WbShape.cpp
+++ b/src/webots/nodes/WbShape.cpp
@@ -389,7 +389,7 @@ void WbShape::pickColor(WbRgb &pickedColor, const WbRay &ray, double *roughness,
       return;  // default value
 
     // combine colors
-    pickedColor.setRed((1 - paintContribution) * diffuseColor.red() * textureColor.red() +
+    pickedColor.setRed((1.0f - paintContribution) * diffuseColor.red() * textureColor.red() +
                        paintContribution * paintColor.red());
     pickedColor.setGreen((1 - paintContribution) * diffuseColor.green() * textureColor.green() +
                          paintContribution * paintColor.green());
--- a/src/webots/nodes/WbShape.cpp
+++ b/src/webots/nodes/WbShape.cpp
@@ -391,7 +391,7 @@ void WbShape::pickColor(WbRgb &pickedColor, const WbRay &ray, double *roughness,
     // combine colors
     pickedColor.setRed((1.0f - paintContribution) * diffuseColor.red() * textureColor.red() +
                        paintContribution * paintColor.red());
-    pickedColor.setGreen((1 - paintContribution) * diffuseColor.green() * textureColor.green() +
+    pickedColor.setGreen((1.0f - paintContribution) * diffuseColor.green() * textureColor.green() +
                          paintContribution * paintColor.green());
     pickedColor.setBlue((1.0f - paintContribution) * diffuseColor.blue() * textureColor.blue() +
                         paintContribution * paintColor.blue());
--- a/src/webots/nodes/WbShape.cpp
+++ b/src/webots/nodes/WbShape.cpp
@@ -284,7 +284,6 @@ void WbShape::pickColor(WbRgb &pickedColor, const WbRay &ray, double *roughness,
   WbRgb paintColor(1.0f, 1.0f, 1.0f);
   pickedColor = WbRgb(1.0f, 1.0f, 1.0f);  // default value
   WbVector2 uv(-1, -1);
-  double paintContribution = 0;
 
   if (roughness)
     *roughness = 0.0;
@@ -292,6 +291,8 @@ void WbShape::pickColor(WbRgb &pickedColor, const WbRay &ray, double *roughness,
     *occlusion = 0.0;
 
   if (geom) {
+    double paintContribution = 0;
+
     WbPaintTexture *paintTexture = WbPaintTexture::findPaintTexture(this);
     if (paintTexture) {
       const bool success = geom->pickUVCoordinate(uv, ray, 0);
--- a/src/webots/nodes/WbShape.cpp
+++ b/src/webots/nodes/WbShape.cpp
@@ -284,6 +284,7 @@ void WbShape::pickColor(WbRgb &pickedColor, const WbRay &ray, double *roughness,
   WbRgb paintColor(1.0f, 1.0f, 1.0f);
   pickedColor = WbRgb(1.0f, 1.0f, 1.0f);  // default value
   WbVector2 uv(-1, -1);
+  double paintContribution = 0;
 
   if (roughness)
     *roughness = 0.0;
@@ -321,6 +322,7 @@ void WbShape::pickColor(WbRgb &pickedColor, const WbRay &ray, double *roughness,
         return;
       }
       // retrieve the corresponding color in the paint texture
+      paintContribution = paintTexture->pickDensity(uv);
       paintTexture->pickColor(paintColor, uv);
     }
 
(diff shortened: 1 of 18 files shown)


stefaniapedrazzi:
:eyes:

src/webots/nodes/WbShape.cpp
```diff
@@ -387,9 +390,12 @@ void WbShape::pickColor(WbRgb &pickedColor, const WbRay &ray, double *roughness,
       return;  // default value
 
     // combine colors
-    pickedColor.setRed(diffuseColor.red() * textureColor.red() * paintColor.red());
-    pickedColor.setGreen(diffuseColor.green() * textureColor.green() * paintColor.green());
-    pickedColor.setBlue(diffuseColor.blue() * textureColor.blue() * paintColor.blue());
+    pickedColor.setRed((1 - paintContribution) * diffuseColor.red() * textureColor.red() +
```
stefaniapedrazzi, line 393: I would use a float value in these instructions so that the expected type is set (sometimes this could lead to errors):
```suggestion
    pickedColor.setRed((1.0f - paintContribution) * diffuseColor.red() * textureColor.red() +
```

stefaniapedrazzi (changes requested):
Thank you! The blending is now working correctly.

I just have a couple of comment about the implementation.

stefaniapedrazzi:
If this is affecting the result of the infra-red sensor, then maybe we should merge it into develop instead of master to be sure we don't break the compatibility (and add a bold warning text in the ChangeLog).

src/webots/nodes/utils/WbPaintTexture.cpp
```diff
@@ -284,6 +286,24 @@ void WbPaintTexture::pickColor(WbRgb &pickedColor, const WbVector2 &uv) const {
...
+  while (y < 0)
+    y += h;
+  while (x >= w)
+    x -= w;
+  while (y >= h)
+    y -= h;
+
+  const int index = (y * w + x) * 4;
+  return mData[index + 3];
+}
```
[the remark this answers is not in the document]
lukicdarkoo: You are completely right, I will change it

src/webots/nodes/utils/WbPaintTexture.cpp
```diff
@@ -237,25 +237,27 @@ void WbPaintTexture::paint(const WbRay &ray, float leadSize, const WbRgb &color,
...
+        mData[dataIndex + 1] = oldDensityRatio * mData[dataIndex + 1] + (1 - oldDensityRatio) * color.green();
+        mData[dataIndex + 2] = oldDensityRatio * mData[dataIndex + 2] + (1 - oldDensityRatio) * color.red();
+        mData[dataIndex + 3] += density;
+        if (mData[dataIndex + 3] > 1)
+          mData[dataIndex + 3] = 1;
+
+        if (mEvaporation) {
+          mEvaporation[rowOffset + tx] += density;
+          if (mEvaporation[rowOffset + tx] > 1)
+            mEvaporation[rowOffset + tx] = 1;
```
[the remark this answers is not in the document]
lukicdarkoo: No, the result should be the same? I will put 1.0f to be consistent

lukicdarkoo:
> If this is affecting the result of the infra-red sensor, then maybe we should merge it into develop instead of master to be sure we don't break the compatibility (and add a bold warning text in the ChangeLog).

I don't know, it makes sense, but it is also a fix. It will affect the infra-red readings only in places where the paint is present. In a way, the previous readings were wrong (the sensors were reading the white color that is not painted).

stefaniapedrazzi:
> > If this is affecting the result of the infra-red sensor, then maybe we should merge it into develop instead of master to be sure we don't break the compatibility (and add a bold warning text in the ChangeLog).
> 
> I don't know, it makes sense, but it is also a fix. It will affect the infra-red readings only in places where the paint is present. In a way, the previous readings were wrong (the sensors were reading the white color that is not painted).

Ok. So it is fine to merge to master.

src/webots/nodes/utils/WbPaintTexture.hpp
```diff
@@ -39,7 +39,7 @@ class WbPaintTexture : public QObject {
 
   static bool isPaintable(const WbShape *shape);
 
-  void pickColor(WbRgb &pickedColor, const WbVector2 &uv) const;
+  void pickColor(WbRgb &pickedColor, const WbVector2 &uv, float *pickedDensity = NULL) const;
```
stefaniapedrazzi, line 42: It is bad to mix input and output arguments, we should also change the function so that first the input arguments are given and then the output arguments:
```suggestion
  void pickColor(const WbVector2 &uv, WbRgb &pickedColor, float *pickedDensity = NULL) const;
```

src/webots/nodes/utils/WbPaintTexture.cpp
```diff
@@ -237,25 +237,27 @@ void WbPaintTexture::paint(const WbRay &ray, float leadSize, const WbRgb &color,
...
-        mData[dataIndex] = existingColorPercentage * mData[dataIndex] + blue;
-        mData[dataIndex + 1] = existingColorPercentage * mData[dataIndex + 1] + green;
-        mData[dataIndex + 2] = existingColorPercentage * mData[dataIndex + 2] + red;
-        mData[dataIndex + 3] = 1.0f;
-
-        if (mEvaporation)
-          mEvaporation[rowOffset + tx] = 1.0f;
+        const float previousDensity = mData[dataIndex + 3];
+        const float oldDensityRatio = previousDensity / (density + previousDensity);
+        mData[dataIndex] = oldDensityRatio * mData[dataIndex] + (1 - oldDensityRatio) * color.blue();
```
stefaniapedrazzi, line 249: I would use 1.0f also here
```suggestion
        mData[dataIndex] = oldDensityRatio * mData[dataIndex] + (1.0f - oldDensityRatio) * color.blue();
```

src/webots/nodes/WbShape.cpp
```diff
@@ -387,9 +389,12 @@ void WbShape::pickColor(WbRgb &pickedColor, const WbRay &ray, double *roughness,
       return;  // default value
 
     // combine colors
-    pickedColor.setRed(diffuseColor.red() * textureColor.red() * paintColor.red());
-    pickedColor.setGreen(diffuseColor.green() * textureColor.green() * paintColor.green());
-    pickedColor.setBlue(diffuseColor.blue() * textureColor.blue() * paintColor.blue());
+    pickedColor.setRed((1.0f - paintContribution) * diffuseColor.red() * textureColor.red() +
+                       paintContribution * paintColor.red());
+    pickedColor.setGreen((1 - paintContribution) * diffuseColor.green() * textureColor.green() +
```
stefaniapedrazzi, line 394:
```suggestion
    pickedColor.setGreen((1.0f - paintContribution) * diffuseColor.green() * textureColor.green() +
```

src/webots/nodes/WbShape.cpp
```diff
@@ -387,9 +389,12 @@ void WbShape::pickColor(WbRgb &pickedColor, const WbRay &ray, double *roughness,
...
+                         paintContribution * paintColor.green());
+    pickedColor.setBlue((1 - paintContribution) * diffuseColor.blue() * textureColor.blue() +
```
stefaniapedrazzi, line 396:
```suggestion
    pickedColor.setBlue((1.0f - paintContribution) * diffuseColor.blue() * textureColor.blue() +
```

src/webots/nodes/utils/WbPaintTexture.cpp
```diff
@@ -237,25 +237,27 @@ void WbPaintTexture::paint(const WbRay &ray, float leadSize, const WbRgb &color,
...
+        mData[dataIndex + 1] = oldDensityRatio * mData[dataIndex + 1] + (1 - oldDensityRatio) * color.green();
```
stefaniapedrazzi, line 250:
```suggestion
        mData[dataIndex + 1] = oldDensityRatio * mData[dataIndex + 1] + (1.0f - oldDensityRatio) * color.green();
```

src/webots/nodes/utils/WbPaintTexture.cpp
```diff
@@ -237,25 +237,27 @@ void WbPaintTexture::paint(const WbRay &ray, float leadSize, const WbRgb &color,
...
+        mData[dataIndex + 2] = oldDensityRatio * mData[dataIndex + 2] + (1 - oldDensityRatio) * color.red();
```
stefaniapedrazzi, line 251:
```suggestion
        mData[dataIndex + 2] = oldDensityRatio * mData[dataIndex + 2] + (1.0f - oldDensityRatio) * color.red();
```

stefaniapedrazzi (changes requested):
Thank you.

I would use 1.0f for all the density computations.

src/webots/nodes/utils/WbPaintTexture.hpp
lukicdarkoo: I was hoping to avoid changing the other parts of the code. I will revert to `pickDensity()` as changing the order will affect other parts of the code and cause inconsistency. For example `WbAppearance::pickColorInTexture(WbRgb &pickedColor, const WbVector2 &uv)` will not be consistent with `WbPaintTexture::pickColor(const WbVector2 &uv, WbRgb &pickedColor, float *pickedDensity = NULL)` (argument order is reversed).

stefaniapedrazzi: I think that we should consider reversing the arguments in all the functions.
It is standard to have first input and then output arguments.

stefaniapedrazzi (approved):
Thank you!

PR #2806, opened 2021-03-02
Fix wireframe rendering

### LINKED ISSUE #1738: The wireframe mode is badly affected by lighting
# description
The wireframe rendering mode looks bad because the lighting is not disabled. Are a consequence, sometimes edges are not visible at all because they are the same color as the background.

# steps to reproduce
Open the [laser_pointer](https://github.com/cyberbotics/webots/blob/master/projects/samples/devices/worlds/laser_pointer.wbt) simulation and enable the wireframe rendering mode (F12). Some edges of the green and red boxes are not visible.

lukicdarkoo:
This PR should eventually fix #1738.

---

Notice: I am directly accessing WREN's low-level functions (e.g. `wr_shader_program_set_custom_uniform_value`) from `WbViewpoint`. Please let me know if it is a good approach or you have some kind of convention for this kind of thing.


omichel (approved):
Thank you.
It works very well and I believe the implementation is good.

PR #2812, opened 2021-03-04
Fix update of supervisor field count

### LINKED ISSUE #1519: Wrong Supervisor Field State when another Supervisor import nodes
**Describe the Bug**
When a supervisor controller imports a node in the scene tree, the number of children of the parent field (``wb_supervisor_field_get_count``) is correctly updated in the supervisor controller.
But if another supervisor controller is using this field, then the number of children is not updated in its reference of the field.

stefaniapedrazzi:
Fix #1519: notify the supervisor controllers when the size of a MF field that is cached changed so that the size in the libController can be updated.

_Changes to existing messages_
Previously the MF field count was correctly updated only for MFNode fields when a node was removed.
To avoid duplicating the update, I changed the `C_SUPERVISOR_NODE_REMOVE_NODE` message so that it just deletes any cached instance of the deleted node. But the field count will be updated using the new `C_SUPERVISOR_FIELD_COUNT_CHANGED` messsage.
Similarly, when importing a node from a supervisor, Webots returned immediately the number of imported nodes to update the cached field count. For MFNode fields this return message has been removed and replaced by the new `C_SUPERVISOR_FIELD_COUNT_CHANGED`.


stefaniapedrazzi:
I fixed the variable name collision.
:+1:

lukicdarkoo (approved):
Very nice implementation! Thank you

PR #2822, opened 2021-03-05
Significantly simplify the usage of the `ros` controller

lukicdarkoo:
This PR significantly simplifies the usage of the `ros` controller:
- Integrates `ros_control` (the `--use-ros-control` flag) to allow usage of a differential drive, joint state publisher, trajectory follower, and other controllers.
- Adds the `--auto-publish` flag that automatically enables all sensors and creates necessary topics.
- Adds the `robot_description` ROS param which contains the robot's URDF (necessary for many ROS packages).

**Impact**
Take [the Pioneer controller](https://github.com/cyberbotics/webots_ros/blob/master/src/pioneer3at.cpp) from the `webots_ros` package as an example. The controller with over 300 lines of code is not needed anymore. Furthermore, to simplify the controller, we "cheated" by calculating odometry from GPS and IMU. There is no way something like that would work in the real world. The odometry is now calculated from the wheel encoders.

**Tasks**
- [x] Proof of concept
- [x] Test with differential drive
  - [x] We need `robot_state_publisher` here. `Robot description couldn't be retrieved from param server.` 
- [ ] ~~Test with MoveIt~~ (I think it is better to finish this in another PR)
- [x] Rethink namespaces (e.g. `/robot_description`, URDF prefix...)
- [x] Document
- [x] Changelog 

**Useful commands**
```bash
rosrun controller_manager controller_manager list
```

### DIFF
--- a/projects/default/controllers/ros/highlevel/WebotsHw.cpp
+++ b/projects/default/controllers/ros/highlevel/WebotsHw.cpp
@@ -18,7 +18,9 @@
 
 namespace highlevel {
 
-  WebotsHw::WebotsHw(webots::Robot *robot) : mRobot(robot) {
+  WebotsHw::WebotsHw(webots::Robot *robot) : mRobot(NULL) {
+    mRobot = robot;
+
     // Find all motors
     const int nDevices = mRobot->getNumberOfDevices();
     for (int i = 0; i < nDevices; i++) {
--- a/projects/default/controllers/ros/highlevel/WebotsHw.cpp
+++ b/projects/default/controllers/ros/highlevel/WebotsHw.cpp
@@ -18,9 +18,7 @@
 
 namespace highlevel {
 
-  WebotsHw::WebotsHw(webots::Robot *robot) : mRobot(NULL) {
-    mRobot = robot;
-
+  WebotsHw::WebotsHw(webots::Robot *robot) : mRobot(robot) {
     // Find all motors
     const int nDevices = mRobot->getNumberOfDevices();
     for (int i = 0; i < nDevices; i++) {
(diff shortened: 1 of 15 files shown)


Simon-Steinmann:
Does the robot description contain the visual data? This was an issue with ROS2 as well like 2 month ago. The new way of storing protos, with the meshes as .obj files, should make this fairly simple.

lukicdarkoo:
> Does the robot description contain the visual data? This was an issue with ROS2 as well like 2 month ago. The new way of storing protos, with the meshes as .obj files, should make this fairly simple.

No :/ It is output from the `wb_robot_get_urdf` function. We need to improve the `wb_robot_get_urdf` function to get the visual data as well. Do you have a use case for this? We deliberately omitted this feature.

projects/default/controllers/ros/highlevel/WebotsHw.cpp
```diff
@@ -0,0 +1,99 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Description: Webots integration with for `ros_control`.
+
+#include "WebotsHw.hpp"
+
+namespace highlevel {
+
+  WebotsHw::WebotsHw(webots::Robot *robot) : mRobot(robot) {
```
omichel, line 21: I would rename this class `WebotsHW` (with a uppercase W) to match ROS control's `hardware_interface::RobotHW`

omichel (changes requested):
It looks very good.
I just have a few suggestions.

projects/default/controllers/ros/highlevel/WebotsHw.cpp
omichel: Webots convention would have led to `WebotsHardware` instead (see https://github.com/cyberbotics/webots/wiki/Coding-Style#cs44-no-abbreviations), but in that case, I believe we should adopt the convention used in the upper class, e..g., `RobotHW`.

omichel (approved):
Thank you.

PR #2824, opened 2021-03-08
Fixed Makefile in case JAVA_HOME is not set and javac is not available on Linux

omichel:
On Linux, if JAVA_HOME is not set and `javac` is not found, the following error is displayed when running the top Makefile:
```bash
readlink: missing operand
Try 'readlink --help' for more information.
```
This PR fixes this problem.


stefaniapedrazzi (approved):
Thank you

Issue #2823, opened 2021-03-08
Hinge2Joint device limits applied on wrong axis


ad-daniel:
**Describe the Bug**
In Hinge2Joint the device2 and jointParameters limits appear to be crossed.

**Steps to Reproduce**
1. Load [hinge2_bug.zip](https://github.com/cyberbotics/webots/files/6100693/hinge2_bug.zip)

2. Note that device2 `minPosition` is -1, and `maxPosition` is +1
3. Set jointParameters position to -2 and save.
4. When reloading it gives a warning that position is outside the limits despite referring to a different axis.

The opposite doesn't appear to be true, meaning that setting min/max position on `device` doesn't trigger a warning if a position out of bounds is set in `jointParameters2`.

**Expected behavior**
device should reflect jointParameters and device2 should reflect jointParameters2.

**System**
 - Operating System: Ubuntu 20.04
 - Graphics Card: AMD Radeon RX

PR #2825, opened 2021-03-08
Fix incorrect device limit constraints in Hinge2Joint and BallJoint

ad-daniel:
**Related Issues**
This pull-request fixes issue #2823.

**Description**
As mentioned in issue #2823, device position limits ( `maxPosition` and `minPosition` on `device2`) appear to be checked against `position` field of `jointParameters` instead of `jointParameters2`.

According to the ODE documentation, this is said about hinge2joint

> Axis 1 can have joint limits and a motor, axis 2 can only have a motor. 

To my understanding, joint limits refers to `minStop` and `maxStop` whereas in this particular case we're talking about _motor limits_, which I believe axis2 can have independently to axis1 and leads me to believe that issue #2823 is indeed a bug.

The bug stems from 2 factors:
- In [updateMinAndMaxPosition](https://github.com/cyberbotics/webots/blob/c0ce8250876997125c4decbeed48bf0f98e39e0b/src/webots/nodes/WbMotor.cpp#L175) no distinction between the axis is made and `jointParameters` is always picked.
- In [WbJointDevice](https://github.com/cyberbotics/webots/blob/c0ce8250876997125c4decbeed48bf0f98e39e0b/src/webots/nodes/WbJointDevice.cpp#L51-L61) the attribution of `mPositionIndex` occurs only in the `postFinalize` step whereas [updateMinAndMaxPosition](https://github.com/cyberbotics/webots/blob/c0ce8250876997125c4decbeed48bf0f98e39e0b/src/webots/nodes/WbMotor.cpp#L99) is called in the `preFinalize` one meaning that it would always default to the first axis anyway.

Technically after fixing the first bug it would've sufficed to call `updateMinAndMaxPosition` in `postFinalize` instead of `preFinalize` to achieve the same result, but the proposed solution seems more consistent with the rest of the code (to me at least).


src/webots/nodes/WbPositionSensor.cpp
```diff
@@ -56,8 +56,11 @@ void WbPositionSensor::init() {
   mResolution = findSFDouble("resolution");
 }
 
+void WbPositionSensor::preFinalize() {
+  WbJointDevice::preFinalize();
+}
+
 void WbPositionSensor::postFinalize() {
-  WbJointDevice::postFinalize();
```
stefaniapedrazzi, line 60: Independently from the other changes, this should be kept in any case because it is important to call the base class method: recursively it should always call `WbBaseNode::postFinalize` to work properly.

src/webots/nodes/WbPositionSensor.cpp
```diff
@@ -56,8 +56,11 @@ void WbPositionSensor::init() {
   mResolution = findSFDouble("resolution");
 }
 
+void WbPositionSensor::preFinalize() {
+  WbJointDevice::preFinalize();
+}
+
```
stefaniapedrazzi, line 62: This is a derived method. There is no need to override it with the exact same value as in the base class.

src/webots/nodes/WbJointDevice.cpp
```diff
@@ -48,8 +48,8 @@ void WbJointDevice::init() {
   mRobot = NULL;
 }
 
-void WbJointDevice::postFinalize() {
-  WbBaseNode::postFinalize();
+void WbJointDevice::preFinalize() {
+  WbBaseNode::preFinalize();
   // Cache position index
   const WbField *const f = parentField(true);
```
stefaniapedrazzi, line 54: Just for info: usually, all the instructions that depends on the parent field/node hierarchy are usually executed in the `postFinalize` method because of this call:
https://github.com/cyberbotics/webots/blob/c0ce8250876997125c4decbeed48bf0f98e39e0b/src/webots/nodes/WbBaseNode.cpp#L101

In this particular case, the `WbNode::parentField()` result should not make use of the field parent value.

ad-daniel: Then if mPositionIndex should be set only in the `postFinalize` it kinda defies the purpose of the rework entirely. So the second option of simply moving the call of  `updateMinAndMaxPosition` to the `WbMotor::postFinalize` to fix the issue would've been better? If not, mPositionIndex will always be 1. (unless there's a better way of knowing which axis is being concerned)

stefaniapedrazzi: It should be checked and tested, but from a first look it seems fine to move this to the `preFinalize()` method because it doesn't use the `WbField::parentNode()` method that is set after the `preFinalize` step.

I just wanted to give a context to the choice of adding this code to the `postFinalize` method.

ad-daniel: Oh I see, thanks for the info. Well as far as I can tell it seems fine, neither this nor the derived ones use it. But just to be sure, it might be better if the reviewer resolves this conversation rather than me (assuming the tests don't break in the meantime).

src/webots/nodes/WbMotor.cpp
```diff
@@ -109,7 +109,6 @@ void WbMotor::preFinalize() {
 }
 
 void WbMotor::postFinalize() {
-  WbJointDevice::postFinalize();
```
stefaniapedrazzi, line 112: This instruction should not be deleted (same reason as for WbPositionSensor)

stefaniapedrazzi:
You should also add entry in the ChangeLog.
If this bug is reproducible programmatically from a Supervisor controller, then it would also be good to test it in the test suite by modifying for example an existing API test or creating a new one.

ad-daniel:
Ok thank you, I'll look into it.

ad-daniel:
I've added a test suite. It checks that the adjustment of the `minPosition` and `maxPosition` is done correctly for all types of joint and all axis and that no criss-crossing occurs between them (jointParameters3 affecting device2 etc).

tests/parser/controllers/joints_motor_minmax_position/joints_motor_minmax_position.c
```diff
@@ -0,0 +1,128 @@
...
+
+int main(int argc, char **argv) {
+  ts_setup(argv[0]);
+
+  // This test suite verifies that if the position field, as set in the jointParameters, is outside of the bounds of the
+  // minPosition and maxPosition of RotationalMotor, these bounds must adadpt accordingly.
+  // This suite also checks that the correct pairing is done (device <-> jointParameters, device2 <-> jointParameters2)
+
+  // SliderJoint Tests -------------------------------------------------------------------------------------------------
+  WbDeviceTag sj_motor_lower_lim = wb_robot_get_device("sj_lower_lim_device");
```
omichel, line 18: [No abbreviations](https://github.com/cyberbotics/webots/wiki/Coding-Style#cs44-no-abbreviations): I would replace "sj" with "slider_joint" or "slider" and "lim" with "limit".
I would also replace "device" with "motor" to be more precise.
This comment also applies to other names below.

omichel (changes requested):

tests/parser/controllers/joints_motor_min_max_position/joints_motor_min_max_position.c
```diff
@@ -0,0 +1,140 @@
...
+
+int main(int argc, char **argv) {
+  ts_setup(argv[0]);
+  ts_disable_output_log();
+  // This test suite verifies that if the position field, as set in the jointParameters, is outside of the bounds of the
+  // minPosition and maxPosition of RotationalMotor, these bounds must adadpt accordingly.
+  // This suite also checks that the correct pairing is done (device <-> jointParameters, device2 <-> jointParameters2)
+
+  // SliderJoint Tests -------------------------------------------------------------------------------------------------
+  WbDeviceTag slider_joint_motor_lower_limit = wb_robot_get_device("slider_joint_lower_limit_motor");
```
omichel, line 18: It looks better here to match the variable name with the device name, isn't it?
```suggestion
  WbDeviceTag slider_joint_lower_limit_motor = wb_robot_get_device("slider_joint_lower_limit_motor");
```

tests/parser/controllers/joints_motor_min_max_position/joints_motor_min_max_position.c
```diff
@@ -0,0 +1,140 @@
...
+
+  // for motor: [minPosition, maxPosition] was [-0.1, 0.1] in world, should be [-0.2, 0.1] after loading
+  ts_assert_double_equal(wb_motor_get_min_position(slider_joint_motor_lower_limit), -0.2,
+                         "'minPosition' of SliderJoint (for motor) not adjusted correctly when position below lower limit.");
+  ts_assert_double_equal(
+    wb_motor_get_max_position(slider_joint_motor_lower_limit), 0.1,
+    "'maxPosition' of SliderJoint (for motor) changed but shouldn't have (testing slider_joint_motor_lower).");
+
+  WbDeviceTag slider_joint_motor_upper_limit = wb_robot_get_device("slider_joint_upper_limit_motor");
```
omichel, line 27:
```suggestion
  WbDeviceTag slider_joint_upper_limit_motor = wb_robot_get_device("slider_joint_upper_limit_motor");
```
etc.

omichel (changes requested):
Thank you, it is much better.
I still have one final request.

tests/parser/controllers/joints_motor_min_max_position/joints_motor_min_max_position.c
ad-daniel: It does, thank you. Also I noticed the messages for the upper limit tests were swapped (correct test, wrong message) hence why there's more changes in  the last commit.

omichel: Thank you. However, I am not very happy with the new names as I believe a device name should end with the device type, e.g., "my special distance sensor" or "left motor". Hence I would prefer its ends with "motor" in this case rather than having the "motor" word in the middle of the name, which looks odd to me.

omichel (approved):
Thank you.

omichel:
I see now that the test was done in the parser folder.
I believe it should rather go to the api folder instead because we are testing the API here.
@ad-daniel: can you open a new PR to fix this?

ad-daniel:
mmh, I don't think any API is being tested, as it's the decoding of the world/proto that was wrong (i.e the `minPosition` / `maxPosition` were adjusted incorrectly when loading) and is also how the test works which is why parser seemed more appropriate, no?

PR #2827, opened 2021-03-09
Fix blog index

omichel:
We currently miss the last post in the index: https://cyberbotics.com/doc/blog/index
This PR fixes it.


lukicdarkoo (approved):
Thank you

PR #2830, opened 2021-03-09
Fixed doc generation script

omichel:
This PR fixes the `generate_objects_doc.py` script:

- make it work on Windows (with correct linefeed and UTF-8 conversion).
- remove the comma in lists with two elements (incorrect in English).
- regenerate the doc pages with the new script.
- improved the image and thumbnail for the FifaSoccerBall proto.

See result at: https://www.cyberbotics.com/doc/guide/object-balls?version=fix-doc-generation-script


docs/guide/generate_objects_doc.py
```diff
@@ -62,7 +62,7 @@
 fileList = sorted(fileList)
 
 # create the 'appearances' page
-with open('appearances.md', 'w', encoding='utf-8') as file:
+with open('appearances.md', 'w', encoding='utf-8', newline='\n') as file:
```
lukicdarkoo, line 65: > Lines in the input can end in '\n', '\r', or '\r\n', and these are translated into '\n' before being returned to the caller.

https://docs.python.org/3/library/functions.html#open

Why does the default value for the new line makes a problem on Windows. Furthermore, on Windows, the new line should be detonated with `\r\n`?

omichel: On Windows, the default is `\r\n` which doesn't match with the files committed which were generated under Linux. For the committed file, we use the Linux standard, hence the `newline='\n'` enforcement for Windows.

lukicdarkoo (approved):
Thank you

PR #2831, opened 2021-03-09
Artificial turf physics

omichel:
This PR aims a developing a realistic model for the physics of artificial turf.
It also includes damping for the ball and changing the controllers of the goal keepers so that they remain in their goal (and can be used to test turf physics by applying forces on them).


lukicdarkoo:
If you leave `softERP` to the default value (0.2) and slightly change `softCFM` to like 0.03 you get "soft" ground.

https://user-images.githubusercontent.com/2135826/110932516-e7c1c400-832b-11eb-9b64-087b1b318eb3.mp4

See the legs in the video. On the grass, the legs sink a bit as I change the center of mass, while it is not the case on the floor.

lukicdarkoo:
What about exposing the `softCFM` and `DEPTH` fields (maybe under different names) instead of `turfPhysics`. It would be more configurable.

omichel:
> What do you think about exposing the `softCFM` and `DEPTH` fields (maybe under different names) instead of `turfPhysics`? It would be more configurable.

I prefer to leave it like that because these parameters are tricky to fine tune, I believe it's better to edit the proto if we want to fine tune them. And also the `turfPhysics` is an easy switch to enable/disable this extra layer and quickly see how the performance and behavior are affected.

lukicdarkoo (dismissed):
Thank you!

(I tried merging locally with the `develop` branch and it was working, so there should not be problems)

lukicdarkoo:
But there is conflict in `kid.wbt`...

omichel:
OK, the conflict should be fixed now.

lukicdarkoo:
> quickly see how the performance and behavior are affected.

I don't see any performance difference. Should it be?

omichel:
> I don't see any performance difference. Should it be?

I could see a tiny 5% performance drop with turfPhysics TRUE, but it's not very clear...

lukicdarkoo (approved):

PR #2836, opened 2021-03-10
Fix test-sources on Windows

omichel:
It seems that chocolately is unable to install cppcheck version 2.2, trying 2.3...


omichel:
That was a temporary problem which seems to be fixed now.

PR #2839, opened 2021-03-10
Add RUB to FLU PROTO conversion script

lukicdarkoo:
To avoid duplication in https://github.com/cyberbotics/webots_ros/pull/60, we convert to FLU here (worlds to ENU)

---

This PR adds the initial version of the PROTO conversion script and it is briefly tested on Pioneer robots. Therefore, improvements will be needed in the future.


lukicdarkoo:
@cyberbotics/maintainers To convert the coordinates, I simply added an upper transform for the robot's children (rotation/translation of the bounding objects had to be converted one by one). This is an easy solution, but not clean. To convert it properly we need a script to handle the `IndexedFaceSet` and `HingeJoint` nodes. Should we leave it as is, or create a script to convert each child independently.

lukicdarkoo:
> in default Webots world (for example in new worlds) the robot appears wrongly rotated.

To me, it looks fine, see default rotation:
image

If you add a FLU robot to the NUE world then it is messed up, but NUE is not the default coordinate system. Maybe we can automatically change the `rotation` parameter once it is added to the world to match ENU/NUE?

> Personally I think we should convert them properly and create a script for this.

Sounds good, I will start making a script

lukicdarkoo:
> > If you add a FLU robot to the NUE world then it is messed up, but NUE is not the default coordinate system
> 
> NUE is the still the default coordinate system in Webots as it is used in new worlds (File > New World).

You are right, but it is the only green thing in the scene tree :)

omichel:
I believe we should add one more design recommendation for all proto nodes: the altitude (e.g., Z coordinate) should be by default 0 and that should correspond to the robot lying on a floor plane at altitude 0.
Also, the default translation/rotation fields should always be:
```js
translation 0 0 0
rotation    0 0 1 0
```

lukicdarkoo:
> I believe we should add one more design recommendation for all proto nodes: the altitude (e.g., Z coordinate) should be by default 0 and that should correspond to the robot lying on a floor plane at altitude 0.
> Also, the default translation/rotation fields should always be:
> 
> ```js
> translation 0 0 0
> rotation    0 0 1 0
> ```

Added a `--z-offset` flag to specify z offset.

omichel (approved):
Thank you.

PR #2840, opened 2021-03-10
Move joints min max position test suite into API folder

ad-daniel:
As mentioned [here](https://github.com/cyberbotics/webots/pull/2825#issuecomment-795200788), `API` folder is more suited for this test instead of `parser`.


lukicdarkoo:
I included the `test suite` flag since you were changing the tests

omichel (approved):
Thank you.

PR #2841, opened 2021-03-10
Fix background in ENU

lukicdarkoo:
The background images in ENU worlds were replaced. It works fine in the `master` branch, so it is only broken in the `develop` branch.


omichel:
cppcheck reported some errors.

omichel (approved):
Thank you!

PR #2845, opened 2021-03-11
Fix ROS api compilation

stefaniapedrazzi:
This is an attempt to fix the ROS API compilation that fails when creating the snap package due to `isnan` not declared.


stefaniapedrazzi:
From this branch the ROS API is now correctly compiled in the snapcraft environment:
https://github.com/cyberbotics/webots-snap/pull/31/checks?check_run_id=2084243404

omichel (approved):
Thank you.

PR #2854, opened 2021-03-15
Use WebGL to draw plots in the generic robot window

stefaniapedrazzi:
Use WebGL to draw the plot instead of drawing directly in the canvas.

Steps:
* [x] add instructions to draw the graph using WebGL  
* [x] create one GL context per canvas
* [x] reduce the number of GL context
* [x] fix wrong callback links

For simple pages with just few plots, there is no difference between using webgl or drawing the graphs directly. But for pages with many graphs (for example the motors tab of the PR2 or Spot robot) the webgl version is running even 3 times faster.


stefaniapedrazzi:
I tested on Linux and I'm now updating my development environment to also test on Windows but it will take some time.

resources/projects/plugins/robot_windows/generic/canvas.js
```diff
@@ -0,0 +1,105 @@
...
+    this.canvasWidth = 0;
+    this.canvasHeight = 0;
+  }
+  return Canvas._instance;
+}
+
+Canvas.prototype.getWebglContext = function() {
+  if (!this.webglContext) {
+    var webglNames = ['webgl', 'experimental-webgl'];
+    for (var i = 0; i < webglNames.length; ++i) {
```
omichel, line 19: Shouldn't we use `let` instead of `var` everywhere?

omichel (commented):
I could test it on Windows and it seems to work.
However, the rescaling of the position sensor graphs is still confusing: previous values disappear suddenly, it looks buggy... I will make a movie to show you this.

omichel:
If you look at the second graph, instead of being rescaled, the graph appears like if it was erased. This happens twice and is confusing. I believe that with WebGL, we could have a much smoother rescaling.

https://user-images.githubusercontent.com/1264964/111502491-2a2d3b80-8746-11eb-9581-8f2da24caebe.mp4

stefaniapedrazzi:
Thank you for the tests.

This is the JUMP rule defined in the generic robot window. The PositionSensor and motors are the only device using this range update method and I think it is the correct way to update the range given that they can continuously growing and it would not make sense to stretch the range.
I will try to improve the used formula so that a bigger portion of the graph is visible and increase the range size.
But we should not stretch the range and extend it because it will produce unusable graphs (same as the encoders for the vehicle robot window).

By the way it seems that there are some issues drawing the white background of the graph on Windows.
Here is what I get on Linux
webgl_plot

omichel:
> This is the JUMP rule defined in the generic robot window. The PositionSensor and motors are the only device using this range update method and I think it is the correct way to update the range given that they can continuously growing and it would not make sense to stretch the range.
I will try to improve the used formula so that a bigger portion of the graph is visible and increase the range size.
But we should not stretch the range and extend it because it will produce unusable graphs (same as the encoders for the vehicle robot window).

I believe that all the Y values in the displayed range should be visible on the graph when we rescale it, which is not the case currently.

stefaniapedrazzi:
> I believe that all the Y values in the displayed range should be visible on the graph when we rescale it, which is not the case currently.

Yes, this can be improved so that _most_ of the Y values are visible.
But the range size should be always kept the same it cannot be stretched so that all the Y values are visible and made bigger.
I updated a version that work much better but stretching the range size for position sensors, motors, and ecoders is very bad and should be avoided.

stefaniapedrazzi:
I added a third range update method that makes sure that all the Y values are visible.
It is currently set as default for motors and position sensors.
The drawback is that it is not as efficient as the others given that every time it needs to loop through the Y values to adjust the range (stretch it when Y is out of range and shrink it when old Y values are discarded).
I expected a much bigger impact on the performance but for simple graphs I could not see any difference in the speedometer.

resources/projects/plugins/robot_windows/generic/timeplot_widget.js
```diff
@@ -364,15 +371,48 @@ TimeplotWidget.prototype.computePlotRect = function() {
...
+      if (max == null || y[j] > max)
+        max = y[j];
+    }
+  }
+  const delta = Math.abs(this.initialYRange['max'] - this.initialYRange['min']);
+  let halfOffset = 0;
+  if (Math.abs(max - min) < delta) {
+    // Minimum size of range is defined by the initial values.
+    halfOffset = (delta - max + min) * 0.5;
+  } else if (y > this.yRange['max'])
```
omichel, line 407:
```suggestion
  if (Math.abs(max - min) < delta)
    halfOffset = (delta - max + min) * 0.5; // Minimum size of range is defined by the initial values.
  else if (y > this.yRange['max'])
```

omichel (changes requested):
Yes, the ADAPT method looks much better than the previous JUMP method.
Thank you.

omichel (approved):

PR #2865, opened 2021-03-16
Fixed balls protos with orientation and position rules

### LINKED ISSUE #2828: What's the best default translation for a ball object?
_Originally posted by @stefaniapedrazzi in https://github.com/cyberbotics/webots/pull/2783#discussion_r589554149_

> Usually the PROTO nodes we provide are by default added above the ground.
With this change the ball will be half below the ground.
Is this really wanted? Because version seems better and this is the only ball from the `objects/balls` section that is not located above the default ground position.

I am opening this issue to discuss this problem, before creating a PR to fix it.

If we want to fix it in a similar way as other ball proto, we should set a default value for the Y coordinate of the ball, equals to the radius of the ball. This has the following drawbacks:
- with ENU coordinate system, it simply doesn't work: we should set Z to the radius.
- if we change the size of the ball, we need to adjust the Y/Z value accordingly, which is not obvious in case of the FIFA ball because it's size (1-5) doesn't give directly the information on the radius.

The idea of setting the default translation value to 0 0 0 fixes the problem with ENU/NUE coordinate systems, but you will have to adjust the height of the ball so that it doesn't appear half in the ground.

Unlike other objects, a ball doesn't have a upside, so offsetting the center of the ball with its radius will not be good after the ball a rolled a little bit: its coordinates (translation field) will become confusing, given its center may be upside down or in between.

I could not find a good solution to this problem and I believe having 0 0 0 as the default position of a ball is something simple and understandable even though it requires to manually offset it from the ground when creating it out of the blue, which seems acceptable.

So, unless we find a better solution, I will fix all the balls we have in Webots to set them at 0 0 0 default position, like the RobocupSoccerBall and the FifaSoccerBall.

omichel:
Fixes #2828.

This PR fixes the default center and orientation for ball protos.
In addition, it fixes the icon for the RobocupSoccerBall proto which was missing from the doc and not nice from the proto icon.
The object doc was regenerated from the script, see https://cyberbotics.com/doc/guide/object-balls?version=fix-ball-protos-with-orientation-and-position-rules


stefaniapedrazzi:
:eyes:

projects/objects/balls/protos/Ball.proto
```diff
@@ -5,8 +5,8 @@
 # You can change the centerOfMass to [0 0 0] to avoid the random rolling effect.
 
 PROTO Ball [
-  field SFVec3f    translation     0 0.0325 0
-  field SFRotation rotation        0 1 0 0
+  field SFVec3f    translation     0 0 0
+  field SFRotation rotation        0 0 1 0
```
stefaniapedrazzi, line 9: Solid default rotation value is `0 1 0 0`.
Is there a particular reason to set a different default rotation for balls?

stefaniapedrazzi (commented):
Everything is fine. I just have one question about the balls default rotation.

projects/objects/balls/protos/Ball.proto
omichel: It's not very important, but means that the balls are now compatible with the [new default ENU orientation recommendations](https://cyberbotics.com/doc/reference/proto-design-guidelines#orientation).

omichel: Webots will soon parse this field to determine if a PROTO has to be rotated when inserted from the GUI in a non-ENU world, but for balls, it doesn't make any sense to rotate them 😉.

stefaniapedrazzi: Again, I think we should be consistent with default rotation value.
For me it is ok to set this here only if we will also change the `Solid` default rotation in R2021b, otherwise it doesn't make sense.

stefaniapedrazzi (approved):

projects/objects/balls/protos/Ball.proto
stefaniapedrazzi: Note that `0 1 0 0` and `0 0 1 0` are both empty rotations. It is just a matter of consistency between default values in the models distributed within Webots.

omichel: Yes, the plan is also to change the default rotation in Solid as well, but this is out of the scope of this PR.

PR #2870, opened 2021-03-17
Change initial position of the Tiago arm

lukicdarkoo:
The initial position of the `ARM_1_JOINT` joint is equal to the `minPosition`. In some scenarios, this throws out-of-bounds errors as the initial position is very close to `minPosition`.


omichel (approved):
Thank you.

PR #2875, opened 2021-03-17
Improve URDF naming convention

lukicdarkoo:
This PR solves the problem of inconsistent link/joint names as a suffix for non-unique names is generated from `WbNode::uniqueId()` (identifies nodes within a whole world -> if the world changes the IDs change as well).

In addition:
- a name generation is much faster as there is no recursive check and
- DEF name is also added as a fallback option (if node name is not defined).


omichel (approved):
The resulting URDF files are indeed way nicer!
Note: this PR is still in Draft mode.

lukicdarkoo:
> The resulting URDF files are indeed way nicer!
> Note: this PR is still in Draft mode.

I forgot to add it to the changelog, will be ready in a sec

omichel (dismissed):

omichel (approved):

PR #2876, opened 2021-03-17
Fix Tiago damping

lukicdarkoo:
Tiago wobbles too much when move, so mapping fails. This PR increases the `suspensionSpringConstant` to make the robot more sturdy.


omichel (approved):

omichel:
We should probably add a line in the change log about it as it may affect previous simulations.

omichel (approved):
Thank you.

Issue #2843, opened 2021-03-10
Question about Charger implementation


stefaniapedrazzi:
The Charger energy is set to 0 if the detected robot current charge is bigger than then maximum charge.
So it means that 
Is this the correct behavior?

Because it doesn't seem this the normal behavior of charger.
Even if not stored in the Robot battery, the energy is taken from the charger but I don't see why the whole energy of the Charger is set to 0.

Usually, the battery charging consumption has a logarithmic curve and not a linear one, but even considering this I cannot understand why the whole charger energy is taken.

https://github.com/cyberbotics/webots/blob/master/src/webots/nodes/WbCharger.cpp#L201-L218

omichel:
We should actually not reset the currentEnergy of the charger to 0.

PR #2879, opened 2021-03-17
Improve Charger energy update when robot battery is full

stefaniapedrazzi:
Address #2843: remove reset of Charger energy to 0 if the robot battery is full.


omichel (approved):
Thank you.

Issue #1428, opened 2020-03-04
Remove 'global' Variables from Python Code


DavidMansolino:
**Describe the Bug**
Some of our python script are usign the `global` keyword, this is known as a bad practice (see for example: https://docs.quantifiedcode.com/python-anti-patterns/maintainability/using_the_global_statement.html) and can be avoided in 99% of the cases (personally I have never used it and never felt the need to use it):
  - projects/languages/ros/controllers/ros_python/ros_controller.py
  - resources/web/server/session_server.py
  - resources/web/server/simulation_server.py
  - tests/test_suite.py

ShuffleWire:
I've just look that up, and there is 3 usage of the global variables : (for instance in resources/web/server/session_server.py)
 - Reference to "constants" like LOAD_THRESHOLD : using `global` is useless, as global variable/constants could be read from anywhere. moreover sometimes LOAD_THRESHOLD is used without being declared `global` before (which is fine, but not homogeneous with the rest of the code). Is the intend to emphasis the fact that we are using "global" values ?
 - Reference to object that are being updated, but the reference of the object stay the same : for instance "config" is only updated, hence, as long as we don't change the reference ie `config = ...`, it's fine to do stuff like `config[...] = ...` without needing the `global` keyword
 - Variable needing the `global` keyword, as `availability`, modified in `update_load`. Is that a symptom of bad function definition / usage / architecture ? Should we update it ?

To generalize, are there some unspoken rules about it in the Webot code. Do you thing it should be relevant to spend a bit of time to rearrange the code to make those global disappear ? I should be able to do so.
The main question is also about useless global statement to **read** global variable : removing those `global` statement will hide the fact that there are global stuff running around. Is that fine ? TBF, it's common in python scripting... And if we don't mess with it (ie don't blindly read/write), it should be fine.

omichel:
This is not an important issue and there is a risk to introduce bugs while modifying this code.
So, I would prefer that you contribute something else.

ShuffleWire:
That's fine, but just for the record, in R2022a, here is the list of python script using the keyword `global`
scripts/converter/convert_geometries.py (1)
projects/robots/mobsya/thymio/controllers/thymio2_aseba/aseba/maintainer/updatedoc/wikidot/structure.py (5)
projects/robots/mobsya/thymio/controllers/thymio2_aseba/aseba/maintainer/translations/translation_tools.py (2)
projects/default/resources/sumo/tools/traci/main.py (1)
[...]

PR #2880, opened 2021-03-17
Global variables fix in session_server.py

bkpcoding:
**Description**
Just tried to create a class for global variables and so as to eliminate the use of globals in the code, although left some of them like(LOAD_THRESHOLD, LOAD_REFRESH), which are mainly constants and therefore won't introduce bugs. This is not the full implementation of the code, there are somethings that I don't know like tornado, thus don't know whether adding arguments to the SessionHandler, MonitorHandler, ClientWebSocketHandler 's __init__ will affect the working of the code, any help would be appreciated. 
I Will take feedbacks positively, even if the changes are not necessary then I can focus on other bugs else if I am going in the right direction then maybe will try to make it more robust and try to implement the same for other files as well where global is used.

**Related Issues**
This pull-request fixes issue #1428

### DIFF
--- a/docs/guide/starting-webots.md
+++ b/docs/guide/starting-webots.md
@@ -44,6 +44,9 @@ Options:
     Choose the startup mode, overriding application preferences. The <mode>
     argument must be either pause, realtime or fast.
 
+  --no-rendering
+    Disable rendering in the main 3D view.
+
   --fullscreen
     Start Webots in fullscreen.
 
--- a/docs/reference/node-chart.md
+++ b/docs/reference/node-chart.md
@@ -5,7 +5,7 @@ The Webots Node Chart outlines all the nodes available to build Webots worlds.
 ## Inheritance
 
 In the chart, an arrow between two nodes represents an inheritance relationship.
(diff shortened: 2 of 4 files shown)


bkpcoding:
I don't know why it's showing my previous commits also, but the main commit that this PR is for the session_server.py file

omichel:
I am not convinced this PR is a progress with respect to the original code.
It adds the complexity of one extra class to get rid of two global variables, which seems overkill.

bkpcoding:
Yeah, it did seem to me that it kind of overcomplicates the things and I was also not able to put in config global variable to it, so just something I tried out to remove global variables. Anyways thank you for your suggestion, maybe will try and fix other bugs or if you have idea about removing these global variables without overcomplicating like adding additional class then I will be more than happy to implement.

omichel:
In case you are wondering which issue to address, I believe this one could be very useful to us: #2884. We would be happy to review your PR about it.

bkpcoding:
Sure, @omichel I will try my best to get the PR ;)

PR #2881, opened 2021-03-17
Darwin-op expose backlash

ad-daniel:
**Description**
This PR exposes a `backlash` field and activates backlash on all joints (amount: 0.04) for the `HingeJoint` based Darwin-OP (the default one).

It should be noted that:

-  The `HingeJointWithBacklash` PROTO already has a mechanism that reverts to a simple `HingeJoint` in case backlash is 0. So in theory there should be no performance difference between the original Darwin and the modified version if backlash is enforced to 0 (condition for it to revert). In practice however there appears to be some overhead in the latter case which makes it more computationally complex, which is why in this PR the switch is done directly in the Darwin PROTO despite this mechanism existing in the `HingeJointWithBacklash`.

**Performance metrics**
Scenario:

- 8 robots walking straight
- timestep 8 [ms]
- run on a (old) laptop (Geforce GT 730M). (Threads = 1, shadows disables, AA disabled)
- 2000 steps

World metrics:
|                            | solids | joints | geometries |
|----------------------------|--------|--------|------------|
| HingeJ (backlash = false)  | 265    | 160    | 305        |
| HingeJWB (backlash = 0)    | 265    | 160    | 305        |
| HingeJWB (backlash = 0.04) | 425    | 320    | 305        |

Performance:
|                            | prePhysics(ms) | physics(ms)     | postPhysics(ms) | mainRendering(ms) |
|----------------------------|----------------|-----------------|-----------------|-------------------|
| HingeJ (backlash = false)  | 655 [0.328]    | 31588 [15.794]  | 6000 [3.000]    | 10048 [5.024]     |
| HingeJWB (backlash = 0)    | 741 [0.371]    | 31766 [15.883]  | 7212 [3.606]    | 9615 [4.796]      |
| HingeJWB (backlash = 0.04) | 1193 [0.597]   | 180843 [90.422] | 13639 [6.820]   | 8609 [4.294]      |

It was obtained using the `--log-performance` argument. The values represent: TOTAL [AVG].


projects/robots/robotis/darwin-op/protos/Darwin-op.proto
```diff
@@ -29,6 +29,7 @@ PROTO Darwin-op [
   field SFNode     jersey          NULL                       # Extends the robot with a jersey: typically RobotisJersey.proto.
   field SFInt32    channel         0                          # Defines the channel for the `Emitter` and `Receiver` devices.
   field SFBool     showWindow      FALSE                      # Is `Robot.showWindow`.
+  field SFBool     enableBacklash  FALSE                      # Enables backlash on all joints.
```
omichel, line 32: I would rename this field to simply "backlash".

projects/robots/robotis/darwin-op/protos/Darwin-op.proto
```diff
@@ -42,6 +43,19 @@ PROTO Darwin-op [
 {
   %{
     local has_jersey = fields.jersey.value
+    local enable_backlash = fields.enableBacklash.value
+    -- tune backlash on joints
+    local backlash_head = 0.04
```
omichel, line 48: I believe we should put here the value taken from in the datasheet of the Darwin-OP servo. Is that correct?
Also, for convenience, we could have single backlash value, like:
```lua
local backlash = 0.04
local backlash_head = backlash
local backlash_neck = backlash
...
```
That would make it easier to adjust.

omichel (changes requested):
Thank you.
We should also update the user guide (ROBOTIS OP2 page).

ad-daniel:
I've added the documentation regarding the new field and since in the other PR it was mentioned that the default rotation should be `0 0 1 0` I changed this one as well.

docs/guide/robotis-op2.md
```diff
@@ -183,10 +186,20 @@ sudo apt-get install espeak
 
 For more information on the use of all of these sensors/actuators refer to the [Reference Manual](http://www.cyberbotics.com/reference) of Webots.
 
+*Self Collision*
+
 The physical model is very realistic and self collision check is available.
 To activate the self collision expand the ROBOTIS OP2 node in the scene tree and set `selfCollision` field to `true`.
 Use the self collision check only if you need it, because it is an expensive computation feature and can therefore significantly slow down the simulation speed.
 
+*Backlash*
```
omichel, line 195:
```suggestion
#### Backlash
```

docs/guide/robotis-op2.md
```diff
@@ -183,10 +186,20 @@ sudo apt-get install espeak
 
 For more information on the use of all of these sensors/actuators refer to the [Reference Manual](http://www.cyberbotics.com/reference) of Webots.
 
+*Self Collision*
```
omichel, line 189:
```suggestion
#### Self Collision
```

docs/guide/robotis-op2.md
```diff
@@ -183,10 +186,20 @@ sudo apt-get install espeak
...
 The physical model is very realistic and self collision check is available.
 To activate the self collision expand the ROBOTIS OP2 node in the scene tree and set `selfCollision` field to `true`.
 Use the self collision check only if you need it, because it is an expensive computation feature and can therefore significantly slow down the simulation speed.
 
+*Backlash*
+
+It is possible to model the effect of backlash present in the motors by setting the `backlash` field to `true`.
+Doing so will add a backlash of 0.01 [rad] to every joint in the robot.
+The modeling of this effect is achieved by replacing the existing joints with [HingeJointWithBacklash](hinge-joint-with-backlash.md) which effectively doubles their number.
+Additionally, for the correct enforcing of this effect, the value of the `basicTimestep` in [WorldInfo](../reference/worldinfo.md) should be set to 8 or lower.
```
omichel, line 200:
```suggestion
Additionally, for the correct enforcing of this effect, the value of the `basicTimestep` in [WorldInfo](../reference/worldinfo.md) should probably be set to a smaller value when enabling the backlash.
```

docs/guide/robotis-op2.md
```diff
@@ -183,10 +186,20 @@ sudo apt-get install espeak
...
+For these reasons, activating this option is very computationally expensive and will significantly slow down the simulation speed.
```
omichel, line 201:
```suggestion
For these reasons, activating this option is computationally expensive and will significantly slow down the simulation speed.
```

docs/guide/robotis-op2.md
```diff
@@ -899,6 +912,10 @@ If you are starting the controller 'manually' by SSH, use the -X option when sta
 export DISPLAY=:0 ./my-controller
 ```
 
+#### Activating the Backlash Flag Renders the Robot Unable to Stand
+
+Due to the small amount of backlash present in the motors, for correct modeling of this effect it is required to lower the `basicTimestep` in [WorldInfo](../reference/worldinfo.md) to 8 or less.
```
omichel, line 917:
```suggestion
Due to the small amount of backlash present in the motors, for correct modeling of this effect it is required to lower the `basicTimestep` in [WorldInfo](../reference/worldinfo.md). It was observed that setting the `basicTimeStep` to 8 ms gives good results.
```

omichel (changes requested):

omichel (approved):
Thank you.

PR #2882, opened 2021-03-18
Retry in case of server error

stefaniapedrazzi:
In the last days the upload of some packages failed.
In this PR I'm trying to better handle the random server errors (500, 502 and 504).

This script already have some custom retry functionality for uploading assets but it only considers the connections errors and not server errors (`requests.exceptions.HTTPError`).
However another solution would be to set the retry rules directly in the Github module.
We could keep them both and check if it ever occurs that the custom retry is called, if not we can remove this later.

### DIFF
--- a/scripts/packaging/publish_release.py
+++ b/scripts/packaging/publish_release.py
@@ -126,9 +126,9 @@ for release in repo.get_releases():
                             release.upload_asset(path)
                             break
                         except requests.exceptions.ConnectionError:
-                            print('Release upload failed due to connection error (remaining trials: %d)' % remainingTrials)
+                            print('Release upload failed due to connection error (remaining trials: %d)' % (4 - retryCount))
                         except requests.exceptions.HTTPError:
-                            print('Release upload failed due to server error (remaining trials: %d)' % remainingTrials)
+                            print('Release upload failed due to server error (remaining trials: %d)' % (4 - retryCount))
                         retryCount += 1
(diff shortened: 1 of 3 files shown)


scripts/packaging/publish_release.py
```diff
@@ -120,7 +124,10 @@
                             remainingTrials = 0
                         except requests.exceptions.ConnectionError:
                             remainingTrials -= 1
-                            print('Release upload failed (remaining trials: %d)' % remainingTrials)
+                            print('Release upload failed due to connection error (remaining trials: %d)' % remainingTrials)
+                        except requests.exceptions.HTTPError:
+                            remainingTrials -= 1
+                            print('Release upload failed due to server error (remaining trials: %d)' % remainingTrials)
                     if (releaseExists and tagName.startswith('nightly_') and not releaseCommentModified and
```
lukicdarkoo, line 131: I think we can afford a short delay here, for example, 1 minute, then 2 minutes, then 3 minutes... This way we will be more robust GitHub failures - if a server is unavailable it is likely it will be unavailable in 2-3 seconds as well.

stefaniapedrazzi: I will add a timeout.
Note that i expect that this custom retry is never called because handled before by the Github library internal retry.

stefaniapedrazzi:
I applied the changes and it is now ready for review.

omichel (approved):
It looks good 🤞🏻

PR #2890, opened 2021-03-21
Fix incorrect stopCFM and stopERP imposition when world uses non-default global CFM/ERP

ad-daniel:
**Description**
The current version loads user-defined `stopERP` and `stopCFM` correctly and also handles invalid values. However, if the world being loaded uses non-default values for global `ERP` and `CFM` (i.e not 0.2 and 1e-5), then in [this](https://github.com/cyberbotics/webots/blob/e80fb802139d4645eed30f9865017a742607c185/src/webots/nodes/WbHingeJoint.cpp#L142) line the `stopERP` and `stopCFM` actually being imposed on the joint are the default ones defined in `HingeJointParameters` [here](https://github.com/cyberbotics/webots/blob/e80fb802139d4645eed30f9865017a742607c185/resources/nodes/HingeJointParameters.wrl#L15) which could be fine as long as the global ones are also at default. If they aren't however, it doesn't respect it and imposes 0.2 and 1e-5 anyway. In other words, currently the global values are imposed to the stops if and only if `jointParameters` isn't defined which isn't enough.

The proposed solution uses `-1` to represent that the global values must be enforced.

### DIFF
--- a/docs/reference/hingejointparameters.md
+++ b/docs/reference/hingejointparameters.md
@@ -9,8 +9,8 @@ HingeJointParameters {
   SFFloat suspensionSpringConstant  0       # [0, inf)
   SFFloat suspensionDampingConstant 0       # [0, inf)
   SFVec3f suspensionAxis            1 0 0   # unit axis
-  SFFloat stopERP                   0.2     # [0, inf)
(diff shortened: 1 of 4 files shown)


omichel (approved):
Well spotted.
Thank you.

PR #2894, opened 2021-03-22
Update c3d robot window

stefaniapedrazzi:
Follow-up #2854 and update the c3d robot window and fix some remaining bugs:
* 3D plot labels not updated if the timeplot is not selected
* PlotWidget and RadarWidget brackground not correctly set to white
* Timeplot graphs redrawn even if nothing changed

### DIFF
--- a/projects/humans/c3d/plugins/robot_windows/c3d_viewer_window/c3d_viewer_window.css
+++ b/projects/humans/c3d/plugins/robot_windows/c3d_viewer_window/c3d_viewer_window.css
@@ -88,6 +88,9 @@
   height: 200px;
   border: 1px solid darkgrey;
 }
+.plot-canvas-background {
+  background-color: white;
+}
 .plot-axis-label {
   font-size: 0.8em;
   position: absolute;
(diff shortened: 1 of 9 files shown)


omichel (commented):
On Windows, when we change the graph from "Time" to "XZ", the background becomes grey:

image

stefaniapedrazzi:
The PlotWidget (showing the XY, YZ and XZ plots) is still drawing the points directly on the canvas.
I checked but in this case using canvas instead of WebGL seems to be faster in most of the use cases because the resize/blit of the plot usually don't occur so frequently and only new points need to be drawn instead of the whole graph.

I'm now comparing the performance of the RadarWidget using WebGL.

By the way, I checked but it doesn't seem possible to share a WebGL context between multiple canvas.

stefaniapedrazzi:
Also in case of the RadarWidget there is no visible performance improvement when drawing with WebGL but it is easier to draw the radar frustum using the cavas functions.

omichel (approved):
Thank you.

PR #2897, opened 2021-03-23
Fix randomBuilding proto

ad-daniel:
**Description**
As discussed in #2893 (specifically [here](https://github.com/cyberbotics/webots/pull/2893#discussion_r599545236)) the `RandomBuilding` proto is supposed to generate (potentially) different buildings for every instance. As such, neither the change introduced in #2893, nor the prior state of the proto, results in the desired behavior.


omichel (approved):
Very good, thank you.

PR #2902, opened 2021-03-24
Improve publish_release.py by catching all the expections

stefaniapedrazzi:
Add additional exception handling to catch all of them.

### DIFF
--- a/scripts/packaging/publish_release.py
+++ b/scripts/packaging/publish_release.py
@@ -132,8 +132,9 @@ for release in repo.get_releases():
                             print('Release upload failed due to server error (remaining trials: %d)' % (4 - retryCount))
                         except GithubException:
                             print('Release upload failed due to GitHub error (remaining trials: %d): %s' % (4 - retryCount))
-                        except:
+                        except Exception as e:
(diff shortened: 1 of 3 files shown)


scripts/packaging/publish_release.py
```diff
@@ -129,6 +130,10 @@
                             print('Release upload failed due to connection error (remaining trials: %d)' % (4 - retryCount))
                         except requests.exceptions.HTTPError:
                             print('Release upload failed due to server error (remaining trials: %d)' % (4 - retryCount))
+                        except GithubException:
+                            print('Release upload failed due to GitHub error (remaining trials: %d): %s' % (4 - retryCount))
+                        except:
+                            print('Release upload failed due to unexpected error (remaining trials: %d)' % (4 - retryCount))
```
lukicdarkoo, line 136:
```suggestion
                        except Exception as e:
                            print('Release upload failed due to unexpected error (remaining trials: %d)' % (4 - retryCount))
                            print('Error message', e)
```
What about printing the exception? Just to know what actually happened in case we end up in the general exception.

lukicdarkoo (approved):
Thank you

PR #2904, opened 2021-03-24
Fixed documentation about getting device tags

omichel:
The Python API should now rely on the generic `getDevice` method.


docs/reference/robot.md
```diff
@@ -588,16 +588,16 @@ Devices are available through their services.
 
 *get a unique identifier to a device*
 
-The `wb_robot_get_device` function (available in C and MATLAB) returns a unique identifier for a device corresponding to a specified `name`.
+The `wb_robot_get_device` function (available in C, Python and MATLAB) returns a unique identifier for a device corresponding to a specified `name`.
 For example, if a robot contains a [DistanceSensor](distancesensor.md) node whose `name` field is "ds1", the function will return the unique identifier of that device.
 This `WbDeviceTag` identifier will be used subsequently for enabling, sending commands to, or reading data from this device.
-If the specified device is not found, the function returns 0.
+If the specified device is not found, the function returns 0 in C and MATLAB or `none` in Python.
```
lukicdarkoo, line 594:
```suggestion
If the specified device is not found, the function returns 0 in C and MATLAB or `None` in Python.
```
It is more consistent with the language

lukicdarkoo (approved):
Thank you

Issue #2898, opened 2021-03-23
The neck of the Darwin-opHinge2 seems broken


omichel:
When playing with the Darwin-opHinge2 model, I found a problem with the neck:
![kid1](https://user-images.githubusercontent.com/1264964/112182135-4f65f200-8bfd-11eb-9a49-f3568b70c5cf.png)

PR #2909, opened 2021-03-24
Rework of hinge2withbacklash sequence

ad-daniel:
**Description**
Issue described in #2898 is a consequence of using a Hinge2Joint for the neck as both the bracket and the head are now in the `endPoint` of the joint they will both move around the anchor. So for the head to behave as it should (which is what matters given the sensors), the bracket will not look right.

To keep the performance gains of the hinge2joint will retaining a visual behavior that is correct for the neck piece, it required reworking the `Hinge2JointWithBacklash` proto by altering the sequence of the joints. From Hinge2Joint -> HingeJoint -> HingeJoint to HingeJoint -> Hinge2Joint -> HingeJoint and using the `startPoint` field for the neck bracket.  

Neck behavior after the change:

https://user-images.githubusercontent.com/44834743/112602860-806c3f80-8e14-11eb-9a5c-5421731638f4.mp4

https://user-images.githubusercontent.com/44834743/112602863-819d6c80-8e14-11eb-8458-3d7bbacddfe4.mp4


omichel (approved):
Thank you.
I will let you apply the Group trick for the startPoint to the HingeWithBacklash as well.

PR #2911, opened 2021-03-26
Remove conditional around startpoint for HingeJointWithBacklash

ad-daniel:
**Description**
The conditional here, as is, doesn't actually work as `startPoint` is actually a MF and can't be checked against `nil`. Irrespective of that, having the conditional to begin with is undesirable as if this block doesn't exist when loading a world, users can't add anything to the `startPoint` since no compatible nodes show up in the interface.


omichel (approved):
Thank you.

PR #2912, opened 2021-03-26
Calibrate the Tiago robot and simulation

### LINKED ISSUE #2908: Tiago arm is missing the `arm_tool_joint` joint
There should be the `arm_tool_joint` after `arm_7_joint`:
https://github.com/pal-robotics/tiago_ikfast/blob/3f1c038d48fd864a41f7d56a01a2f3c2acb49cd8/only_arm.urdf.clean#L333-L337

As the result, the generated URDF is not fully compatible with the official description.

Furthermore, we should also add the `Slot` node in the gripper.

### LINKED ISSUE #2910: Add a `base_footprint` link to Tiago Base
We are missing the `base_footprint` link in Tiago Base. We need it, so the exported URDF corresponds to the [REP 120](https://www.ros.org/reps/rep-0120.html#base-footprint).

**Related problem**
Currently, the robot origin is around 8cm above the ground level. This corresponds to the REP 120, but to [our convention](https://cyberbotics.com/doc/reference/proto-design-guidelines#translation-and-rotation). Note that the exported `base_link` is created out of the Robot node.

lukicdarkoo:
This PR introduces the following improvements:
1. Fixes #2910
image
2. Decreases `softCFM` so the robot sinks less into the ground

### DIFF
--- a/projects/robots/pal_robotics/tiago_base/protos/TiagoBase.proto
+++ b/projects/robots/pal_robotics/tiago_base/protos/TiagoBase.proto
@@ -46,6 +46,10 @@ PROTO TiagoBase [
       }
       Gyro {
       }
+      Solid {
+        name "base_footprint"
+        translation 0 0 -0.095
+      }
       DEF BASE_MOBILE_SHAPE Group {
         children [
           DEF BASE Shape {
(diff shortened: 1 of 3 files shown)


omichel (approved):
Thank you.

PR #2915, opened 2021-03-27
RoboCup controllers: correction of halftimeStartingPose in config files

medrimonia:
In the files provided initially, the initial pose of the robot at start of half time were invalid because they were inside the field.


omichel (approved):

omichel:
Thank you!

PR #2916, opened 2021-03-27
Lidar endpoint for remote control api

RTrioux:
**Description**
I'm trying to add support for lidar in remote control API such that we can get and visualize real time point cloud from lidar.

**Current Issue**
I'm currently facing some problems concerning this feature as discussed on your discord.
The tests I have made so far shows that the internal data container of the lidar remains null (c->image->data).
The memory can't be **allocated** in the remote plugin like this as suggested:

```
const unsigned char *lidar_data_from_robot = malloc(...);
wbr_lidar_set_image(camera->tag(), lidar_data_from_robot);
```
because:

```
wbr_lidar_set_image(WbDeviceTag t, const float *image) => wbr_abstract_camera_set_image(WbDevice *d, const unsigned char *image)

void wbr_abstract_camera_set_image(WbDevice *d, const unsigned char *image) {
  AbstractCamera *c = d->pdata;
  if (c && c->image->data)
    memcpy(c->image->data, image, 4 * c->height * c->width);
}
```

See if "c->image->data" has not been allocated before, calling wbr_lidar_set_image is pointless.
I don't see where this `c->image->data` is supposed to be allocated at least in remote control mode.

**Suggestion**

I'm quiet sure that's not the correct way of doing because it's way too obvious but why not allocating the memory directly while the abstract camera is being instantiated like this ? After all the image's size is known at compilation time.

```
void wb_abstract_camera_new(WbDevice *d, unsigned int id, int w, int h, double fov, double camnear, bool spherical) {
  wb_abstract_camera_cleanup(d);
  AbstractCamera *c = malloc(sizeof(AbstractCamera));
  c->enable = false;
  c->unique_id = id;
  c->width = w;
  c->height = h;
  c->fov = fov;
  c->camnear = camnear;
  c->spherical = spherical;
  c->sampling_period = 0;
  c->image = image_new();
  c->image->data = (unsigned char*) calloc(w*h,4);

  d->pdata = c;
}
```

by doing this I'm now able to read my sensor data in webots with `wb_lidar_get_range_image` **but** the point cloud is not 
rendered in remote control mode, only in simulation mode.
Thanks in advance for considering this.

[Question-remote-api.pdf](https://github.com/cyberbotics/webots/files/6216224/Question-remote-api.pdf)

### DIFF
--- a/include/controller/c/webots/remote_control.h
+++ b/include/controller/c/webots/remote_control.h
@@ -60,6 +60,7 @@ void wbr_display_save_image(WbDeviceTag tag, int id, int width, int height, unsi
 void wbr_camera_set_image(WbDeviceTag tag, const unsigned char *image);
 unsigned char *wbr_camera_get_image_buffer(WbDeviceTag tag);
 
+void wbr_lidar_set_image(WbDeviceTag tag, const float *image);
 // Actuator functions (values written by the controller)
 typedef struct WbrInterface {
   // mandatory functions :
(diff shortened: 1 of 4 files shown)


omichel:
It's been a long time since this PR was open.
@RTrioux: are you still interested to work on this or did you find another solution?
I will close it for now. Feel free to re-open it if needed.

PR #2923, opened 2021-03-31
Fix CameraRecognitionObject API in Python and Java

### LINKED ISSUE #2922: CameraRecognitionObject.size is only accessible as an integer in Python
**Describe the Bug**
In Python, the function `CameraRecognitionObject.get_size` returns an array of integers instead of an array of floats. There seems to be a typo in https://github.com/cyberbotics/webots/blob/8f1b58f42565dd926a23bac441f27e999d835da0/src/controller/python/controller.i#L234-L240
where the `double` attribute `size` is converted to an `int` using `PyInt_FromLong`.

**Steps to Reproduce**
1. Open camera recognition sample world `webots/projects/samples/devices/worlds/camera_recognition.wbt`
2. Create a new Python controller:
```python
from controller import Robot


robot = Robot()

camera = robot.getDevice("camera")
camera.recognitionEnable(timestep)

    objects = camera.getRecognitionObjects()
    print("Size:", objects[0].get_size())

```
3. Set the controller of the robot to the newly created one and run the simulation.
4. "Size: [0, 0]" is printed instead of the real size.

**Expected behavior**
The floating size should be accessible in Python.

stefaniapedrazzi:
Fix #2922: wrong return type for `CameraRecognitionObject.get_size` in Python API.
The Java API also has some issues.

Tasks:
* [x] fix `CameraRecognitionObject.get_size` return type in Python API
* [x] rename `Camera.getCameraRecognitionObjects` to `Camera.getRecognitionObjects` in Java API (name used in the documentation and in the Python API)
* [x] fix return type of `Camera.getPositionOnImage` and `Camera.getSizeOnImage` in Java API (SWIGTYPE_p_int instead of int[])
* [x] improve Camera Recognition Object section level to pass the docs tests.
       This is better than changing the test_references.py code because in general we want to skip h5 anchors but in this case.

### DIFF
--- a/docs/reference/camera.md
+++ b/docs/reference/camera.md
@@ -1258,7 +1258,7 @@ Sample code showing how to use the returned segmentation image object in the dif
 The `wb_camera_recognition_save_segmentation_image` function allows the user to save the latest segmentation image.
 Further details about the arguments and the return value can be found in the description of the [`wb_camera_save_image`](#wb_camera_save_image) function.
 
-##### Camera Recognition Object
+### Camera Recognition Object
 
 A camera recognition object is defined by the following structure:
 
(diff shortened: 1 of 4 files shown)


omichel (approved):
Thank you!

PR #2927, opened 2021-03-31
Fix performance disparity when using nested PROTO

ad-daniel:
**How to reproduce**

Suppose you have a PROTO that looks like this.
```
#VRML_SIM R2021b utf8
PROTO MainProtoSimple [
  field SFBool     condition         FALSE
]
{
  %{
    local condition = fields.condition.value
  }%
  Robot {
    children [
      %{ if condition then }%     <----
        MyHingeJoint { 
      %{ else }%                  <----
        HingeJoint {
      %{ end }%
          jointParameters ....
          device [ ... ]
          endPoint Solid { ... }
    ]
  }
}
```
where the `MyHingeJoint` PROTO is a straightforward `HingeJoint`
```
PROTO MyHingeJoint [
  field SFNode                                 jointParameters   NULL
  field MFNode                                 device            [ ]
  field SFNode                                 endPoint          NULL
]
{
  HingeJoint {
    jointParameters IS jointParameters
    device IS device
    endPoint IS endPoint
  }
}
```
Whether the condition is set to `TRUE` or `FALSE`, the PROTO yields the same result. However, when running the performance logger a clear disparity is present in the two situations, the nested version being slower. For reference, here's the result obtained on a similar scenario (it has 5 nested layers instead of 1).
condition FALSE:
```
<mode> <stepsCount> <prePhysics(ms)> <physics(ms)> <postPhysics(ms)> <mainRendering(ms)> 
TOT            5000            9.119       223.801            50.618              12.670  
```
condition TRUE:
```
<mode> <stepsCount> <prePhysics(ms)> <physics(ms)> <postPhysics(ms)> <mainRendering(ms)> 
TOT            5000           11.293       240.647           114.320              12.788
```
There is some variability to the results, but what remains consistently worse is the `postPhysics` step, which entails doing the `postPhysicsStep` for all the `mSolidChildren` and `mJointChildren` of all the top solids. If needed, the here's a ZIP of the tests [nested_proto_bug.zip](https://github.com/cyberbotics/webots/files/6237298/nested_proto_bug.zip)


The reason for this disparity is because in the nested case, a lot of overhead is added by the emission of signals that  propagate changes in a node field back to PROTO parameter.  In reality, the same overhead exists in the `prePhysicsStep`, but it only manifests in kinematic simulations, so the problem could be worse still.

The current solution only addresses part of the difference, dropping to around:
```
<mode> <stepsCount> <prePhysics(ms)> <physics(ms)> <postPhysics(ms)> <mainRendering(ms)> 
TOT            5000           10.273       222.292            64.199              12.607
```

**Culprits**
A) `WbRobot::postPhysicsStep > mJointChildren[i].postPhysicsStep > setPositionFromOde`.
- condition FALSE: only executes the steps with `(#)`
- condition TRUE: full chain
```
|HingeJoint (EACH POST_PHYSICS_STEP)
|                                                            |WbSFDouble                  |WbField                                |WbValue                                               |WbField                                  |WbAnimationCommand
|p = JointParameters                                         |                            |                                       |                                                      |                                         |
|p->setPositionFromOde ----- mPosition->setValueFromOde() -- | -- emit changedByOde() --- | -- [SLT] WbField::fieldChangedByOde --|-- WbValue::emitChangeByOde ... emit changedByOde() --|-- [SIG] WbField::valueChangedByOde ( -- | -- [SLT] WbAnimationCommand::updateValue )
|                             |            (#)               |              (#)           |                                       |                                                      |                                         |
|                             |_ WbSFDouble                  |                            |                                       |                                                      |                                         |
```


B) `WbRobot::postPhysicsStep > mSolidChildren[i].postPhysicsStep > applyPhysicsTransform > mLinearVelocity/mAngularVelocity->setValue`.
- condition FALSE:  only executes the steps with `(#)`
- condition TRUE: full chain
```
|Solid (EACH POST_PHYSICS_STEP)
|                              |WbSFVector3            |WbSolid
|                              |                       |
|mLinearVelocity->setValue() --|-- emit changed() -----|-- updateIsLinearVelocityNull()
|    |                                  (#)         :  |              (#)
|    |_ WbSFVector3                                 :
|                                                   :  |WbField                                                                                                                     |WbSFVector3                                                 |WbNode
|                                                   :  |                                                                                                                            |                                                            |
|                                                   -- | -- [SIG] WbValue::changed -- [SIG] WbField::valueChanged ------- [SLT] WbField::fieldChanged ... WbField::copyValueFrom -- |-- mValue::copyFrom  ... operator= ---- emit changed() ---- | -- [SLT] WbNode::notifyFieldChanged ... emit fieldChanged(field) --<|
|                                                   :                                                                                                                                                                                         :  |
|                                                   :  |WbNode                                                                                                                                                                                :
|                                                   :  |                                                                                                                                                                                      :  |WbField                                                           |WbSFVector3
|                                                   -- | -- [SLT] WbNode::notifyFieldChanged ... emit fieldChanged(field) --<|                                                                                                                :  |                                                                  |
|                                                                     (#)                                                                                                                                                                     -- | -- [SLT] WbField::parameterChanged ... WbField::copyValueFrom -- |-- mValue::copyFrom  ... operator= --<|
|                                                                                                                                                                                                                                             :
|                                                                                                                                                                                                                                             :  |WbSolid
|                                                                                                                                                                                                                                             :  |
|                                                                                                                                                                                                                                             -- | -- updateIsLinearVelocityNull
|                               |WbSFVector3           |WbSolid
|                               |                      |
|mAngularVelocity->setValue() --|-- emit changed() ----|-- updateIsAngularVelocityNull()
|    |                                 (#)          :  |              (#)
|    |_ WbSFVector3                                 :
|                                                   :  |WbField                                                                                                                     |WbSFVector3                                                 |WbNode
|                                                   :  |                                                                                                                            |                                                            |
|                                                   -- | -- [SIG] WbValue::changed -- [SIG] WbField::valueChanged ------- [SLT] WbField::fieldChanged ... WbField::copyValueFrom -- |-- mValue::copyFrom  ... operator= ---- emit changed() ---- | -- [SLT] WbNode::notifyFieldChanged ... emit fieldChanged(field) --<|
|                                                   :                                                                                                                                                                                         :  |
|                                                   :  |WbNode                                                                                                                                                                                :
|                                                   :  |                                                                                                                                                                                      :  |WbField                                                           |WbSFVector3
|                                                   -- | -- [SLT] WbNode::notifyFieldChanged ... emit fieldChanged(field) --<|                                                                                                                :  |                                                                  |
|                                                                     (#)                                                                                                                                                                     -- | -- [SLT] WbField::parameterChanged ... WbField::copyValueFrom -- |-- mValue::copyFrom  ... operator= --<|
|                                                                                                                                                                                                                                             :
|                                                                                                                                                                                                                                             :  |WbSolid
|                                                                                                                                                                                                                                             :  |
|                                                                                                                                                                                                                                             -- | -- updateIsAngularVelocityNull
```


C) `WbRobot::postPhysicsStep > mSolidChildren[i].postPhysicsStep > applyPhysicsTransform > setTransformFromOde`.
- condition FALSE: only executes the steps with `(#)`
- condition TRUE: possibly full chain
```
|Solid (EACH POST_PHYSICS_STEP)
|                      |WbAbstractTransform                                            |WbField                                 |WbSFVector3
|                      |                                                               |                                        |
|setTransformFromOde --|--- mTranslation->setValueFromOde ... emit changedByOde() ---- | -- [SLT] WbField::fieldChangedByOde .. | .... mValue::copyFrom  ... operator= ---- emit changed() --<|
|                      | :      |                                  (#)              :  |                                        |  :
|                      | :      |_ WbSFVector3                                      :                                           |  :
|                      | :                                                          :                                           |  ... mValue::emitChangedByOde .... emit changedByOde() --<|
|                      | :                                                          :                                           |
|                      | :                                                          :
|                      | :                                                          :  |WbField                           |WbAnimationCommand
|                      | :                                                          :  |                                  |
|                      | :                                                          -- | --[SIG] WbField::valueChanged -- | -- [SLT] WbAnimationCommand::updateValue
|                      | :                                                             |                                  |
|                      | :
|                      | :
|                      | :                                                            |WbField                                 |WbSFRotation
|                      | :                                                            |                                        |
|                      | --- mRotation->setValueFromOde ... emit changedByOde() ----- | -- [SLT] WbField::fieldChangedByOde .. | .... mValue::copyFrom  ... operator= ---- emit changed() --<|
|                      |       |                                   (#)             :  |                                        |  :
|                      |       |_ WbSFRotation                                     :                                           |  :
|                      |                                                           :                                           |  ... mValue::emitChangedByOde .... emit changedByOde() --<|
|                      |                                                           :                                           |
|                      |                                                           :
|                      |                                                           :  |WbField                           |WbAnimationCommand
|                                                                                  :  |                                  |
|                                                                                  -- | --[SIG] WbField::valueChanged -- | -- [SLT] WbAnimationCommand::updateValue
|                                                                                     |                                  |
```

As mentioned, if the simulation is run in kinematic mode, we'd have similar issues in the `prePhysicsStep` as well. This is it's chain:

D) `WbRobot::prePhysicsStep > mJointChildren[i].prePhysicsStep > setPosition`.
- condition FALSE: only executes the steps with `(#)`
- condition TRUE: full chain

```
|HingeJoint (EACH PRE_PHYSICS_STEP)
|                                              |WbSFDouble             |WbField
|p = JointParameters                           |                       |
|p->setPosition ----- mPosition->setValue() -- | -- emit changed() --- | -- [SIG] WbValue::changed -- [SIG] WbField::valueChanged ------- [SLT] WbField::fieldChanged
|                      |                       |         (#)        :  |              (#)                                           :
|                      |_ WbSFDouble           |                    :                                                               :
|                                              |                    :                                                               ----- [SLT] WbField::parameterChanged
|                                                                   :                                                               :
|                                                                   :                                                               : |WbNode
|                                                                   :                                                               : |
|                                                                   :                                                               --|-- [SLT] WbNode::notifyFieldChanged ... emit fieldChanged(field) --<|
|                                                                   :                                                                 |
|                                                                   :
|                                                                   :  |WbJointParameters          |WbJoint                    |WbAbstractTransform                 |WbSFVector3            |WbField
|                                                                   :  |                           |                           |                                    |                       |
|                                                                   -- |-- [SIG] positionChanged --|-- [SLT] updatePosition -- | -- setTranslationAndRotation  ---- | -- emit changed() --- | -- [SIG] WbValue::changed -- [SIG] WbField::valueChanged ----- [SLT] WbField::fieldChanged
|                                                                      |            (#)            |          (#)              |                                 :  |                       |                                                            :
|                                                                      |                           |                           |                                 :                                                                                       :
|                                                                                                                                                                :                                                                                       --- [SLT] WbField::parameterChanged
|                                                                                                                                                                :                                                                                       :
|                                                                                                                                                                :                                                                                       : |WbNode
|                                                                                                                                                                :                                                                                       : |
|                                                                                                                                                                :                                                                                       --|-- [SLT] WbNode::notifyFieldChanged ... emit fieldChanged(field) --<|
|                                                                                                                                                                :                                                                                         |
|                                                                                                                                                                :
|                                                                                                                                                                :  |WbSFRotation           |WbField
|                                                                                                                                                                :  |                       |
|                                                                                                                                                                -- | -- emit changed() --- | -- [SIG] WbValue::changed -- [SIG] WbField::valueChanged ----- [SLT] WbField::fieldChanged
|                                                                                                                                                                   |                       |                                                            :
|                                                                                                                                                                                                                                                        :
|                                                                                                                                                                                                                                                        --- [SLT] WbField::parameterChanged
|                                                                                                                                                                                                                                                        :
|                                                                                                                                                                                                                                                        : |WbNode
|                                                                                                                                                                                                                                                        : |
|                                                                                                                                                                                                                                                        --|-- [SLT] WbNode::notifyFieldChanged ... emit fieldChanged(field) --<|
|                                                                                                                                                                                                                                                          |
```
As can be seen, there's a lot of overhead that isn't otherwise there in the `condition = FALSE` case, and it should be noted that all this happens every timestep.

The current solution only addresses (A) and (B):
- This alone accounts for a large part of the overhead
- The tests run fine
- Saving the world mid-execution, the hidden parameters `linearVelocity` and `angularVelocity` are still saved correctly despite not notifying the PROTO parameters about it.
- Since `WbField::valueChanged` is no longer reached, in theory I expected it to break AnimationRecording but from the interface it seems like I can still record just fine.

This solution however feels a little hacky and doesn't address (C) and (D). These last two bullet points also raise a bigger question. The fact is, if the `condition = FALSE` case works both in terms of AnimationRecording and mid-execution save without going through all this craziness about notifying the PROTO parameters about changes to the fields, I don't see why the `condition = TRUE` needs to.

A cleaner solution might involve preventing this propagation of the change to begin with, but every attempt I've made ends up breaking a lot of testes. The main issue I think is that somehow changes to the `translation` and `rotation` fields must be notified, and is why (C) and (D) currently remain unaddressed, as mid-way through their chain they involve this notification. But once again, the `condition = FALSE` doesn't go through any of it either, so there has to be a way to make it work.


stefaniapedrazzi:
Other than the tests there are a couple of other functionalities that should also be tested to make sure that nothing is broken:
* animation recording (as you already checked)
* update of field values in the scene tree (and field editor)
* saving and reloading a world after the simulation started (especially for the linear and angular velocities)

Here are some initial notes about the resulting behavior:
* animation recording only exports the translation/rotation of the `endPoint Solid` and not the joint position (or solid angular/linear velocity) and it is notified directly when the `endPoint Solid` translation/rotation changes.
* animation recording is based on the finalized (with physics and graphical appearance) nodes and not the PROTO parameters or internal nested PROTO nodes, that is why it works event if the field changes are not propagated (changes are applied directly on these finalized nodes).
* `linearVelocity` and `angularVelocity` are hidden fields that doesn't need to be propagated to the PROTO node instance or parameter because they are used only by the current Solid node instance during the Webots execution. These values are stored in the WBT file when saving the world, the solid index is also stored, so I don't think that any other node needs to be notified about changes in velocities.

(Nested) PROTO field changes propagation is very important in many cases.
The only way to go is to analyze and compare the workflow for each instruction to understand what is needed, what can be skipped, and where does the update come from when removing the field change notification.
So I would first continue with this approach and we will see only at the end if there is a more generic approach to skip the field change propagation.

stefaniapedrazzi:
Just as side node: I quickly inspected the update of the scene tree and it seems that since some years the scene tree update is trigged regularly based on the refresh timer independently if the field changed signal is received or not.
https://github.com/cyberbotics/webots/blob/ab8b93ff30523825a8193327db756f44f12d390c/src/webots/scene_tree/WbSceneTree.cpp#L1200-L1204

So currently the field changed signals doesn't seem necessary to notify the scene tree that something changed.
Not sure which approach is better, but if the scene tree update is not the bottleneck we could think about removing all the field change notification in the WbTreeItem class.

ad-daniel:
Upon further inspection, it turns out that all the overhead mentioned here is a consequence of the way the node structure is built. Since the `MyHingeJoint` being used here exports SFNode/MFNode parameters, during the parsing, full nodes are created for each of these and once the internal structure of the PROTO is decoded they are linked to the external ones. This link is the source of the overhead since the two sides need to be notified of eventual changes (be it changes from the scene tree or like in this case from ODE side). In order to avoid this overhead therefore it's necessary to find a way of not having to rely on this link.

A possible way of fixing this is that in a first pass the node structure is allowed to be built as it currently does and when that's done and the finalized version is available, we go through the structure a second time and simplify/collapse these links. The reason it cannot be done during the parsing (or at least not easily) is that potentially these exposed SFNodes/MFNodes could be in their own way be other nested PROTO or just as complicated structures.

stefaniapedrazzi (changes requested):
I just checked the structure of the change and I have some comments on the class and folders chosen to add the new code and the test.

The approach seems fine, but I didn't check the implementation or test it because if there are major changes to apply it should be retested again later.

ad-daniel:
The PR is ready for review. If you prefer I remove the commented out debug functions let me know.

stefaniapedrazzi:
> The PR is ready for review. If you prefer I remove the commented out debug functions let me know.

It is fine to keep them for now. They can be removed later when the PR is approved.

src/webots/vrml/WbNode.hpp
```diff
@@ -276,8 +286,8 @@ class WbNode : public QObject {
   virtual const QString &stateId() const { return mCurrentStateId; };
 
   // debug utility functions
-  // void printDebugNodeStructure(int level = 0);
-  // void printDebugNodeFields(int level, bool printParameters);
+  void printDebugNodeStructure(int level = 0);
+  void printDebugNodeFields(int level, bool printParameters);
```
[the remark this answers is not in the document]
ad-daniel: Yes, I also modified part of the function body which I've now restored.

src/webots/vrml/WbNode.cpp
```diff
@@ -14,6 +14,7 @@
 
 #include "WbNode.hpp"
 
+#include "../nodes/utils/WbNodeUtilities.hpp"
```
stefaniapedrazzi, line 17: Unfortunately this is not good because it will cause a circular dependency between `nodes/utils` and `vrml` module.
Sorry but I did see this issue when suggesting to move the functionality.

Other alternatives would be to remove this dependency by implementing the functionality to search for hidden fields in WbNode class or to move the `removeInvisibleProtoNodes` to the WbBaseNode class that is in the `nodes` module.


Note that if the `removeInvisibleProtoNodes` functionality reamains in WbNode class, then you should also check the visibility of some functions (for example `WbNode::disconnectInternalNode`, `WbNode::removeFromFieldsOrParameters`, etc.) that can be made private.

stefaniapedrazzi:
I just started testing the functionality and I won't probably complete the tests today.
The approach seems the good one, but should be deeply tested.

Here is a first crash I got (I checked that on the develop branch it works):
1) open the `city.wbt` simulation
2) set the `BmwX5.interior` field to False and then back to True
```
0x00005555558ee4f8 in WbSFNode::value (this=0x0) at vrml/WbSFNode.hpp:40
40	  WbNode *value() const { return mValue; }
(gdb) bt
#0  0x00005555558ee4f8 in WbSFNode::value() const (this=0x0)
    at vrml/WbSFNode.hpp:40
#1  0x0000555555a0f2e7 in WbJoint::parameters() const (this=0x55559d7e3910)
    at nodes/WbJoint.cpp:171
#2  0x0000555555a0f10f in WbJoint::updateParameters() (this=0x55559d7e3910)
    at nodes/WbJoint.cpp:156
#3  0x00005555559f6274 in WbHingeJoint::updateParameters() (this=0x55559d7e3910) at nodes/WbHingeJoint.cpp:406
#4  0x000055555594d7b9 in WbBasicJoint::preFinalize() (this=0x55559d7e3910)
    at nodes/WbBasicJoint.cpp:94
```

ad-daniel:
You're right, I didn't even notice the circular dependency, no problem I'll figure out a way without having to rely on `NodeUtilities`, it does make more sense to have the function in `WbNode`.

As for the crash / review, how do you reckon is better to handle it? Should I wait for you and perhaps somebody else to review the code implementation itself first (after I fix the circular dependency of course) just to make sure there's no glaring errors there (or things I should've done but didn't) and only when that looks good I focus afterwards on finding why this specific world crashes? I think it would be pointless for me to try and fix that crash if there are fundamental problems in the implementation itself (and would waste your time too). What do you suggest?

stefaniapedrazzi:
I would first fix all the crashes and bugs because if there are issues it means that there are problems in the implementation and should be adjusted.

src/webots/vrml/WbNode.cpp
```diff
@@ -1881,6 +1903,264 @@ bool WbNode::isProtoParameterNode() const {
...
+  for (int i = 0; i < invisibleProtoParameterNodes.size(); ++i) {
+    WbNode *parameterNode = invisibleProtoParameterNodes[i];
+    WbNode *parent = parameterNode->parentNode();
+
+    const QVector<WbField *> fields = parent->fields();
+    for (int j = 0; j < fields.size(); ++j) {
+      WbSFNode *sfnode = dynamic_cast<WbSFNode *>(fields[j]->value());
+      WbMFNode *mfnode = dynamic_cast<WbMFNode *>(fields[j]->value());
+      if (sfnode && sfnode->value() == parameterNode) {
+        sfnode->setValueNoSignal(NULL);
```
stefaniapedrazzi, line 2134: Note that it is not necessary to duplicate the method so that it doesn't emit the signal, but you can simply use:
```
sfnode->blockSignals(true);
sfnode->setValue(NULL);
sfnode->blockSignals(false);
```

The methods that don't emit signals are needed when called indirectly by slots where it is not possible to block the signals.

stefaniapedrazzi:
I tested another world and I got a maybe different crashe:

in `project/robots/robotis/darwin-op/worlds/soccer.wbt`, change the `jerseyTexture` or `jerseyColor` field of the `RobotisOp2.jersey` node.

stack:
```
0x00005555558ee4f8 in WbSFNode::value (this=0x0) at vrml/WbSFNode.hpp:40
40	  WbNode *value() const { return mValue; }
(gdb) bt
#0  0x00005555558ee4f8 in WbSFNode::value() const (this=0x0)
    at vrml/WbSFNode.hpp:40
#1  0x00005555559e1515 in WbShape::geometry() const (this=0x55555d05ff70)
    at nodes/WbShape.hpp:55
#2  0x0000555555c4943d in (anonymous namespace)::isAllowedToInsert(QString const&, QString const&, WbNode const*, QString&, WbNode::NodeUse, QString const&, bool, bool) (fieldName=..., nodeName=..., node=0x55555d05ff70, errorMessage=..., nodeUse=WbNode::STRUCTURE_USE, type=..., automaticBoundingObjectCheck=true, areSlotAllowed=true) at nodes/utils/WbNodeUtilities.cpp:309
#3  0x0000555555c4f5f2 in WbNodeUtilities::validateExistingChildNode(WbField const*, WbNode const*, WbNode const*, bool, QString&) (field=0x55556a6a01e0, childNode=0x5555610dbeb0, node=0x55555d05ff70, isInBoundingObject=false, errorMessage=...) at nodes/utils/WbNodeUtilities.cpp:1459
#4  0x0000555555c1feff in WbConcreteNodeFactory::validateExistingChildNode(WbField const*, WbNode const*, WbNode const*, bool, QString&) const (this=0x5555561e74b8 <WbConcreteNodeFactory::gFactory>, field=0x55556a6a01e0, childNode=0x5555610dbeb0, node=0x55555d05ff70, isInBoundingObject=false, errorMessage=...)
    at nodes/utils/WbConcreteNodeFactory.cpp:476
#5  0x0000555555a7a4a2 in WbNode::validate(WbNode const*, WbField const*, bool) const (this=0x55555d05ff70, upperNode=0x0, upperField=0x0, isInBoundingObject=false) at vrml/WbNode.cpp:869
```

I will wait for the fix of these first crashes to continue with the tests.

ad-daniel:
Both crashes ended up being caused by the same issue. Basically, visible candidates are ignored and when we do so, we also remove from the candidate list any ancestors to it so that they wouldn't get deleted indirectly. However, the opposite can be true as well, meaning that the descendants that are also candidates shouldn't be deleted either as they might be referenced or related to the visible node. 

The base node of this PROTO is a Slot but in practice when building the structure I believe it takes the endPoint type as higher level of the PROTO (in this case it treats the PROTO as being of type Shape). That's the only reasonable explanation I can find to explain why it creates  parameter nodes for things such as `TextureCoordinate`, `Coordinate` and `IndexedFaceSet` which are fully internal unless the PROTO base node is a Shape. These nodes also appeared among the candidates for removal, however since the Shape cannot be removed from the candidate list as it exposes the PBR fields, nor should these other nodes be removed but they were.

A solution that appears to be fixing both crashes is to remove from the candidate list both ancestors to the visible node as well as descendants.

stefaniapedrazzi:
:eyes:

src/webots/nodes/WbBaseNode.cpp
```diff
@@ -124,6 +125,269 @@ void WbBaseNode::validateProtoNodes() {
...
+  /*
+  // TODO: remove before merge
+  printf("\nINVISIBLE PROTO PARAMETER NODES (WHAT WILL BE REMOVED)\n");
+  for (int i = 0; i < invisibleProtoParameterNodes.size(); ++i) {
+    printf("  [L%d] %s [%p]\n", invisibleProtoParameterNodes[i]->level(),
+           invisibleProtoParameterNodes[i]->usefulName().toUtf8().constData(), invisibleProtoParameterNodes[i]);
+  }
+  */
+
+  if (invisibleProtoParameterNodes.size() == 0)
```
stefaniapedrazzi, line 319: usually we use the isEmpty() method
```suggestion
  if (invisibleProtoParameterNodes.isEmpty())
```

src/webots/nodes/WbBaseNode.cpp
```diff
@@ -124,6 +125,269 @@ void WbBaseNode::validateProtoNodes() {
...
+
+    internalProtoNodes[i]->setProtoParameterNode(NULL);  // break link with proto parameter node
+  }
+
+  // break link [parameter] -> [internal field] and [parameter node] -> [internal node] (from parameter node side)
+  for (int i = 0; i < invisibleProtoParameterNodes.size(); ++i) {
+    invisibleProtoParameterNodes[i]->clearProtoParameterNodeInstances();  // clear downward references
+
+    // clear internal field references (for protoParameterNodes the reference is kept in its fields)
+    QVector<WbField *> fields = invisibleProtoParameterNodes[i]->fields();
```
stefaniapedrazzi, line 338:
```suggestion
    const QVector<WbField *> fields = invisibleProtoParameterNodes[i]->fields();
```

src/webots/nodes/WbBaseNode.cpp
```diff
@@ -124,6 +125,269 @@ void WbBaseNode::validateProtoNodes() {
...
+        parent->removeFromFieldsOrParameters(parameters[j]);
+      } else {
+        if (mfnode && mfnode->nodeIndex(parameterNode) != -1) {
+          mfnode->blockSignals(true);
+          mfnode->removeNode(parameterNode);
+          mfnode->blockSignals(false);
+          parent->removeFromFieldsOrParameters(parameters[j]);
+        }
+      }
+    }
```
stefaniapedrazzi, line 386: Instead of duplicating this block you should merge fields and parameters in the same QVector and run the loop only once.

stefaniapedrazzi (changes requested):
I tried different tests with nested and complex PROTOs and everything seems to work fine.
I just have a couple of change requests on the code and then it remains to remove all the debugging functions.

Are the speedup measurements in the description still up-to-date to the latest changes?

stefaniapedrazzi:
Please write a message here when the PR is ready for review.
Note that some tests failed.

ad-daniel:
Since the performance gained at runtime by applying this trick is dwarfed by the increased load time necessary to take advantage of it (going through the entire node tree and removing invisible nodes), it does not make much sense. Therefore it can be closed for now.

PR #2928, opened 2021-03-31
Fix cppcheck 2.4.1 errors

stefaniapedrazzi:
New cppcheck errors are thrown during the macOS sources tests where the new version 2.4.1 is used.

When checking the projects sources, many `ctuOneDefinitionRuleViolation` errors are thrown because the same class name is used in many plugins and libraries for different robots.
For example the Qt robot windows of the e-puck, RobotisOP2, automobile have classes named `Viewer`.
However given that these are different programs this doesn't cause any issues and I disabled the check.

### DIFF
--- a/src/webots/gui/WbMultimediaStreamingLimiter.cpp
+++ b/src/webots/gui/WbMultimediaStreamingLimiter.cpp
@@ -44,8 +44,6 @@ QSize WbMultimediaStreamingLimiter::fullResolution() const {
 void WbMultimediaStreamingLimiter::recomputeStreamingLimits(int skippedImages) {
   mIsStopped = false;
   mResolutionChanged = false;
-  if (mIsStopped)
-    return;
 
   if (skippedImages > 0)
     mIncreasingSteps++;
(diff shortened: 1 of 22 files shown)


src/webots/gui/WbMultimediaStreamingLimiter.cpp
```diff
@@ -44,8 +44,6 @@ QSize WbMultimediaStreamingLimiter::fullResolution() const {
 void WbMultimediaStreamingLimiter::recomputeStreamingLimits(int skippedImages) {
   mIsStopped = false;
   mResolutionChanged = false;
-  if (mIsStopped)
```
stefaniapedrazzi, line 47: This should be the only change where the behavior is different than before.
I tested on my system and it still works as expected, but locally the connection is good.

stefaniapedrazzi:
It seems that some tests have been skipped.
I'm checking if there are some errors.

src/webots/nodes/utils/WbDictionary.cpp
```diff
@@ -87,20 +87,22 @@ bool WbDictionary::updateDef(WbBaseNode *&node, WbSFNode *sfNode, WbMFNode *mfNo
   } else if (useCase) {
     if (isAValidUseableNode) {
       const WbNode::NodeUse nodeUse = node->nodeUse();
-      bool match = false;
+      bool match;
```
omichel, line 90: If defNodes.size() == 0 match will be undefined and tested at line 109...
```suggestion
      bool match = false;
```

stefaniapedrazzi: Reintroducing this definition is throwing error.
I will check for other solutions to avoid suppressing the error.

omichel (approved):
Thank you.

PR #2932, opened 2021-04-01
Add `wb_supervisor_node_get_pose` function

lukicdarkoo:
Often, users ask for an easy way to get a relative translation and rotation between two objects. This function resolves that issue.

Furthermore, this function is more efficient than manually calculating transformation between two objects as everything is done in a single request (compared to 4 requests otherwise, 2 x (orientation + position)).

**Documentation**
https://cyberbotics.com/doc/reference/supervisor?version=feature-relative-pose

**Tasks**
- [x] Add Supervisor C API
  - [x] C++
  - [x] Python
  - [x] Java
  - [x] Matlab
  - [x] ROS
- [x] Documentation
- [x] Tests
- [x] Check scale
- [x] ROS tests
- [x] Controller.def + WbLanguage.cpp
- [x] Changelog

**Test**
```python


def check(a, b):
    assert abs(a - b) < 0.005, f'{a} != {b}'


robot = Supervisor()
while robot.step(int(robot.getBasicTimeStep())) != -1:
    pose = robot.getSelf().getPose(robot.getRoot())
    # pose = robot.getSelf().getPose()
    orientation = robot.getSelf().getOrientation()
    position = robot.getSelf().getPosition()
    
    for i, j in enumerate([0, 1, 2, 4, 5, 6, 8, 9, 10]):
        check(pose[j], orientation[i])
    
    check(pose[3], position[0])
    check(pose[7], position[1])
    check(pose[11], position[2])
```


docs/reference/changelog-r2021.md
```diff
@@ -33,6 +33,7 @@ Released on June, Xth, 2021.
     - Altered the collision detection logic for [Robot.selfCollision](robot.md) to ignore chains of joints if the intermediary joints all share the same `anchor` point ([#2868](https://github.com/cyberbotics/webots/pull/2868)).
     - Added conversion from PROTO to URDF from the Webots command line ([#2885](https://github.com/cyberbotics/webots/pull/2885)).
     - Added flag to [RobotisOp2](../guide/robotis-op2.md) that enables the modeling of backlash in the robot ([#2881](https://github.com/cyberbotics/webots/pull/2881)).
+    - Added the `wb_supervisor_node_get_pose` function that retrives an absolute pose or a pose between two nodes ([#2932](https://github.com/cyberbotics/webots/pull/2932)).
```
omichel, line 36:
```suggestion
    - Added the `wb_supervisor_node_get_pose` function that retrieves an absolute or relative pose. Relative pose is expressed in the coordinate system of another node specified as an argument ([#2932](https://github.com/cyberbotics/webots/pull/2932)).
```

docs/reference/supervisor.md
```diff
@@ -868,48 +875,9 @@ p' = R * p + T
...
-box_pos_robot = np.dot(rot_ur10e, box_pos_world)
-
-# Calculate the orientation of the box, relative to the robot, all in one line.
-box_rot_robot = np.dot(rot_ur10e, np.array(box.getOrientation()).reshape(3, 3))
-```
-
-%end
+The `wb_supervisor_node_get_pose` function returns an array of 16 values.
+The array shall be interpreted as a 4 x 4 [transformation matrix](https://en.wikipedia.org/wiki/Transformation_matrix) that represents an absolute transform of the node.
+If the `from_node` argument is given than the function returns a relative pose of the node in respect of the node `from_node`.
```
omichel, line 880:
```suggestion
The function returns the relative pose of the node with respect to the node specified in `from_node`. If `from_node` is null, it returns the absolute pose of the node in the global coordinate system.
```

docs/reference/supervisor.md
```diff
@@ -820,6 +825,7 @@ public class Node {
 ```MATLAB
 position = wb_supervisor_node_get_position(node)
 orientation = wb_supervisor_node_get_orientation(node)
+pose = wb_supervisor_node_get_pose(node, from_node=0)
```
omichel, line 828: I am not sure the default value of 0 is supported in MATLAB. Did you check it?

src/controller/c/supervisor.c
```diff
@@ -387,6 +387,9 @@ static WbNodeRef self_node_ref = NULL;
 static WbNodeRef position_node_ref = NULL;
 static WbNodeRef export_string_node_ref = NULL;
 static WbNodeRef orientation_node_ref = NULL;
+static double *relative_pose = NULL;
```
omichel, line 390: Instead of dynamic allocation / free, it would be simpler and faster to have a stack allocation, isn't it?
```suggestion
static double relative_pose[16];
```

src/controller/c/supervisor.c
```diff
@@ -965,6 +973,12 @@ static void supervisor_read_answer(WbDevice *d, WbRequest *r) {
       for (i = 0; i < 9; i++)
         orientation_node_ref->orientation[i] = request_read_double(r);
       break;
+    case C_SUPERVISOR_NODE_GET_POSE:
+      free(relative_pose);
+      relative_pose = malloc(16 * sizeof(double));
```
omichel, line 978: Not needed in case of stack allocation:
```suggestion
```

include/controller/cpp/webots/Node.hpp
```diff
@@ -130,6 +130,8 @@ namespace webots {
     Field *getProtoField(const std::string &fieldName) const;
     const double *getPosition() const;
     const double *getOrientation() const;
+    const double *getPose() const;
+    const double *getPose(const Node *fromNode) const;
```
omichel, line 134: Isn't that simpler?
```suggestion
    const double *getPose(const Node *fromNode = NULL) const;
```

src/controller/cpp/Node.cpp
```diff
@@ -105,6 +105,14 @@ const double *Node::getOrientation() const {
   return wb_supervisor_node_get_orientation(nodeRef);
 }
 
+const double *Node::getPose() const {
+  return wb_supervisor_node_get_pose(nodeRef, NULL);
+}
+
```
omichel, line 111: Using a default argument would save us from this implementation:
```suggestion
```

omichel (changes requested):

include/controller/cpp/webots/Node.hpp
lukicdarkoo: Unfortunately no because SWIG gets confused and doesn't know how to make Python and JAVA interface with default values. Alternatively, I write a custom code in `controller.i` but then it is not consistent with `worldSave`

src/controller/cpp/Node.cpp
lukicdarkoo: See above

src/controller/c/supervisor.c
lukicdarkoo: Hmm, that is a good point, but it is strange that there are a lot of variables doing the opposite like `solid_velocity`, `virtual_reality_headset_position`, `virtual_reality_headset_orientation`... I am not sure about the reason, maybe we wanted to make sure the memory chunk is invalid after a new value is requested?

docs/reference/supervisor.md
lukicdarkoo: I think it is better now

src/controller/c/supervisor.c
omichel: I don't see a good reason for that.
I believe we should also simplify the other cases (`solid_velocity`, etc.) in a different PR.

docs/reference/supervisor.md
omichel: Yes, it looks very good.

src/controller/c/supervisor.c
stefaniapedrazzi: I don't remember exactly, but one reason is that a pointer also contains one more information than an array, i.e. if the value is set or not.
For example if some errors occurs and the value could not be computed by Webots, with a null pointer it is easier to detect if the call was invalid or not (see comment on the following line). 
https://github.com/cyberbotics/webots/blob/49573cdfa90e59c7db9f7b3de764ad4d1610c65e/src/controller/c/supervisor.c#L2198
With an array, Webots should send an aswer to the libController and explicitly set the array valued to NaN (same as in the `invalid_vector`)

omichel (approved):
Thank you.

PR #2935, opened 2021-04-05
Updating Darwin-opHinge2.proto with virtual RoboCup procedure

medrimonia:
The main parameters of the joints and motors are now accessible using local
variables (as it was done for backlash).

This change has the following impacts for all joints:

- maxTorque is slightly decrease (2.5 -> 2.3)
- maxVelocity is strongly decreased (12.26 -> 5.24)
- dampingConstant is strongly increased (0.002 -> 0.27)
- staticFriction is strongly increased (0.025 -> 0.80)
- resolution of the sensor is now limited

Impact of the behavior has been tested in the following way: on a RoboCup field,
the robot can still walk and stand-up from both side.

**Additional context**

Note that the procedure might still evolve, but the parameters are now much closer to what should be announced by the teams for this kind of motors.

### DIFF
--- a/projects/robots/robotis/darwin-op/protos/Darwin-opHinge2.proto
+++ b/projects/robots/robotis/darwin-op/protos/Darwin-opHinge2.proto
@@ -232,7 +232,7 @@ PROTO Darwin-opHinge2 [
             }
             PositionSensor {
               name "NeckS"
-              resolution %{= mx28_resolution}%
+              resolution %{= mx28_resolution }%
             }
           ]
           jointParameters HingeJointParameters {
@@ -252,7 +252,7 @@ PROTO Darwin-opHinge2 [
             }
             PositionSensor {
               name "HeadS"
-              resolution %{= mx28_resolution}%
+              resolution %{= mx28_resolution }%
             }
           ]
           jointParameters2 JointParameters {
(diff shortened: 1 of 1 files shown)


ad-daniel (approved):
Looks good to me, I've only fixed a minor formatting issue. The robot appears to still walk/stand without problems in the other DarwinOp worlds.

jgueldenstein:
I can confirm that these values correspond to those from the performance graph of the dynamixel motors https://emanual.robotis.com/docs/en/dxl/mx/mx-28/.

jgueldenstein:
One thing I am uncertain of is, weather the values for a supply voltage of 11.1V (current values by @medrimonia) or the values for 12V should be used. From the Dynamixel documentation it is not clear to me to which voltage the performance graph corresponds.

medrimonia:
Thanks for fixing the formatting issue!

> From the Dynamixel documentation it is not clear to me to which voltage the performance graph corresponds.

I think it's rather 11.1V on this performance graph because the speed seems to converge to 50 rpm (11.1V) rather than 55rpm (12V).

jgueldenstein:
I believe you are right to use these values. the nominal voltage of the Darwin battery is 11.1V

PR #2939, opened 2021-04-06
Addition of coupled motors

ad-daniel:
**Description**
This PR introduces the ability to control multiple motors, that share the same name, at the same time.

**Working principle**
- Webots side, each motor keeps track of all other motors that have the same name as itself (based on `cMotors`). They all need to have this list because in principle any of them could be directly called (if `get_device_by_index` is used, if `get_device` is used, it will usually always be the first one)
- Whenever an API call is made, for e.g., `wb_motor_set_velocity`, the motor under question handles the message and also relays the same command to its siblings, if any. Each will adapt this command according to their own `multiplier`.
- Not restrained by node structure, potentially any two motors can be coupled. 
- No changes to the tag assignment/indexing logic to maintain compatibility

**Tasks**
  - [ ] Core functionality
    - [x] velocity control
    - [x] torque control
    - [ ] position control
  - [ ] Demo
  - [ ] Unit test
  - [ ] Documentation
  - [ ] Cleanup

**Screenshots**
TODO


ad-daniel:
Concerning the motor [API](https://cyberbotics.com/doc/reference/motor), this feature will require some adjustment:
```
wb_motor_set_position
wb_motor_set_velocity
wb_motor_set_acceleration
wb_motor_set_available_force
wb_motor_set_available_torque
wb_motor_set_force
wb_motor_set_torque
```
These functions are fine as the imposition of these values is automatically relayed to the sibling motors.
```
wb_motor_set_control_pid
```
Forcing all motors to have the same PID parameters seems restrictive. So a different approach is necessary.
```
wb_motor_get_target_position
wb_motor_get_min_position
wb_motor_get_max_position
wb_motor_get_velocity
wb_motor_get_max_velocity
wb_motor_get_acceleration
wb_motor_get_available_force
wb_motor_get_max_force
wb_motor_get_available_torque
wb_motor_get_max_torque
wb_motor_enable_force_feedback
wb_motor_disable_force_feedback
wb_motor_get_force_feedback
wb_motor_get_force_feedback_sampling_period
wb_motor_enable_torque_feedback
wb_motor_disable_torque_feedback
wb_motor_get_torque_feedback_sampling_period
wb_motor_get_torque_feedback
wb_motor_get_type
wb_motor_get_brake
wb_motor_get_position_sensor
```
Similarly, all the getters would return the value/references only of the first motor. It could be argued that if we're using coupled motors is because we don't care to command each of them individually but I think we should add a way of accessing the others anyway. The last four especially need it.

A minimally invasive solution could be to add:
```
int wb_robot_get_number_of_coupled_motors(const char *name)
WbDeviceTag * wb_robot_get_motor_siblings(const char *name)
```
or instead of the tags, it returns the indexes but that would require an additional call to `wb_robot_get_device_by_index` to make use of it. Either option however suffers a bit from lacking clarity, we have the tags/indexes but no clear indication of what or where this motor is. That's a consequence of dropping the unique names, and I'm not sure much can be done about it, unless a middle ground of enforcing name a structure in the world file such as `"motor_name:0"`, `"motor_name:1"`, etc which when using setters applies the command to all motors called `"motor_name"` but when using getters, it returns either the first if calling `getter("motor_name")` or the specific one if calling `getter("motor_name:1")` (where the number refers to the order of appearance in the structure).

Also, should be added:
```
double wb_motor_get_multiplier(WbDeviceTag tag);
```

what approach do you think is best @omichel ?

ad-daniel:
https://user-images.githubusercontent.com/44834743/116724996-a53e6e80-a9e1-11eb-959f-71225a7314b3.mp4

ad-daniel:
The PR is ready for review. At the level of `WbMotor` both plain setter functions (to change the limits of the motors) and functions that enforce the relayed commands are needed. The nomenclature is already fairly confusing and is due to the fact that the effect of the API setters varies depending on which control strategy is currently active. For instance, in position control an API setVelocity sets the maximal velocity whereas in velocity control it sets the target velocity.

In this class, associated to the API position setter the function `setTargetPosition` already existed and I've considered calling the others something similar like `setTargetVelocity` and `setTargetAcceleration` etc but for the reasons mentioned I don't think it would've been an improvement towards making it more clear. Therefore:

- For the other API callback functions, I've called them a generic `setVelocity`, `setForceOrTorque` as their effect depends on the context. 
- I've left the `setTargetPosition` name unchanged as it's public and used elsewhere (calling it `setPosition` might bring confusion with respect to the joint `setPosition`)
- The plain setters are called according to the limit they change, so `setMaxVelocity`, `setMaxAcceleration` (although the acceleration variable is called `mAcceleration` ...), etc.

- `setMaxVelocity` existed already and was doing something else. However as far as I could tell it wasn't used anywhere so it was repurposed for the corresponding plain setter.

omichel (changes requested):
Are we getting a warning in Webots if a motor has no coupled devices and has a name including the delimiter `::`?
Are we getting a warning in Webots if a motor has no coupled siblings but has a multiplier != 1?

Also, I have some doubts about the `multiplier` usage.
I believe it would be simpler to define it as the following:
- it applies to all motors (including non-coupled motors).
- it defines by which factor the position, velocity and force/torque command will be multiplied (or divided in case of force/torque) regardless of the siblings if any.
This way, if ever you remove a motor from a coupling set, it will behave the same as before if controlled individually. It will also simplify the understanding of the effect of the multiplier.

Finally, I believe you were right when suggesting the force/torque should be multiplied by the multiplier, like the position and velocity. It is simpler to understand and I believe it corresponds better to real logically coupled motors. On the other hand, I would remove the constraint on the maxForce/maxTorque not being consistent with the multiplier and I would silence the warnings about sending a force/torque command beyond the maxForce/maxTorque in case of coupled motors. That seems a better modeling of the fact that even though its multiplier is high (maybe due to a gear box), a coupled motor doesn't necessary have more force/torque than its sibling.

src/webots/nodes/WbMotor.cpp
```diff
@@ -127,20 +130,31 @@ void WbMotor::postFinalize() {
...
     dynamic_cast<WbMuscle *>(it.next())->postFinalize();
 
-  connect(mMaxVelocity, &WbSFDouble::changed, this, &WbMotor::updateMaxVelocity);
-  connect(mAcceleration, &WbSFDouble::changed, this, &WbMotor::updateMaxAcceleration);
+  connect(mMultiplier, SIGNAL(changed()), this, SLOT(updateMultiplier()));
+  connect(mMaxVelocity, SIGNAL(changed()), this, SLOT(updateMaxVelocity()));
+  connect(mMinPosition, SIGNAL(changed()), this, SLOT(updateMinAndMaxPosition()));
+  connect(mMaxPosition, SIGNAL(changed()), this, SLOT(updateMinAndMaxPosition()));
+  connect(mAcceleration, SIGNAL(changed()), this, SLOT(updateMaxAcceleration()));
+  connect(mMaxForceOrTorque, SIGNAL(changed()), this, SLOT(updateMaxForceOrTorque()));
```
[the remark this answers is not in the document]
ad-daniel: The reason I had to change it to this was because the slot functions had default values which isn't otherwise supported in the new syntax. Either way, I can do it without having to rely on default parameters so I changed it back to what it was.

projects/samples/devices/worlds/coupled_motors.wbt
```diff
@@ -0,0 +1,444 @@
...
+          }
+          DEF RIGHT_ACTUATED HingeJoint {
+            jointParameters HingeJointParameters {
+              axis 0 1 0
+              anchor -0.008 0 0
+            }
+            device [
+              RotationalMotor {
+                name "motor::right finger"
+                maxVelocity 10.000000000000004
```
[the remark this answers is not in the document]
ad-daniel: I changed it back and tried saving and it didn't return. No clue where that came from

src/controller/c/motor.c
```diff
@@ -533,6 +537,18 @@ double wb_motor_get_available_torque(WbDeviceTag tag) {
...
 }
 
+double wb_motor_get_multiplier(WbDeviceTag tag) {
+  double multiplier = NAN;
+  robot_mutex_lock_step();
+  Motor *m = motor_get_struct(tag);
+  if (m)
+    multiplier = m->multiplier;
+  else
+    fprintf(stderr, "Error: %s(): invalid device tag.\n", __FUNCTION__);
```
omichel, line 547: This saves an unneeded double initialization:
```suggestion
  double multiplier;
  robot_mutex_lock_step();
  Motor *m = motor_get_struct(tag);
  if (m)
    multiplier = m->multiplier;
  else {
    fprintf(stderr, "Error: %s(): invalid device tag.\n", __FUNCTION__);
    multiplier = NAN
  }
```

src/controller/c/motor.c
```diff
@@ -174,6 +176,8 @@ static void motor_read_answer(WbDevice *d, WbRequest *r) {
       m->control_i = m->previous_control_i;
       m->control_d = m->previous_control_d;
       m->position = request_read_double(r);
+      m->velocity = request_read_double(r);
```
omichel, line 179: Why did you add velocity here? It doesn't seem to be useful to anything.

omichel (changes requested):

src/controller/c/motor.c
ad-daniel: Because when the controller sets the velocity, webots is informed about it and in turn it relays the command to its siblings. If these siblings don't notify libController about the change in velocity, any call to `wb_motor_get_velocity` that directly targets one of these siblings would return their old value, not the new one.

omichel (approved):
Thank you!

lukicdarkoo:
We are missing ROS tests for the `wb_motor_get_multiplier` function

PR #2940, opened 2021-04-06
Improve controller linking

lukicdarkoo:
We add the `rpath` to better support a dynamic library loading in Python. Otherwise, we have to configure `LD_LIBRARY_PATH` before the `python3 myscript.py` call.

Unfortunately, this works fine only for Linux. For macOS, it may be possible (https://longwei.github.io/rpath_origin/), but Windows doesn't support `rpath` at all (https://stackoverflow.com/questions/42497999/build-project-with-libraries-in-other-directories-windows-mingw).

### DIFF
--- a/docs/guide/running-extern-robot-controllers.md
+++ b/docs/guide/running-extern-robot-controllers.md
@@ -38,10 +38,10 @@ Generic Webots environment variables needed for all the controller languages:
 
 %tab "Linux"
 
-| Environment Variable     | Typical Value                                    |
-|--------------------------|--------------------------------------------------|
-| WEBOTS\_HOME             | `/usr/local/webots`                              |
-| LD\_LIBRARY\_PATH        | add `${WEBOTS_HOME}/lib/controller`              |
+| Environment Variable                                  | Typical Value                                    |
+|-------------------------------------------------------|--------------------------------------------------|
+| WEBOTS\_HOME                                          | `/usr/local/webots`                              |
+| LD\_LIBRARY\_PATH (not needed for Python controllers) | add `${WEBOTS_HOME}/lib/controller`              |
 
 %tab-end
 
--- a/docs/guide/running-extern-robot-controllers.md
+++ b/docs/guide/running-extern-robot-controllers.md
@@ -28,11 +28,11 @@ Generic Webots environment variables needed for all the controller languages:
 %tab-component "os"
 
 %tab "Windows"
-| Environment Variable                        | Typical Value                                                             |
-|---------------------------------------------|---------------------------------------------------------------------------|
-| WEBOTS\_HOME                                | `C:\Program Files\Webots`                                                 |
-| Path (all controllers)                      | add `%WEBOTS_HOME%\lib\controller` and `%WEBOTS_HOME%\msys64\mingw64\bin` |
-| Path (for C++, Python and Java controllers) | add `%WEBOTS_HOME%\msys64\mingw64\bin\cpp`                                |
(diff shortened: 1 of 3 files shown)


omichel (commented):
Shall we also modify the documentation accordingly?
https://cyberbotics.com/doc/guide/running-extern-robot-controllers?tab-language=python&tab-os=linux#environment-variables

lukicdarkoo:
Long-term it would be very nice to have statically linked controllers for Python. Regarding the RAM usage, it should make a big impact as the controller library is small, without external dependencies.

docs/guide/running-extern-robot-controllers.md
```diff
@@ -38,10 +38,10 @@ Generic Webots environment variables needed for all the controller languages:
...
 %tab "Linux"
 
-| Environment Variable     | Typical Value                                    |
-|--------------------------|--------------------------------------------------|
-| WEBOTS\_HOME             | `/usr/local/webots`                              |
-| LD\_LIBRARY\_PATH        | add `${WEBOTS_HOME}/lib/controller`              |
+| Environment Variable                                  | Typical Value                                    |
+|-------------------------------------------------------|--------------------------------------------------|
+| WEBOTS\_HOME                                          | `/usr/local/webots`                              |
+| LD\_LIBRARY\_PATH (not needed for Python controllers) | add `${WEBOTS_HOME}/lib/controller`              |
```
omichel, line 44: That applies from Python 3.8 on:
```suggestion
| LD\_LIBRARY\_PATH (not needed for Python >= 3.8 controllers) | add `${WEBOTS_HOME}/lib/controller`              |
```
And it also applies to Windows and macOS.
Moreover, for Windows, we should also document the `MSYS2_HOME` environment variable in case of development environment.

omichel (changes requested):

docs/guide/running-extern-robot-controllers.md
lukicdarkoo: Why does it apply only to Python 3.8? It should work with any version.

`rpath` is not supported for Windows and macOS, so it will not work there.

omichel: Because os.add_dll_directory() was introduced only in this version of Python, see: https://github.com/cyberbotics/webots/blob/master/src/controller/python/controller.i#L28-L37

omichel: See the explanation here: https://stackoverflow.com/questions/60509120/python-3-8-not-loading-c-dll

omichel: And that is specific to Windows: https://github.com/Toblerity/Fiona/issues/851

omichel: https://docs.python.org/3/library/os.html#os.add_dll_directory

lukicdarkoo: Oh nice, since Windows doesn't support $ORIGIN this is great. Let me try to make $ORIGIN work for macOS, so we have all platforms covered.

lukicdarkoo: It indeed works on macOS as well:
```
lukic@Cyberboticss-Mac-mini webots % echo $DYLD_LIBRARY_PATH

lukic@Cyberboticss-Mac-mini webots % python3.7 -c "import controller; controller.Supervisor()"
Webots doesn't seems to be ready yet. Retrying in 1 second.
...
```

omichel (approved):
Thank you.

PR #2941, opened 2021-04-06
fix net

BenjaminDeleze:
**Description**
The streaming viewer cannot load images that have an URL of the form ```../../../textures/net.png``` because it does not support the ```../```

We currently have this format for the net of the robocup goal and for the white grass of the robocup soccerfield.

- [x] Fix the net by adding the net texture image in the general texture folder.
- [x] Add a new option "white artificial" to the grass proto instead of having a dedicated proto for this kind of grass.
- [x] Update the robocup soccerfield with the new grass

Concerning the net fix, this may not be the most elegant solution but it is supposed to be temporary pending the implementation of getting images with full URL.


omichel:
It seems the RobocupSoccerField proto was not update accordingly.

omichel (approved):
Thank you!

PR #2943, opened 2021-04-07
update tiago_reference

BenjaminDeleze:
Fix the tiago_reference.urdf to match the newest tiago definition. Fix #2942

PR #2944, opened 2021-04-07
Merge master into develop

lukicdarkoo:
No conflicts.


stefaniapedrazzi:
`robot_urdf_tiago` failed so it seems there will be issues in the develop branch after merging this.

lukicdarkoo:
> `robot_urdf_tiago` test failed so it seems there will be issues in the develop branch after merging this.

I thought it was fixed in https://github.com/cyberbotics/webots/pull/2943, but I have just noticed that the master branch is failing because of the same reason.

lukicdarkoo:
It looks like there is a random Tiago failure. I suggest merging this. Then, I will create another PR to print the two files if the reference and the generated files are not the same. It will give us a better insight into what is happening and whether it is a random failure.

stefaniapedrazzi (approved):

PR #2950, opened 2021-04-07
Print generated URDF test file if test fails

lukicdarkoo:
We want to get more details when URDF export is broken


stefaniapedrazzi:
Could you please add more details why the `compare_files` function is not working?
Do we really need to add an additional block? Wouldn't be better to fix directly the `compare_files` code?

stefaniapedrazzi:
Oops sorry, this PR is just about printing the resulting URDF.
We should add a comment explaining what does the new block.

stefaniapedrazzi (approved):
Thank you

PR #2951, opened 2021-04-07
Add JetBot robot

stefaniapedrazzi:
Add the JetBot robot.

* [x] add robot PROTO
* [x] create icon
* [x] calibrate the physical behavior
* [x] add an example
* [x] create the web component studio 
* [x] documentation
* [x] update ChangeLog


stefaniapedrazzi:
The JetBot robot model and sample simulations are ready for review.

There are just two points I'm not sure what is the best choice.
* it would be good to provide some pre-trained model for the "jetbot_collision_avoidance" controller, but the file is too large and cannot be added to the GitHub repo. So we could add a link the controller documentation from which the user can download the model and dataset. But where should this be stored? Should we restored a projects folder on the cyberbotics.com server?
  Currently the user has to create its own dataset and train it using the provided controller.
* I created a separate PROTO for the JetsonNano embedded module. I put it in the `projects/devices` folder even if it is a sully graphical object, because it made to me more sense than adding to the `projects/objects`, but it is neither an actuator nor a sensor. So I'm not sure which the best location (once defined I will also add the documentation of this PROTO).

lukicdarkoo:
> * I created a separate PROTO for the JetsonNano embedded module. I put it in the `projects/devices` folder even if it is a sully graphical object, because it made to me more sense than adding to the `projects/objects`, but it is neither an actuator nor a sensor. So I'm not sure which the best location (once defined I will also add the documentation of this PROTO).

We can replace the AlexNet with MobileNet which should be a few MBs (~5MB). So we can keep it in the repository, but not include it in the distribution, just automatically download it when needed. In addition, the MobileNet is faster to train because of the size and residual layers.

stefaniapedrazzi:
> We can replace the AlexNet with MobileNet which should be a few MBs (~5MB). So we can keep it in the repository, but not include it in the distribution, just automatically download it when needed. In addition, the MobileNet is faster to train because of the size and residual layers.

That's a good idea.
But one of the point was to implement exactly the same code as provided in the robot tutorial Jupyter notebooks.
I will try to check the file size with the other version using resnet18 that they also provide.

stefaniapedrazzi:
We decided to create a GitHub repo just for storing this kind of controllers taking lot of disk space and let the users manually download them and avoid adding big files into the repo.

For the JetsonNano PROTO, it was decided to change the base node to Robot, keep it in the "devices" folder and add a special section in the documentation names "Single Board Computers".

docs/guide/single-board-computers.md
```diff
@@ -0,0 +1,39 @@
+# Single Board Computers
+
+## Single Board Computer List
+
+All the commercially available single board computers models included in Webots are included here.
+Users may also add their own custom models.
+Please [contact us](https://www.cyberbotics.com/contact) if you would like to see your favorite single board computer here.
```
BenjaminDeleze, line 7:
```suggestion
Please [contact us](https://www.cyberbotics.com/#contact) if you would like to see your favorite single board computer here.
```

docs/guide/single-board-computers.md
```diff
@@ -0,0 +1,39 @@
...
+  SFRotation rotation       0 1 0 0
+  SFString   name           "NVIDIA Jetson Nano"
+  SFNode     physics        NULL
+  SFString   controller     ""
+  SFString   window         ""
+  MFNode     extensionSlot  []
+]
+```
+
+> **File location**: "WEBOTS\_HOME/projects/devices/nvidia/protos/JetsonNano.proto"
```
BenjaminDeleze, line 35:
```suggestion
> **File location**: "[WEBOTS\_HOME/projects/devices/nvidia/protos/JetsonNano.proto]({{ url.github_tree }}/projects/devices/nvidia/protos/JetsonNano.proto)"
```

BenjaminDeleze (commented):
I only found some minor corrections to do

The web component studio needs to be regenerated in order to work with the current version of the viewer. 

The `scrips/web_component_studio/components.json` has no entry to generate the x3d file for the JetBot.

docs/guide/single-board-computers.md
omichel: I would also remove the `www.` prefix from the URL.

stefaniapedrazzi:
> The web component studio needs to be regenerated in order to work with the current version of the viewer.

Ok, I will do it now if you don't have any other comment about the robot model to avoid regenerating multiple times.

stefaniapedrazzi:
> The scrips/web_component_studio/components.json has no entry to generate the x3d file for the JetBot.

I will add it, but it seems it is not listing all the available robots.

docs/guide/single-board-computers.md
stefaniapedrazzi: We have to apply these changes also 4 other pages. I will do these changes in a separate PR.

BenjaminDeleze:
> I will add it, but it seems it is not listing all the available robots.

I think it contains all the robots present in the docs because I was able to regenerate them all using this .json when I transitioned to the new viewer.

stefaniapedrazzi:
Web component studio doesn't seem to correctly support the URL images of the format "https://raw.githubusercontent.com/cyberbotics/webots/<...>.png".
Given that the robot uses both the `BrushedAluminium` PROTO and the individual textures retrieved from GitHub, the exporter handles them as different instances (with numbering) but only one instance is automatically copied in the docs folder. Manually copying the missing duplicated textures solves the loading issues.

stefaniapedrazzi:
All the comments should be addressed.

BenjaminDeleze (dismissed):
Seems fine for me, thank you!
do not forget to also merge the PR on webots-projects

omichel (approved):
Thank you.

PR #2953, opened 2021-04-09
Enhancement defaultShader for webgl

BenjaminDeleze:
**Description**
We need to add a fragNormal output to the defaultShader.frag in order for it to be usable in WebGL.
I implemented the calculation of the fragNormal in the same way as it is in Phong.

defaultShader is use when there is no appearance or when there is a Phong appearance with a texture but not material.

However it does change a bit the rendering, especially when there is a texture. 
It does not really change when there is not texture :
Before:
simple_cube_1_white

After:
simple_cube_3_white

With a texture:
Before:
simple_cube

After:
simple_cube_3

With a Phong materiel (so phong shaders) to compare:
simple_cube_Phong

### DIFF
--- a/resources/wren/shaders/default.frag
+++ b/resources/wren/shaders/default.frag
@@ -6,10 +6,12 @@ const int mainTextureIndex = 0;
 const int penTextureIndex = 1;
 const int backgroundTextureIndex = 2;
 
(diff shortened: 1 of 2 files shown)


omichel (approved):
Thank you.

PR #2954, opened 2021-04-09
Fix joint positions in URDF CLI export

lukicdarkoo:
I missed some steps from the `preFinalize` stage and translation fields have never been updated. It should be fine now:
image

### DIFF
--- a/src/webots/maths/WbVector3.hpp
+++ b/src/webots/maths/WbVector3.hpp
@@ -222,7 +222,7 @@ public:
   }
 
   // text conversion
-  QString toString(WbPrecision::Level level) const {
+  QString toString(WbPrecision::Level level=WbPrecision::DOUBLE_MAX) const {
     return QString("%1 %2 %3")
       .arg(WbPrecision::doubleToString(mX, level))
       .arg(WbPrecision::doubleToString(mY, level))
--- a/src/webots/maths/WbVector3.hpp
+++ b/src/webots/maths/WbVector3.hpp
@@ -222,7 +222,7 @@ public:
   }
 
   // text conversion
-  QString toString(WbPrecision::Level level=WbPrecision::DOUBLE_MAX) const {
+  QString toString(WbPrecision::Level level = WbPrecision::DOUBLE_MAX) const {
     return QString("%1 %2 %3")
       .arg(WbPrecision::doubleToString(mX, level))
       .arg(WbPrecision::doubleToString(mY, level))
--- a/src/webots/nodes/utils/WbNodeUtilities.cpp
+++ b/src/webots/nodes/utils/WbNodeUtilities.cpp
@@ -819,7 +819,7 @@ WbRobot *WbNodeUtilities::findRobotAncestor(const WbNode *node) {
   if (!node)
     return NULL;
 
-  while (node->parentNode()) {
+  while (node) {
     if (isRobotTypeName(node->nodeModelName())) {
       const WbRobot *robot = reinterpret_cast<const WbRobot *>(node);
       return const_cast<WbRobot *>(robot);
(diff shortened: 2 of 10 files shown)


omichel (approved):

stefaniapedrazzi:
I'm reviewing this PR and I have some comments. I will be done soon..

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -819,7 +819,7 @@ WbRobot *WbNodeUtilities::findRobotAncestor(const WbNode *node) {
   if (!node)
     return NULL;
 
-  while (node->parentNode()) {
+  while (node) {
```
stefaniapedrazzi, line 822: After this change, the function name is no longer consistent with the functionality.

I didn't check why this change is needed but previously the check if the node was a robot was done outside this function. So we should be consistent and keep the function as it was before, or change the name of the function and adapt the previous calls of this function.

stefaniapedrazzi (commented):
I have a couple of comments and change requests on this PR.

src/webots/maths/WbVector3.hpp
```diff
@@ -222,7 +222,7 @@ class WbVector3 {
   }
 
   // text conversion
-  QString toString(WbPrecision::Level level) const {
+  QString toString(WbPrecision::Level level = WbPrecision::DOUBLE_MAX) const {
```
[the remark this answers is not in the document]
lukicdarkoo: I am not sure what is a good approach. It's annoying when debugging. I believe it is important to have the option to change the precision, but having a default one is handy as very often we don't care about the precision.

src/webots/nodes/utils/WbNodeUtilities.cpp
lukicdarkoo: Without the fix, if the robot node is the root (it always used to be Group) this function fails to find the robot ancestor.

But you are right, maybe I should detect that edge case instead of doing a quick fix.

lukicdarkoo: Actually, it doesn't change the behavior, but only fixes the edge case. The behavior and the function don't match in any case (if I am not missing something). If the passed node is a robot it will return the node itself.

stefaniapedrazzi: It changes the behavior because if the node is robot but doesn't have any robot ancestor previously it was returning NULL and now it returns the node.
A function that searches for ancestors (as described by the name) is not meant to detect if the node is a robot or not. Note that this check is almost always done externally before calling this function.

lukicdarkoo: Oh, I see, I thought you were pointing at something else

src/webots/maths/WbVector3.hpp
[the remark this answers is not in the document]
stefaniapedrazzi: Ok, but I'm worried that we will always use the default option without thinking about which one is the best option to use.

src/webots/nodes/utils/WbNodeUtilities.cpp
stefaniapedrazzi: Actually, I'm wrong. This is only the condition that assumes that every nodes as a parent (top parent is Group). So the behavior is unchanged.

src/webots/maths/WbVector3.hpp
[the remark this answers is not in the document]
lukicdarkoo: That would be a problem if default precision was low, but since it is `DOUBLE_MAX` it is not harmful.

PR #2957, opened 2021-04-12
Fixed bug in Robot::getDevice()

omichel:
The C++ API functions Robot::getDevice() and similar (Robot.getCamera(), Robot.getPositionSensor(), etc.) were behaving wrong as they were sending a new pointer each time they were called with the same device name as reported [here](https://github.com/RoboCup-Humanoid-TC/webots/issues/32).
This PR fixes this bug.


stefaniapedrazzi:
Does it work correctly in Python and Java? Because the exact same IF condition is applied

omichel:
Yes, the fix needed to be applied in Python and Java as well.

stefaniapedrazzi:
:eyes:

stefaniapedrazzi (approved):
Thank you

Issue #2934, opened 2021-04-02
Incorrect lens distortion


jnoyola:
**Describe the Bug**
Lens distortion appears to incorrectly transform vertical lines. I only started experimenting with lens distortion today, so please correct me if I misunderstood, but based on the documentation and other resources I found, it appears incorrect.

**Steps to Reproduce**
1. Create a camera.
2. Add a lens.
3. Set the `radialCoefficients.x` to a positive value (e.g. 1).
4. Place an object in front of the camera and observe it near the left and the right of the image.

**Expected behavior**
Horizontal lines are bowed out from the center as expected.
Vertical lines on the right side of the image are bowed out from the center as expected.
Vertical lines on the left side of the image should also be bowed out from the center, but instead remain vertical.

![image](https://user-images.githubusercontent.com/4064932/113399010-5b1b8c00-9354-11eb-8bbe-5234d3abbbec.png)

**Screenshots**
When the object is near the right of the image, it appears as expected.
![image](https://user-images.githubusercontent.com/4064932/113389662-44216d80-9345-11eb-97d5-81fd61542a21.png)
When the object is near the left of the image, vertical lines remain vertical. Note the right side of the checkerboard in fact bows to the left.
![image](https://user-images.githubusercontent.com/4064932/113389872-9bbfd900-9345-11eb-8fe0-f12593f4a7d1.png)

**System**
 - Operating System: Windows 10
 - Graphics Card: NVIDIA GeForce RTX 2080 8 GB

**Additional context**
Not sure if it's related, but sources such as [this](https://www.sciencedirect.com/topics/engineering/lens-distortion) suggest that barrel distortion would have a negative K<sub>1</sub>, but that appears to create pincushion distortion in Webots.

omichel:
Relates with #2295.

PR #2961, opened 2021-04-12
Fix Camera lens distortion

### LINKED ISSUE #2295: Vertical effect in spherical camera image
A user asked for vertical effect in spherical image because currently only horizontal effect is applied:

horizontal_effect
vertical_effect

It seems correct to also expect a vertical spherical effect.
So we should maybe evaluate:
* if it is interesting to add a vertical effect
* if `spherical` is the best description for the image effect or if we could choose a better name (for example `equirectangular` or `cylindric` projection instead of `spherical` projection)

stefaniapedrazzi:
Fix #2295 and #2934:
Camera lens distortion uses the Brown's distortion model, but there were bugs in the implementation and the center was not used for the radial and tagential components.
Additional, to apply lens distortion it is needed to apply an inverted Brown's distortion model.

Now, the distortion is correctly applied both in X and Y directions, as well as symmetrically on the right/left of the image. Here is the result using `Lens { center = [0.5, 0.5], radialCoefficients = [-0.5, 0], tangetialCoefficients = [0, 0]}`:

lens_distortion

### DIFF
--- a/docs/reference/changelog-r2021.md
+++ b/docs/reference/changelog-r2021.md
@@ -13,6 +13,7 @@ Released on XX Xth, 2021.
     - Added a nice looking FIFA soccer ball proto ([#2782](https://github.com/cyberbotics/webots/pull/2782)).
     - Added an `allowedChannels` field in the [Emitter](emitter.md) and [Receiver](receiver.md) nodes to restrict the channel usage ([#2849](https://github.com/cyberbotics/webots/pull/2849)).
   - Bug fixes
+    - Fixed [Lens](lens.md) distortion ([#2961](https://github.com/cyberbotics/webots/pull/2961)).
     - Fixed return value type of [`CameraRecognitionObject.get_size`](camera.md#camera-recognition-object) Python function ([#2923](https://github.com/cyberbotics/webots/pull/2923)).
     - Fixed [`Camera.getRecognitionObjects`](camera.md#wb_camera_recognition_get_objects) function not available and the return value of [`CameraRecognitionObject.getPositionOnImage`](camera.md#camera-recognition-object) and [`CameraRecognitionObject.getSizeOnImage`](camera.md#camera-recognition-object) in Java API ([#2923](https://github.com/cyberbotics/webots/pull/2923)).
     - Fixed detection of scaled objects in the [Camera](camera.md) image using the [Recognition](recognition.md) functionality ([#2921](https://github.com/cyberbotics/webots/pull/2921)).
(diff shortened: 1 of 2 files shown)


stefaniapedrazzi:
[Brown distortion model](https://en.wikipedia.org/wiki/Distortion_(optics)), that is used in the lens distortion fragment shader, applies lens correction and not lens distortion
So I'm now trying to compare the results with other tools that perform lens distortion/correction to check if the results are matching or if we need to invert the formula.

stefaniapedrazzi:
The issue with the used formula was already raised here: https://github.com/cyberbotics/webots-dev/issues/5478

I found two papers providing formulas to render lens distortion:
1) [Realistic Lens Distortion Rendering](https://www.researchgate.net/publication/326968643_Realistic_Lens_Distortion_Rendering)
2) [An Exact Formula for Calculating Inverse Radial Lens Distortions](https://www.researchgate.net/publication/303703794_An_Exact_Formula_for_Calculating_Inverse_Radial_Lens_Distortions)

Formula in 2) only considers the radial distortion and not the tangential but it also generates wrong pincushion distortion if K1 is greater than 0.4.
This is why I chose formula 1) for the lens distortion shader.

However, with any of the two formulas (that are approximations) it is not possible to use the exact same parameters to correct the lens distortion.
And I didn't find any implementation applying the lens distortion to compare the results.
The implementations I found are using the Brown distortion model as-is to apply lens distortion even if it is meant for distortion correction.

omichel (approved):
Very good, thank you.

egordv:
Hello!
The lens distortion model now seems to be correct, at least in the feature-robocup-controllers branch. 
But the output image with radial distortion doesn't take into account pixels that are outside of camera's visible area and thus has a heavy artefacts around it's borders. Please check the image provided:
image

Left is an undistorted image, right is distorted.  
I'm using PROTO with camera fieldOfView=1.762 and radialCoefficients = [-0.26, -0.26]
If resulting distorted pixel coordinate is outside the visible area - it will have a colour of the closest pixel inside the visible area.
This leads to heavy artifacts around distorted image area as in right part of an image provided.
The partially visible robocup soccer ball in the corner of an image looks really strange and confusing our detection pipeline (check the left lower corner of an image).
What I'm expecting is the pixels outside visible area to be black, not repeating closest visible pixel's colour.
The image without this artifact should look like this: 
image

omichel:
See also this post: https://hl.forum.robocup.org/t/modeling-wideangle-lenses-in-webots/495/3

PR #2972, opened 2021-04-14
Fix typo verticalfieldofview

mael25:
The verticalFieldOfView of the HokuyoUtm30lx was computed using a value for the fieldOfView that was corresponding to other Hokuyo lidars.

Datasheet : https://www.hokuyo-aut.jp/search/single.php?serial=169
Webots doc : https://www.cyberbotics.com/doc/guide/lidar-sensors#hokuyo-utm-30lx


omichel (changes requested):
Thank you!
We should probably add a line in the change log about this fix as it may break existing user simulations.

omichel (dismissed):
Thank you.

mael25:
I had to merge master into this branch as a new bug entry was added in the meantime...

omichel (approved):

PR #2769, opened 2021-02-19
Feature new streaming-viewer

BenjaminDeleze:
**Description**
This PR introduces the new streaming-viewer based on WREN.

**Tasks**
Functionalities
- [x] be able to display the same objects/worlds as the current version but with almost the same graphics as in Webots
- [x] support for animation
- [x] finish last todos in the code
- [x] Test on Mac
- [x] Capsule primitive
- [x] improve the Makefile to make the new streaming-viewer
- [x] support for other coordinate systems
- [x] Installation instruction for Linux
- [x] Installation instruction for MacOS
- [x] Installation instruction for Windows
- [ ] support for the documentation viewer
- [x] add support for different fonts (in PR #2976 )

Web interface
- [x] Improve the disconnect / reconnect
- [x] Add loading gif in animation
- [x] Reset the viewpoint when resetting the simulation 
- [ ] improve the navigation (rotation) in the viewer for mobile phone
- [x] improve the animation viewer with fullscreen mode and better looking playbar (in PR #2979 )

Cleanup
- [x] initial cleanup
- [x] Restore the ros submodule (it was deleted because it was creating issues with githubpages)
- [x] Remove generated files (needed for the use with github pages now)
- [x] Restore the files that have modifications unrelated to this PR
- [x] Cleanup the old code of Webotsjs
  - [x] Replace ```var``` by ```const``` and ```let```
  - [x] Delete unused files
  - [x] Get rid of the external dependency to load .hdr file
  - [x] Replace ```XMLHttpRequest``` with the ```fetch``` API when possible
- [x] Cleanup of the new code
  - [x] Cleanup of utils/
  - [x] Cleanup of nodes/
  - [x] Cleanup of wren/
- [x] Change animation template to point towards the right urls
- [x] Change index.html to point towards remote links
- [x] Add necessary components to webots server

Know Bugs
  - [x] Restore mjpeg streaming
  - [x] Viewpoint not reset correctly in animation with camera following an object
  - [ ] Viewpoint follow problem with animation
  - [ ] GTAO is lighter than the one of Webots due to glClipControl
  - [ ] Slowdown on Windows 10 and macOS with Firefox. Due to glLinkProgram on phongShader, phongStencilAmbientEmissiveShader, phongStencilDiffuseSpecularShader, pbrShader and pbrStencilDiffuseSpecularShader

PR #2976, opened 2021-04-16
Feature fonts web

BenjaminDeleze:
Add the support for different fonts in the web interface.
Better handle the background of labels


resources/web/templates/x3d_playback.html
```diff
@@ -5,7 +5,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <title>%title%</title>
     <link type="text/css" href="https://www.cyberbotics.com/jquery-ui/1.11.4/jquery-ui.min.css" rel="stylesheet"/>
-    <link type="text/css" href="https://www.cyberbotics.com/wwi/R2021b/wwi.css" rel="stylesheet"/>
+    <link type="text/css" href="https://www.cyberbotics.com/wwi/Wrenjs/wwi.css" rel="stylesheet"/>
```
stefaniapedrazzi, line 8: This path is ok for testing.
But when in production we need to have a folder for each major version so that previously Webots releases are not broken when applying changes to the develop branch.

BenjaminDeleze: Indeed, but we needed to share the animation player with the people of robocup so that they can use it and also for testing.
But we should revert it back once #2769 is ready to be merged

Issue #2982, opened 2021-04-19
Webots crashes after changing Lidar-type


KNTRL9:
**Describe the Bug**
Webots R2021a crashes consistently after changing the "type" of the Lidar-Node from "rotating" to "fixed" or vice versa and pressing "Reset Simulation".

**Steps to Reproduce**
1. Go to 'Webots\projects\samples\devices\worlds\lidar.wbt'
2. Change the type of the Lidar-Node to "fixed"
3. Reset the simulation
4. Webots Crashes either immediately or after a short freeze

Out of 10 times, it crashed 8 times. The other 2 times it took several more seconds where I could Play/Pause/Reset until it crashed.

**Expected behavior**

**Screenshots**

**System**
 - Windows 10 (Same for Ubuntu 18.04.5)
 - Nvidia Quadro K2100M on Windows Device

**Additional context**

stefaniapedrazzi:
After changing the `Lidar.type` value, the internal image size should be updated but it is not the case.
So when resetting the internal image the wrong size is used causing the crash due to write outside the boundaries.

I'm also inspecting the other issues caused by the missing update in order to find the better fix.

omichel:
@stefaniapedrazzi: can we close this issue?

stefaniapedrazzi:
This issue is fixed in #2983.
Given that the dimensions of the returned data depends on the `Lidar.type`, we had to remove the possibility to change it during simulation run and force saving and reverting the simulation in order to apply the type change.

PR #2983, opened 2021-04-19
Fix updating Lidar.type during simulation run

stefaniapedrazzi:
Many internal structures (WREN camera, image arrays, etc.) depends on the `Lidar.type` value but the size of these structure was not updated when then `Lidar.type` was changed causing some issues in particular when switching from "rotating" to "fixed" (height and width of the lidar is bigger for the "fixed" type).
Here are some of the issues:
- crash #2982 when executing the `memset` instruction in the `WbLidar::reset` method
- wrong graphical point cloud representation

There are already a number of fields for Camera (width, height) and Lidar (fieldOfView, verticalFieldOfView, horizontalResolution, numberOfLayers) nodes whose modification is only applied after saving and reloading the world.
So the best and more consistent fix is to require saving and reloading the world also to change the Lidar type.

### DIFF
--- a/src/webots/nodes/WbLidar.cpp
+++ b/src/webots/nodes/WbLidar.cpp
@@ -76,6 +76,7 @@ void WbLidar::init() {
   mActualHorizontalResolution = mHorizontalResolution->value();
   mActualVerticalFieldOfView = mVerticalFieldOfView->value();
   mActualFieldOfView = mFieldOfView->value();
+  mActualType = mType->value();
 }
 
 WbLidar::WbLidar(WbTokenizer *tokenizer) : WbAbstractCamera("Lidar", tokenizer) {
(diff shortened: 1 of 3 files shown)


src/webots/nodes/WbLidar.cpp
```diff
@@ -679,7 +681,10 @@ void WbLidar::updateType() {
     parsingWarn(tr("'type' should either be 'fixed' or 'rotating', reset to 'fixed'"));
     mType->setValue("fixed");
   }
-  if (areWrenObjectsInitialized())
+  if (hasBeenSetup())
+    warn(tr("'type' has been modified. This modification will be taken into account after saving and reloading "
+            "the world."));
```
omichel, line 686: Can't we have this string on a single line, it seems it fits within our max line length coding style, isn't it?
```suggestion
    warn(tr("'type' has been modified. This modification will be taken into account after saving and reloading the world."));
```

omichel (changes requested):

omichel (approved):
Thank you!

PR #2988, opened 2021-04-21
Add namespace to model_name topic for ROS controller

eborghi10:
**Description**

For multiple robots with ROS controllers, it's necessary to have a namespace for the `model_name` topic name so they don't collide with each other.

**Tasks**

  - [x] Add namespace to topic name
  - [x] [Make sure other parts of the repo are not using it](https://github.com/cyberbotics/webots/search?q=model_name&type=code)

### DIFF
--- a/projects/default/controllers/ros/Ros.cpp
+++ b/projects/default/controllers/ros/Ros.cpp
@@ -159,7 +159,7 @@ void Ros::launchRos(int argc, char **argv) {
   mNodeHandle = new ros::NodeHandle;
   ROS_INFO("The controller is now connected to the ROS master.");
 
-  mNamePublisher = mNodeHandle->advertise<std_msgs::String>("model_name", 1, true);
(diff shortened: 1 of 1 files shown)


stefaniapedrazzi:
The purpose of this function was to retrieve the robot name from the code itself without having to hard code it.
If this function now requires the robot name, how can the robot name be retrieved programmatically?

Here you can find some sources using it:
https://github.com/cyberbotics/webots_ros/search?q=model_name

By the way this kind of change that breaks the compatibility have to be merged in the `develop` branch.

stefaniapedrazzi:
Here you can also find an example handling multiple ros controllers:
https://github.com/cyberbotics/webots_ros/blob/master/src/robot_information_parser.cpp#L68

eborghi10:
I tested it and the proposed solution is enough. Ideally, each controller should publish its own topic in my opinion, but I can deal with it. Thanks, @stefaniapedrazzi!

PR #2991, opened 2021-04-21
Improve wb_supervisor_node_reset_physics

### LINKED ISSUE #2899: SoccerBall is moving after setting its position and resetting velocity
**Describe the Bug**

**Steps to Reproduce**
1. Have a SoccerBall object

**Expected behavior**

**Video**

**System**

**Additional context**

**Workaround**

stefaniapedrazzi:
Fix #2899:
Independently if the node physics is reset before or after moving it artificially with the supervisor, during the physics step ODE handles the jerk and overwrites the linear/angular velocities, etc.
The only alternative to run a simulation step between the artificial move and the physics reset, is to reset the node physics again after the ODE step computations.

Additionally I also found that the ball joint was not handled in the ODE joint physics reset.


stefaniapedrazzi:
All the tests passed but the Windows one due to VS 10 installation.

omichel (approved):
Thank you!

PR #2995, opened 2021-04-21
Remove blank line proto info

mael25:
Remove the blank line appearing for PROTO nodes in the small info window that appears on hover. This is done by removing the last character which is in this case always '\n' in non-empty strings.

before
after


omichel (approved):
Thank you!

PR #2997, opened 2021-04-22
Hotfix lib controller update after reset

### LINKED ISSUE #2992: Crash using negative index in field operations after reset
The field count on the controller side is not updated when nodes are removed due to simulation reset.
As a consequence subsequent calls of field operations using negative index may compute an invalid index that makes Webots crash.

Example:
1) At simulation start a world has X nodes -> libController root children field count set to X
2) Supervisor controller imports a new node at index `-1` that is converted to `X`
3) Webots imports the new node at position `X` -> libController root children field count set to X+1
4) Supervisor controller resets the simulation
5) Webots deletes the latest added node and world has `X` nodes but libController is not notified
6) Supervisor controller imports a new node at index `-1` that is converted to `X+1`
3) Webots imports the new node at position `X+1`
     -> Webots crashes because valid index should be in range `[0, X]`

stefaniapedrazzi:
Fix #2992: when the simulation is reset, all the pending answers to the controllers were also reset including the fields and nodes updates due to the nodes deleted during the reset.
But if the controller is not restarted, we should not reset the pending answers because the controller is still running and it is still waiting for these messages.

The new message to update the fields count in the libController has been added in R2021b.
Some workarounds exists to prevent Webots to crash that can be applied on R2021a rev1.


omichel (approved):
Thank you!

PR #2999, opened 2021-04-22
Add Fabtino robot

mael25:
Adding [Robotics Equipment Corporation](http://servicerobotics.eu/en/homepage/)'s Fabtino robot.

fabtino

**Tasks**
  - [x] Finish appearance
  - [x] Clean code
  - [x] Doc
  - [x] Get info about physics of robot / Update doc accordingly
  - [x] Fix `contactProperties` issue


mael25:
@BenjaminDeleze On this robot, the mesh for the wheels was already offseted, so that adding it without translation as geometry would put it at the exact position. However, I have just seen that on the x3d, when we hover on the wheels, the little circle that appears is at the origin, as I guess it checks the position of the endPoint Solid and not the anchor of the HingeJoint.

Do you think it makes sense to change how the position is computed ?

BenjaminDeleze (changes requested):
> On this robot, the mesh for the wheels was already offseted, so that adding it without translation as geometry would put it at the exact position.


As discussed it is better if you delete the offset such that the endpoint position correspond to the position of the wheel.

On the image, there is a cylinder bounding object at each corner of the robot. I think you did that to approximate the corners. But since the curvature is not too big, I don't know if it would maybe be better to just approximate the whole frame of the robot with one single box. It would be a bit less precise but would reduce the number of needed bounding objects. What do you think @mael25, @omichel ?
fabtino_bo

mael25:
@BenjaminDeleze I have centered the wheelsmeshes and removed the unnecessary pillar in the world file.

BenjaminDeleze (approved):
Thank you!

PR #2964, opened 2021-04-12
Fix the saving of hidden field for BallJoint

ad-daniel:
**Related Issues**
This pull-request addresses issue #2963

**Description**
It appears that indeed the saving of the hidden parameters for `BallJoint` wasn't working. 
Before the fix:
```
#VRML_SIM R2021a utf8
WorldInfo {
  coordinateSystem "NUE"
  lineScale 0.4
}
Viewpoint {
  orientation -0.36364034235430376 -0.9164891884416874 -0.16677310599116954 0.9279522774354931
  position -1.0729115992676148 0.5311913032755937 0.9792082997483017
}
TexturedBackground {
}
TexturedBackgroundLight {
}
RobotWithHinge2Joint {
  hidden position_0_0 0.12800000000000003
  hidden position2_0_0 0.25344000000000005
  hidden rotation_1 0.44864394902849164 0.8918808888552964 0.05715843852666757 0.28377458981805487
}
RobotWithBallJoint {
  hidden position_0_0 0.19199999999999998
  hidden rotation_1 -0.436450151027134 0.4018780050889322 0.8049877854316374 0.7167288954785301
  translation 0.5 0 0
}
```
After the fix:
```
#VRML_SIM R2021a utf8
WorldInfo {
  coordinateSystem "NUE"
  lineScale 0.4
}
Viewpoint {
  orientation -0.36364034235430376 -0.9164891884416874 -0.16677310599116954 0.9279522774354931
  position -1.0729115992676148 0.5311913032755937 0.9792082997483017
}
TexturedBackground {
}
TexturedBackgroundLight {
}
RobotWithHinge2Joint {
  hidden position_0_0 0.12160000000000004
  hidden position2_0_0 0.24064000000000008
  hidden rotation_1 0.4490486589044304 0.8918563250593967 0.05429178011154328 0.2694860452190886
}
RobotWithBallJoint {
  hidden position_0_0 0.18239999999999998
  hidden position2_0_0 0.35583999999999993
  hidden position3_0_0 0.52032
  hidden rotation_1 -0.44247596067295786 0.3964965005313276 0.8043665515751843 0.6785561719474464
  translation 0.5 0 0
}
```
**Test World**
World used for tests [ball_joint_save_hidden.zip](https://github.com/cyberbotics/webots/files/6297851/ball_joint_save_hidden.zip)

PR #3001, opened 2021-04-23
Fix loading of hidden position3 for Ball

ad-daniel:
**Description**
PR #2964 only addressed the saving of the hidden position to file, in practice however while loading it also purposefully ignores the possibility of a `position3` (which ended up overwriting `position`).
Since it slipped through, I've also a unit test that verifies the loading of hidden position for all joints.


src/webots/nodes/WbBallJoint.cpp
```diff
@@ -381,6 +381,8 @@ double WbBallJoint::initialPosition(int index) const {
 }
 
 void WbBallJoint::setPosition(double position, int index) {
+  WbHinge2Joint::setPosition(position, index);
+
   if (index == 3) {
```
omichel, line 386: To improve the code readability, I would rather do something like:
```C++
  if (index != 3)
    return;
```
(the same applies for the other types of joints)

tests/parser/controllers/generic_sensor/generic_sensor.c
```diff
@@ -0,0 +1,44 @@
...
+#include <webots/robot.h>
+
+#define TIME_STEP 64
+
+int main(int argc, char **argv) {
+  wb_robot_init();
+
+  WbDeviceTag emitter = wb_robot_get_device("emitter");
+
+  int nb_devices = atoi(argv[1]);
```
omichel, line 14:
```suggestion
  const int nb_devices = atoi(argv[1]);
```

tests/parser/controllers/generic_sensor/generic_sensor.c
```diff
@@ -0,0 +1,44 @@
...
+      positions[i] = -1.0;
+  }
+
+  char outbuffer[50];
+
+  sprintf(outbuffer, "%.4f %.4f %.4f %s\n", positions[0], positions[1], positions[2], wb_robot_get_name());
+
+  while (wb_robot_step(TIME_STEP) != -1.0) {
+    wb_emitter_send(emitter, outbuffer, 50 * sizeof(char));
+  }
```
omichel, line 40:
```suggestion
  while (wb_robot_step(TIME_STEP) != -1.0)
    wb_emitter_send(emitter, outbuffer, 50 * sizeof(char));
```

tests/parser/controllers/generic_sensor/generic_sensor.c
```diff
@@ -0,0 +1,44 @@
...
+  for (int i = 0; i < 3; ++i) {
+    if (i < nb_devices)
+      positions[i] = wb_position_sensor_get_value(sensors[i]);
+    else
+      positions[i] = -1.0;
+  }
+
+  char outbuffer[50];
+
+  sprintf(outbuffer, "%.4f %.4f %.4f %s\n", positions[0], positions[1], positions[2], wb_robot_get_name());
```
omichel, line 36: We should probably use `snprintf` in general instead of `sprintf` (together with `sizeof(outbuffer)`).

tests/parser/controllers/generic_sensor/generic_sensor.c
```diff
@@ -0,0 +1,44 @@
...
+
+  while (wb_robot_step(TIME_STEP) != -1.0) {
+    wb_emitter_send(emitter, outbuffer, 50 * sizeof(char));
```
omichel, line 39:
```suggestion
    wb_emitter_send(emitter, outbuffer, strlen(outbuffer) + 1);
```

tests/parser/controllers/hidden_parameter_single/hidden_parameter_single.c
```diff
@@ -27,6 +28,43 @@ int main(int argc, char **argv) {
...
+      else if (strcmp(robot_name, "Hinge2Joint") == 0) {
+        ts_assert_double_in_delta(position, 0.333, delta, "Hinge2Joint position is %.4f instead of %.4f", position, 0.333);
+        ts_assert_double_in_delta(position2, 0.444, delta, "Hinge2Joint position2 is %.4f instead of %.4f", position2, 0.444);
+      } else if (strcmp(robot_name, "BallJoint") == 0) {
+        ts_assert_double_in_delta(position, 0.555, delta, "BallJoint position is %.4f instad of %.4f", position, 0.555);
+        ts_assert_double_in_delta(position2, 0.666, delta, "BallJoint position2 is %.4f instead of %.4f", position2, 0.666);
+        ts_assert_double_in_delta(position3, 0.777, delta, "BallJoint position3 is %.4f instead of %.4f", position3, 0.777);
+      } else {
+        ts_assert_boolean_equal(0, "Message unknown.");
+      }
```
omichel, line 60:
```suggestion
      } else
        ts_assert_boolean_equal(0, "Message unknown.");
```

tests/parser/controllers/hidden_parameter_single/hidden_parameter_single.c
```diff
@@ -27,6 +28,43 @@ int main(int argc, char **argv) {
...
+
+      wb_receiver_next_packet(receiver);
+    }
+  }
+
+  ts_assert_boolean_equal(received_message, "Should've received at least one message, but didn't.");
```
omichel, line 66:
```suggestion
  ts_assert_boolean_equal(received_message, "Should have received at least one message, but didn't.");
```

omichel (changes requested):
Very good. I only have a few minor suggestions.

stefaniapedrazzi (changes requested):
Parser tests are different than the other tests because they mainly focus on detecting crashes at load or checking the messages printed in the console without running the simulation.

This kind of tests on hidden parameters should probably be moved to the `tests/physics` folder where the other tests about hidden fields are already located.

ad-daniel:
> This kind of tests on hidden parameters should probably be moved to the `tests/physics` folder where the other tests about hidden fields are already located.

I've modified an existing test which already verified the loading of hidden parameters and was in parser folder. Should the entire test unit be moved then?

stefaniapedrazzi:
> I've modified an existing test which already verified the loading of hidden parameters and was in parser folder. Should the entire test unit be moved then?

Yes. The whole `hidden_parameter_single` test should be moved to the `physics` folder because it is now also testing the physics behavior after loading the model.

ad-daniel:
Moved and removed the redundant returns. Thank you.

stefaniapedrazzi (approved):
Thank you

omichel (approved):

PR #3002, opened 2021-04-23
Add loading time and average real time factor to performance log

ad-daniel:
**Description**
This PR adds additional info to the performance log.

Loading time options:
1. Measure time only for node structure creation
2. Entire load process (basically from [here](https://github.com/cyberbotics/webots/blob/1367ab2560688c435afa50b8713522b9897afe7c/src/webots/nodes/utils/WbWorld.cpp#L102) to [here](https://github.com/cyberbotics/webots/blob/1367ab2560688c435afa50b8713522b9897afe7c/src/webots/engine/WbSimulationWorld.cpp#L106)) including things like `handleInitialCollisions`, loading plugins, etc

The current solution corresponds to option (2)  for loading timing.

### DIFF
--- a/src/webots/engine/WbSimulationWorld.cpp
+++ b/src/webots/engine/WbSimulationWorld.cpp
@@ -172,15 +172,14 @@ WbSimulationWorld::~WbSimulationWorld() {
 
 void WbSimulationWorld::step() {
   WbPerformanceLog *log = WbPerformanceLog::instance();
-  if (log) {
+  if (log)
     log->stepChanged();
-    log->lapTime(WbPerformanceLog::SPEED_FACTOR);
-  }
 
   const double timeStep = basicTimeStep();
 
   if (WbSimulationState::instance()->isRealTime()) {
     const int elapsed = mRealTimeTimer.restart();
+
     // computing the mean of an history of several elapsedTime
     // improves significantly the stability of the algorithm
     // in case of simulations where elapsedTime oscillates often
@@ -304,7 +303,7 @@ void WbSimulationWorld::modeChanged() {
       WbSoundEngine::setPause(true);
       WbSoundEngine::setMute(WbPreferences::instance()->value("Sound/mute").toBool());
       if (log)
-        log->invalidateMeasure(WbPerformanceLog::SPEED_FACTOR);
+        log->stopMeasure(WbPerformanceLog::SPEED_FACTOR);
       break;
     case WbSimulationState::STEP:
       WbSoundEngine::setMute(WbPreferences::instance()->value("Sound/mute").toBool());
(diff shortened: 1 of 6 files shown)


src/webots/engine/WbSimulationWorld.cpp
```diff
@@ -165,14 +172,15 @@ WbSimulationWorld::~WbSimulationWorld() {
 
 void WbSimulationWorld::step() {
   WbPerformanceLog *log = WbPerformanceLog::instance();
-  if (log)
+  if (log) {
     log->stepChanged();
+    log->lapTime(WbPerformanceLog::SPEED_FACTOR);
```
lukicdarkoo, line 177: Why not merging `lapTime` and `stepChanged` functions?

src/webots/engine/WbSimulationWorld.cpp
```diff
@@ -165,14 +172,15 @@ WbSimulationWorld::~WbSimulationWorld() {
...
+  }
 
   const double timeStep = basicTimeStep();
 
   if (WbSimulationState::instance()->isRealTime()) {
     const int elapsed = mRealTimeTimer.restart();
-
```
lukicdarkoo, line 175: Not needed change

src/webots/engine/WbSimulationWorld.cpp
```diff
@@ -287,12 +295,16 @@ void WbSimulationWorld::restartStepTimer() {
...
+  WbPerformanceLog *log = WbPerformanceLog::instance();
+
   const WbSimulationState::Mode mode = WbSimulationState::instance()->mode();
   switch (mode) {
     case WbSimulationState::PAUSE:
       mTimer->stop();
       WbSoundEngine::setPause(true);
       WbSoundEngine::setMute(WbPreferences::instance()->value("Sound/mute").toBool());
+      if (log)
+        log->invalidateMeasure(WbPerformanceLog::SPEED_FACTOR);
```
lukicdarkoo, line 307: Can we pause the measurement here? Or, just not stop the timer?

lukicdarkoo (changes requested):
Looks very good, but there are a few things I don't understand (more details in the comments).

Don't forget to add this to the changelog.

src/webots/engine/WbSimulationWorld.cpp
ad-daniel: You're right, the speed factor is tied to the step change and there's no other uses for the `lapTime` function so there's no need to make a function for it.

src/webots/engine/WbSimulationWorld.cpp
ad-daniel: You're right, I thought it would pollute the time accumulator variable if paused mid-run but I checked and it doesn't. Even pausing for a long time and resuming still yields the right result.

lukicdarkoo (approved):
Thank you

PR #3003, opened 2021-04-23
Support for label in animation

BenjaminDeleze:
**Description**
Currently the labels are not recorded in animation, the PR aims to change that.

For animations to be able to read and update correctly the labels, we need to change the way they are transmitted to webots.js. An update of label will be transmitted along the updates of other objects, with a timestamp, instead of alone without timestamp

**Tasks**
  - [x] Modify how webots send the labels
  - [x] Modify X3dScene.js to read the new format of labels 
  - [x] Support for multiple labels
  - [x] Modify WbAnimationRecorder to take labels into account
  - [x] Send labels only when they are updated
  - [x] Remove the old way of sending labels


BenjaminDeleze:
duplicate of #3019

PR #2152, opened 2020-08-27
Fix robot-window not up to date.

DavidMansolino:
**Description**
When a extern controller is stopped and restarted the robot-window was not linked to the new controller.

**Related Issues**
This pull-request fixes issue #2080

PR #3005, opened 2021-04-23
Allows failed controller to reconnect to a robot

### LINKED ISSUE #1903: shutdown of python controller does not close cleanly connection
When we run an extern python controller, sometimes (well most of the time) the communication is not correctly closed and restarting the controller leads to an error message "Failed to attach extern robot controller" image

The message "INFO '<extern>' controller existed successfully" does not appear on logging windows

We started discussing this on discord but I think it is easier here.

The controller is not my code so I am going through it but it seems there is no RobotController only a supervisor and some cameras. There is also no thread running.

to try to debug I made a simple script creating a supervisor and deleting it with del. it make the controller disconnect, so that works for a simple case

in our application I create the main object, close all threads then delete the supervisor but that does not disconnect. There is probably some dependencies preventing the C object to be deleted. so I tried to go through all objects and delete them by hand but it does not trigger the connection close....

Is there a way to find out what is keeping the connection open?

although generally I would suggest to allow to disconnect manually instead of using the delete methods of python objects. you could use the "with" statement for example. Implicit connections in python is a recipe for nightmares...

UPDATE 2002-08-02:

Can be easily reproduced by starting webots then running following code:

```
import time
from threading import Thread

class T(Thread):
    def run(self):
        while not self.ev_stop:
            time.sleep(0.01)
            print(time.time())


s = Supervisor()

while s.step(5000) != -1:
    pass
```

then interupting with two times ctrl-C
If you restart the script one more time you get the Error on webots

Since this require two times ctrl-C it is very easy to avoid, thus not very high priority

### LINKED ISSUE #3000: Allows external controller to reconnect after been killed
Hello, 
First of all, I wish to say that Webot is so nice, thank a lot for it !
Moreover I wish to contribute, but I'm new to it, and the solution I've found could be pushed with a regular PR, but my brain melt when I've tried to do so. Please help if possible :)

**Describe the Bug**
I'm working with external C controller, which work very well.
However, if the controller died, it can't call the cleanup function, and when I restart it, I get a error message from webots because he think that the controller is still running, so he can't find a spot for the new incoming connection.
I'm often in this configuration, because I'm debugging with gdb, and we kill everything when I've found the location of a bug, for instance.

**Workaround**
The workaround is to restart the simulation via the GUI each time, but it's not very handy...

**Solution**
I wish the controller could reconnect automatically.
I've tested a solution, which is to add the following code in `src/webots/control/WbControlledWorld.cpp`, between lines : 222 and 223
It simply remove the dead controller connection as usual, then recreate a new one. (a bit of code duplication from the line above, my bad... It might be possible to avoid that).
The trick is to use the name of the robot to know which dead controller should be kill. (which is fine for me). Otherwise, if no name is provided, the code execute as usual (scanning for free spot).
Maybe a more elegant solution is possible, by reassigning a socket id, for instance, but I'm first trying to get it working. I've noticed the robot_id, which maybe could be used for that, but I've not investigated in that direction.
It's have not been tested in every way possible, but I'm not seeing major issues coming.
The main thing to check would be leak of memory, unclosed socket and so on, because I'm not sure how is it performed.

```c
     robotName = buffer;
      delete[] buffer;
    }
    foreach (WbRobot *const robot, robots()) {
+      if (robotName == robot->name() && robot->controllerName() == "<extern>" && robot->isControllerStarted()) {
+        updateRobotController(robot);
+        WbLog::info(tr("Restarting extern controller connection for robot \"%1\".").arg(robot->name()));
+        mRobotsWaitingExternController.append(robot);
+        startControllerFromSocket(robot, socket);
+        return;
+      }
      if (robot->isControllerStarted())
        continue;
```
I would be glad if I got feedback about it :) And if this is a correct solution, I would be happy to get some help to push it properly, because I would enjoy contribute, but a need to get the whole PR process working!

Thank in advance :)

PS : I've noticed later than could solve this issue : 
https://github.com/cyberbotics/webots/issues/1903
PS2 : This pull request is also about it, however having built from origin/master don't seems to solve the issue
https://github.com/cyberbotics/webots/pull/2152

ShuffleWire:
If the controller died, it can't call the cleanup function, and when I restart it, I get a error message from webots because he think that the controller is still running, so he can't find a spot for the new incoming connection.

It simply remove the dead controller connection as usual, then recreate a new one. (a bit of code duplication from the line above, my bad... It might be possible to avoid that).
The trick is to use the name of the robot to know which dead controller should be kill. (which is fine for me). Otherwise, if no name is provided, the code execute as usual (scanning for free spot).

Maybe a more elegant solution is possible, by reassigning a socket id, for instance, but I'm first trying to get it working. I've noticed the robot_id, which maybe could be used for that, but I've not investigated in that direction.
It's have not been tested in every way possible, but I'm not seeing major issues coming.
The main thing to check would be leak of memory, unclosed socket and so on, because I'm not sure how is it performed.

I'm open to any advice,  it's my first PR here ! :)
This PR close #3000
It might close #1903 
#2152 is also about it, however having built from origin/master don't seems to solve the issue

PS : I'm not sure about merging to master. Is it on develop ?

### DIFF
--- a/src/webots/control/WbControlledWorld.cpp
+++ b/src/webots/control/WbControlledWorld.cpp
@@ -300,7 +300,7 @@ void WbControlledWorld::reset(bool restartControllers) {
 }
 
 void WbControlledWorld::checkIfReadRequestCompleted() {
-  assert(!mControllers.isEmpty());
+  //assert(!mControllers.isEmpty());
   if (!needToWait()) {
     WbSimulationState *state = WbSimulationState::instance();
     emit state->controllerReadRequestsCompleted();
--- a/src/webots/control/WbControlledWorld.cpp
+++ b/src/webots/control/WbControlledWorld.cpp
@@ -219,9 +219,22 @@ void WbControlledWorld::addControllerConnection() {
       robotName = buffer;
       delete[] buffer;
     }
+    int nbExternalController = 0;
     foreach (WbRobot *const robot, robots()) {
-      if (robot->isControllerStarted())
-        continue;
+      if (robot->controllerName() == "<extern>")
+	    nbExternalController++;
+    }
+    foreach (WbRobot *const robot, robots()) {
+      if (robot->isControllerStarted()) {
+        // if a specific robot have been targeted by giving a name, or only one external controller is in the scene (hence it's the target)
+        if (robot->controllerName() == "<extern>" && (robotName == robot->name() || nbExternalController == 1)) {
+          // automatically disconnect from previous extern controller that may have failed, the slot could be immediately used to restart
+          WbLog::info(tr("Closing extern controller connection for robot \"%1\".").arg(robot->name()));
+          updateRobotController(robot);
+        }
+        else // this controller seems to be in active use, ignore it
+          continue;
+      }
       if ((robotName == robot->name() || robotName.isEmpty()) && robot->controllerName() == "<extern>") {
         WbLog::info(tr("Starting extern controller for robot \"%1\".").arg(robot->name()));
         mRobotsWaitingExternController.append(robot);
--- a/src/webots/control/WbControlledWorld.cpp
+++ b/src/webots/control/WbControlledWorld.cpp
@@ -192,64 +192,71 @@ void WbControlledWorld::deleteController(WbController *controller) {
     step();
 }
 
+WbRobot *WbControlledWorld::findRobotToReconnect(QString robotName) {
+  int nbExternalController = 0;
+  WbRobot *matchingRobot = NULL;
+  foreach (WbRobot *const robot, robots()) {
+    if (robot->controllerName() == "<extern>") {
+      nbExternalController++;
+      if (robotName == robot->name()) {
+        // find first robot targeted with the given name
+        matchingRobot = robot;
+        break;
+      } else if ((!matchingRobot || matchingRobot->isControllerStarted()) && robotName.isEmpty() &&
+                  (nbExternalController == 1 || !robot->isControllerStarted()))
+        // if no robot name is specified and only one extern controller is used,
+        matchingRobot = robot;
+      }
+    }
+  return matchingRobot;
+}
+
+void readSocket(QLocalSocket *s, char *data, qint64 len) {
+  qint64 i = 0, rc;
+  while ((rc = s->read(&data[i], len - i)) >= 0) {
+    i += rc;
+    if (i >= len)
+      break;
+    s->waitForReadyRead();
+  }
+}
+
 void WbControlledWorld::addControllerConnection() {
   QLocalSocket *socket = mServer->nextPendingConnection();
   int robotId = 0;
-  int n, i = 0;
-  while ((n = socket->read((char *)&robotId, sizeof(robotId) - i)) != (int)sizeof(robotId) - i) {
-    i = n;
-    socket->waitForReadyRead();
-  }
-  QString robotName;
+  readSocket(socket, (char *)&robotId, sizeof(robotId));
   if (robotId == 0) {  // the Robot.name should be sent
+    QString robotName;
     int size = 0;
-    i = 0;
-    while ((n = socket->read((char *)&size, sizeof(size) - i)) != (int)sizeof(size) - i) {
-      i = n;
-      socket->waitForReadyRead();
-    }
+    readSocket(socket, (char *)&size, sizeof(size));
     if (size) {
       char *buffer = new char[size + 1];
-      i = 0;
-      while ((n = socket->read(&buffer[i], size - i)) != size - i) {
-        i = n;
-        socket->waitForReadyRead();
-      }
+      readSocket(socket, buffer, size);
       buffer[size] = '\0';
       robotName = buffer;
       delete[] buffer;
     }
-    int nbExternalController = 0;
-    foreach (WbRobot *const robot, robots()) {
-      if (robot->controllerName() == "<extern>")
-	    nbExternalController++;
-    }
-    foreach (WbRobot *const robot, robots()) {
-      if (robot->isControllerStarted()) {
-        // if a specific robot have been targeted by giving a name, or only one external controller is in the scene (hence it's the target)
-        if (robot->controllerName() == "<extern>" && (robotName == robot->name() || nbExternalController == 1)) {
-          // automatically disconnect from previous extern controller that may have failed, the slot could be immediately used to restart
-          WbLog::info(tr("Closing extern controller connection for robot \"%1\".").arg(robot->name()));
-          updateRobotController(robot);
-        }
-        else // this controller seems to be in active use, ignore it
-          continue;
+    if (WbRobot *matchingRobot = findRobotToReconnect(robotName)) {
+      if (matchingRobot->isControllerStarted()) {
+        // automatically disconnect from previous extern controller that may have failed,
+        // the slot could be immediately used to restart
+        WbLog::info(tr("Closing extern controller connection for robot \"%1\".").arg(matchingRobot->name()));
+        updateRobotController(matchingRobot);
       }
-      if ((robotName == robot->name() || robotName.isEmpty()) && robot->controllerName() == "<extern>") {
-        WbLog::info(tr("Starting extern controller for robot \"%1\".").arg(robot->name()));
-        mRobotsWaitingExternController.append(robot);
-        startControllerFromSocket(robot, socket);
-        return;
-      }
+      WbLog::info(tr("Starting extern controller for robot \"%1\".").arg(matchingRobot->name()));
+      mRobotsWaitingExternController.append(matchingRobot);
+      startControllerFromSocket(matchingRobot, socket);
     }
-    socket->close();
-    if (robotName.isEmpty())
-      WbLog::warning(tr("Failed to attach extern robot controller: no available \"<extern>\" robot controller found."), true);
-    else
-      WbLog::warning(
-        tr("Failed to attach extern robot controller: no available \"<extern>\" robot controller named \"%1\" found.")
-          .arg(robotName),
-        true);
+    else {
+      socket->close();
+      if (robotName.isEmpty())
+        WbLog::warning(tr("Failed to attach extern robot controller: no available \"<extern>\" robot controller found."), true);
+      else
+        WbLog::warning(
+          tr("Failed to attach extern robot controller: no available \"<extern>\" robot controller named \"%1\" found.")
+            .arg(robotName),
+          true);
+    }
     return;
   }
   foreach (WbController *const controller, mControllers) {
--- a/src/webots/control/WbControlledWorld.hpp
+++ b/src/webots/control/WbControlledWorld.hpp
@@ -60,7 +60,7 @@ private:
   void startControllerFromSocket(WbRobot *robot, QLocalSocket *socket);
   void updateRobotController(WbRobot *robot);
   void handleRobotRemoval(WbBaseNode *node);
-
+  WbRobot *findRobotToReconnect(QString robotName);
   QLocalServer *mServer;
   QList<WbController *> mControllers;
   QList<WbController *> mWaitingControllers;  // controllers inserted in previous step and waiting to be started in current step
--- a/docs/reference/changelog-r2021.md
+++ b/docs/reference/changelog-r2021.md
@@ -18,6 +18,7 @@ Released on XX Xth, 2021.
     - Fixed [Supervisor](supervisor.md) API operations on nodes defined in multiple nested PROTO fields ([#3036](https://github.com/cyberbotics/webots/pull/3036)).
     - Fixed crash applying [`wb_supervisor_node_add_force`](supervisor.md#wb_supervisor_node_add_force) on kinematic objects ([#3036](https://github.com/cyberbotics/webots/pull/3036)).
     - Fixed mecanum wheels [ContactProperties](contactproperties.md) in [YouBot](../guide/youbot.md) worlds ([#3025](https://github.com/cyberbotics/webots/pull/3025)).
+    - Fixed various issues when an extern [Robot](robot.md) controller exits without disconnecting and reconnects to the same [Robot](robot.md) ([#3005](https://github.com/cyberbotics/webots/pull/3005)).
     - Fixed [`wb_supervisor_node_reset_physics`](supervisor.md#wb_supervisor_node_reset_physics) not working if during the same step the node is also artificially moved with the [Supervisor](supervisor.md) API ([#2991](https://github.com/cyberbotics/webots/pull/2991)).
     - Fixed crash changing the [`Lidar.type`](lidar.md) field during the simulation run by requiring to save and reload the world ([#2983](https://github.com/cyberbotics/webots/pull/2983)).
     - Fixed value of the `verticalFieldOfView` for the [Hokuyo UTM-30LX](../guide/lidar-sensors.md#hokuyo-utm-30lx) ([#2972](https://github.com/cyberbotics/webots/pull/2972)).


src/webots/control/WbControlledWorld.cpp
```diff
@@ -220,6 +220,13 @@ void WbControlledWorld::addControllerConnection() {
...
     foreach (WbRobot *const robot, robots()) {
+      if (robotName == robot->name() && robot->controllerName() == "<extern>" && robot->isControllerStarted()) {
+        updateRobotController(robot);
+        WbLog::info(tr("Restarting extern controller connection for robot \"%1\".").arg(robot->name()));
+        mRobotsWaitingExternController.append(robot);
+        startControllerFromSocket(robot, socket);
+        return;
+      }
       if (robot->isControllerStarted())
         continue;
```
stefaniapedrazzi, line 231: It is ok to duplicate the two commands, but I would at least  try to limit the duplication of the tests and add some comments:
```suggestion
      if (robot->isControllerStarted()) {
        if (robotName == robot->name() && robot->controllerName() == "<extern>") {
          // automatically disconnect from previous extern controller that may have failed
          updateRobotController(robot);
          WbLog::info(tr("Restarting extern controller connection for robot \"%1\".").arg(robot->name()));
          mRobotsWaitingExternController.append(robot);
          startControllerFromSocket(robot, socket);
        } // else ignore running controllers
        return;
      }
```

stefaniapedrazzi (changes requested):
It is fine and it is correct to commit it  in the `master` branch.

Just note that this fix is working but only if the `WEBOTS_ROBOT_NAME` is set (this is not mandatory when using a single extern controller).
Otherwise if `WEBOTS_ROBOT_NAME` is not set, it is in any case not possible to correctly disconnect to the controller.
But I confirm that if `WEBOTS_ROBOT_NAME` both #3000 and #1903 are fixed.

To complete the fix is also needed to add an entry in the ChangeLog:
https://github.com/cyberbotics/webots/blob/master/docs/reference/changelog-r2021.md#webots-r2021a-revision-1

src/webots/control/WbControlledWorld.cpp
ShuffleWire: Thanks a lot for your feedback :) ! I just saw your reply when I was thinking about extending this to the case with only 1 robot in the world. In this case, webots could know the controller to reboot.
I will integrate your change in the meantime.

ShuffleWire:
I think I've a nice solution regarding code duplication !
I'm not huge fan of the "else continue;", I would instead duplicate and add the "robot->isControllerStarted()" on line 233, especially because "updateRobotController" actually set this flag to false. But that not a big deal :) 
It also integrate the special case with only 1 robot, so we don't need to specify name.
I've tested this with one robot, but not with multiples, however.

PS : I'm used to force-push new code when I've just "fixed" previous commit, in order to avoid bunch of fix commit. Is it fine for you ?
Moreover, I've mess up a bit here, hence the multiple "push", hope you don't get notification each time :/

PS2 : Test Suite Windows / build (windows-2019) (pull_request)  don't seems to pass, but as far as I know, I'm not the only one. Should I fix something ?

src/webots/control/WbControlledWorld.cpp
ShuffleWire: I think I've a nice solution regarding code duplication !
I'm not huge fan of the "else continue;", I would instead duplicate and add the "robot->isControllerStarted()" on line 233, especially because "updateRobotController" actually set this flag to false. But that not a big deal :)
It also integrate the special case with only 1 robot, so we don't need to specify name.
I've tested this with one robot, but not with multiples, however.

PS : I'm used to force-push new code when I've just "fixed" previous commit, in order to avoid bunch of fix commit. Is it fine for you ?
Moreover, I've mess up a bit here, hence the multiple "push", hope you don't get notification each time :/

PS2 : Test Suite Windows / build (windows-2019) (pull_request) don't seems to pass, but as far as I know, I'm not the only one. Should I fix something ?

src/webots/control/WbControlledWorld.cpp
```diff
@@ -220,8 +220,16 @@ void WbControlledWorld::addControllerConnection() {
       delete[] buffer;
     }
     foreach (WbRobot *const robot, robots()) {
-      if (robot->isControllerStarted())
-        continue;
+      if (robot->isControllerStarted()) {
+        // if a specific robot have been targeted by giving a name, or only one robot is in the scene (hence it's the target)
+        if ((robotName == robot->name() || robots().size() == 1) && robot->controllerName() == "<extern>") {
+          // automatically disconnect from previous extern controller that may have failed, the slot could be immedialy used to restart
```
stefaniapedrazzi, line 226:
```suggestion
          // automatically disconnect from previous extern controller that may have failed, the slot could be immediatly used to restart
```

stefaniapedrazzi (changes requested):
Again the fix works only in two particular cases:
* only one robot in the simulation or
* the `WEBOTS_ROBOT_NAME` for the extern controller is set

but it still doesn't work if the simulation contains multiple robots but only one is using extern controller.
So rather than checking for the number of robots only, it would be better to check for the number of extern controllers so that the fix works in all the cases.

This could be achieved by using the for loop just to identify the robot whose controller needs to be started and count the number of external controller.
Disconnect and controller start can be applied just after the loop.

src/webots/control/WbControlledWorld.cpp
stefaniapedrazzi: > PS : I'm used to force-push new code when I've just "fixed" previous commit, in order to avoid bunch of fix commit. Is it fine for you ?
Moreover, I've mess up a bit here, hence the multiple "push", hope you don't get notification each time :/

Given that the branch will be squashed and merged we don't care too much about the number of commits.

> PS2 : Test Suite Windows / build (windows-2019) (pull_request) don't seems to pass, but as far as I know, I'm not the only one. Should I fix something ?

The Windows tests is temporary broken due to some changes that we are not responsible for, so it is normal that it fails.

src/webots/control/WbControlledWorld.cpp
ShuffleWire: And what if there is only one external controller, and the name of the robot doesn't match ? Should not happen often, cause in this case user might better not provide any name, but still...

ShuffleWire: I've still made the change, that a nice idea

stefaniapedrazzi: > And what if there is only one external controller, and the name of the robot doesn't match ? Should not happen often, because in this case user might better not provide any name, but still...

It should work exactly as for the first start: if the `robotName` is set but differs from `robot->name()` it is not disconnected.

ShuffleWire: BTW, I've tried to used this sort of construct, which should appropriate : 
`int nbExternalController = std::count_if(robots().begin(), robots().end(), [](WbRobot *const r) { return r->controllerName() == "<extern>"; });`
But it actually cause segfault :/ not sure what happened, I'm not use to C++ that much...

src/webots/control/WbControlledWorld.cpp
```diff
@@ -219,9 +219,22 @@ void WbControlledWorld::addControllerConnection() {
...
+      if (robot->isControllerStarted()) {
+        // if a specific robot have been targeted by giving a name, or only one robot is in the scene (hence it's the target)
+        if (robot->controllerName() == "<extern>" && (robotName == robot->name() || nbExternalController == 1)) {
+          // automatically disconnect from previous extern controller that may have failed, the slot could be immediatly used to restart
+          WbLog::info(tr("Closing extern controller connection for robot \"%1\".").arg(robot->name()));
+          updateRobotController(robot);
+        }
+        else // this controller seems to be in active use, ignore it
+          continue;
+      }
```
stefaniapedrazzi, line 237: The two for-loops can be merged into a single one so that if it is extern and the robot name matches the loop can exit immediately.

ShuffleWire: To be fair, I'm confused : 
As I understand it, you advice to move lignes 224-225
```
      if (robot->controllerName() == "<extern>")
	    nbExternalController++;
```
inside and at the start of the foreach starting at 227
I'm right ? 
However, let's consider the following case : 
2 robots (A and B in this order) with external controllers (resp c_A, c_B), both started, and robotName == "B" (user want to reconnect c_B)
On the first loop, we would get : on line 230 : `nbExternalController == 1` (because the `if (robot->controllerName() == "<extern>")` check have been placed at the beginning of the loop)
This will allows the `if` to trigger the reconnect of c_B on robot A (and kicking c_A), which is not what we want.
This is why we need to know in advance how many external we have, to be able to trigger this special case of only one external controller (ie`nbExternalController == 1` ) correctly.
I do think that two separate loops are the way to go. Did I get it right ?

stefaniapedrazzi: What I mean is, in the for loop, count the extern controllers and save the WbRobot instance that matches so that disconnect and restart can be applied just after the loop:
* if `WEBOTS_ROBOT_NAME` is set, you don't need to count the extern controller, you can store the current WbRobot instance and break immediately the loop
* if `WEBOTS_ROBOT_NAME` is empty you store the WbRobot instance of the first started extern controller but continue within the loop
After the loop is over you can use the stored WbRobot instance to apply the disconnect/controller start.

ShuffleWire: Well, that not went that well... I've been trying to implement what you was explaining, but even if I succeeded, I ended up with wacky code and strange construct.
If I find some time tomorrow I will try to think a bit more about it.

ShuffleWire: I must say that I've not "clean" idea...
I would be interesting to see a nicer solution, but for the moment I prefer just request review again, to not stale the PR.

src/webots/control/WbControlledWorld.cpp
```diff
@@ -219,9 +219,22 @@ void WbControlledWorld::addControllerConnection() {
...
     foreach (WbRobot *const robot, robots()) {
-      if (robot->isControllerStarted())
-        continue;
+      if (robot->controllerName() == "<extern>")
+	    nbExternalController++;
+    }
+    foreach (WbRobot *const robot, robots()) {
+      if (robot->isControllerStarted()) {
+        // if a specific robot have been targeted by giving a name, or only one external controller is in the scene (hence it's the target)
+        if (robot->controllerName() == "<extern>" && (robotName == robot->name() || nbExternalController == 1)) {
```
stefaniapedrazzi, line 230: This test is wrong because if there is only one controller but the robotName doesn't match, the controller should not be disconnected.
This is important as we want to have the exact same behavior as at first connect.
```suggestion
        if (robot->controllerName() == "<extern>" && (robotName == robot->name() || (robotName.isEmpty() && nbExternalController == 1))) {
```

stefaniapedrazzi:
Here is what I meant with a single loop through the robot list:

```
    int nbExternalController = 0;
    const int robotsSize = robots().size();
    WbRobot *matchingRobot = NULL;
    for (int i = 0; i < robotsSize; i++) {
      WbRobot *robot = robots().at(i);
      if (robot->controllerName() == "<extern>") {
        nbExternalController++;
        if (robotName == robot->name()) {
          // find first robot targeted with the given name
          matchingRobot = robot;
          break;
        } else if ((!matchingRobot || matchingRobot->isControllerStarted()) && robotName.isEmpty() &&
                   (nbExternalController == 1 || !robot->isControllerStarted()))
          // if no robot name is specified and only one extern controller is used,
          matchingRobot = robot;
      }
    }
    if (matchingRobot) {
      if (matchingRobot->isControllerStarted() && (nbExternalController == 1 || robotName == matchingRobot->name())) {
        // automatically disconnect from previous extern controller that may have failed, the slot could be immedialy used to
        // restart
        WbLog::info(tr("Closing extern controller connection for robot \"%1\".").arg(matchingRobot->name()));
        updateRobotController(matchingRobot);
      }
      if (!matchingRobot->isControllerStarted()) {
        WbLog::info(tr("Starting extern controller for robot \"%1\".").arg(matchingRobot->name()));
        mRobotsWaitingExternController.append(matchingRobot);
        startControllerFromSocket(matchingRobot, socket);
        return;
      }
    }
```

stefaniapedrazzi (changes requested):
The disconnect behavior should be exactly the same as the first controller start.

ShuffleWire:
First, sorry for the lag, been busy and having trouble to find nice solution !
Second, I've integrate your solution, but I'm not huge fan of it, because the main `if` condition doesn't allows to guess immediately what the code is trying to achieve (IMO).
That why I've split it into smaller function, which could be tested more easily, compensating for the difficult `if` construct.
Moreover, the readSocket function that I've created could be use elsewhere. Additionally, in this one, I've note made (yet?) a clean error check, but neither the original code does.
I've tried to check in QT docs, I've not found a similar function in the QLocalSocket class, which is rather surprising :/ If I just missed it, than It would be better to simply use it.

Moreover, I'd to remove an assert in checkIfReadRequestCompleted, which was causing me trouble, but that maybe just me... What is his role ? it was introduce in 785ece7cef9d322af09cca786bb66f4144595e56, when moving code, but the original code doesn't use it.

ShuffleWire:
Side question, when I'm pushing, I always end up when commit from others... what is causing that ?

stefaniapedrazzi:
> Side question, when I'm pushing, I always end up when commit from others... what is causing that ?

It seems that your local branch is not correctly synchronized with `origin/master`.
Maybe you merged it with other branches before pushing.
Rebasing and overwriting commits that have already been pushed to origin could also be the cause of this.

I will now check the latest changes.
However, currently this PR is not in a good shape and cannot be merged as-is.
You should revert all the unrelated changes or create a new fresh PR containing only the changes related to the extern controller connection fix.

ShuffleWire:
Got it !
I might delay a bit update on this PR, cause I'm rather busy, but I'm still keeping an eye on it.

src/webots/control/WbControlledWorld.cpp
```diff
@@ -192,51 +192,71 @@ void WbControlledWorld::deleteController(WbController *controller) {
...
-        startControllerFromSocket(robot, socket);
-        return;
+    if (WbRobot *matchingRobot = findRobotToReconnect(robotName)) {
+      if (matchingRobot->isControllerStarted()) {
+        // automatically disconnect from previous extern controller that may have failed,
+        // the slot could be immediately used to restart
+        WbLog::info(tr("Closing extern controller connection for robot \"%1\".").arg(matchingRobot->name()));
+        updateRobotController(matchingRobot);
       }
+      WbLog::info(tr("Starting extern controller for robot \"%1\".").arg(matchingRobot->name()));
```
stefaniapedrazzi, line 246: here it seems that the check `!matchingRobot->isControllerStarted()` is now missing.

stefaniapedrazzi (changes requested):
Here my comments on the latest version.
It seems that some important checks have been lost while refactoring the code.

Moreover all the changes related to the `readSocket` function are not needed to fix this issue and would be better to remove them and restore the original version.

Please make sure that all the unrelated changes have been removed before requesting again a review.

src/webots/control/WbControlledWorld.cpp
```diff
@@ -280,7 +300,7 @@ void WbControlledWorld::reset(bool restartControllers) {
 }
 
 void WbControlledWorld::checkIfReadRequestCompleted() {
-  assert(!mControllers.isEmpty());
+  //assert(!mControllers.isEmpty());
```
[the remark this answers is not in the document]
ShuffleWire: yes, this one that rather strange, and not meant to be merge, cf my comment above : 
> Moreover, I'd to remove an assert in checkIfReadRequestCompleted, which was causing me trouble, but that maybe just 
> me... What is his role ? it was introduce in 785ece7, when moving code, but the original code doesn't use it.

src/webots/control/WbControlledWorld.cpp
```diff
@@ -192,51 +192,71 @@ void WbControlledWorld::deleteController(WbController *controller) {
...
+                  (nbExternalController == 1 || !robot->isControllerStarted()))
+        // if no robot name is specified and only one extern controller is used,
+        matchingRobot = robot;
+      }
+    }
+  return matchingRobot;
+}
+
+void readSocket(QLocalSocket *s, char *data, qint64 len) {
+  qint64 i = 0, rc;
```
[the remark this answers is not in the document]
ShuffleWire: Return Code, but I could use any thing else, not a issue

src/webots/control/WbControlledWorld.cpp
```diff
@@ -192,51 +192,71 @@ void WbControlledWorld::deleteController(WbController *controller) {
     step();
 }
 
+WbRobot *WbControlledWorld::findRobotToReconnect(QString robotName) {
+  int nbExternalController = 0;
+  WbRobot *matchingRobot = NULL;
+  foreach (WbRobot *const robot, robots()) {
```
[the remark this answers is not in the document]
ShuffleWire: didn't know that.
however, this isn't a time critical loop, which won't be called often.
but for the sake of consistency, I will use for loops.

src/webots/control/WbControlledWorld.cpp
```diff
@@ -192,51 +192,71 @@ void WbControlledWorld::deleteController(WbController *controller) {
     step();
 }
 
+WbRobot *WbControlledWorld::findRobotToReconnect(QString robotName) {
```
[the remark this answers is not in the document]
ShuffleWire: yep, that better

src/webots/control/WbControlledWorld.cpp
```diff
@@ -192,64 +192,71 @@ void WbControlledWorld::deleteController(WbController *controller) {
...
-        }
-        else // this controller seems to be in active use, ignore it
-          continue;
-      }
-      if ((robotName == robot->name() || robotName.isEmpty()) && robot->controllerName() == "<extern>") {
-        WbLog::info(tr("Starting extern controller for robot \"%1\".").arg(robot->name()));
-        mRobotsWaitingExternController.append(robot);
-        startControllerFromSocket(robot, socket);
-        return;
+    if (WbRobot *matchingRobot = findRobotToReconnect(robotName)) {
```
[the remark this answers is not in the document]
ShuffleWire: my idea was to allows way simpler testing of this function in particular, because it actually make non trivial work, and ensuring that the right robot is chosen might justify the use of some tests case.
Additionally, the overhead is rather negligible, this function will not be called often (one times for each robot re-connection, which not occur often)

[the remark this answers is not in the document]
ShuffleWire: ok!

src/webots/control/WbControlledWorld.cpp
```diff
@@ -192,51 +192,71 @@ void WbControlledWorld::deleteController(WbController *controller) {
...
-    foreach (WbRobot *const robot, robots()) {
-      if (robot->isControllerStarted())
-        continue;
-      if ((robotName == robot->name() || robotName.isEmpty()) && robot->controllerName() == "<extern>") {
-        WbLog::info(tr("Starting extern controller for robot \"%1\".").arg(robot->name()));
-        mRobotsWaitingExternController.append(robot);
-        startControllerFromSocket(robot, socket);
-        return;
+    if (WbRobot *matchingRobot = findRobotToReconnect(robotName)) {
+      if (matchingRobot->isControllerStarted()) {
```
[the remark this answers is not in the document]
ShuffleWire: I thought I've moved them in `findRobotToReconnect`, isn't ?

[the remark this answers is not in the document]
ShuffleWire: I thought I've moved them in `findRobotToReconnect`, isn't ?

src/webots/control/WbControlledWorld.cpp
ShuffleWire: I thought I've moved them in `findRobotToReconnect`, isn't ?

src/webots/control/WbControlledWorld.cpp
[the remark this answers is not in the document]
stefaniapedrazzi: The checks in `findRobotReconnet` are not enough (and where already there before).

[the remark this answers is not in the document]
ShuffleWire: ok

src/webots/control/WbControlledWorld.hpp
```diff
@@ -60,7 +60,7 @@ public slots:
   void startControllerFromSocket(WbRobot *robot, QLocalSocket *socket);
   void updateRobotController(WbRobot *robot);
   void handleRobotRemoval(WbBaseNode *node);
-
+  WbRobot *findRobotToReconnect(QString robotName);
```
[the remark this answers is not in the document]
ShuffleWire: agreed

stefaniapedrazzi:
Version at commit 3757672 was fine apart from the fix to the if condition I reported.
I don't like so much to have the two loops, but given that this code is not called too often it is acceptable.
But honestly it seems a waste of time to continuously refactoring the code of this quite simple fix.

stefaniapedrazzi:
Replaced by #3189.

PR #3006, opened 2021-04-23
Add two missing package to build on Debian 11

ShuffleWire:
Building on Debian miss some instructions.
I've built Webots from source on a clean install of Debian 11, and I needed to issue some additionnal commands.

**Solution**
Got the followings messages during make :

`/usr/bin/ld: cannot find -lusb`
solution : 
`sudo apt install libusb-dev`

`Can't find Python.h`
solution : 
`sudo apt install python3-dev`

I suggest adding both those packages in `sudo scripts/install/linux_compilation_dependencies.sh`

`can't find execstack`
solution:
didn't need to make anything, it seems to be working fine without it. Any idea ?

### DIFF
--- a/scripts/install/linux_compilation_dependencies.sh
+++ b/scripts/install/linux_compilation_dependencies.sh
@@ -6,7 +6,7 @@ if [[ $EUID -ne 0 ]]; then
 fi
 
 apt update
-apt install --yes git lsb-release cmake swig python2.7-dev libglu1-mesa-dev libglib2.0-dev libfreeimage-dev libfreetype6-dev libxml2-dev libzzip-0-13 libboost-dev libgd3 libssh-gcrypt-dev libzip-dev libreadline-dev pbzip2 libpci-dev wget libssl-dev zip unzip
+apt install --yes git lsb-release cmake swig python2.7-dev python3-dev libusb-dev libglu1-mesa-dev libglib2.0-dev libfreeimage-dev libfreetype6-dev libxml2-dev libzzip-0-13 libboost-dev libgd3 libssh-gcrypt-dev libzip-dev libreadline-dev pbzip2 libpci-dev wget libssl-dev zip unzip
 
 UBUNTU_VERSION=$(lsb_release -rs)
 if [[ $UBUNTU_VERSION == "16.04"  || $UBUNTU_VERSION == "18.04" ]]; then


stefaniapedrazzi:
These libraries are already installed by the Webots installation system and it is not necessary to duplicate them in the `linux_compilation_dependencies.sh`.

`libusb-dev` should already be installed by the `linux_compilation_dependencies.sh` scritps because it is included in the  `linux_runtime_dependencies.sh` script that is called just at the end.
So it doesn't seem necessary to add it also in the `linux_compilation_dependencies.sh`.
If you have issues with the `libusb-dev`, I would suggest you to also inspect if the `linux_runtime_dependencies.sh` script is called correctly.

The Python dev packages are installed in the `linux_optional_compilation_dependencies.sh` script and Webots can be compiled even without it. They are just needed if you want to use the Python API.

ShuffleWire:
Good news, I've got explanations !

**First part : `libusb-dev`**

The `libusb-dev` is indeed in `linux_runtime_dependencies.sh`, but the issue is that the `execstack` does not exist in the regular Debian 11 repos, and thus the whole `apt install` command failed.
To be sure, I've install a new Debian 11 on my computer on a separate partition, then :
```
apt update
apt upgrade -y
apt install git
git clone --recurse-submodules -j8 https://github.com/cyberbotics/webots.git
cd webots
sudo scripts/install/linux_compilation_dependencies.sh
```
Here I got 2 things : 
 -> `linux_runtime_dependencies.sh` yield : `Unable to find package execstack`
Which abort the whole apt command
 -> the If/else construct in `linux_compilation_dependencies.sh` and `linux_runtime_dependencies.sh` both yield `Unsupported Linux version` which is worrying, even if the build actually work on it :)

```
cat scripts/install/bashrc.linux >> ~/.bashrc
source ~/.bashrc
```
(I was cloning in my home folder, but maybe add in the https://github.com/cyberbotics/webots/wiki/Linux-installation that 
`Check the content of scripts/install/bashrc.linux ` first of all means "change the home directory path" ? New webots' users like me would be aware of the thing they should change when trying to build for the first time, or advice the beginning to clone webots in the home directory)

```
make
```
which trigger `ld: cannot find -lusb`

The solution is to remove the execstack package from the list, allowing to install the necessary package, and doesn't seems necessary to build/run webots. (Does webots actually need it ?)

**Second part : 'python3.9-dev'**

I've check just after the install of my Debian, but it seems that python3 (python3.9) it embedded with the regular installer.
Thu, if I understand correctly, because I've have it, webots is trying to build the python3.9 API, but could not find the development files, being in the optional script, which I've not install for a first build.

In fact, executing the `linux_optional_compilation_dependencies.sh` doesn't work neither, it install bunch of package, but the
`apt install ... python3.9-dev ...` failed because of `python3.6/7/8` could not be found. otherwise it may have work.

The solution for me was to install manually to `python3.9-dev` package manually, which allows me to build and run (hooray! :) )
It's not a big deal, and seems logical once I've experienced it, that having `python3.9` would trigger the installation of the `python3.9-dev`,
but in the other hand, I wish that building could work out-of-the-box, at least for a limited number of features.
Or explicitly state that in the building steps, that the user should himself take care of installing the right version.

I get that on Debian we would be missing stuff, and errors messages from apt from the optional script allows me to say that there is much more work to port fully webots to Debian (ROS etc) for instances :

```
Ign:4 http://ppa.launchpad.net/deadsnakes/ppa/ubuntu impish InRelease                                                                      
Err:5 http://ppa.launchpad.net/deadsnakes/ppa/ubuntu impish Release                                                  
  404  Not Found [IP: 2001:67c:1560:8008::19 80]
Ign:6 http://packages.ros.org/ros/ubuntu bullseye InRelease
Err:7 http://packages.ros.org/ros/ubuntu bullseye Release
  404  Not Found [IP: 2600:3402:200:227::2 80]
Reading package lists... Done
E: The repository 'http://ppa.launchpad.net/deadsnakes/ppa/ubuntu impish Release' does not have a Release file.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
E: The repository 'http://packages.ros.org/ros/ubuntu bullseye Release' does not have a Release file.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
```

For peace of mind and after all of it, I've removed python3.9 and python3.9-dev (which could cause some mess in the regular Debian install, but still) And I was able to build nicely :) Of course I could not use python script, but that fine in this case, obviously !

PS : 
The fact to not be able to build easily by following step by step https://github.com/cyberbotics/webots/wiki/Linux-installation once could be a deal breaker for people wanting to contribute :)
After that step, it's easier to go deeper in it.

stefaniapedrazzi:
Thank you for you feedback!

We try to be compatible with as many Linux distribution as possible.
However, we only support and maintain instructions for the two latest Ubuntu LTS releases.
I will make sure that this is clearly stated in the installation documentation and that users using other Linux distribution will be aware that some modifications are needed.

I will also double check where is the `execstack` library needed.

stefaniapedrazzi:
Installation instructions improved in #3030.

PR #3010, opened 2021-04-26
Add MiR100 robot

mael25:
Add [Mobile Industrial Robots' MiR100](https://www.mobile-industrial-robots.com/en/solutions/robots/mir100/) robot.

.

**Tasks**
  - [x] Fix  wheels
  - [x] Doc / Icon
  - [x] Add sensors
  - [x] Correct controller

MiR100 User guide (pdf): [https://www.mobile-industrial-robots.com/media/11446/mir100-user-guide_31_en.pdf](https://www.mobile-industrial-robots.com/media/11446/mir100-user-guide_31_en.pdf)


projects/robots/mir/mir100/protos/Mir100.proto
```diff
@@ -0,0 +1,668 @@
...
+        name "us_2_frame"
+      }
+      DEF LEFT_WHEEL_JOINT HingeJoint {
+        jointParameters HingeJointParameters {
+          axis 0 1 0
+          anchor 0 0.222604 0.0625
+        }
+        device [
+          RotationalMotor {
+            name "mir100_middle_left_wheel_joint"
```
BenjaminDeleze, line 130: Other robots don't have their name as prefix in names. You should just name them `middle_left_wheel_joint`, `middle_left_wheel_joint_sensor`, ...

projects/robots/mir/mir100/controllers/keyboard_control/Makefile
```diff
@@ -0,0 +1,72 @@
...
+### if you want to use the Webots C API in your C++ controller program:
+### USE_C_API = true
+###
+### ---- Debug mode ----
+### if you want to display the gcc command line for compilation and link, as
+### well as the rm command details used for cleaning:
+### VERBOSE = 1
+###
+###-----------------------------------------------------------------------------
+
```
BenjaminDeleze, line 67:
```suggestion
```

BenjaminDeleze (changes requested):
Thanks! Here is a first round remarks.

projects/robots/mir/mir100/controllers/keyboard_control/Makefile
ad-daniel: Normally for robot controllers a shortened Makefile header is used, which typically just mentions copyright dates & license. You can copy one from another robot, like khepera3 for example.

BenjaminDeleze (changes requested):
I think this PR should wait on the Summit-XL PR, such that you could add the missing lidars.

Moreover: 

- I found out on pictures of the MiR100 that it seems to have several sensors both on front, on top and on back. However they seems to be absent from the model or not graphically visible.


- Both logos seems to have triangle problem with the R. I can add pictures if needed.
- You should try to make the logo more visible. One good technic would be to use Blender to add a ambiant occlusion texture to the logo.

projects/robots/mir/mir100/controllers/keyboard_control/keyboard_control.c
```diff
@@ -0,0 +1,112 @@
...
+          target_speed = target_speed + SPEED_INCREMENT;
+          target_speed = target_speed > SPEED_MAX ? SPEED_MAX : target_speed;
+          break;
+
+        case WB_KEYBOARD_DOWN:
+          target_speed = target_speed - SPEED_INCREMENT;
+          target_speed = target_speed < SPEED_MIN ? SPEED_MIN : target_speed;
+          break;
+
+        case WB_KEYBOARD_PAGEUP:
```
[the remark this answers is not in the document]
mael25: Yes, it was to stay consistent with `YouBot` for instance and the new mecanum wheels robot where left/right is already for moving sideways...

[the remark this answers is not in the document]
mael25: But it may indeed make more sense to do like other differential drive robots

mael25:
@BenjaminDeleze I have updated the controller and put additional sensors (4 small range ultrasonic sensors as well as two depth camera implemented by a unique `RangeFinder` at the front of the robot).

BenjaminDeleze (changes requested):
The new mesh looks really nice!

I still have some remarks:

- you need to add a .gitignore in the `keyboard_control` folder so that the binary is not pushed on github.
- In `mir.x3d`, the link to the background (12 links) has changed with the new online documentation, they now point directly on the ressources on github, you need to either regenerate the .x3d file or you can copy the ` <background>...</background> ` from another robot model.
- I got some errors when launching the `mir100.wbt`
  - `ERROR: '/home/benjamin/webots/projects/robots/mir/mir100/protos/Mir100.proto':164:9: error: Skipped unknown 'lookupTable' parameter in Mir100UltrasonicSensor PROTO.` four times (once for each ultrasonic sensor)
  - `WARNING: Mir100 (PROTO) > Robotino3Platform (PROTO) > ConveyorPlatform (PROTO): The controller directory has not been found, searched the following locations`

mael25:
Thanks for the review, I have made the changes, the new documentation viewer looks great!

BenjaminDeleze (changes requested):
Thanks, it seems like all the previous errors have been fixed, however I had some problems when testing the sensors:
-Accelerometer: `WARNING: Mir100 (PROTO) > DEF IMU_LINK Solid > Accelerometer: Parent of Accelerometer node has no physics: measurements may be wrong.`
-Gyro: `WARNING: Mir100 (PROTO) > DEF IMU_LINK Solid > Gyro: this node or its parents requires a 'physics' field to be functional.`

I thought at first that the distance sensors on front where not working but it seems that the problem come from their orientation (as they do not point towards, they will not detect an obstacle in front of the robot), but I think it is fine if it is like that on the real robot.

A small detail, but is it intentional that the reset key for both the robot and its conveyor belt are the same such that when you reset one it reset also the other?

BenjaminDeleze (approved):
Thank you!

PR #3013, opened 2021-04-27
Avoid duplicate device names in the exported URDF

lukicdarkoo:
**Description**
In case there was a device and solid with the same name there used to be a conflict in the generated URDF. In general, we should make sure there is no device and solid with the same name, but this PR makes the URDF exporter more robust to these cases.

### DIFF
--- a/src/webots/nodes/WbSolidDevice.cpp
+++ b/src/webots/nodes/WbSolidDevice.cpp
@@ -58,7 +58,3 @@ QString WbSolidDevice::computeShortUniqueName() const {
     return QString("%1:%2").arg(robot->computeUniqueName()).arg(WbSolid::name());
   return computeUniqueName();
 }
-
-const QString WbSolidDevice::urdfName() const {
-  return getUrdfPrefix() + name();
-}
--- a/src/webots/nodes/WbSolidDevice.hpp
+++ b/src/webots/nodes/WbSolidDevice.hpp
@@ -47,8 +47,6 @@ protected:
   void subscribeToRaysUpdate(dGeomID ray);
   virtual void updateRaysSetupIfNeeded() {}
 
-  const QString urdfName() const override;
-
 private:
   static QList<QPair<WbSolidDevice *, dGeomID>> cDirtySensors;
 };


ad-daniel:
For coupled motors (https://github.com/cyberbotics/webots/pull/2939), motors having the same name that belong to the same robot will be considered as siblings so giving commands to one of them will relay the command to all others. Does this fact of having multiple devices (motors specifically) with the same name will conflict with this?

lukicdarkoo:
> Does this fact of having multiple devices (motors specifically) with the same name will conflict with this?

No, it will just add a suffix to duplicates.

> EDIT: the title mentions duplicated device names, and in the message you mention device and solid sharing the same name. It's both or just the latter?

It is for both, but it essentially tries to solve the generated URDF for the e-puck PROTO in which solids and devices have the same name.

mael25:
I have checked and it works well for me!

ad-daniel:
I have tested with this Robot:
```
Robot {
  children [
    HingeJoint {
      device [
        Brake {
          name "UNIQUE_NAME"
        }
        PositionSensor {
          name "UNIQUE_NAME"
        }
        RotationalMotor {
          name "UNIQUE_NAME"
        }
      ]
      endPoint Solid {
        children [
          SliderJoint {
            device [
              LinearMotor {
                name "UNIQUE_NAME"
              }
            ]
            endPoint Solid {
              name "UNIQUE_NAME"
            }
          }
        ]
        name "UNIQUE_NAME"
      }
    }
  ]
}
```
and this was the result:
```
<?xml version="1.0"?>
<robot name="/home/daniel/webots_develop/Robot.urdf" xmlns:xacro="http://ros.org/wiki/xacro">
  <link name="base_link">
  </link>
  <joint name="UNIQUE_NAME" type="continuous">
    <parent link="base_link"/>
    <child link="UNIQUE_NAME"/>
    <axis xyz="1 0 0"/>
    <limit effort="10" velocity="10"/>
    <origin xyz="0 0 0" rpy="0 0 0"/>
  </joint>
  <link name="UNIQUE_NAME">
  </link>
  <joint name="UNIQUE_NAME" type="prismatic">
    <parent link="UNIQUE_NAME"/>
    <child link="UNIQUE_NAME_0"/>
    <axis xyz="0 0 1"/>
    <origin xyz="0 0 0" rpy="0 0 0"/>
    <limit effort="10" velocity="10"/>
  </joint>
  <link name="UNIQUE_NAME_0">
  </link>
</robot>
```
For the inner slider joint it's adding the suffix, but for the top layer shouldn't the link name and the joint_name differ? Or have I misunderstood where the name conflict was?

lukicdarkoo:
> For the inner slider joint it's adding the suffix, but for the top layer shouldn't the link name and the joint_name differ? Or have I misunderstood where the name conflict was?

Joint and link names are allowed to be the same in URDF. However, two link nodes are not. So, if you have `Link (UNIQUE_NAME) > Joint (UNIQUE_NAME) > Link (UNIQUE_NAME)` then the last link should receive a prefix.

In particular, with this change, the URDF generated from the e-puck is now valid. In the e-puck PROTO, there is e.g. `Solid (ps0) > Device (ps0)` which translates to `Link (ps0) > Link (ps0)` which is invalid. Keep in mind that handling duplicate names will always generate some strange side-effects in URDF, so PROTO designers should be careful when naming links and joints.

Issue #2996, opened 2021-04-22
Fail to load generic window library (MATLAB)


mael25:
**Describe the Bug**
The MATLAB controllers crash in the develop branch, because they fail to load the generic window library (see Screenshots).

**Steps to Reproduce**
1. Open and run `nao_matlab.wbt `or `youbot_matlab.wbt`
2. See error

**Expected behavior**
The library loads correctly and the controller does not crash.

**Screenshots**
![Screenshot from 2021-04-22 10-53-50](https://user-images.githubusercontent.com/38250944/115686946-fdc69980-a359-11eb-84e7-df140c4aec9d.png)

**System**
 - Operating System: Linux Ubuntu 20.04.2 LTS
 - Graphics Card: NVIDIA GeForce GTX 1650 Mobile / Max-Q

**Additional context**
I think the issue might be due to the fact that the files of the new robot window in `include/controller/c/webots/plugins/robot_window` are not included in `lib/controller/matlab/allincludes.h` and not mentioned in `lib/controller/matlab/launcher.m`.

mael25:
Hello @stefaniapedrazzi, so I have checked in more details where the error happens:

1. `lib/controller/matlab/launcher.m` (l.150) 
- calllib('libController', 'wb_robot_init');
2. `projects/robots/kuka/youbot/controllers/youbot_matlab/youbot_matlab.m` (l.14)
- passive_wait(4.0)
3. `projects/robots/kuka/youbot/controllers/youbot_matlab/passive_wait.m` (l.4)
- if (wb_robot_step(32)==-1)
4. `resources/projects/plugins/robot_windows/generic/generic.c` (l.16): 
- if(!wbu_generic_robot_window_handle_messages(message)) 
5. At this place, the error (see above) appears.

So basically, the error occurs the first time `wb_robot_step()`  is called inside the Matlab controller. Then, it looks in `libgeneric.so ` (generated from `resources/projects/plugins/robot_windows/generic/generic.c`) for `wb_robot_window_step()`. Finally, inside `wb_robot_window_step()`, it fails to find `libgeneric_robot_window.so` (generated from `resources/projects/libraries/generic_robot_window/generic.c`) necessary to use `wbu_generic_robot_window_handle_messages()` and the other generic window functions.

I guess it might be because it looks inside the wrong place for `libgeneric_robot_window.so`, but I have trouble to find where/how to define the search location of the library... Do you have any idea ?
Thanks!

[...]

PR #3021, opened 2021-04-29
Fix crash Matlab controller

mael25:
**Description**
The MATLAB controllers crash in the _develop_ branch, because they fail to load `lib/controller/libgeneric_robot_window.so`  from `resources/projects/plugins/robot_windows/generic/libgeneric.so`.

**Related Issues**
This pull-request fixes issue #2996.

**Tasks**

- [x] Compatibility MacOS
- [x] Compatibility Windows

**Additional context**
This has been solved by adding `lib/controller` in the rpath of `libgeneric.so`.

### DIFF
--- a/resources/projects/plugins/robot_windows/generic/Makefile
+++ b/resources/projects/plugins/robot_windows/generic/Makefile
@@ -23,8 +23,8 @@ LIBRARIES = -lgeneric_robot_window
 null :=
 space := $(null) $(null)
 WEBOTS_HOME_PATH=$(subst $(space),\ ,$(strip $(subst \,/,$(WEBOTS_HOME))))
-
 include $(WEBOTS_HOME_PATH)/resources/Makefile.os.include
+
 ifeq ($(OSTYPE),linux)
  LIBRARIES += -Wl,-rpath,$(WEBOTS_HOME_PATH)/lib/controller
 endif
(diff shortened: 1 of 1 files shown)


mael25:
Note: I had to add the lines after `include $(WEBOTS_HOME_PATH)/resources/Makefile.include`, otherwise the `OSTYPE` was not set yet to `linux`.

stefaniapedrazzi:
> Note: I had to add the lines after include $(WEBOTS_HOME_PATH)/resources/Makefile.include, otherwise the OSTYPE was not set yet to linux.

It seems wrong to set it after including `Makefile.include` and it would be better to set `LIBRARY` before.

You can retrieve the OSTYPE in the Makefile even before the include.
See for example: https://github.com/cyberbotics/webots/blob/master/projects/samples/devices/controllers/lidar/Makefile

stefaniapedrazzi:
We should probably also check that it works properly on Windows and macOS.

mael25:
> > Note: I had to add the lines after include $(WEBOTS_HOME_PATH)/resources/Makefile.include, otherwise the OSTYPE was not set yet to linux.
> 
> It seems wrong to set it after including `Makefile.include` and it would be better to set `LIBRARY` before.
> 
> You can retrieve the OSTYPE in the Makefile even before the include.
> See for example: https://github.com/cyberbotics/webots/blob/master/projects/samples/devices/controllers/lidar/Makefile

Right, in fact only `resources/Makefile.os.include` is needed before, I have updated the code.
I will also test on Windows afterwards.

stefaniapedrazzi (approved):
Thank you

stefaniapedrazzi:
Is something still missing or is it possible to merge this PR?

mael25:
> Is something still missing or is it possible to merge this PR?

I have not tested on MacOS and Windows yet...

omichel (approved):
Thank you for the tests!

Issue #3024, opened 2021-04-30
youBot mecanumWheel


Thomas-Feldmeier:
https://www.kuka.com/en-de/products/mobility/kuka-omnimove-drive-technology:
The wheel consists of two rims and nine free-running rollers, which are mounted at a 45 degree angle

the contactProperties in sampleWorld of youbot are strange, the mecanum wheel angle there is not 45 degrees but something else, so the speed front/back is not same as left/right. i changed them simular to robotino, and it worked. 

change contactProperties in projects/robots/kuka/youbot/worlds/youbot.wbt to this:

contactProperties [
    ContactProperties {
      material1 "InteriorWheelMat"
      coulombFriction [
        0, 2, 0
      ]
      frictionRotation -0.785398 0
      bounce 0
    }
    ContactProperties {
      material1 "ExteriorWheelMat"
      coulombFriction [
        0, 2, 0
      ]
      frictionRotation 0.785398 0
      bounce 0
    }
  ]

the controller could be improved to use something like this (simplified pseudocode of running function)

vx = speed forward/backward, m/s
vy = speed left/right, m/s
vrotation = turning speed, radians/s

move(double vx, vy, vrotation){
  for(int i=0; i<4; i++) {
    motor[i]->setVelocity(
		  (vx
		   + vy * ( (i==1 || i==2) ? 1 : -1)
		   + vrotation * ( (i==1 || i==3) ? 1 : -1) * distanceToRobotCentre)
		  / radius);
  }
}

with motor[0] = left front, motor[1] = right front, motor[2] = left back, motor[3] = right back (watch out, different order than the wheels in youBot)

radius = 0.05m = radius of boundingObject wheels
distanceToRobotCentre = 0.228 + 0.158 = x-offset + y-offset of wheel relative to robot center point

DEF WHEEL1 InteriorWheel {
      translation 0.228 -0.158 -0.055
      anchor 0.228 -0.158 -0.055
      name "wheel1"
      sensorName "wheel1sensor"
    }

Thomas-Feldmeier:
in https://cyberbotics.com/doc/guide/youbot should be some documentation about the contactProperties (like in https://cyberbotics.com/doc/guide/robotino3)

mael25:
Thank you @Thomas-Feldmeier for raising this issue. The angle of the mecanum wheels has been corrected and the controller improved according to your suggestion in #3025.

PR #3025, opened 2021-04-30
Correct Youbot wheels

mael25:
**Description**
Correct contactProperties of Kuka's YouBot robot and improve the keyboard controller.

**Related Issues**
This pull-request fixes issue #3024.

**Tasks**
- [x] Correct contactProperties
- [x] Controller: Velocity control with keyboard
- [x] Add keyboard to Matlab controller (currently only demo)
- [x] Fix keys with Shift modifier (Matlab)
- [x] Correct turning direction
- [x] Prevent auto-colliding configurations from happening
- [x] Update YouBot doc (contactProperties, etc.)


stefaniapedrazzi:
:eyes:

stefaniapedrazzi (commented):
Thank you!
I could not test the matlab controller but the code seems fine.

Just note that also the "tower_of_hanoi.wbt" is using the youBot robot and it would be good to update the ContactPropetries in this world as well, so that the same are used for all the distributed worlds containing the youBot.

mael25:
> Thank you!
> I could not test the matlab controller but the code seems fine.
> 
> Just note that also the "tower_of_hanoi.wbt" is using the youBot robot and it would be good to update the ContactPropetries in this world as well, so that the same are used for all the distributed worlds containing the youBot.

Thank you for the review, I have now updated this world as well!

stefaniapedrazzi (approved):
Thank you!

PR #3027, opened 2021-05-01
Adds functions to the supervisor API to get Fields and ProtoFields by their ID

jgueldenstein:
**Description**
Currently fields of a node can only be accessed using the field name. This PR adds functions to get the fields by their ID as well. Furthermore a function to get the name of a field and the count of fields is included.

**Tasks**
  - [x] Basic Implementation
  - [ ] API for other languages
  - [ ] Documentation


jgueldenstein:
I built this for a tool to automatically verify robot models for the RoboCup competition.
I would appreciate feedback if you would also like to see this merged in the devel branch.

src/webots/nodes/utils/WbSupervisorUtilities.cpp
```diff
@@ -1071,6 +1092,42 @@ void WbSupervisorUtilities::handleMessage(QDataStream &stream) {
...
+              field->listenToValueSizeChanges();
+              connect(field, &WbField::valueSizeChanged, this, &WbSupervisorUtilities::notifyFieldUpdate, Qt::UniqueConnection);
+            }
+          }
+        }
+      }
+      return;
+    }
+    case C_SUPERVISOR_FIELD_GET_FROM_ID: {
+      int node_id, field_id;
```
stefaniapedrazzi, line 1108: Nodes are searched using a unique id, but fields don't have ids but only an index in the nodes fields list.
```suggestion
    case C_SUPERVISOR_FIELD_GET_FROM_INDEX: {
```

src/controller/c/messages.h
```diff
@@ -115,6 +115,8 @@
 #define C_SUPERVISOR_VIRTUAL_REALITY_HEADSET_IS_USED 98
 #define C_SUPERVISOR_VIRTUAL_REALITY_HEADSET_GET_POSITION 99
 #define C_SUPERVISOR_VIRTUAL_REALITY_HEADSET_GET_ORIENTATION 100
+#define C_SUPERVISOR_FIELD_GET_COUNT 101
```
stefaniapedrazzi, line 118: This message name could be misunderstood I would rename it to `C_SUPERVISOR_NODE_GET_FIELD_COUNT`.

src/controller/c/messages.h
```diff
@@ -115,6 +115,8 @@
...
+#define C_SUPERVISOR_FIELD_GET_FROM_ID 102
```
stefaniapedrazzi, line 119: I would rename this to `C_SUPERVISOR_FIELD_GET_FROM_INDEX` given that fields don't have ids.

src/controller/c/supervisor.c
```diff
@@ -380,6 +391,9 @@ static bool allow_search_in_proto = false;
 static const char *node_def_name = NULL;
 static int proto_id = -1;
 static const char *requested_field_name = NULL;
+static bool requested_field_count = false;
```
stefaniapedrazzi, line 394: I would rename this so that it is more self-explanatory otherwise it may be confused with the number of items in a MF field.
```suggestion
static bool requested_node_field_count = false;
```

include/controller/c/webots/supervisor.h
```diff
@@ -94,7 +94,9 @@ WbNodeRef wb_supervisor_node_get_parent_node(WbNodeRef node);
 WbNodeRef wb_supervisor_node_get_selected();
 WbNodeType wb_supervisor_node_get_type(WbNodeRef node);
 WbFieldRef wb_supervisor_node_get_field(WbNodeRef node, const char *field_name);
+WbFieldRef wb_supervisor_node_get_field_by_index(WbNodeRef node, const int index);
```
stefaniapedrazzi, line 97: It is useless to define an argument passed by value as constant:
```suggestion
WbFieldRef wb_supervisor_node_get_field_by_index(WbNodeRef node, int index);
```

Same for `wb_supervisor_node_get_proto_field_by_index`.

src/controller/c/supervisor.c
```diff
@@ -2055,6 +2083,61 @@ const char *wb_supervisor_node_get_base_type_name(WbNodeRef node) {
 
   return wb_node_get_name(node->type);
 }
+WbFieldRef wb_supervisor_node_get_field_by_index(WbNodeRef node, const int id) {
```
stefaniapedrazzi, line 2086: Argument name should match the one in `supervisor.h`
```suggestion
WbFieldRef wb_supervisor_node_get_field_by_index(WbNodeRef node, int index) {
```

same in `wb_supervisor_node_get_proto_field_by_index`.

stefaniapedrazzi, line 2086: Missing empty new line between this and the previous function.

src/controller/c/supervisor.c
```diff
@@ -2055,6 +2083,61 @@ const char *wb_supervisor_node_get_base_type_name(WbNodeRef node) {
...
+    return NULL;
+
+  if (!is_node_ref_valid(node)) {
+    if (!robot_is_quitting())
+      fprintf(stderr, "Error: %s() called with a NULL or invalid 'node' argument.\n", __FUNCTION__);
+    return NULL;
+  }
+
+  robot_mutex_lock_step();
+  // yb: search if field is already present in field_list
```
stefaniapedrazzi, line 2097:
```suggestion
  // search if field is already present in field_list
```

src/controller/c/supervisor.c
```diff
@@ -2055,6 +2083,61 @@ const char *wb_supervisor_node_get_base_type_name(WbNodeRef node) {
...
```
stefaniapedrazzi, line 2124:
```suggestion
  // search if field is already present in field_list
```

include/controller/c/webots/supervisor.h
```diff
@@ -125,6 +127,9 @@ void wb_supervisor_node_add_force(WbNodeRef node, const double force[3], bool re
 void wb_supervisor_node_add_force_with_offset(WbNodeRef node, const double force[3], const double offset[3], bool relative);
 void wb_supervisor_node_add_torque(WbNodeRef node, const double torque[3], bool relative);
 
+int wb_supervisor_node_get_field_count(WbNodeRef node);
```
stefaniapedrazzi, line 130: I think it would be better to follow the style used by the Robot API for the devices:
https://www.cyberbotics.com/doc/reference/robot#wb_robot_get_device_by_index

and rename these functions
```suggestion
int wb_supervisor_node_get_number_of_fields(WbNodeRef node);
```
same for `wb_supervisor_node_get_proto_field_count`.

src/controller/c/supervisor.c
```diff
@@ -2055,6 +2083,61 @@ const char *wb_supervisor_node_get_base_type_name(WbNodeRef node) {
 
   return wb_node_get_name(node->type);
 }
+WbFieldRef wb_supervisor_node_get_field_by_index(WbNodeRef node, const int id) {
+  if (!robot_check_supervisor(__FUNCTION__))
+    return NULL;
+
+  if (!is_node_ref_valid(node)) {
```
stefaniapedrazzi, line 2090: You should also check if the field index is valid, i.e. at least that index is >= 0.

src/controller/c/supervisor.c
```diff
@@ -2055,6 +2083,61 @@ const char *wb_supervisor_node_get_base_type_name(WbNodeRef node) {
...
+  robot_mutex_lock_step();
+  // yb: search if field is already present in field_list
+  WbFieldRef result = find_field_by_id(id, node->id);
+  if (!result) {
+    // otherwise: need to talk to Webots
+    requested_field_id = id;
+    node_ref = node->id;
+    wb_robot_flush_unlocked();
+    requested_field_id = -1;
+    result = field_list;  // was just inserted at list head
```
stefaniapedrazzi, line 2105: If there were some errors it is possible that the new field has not been added to the list.
You should check that the `field_list` changed after calling `wb_robot_flush_unlocked` (as in `node_get_from_id` ) or reset `requested_field_id` when handling the `C_SUPERVISOR_FIELD_GET_FROM_ID` answer.

Same for `wb_supervisor_node_get_proto_field_by_index`.

src/controller/c/supervisor.c
```diff
@@ -2055,6 +2083,61 @@ const char *wb_supervisor_node_get_base_type_name(WbNodeRef node) {
...
+  if (!result) {
+    // otherwise: need to talk to Webots
+    requested_field_id = id;
+    node_ref = node->id;
+    allow_search_in_proto = true;
+    wb_robot_flush_unlocked();
+    requested_field_id = -1;
+    result = field_list;  // was just inserted at list head
+    if (result && node->is_proto_internal)
+      result->is_proto_internal = true;
```
stefaniapedrazzi, line 2135: The retrieved field will be internal to PROTO.
```suggestion
    if (result)
      result->is_proto_internal = true;
```

src/controller/c/supervisor.c
```diff
@@ -2091,6 +2174,32 @@ WbFieldRef wb_supervisor_node_get_field(WbNodeRef node, const char *field_name)
   return result;
 }
 
+const int wb_supervisor_node_get_field_count(WbNodeRef node) {
+  robot_mutex_lock_step();
+  requested_field_count = true;
```
stefaniapedrazzi, line 2179: You should also reset `node_field_count = -1` before calling `wb_robot_flush_unlocked()` to handle correctly errors is the node is not found or communication errors.

Same for `wb_supervisor_node_get_proto_field_count`.

include/controller/cpp/webots/Node.hpp
```diff
@@ -126,8 +126,12 @@ namespace webots {
     Node *getParentNode() const;
     bool isProto() const;
     Node *getFromProtoDef(const std::string &name) const;
+    int getFieldCount() const;
```
stefaniapedrazzi, line 129: I would rename this function
```suggestion
int Node::getNumberOfFields() const {
```

same for `getProtoFieldCount`

stefaniapedrazzi (changes requested):

The implementation is globally fine.

There is some confusion between `index` and `id`.
Fields don't have ids but only an index in the node's field list. The existing variable names are not consistent with this, but at least the new libController messages and API arguments names should be adjusted to use only the `index` term and not `id`.
I will then fix the variable and function names in the existing code later.

Please also do not forget to add a test in the test_suite:
https://github.com/cyberbotics/webots/wiki/Adding-New-Node-and-API-Function#adding-a-new-api-function (last point)
You can extend an existing test, for example `tests/api/worlds/supervisor_field.wbt`.

jgueldenstein:
Thank you very much for you feedback @stefaniapedrazzi. I am quite busy with other projects at the moment but I will get back to this definitely.

medrimonia:
@stefaniapedrazzi Thank you very much for the feedback, I will take over to try making this PR compatible with the elements you mentioned.

I pushed a first set of modification but I am aware that there are multiple elements to address before merging it into develop (other languages/documentation/tests).

I had a few questions on some of your comments:

> It is better to order the messages by type and move the close to other related messages instead of appending them at the end of the list.

I did this, but the numbers are not ordered anymore due to this change and updating them would include changing a lot of lines: what is the procedure regarding the number of those messages.

> Fields don't have ids but only an index in the node's field list.

I have seen the following in the code:
```
typedef struct WbFieldStructPrivate {
  ...
  int id;                  // attributed by Webots
  ...
} WbFieldStruct;
```
Which led me to change the existing condition to something that sounds suspicious to me: `index == field->id` (in `find_field_by_index`). What is the meaning of this `id` in WbFieldStruct and how is it attributed by Webots?

> This check is not in the correct position.
> If only positive indices are valid, then it should be the libController that checks that the field index is not < 0 and returns immediately without communicating with the Webots main application.
>
> Note that in any case, even if this won't be checked in the libController, the check should be performed before calling WbNode::findNode for performance reasons.

I am not sure I fully understood this one. I still lack an understanding of the inner mechanism of Webots with respect to this and I added the suggested tests in `wb_supervisor_node_get_field_by_index` and `wb_supervisor_node_get_proto_field_by_index`. Does this mean that the tests here should be removed?

stefaniapedrazzi:
Thank you for the changes.
Here are some initial aswers, I will now check the changes and make a new review.

> > It is better to order the messages by type and move the close to other related messages instead of appending them at the end of the list.
> 
> I did this, but the numbers are not ordered anymore due to this change and updating them would include changing a lot of lines: what is the procedure regarding the number of those messages.

Numbering of lines should be consecutive, so you should not only move the line but also change the message type.
It is not a problem if multiple lines have to be changes.
Note that the messages are still not ordered correctly.


> > Fields don't have ids but only an index in the node's field list.
> 
> I have seen the following in the code:
> 
> ```
> typedef struct WbFieldStructPrivate {
>   ...
>   int id;                  // attributed by Webots
>   ...
> } WbFieldStruct;
> ```
> 
> Which led me to change the existing condition to something that sounds suspicious to me: `index == field->id` (in `find_field_by_index`). What is the meaning of this `id` in WbFieldStruct and how is it attributed by Webots?

Yes, id and index are the same internally, but we should be consistent with the exposed API and only use the `index` term for everything that is visible to users.

> > This check is not in the correct position.
> > If only positive indices are valid, then it should be the libController that checks that the field index is not < 0 and returns immediately without communicating with the Webots main application.
> > Note that in any case, even if this won't be checked in the libController, the check should be performed before calling WbNode::findNode for performance reasons.
> 
> I am not sure I fully understood this one. I still lack an understanding of the inner mechanism of Webots with respect to this and I added the suggested tests in `wb_supervisor_node_get_field_by_index` and `wb_supervisor_node_get_proto_field_by_index`. Does this mean that the tests here should be removed?

Yes, the check for `field_index != -1` in `WbSupervisorUtilities.cpp` can be removed.
I will now review more deeply the new checks you added in `supervisor.c` because requesting the fields' count needs to communicate to the Webots main process, given that the value is not cached, and it won't be very efficient.
I will probably come with a different suggestion very soon.

src/controller/c/messages.h
```diff
@@ -81,6 +81,8 @@
 #define C_SUPERVISOR_FIELD_SET_VALUE 62
 #define C_SUPERVISOR_FIELD_IMPORT_NODE_FROM_STRING 63
 #define C_SUPERVISOR_FIELD_REMOVE_VALUE 64
+#define C_SUPERVISOR_NODE_GET_FIELD_COUNT 101
+#define C_SUPERVISOR_FIELD_GET_FROM_INDEX 102
```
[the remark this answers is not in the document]
medrimonia: Ok, thanks for your comments, I believe it's now properly done.

stefaniapedrazzi (changes requested):
I just reviewed the latest changes.
From the code side, these are the remaining changes.

src/webots/nodes/utils/WbSupervisorUtilities.cpp
```diff
@@ -1071,6 +1084,40 @@ void WbSupervisorUtilities::handleMessage(QDataStream &stream) {
...
+      stream >> allowSearchInProto;
+
+      mFoundFieldId = -1;
+      mFoundFieldType = 0;
+      mFoundFieldCount = -1;
+      mFoundFieldIsInternal = false;
+
+      WbNode *const node = WbNode::findNode(node_id);
+      if (node) {
+        if (field_index != -1) {
```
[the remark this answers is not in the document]
medrimonia: Thanks for the suggestion, I took it, just changing:
`const QVector<WbField *> fields(fieldsOrParameters());`
to
`const QVector<WbField *> &fields = fieldsOrParameters();`

Which looks more consistent with the rest of the file and avoid the call to a copy constructor according to my understanding.

src/controller/c/supervisor.c
```diff
@@ -2056,6 +2083,84 @@ const char *wb_supervisor_node_get_base_type_name(WbNodeRef node) {
...
+WbFieldRef wb_supervisor_node_get_field_by_index(WbNodeRef node, int index) {
+  if (!robot_check_supervisor(__FUNCTION__))
+    return NULL;
+
+  if (!is_node_ref_valid(node)) {
+    if (!robot_is_quitting())
+      fprintf(stderr, "Error: %s() called with a NULL or invalid 'node' argument.\n", __FUNCTION__);
+    return NULL;
+  }
+  const int node_fields = wb_supervisor_node_get_number_of_fields(node);
```
stefaniapedrazzi, line 2095: You also have to remove this line given that `node_fields` is no longer used and adjust the error message.

src/controller/c/supervisor.c
```diff
@@ -2056,6 +2083,84 @@ const char *wb_supervisor_node_get_base_type_name(WbNodeRef node) {
...
+WbFieldRef wb_supervisor_node_get_proto_field_by_index(WbNodeRef node, int index) {
+  if (!robot_check_supervisor(__FUNCTION__))
+    return NULL;
+
+  if (!is_node_ref_valid(node)) {
+    if (!robot_is_quitting())
+      fprintf(stderr, "Error: %s() called with a NULL or invalid 'node' argument.\n", __FUNCTION__);
+    return NULL;
+  }
+  const int node_fields = wb_supervisor_node_get_proto_number_of_fields(node);
```
stefaniapedrazzi, line 2133: You also have to remove this line given that `node_fields` is no longer used and adjust the error message.

stefaniapedrazzi (changes requested):
Thank you for applying the changes.
It just remains to remove the costly call of `wb_supervisor_node_get_number_of_fields` in `supervisor.c`.

Then the basic implementation is done and it remains just the documentation and the other APIs.

medrimonia:
Thank you very much for the support. Given the current workload for the competition we will apply the approach discussed with @omichel and merge this in the [RoboCup-Humanoid-TC fork](https://github.com/RoboCup-Humanoid-TC/webots/) for now and once we have more time (hopefully after the mock-competition on June 6th), I will get back to this PR to add the documentation, other API and tests.

stefaniapedrazzi:
We will be interested in including this PR in the next version of Webots (R2021b) that will be released in two weeks.
So if you don't have time to complete it, we could help and taking care of some missing parts (documentation, other APIs and tests).

medrimonia:
I think it is very unlikely that I will have some time to work on this before the end of the RoboCup competition (June 28th). Therefore if you can take care of some of the missing parts, that would definitely be great!

jgueldenstein:
Same problem for me as @medrimonia  unfortunately.

stefaniapedrazzi:
Ok, so I will take care of completing this PR.

jgueldenstein:
Thank you very much @stefaniapedrazzi edits by maintainers are set to allowed.

stefaniapedrazzi:
To complete the implementation we will need to synchronize with the Webots develop branch (in particular for the Matlab and ROS APIs).
Given that it is better to wait until the end of the competition to update this branch, I will first create a separate PR (#3202) and merge it directly to the Webots develop branch.

jgueldenstein:
closed since #3202 is merged

PR #3030, opened 2021-05-03
Improve Linux installation instructions

stefaniapedrazzi:
Following the feedback in #3006, I tried to improve the Linux installation instructions:
* [x] improve the wiki page:
    * clearly state that automatic scripts only fully works on Ubuntu LTS
    * clearly state that the `.bash` configuration should be adapted to the user system
    * clearly state that on other Linux distribution python-dev package 
    https://github.com/cyberbotics/webots/wiki/Linux-installation/_compare/ac6b97088b70a397ad9f1ebcd0ff0c1265c1eb54...3e45f2826bcb70df7025a7c6d75f44e30bad6d34
* [x] improve "Unsupported Linux version" message
* [x] `execstack` library is not available on some Linux distribution (e.g., Debian 11) and could be needed for e-puck cross compilation but seems safe to move it from `compilation` to  `optional` dependency
* [x] add python-dev availability check in Python API Makefile

### DIFF
--- a/scripts/install/linux_compilation_dependencies.sh
+++ b/scripts/install/linux_compilation_dependencies.sh
@@ -14,7 +14,7 @@ if [[ $UBUNTU_VERSION == "16.04"  || $UBUNTU_VERSION == "18.04" ]]; then
 elif [[ $UBUNTU_VERSION == "20.04" ]]; then
        apt install --yes libzip5 python-pip-whl
 else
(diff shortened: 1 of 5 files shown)


ShuffleWire:
Thanks for taking this in consideration ! :) It should help newcomers and others distros users !

omichel (approved):
Thank you.

PR #3031, opened 2021-05-03
Fix initial labels in streaming viewer mode

BenjaminDeleze:
**Description**
The labels are now send along the updates of positions/rotations/...

The labels that are set before the connection with the streaming viewer are sent with the first update.

However, it happens that this first update is droped by the streaming viewer because the scene is not fully loaded.

As labels are HTML elements, it is not a problem if some nodes or textures are not loaded, we can already draw them.

**Fix**
Instead of fully dropping an update if the scene is still loading, we drop everything but the labels.

Additionally, we restore the correct link in the animation template (the file will be changed on the server when merging)


omichel (approved):
Thank you.

PR #3032, opened 2021-05-03
Improve camera performance

lukicdarkoo:
**Description**
This PR improves performances of the `wb_camera_get_image` function by 1000x (from ~2000us to ~2us). At the moment, it is a raw implementation, I still have to check for side effects.

Interestingly, in our ROS 2 performance benchmark, there are no performance improvements. It is probably because the bottleneck is the mutex lock rather than unnecessary processing.


src/webots/nodes/WbCamera.cpp
```diff
@@ -533,6 +533,15 @@ void WbCamera::writeConfigure(QDataStream &stream) {
...
   WbAbstractCamera::writeAnswer(stream);
+
+  if (!mImageWritten) {
+    if (isEnabled()) {
+      copyImageToSharedMemory(mWrenCamera, image());
+      mImageChanged = false;
+    }
+    mImageReady = true;
+  }
+
```
omichel, line 544: We should probably do this in Lidar and RangeFinder as well.
So, it could be better to put this code in `WbAbstractCamera.cpp` instead, isn't it?

omichel (commented):
Thank you, this is great.
I am wondering why you didn't apply this fix also to the RangeFinder and Lidar nodes.

src/controller/c/camera.c
```diff
@@ -707,8 +707,7 @@ const unsigned char *wb_camera_get_image(WbDeviceTag tag) {
     return ac->image->data;
 
   robot_mutex_lock_step();
-  bool success = image_request(ac->image, __FUNCTION__);
-  if (!ac->image->data || !success) {
+  if (!ac->image->data) {
```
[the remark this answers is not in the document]
lukicdarkoo: Tried here 691dc1b (Lidar and range finders)

The implementation of rotational lidars is convoluted, and I cannot make them work. It crashes and it doesn't give a correct result. It is probably due to the lifecycle of the device that I don't fully understand.

The `image_request` is used by the image segmentation, so I should check with @stefaniapedrazzi why is it used there.

Let me know if I should invest more time.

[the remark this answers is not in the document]
omichel: I believe we should apply the same mechanism to the segmented image as well.
As for the Lidar, I understand the rotational aspect of Lidar may induce some extra care, but in principle, it could work the same way as normal cameras when it comes to sending the result to the controller.
Also, the range-finder should be pretty similar to the camera.
I believe it is important to invest more time on this to be sure we are doing the best performance wise.
A very useful test would be to verify that we don't write twice the camera image in the shared memory segment (`copyImageToSharedMemory`) during the same step.

stefaniapedrazzi:
Just be aware that as-is this optimization also have performance drawbacks if the camera is enabled but the image is not used in the controller or very rarely used.
In this uses cases, to not loose performance, users will continuously have to enable/disable the camera when some measurements are needed (and wait at least one simulation step before the image is available).
Thus, I think it is very important to document this in the Reference Manual and warn users on the new best practice.

This could for example happen in case users are mainly using the recognition and segmentation functions for which the camera doesn't need to be enabled but could be seldomly used for visualization or other.

lukicdarkoo:
> Just be aware that as-is this optimization also have performance drawbacks if the camera is enabled but the image is not used in the controller or very rarely used.
> In this uses cases, to not loose performance, users will continuously have to enable/disable the camera when some measurements are needed (and wait at least one simulation step before the image is available).
> Thus, I think it is very important to document this in the Reference Manual and warn users on the new best practice.
> 
> This could for example happen in case users are mainly using the recognition and segmentation functions for which the camera doesn't need to be enabled but could be seldomly used for visualization or other.

I was thinking to introduce another function, but I was afraid it would be confusing for users.

omichel:
I agree that having a camera enabled and not using its image from the controller side will have a performance impact with this change. But this is more consistent with other sensors, and also more optimized as it will save messages exchanges between the libController and Webots. I believe we should document that if a camera is enabled, the images will be generated and copied to the shared memory segment regardless of calls to `wb_camera_get_image()`. If users they need speed, they could enable the camera only when needed and the new version should perform slightly better than the current one.

stefaniapedrazzi:
Seems fine. But we should really clearly document it (and add it in the ChangeLog) to warn users about the impact of enabling the camera without using it.

lukicdarkoo:
The tests are passing on my PC. The following devices are affected by the changes: Camera, Camera Segmentation, Lidars (static and rotational), and Range Finder.

src/controller/c/lidar.c
```diff
@@ -391,6 +390,7 @@ double wb_lidar_get_max_range(WbDeviceTag tag) {
 
 const float *wb_lidar_get_range_image(WbDeviceTag tag) {
   AbstractCamera *ac = lidar_get_abstract_camera_struct(tag);
+  robot_mutex_lock_step();
```
omichel, line 393: I would move this call to `robot_mutex_lock_step()` down as I fear it could lock the mutex and return (if invalid device tag or remote control mode) without unlocking it...

src/controller/c/range_finder.c
```diff
@@ -228,6 +227,8 @@ double wb_range_finder_get_max_range(WbDeviceTag tag) {
 
 const float *wb_range_finder_get_range_image(WbDeviceTag tag) {
   AbstractCamera *ac = range_finder_get_abstract_camera_struct(tag);
+
+  robot_mutex_lock_step();
```
omichel, line 231: Same concern here.

omichel (changes requested):
I like very much this PR as the number of deleted lines is way higher than the number of added lines. We should also document it properly in the change log and maybe also reference manual.

src/controller/c/camera.c
```diff
@@ -911,10 +883,6 @@ int wb_camera_recognition_save_segmentation_image(WbDeviceTag tag, const char *f
   }
 
   // make sure image is up to date before saving it
```
stefaniapedrazzi, line 885: This line needs to be delete as well

src/webots/nodes/WbLidar.cpp
```diff
@@ -208,8 +208,10 @@ void WbLidar::prePhysicsStep(double ms) {
 
 void WbLidar::postPhysicsStep() {
   WbSolid::postPhysicsStep();
-  if (isRotating() && mSensor->isEnabled())
+  if (isRotating() && mSensor->isEnabled()) {
+    // copyImageToSharedMemory(mWrenCamera, image());
```
stefaniapedrazzi, line 212: Why this line is commented?

stefaniapedrazzi (changes requested):
The device reset functionality in the libController is now unused and should be completely removed from `robot.c` and `device_private.h` as well.

lukicdarkoo:
I am getting segmentation faults on the other PC for unrelated worlds :/ Investigating...

lukicdarkoo:
0552b18 This commit fixes the random Lidar Segmentation faults (for simulations in which I managed to reproduce the error).

The method returns the amount of shared memory that has to be allocated. `WbAbstractCamera` fills it depending on the resolution of the `WbWrenCamera` which is >= lidar's resolution (typically the physical lidar has fewer rows). Since the image size is often bigger than the size returned by `WbLidar::size()` the `FrameBuffer::copyContents` fails to perform the `memcpy`. Fixing the value returned by the `WbLidar::size()` fixes the problem as well.

However, I don't understand how it was working in the first place. The segmentation fault was hard to reproduce locally and maybe we were lucky or I have been missing something.

**General Memory Management Issues**
There are other questions now. The `WbLidar::copyAllLayersToSharedMemory` method copies the range layers over the image received by the `WbWrenCamera` which forces developers to be very careful about memory management. I guess it is done to reduce the heap size. But storing the point cloud in the same memory chunk seems unnecessary, I couldn't find a reason why it is done like that, it just makes the code hard to read.

**General Processing Issues** 
Some performance optimization opportunities I noticed. The `WbLidar::copyAllLayersToSharedMemory` does many `memcpy()` (those are RAM to RAM memory copies). Maybe it is possible to avoid those by returning pointers to parts of memory in which the actual range layers are stored. Then, the point clouds are calculated on the CPU in a loop for each point. I guess this can be parallelized with OpenMP or Intel TBB. Each point can be calculated and stored independently, so it should not be very hard.

omichel:
I fully agree with your comments.
1. The point cloud data should probably be stored in another shared memory segment or written to the pipe. It should not overwrite the range image.
2. Yes, we should avoid memcpy as much as possible and use OpenMP as much as possible. However, we investigated using OpenMP some time ago and although it was working fine on Linux and Windows, we had trouble in getting it working on macOS. But meanwhile the situation may have improved on macOS.

May I let you open two issues about it, as I believe it is out of the scope of this PR?

src/controller/c/lidar.c
lukicdarkoo: Done for all devices

src/controller/c/camera.c
lukicdarkoo: Missed it, fixed

src/webots/nodes/WbLidar.cpp
lukicdarkoo: Leftover, thanks for pointing

lukicdarkoo:
> May I let you open two issues about it, as I believe it is out of the scope of this PR?

Sure

omichel:
It seems that it is possible (although not straightforward) to use OpenMP on macOS: https://mac.r-project.org/openmp/

lukicdarkoo:
> It seems that it is possible (although not straightforward) to use OpenMP on macOS: https://mac.r-project.org/openmp/

According to [this](https://stackoverflow.com/a/29109926/1983050) it is just about adding a flag.

omichel:
> According to this it is just about adding a flag.

Yes, but that means moving from clang to gcc (we currently use clang on macOS, although the command line is called gcc for compatibility reasons).

docs/reference/camera.md
```diff
@@ -293,6 +293,8 @@ period = wb_camera_get_sampling_period(tag)
 *enable and disable camera updates*
 
 The `wb_camera_enable` function allows the user to enable a camera.
+Once the camera is enabled, it will copy images from VRAM to RAM at each time step, regardless of `wb_camera_get_image` calls.
```
omichel, line 296:
```suggestion
Once the camera is enabled, it will copy images from GPU memory to CPU memory at each time step, regardless of `wb_camera_get_image` calls.
```

docs/reference/lidar.md
```diff
@@ -292,6 +292,8 @@ period = wb_lidar_get_sampling_period(tag)
 *enable and disable lidar updates*
 
 The `wb_lidar_enable` function allows the user to enable lidar updates.
+Once the lidar is enabled, it will copy range images from VRAM to RAM at each time step, regardless of `wb_lidar_get_range_image` calls.
```
omichel, line 295:
```suggestion
Once the lidar is enabled, it will copy range images from GPU memory to CPU memory at each time step, regardless of `wb_lidar_get_range_image` calls.
```

omichel (commented):
Thank you. It looks good to me.
I just have minor suggestion for the documentation.

omichel (approved):
Thank you.

stefaniapedrazzi (changes requested):
Thank you.
Everything seems fine, maybe it is just missing a sentence in the documentation for the camera segmentation.

But when testing a little bit I found a crash:
1. open `range_finder.wbt`
2. change the RangeFinder image width and set it to a bigger value (for example 257)
3. change the RangeFinder.spherical value to TRUE

-> Webots crashes
```
(gdb) bt
#0  0x00007ffff27aba5f in __memmove_avx_unaligned_erms () at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:257
#1  0x0000555555d8d6e2 in wren::FrameBuffer::copyContents(unsigned long, void*) (this=0x5555733e8670, index=1, data=0x7fffb02d4000) at FrameBuffer.cpp:160
#2  0x0000555555d8e505 in wr_frame_buffer_copy_contents(WrFrameBuffer*, int, void*) (frame_buffer=0x5555733e8670, index=1, data=0x7fffb02d4000) at FrameBuffer.cpp:368
#3  0x0000555555c045ce in WbWrenCamera::copyContentsToMemory(void*) (this=0x55555d514720, data=0x7fffb02d4000) at wren/WbWrenCamera.cpp:508
#4  0x00005555558efdb5 in WbAbstractCamera::copyImageToSharedMemory(WbWrenCamera*, unsigned char*) (this=0x55555c137630, camera=0x55555d514720, data=0x7fffb02d4000 "") at nodes/WbAbstractCamera.cpp:277
#5  0x00005555558f0006 in WbAbstractCamera::writeAnswer(QDataStream&) (this=0x55555c137630, stream=...) at nodes/WbAbstractCamera.cpp:311
```

I also tested on a different branch based on develop and it seems to work without crashing.

docs/reference/camera.md
```diff
@@ -293,6 +293,8 @@ period = wb_camera_get_sampling_period(tag)
 *enable and disable camera updates*
 
 The `wb_camera_enable` function allows the user to enable a camera.
+Once the camera is enabled, it will copy images from GPU memory to CPU memory at each time step, regardless of `wb_camera_get_image` calls.
```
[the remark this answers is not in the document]
lukicdarkoo: > But when testing a little bit I found a crash:
> 
> 1. open `range_finder.wbt`
> 2. change the RangeFinder image width and set it to a bigger value (for example 257)
> 3. change the RangeFinder.spherical value to TRUE
> 
> -> Webots crashes

Thank you for discovering this, I can reproduce it by running the simulation and then changing the width and spherical parameters. However, the same happens on the develop branch, so it is not regression.

Should I fix it in this PR or another?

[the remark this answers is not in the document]
omichel: > Should I fix it in this PR or another?

If it doesn't create any conflicts with this branch, it should be fixed in another PR, otherwise it is probably better to fix it in this PR.

lukicdarkoo:
> But when testing a little bit I found a crash:
> 
> 1. open `range_finder.wbt`
> 2. change the RangeFinder image width and set it to a bigger value (for example 257)
> 3. change the RangeFinder.spherical value to TRUE
> 
> -> Webots crashes

Thank you for discovering this, I can reproduce it by running the simulation and then changing the width and spherical parameters. However, the same happens on the develop branch, so it is not regression.

Should I fix it in this PR or another?

omichel:
> Should I fix it in this PR or another?

If it doesn't create any conflicts with this branch, it should be fixed in another PR, otherwise it is probably better to fix it in this PR.

stefaniapedrazzi (approved):
Thank you

PR #3036, opened 2021-05-05
Recursively search for finalized PROTO node

### LINKED ISSUE #3034: Supervisor API caches positions of nodes
**Describe the Bug**
I am using the supervisor API and calling `getPosition()` on nodes to get their position in the world. Now, when the robot is moved in the simulation, the result of `getPosition()` is still the old result even though it has obviously changed.

**Steps to Reproduce**
1. Open Webots
2. Start an external supervisor controller
3. In the controller, get the robot with `robot = supervisor.getFromDef('robot_def')` and the link with `link = robot.getFromProtoDef('link_def')`
4. Call `link.getPosition()`
4. Move the robot with the node in the simulation
5. Step a few times and call `link.getPosition()` again

**Expected behavior**
The returned position should change.

**System**
 - Operating System: Debian Bullseye
 - Graphics Card: –
 - Webots version: 2021b Nightly Build (30-4-2021)

**Additional context**

stefaniapedrazzi:
Fix #3034:
fix the search for PROTO parameter node instance considering only one nested layer.
So accessing parameters of multiple nested PROTO nodes were not correctly handled causing assertion failures and crashes due to request of physical properties to non-finalized intermediate nodes.

This will be the simpler solution for users because they do not need to know the internal sub PROTO node definitions.
However, it is faster to already get the parameter node instance (or the closest node as possible) from the robot controller.


stefaniapedrazzi:
All the tests passed when the PR was set to "ready for review".
The Linux failure comes from a test I started later to check the status of the master branch.

ad-daniel (changes requested):
The implementation looks good and appears to be working but I'm having a crash when requesting the velocity:

1. give a def name to innermost `ProtoNestedMultipleInstances`, say `INNER`.
2.  get node ref `WbNodeRef inner_node = wb_supervisor_node_get_from_proto_def(parent_node, "INNER");`, works fine and isn't NULL.
3. being a Solid, I should be able to get it's velocity: `wb_supervisor_node_get_velocity(inner_node);` but it segfaults.

The crash appears to be limited to the `get_velocity` API function, I quickly tested others like `get_center_of_mass` or `get_number_of_contact_points` and they work without issues.

stefaniapedrazzi:
The crash is related to the search of the top solid and fixed in #3039.
Once #3039 is merged and I will synchronize this branch and can be retested.

ad-daniel:
I see, I checked a few more and these too might be worth testing, currently:

- `get_velocity` fails but the `set_velocity` works (or at least doesn't crash). 
- `add_force` crashes but `add_torque` and `add_force_with_offset` doesn't.

stefaniapedrazzi:
Both the reported crashes should now be fixed.

ad-daniel (commented):
Yes they appear to be working now, thank you.
I just noticed that the test unit doesn't actually reset using the GUI button (i.e the proto doesn't revert back to the initial position) but I don't think any of these changes would break that, could it be related to the 'recent' changes to the reset logic?

stefaniapedrazzi:
Without inspecting I don't know what is the source of the reset issue, but I can reproduce the master branch.
So it is not related to this PR.
It should be reported in a new issue.

Issue #3037, opened 2021-05-05
Webots crash when inserting Shrimp inside a transform/group on develop


BenjaminDeleze:
**Describe the Bug**
Webots crash when adding the Shrimp robot to the children of a transform/group

**Steps to Reproduce**
1. Build Webots from the develop branch
2. Open a world
3. Add a group or a transform
4. Add the Shrimp robot to the children of the group/transform
5. Webots crash

**Expected behavior**
The robot is supposed to be added normally to the scene

**System**
 - Operating System: Linux Ubuntu 20.04
 - Graphics Card: Nvidia GeForce GTX 1060 3GB

PR #3039, opened 2021-05-05
Fix crash solidReference

BenjaminDeleze:
**Description**
Webots crashed when we were inserting a solidReference and the top node of this solidReference was a transform or a group.
For example: Transform->Solid->HingeJoint->SolidReference.

With the introduction in developp of the possibility to introduce robots in transform, Webots would crash if we inserted the Shrimp robot in a transform.

CAUSE: to get the top solid, Webots was retrieving the top node and casting it to solid. But in the case of a group or transform as the top node, it would fail.

FIX: instead of retrieving the top node, we retrieve the uppest solid.

**Related Issues**
This pull-request fixes issue #3037

### DIFF
--- a/src/webots/nodes/utils/WbNodeUtilities.cpp
+++ b/src/webots/nodes/utils/WbNodeUtilities.cpp
@@ -663,7 +663,24 @@ WbMatter *WbNodeUtilities::findUppermostMatter(WbNode *node) {
 }
 
 WbSolid *WbNodeUtilities::findTopSolid(const WbNode *node) {
-  return dynamic_cast<WbSolid *>(const_cast<WbNode *>(findTopNode(node)));
+  if (node == NULL || node->isWorldRoot())
+    return NULL;
+
+  const WbNode *n = node;
+  const WbNode *parent = n->parentNode();
+  WbSolid *topSolid = NULL;
+  WbSolid *currentNode;
+  while (parent) {
+    currentNode = dynamic_cast<WbSolid *>(const_cast<WbNode *>(n));
+    if (currentNode)
+      topSolid = currentNode;
+    if (parent->isWorldRoot())
+      return topSolid;
+
+    n = parent;
+    parent = n->parentNode();
+  }
+  return NULL;
 }
 
 WbTransform *WbNodeUtilities::findUpperTransform(const WbNode *node) {
--- a/src/webots/nodes/utils/WbNodeUtilities.cpp
+++ b/src/webots/nodes/utils/WbNodeUtilities.cpp
@@ -663,24 +663,23 @@ WbMatter *WbNodeUtilities::findUppermostMatter(WbNode *node) {
 }
 
 WbSolid *WbNodeUtilities::findTopSolid(const WbNode *node) {
-  if (node == NULL || node->isWorldRoot())
+  if (node == NULL)
     return NULL;
 
   const WbNode *n = node;
   const WbNode *parent = n->parentNode();
   WbSolid *topSolid = NULL;
-  WbSolid *currentNode;
   while (parent) {
-    currentNode = dynamic_cast<WbSolid *>(const_cast<WbNode *>(n));
-    if (currentNode)
-      topSolid = currentNode;
+    WbSolid *currentSolid = dynamic_cast<WbSolid *>(const_cast<WbNode *>(n));
+    if (currentSolid)
+      topSolid = currentSolid;
     if (parent->isWorldRoot())
-      return topSolid;
+      break;
 
     n = parent;
     parent = n->parentNode();
   }
-  return NULL;
+  return topSolid;
 }
 
 WbTransform *WbNodeUtilities::findUpperTransform(const WbNode *node) {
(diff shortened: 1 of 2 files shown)


src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -663,7 +663,24 @@ WbMatter *WbNodeUtilities::findUppermostMatter(WbNode *node) {
...
-  return dynamic_cast<WbSolid *>(const_cast<WbNode *>(findTopNode(node)));
+  if (node == NULL || node->isWorldRoot())
+    return NULL;
+
+  const WbNode *n = node;
+  const WbNode *parent = n->parentNode();
+  WbSolid *topSolid = NULL;
+  WbSolid *currentNode;
+  while (parent) {
+    currentNode = dynamic_cast<WbSolid *>(const_cast<WbNode *>(n));
```
ShuffleWire, line 674: I think that it would be nice to declare and initialize `currentNode` on line 674, because you don't use it outside.
Moreover, I would name it `currentSolid`, as it a pointer to a `WbSolid`. (or a `NULL` if it's not a `WbSolid`)

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -663,7 +663,24 @@ WbMatter *WbNodeUtilities::findUppermostMatter(WbNode *node) {
...
+    if (currentNode)
+      topSolid = currentNode;
+    if (parent->isWorldRoot())
+      return topSolid;
+
+    n = parent;
```
ShuffleWire, line 680: I'm not convinced you actually need `n`, you might use `node`, and `node->parentNode()` each time you need the parent, instead of storing it in `parent`.

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -663,7 +663,24 @@ WbMatter *WbNodeUtilities::findUppermostMatter(WbNode *node) {
...
+    parent = n->parentNode();
+  }
+  return NULL;
```
ShuffleWire, line 683: I don't know if you must return `NULL` or `topSolid`. Could the code have found a top Solid without having meet the WorldRoot ?
Moreover, if you return `topSolid`, you could make a `break` instead of the `return` on line 678, just to get only one exit point in the function, which I found to be kind of nicer :)

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -663,7 +663,24 @@ WbMatter *WbNodeUtilities::findUppermostMatter(WbNode *node) {
 }
 
 WbSolid *WbNodeUtilities::findTopSolid(const WbNode *node) {
-  return dynamic_cast<WbSolid *>(const_cast<WbNode *>(findTopNode(node)));
+  if (node == NULL || node->isWorldRoot())
```
ShuffleWire, line 666: Testing `node->isWorldRoot()` might not be necessary, as long as a node which get `node->isWorldRoot() == true` also provide `node->parentNode() == NULL` (which must be confirmed, I'm not certain on that one, event if it sound logical). In that case, your code handle nicely this case, and return `NULL` as well.

stefaniapedrazzi (changes requested):
Thank you!

I agree with comments to rename `currentNode` to `currentSolid` (or anything else containing `solid`) and moving the declaration inside the loop, and that the code could be simplified by removing the `return topSolid` in the loop and change the last return from `return NULL;` to `return topSolid;`. It should be confirmed by the tests, but I don't think there should be any difference in the functionality.

Even if the the other 2 comments of @ShuffleWire are correct, checking immediately for `node->isWorldRoot` and using a different variable `n` than the argument for the loop is fine.
I will let you decide if you want to change it or not.

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -663,24 +663,23 @@ WbMatter *WbNodeUtilities::findUppermostMatter(WbNode *node) {
...
   while (parent) {
-    currentNode = dynamic_cast<WbSolid *>(const_cast<WbNode *>(n));
-    if (currentNode)
-      topSolid = currentNode;
+    WbSolid *currentSolid = dynamic_cast<WbSolid *>(const_cast<WbNode *>(n));
+    if (currentSolid)
+      topSolid = currentSolid;
     if (parent->isWorldRoot())
-      return topSolid;
+      break;
```
stefaniapedrazzi: What I meant is that these lines can be completely removed, not simply removing the return with a break.
In any case the while loop will break because in the next iteration `parent` will be NULL.

stefaniapedrazzi (changes requested):
One last thing to change.

src/webots/nodes/utils/WbNodeUtilities.cpp
BenjaminDeleze: You are right.

stefaniapedrazzi (approved):
Thank you

PR #3040, opened 2021-05-05
Added mStep increment in ros.cpp

kyax32:
**Description**
Fixing a bug which affects sampling period of sensors enabled from a ROS service. Sensors enabled from ROS are publishing values based on the basic time step of the world and not on the given value when enabling the sensor. mStep was not increased in ROS::run method. Additionaly I changed mStepSize by mRobot->getBasicTimeStep() in ROS::run, when passing the elapsed time to the publishValues method. mStepSize seems to be always equal to 1, even if the basic time step is bigger than 1ms.

**Related Issues**
None (Discord discussion)

**Tasks**
  - [ ] Task 1
  - [ ] Task 2

**Documentation**
None

**Screenshots**
None

**Additional context**
Bug was discovered while creating a VIO module using ROS environment.

### DIFF
--- a/projects/default/controllers/ros/Ros.cpp
+++ b/projects/default/controllers/ros/Ros.cpp
@@ -112,6 +112,8 @@ void Ros::launchRos(int argc, char **argv) {
   fixName();
   bool rosMasterUriSet = false;
 
+  mStepSize = mRobot->getBasicTimeStep();
+
   for (int i = 1; i < argc; ++i) {
     const char masterUri[] = "--ROS_MASTER_URI=";
     const char name[] = "--name=";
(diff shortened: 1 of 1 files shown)


projects/default/controllers/ros/Ros.cpp
```diff
@@ -447,7 +447,7 @@ void Ros::run(int argc, char **argv) {
     ros::spinOnce();
     publishClockIfNeeded();
     for (unsigned int i = 0; i < mSensorList.size(); i++)
-      mSensorList[i]->publishValues(mStep * mStepSize);
+      mSensorList[i]->publishValues(mStep * mRobot->getBasicTimeStep());
```
lukicdarkoo, line 450: The `mStepSize` is configured in the `timeStepCallback` function. To preserve the functionality of the `timeStepCallback`, I would rather set default value of the `mStepSize` to `mRobot->getBasicTimeStep()` as soon as we have access to the `Robot` object.

lukicdarkoo (changes requested):
Thank you for this!

I have a minor comment about the `mStepSize`.

lukicdarkoo (approved):
Thank you!

omichel:
@kyax32: I just sent you an invite to become a member of the Webots Committers, so that you should be able to merge this PR.

kyax32:
Thanks @omichel ! The PR is now merged.

PR #3042, opened 2021-05-06
Disable broken windows tests for test suite label

stefaniapedrazzi:
Temporarily disable the broken Windows tests in case the "test suite" label is set to have more relevant tests outputs.


omichel (approved):
Thank you.

PR #3043, opened 2021-05-06
Uses Visual C++ 2017 instead of deprecated 2010

omichel:
Since Visual Studio Express 2010 seems deprecated, this PR tries to use Visual C++ 2017 instead.
I also updated the [wiki](https://github.com/cyberbotics/webots/wiki/Windows-Optional-Dependencies/_compare/85aa2fd8026903ba8d334e2cb80f87a745c247b4).

### DIFF
--- a/.github/workflows/test_suite_windows.yml
+++ b/.github/workflows/test_suite_windows.yml
@@ -22,7 +22,7 @@ jobs:
         GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
   build:
     needs: cleanup-runs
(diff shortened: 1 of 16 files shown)


stefaniapedrazzi:
#3042 should be reverted to enable the Windows tests now that the package compilation is fixed.

We should also understand if the Linux tests failure are related to this branch.

omichel:
> #3042 should be reverted to enable the Windows tests now that the package compilation is fixed.

Done.

> We should also understand if the Linux tests failure are related to this branch.

It seems unrelated as the failure is related with some OpenGL include file missing.

stefaniapedrazzi (approved):
Thank you

PR #3056, opened 2021-05-11
Enhancement in Makefile : add GDB option

ShuffleWire:
Adding -ggdb option in makefile help (a lot) GDB to debug, which it needed because of the usage of QT, probably.


omichel:
It would be probably useful to add it also to the `ode` and `wren` Makefiles.

ShuffleWire:
I will do so as soon as I've some time.
PS : Sorry for closing/reopening the PR, i've missclicked...

ShuffleWire:
As a side note, I'm not sure if we need this option for `ode` and `wren`, because it's initially caused by the use of Qt, and they don't depend on it, exact ? I've not tried to build and debug them, to be fair.
If needed, I'm not sure were to add it, because in both `ode` and `wren` there is no place to add debbuging options. Found one in Makefile.include, however. So I've made both modification, but only one of each should be enough.

src/wren/Makefile
```diff
@@ -25,6 +25,10 @@ STATIC_LIBRARY = true
 CFLAGS = -std=c++11 -DGLM_ENABLE_EXPERIMENTAL
 INCLUDE = -I$(WEBOTS_HOME_PATH)/include -I$(WEBOTS_HOME_PATH)/include/glad -I../glm -I../webots/external/siphash/
 
+ifeq ($(MAKECMDGOALS),debug)
+  CFLAGS += -g -ggdb
+endif
+
```
omichel, line 31:
```suggestion
```

src/ode/Makefile
```diff
@@ -63,6 +63,9 @@ endif
 ifeq ($(MAKECMDGOALS),release)
   CFLAGS += -DdNODEBUG
 endif
+ifeq ($(MAKECMDGOALS),debig)
+  CFLAGS += -g -ggdb
+endif
```
omichel, line 68:
```suggestion
```

omichel (changes requested):
I believe it is sufficient to apply it in the Makefile.include as this one is included by the ODE and WREN makefiles.

omichel:
Also, I believe that we should remove the `-g` option and leave only the `-ggdb` one, which is seems indeed better, see: https://stackoverflow.com/questions/668962/what-is-the-difference-between-gcc-ggdb-and-gcc-g

ShuffleWire:
As far as I understand, the -g option is more "standard". Whould removing it could cause issue when debugging with any thing others than gdb ? If not, or if it's not a issue, I would remove it.

ShuffleWire:
That it, finally just replacing -g with -ggdb in two location

omichel (approved):
Thank you.

PR #3057, opened 2021-05-11
Enhancement docs generation

ShuffleWire:
Fix makefile used to generate docs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,9 +143,9 @@ endif
 THREADS = $$(($(NUMBER_OF_PROCESSORS) * 3 / 2))
 
 docs:
-ifneq (, $(shell which python 2> /dev/null))
+ifneq (, $(shell which python3 2> /dev/null))
 	@+echo "#"; echo "# * documentation *";
-	-@+python docs/local_exporter.py --silent
+	-@+python3 docs/local_exporter.py --silent
 else
 	@+echo "#"; echo -e "# \033[0;33mPython not installed, skipping documentation\033[0m";
 endif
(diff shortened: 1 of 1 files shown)


omichel:
Did you experience problems with this script running python 2.7?
It was originally designed to support both python 2.7 and python 3.x:
```python
try:
    # For Python 3.0 and later
    from urllib.request import urlopen, HTTPError, URLError, Request
except ImportError:
    # Fall back to Python 2's urllib2
    from urllib2 import urlopen, HTTPError, URLError, Request
```
If we go for python3 only (which is good), we could also simplify this script.

ShuffleWire:
Python 2.7 is not on my PC, and the `python` command don't exist.
Using an alias or the package `python-is-python3` solve the issue, but it's not very handy.

ad-daniel:
I believe it's the default behavior now in ubuntu 20. If you don't install the `python-is-python3` package the command `which python` returns nothing even if you have python 2.7 installed.

ShuffleWire:
We could just update build instruction (just a little instruction for others distros than Ubuntu) to let the user provide the `python` command, which could be achieved by installing various packages. I'll reinstall later a fresh Ubuntu on my PC, and check what happen when building docs.
In  https://github.com/cyberbotics/webots/wiki/Linux-installation : Building on a system different than Ubuntu LTS
We could add : `Add python-is-pythonX` with X=2 or 3 depending of your python installed version.

omichel:
I wrote [this precision](https://github.com/cyberbotics/webots/wiki/Linux-installation/_compare/2696b72610877fae2dbfdaae41ba8580ae9012cd) to address the problem. Let me know if it looks good to you. Meanwhile, I am closing the issue for now. Feel free to re-open it if needed.

PR #3059, opened 2021-05-12
Fix wrong  implementation of `wbrotation.equal` in LUA module

ad-daniel:
**Description**
The `wbrotation.equal` function refers to non-existing variables instead of function arguments. 
Also fixes some trailing spaces present in these modules.

Not sure if it warrants a changelog entry, none of the PROTO is using it and if nobody reported it so far probably doesn't have much use.


omichel (approved):
Thank you.

PR #3067, opened 2021-05-17
Fixed Windows packaging

omichel:
The nightly builds were failing at creating the Windows packages because libzzip was recently renamed after a MSYS2 update.


lukicdarkoo (approved):
Thank you

PR #3074, opened 2021-05-18
Change macro definition

ShuffleWire:
**Description**
Change macro definition to use a do{}while(0) construct
In particular, it avoid duplicate `wb_robot_step(d)`
Note that modern compiler recognize do{}while(0) construct and optimize it away without emitting extra code.

### DIFF
--- a/include/controller/c/webots/robot.h
+++ b/include/controller/c/webots/robot.h
@@ -74,7 +74,7 @@ int wb_robot_step(int duration);  // milliseconds
 #ifdef __CYGWIN__  // In that case, we need to flush explicitly the stdout/stdin streams otherwise they are buffered
 // We cannot call fflush from the libController as libController is compiled with gcc8 and won't flush the stdout/stderr
 // of a gcc7 (cygwin) compiled binary. Therefore, we need to perform the fflush in a gcc7 compiled code, e.g., in a macro here.
-#define wb_robot_step(d) (fflush(NULL) ? wb_robot_step(d) : wb_robot_step(d))
+#define wb_robot_step(d) (fflush(NULL), wb_robot_step(d))
 #endif
 
 WbUserInputEvent wb_robot_wait_for_user_input_event(WbUserInputEvent event_type, int timeout);  // milliseconds


include/controller/c/webots/robot.h
```diff
@@ -66,15 +66,15 @@ int wb_robot_init();
  */
 #else
 int wb_robot_init_msvc();  // internally, this function just calls wb_robot_init()
-#define wb_robot_init() (setvbuf(stdout, NULL, _IONBF, 0), setvbuf(stderr, NULL, _IONBF, 0), wb_robot_init_msvc())
+#define wb_robot_init() do{setvbuf(stdout, NULL, _IONBF, 0); setvbuf(stderr, NULL, _IONBF, 0); wb_robot_init_msvc()}while(0)
```
omichel, line 69: Shouldn't we add a semi-colon here?
```suggestion
#define wb_robot_init() do{setvbuf(stdout, NULL, _IONBF, 0); setvbuf(stderr, NULL, _IONBF, 0); wb_robot_init_msvc();}while(0)
```

include/controller/c/webots/robot.h
```diff
@@ -66,15 +66,15 @@ int wb_robot_init();
...
 #endif
 
 int wb_robot_step(int duration);  // milliseconds
 
 #ifdef __CYGWIN__  // In that case, we need to flush explicitly the stdout/stdin streams otherwise they are buffered
 // We cannot call fflush from the libController as libController is compiled with gcc8 and won't flush the stdout/stderr
 // of a gcc7 (cygwin) compiled binary. Therefore, we need to perform the fflush in a gcc7 compiled code, e.g., in a macro here.
-#define wb_robot_step(d) (fflush(NULL) ? wb_robot_step(d) : wb_robot_step(d))
+#define wb_robot_step(d) do{fflush(NULL); wb_robot_step(d)}while(0)
```
omichel, line 77:
```suggestion
#define wb_robot_step(d) do{fflush(NULL); wb_robot_step(d);}while(0)
```

include/controller/c/webots/robot.h
ShuffleWire: yes indeed ! crazy that it build without it, thought

omichel:
The tests are failing (due to clang format). Can you please fix your code accordingly so that the tests pass?

ShuffleWire:
It seems that tests doesn't pass because cppcheck complain, but I've not changed thoses lines
```
src/controller/c/robot.c:644:3: error: syntax error [syntaxError]
  if (wb_robot_step(0) == -1) {
```
Should I do something about it ?

omichel:
> It seems that tests doesn't pass because cppcheck complain, but I've not changed thoses lines
> 
> ```
> src/controller/c/robot.c:644:3: error: syntax error [syntaxError]
>   if (wb_robot_step(0) == -1) {
> ```
> 
> Should I do something about it ?

That's probably because the `wb_robot_step()` function is expanded using your new macro and there is a syntax error in there... Does it compile properly on your machine?

omichel:
In fact, this is simply because your new robot_step() macro doesn't return a value as the previous one and also generates a syntax error when testing the return value... Reverting to the original macro seems to be the best solution unless you have a better idea.

ShuffleWire:
I've completely overlooked that, I've completely forgot to built it lately, and in fact the PR basically doesn't work ! because it while expand to something like : 
` if ( do{...}while(0) == -1) ` which is obviously not correct
If do think this PR is not a big deal, and we should discard it :) 
I will be more systematic in building it before committing, next !

omichel:
Yes, to me this PR is not a big deal, so, if you prefer to close it, that's fine with me.

ShuffleWire:
Should I delete the branch ?

ShuffleWire:
After the fact, I've realized that it still could be useful : 
`#define wb_robot_step(d) (fflush(NULL) ? wb_robot_step(d) : wb_robot_step(d))`
might be replaced with
`#define wb_robot_step(d) (fflush(NULL), wb_robot_step(d))`
(which use the same syntax as the other macro)
Or I don't get the ternary condition usage, and why it's mandatory here.

omichel:
> 
> 
> After the fact, I've realized that it still could be useful :
> `#define wb_robot_step(d) (fflush(NULL) ? wb_robot_step(d) : wb_robot_step(d))`
> might be replaced with
> `#define wb_robot_step(d) (fflush(NULL), wb_robot_step(d))`
> (which use the same syntax as the other macro)
> Or I don't get the ternary condition usage, and why it's mandatory here.

Yes, `#define wb_robot_step(d) (fflush(NULL), wb_robot_step(d))` looks simpler indeed, and less hacky than the ternary operator.

omichel (approved):
Thank you.

PR #3076, opened 2021-05-18
Replace the three.js part of documentation viewer

BenjaminDeleze:
**Description**
Replace the three.js parts of the documentation viewer with wrenjs.

**Tasks**
  - [x] Replace webots.min.js by wrenjs
  - [x] Fix wrong location of .hdr background
  - [x] Hardcode 'NUE' in worldinfo
  - [x] Add mention of the pip library needed by `web_component_studio` 
  - [x] Script loader
  - [x] Adress the problem when switching between robot
  - [x] Modify JSon and .x3d of ENU robots (robotis op2, pioneer 3 AT and tinkerbots)
  - [x] Convert billboard panel in non three.js code
  - [x] Improve it
  - [x] Interface billboard and scene
  - [x] Fix position of loading gif
  - [x] Add sphere in the scene according to the billboard
  - [x] Modernize the code of viewer.js
    - [x] Var to const/let
    - [x] Get rid of JQuery

**Before merging**
Rename request_methods_module.js en  request_methods.js

Change the branch on doc.php

**Link to test**
https://cyberbotics.com/doc/guide/aibo-ers7?version=feature-docs-wrenjs

**Remarks**

- To not break the offline documentation on develop, I was forced to restore the old request_methods on the server. The new request_methods is called request_methods_module on the server.

- The offline documentation does not seem compatible with the use of wrenjs due to the presence of modules.

- I had to force the generated x3d files to be 'NUE' to match the robots, but we should remove this in `webots/scripts/web_component_studio/worlds/web_component_studio_template.wbt` when the conversion to 'ENU' is done.

- When generating the .x3d world file for an ENU robot, we also need to rotate the floor. To do that, add `rotation 1 0 0 1.5708` in the Solid of `webots/scripts/web_component_studio/worlds/web_component_studio_template.wbt`. We also need to change the directional light


omichel (dismissed):
It looks good to me.
I saw a number of problems, like when rotating the propellers of the Mavic drone, but none of them are a regression, so this is fine.

omichel (approved):

PR #3079, opened 2021-05-18
Publish Webots controller to independent repository

lukicdarkoo:
We need C/C++ libs to build the new `webots_ros2` interface:
https://github.com/cyberbotics/webots_ros2/pull/221

### DIFF
--- a/.github/workflows/test_suite_linux.yml
+++ b/.github/workflows/test_suite_linux.yml
@@ -55,6 +55,8 @@ jobs:
         export LIBGL_ALWAYS_SOFTWARE=true
         export ROS_DISTRO=${{ matrix.ROS_DISTRO }}
         xvfb-run --auto-servernum make distrib -j4
+    - name: Deploy Webots Controller
+      run: scripts/packaging/sync_controller_lib.sh
     - name: Rename Webots Tarball Package
       if: ${{ matrix.os == 'ubuntu-18.04' }}
       run: |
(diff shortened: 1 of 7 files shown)


lukicdarkoo:
The Windows deployment has passed in bc537b1 and Linux/macOS in e4d846a. Deployment is limited now only during the nightly builds.

omichel:
That looks very good.
However, I have a question: did you consider committing the source files instead of the binary files for the libController? Committing source files would be easier (it has to be done only once, not for all platforms) and the binaries should be easy to reconstruct from the source on the webots_ros2 side. Also committing binary files on GitHub looks a little bit odd...

lukicdarkoo:
> That looks very good.
> However, I have a question: did you consider committing the source files instead of the binary files for the libController? Committing source files would be easier (it has to be done only once, not for all platforms) and the binaries should be easy to reconstruct from the source on the webots_ros2 side. Also committing binary files on GitHub looks a little bit odd...

I considered it, but I haven't done an in-depth evaluation. Since we are already compiling Webots here I thought that is simpler for the end-users to have dynamic libraries.

Should I reconsider it? If you don't foresee problems with compilation I can try deploying c/cpp files.

omichel:
I don't forsee any problem with the compilation of the C and C++ libraries. But my question was just out of curiosity. To me both options are equally fine, and in any case, it will always be possible to change later on if needed.

omichel (approved):

lukicdarkoo:
If we choose to distribute .c/.cpp files then I would rather configure the CI in the `webots-controller` repo. That way we keep the `webots-controller` repo self-contained and independent from the main repository.

PR #3080, opened 2021-05-19
Increase default mass for intemediary solids in `HingeJointWithBacklash` and `Hinge2JointWithBacklash`

ad-daniel:
**Description**
As reported by a Robocup user (Uli), the default mass used in the intermediary solids of the `HingeJointWithBacklash` and `Hinge2JointWithBacklash` might be too small for most use-cases and could lead to instabilities.


omichel (approved):
Thank you.

PR #3083, opened 2021-05-20
Fixed compilation problem in WbPrecision.hpp

omichel:
See https://github.com/RoboCup-Humanoid-TC/webots/pull/112
Apparently this include is needed on some systems.


lukicdarkoo (approved):
Thank you

PR #3084, opened 2021-05-20
Fix potential crash on Windows

ShuffleWire:
Add a missing break statement.
Otherwise on Windows displaying we run the default statement, which raise an assert.


omichel (approved):
Well spotted. Thank you!

PR #3085, opened 2021-05-21
Run Webots step() in ROS controller by default

lukicdarkoo:
Incrementing `mStep` in the #3040 made the `ros` controller wait for the `time_step` service by default instead of running normally. This PR reverts the behavior, but I am not sure what was the purpose of the `mStep != 0` in the first place.


omichel (approved):
Thank you.

PR #1456, opened 2020-03-13
Enhancement display triangles geometry node

remi-cyberbotics:
Geometry node now displays the triangle count for its element.

This pull-request fixes issue #1323

Triangle Count

PR #3086, opened 2021-05-21
Fix triangle count

mael25:
**Description**
The triangle count for `Mesh`/`IndexedFaceSet` shows `0` when there is too many triangles. 

noshadow

This is because the count function relies on the size of the `triangles` vector from the `wren`. However, the renderer currently does not  display shadows  for meshes with more than (2^16-1)/3 triangles, to avoid an overflow of the index. Thus the `triangles` vector is empty.

This PR instead directly use the `indexCount` (divided by 3) to get the number of triangles. Also, it displays `(no shadow)` next to the number of triangles if it is beyond the limit for which Webots casts shadows.

corrected

**Linked PR**
[Enhancement display triangles geometry node](https://github.com/cyberbotics/webots/pull/1456)

### DIFF
--- a/src/webots/nodes/WbGeometry.cpp
+++ b/src/webots/nodes/WbGeometry.cpp
@@ -44,7 +44,7 @@
 // Constant used to scale down the line scale property
 const float WbGeometry::LINE_SCALE_FACTOR = 250.0f;
 
-const int gMaxIndexNumberToCastShadows = (2 << 14) - 1;  // 2^15 - 1 (16-bit resolution)
+const int gMaxIndexNumberToCastShadows = (2 << 15) - 1;  // 2^16 - 1 (16-bit resolution)
 
 int WbGeometry::maxIndexNumberToCastShadows() {
   return gMaxIndexNumberToCastShadows;
@@ -559,7 +559,7 @@ bool WbGeometry::isAValidBoundingObject(bool checkOde, bool warning) const {
 
 int WbGeometry::triangleCount() const {
   if (areWrenObjectsInitialized() && this->wrenMesh())
-    return wr_static_mesh_get_triangle_count(this->wrenMesh());
+    return wr_static_mesh_get_index_count(this->wrenMesh()) / 3;
   else
     return 0;
 }
--- a/src/webots/nodes/WbGeometry.cpp
+++ b/src/webots/nodes/WbGeometry.cpp
@@ -44,7 +44,7 @@
 // Constant used to scale down the line scale property
 const float WbGeometry::LINE_SCALE_FACTOR = 250.0f;
 
-const int gMaxIndexNumberToCastShadows = (2 << 15) - 1;  // 2^16 - 1 (16-bit resolution)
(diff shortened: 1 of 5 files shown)


mael25:
Let me know if you have another idea on how to notify the user when the `Mesh`/`IndexedFaceSet` is too big to display a shadow.

src/webots/nodes/WbGeometry.cpp
```diff
@@ -44,7 +44,7 @@
 // Constant used to scale down the line scale property
 const float WbGeometry::LINE_SCALE_FACTOR = 250.0f;
 
-const int gMaxIndexNumberToCastShadows = (2 << 14) - 1;  // 2^15 - 1 (16-bit resolution)
```
ShuffleWire, line 47: Would not be better to use `(1 << 16) - 1` instead of `(2 << 15) - 1` ? imo, it match more closely to the `16` in `2^16 -1`

src/webots/scene_tree/WbNodeEditor.cpp
```diff
@@ -146,9 +146,14 @@ void WbNodeEditor::update() {
...
   const WbGeometry *node = dynamic_cast<WbGeometry *>(mNode);
-  if (node && !node->isUseNode())
-    mNbTriangles->setText(tr("Triangle count: %1").arg(node->triangleCount()));
-  else
+  if (node && !node->isUseNode()) {
+    const int maxTriangleNumberToCastShadows = node->maxIndexNumberToCastShadows() / 3;
+    int triangleCount = node->triangleCount();
+    if (triangleCount > maxTriangleNumberToCastShadows) {
+      mNbTriangles->setText(tr("Triangle count: %1 (no shadow)").arg(triangleCount));
+    } else
```
omichel, line 154:
```suggestion
    if (triangleCount > maxTriangleNumberToCastShadows)
      mNbTriangles->setText(tr("Triangle count: %1 (no shadow)").arg(triangleCount));
    else
```

src/webots/nodes/WbGeometry.cpp
```diff
@@ -44,7 +44,7 @@
...
+const int gMaxIndexNumberToCastShadows = (2 << 15) - 1;  // 2^16 - 1 (16-bit resolution)
```
omichel, line 47: Why did you increase this value?

src/webots/nodes/WbGeometry.cpp
omichel: I agree  `(1 << 16) - 1` is better than `(2 << 15) - 1`. However `(uint16_t)(-1)` looks a little bit too hacky...

omichel (changes requested):

src/webots/nodes/WbGeometry.cpp
mael25: The shadow is disappearing when there is more than 21845 triangles, which means maximum 65535 = 2^16-1 indices. So I deduced from this there was an error in the value of `gMaxIndexNumberToCastShadows` (which was not used anyway in the original version of the code).

omichel (approved):
Thank you!

PR #3087, opened 2021-05-25
Add JavaScript support for procedural PROTO

ad-daniel:
**Description**
This PR adds JavaScript support for Procedural PROTO files. Here, only the test suite has been converted. The conversion of the remaining library is done in a different PR.

As it will become clear in the review, supporting two different scripting languages (and specifically two different tokens) leads to some awkward choices like having to set the tokens multiple times during the execution (assumptions about the Template tokens are made long before a procedural protos are generated). To completely refactor the structure didn't make much sense since ultimately Lua will be dropped in the future. This means that until then there will be some additional overhead because of this.

**Performance**
Performance-wise, some degradation compared to Lua has been identified in certain cases:
- PROTO files that have many expressions. A typical case is where the IndexedFaceSet is built in a procedural way, coordinate by coordinate. A notable example is the `Fence.proto` that involves 20'000-50'000 javascript expression evaluations just to define the shape.
- Worlds that rely on the `Forest.proto`. Since it isn't possible to natively manipulate files in the engine, the text files that define the forests have been moved inside the world file. Since some of them involve 50'000+ trees, the parsing of these worlds drops to unacceptable levels.

Both issues are addressed in the relevant PR #3156 as it's more pertinent.

**Docs**
Concerning the docs, the functions exported by the modules have been also documented. Since this did not exist for Lua and nothing similar exists for JavaScript in the current docs, the route I took was to do this like this:

I believe having these tabs makes it more intuitive rather than having one huge list, but I'm open to do it differently if you don't think it's a good idea

docs

**Lua end of support**

When Lua support is dropped, the following must be done:
- in the test suite:
   - remove all PROTO files ending in `*Lua.proto`
   - remove all worlds ending in `*lua.wbt`
- in all PROTO files, remove the `# template language: javascript` from the header
- search if any PROTO files with template tokens `%{   }%` persist. If so, delete or convert them.
- in `WbProtoModel.cpp`
   - in `WbProtoModel::WbProtoModel`: basically undo all changes made in this PR and replace the hard-coded tokens with the new ones
   - in `WbProtoModel::generateRoot`: remove member variable mTemplateLanguage and anything else that uses it
- in `WbTokenizer.cpp`: undo all changes made by this PR
- in `WbProtoTemplateEngine.cpp`: remove the `convertStatementFromJavaScriptToLua` function.
- in `WbTemplateEngine.cpp`: 
   - remove references to lua (`generateLua`, `initializeLua` and global variables.)
   - remove `copyModuleToTemporaryFile`


ad-daniel:
The failing test is unrelated to this PR, the test suite passes successfully.

docs/reference/javascript-procedural-proto.md
```diff
@@ -0,0 +1,833 @@
+## Procedural PROTO Nodes
+
+The descriptive power of PROTO files can be significantly extended by generating object in a procedural way.
+Procedural PROTO nodes can be created using JavaScript as a scripting language.
+Introducing and learning JavaScript is outside the scope of this document.
+
+### Template Engine
+
+What determines if a PROTO file is procedural or not, is whether template statements exist in its body, in other words if JavaScript expressions encapsulated either between the tokens `%<` and `>%` or `%<=` and `>%`.
+Webots cannot load procedural PROTO nodes directly, therefore the procedural PROTO is first translated to pure JavaScript and the resulting script is evaluated by a JavaScript Engine, in Webot's case using QJSEngine, and the result of this process yields a VRML97 compatible PROTO file which is then loaded by Webots as any other non-procedural would.
```
omichel, line 10:
```suggestion
Webots cannot load procedural PROTO nodes directly, therefore the procedural PROTO is first translated to pure JavaScript and the resulting script is evaluated by a JavaScript engine, in Webot's case using QJSEngine, and the result of this process yields a VRML97 compatible PROTO file which is then loaded by Webots as any other non-procedural would.
```

docs/reference/javascript-procedural-proto.md
```diff
@@ -0,0 +1,833 @@
...
+
+  // access functions
+  let number = wbrandom.integer(0, 10);
+
+  // to import only specific functions instead of the entire module
+  import {integer} from 'wbrandom.js';
+}%
+```
+
+The available modules are the following.
```
omichel, line 114:
```suggestion
The available modules are the following:
```

docs/reference/procedural-proto-nodes.md
```diff
@@ -5,197 +5,10 @@ In this way, the PROTO node may contain constants, mathematic expressions, loops
...
-      }
-    ]
-    name IS name
-    boundingObject USE SIMPLE_STAIRS_GROUP
-    physics IS physics
-  }
-  # template statements can be used up to there
-}
-```
+> **Note**: Lua support will be dropped in future releases, wherever possible [JavaScript](javascript-procedural-proto.md) should be preferred instead.
```
omichel, line 14: That seems a bit redundant with line 9...

tests/protos/worlds/template_proto_follow_solid_LUA.wbt
```diff
@@ -0,0 +1,66 @@
+#VRML_SIM R2021b utf8
```
omichel, line 1: The name of this file doesn't seem to be consistent with others (LUA in capital letters).

src/webots/core/WbTemplateEngine.cpp
```diff
@@ -108,7 +138,183 @@ const QString &WbTemplateEngine::closingToken() {
...
+    gOpeningToken = "%{";
+    gClosingToken = "}%";
+
+    result = generateLua(tags, logHeaderName);
+  } else {
+    static bool firstJavaScriptCall = true;
+    if (firstJavaScriptCall) {
+      initializeJavaScript();
+      firstJavaScriptCall = true;
+    }
```
BenjaminDeleze, line 160:
```suggestion
    static bool firstJavaScriptCall = true;
    if (firstJavaScriptCall) {
      initializeJavaScript();
      firstJavaScriptCall = false;
    }
```

BenjaminDeleze (changes requested):
Seems good to me, there is just two small things that I am not sure about

BenjaminDeleze (approved):
Seems good to me

omichel (approved):
I tested it on Windows and it works nicely.

PR #3089, opened 2021-05-25
Upgrade to Qt 5.15.2 on linux and add Qml module

ad-daniel:
**Description**
Upgrade Qt to 5.15.2 under linux and add Qml module needed for #3087.


omichel (approved):
Thank you.

PR #3098, opened 2021-05-26
Hotfix crash due to relative velocities viewer

### LINKED ISSUE #3041: Crash when changing velocity reference
**Describe the Bug**
Webots crash when trying to get the velocity of an object relative to is parent.

**Steps to Reproduce**
1. Load the following world, no need to run simulation : 
```
#VRML_SIM R2021a utf8
WorldInfo {
}
Viewpoint {
}
Robot {
  children [
    Solid {
      children [
        Solid {
          physics Physics {
          }
        }
      ]
    }
  ]
}
```
2. Expand the whole tree
3. Click on the most inner Solid
4. Go to the "velocity" tab
5. Change "absolute" to "Relative to Solid"
6. Webots crash (reproducible crash, got it each time)

**Expected behavior**
Velocity should be displayed relatively to the parent solid.

**System**
 - Operating System: Ubuntu 20.04 LT and Debian 11 (both crash the same way)
 - Webots built today from master

**Additional context**
Running webots under gdb bring me some infos : 
1. The bug is a segfault, due to NULL pointer being de-referenced.
2. Adding an assert on line 1131 on WbSolid.cpp, it's triggered when following the step, which should not.
```
 } else if (parentSolid != NULL) {  // in case of dynamic solid, the velocity is already absolute
    while (!solid->isTopSolid() && solid != parentSolid)
      solid = solid->upperSolid();
+   assert(solid->solidMerger() != NULL);
    l -= solid->solidMerger()->solid()->linearVelocity();
  }
```
When removing the physic node, I don't get a crash.
It seems that the object become (correct me if I'm wrong) a kinematic object, hence `solid->isKinematic()` is true and the code reported above is not ran.

stefaniapedrazzi:
Fix #3041: fix use of invalid solid merger.

### DIFF
--- a/docs/reference/changelog-r2021.md
+++ b/docs/reference/changelog-r2021.md
@@ -51,6 +51,7 @@ Released on June, Xth, 2021.
     - Added a version of the [RobotisOp2](../guide/robotis-op2.md) modeled using [Hinge2Joint](hinge2joint.md) on the ankles, hips, and neck ([#2861](https://github.com/cyberbotics/webots/pull/2861)).
     - Made the `static` behavior the default for PROTO files and removal of the tag. Non static cases must be labeled as such using the `nonDeterministic` tag instead ([#2903](https://github.com/cyberbotics/webots/pull/2903)).
   - Bug fixes:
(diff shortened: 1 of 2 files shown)


ShuffleWire:
So nice, thank !

omichel (approved):
Thank you.

PR #3099, opened 2021-05-26
Fix mismatch between slugify function in documentation test

stefaniapedrazzi:
The `slugify` function used in `showdown-extension-js` to create the documentation HTML code doesn't match the one used in the documentation tests.
As a result wrong anchors that won't work in the documentation are passing the tests instead of the correct ones.
In particular there are issues with text in parenthesis.

Also `showdown-extensions.js` doesn't properly convert the `+` symbols in its slugify function but just the first occurrence.


stefaniapedrazzi:
I tested locally and all the modified references are working correctly.
But on the online version based on the cdn files the `showdown-extension.js` has still not been updated and the figure reference is still wrong for tutorial 6.

omichel (approved):
Thank you.

PR #3103, opened 2021-05-28
Release python global interpreter lock when running c++ code

### LINKED ISSUE #2057: Python interpreter does not runs between controller steps
**Describe the Bug**
If starting  a thread in a python controller, the thread is not allowed to run between controller steps. This is not the expected behaviour of a python script and breaks communication with external sytems. For example if the thread is supposed to receive UDP messages from another system, the UDP messages are not going to be received when expected

Most probably there is some C/C++ code sleeping with the Python lock. This must be fixed

**Steps to Reproduce**

Start webots and run the following script. The Thread is supposed to increment a counter 100 times a second, but in reality it does not even run once every second controller step (This is in fact even stranger, I was expected it to run a little at every step))

```
import time
from threading import Thread

class T(Thread):
    def __init__(self):
        Thread.__init__(self)
        self.start_time = None
        self.counter = 0
        self.ev_stop = False

    def run(self):
        self.start_time = time.time()
        while not self.ev_stop:
            time.sleep(0.01)
            self.counter += 1
            print(time.time())

    def __enter__(self):
        self.start()

    def __exit__(self, *args):
        self.ev_stop = True
        self.join()
        total = time.time() - self.start_time
        print(f"\n\n time elapsed: {total}s, nb counts: {self.counter}\n, counter per seconds: {self.counter / total}\n\n")


s = Supervisor()

with T() as t:
    while s.step(500) != -1:
        pass

```

**Expected behavior**

The Python thread should run freely independently of the controller period

**Additional context**
This seems to be related to https://github.com/cyberbotics/webots/issues/1903
With a very long controller period it becomes easy to reproduce it. Probably when the ctrl-C happens while the C code has the python lock or something similar

sigmunau:
Fixes #2057

### DIFF
--- a/src/controller/python/controller.i
+++ b/src/controller/python/controller.i
@@ -16,7 +16,7 @@
 /* Description:  Swig interface which maps the OO C++ files into a python module called "controller"   */
 /*******************************************************************************************************/
 
-%module controller
+%module(threads=1) controller
 
 %begin %{
 #define SWIG_PYTHON_2_UNICODE


omichel:
Thank you for this nice contribution. I would however prefer to target it at the develop branch since we are not going to release any revision for R2021a now (master branch) and we are going to release R2021b (based on develop) in June. So I moved your patch to #3104.

PR #3106, opened 2021-05-28
Feature api webcomponent

BenjaminDeleze:
Add the following API functions to the WebotsStreaming webcomponent:

- HideToolbar
- ShowToolbar
- SendMessage: allow the user to send message through the websocket, to play, pause, change the controller,...

An example of usage is present here:  https://cyberbotics1.epfl.ch/open-roberta
with the configuration file here : https://cyberbotics1.epfl.ch/open-roberta/setup_viewer.js


omichel (approved):
Thank you.

PR #3109, opened 2021-05-31
Fix search for proto parameter node instance

stefaniapedrazzi:
Fix crashes and assertion failures with deeply nested PROTO parameters
* [x] fix bug introduced in #3036:
    it is not enough to only inspect the first PROTO parameter instance because it could not be finalized, but all the instances have to be inspected.
* [x] fix search of PROTO parameter node instance in node pane causing assertion failures when selecting for example the "Position" tab:
    I now used the same function to get the finalized PROTO parameter instance everywhere in the code. The scene tree code is no longer highlighting the selected parameter in the 3D only if it used only once, but it is now highlighting the first instance.


stefaniapedrazzi:
All the tests passed but on Windows.
There were also issues deleting the macOS artifact but this is not related to this PR.

PR #3119, opened 2021-06-01
Enhancement guided tour

BenjaminDeleze:
I propose to improve the guided tour by :

- Replacing the nao_demo.wbt by nao_room.wbt. They have both a nao with the same controller, but the room world is nicer than a simple floor.
- Modernizing the soccer.wbt by changing the field for the new one made for robocup.

Old:

New:

Old:

New:


omichel (approved):
Thank you.

PR #3124, opened 2021-06-02
Removed use of undefined variable

omichel:
This `$(var)` variable is not defined and seems useless.


BenjaminDeleze (approved):
Thank you!

PR #3125, opened 2021-06-02
Fixed several memory leaks in supervisor API

omichel:
Some small memory chunks for strings were allocated but not free as they should.
This PR fixes 3 different leaks.


BenjaminDeleze (approved):
Thank you!

omichel:
The test suite on Linux is returning a wrong status although all the tests passed successfully.
This problem is not specific to this PR and was seen in other recent PRs.
I will merge it and open an issue about this problem.

PR #3126, opened 2021-06-02
Fix test suite on windows

omichel:
The test suite on Windows was not running smoothly. This PR fixes the following issues:

- [x] The Makefile targets are non-consistent and wrong, typing `make` without a target is causing failures.
- [x] The `generic_sensor.c` is producing compilation warnings.
- [x] The `extern` test is hanging.
- [x] The `robot_urdf` test is failing.


omichel:
The Linux test fails because of #3127.

src/webots/maths/WbPrecision.cpp
```diff
@@ -46,6 +48,15 @@ QString WbPrecision::doubleToString(double value, Level level) {
...
       return r;
     }
+    case FLOAT_ROUND_6: {
+      QString str = QString::number(value, 'f', 6);
+      str.remove(QRegExp("0+$"));   // Remove any number of trailing 0's
+      str.remove(QRegExp("\\.$"));  // If the last character is just a '.' then remove it
+      if (str == "-0")
+        return "0";
+      else
+        return str;
```
lukicdarkoo, line 58:
```suggestion
      return str;
```

src/webots/maths/WbPrecision.cpp
```diff
@@ -46,6 +48,15 @@ QString WbPrecision::doubleToString(double value, Level level) {
       assert(0);
       return r;
     }
+    case FLOAT_ROUND_6: {
+      QString str = QString::number(value, 'f', 6);
+      str.remove(QRegExp("0+$"));   // Remove any number of trailing 0's
+      str.remove(QRegExp("\\.$"));  // If the last character is just a '.' then remove it
```
lukicdarkoo, line 54: Since this is often used it might be a good idea to store the compiled expression (because of performance). Also, I would merge those two (if possible) because of performance.

lukicdarkoo (changes requested):
Looks good. I added a few suggestions for small optimizations

lukicdarkoo:
> The Linux test fails because of #3127.

The Linux test fails due to:
```
FAILURE with robot_urdf_tiago: URDF verification failed
```
Maybe there are hidden errors, but this one should be solved

omichel:
The tests are now all passing.

lukicdarkoo (approved):
Thank you

PR #3134, opened 2021-06-04
Fix world update error Mir100

ad-daniel:
**Description**
World-update test fails on develop because the `keybody_controller` is already the default controller in the PROTO so no need to have it in the world as well.


mael25:
Thank a lot! I just noticed that there is a missing end parenthesis in the description of the PROTO that is displayed on the Webots interface. Could you add it on `l.6` of MiR.proto, so that we don't need to open another PR ?

l.5 Designed by Mobile Industrial Robotics (MIR), MiR100 is a six-wheeled robot (4 caster wheels and 2 regular ones at the
l.6 center`)`. Used in industrial environments, it can carry a payload of up to 100kg.

ad-daniel:
Done, I put it ready for review since the world update test takes hours

mael25 (approved):
Thank you!

PR #3137, opened 2021-06-04
fix matlab makefile

BenjaminDeleze:
**Description**
It seems that `which python` is not available by default on ubuntu 20.
We should use `which python3` instead, moreover it is more pertinent as the mgenerate.py script is not compatible with python 2.7.

### DIFF
--- a/src/controller/matlab/Makefile
+++ b/src/controller/matlab/Makefile
@@ -26,7 +26,7 @@ PYTHON_OK := $(shell which python3)
 release debug profile:
 	@echo "# generating matlab"
 ifeq ('$(PYTHON_OK)','')
-	@echo "# \033[0;33mUnable to find python, skipping matlab API\033[0m"
+	@echo "# \033[0;33mUnable to find python3, skipping matlab API\033[0m"
 else
 	$(shell python3 mgenerate.py;)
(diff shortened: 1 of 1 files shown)


mael25:
I think you have also to run the script with python3 in `l.31` for it to work: `$(shell python3 mgenerate.py;)`

mael25:
Thanks it works! Do you think it would be better to separate the two cases so that the script is not called twice?

src/controller/matlab/Makefile
```diff
@@ -22,13 +22,13 @@ endif
 
 include $(WEBOTS_HOME_PATH)/resources/Makefile.os.include
 
-PYTHON_OK := $(shell which python)
+PYTHON_OK := $(shell which python3)
 release debug profile:
 	@echo "# generating matlab"
 ifeq ('$(PYTHON_OK)','')
 	@echo "# \033[0;33mUnable to find python, skipping matlab API\033[0m"
```
omichel, line 29:
```suggestion
	@echo "# \033[0;33mUnable to find python3, skipping matlab API\033[0m"
```

omichel (approved):
That looks good to me.

PR #3140, opened 2021-06-07
Generate mir100 and ned web robots components.

BenjaminDeleze:
**Description**
I added a description of the mir100 and the ned in the `components.json` file such that we are now able to automatically regenerate them (it will be useful in #2787). 

Moreover, the visualization of the ned in the docs was not updated for the new version version of the viewer and was not working anymore on develop. This PR correct that.

You can see the result here:
-  ned: https://cyberbotics.com/doc/guide/ned?version=docs-generate-mir-ned
- mir100: https://cyberbotics.com/doc/guide/mir100?version=docs-generate-mir-ned


mael25 (approved):
Thank you!

PR #3147, opened 2021-06-09
remove nodejs

BenjaminDeleze:
Nodejs was used to minimify webots.min.js, but as we do not use minification anymore, it is now useless and we can remove it

- [x] Remove it from the sources
- [x] Modify the wiki to inform that nodejs is needed only for master

It will fix #3147


lukicdarkoo (approved):
Thank you, it looks good (let's just see what the tests say)

PR #3150, opened 2021-06-09
Fix asymmetric friction

mael25:
**Description**
The asymmetric friction force direction is wrongly computed for `Cylinder` and `Box`.

Currently, the `computeFrictionDirection(const WbVector3 &normal)` of both of them try first to express the normal to the contact point into the local coordinates system of the `Cylinder`/`Box`.  However to do so, it uses `matrix().extracted3x3Matrix()` as transformation matrix, which corresponds to Local-->Global.

`WbVector3 localNormal =  matrix().extracted3x3Matrix() * normal;`

The issue is corrected by using the inverse matrix, which is here simply the transpose:

`WbVector3 localNormal = normal * matrix().extracted3x3Matrix();`.

Besides, in SimulationCluster.cpp, the naming of 

`localFdirS1 = wg2->matrix().extracted3x3Matrix() * wg2->computeFrictionDirection(contactNormal);`

has been changed to globalFdirS1 for clarification (it is expressed in the global coordinates system, not local)

**Tasks**
  - [x] Correct localNormal 
  - [x] Changed localFdirSx to globalFdirSx


omichel (changes requested):
Thank you.
I believe we should also add a line in the change log as it is an important bug fix.

omichel (approved):
Thank you.

PR #3154, opened 2021-06-10
Enhancement converter worlds

mael25:
**Description**
Added the conversion of the `Viewpoint` when running the scripts to convert the `coordinateSystem` of a world.


mael25:
Wrong targeted branch

PR #3155, opened 2021-06-10
Added Viewpoint in world converter scripts

mael25:
**Description**
Added the conversion of the `Viewpoint` in the scripts to convert the `coordinateSystem` of a world.

**Tasks**

- [x] Viewpoint
- [x] Correct doc of viewpoint.md (default `position` field) to match reality

Note: if no `orientation`/`position` are given, it will adopt the `Viewpoint` as seen from a `NUE` world with the default Webots parameters. So for EUN-->NUE nothing is done, and for NUE-->ENU the default Viewpoint is adapted.

### DIFF
--- a/scripts/converter/convert_eun_to_nue.py
+++ b/scripts/converter/convert_eun_to_nue.py
@@ -58,7 +58,7 @@ def convert_to_nue(filename):
                     field['value'] = rotation(field['value'], [0, 1, 0, 0.5 * math.pi])
                 elif field['name'] in ['position']:
                     field['value'] = [field['value'][2], field['value'][1], str(-float(field['value'][0]))]
-        elif node['name'] not in ['TexturedBackground', 'TexturedBackgroundLight']:
+        elif node['name'] not in ['TexturedBackground', 'TexturedBackgroundLight', 'PointLight']:
             print('Rotating', node['name'])
             rotation_found = False
             for field in node['fields']:
--- a/scripts/converter/convert_eun_to_nue.py
+++ b/scripts/converter/convert_eun_to_nue.py
@@ -53,23 +53,11 @@ def convert_to_nue(filename):
                                        'type': 'SFString'})
         elif node['name'] == 'Viewpoint':
             print('Rotating', node['name'])
-            orientation_found = False
-            position_found = False
             for field in node['fields']:
                 if field['name'] in ['orientation']:
-                    orientation_found = True
                     field['value'] = rotation(field['value'], [0, 1, 0, 0.5 * math.pi])
                 elif field['name'] in ['position']:
-                    position_found = True
                     field['value'] = [field['value'][2], field['value'][1], str(-float(field['value'][0]))]
-            if not orientation_found:
-                node['fields'].append({'name': 'orientation',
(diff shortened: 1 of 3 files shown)


scripts/converter/convert_eun_to_nue.py
```diff
@@ -51,7 +51,14 @@ def convert_to_nue(filename):
...
                                        'value': 'NUE',
                                        'type': 'SFString'})
-        elif node['name'] not in ['Viewpoint', 'TexturedBackground', 'TexturedBackgroundLight']:
+        elif node['name'] == 'Viewpoint':
+            print('Rotating', node['name'])
+            for field in node['fields']:
+                if field['name'] in ['orientation']:
+                    field['value'] = rotation(field['value'], [0, 1, 0, 0.5 * math.pi])
+                elif field['name'] in ['position']:
+                    field['value'] = [field['value'][2], field['value'][1], str(-float(field['value'][0]))]
```
lukicdarkoo, line 60: Here we are missing a conversion when there is no orientation or position defined

lukicdarkoo (changes requested):
Looks good, there only two small details

scripts/converter/convert_eun_to_nue.py
mael25: If no orientation or position are defined, I chose to set the `Viewpoint` to be as we would see it with the default parameters of Webots in a `NUE` world. So for EUN-->NUE nothing is done, and for NUE-->ENU the default Viewpoint is adapted.

Let me know if you have another idea of implementation!

scripts/converter/convert_nue_to_enu.py
```diff
@@ -42,7 +42,26 @@ def convert_to_enu(filename):
...
+                    field['value'] = [field['value'][0], str(-float(field['value'][2])), field['value'][1]]
+            if not orientation_found:
+                node['fields'].append({'name': 'orientation',
+                                       'value': ['1', '0', '0', str(0.5 * math.pi)],
+                                       'type': 'SFRotation'})
+            if not position_found:
+                node['fields'].append({'name': 'position',
+                                       'value': ['0', '-10', '0'],
+                                       'type': 'SFVec3f'})
+        elif node['name'] not in ['TexturedBackground', 'TexturedBackgroundLight']:
```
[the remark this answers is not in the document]
mael25: I guess we can also add it to convert_eun_to_nue.py ?

lukicdarkoo (approved):
Thank you

mael25:
> Thank you

Thanks for the review! Do you think it is needed to also make this change for `convert_eun_to_nue.py `?

`elif node['name'] not in ['TexturedBackground', 'TexturedBackgroundLight', 'PointLight']:`

lukicdarkoo:
> > Thank you
> 
> Thanks for the review! Do you think it is needed to also make this change for `convert_eun_to_nue.py `?
> 
> `elif node['name'] not in ['TexturedBackground', 'TexturedBackgroundLight', 'PointLight']:`

Yes, it would be good

lukicdarkoo (approved):

Issue #3052, opened 2021-05-10
Add Supervisor API function to artificially set joints position


stefaniapedrazzi:
Currently with the Supervisor API it is possible to artificially set the joint position but only if the `JointParameter` node is visible in the scene tree. Internal PROTO nodes can be retrieved using the Supervisor API but they are read-only and thus cannot be modified. This means that this approach doesn't work for most of the built-in Webots robots where the `JointParameter` is defined inside the PROTO definition.

The alternative method is to use the Motor API to move the robot to the initial position, but:
  - even setting the PID to high value it may take at least one step or more: i.e. may be needed to reset the physics of the robot and reset other nodes that moved in the meanwhile
  - if separate Supervisor and Robot controller are used it requires additional communication

So, now that almost all the robots are PROTO nodes, it seems that will be helpful for users to have a method to artificially set the position of joints, something like:
`wb_supervisor_node_set_joint_position(WbNodeRef joint_node, double position, int index = 0)`
where the index > 0 will be used for Hinge2Joint and BallJoint nodes.

PR #3160, opened 2021-06-10
Feature wb_supervisor_node_set_joint_position

stefaniapedrazzi:
Addresses #3052.
Add Supervisor function to set the joint position of active/passive joints visible in the scene tree or hidden in PROTO body definition.

Tasks:
* [x] Add functionality:
  * [x] if corresponding `JointParameters` node is set
  * [x] if corresponding `JointParameters` node is not set
* [x] C API
* [x] C++ API
* [x]  Java API
* [x] Python API
* [x] matlab API
* [x] ROS API 
* [x] documentation
* [x] add test

To be able to access at least active joint in a user friendly way, these additional changes are applied:
* ``wb_supervisor_node_get_from_device`` is now working also with internal PROTO nodes
* ``wb_supervisor_node_get_parent_node`` is now working also for internal PROTO nodes


docs/reference/supervisor.md
```diff
@@ -1400,6 +1399,102 @@ If the specified node is physics-enabled, i.e. it contains a [Physics](physics.m
...
+%tab-end
+
+%tab "C++"
+
+```cpp
+#include <webots/Node.hpp>
+
+namespace webots {
+  class Node {
+    void setJointPosition(double position, int index = 1);
```
ad-daniel, line 1426:
```suggestion
    void setJointPosition(double position, int index = 0);
```

ad-daniel: On the C++ controller API the default value being applied is 0, not 1. Also, does the C version not impose the same? (and therefore python etc)

ad-daniel (changes requested):
Looks good to me, only a few minor questions/doubts:

- I believe you forgot to commit the `wb_supervisor_node_set_joint_position.m`
- The setting of the joint position appears to ignore any `maxStop`/`minStop` set in the JointParameters. From the tests I made it sets the position beyond these limits, and it stays there for 1 timestep before bouncing away from it, probably due to the violation. Is it something the API should warn/prevent against or is it up to the user? I'm wondering if it might end up "locking" a robot if [these](https://cyberbotics.com/doc/reference/motor?tab-language=c#motor-limits) rules are bypassed by the API

stefaniapedrazzi:
> Looks good to me, only a few minor questions/doubts:
> 
>     * I believe you forgot to commit the `wb_supervisor_node_set_joint_position.m`
After the latest changes on the Matlab API generation, automatically generated files are ignored and no longer committed on the repository. So `wb_supervisor_node_set_joint_position.m` that is automatically generated doesn't need to be committed.

> 
>     * The setting of the joint position appears to ignore any `maxStop`/`minStop` set in the JointParameters. From the tests I made it sets the position beyond these limits, and it stays there for 1 timestep before bouncing away from it, probably due to the violation. Is it something the API should warn/prevent against or is it up to the user? I'm wondering if it might end up "locking" a robot if [these](https://cyberbotics.com/doc/reference/motor?tab-language=c#motor-limits) rules are bypassed by the API

I will check this.

docs/reference/supervisor.md
stefaniapedrazzi: In joints the "position" fields are called (1), 2, and 3.
So the index here is correct and I just changed the code that was not updated.

C, Java, Matlab and ROS APIs don't support default values.
But I will add in the default value in the Python API.

stefaniapedrazzi:
Thank you for the review.
All the comments should be addressed.

ad-daniel (approved):
Looks good to me, thank you!

PR #3161, opened 2021-06-10
Fix `test suite` CI on ubuntu 18

ad-daniel:
**Description**
Dummy branch to figure out why test suite step fails in CI (passes locally) despite passing all the tests.

### DIFF
--- a/tests/test_suite.py
+++ b/tests/test_suite.py
@@ -297,7 +297,7 @@ for groupName in testGroups:
         else:
             lines = open(webotsStdErrFilename, 'r').readlines()
             for line in lines:
-                if 'Failures' in line:
+                if 'Failure' in line:
                     # check if it should be ignored
                     if not any(item in line for item in whitelist):
                         failures += 1
--- a/tests/test_suite.py
+++ b/tests/test_suite.py
@@ -321,6 +321,13 @@ for groupName in testGroups:
                         appendToOutputFile(
                             'Cannot get the core dump file: "%s" does not exist.' % core_dump_file
                         )
+    print('-- STDOUT -----------------------------------------------------------------------------------')
+    with open(webotsStdOutFilename, 'r') as file:
+        print(file.read())
+    print('-- STDERR ------------------------------------------------------------------------------------')
+    with open(webotsStdErrFilename, 'r') as file:
+        print(file.read())
+    print('>> failures: %d' % failures)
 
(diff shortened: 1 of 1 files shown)


ad-daniel:
The culprit is this step [here](https://github.com/cyberbotics/webots/blob/8c347283fb59acd5edef39890354ce6f80508ea7/tests/test_suite.py#L297-L299). The reason it is triggered is because of this line showing up in Webots's stderr:

> 2021-06-10T19:06:37.9977970Z [7201:7208:0610/190537.908943:ERROR:command_buffer_proxy_impl.cc(141)] ContextResult::kTransient**Failure**: Failed to send GpuChannelMsg_CreateCommandBuffer.

I'm not sure if there's anything we can do about it, however whether these sort of failures should interrupt the `test suite` is debatable I suppose.

Either way, for the future at least [this](https://github.com/cyberbotics/webots/blob/8c347283fb59acd5edef39890354ce6f80508ea7/tests/test_suite.py#L301) condition should instead be triggered if `failures > 0` in order to make the debugging more easy.

ad-daniel:
I can only reproduce the problem on ubuntu 18 docker, and the world causing it is `robot_window_html.wbt`.
Even commenting out the send and receive function in the controller's body doesn't prevent it from showing up. The header inclusion is sufficient. 
I've replaced the header `#include <webots/plugins/robot_window/default.h>` with `#include <webots/plugins/robot_window/robot_wwi.h>` to narrow it down further and it still shows up.

ad-daniel:
With this PR if a system failure occurs, it still interrupts the `test suite`, that way we are aware a problem does indeed exist and we print the line that triggered it. However, if like in this case the problem should be ignored, the relevant message can be added to the whitelist so that it doesn't interrupt the `test suite` further.

tests/test_suite.py
```diff
@@ -294,9 +295,13 @@ def monitorOutputFile(finalMessage):
             appendToOutputFile('- expected number of worlds: %d\n' % (worldsCount))
             appendToOutputFile('- number of worlds actually tested: %s)\n' % (counterString))
         else:
-            with open(webotsStdErrFilename, 'r') as file:
-                if 'Failure' in file.read():
-                    failures += 1
+            lines = open(webotsStdErrFilename, 'r').readlines()
+            for line in lines:
+                if 'Failures' in line:
```
omichel, line 300:
```suggestion
                if 'Failure' in line:
```

PR #3162, opened 2021-06-11
Add Contact Points Tracking

lukicdarkoo:
Similar to #2279 we want to provide an API function for contact points tracking.

This PR adds the following API functions:
- `wb_supervisor_node_enable_contact_point_tracking(WbNodeRef node, int sampling_period, bool include_descendants)`
- `wb_supervisor_node_disable_contact_point_tracking(WbNodeRef node, bool include_descendants)`
- `wb_supervisor_node_get_contact_points(WbNodeRef node, bool include_descendants, int *size)`

The `wb_supervisor_node_get_contact_points` function is a substitute for:
- `wb_supervisor_node_get_contact_point`
- `wb_supervisor_node_get_contact_point_node`
- `wb_supervisor_node_get_number_of_contact_points`

Therefore, these functions are deprecated.

- Tracking
  - [x] Basic implementation
    - [x] C++
    - [x] Python
    - [x] Java
    - [x] Matlab
    - [x] ROS
  - [x] Documentation
  - [x] Tests
  - [x] ROS tests
  - [x] Controller.def + WbLanguage.cpp
- `wb_supervisor_node_get_contact_points`
  - [x] Basic implementation
    - [x] C++
    - [x] Python
    - [x] Java
    - [x] Matlab
    - [x] ROS
  - [x] Documentation
  - [x] Tests
  - [x] ROS tests
  - [x] Controller.def + WbLanguage.cpp
- [x] Changelog
- Verify 
  - [ ] Matlab `wb_supervisor_node_get_contact_points`
- Non related fixes
  - [x] Change the ROS sampling_period type for pose tracking
  - [x] Fix pose and field tracking reset handling 
  - [x] Ignore `wb_supervisor_node_set_joint_position` 
  - [x] Change Python API to return an object instead of dictionary

**Documentation**

https://cyberbotics.com/doc/reference/supervisor?version=feature-contact-point-tracking#wb_supervisor_node_get_contact_points

**Development Scenario**

```
#VRML_SIM R2021b utf8
WorldInfo {
  basicTimeStep 16
  coordinateSystem "NUE"
}
Viewpoint {
  orientation 0.2744370696263734 -0.9240498909372512 -0.2661129344352567 4.66421960062771
  position 3.2657338363728203 2.5300277265960878 0.4360826908960025
}
TexturedBackground {
}
TexturedBackgroundLight {
}
Floor {
}
WoodenBox {
}
DEF BOX WoodenBox {
  translation 0 1.31 0.28
  name "wooden box(1)"
  mass 1
}
Robot {
  controller "contacttracking"
  supervisor TRUE
}
```

```python
robot = Supervisor()
use_new_api = True

box = robot.getFromDef('BOX')
box.enableContactPointsTracking(timestep)

    if use_new_api:
        print(box.getContactPoints())
    else:
        n_points = box.getNumberOfContactPoints()
        for i in range(n_points):
            print(f'#{i}: {box.getContactPoint(i)}')
            print(f'#{i}: {box.getContactPointNode(i)}')
```

```java
import com.cyberbotics.webots.controller.Supervisor;
import com.cyberbotics.webots.controller.Node;
import com.cyberbotics.webots.controller.ContactPoint;

public class contacttracking_java {
  public static void main(String[] args) {
    Supervisor robot = new Supervisor();

    int timeStep = (int) Math.round(robot.getBasicTimeStep());

    Node box = robot.getFromDef("BOX");
    
    while (robot.step(timeStep) != -1){
      ContactPoint[] points = box.getContactPoints(false);
      for (int i = 0; i < points.length; i++) {
        System.out.println(points[i].getNodeId());
        System.out.println(points[i].getPoint()[1]);
        }
    }
  }
}
```


lukicdarkoo:
@omichel Each contact point is described with `node_id` and `point` (x, y, z) properties.
https://github.com/cyberbotics/webots/blob/7525206de993a76f6030cee241c2e7c3b843df58/include/controller/c/webots/contact_point.h#L21-L22
I am a bit worried about the `node_id` as it is not a reference, but a node ID. Resolving references in the libcontroller would take additional time, but if users often need to convert the ID to a node reference maybe it is better to do it on the libcontroller side. If it is not something users often do I would prefer to keep `node_id` only.

omichel:
I believe it's better to use a node reference rather than a node ID, as it the default way to refer to a node in the supervisor API and users will most likely need the reference rather than the ID. I believe we can implement a fast way to resolve references (like an array indexed with node IDs that points to node references). But we can do this later if it is truly a bottleneck.

lukicdarkoo:
> I believe it's better to use a node reference rather than a node ID, as it the default way to refer to a node in the supervisor API and users will most likely need the reference rather than the ID. I believe we can implement a fast way to resolve references (like an array indexed with node IDs that points to node references). But we can do this later if it is truly a bottleneck.

Once the `node_id` is received from the simulation its reference has to be found by iterating through a list. Then, in C++, it has to be resolved using the `findNode` method which again iterates through a list:
https://github.com/cyberbotics/webots/blob/55f079a24abddefa9ac5cc4dad7afee98975d990/src/controller/cpp/Node.cpp#L27
Finally, in Python, we have to resolve the reference again (this time not by iterating through a list but adds up)

docs/reference/supervisor.md
```diff
@@ -1113,6 +1149,88 @@ The "[WEBOTS\_HOME/projects/samples/howto/cylinder_stack/worlds/cylinder\_stack.
...
+```
+
+%tab-end
+
+%tab "Java"
+
+```java
+import com.cyberbotics.webots.controller.ContactPoint;
+
+public class CameraRecognitionObject {
```
omichel, line 1202:
```suggestion
public class ContactPoint {
```

omichel:
The [documentation page](https://cyberbotics.com/doc/reference/supervisor?version=feature-contact-point-tracking#contact-point) looks odd:

image

lukicdarkoo:
> The [documentation page](https://cyberbotics.com/doc/reference/supervisor?version=feature-contact-point-tracking#contact-point) looks odd:
> 
> image

Yes, but if you check the TODO list you will see I didn't do the documentation.

src/controller/c/supervisor.c
```diff
@@ -1077,25 +1103,44 @@ static void supervisor_read_answer(WbDevice *d, WbRequest *r) {
...
+        WbNodeRef tmp_node = node_list;
+        while (tmp_node) {
+          if (tmp_node->id == node_id) {
+            contact_point_node = tmp_node;
+            break;
+          }
+          tmp_node = tmp_node->next;
         }
       }
+      assert(contact_point_node);
```
lukicdarkoo, line 1124: @omichel Is it possible it is crashing here?

omichel: It's actually not crashing any more...

docs/reference/changelog-r2021.md
```diff
@@ -129,6 +130,7 @@ Released on June, Xth, 2021.
   - Cleanup
     - Deleted deprecated DifferentialWheels node ([#2749](https://github.com/cyberbotics/webots/pull/2749)).
     - Changed structure of the [projects/samples/howto]({{ url.github_tree }}/projects/samples/howto) directory, so each demonstration is in a dedicated directory ([#2639](https://github.com/cyberbotics/webots/pull/2639)).
+    - Deprecated [`wb_supervisor_node_disable_contact_point_tracking`](supervisor.md#wb_supervisor_node_disable_contact_point_tracking), [`wb_supervisor_node_get_contact_point_node`](supervisor.md#wb_supervisor_node_get_contact_point_node), and [`wb_supervisor_node_get_number_of_contact_points`](supervisor.md#wb_supervisor_node_get_number_of_contact_points) functions in favor of [`wb_supervisor_node_get_contact_points`](supervisor.md#wb_supervisor_node_get_contact_points) ([#3162](https://github.com/cyberbotics/webots/pull/3162)).
```
omichel, line 133: There was no such function: wb_supervisor_node_disable_contact_point_tracking

docs/reference/supervisor.md
```diff
@@ -1094,6 +1122,14 @@ number_of_contacts = wb_supervisor_node_get_number_of_contact_points(node, inclu
 
 *get the contact point with given index in the contact point list of the given solid.*
 
+The `wb_supervisor_node_get_contact_points` function returns a list of contact points.
+
+The `wb_supervisor_node_enable_contact_point_tracking` function forces Webots to stream contact point data to the controller.
+It improves the performance as the controller by default uses a request-response pattern to get data from the field.
+The `sampling_period` argument determines how often the field data should be sent to the controller.
```
omichel, line 1129:
```suggestion
The `sampling_period` argument determines how often the contact point data should be sent to the controller.
```

lib/controller/matlab/.gitignore
```diff
@@ -277,6 +277,7 @@ wb_supervisor_node_is_proto.m
 wb_supervisor_node_move_viewpoint.m
 wb_supervisor_node_set_visibility.m
 wb_supervisor_node_set_velocity.m
+wb_supervisor_node_set_joint_position.m
```
omichel, line 280: This looks unrelated...

src/controller/c/supervisor.c
```diff
@@ -89,21 +89,26 @@ static WbFieldRequest *field_requests_garbage_list = NULL;
 static WbFieldRequest *sent_field_get_request = NULL;
 static bool is_field_immediate_message = false;
 
+typedef struct WbNodeWbContactPointListStructPrivate {
+  int n;
+  WbContactPoint *points;
+  double timestamp;  // TODO: Delete with `wb_supervisor_node_get_contact_point`
```
omichel, line 95: Can't we address this TODO?

src/controller/c/supervisor.c
```diff
@@ -89,21 +89,26 @@ static WbFieldRequest *field_requests_garbage_list = NULL;
...
-  int number_of_contact_points;
-  double *contact_points;           // double[3 * number_of_contact_points]
-  int *node_id_per_contact_points;  // int[number_of_contact_points]
-  double contact_points_time_stamp;
-  bool contact_points_include_descendants;
+  double *position;                                  // double[3]
+  double *orientation;                               // double[9]
+  double *center_of_mass;                            // double[3]
+  WbNodeWbContactPointListStruct contact_points[2];  // 0 -> without descendants, 1 -> with descendants
+  bool contact_points_include_descendants;           // TODO: Delete with `wb_supervisor_node_get_contact_point`
```
omichel, line 111: Can't we address this TODO?

src/controller/c/supervisor.c
```diff
@@ -2079,11 +2152,44 @@ int wb_supervisor_node_get_number_of_contact_points(WbNodeRef node, bool include
...
+
+  const double t = wb_robot_get_time();
+  const int descendants = include_descendants ? 1 : 0;
+
+  if (t <= node->contact_points[descendants].last_update) {
+    *size = node->contact_points[descendants].n;
+    return node->contact_points[descendants].points;
+  }
+
+  // TODO: Delete with `wb_supervisor_node_get_contact_point`
```
omichel, line 2191: We should address this TODO.

omichel (changes requested):
It looks good to me.
From my experiments, the performance increases only if `include_descendants` is False.
I have a few questions and suggestions.

lib/controller/matlab/.gitignore
lukicdarkoo: It was creating an additional file, Benjamin fixed it today in another PR. After a sync we will not need it any more

src/controller/c/supervisor.c
lukicdarkoo: We can delete the `wb_supervisor_node_get_contact_point` functions, but we don't want that :)

src/controller/c/supervisor.c
lukicdarkoo: Those are just guidelines when deleting the `wb_supervisor_node_get_contact_point` function

omichel (approved):
Thank you.

PR #3166, opened 2021-06-11
Fix size animation web component

### LINKED ISSUE #3159: Size of animation canvas
The size of the animation window is set by the web-component whereas the user should be able to chose.
https://github.com/cyberbotics/webots/blob/e49e23e96c628250b330e5c29b529a2e319f952c/resources/web/wwi/WebotsAnimation.js#L5

BenjaminDeleze:
**Description**
Fix the size of the animation to 100% of the web-component. 

It was necessary to also set the width to 100% to allow both `flex` and `block` display to work.

The user can know determine the size of the web-component by changing it in the .html.

Fix #3159.

### DIFF
--- a/resources/web/templates/x3d_playback.html
+++ b/resources/web/templates/x3d_playback.html
@@ -14,7 +14,7 @@
   </head>
   <body>
     <h1 style="height:5%">%title%</h1>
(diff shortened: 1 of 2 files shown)


lukicdarkoo (approved):
Thank you

Issue #3051, opened 2021-05-10
Reset from GUI still not working on certain cases


ad-daniel:
**Describe the Bug**
As it surfaced here #3036, some issues persist on the master branch regarding the reset.

How to reproduce:
1. Load the ` tests/protos/worlds/nested_parameter_transform_node.wbt ` world.
2. Run it and notice the PROTO node move.
3. Reset. The node doesn't return to the initial position.

It's unclear what about it makes it fail, the specific case is a heavily nested PROTO with multiple instances.

ShuffleWire:
I have another case : 
Using situation described in https://github.com/cyberbotics/webots/issues/3163
Proceed until step 4 (create model and controller, and remove the JointParameters)
Run simulation, cylindre should move.
Pause the simulation.
Restart the simulation.
Cylinder don't go back to the original position.
BUT
Hit a few time the "Execute one simulation step" button, and the cylinder will eventually go back to position. Note that when it come back, it doesn't go at the position 0, but a bit more, like if the simulation have been actually running during those few step, and the cylinder "catch up" to the correct position only at this moment.

PR #3167, opened 2021-06-11
Fix incorrect resetting of Transform

ad-daniel:
**Description**
As surfaced in #3051, under certain conditions the resetting (from GUI or from supervisor) does not appear to be working. 
In some cases the objects don't return to their initial position, in others they do but when the position is queried the value is incorrect.

The issue doesn't not appear to be related to the nested-ness of PROTO files as previously thought, chipping away at the example mentioned in #3036 revealed that the MRE is just having a simple structure like 
```
DEF OUTER_NODE Transform {
  children [
    DEF INNER_NODE Transform {
    
    }
  ]
}
```  
In this case, the objects return to their initial position but after a reset their position values are incorrect. Similar situations to `T { T {} }` that also fail are: `T { S {} }`, `G { T { S { } } }`, `S { T { S { } } }` , and `S { T { T { } } }`

Specifically, what appears to be failing is the following condition:
- pick one of the structures mentioned above, let's call the inner node INNER and the outer OUTER.
- get their position: `positions_A`
- move the OUTER (for example using `set_sf_vec3f`)
- get their positions: `positions_B`

So far so good, the INNER is translated as well.

- reset and get the positions: `position_C`. This should be equal to `positions_A` but **isn't.**
- move the OUTER (for example using `set_sf_vec3f`)
- get their positions: `positions_D`. This should be equal to `position_B` and **it is**.

So the issue really is just the first value after the reset.


ShuffleWire:
When you feel the PR ready, please ping me, so I could test it on a project which have a lot of trouble with resetting.

ad-daniel:
@ShuffleWire The PR is ready for review, so you can check on your side. However I updated the description, the conditions of this bug appear to be very specific so it might not apply to your case. Still worth a try, if it doesn't work it might be better to open it in a different issue, preferably with a minimal reproducible example

ShuffleWire:
First, thank for your work :) 
However, even if I've noticed improvement over my use case, I've still have issues.
I've try to forge a minimal example which fail to reset properly.
Per say it's not really "minimal", I will try to work more on it, if you need? However I want to get you quick feedback 

<details>
<summary>world file</summary>
<pre>
#VRML_SIM R2021b utf8
WorldInfo {
}
Viewpoint {
  orientation 0.19943768906045042 -0.4859795677615011 -0.850910375950706 2.295629865564844
  position -8.59124820629042 4.227000712690205 6.770455272724853
}
TexturedBackground {
}
TexturedBackgroundLight {
}
Robot {
  rotation 0 0 1 -1.570865597279716
  children [
    Solid {
      children [
        Solid {
          children [
            Transform {
              children [
                Shape {
                  appearance PBRAppearance {
                  }
                  geometry Box {
                  }
                }
              ]
            }
          ]
          physics Physics {
            density -1
            mass 20
          }
        }
        Solid {
          children [
            Transform {
              translation 3 0 0
              children [
                Shape {
                  appearance PBRAppearance {
                  }
                  geometry Sphere {
                  }
                }
              ]
            }
            HingeJoint {
              jointParameters HingeJointParameters {
                axis 0 1 0
              }
              endPoint SolidReference {
                solidName "solid"
              }
            }
          ]
          name "solid1"
        }
      ]
      name "solid2"
    }
  ]
  physics Physics {
    density -1
    mass 200
  }
}
</pre>
</details>

ad-daniel:
I don't believe the odd behavior of this world is related to the resetting. By adding a Physics node to the solid `solid1`, for instance 
```
physics Physics {
    density -1
    mass 1
}
```
it appears to be working. I suspect closed chain links that rely on solid references as in this case you can't have parts of it without Physics. I don't know if it's a rule, but I've encountered a similar situation before while making a gripper. Either way, let me know if that fixes it for you.

omichel:
> I suspect closed chain links that rely on solid references as in this case you can't have parts of it without Physics. I don't know if it's a rule, but I've encountered a similar situation before while making a gripper. Either way, let me know if that fixes it for you.

Yes, in other words, mechanical closed-loops are not supported in kinematic mode, they work only in dynamic mode.

ShuffleWire:
That not a good news, I've would like to use assembly with close loop without having every part assigned with a physic node (performance reason)...
However, I still don't understand why the reset operation depend on the presence of physic node or not :/

omichel:
> That not a good news, I've would like to use assembly with close loop without having every part assigned with a physic node (performance reason)...

Unfortunately, the kinematics engine doesn't support closed loops...

ShuffleWire:
> > That not a good news, I've would like to use assembly with close loop without having every part assigned with a physic node (performance reason)...
> 
> Unfortunately, the kinematics engine doesn't support closed loops...

Whould it be a lot of work to make it work ?

omichel:
> Would it be a lot of work to make it work ?

I believe it's indeed tricky to implement, and that's not a priority for us right now. But feel free to try to implement it to see by yourself. If you succeed, we will be happy to review your contribution.

ShuffleWire:
Ok
Regarding the initial subject, I guess I will not block this PR, even if I still suspect that their is more about it. However, for now I simply reset the whole world, which does the trick. Will be using it in a CI pipeline (docker), also, and that will definitely not be an issue.

tests/api/worlds/supervisor_reset_transform.wbt
```diff
@@ -0,0 +1,75 @@
+#VRML_SIM R2021b utf8
+WorldInfo {
+  info [
+    "Test world to validate correct resetting of transforms"
```
stefaniapedrazzi, line 4:
```suggestion
    "Test world to validate correct reset of transforms"
```

src/webots/nodes/utils/WbTemplateManager.cpp
```diff
@@ -233,12 +233,13 @@ void WbTemplateManager::regenerateNode(WbNode *node, bool restarted) {
   }
   WbNode::setRestoreUniqueIdOnClone(false);
   int uniqueId = node->uniqueId();
+  QString stateId = node->stateId();
```
stefaniapedrazzi, line 236: These can be const
```suggestion
  const int uniqueId = node->uniqueId();
  const QString &stateId = node->stateId();
```

src/webots/nodes/WbTransform.cpp
```diff
@@ -52,6 +60,24 @@ WbTransform::~WbTransform() {
     wr_node_delete(WR_NODE(wrenNode()));
 }
 
+void WbTransform::reset(const QString &id) {
+  WbGroup::reset(id);
+  // note: for solids, the setting of these parameters has to occur only if mJointParents.size() == 0 and is taken care of in
+  // WbSolid::reset, otherwise it breaks the resetting of hinge based joints
```
stefaniapedrazzi, line 66:
```suggestion
  // note: for solids, the set of these parameters has to occur only if mJointParents.size() == 0 and it is handled in
  // WbSolid::reset, otherwise it breaks the reset of hinge based joints
```

src/webots/nodes/WbTransform.cpp
```diff
@@ -52,6 +60,24 @@ WbTransform::~WbTransform() {
...
+  if (nodeModelName() != "TrackWheel" && nodeModelName() != "Solid") {
+    setTranslation(mSavedTranslations[id]);
+    setRotation(mSavedRotations[id]);
+  }
+}
+
+void WbTransform::save(const QString &id) {
+  WbGroup::save(id);
+  if (nodeModelName() != "TrackWheel") {
```
stefaniapedrazzi, line 75: Why don't you use a dynamic case instead of checking the model name?

Same comment for `WbTransform::reset`.

stefaniapedrazzi (changes requested):
This PR seems good.
I just have some minor comments on the code

src/webots/nodes/WbTransform.cpp
ad-daniel: I couldn't use the nodetype in the init because it wasn't set yet, and ended up doing the same in these functions but yes, don't need to check the model name in the reset/save.

src/webots/nodes/WbTransform.cpp
```diff
@@ -52,6 +60,24 @@ WbTransform::~WbTransform() {
     wr_node_delete(WR_NODE(wrenNode()));
 }
 
+void WbTransform::reset(const QString &id) {
+  WbGroup::reset(id);
+  // note: for solids, the set of these parameters has to occur only if mJointParents.size() == 0 and it is handled in
+  // WbSolid::reset, otherwise it breaks the reset of hinge based joints
+  if (nodeType() != WB_NODE_TRACK_WHEEL && nodeType() != WB_NODE_SOLID) {
```
stefaniapedrazzi, line 67: What I meant was to use dynamic cast and not the `nodeType()` method, i.e.
```suggestion
  if (!dynamic_cast<const WbTrackWheel *>(this) && !dynamic_cast<const WbSolid *>(this)) {
```

Do you really want to skip only WbSolid nodes or any WbSolid-derived nodes (for example devices) that could also have parent or descendant joint nodes?

ad-daniel: Good point, doing so would alter the reset logic for those cases. As `SolidDevices` inherit the reset from `Solid` the change must apply to them as well. However can't use the cast for track wheel as it would introduce a circular dependency of `WbTransform`. Since nothing else depends on it (to my knowledge) it should be fine to leave that one as a nodeType check unless there's a better way.

ad-daniel:
I made a few tests with a `GPS` and appears to be behaving as intended.

stefaniapedrazzi (approved):
Thank you

PR #3171, opened 2021-06-14
Fix export of & in x3d

BenjaminDeleze:
The `&` was not sanitized correctly in solid names when exported in x3d and causes XML parsing errors.

It aims to fix https://github.com/RoboCup-Humanoid-TC/webots/issues/200.


omichel (approved):
Thank you!

PR #3173, opened 2021-06-14
Fix `test ros` CI for ubuntu 18

ad-daniel:
**Description**
Scheduled workflows have been failing for the past 2 [weeks](https://github.com/cyberbotics/webots/actions/workflows/test_suite_linux_develop.yml) due to the ROS-related key issue. Since the instructions have been updated [here](http://wiki.ros.org/Installation/Ubuntu/Sources) might as well do the same.


lukicdarkoo:
Thanks!

I have just realized that we have been installing the ROS dependency twice:
- https://github.com/cyberbotics/webots/blob/454d4536a366b4d0b73b556e03112f857577d6f0/.github/workflows/test_suite_linux.yml#L106
- https://github.com/cyberbotics/webots/blob/454d4536a366b4d0b73b556e03112f857577d6f0/.github/workflows/test_suite_linux.yml#L146

I think we need the ROS dependency only during the build and `test-ros` job.

What about deleting the `scripts/install/linux_test_dependencies.sh` from the `test-suite` job and replace the ros-tooling action with the `scripts/install/linux_test_dependencies.sh`?

.github/workflows/test_suite_linux.yml
```diff
@@ -102,8 +102,6 @@ jobs:
         path: artifact
     - name: Extract Webots
       run: tar xjf artifact/webots-*-x86-64*.tar.bz2 -C artifact
-    - name: Install Webots Dependencies
-      run: sudo scripts/install/linux_test_dependencies.sh
```
lukicdarkoo, line 106: This script installs runtime dependencies as well. We need an argument (e.g. `--exclude-ros`) to exclude only the ROS dependency.

lukicdarkoo (changes requested):

ad-daniel:
I don't have much time to look into this, do you want to @lukicdarkoo or should I revert back to the initial commit so that at least the scheduled github actions pass and open an issue about the possible optimization you proposed?

lukicdarkoo:
The ROS tests should be working, e.g. this is passing:
https://github.com/cyberbotics/webots/runs/2882570086?check_suite_focus=true

We can continue the implementation of the optimization (on this PR) once we are ready to tackle it

ad-daniel:
I mean these: https://github.com/cyberbotics/webots/actions/workflows/test_suite_linux_develop.yml

lukicdarkoo:
It should be ready for review now.

A few not-strictly related fixes:
- `python3-setuptools` is added because (https://github.com/cyberbotics/webots/pull/3173/checks?check_run_id=2885654329):
  ```
  ERROR: Command errored out with exit status 1:
       command: /usr/bin/python3 -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install- 
  1wex7u35/future_1108251f13264dc4bbcee02c04962307/setup.py'"'"'; __file__='"'"'/tmp/pip-install- 
  1wex7u35/future_1108251f13264dc4bbcee02c04962307/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if 
  os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', 
  '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-jltig8s6
           cwd: /tmp/pip-install-1wex7u35/future_1108251f13264dc4bbcee02c04962307/
      Complete output (3 lines):
        File "<string>", line 1, in <module>
      ModuleNotFoundError: No module named 'setuptools'
  ```
- `git` is added to the MSYS2 dependencies because (https://github.com/cyberbotics/webots/runs/2888587906):
  ```
  # removing files and folders not in git repository
  /bin/sh: line 1: git: command not found
  # generating project files
  # updating proto cache
  # checking submodules update
  ./check_submodules_update.sh: line 29: git: command not found
  ./check_submodules_update.sh: line 30: git: command not found
  ./check_submodules_update.sh: line 33: git: command not found
  ```

The Windows failure (https://github.com/cyberbotics/webots/runs/2888979665) shouldn't be related to this PR as it fails in the other PRs as well:
```
D:/a/_temp/msys/msys64/mingw64/bin/ffmpeg.exe: error while loading shared libraries: ?: cannot open shared object file: No such file or directory
  File "./copy_msys64_files.py", line 81, in <module>
    for ffmpeg_dll in subprocess.check_output(['bash', 'ffmpeg_dependencies.sh'], shell=True).decode('utf-8').split():
  File "C:\hostedtoolcache\windows\Python\3.8.10\x64\lib\subprocess.py", line 415, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "C:\hostedtoolcache\windows\Python\3.8.10\x64\lib\subprocess.py", line 516, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['bash', 'ffmpeg_dependencies.sh']' returned non-zero exit status 127.
make[1]: *** [Makefile:80: all] Error 1
make: *** [Makefile:62: distrib] Error 2
```

ad-daniel:
looks good to me, awesome. As for how much we gain in terms of time is hard to say, comparing the time it took to build ubuntu 18 here to other ones like [this](https://github.com/cyberbotics/webots/runs/2881466180?check_suite_focus=true) for instance it doesn't appear to make much of a difference. But probably depends on the load on github side as well, either way this approach makes more sense than the previous

lukicdarkoo:
> looks good to me, awesome. As for how much we gain in terms of time is hard to say, comparing the time it took to build ubuntu 18 here to other ones like [this](https://github.com/cyberbotics/webots/runs/2881466180?check_suite_focus=true) for instance it doesn't appear to make much of a difference. But probably depends on the load on github side as well, either way this approach makes more sense than the previous

Definitely, not a significant speed-up :/

scripts/install/msys64_installer.sh
```diff
@@ -20,6 +20,7 @@ declare -a BASE_PACKAGES=(
 )
 
 declare -a OPTIONAL_PACKAGES=(
+  "git"                       # Distribution script (check_submodules_update.sh)
```
omichel, line 23: Is that really needed? Git should be anyhow installed, otherwise the checkout of the repo is not possible.

lukicdarkoo: The git dependency is not related to the ROS fix, but I found some errors while reviewing the logs:

> * `git` is added to the MSYS2 dependencies because (https://github.com/cyberbotics/webots/runs/2888587906):
>   ```
>   # removing files and folders not in git repository
>   /bin/sh: line 1: git: command not found
>   # generating project files
>   # updating proto cache
>   # checking submodules update
>   ./check_submodules_update.sh: line 29: git: command not found
>   ./check_submodules_update.sh: line 30: git: command not found
>   ./check_submodules_update.sh: line 33: git: command not found
>   ```

— https://github.com/cyberbotics/webots/pull/3173#issuecomment-866570817

It is not necessary to include, the CI doesn't fail, it just displays the warnings

omichel: That is probably because the original git is not the one of MSYS2 and cannot be found from the MSYS2 console due to a missing PATH. Adding it here doesn't hurt too much as it's not a big download.

omichel (approved):

lukicdarkoo (approved):

PR #3175, opened 2021-06-15
Add force/torque scale factor to Worldinfo

mael25:
**Description**
This PR adds two new fields to `WorldInfo`: `dragForceScale` and `dragTorqueScale`. They allow the user to set the order of magnitude of the force/torque applied to a `Solid` inside the Webots interface when doing (Alt+Click). Indeed, it is sometimes difficult to apply a force/torque in the desired range when dragging the mouse. The force is computed as follow: 

*F* [N] = `dragForceScale` * `Solid.mass` * *d*<sup> 3</sup>, where *d* corresponds to the vector created by dragging the mouse (in meters),

instead of F [N] = _constant_ * *d*<sup> 3</sup> (nothing changes except we can set the value of the _constant_).

The torque is computed in the same idea:

*T* [Nm] = `dragTorqueScale` * `Solid.mass` * *d*<sup> 3</sup>,

instead of T [Nm] = _constant_  * `Solid.mass` * *d*   (we have the cube of *d* here to be consistent with the computation of the force. Also, it enables to increase slowly when *d*<1 which is good as the object goes easily out of bounds when the applied torque is too high).

The default values of the added fields corresponds to the _constant_ values that are currently used (`dragForceScale` = 30 and `dragTorqueScale` = 5)


omichel (changes requested):

ShuffleWire:
For the torque part, should we not better use the rotational inertia (from the inertia matrix) instead of mass ?

src/webots/gui/WbDragSolidEvent.cpp
```diff
@@ -235,26 +234,19 @@ void WbDragForceEvent::updateOrigin() {
...
 QString WbDragForceEvent::magnitudeString() {
-  const double lengthSquared = mVector.length2();
-  mScalingFactor = FORCE_SCALING_FACTOR * lengthSquared;
-  mMagnitude = mScalingFactor * sqrt(lengthSquared);
-  QString returnValue = WbPrecision::doubleToString(mMagnitude, WbPrecision::GUI_MEDIUM) + " N";
-  mMagnitude = mScalingFactor * mVector.length();
-  return returnValue;
+  mMagnitude = mScalingFactor * mVector.length2() * mVector.length();
+  return WbPrecision::doubleToString(mMagnitude, WbPrecision::GUI_MEDIUM) + " N";
 }
```
omichel, line 245: Code clean-up: I would delete this apparently useless `mMagnituteString` field and make this method const.
The same applies to the `WbDragTorqueEvent` object.

omichel (changes requested):
The behavior looks good to me.
However, I could clean-up a little bit the code.

mael25:
> For the torque part, should we not better use the rotational inertia (from the inertia matrix) instead of mass ?

@ShuffleWire This would indeed give a more relevant range of torque and would enable to adapt it according to the direction of the torque to apply, but in most use cases the mass gives us already a good idea of the order of magnitude of the torque to apply.

However, if this is a feature that you would really like to see into Webots, feel free to implement it ! ;)

ShuffleWire:
> > For the torque part, should we not better use the rotational inertia (from the inertia matrix) instead of mass ?
> 
> @ShuffleWire This would indeed give a more relevant range of torque and would enable to adapt it according to the direction of the torque to apply, but in most use cases the mass gives us already a good idea of the order of magnitude of the torque to apply.
> 
> However, if this is a feature that you would really like to see into Webots, feel free to implement it ! ;)

Yes, I do agree, we are not doing accurate physic here, and it would be relevant only for part with mass very far from center of mass, which is not common. Mass will be nice, then :)

Issue #3158, opened 2021-06-10
Python controller segfaults in RangeFinder::__getRangeImageList


sigmunau:
**Describe the Bug**
When running a python controller that uses threads (even if the other threads does not touch webots api) the controller process segfaults in PyFloat_FromDouble called from __getRangeImageList
Tried on both https://github.com/cyberbotics/webots/releases/download/nightly_9_6_2021/webots_2021a-rev1_amd64.deb and https://github.com/cyberbotics/webots/releases/download/nightly_9_6_2021/webots_2021b_amd64.deb with seemingly the same results. The stacktrace below is from 2021b:

Most likely this is caused by a previous change to release python GIL while in c++ code. This also (seemingly) releases the log inside the %extends blocks which may call non-threadsafe python functions. The solution to this seems to be to wrap such code like this
```
%nothread;
<extension code>
%thread;
```
gdb stack trace:
```
#0  0x00000000005bc013 in PyObject_Malloc ()
[Current thread is 1 (Thread 0x7fdc41323740 (LWP 79169))]
(gdb) bt
#0  0x00000000005bc013 in PyObject_Malloc ()
#1  0x00000000005ed03e in PyFloat_FromDouble ()
#2  0x00007fdc3fbd84c8 in webots_RangeFinder___getRangeImageList(webots::RangeFinder*) () from /usr/local/webots/lib/controller/python38/_controller.so
#3  0x00007fdc3fc100a8 in _wrap_RangeFinder___getRangeImageList ()
   from /usr/local/webots/lib/controller/python38/_controller.so

```

**Steps to Reproduce**
Load world file
Start controller
Controller crashes when it calls RangeFinder.getRangeImage('list')

**Expected behavior**
Controller does not crash

**Screenshots**

**System**
Ubuntu 20.04 webots nightly 20
 - Operating System: [e.g., Windows 10, Linux Ubuntu 18.04, macOS Mojave]
 - Graphics Card: [e.g., NVIDIA GeForce RTX 2080 11 GB, AMD Radeon RX 580 8GB, etc.]

**Additional context**

omichel:
Does adding:
```
%nothread __getRangeImageList;
```
right after:
```
%module(threads=1) controller
```
fixes the problem for you?

sigmunau:
[...]

PR #3176, opened 2021-06-15
Re-acquire the python GIL before calling libpython functions

sigmunau:
This fixes a segfault when calling RangeSensor.getRangeImage('list')
and possilby others

**Related Issues**
This pull-request fixes issue #3158

I think we only really need to release the GIL when calling Supervisor.step() and that would make for a much cleaner change, but I couldn't find which include contains it.

### DIFF
--- a/src/controller/python/controller.i
+++ b/src/controller/python/controller.i
@@ -82,6 +82,7 @@ using namespace std;
 //  Miscellaneous - controller module's level
 //----------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


omichel:
> I think we only really need to release the GIL when calling Supervisor.step() and that would make for a much cleaner change, but I couldn't find which include contains it.

It's in `Robot.hpp`.

omichel:
Replaced by #3389.

Issue #3179, opened 2021-06-16
The Windows nightly build is broken


omichel:
I guess the problem was introduced in #3087:
```bash
Error on line 2301 in D:\a\webots\webots\scripts\packaging\webots.iss: Source file "D:\a\webots\webots\resources\javascript\modules" does not exist.
```

See full log [here](https://github.com/cyberbotics/webots/runs/2834339850?check_suite_focus=true).

ad-daniel:
Appears to be working now, so I'm closing this issue.

PR #3181, opened 2021-06-16
Fix windows nightly build

ad-daniel:
**Description**
Addresses issue #3179.
It appears that for windows it's necessary to be verbose and include all folders.


omichel (approved):
It seems to work.
Thanks.

PR #3186, opened 2021-06-16
Removed obsolete Python 2.7 tests

omichel:
These tests were causing PEP8 failures with python 3 code, such as some f-strings.
Since Python 2.7 is deprecated since January 2021, we should get rid of it.


BenjaminDeleze (approved):
Thank you

Issue #2994, opened 2021-04-21
Wrapping endPoint in a LUA statement results in freeze/crash


ad-daniel:
**Describe the Bug**
Wrapping the endpoint of a hinge joint with LUA statements leads to a crash

**Steps to Reproduce**
1. load [bug_reset_with_lua.zip](https://github.com/cyberbotics/webots/files/6351230/bug_reset_with_lua.zip)

2. run a few seconds
3. reset

The interface freezes and eventually crashes. Code-side, I don't know where it triggers but it results in the function [WbField::fieldChanged](https://github.com/cyberbotics/webots/blob/1a090134c5451bacd35b9e3f26e773c3d411c329/src/webots/vrml/WbField.cpp#L396) being called infinite times.

4. remove/comment the lua code from `ProtoHingeJoint.proto` and it no longer does it.

```
PROTO ProtoHingeJoint [
  field SFNode                                 jointParameters   NULL
  field MFNode                                 device            [ ]
  field SFNode                                 endPoint          NULL
]
{
  HingeJoint {
    jointParameters IS jointParameters
    device IS device
    %{ if fields.endPoint.value == nil then }%
      %{ print("a") }%
    %{ else }%
      %{ print("b") }%
      endPoint IS endPoint
    %{ end }%
  }
}
```

The bug appears to be present both in develop and master, so probably not a regression (not a recent one at least).

ad-daniel:
It appears that fix #3227 also ended up fixing this specific case, the other case similar to this mentioned here #3215 on the other hand still appears to regenerate infinitely. This issue can therefore be closed.

PR #3188, opened 2021-06-16
Fix endless regeneration resulting in a crash

ad-daniel:
**Description**
Cases of endless regeneration:
- ~~Addresses #2994 - work in progress~~ 
   - no longer needed, fixed by  #3227.
- The first example #3215 however still appears to be infinitely regenerating.

The smallest reproducible example I could manage for the second issue is this: [bug_mf_deletion.zip](https://github.com/cyberbotics/webots/files/6717186/bug_mf_deletion.zip)

**Steps to reproduce:**
Open world, delete the second `RoadLine` in the `lines` field. Webots should crash.
Or do the same in a world that has a `AddLaneRoadSegment.proto`

Basically it involves: 
- nested PROTO
- template regenerator fields on both levels of the PROTO
- IS statements involving this fields (which needs to be in a `Solid { Group { children [] } }` structure for some reason).

I couldn't simplify it further without loosing the crash, which makes the debugging rather tricky given how specific of a structure it is. However, basically what appears to going on is that:

When removing the the `RoadLine` from the `MFField`, the field `lines` being a template regenerator, triggers a regeneration which propagates to the most internal PROTO first, during this process, after creating the new node, the parameter fields of this newly created node are redirected to the parent parameters [here](https://github.com/cyberbotics/webots/blob/ecbbc1642d80f3e702bfbbf619d544f85e4ceb06/src/webots/nodes/utils/WbTemplateManager.cpp#L390-L398). In the process of doing this redirection however, the copy of this parameter [here](https://github.com/cyberbotics/webots/blob/da4f3d4e26be0b5da401547daef7afc4374743f6/src/webots/vrml/WbField.cpp#L325-L326) is not protected against generating new signals which triggers a new regeneration. Because of this, the inner PROTO regenerates infinitely until at some point it crashes webots.

The current implementation fixes the issue and the blocking of the signal occurs only during the regeneration of a node as doing so every time would break several cases in the test suite as this function also plays a role during the building of the node structure.

Despite being reasonable and despite fixing both the MRE case and issue #3215, I definitely would like a second opinion on the approach or possible side-effects.


src/webots/nodes/utils/WbTemplateManager.cpp
```diff
@@ -393,7 +393,7 @@ void WbTemplateManager::regenerateNode(WbNode *node, bool restarted) {
     foreach (WbField *parentParameter, previousParentRedirections) {
       foreach (WbField *newParam, newNode->parameters()) {
         if (parentParameter->name() == newParam->alias())
-          newParam->redirectTo(parentParameter);
+          newParam->redirectTo(parentParameter, true);
```
stefaniapedrazzi, line 396: Given that this a special case, I would probably block the signals just here without modifying the `WbField::redirectTo` function and adding a default argument.

stefaniapedrazzi (commented):
Thank you. This fix seems fine.

I checked the original use case where the redirection was needed and it still works correctly.
I'm just doing some more tests.

stefaniapedrazzi:
I found some more crash related to PROTOs but nothing caused by this PR.
I will wait for the tests result for the approval.

stefaniapedrazzi (approved):
Thank you

PR #3189, opened 2021-06-16
Fix reconnecting extern controllers

stefaniapedrazzi:
Replaces #3005.
Fixes: #3000 and #1903

> If the controller died, it can't call the cleanup function, and when I restart it, I get a error message from webots because he think that the controller is still running, so he can't find a spot for the new incoming connection.

> It simply remove the dead controller connection as usual, then recreate a new one.
The trick is to use the name of the robot to know which dead controller should be kill.
Otherwise, if no name is provided, the code execute as usual (scanning for free spot).

In case a single extern controller is defined, then the robot name is not needed and the currently connected controller is terminated and reconnected.

### DIFF
--- a/docs/reference/changelog-r2021.md
+++ b/docs/reference/changelog-r2021.md
@@ -58,6 +58,7 @@ Released on June, Xth, 2021.
     - Added a version of the [RobotisOp2](../guide/robotis-op2.md) modeled using [Hinge2Joint](hinge2joint.md) on the ankles, hips, and neck ([#2861](https://github.com/cyberbotics/webots/pull/2861)).
     - Made the `static` behavior the default for PROTO files and removal of the tag. Non static cases must be labeled as such using the `nonDeterministic` tag instead ([#2903](https://github.com/cyberbotics/webots/pull/2903)).
   - Bug fixes:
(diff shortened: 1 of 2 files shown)


ShuffleWire:
Thanks for taking it ! Not been able to work on it for a while...

lukicdarkoo:
It looks very good, it makes the external controllers much more convenient to use. However, I am not sure about the following edge case: 
- If there is a single robot (with an external controller) in the world and I run two external controllers then the second one kills the first one.
Should we handle such use-case or let the users worry not to kill the other controller?

stefaniapedrazzi:
Yes. Currently no way to avoid this because Webots doesn't know if the external controller crashed or is still running properly.
But note that this is not a particular edge case, as the same behavior occurs in all the cases including simulations with multiple extern controllers if the `WEBOTS_ROBOT_NAME` is set.

I think that the question is if this "edge-case" is a real use case.
Why exactly a user should start a second extern controller?
If multiple simulations are running concurrently, then is it not needed to set the PID to correctly connect to the Webots instance?

We should probably evaluate which use-case is more affecting users.

lukicdarkoo (approved):
> Why exactly a user should start a second extern controller?

I think the question is more like, Should an accidental run of the second controller crash the one running properly.

However, I agree with you, there are way more benefits with this approach than cons.

PR #3190, opened 2021-06-16
Don't fire tests on ready_for_review event

omichel:
I don't understand the need for launching the sources and doc tests on the `ready_for_review` event, given that there is no code modification at this point and tests are run on each commit (`synchronize` event). We would save time by disabling this.


lukicdarkoo (approved):
Thank you!

PR #3192, opened 2021-06-17
Fix robot windows of robotbenchmark simulations

### LINKED ISSUE #3182: Robot window layout in robotbenchmark simulations is broken
Info window layout in robotbenchmark simulation is broken and the tabs are no longer correctly displayed.
I would expect that after compiling webots from the main Makefile all the simulation works correctly including these ones.
robotbechmark_info_window

BenjaminDeleze:
**Description**
Restore the robot windows of the robotbenchmark simulations that were broken with the removal of JQuery:

- [x] Change JQuery expression with vanilla JS
- [x] Change `var` to `let`/`const`

Fixes #3182


stefaniapedrazzi (changes requested):
Thank you. Globally seems fine.
I started testing the functionality, I will then check the code changes.

I just found some issues with images and the width of the robot window content.
In the `humanoid_sprint.wbt` and `wall_following.wbt` robot windows, the content width is not correctly updated when increasing the width of the robot window widget and images are partially outside the content area.

Other issues were already reproducible in the master branch.

projects/samples/robotbenchmark/include/tabs.css
```diff
@@ -0,0 +1,64 @@
...
+}
+.radio{
+  display:none;
+}
+#one:checked ~ .panels #one-panel,
+#two:checked ~ .panels #two-panel,
+#three:checked ~ .panels #three-panel,
+#four:checked ~ .panels #four-panel{
+  display:block
+}
```
stefaniapedrazzi, line 56: Is not possible to use a class instead of ids?
That would be more generic and scalable.

projects/samples/robotbenchmark/include/tabs.css
```diff
@@ -0,0 +1,64 @@
...
+#one:checked ~ .tabs #one-tab,
+#two:checked ~ .tabs #two-tab,
+#three:checked ~ .tabs #three-tab,
+#four:checked ~ .tabs #four-tab{
```
stefaniapedrazzi, line 60: Is not possible to use classes selectors instead of ids?
That would be more generic and scalable.

projects/samples/robotbenchmark/highway_driving/plugins/robot_windows/highway_driving/highway_driving.html
```diff
@@ -2,57 +2,62 @@
...
-          This benchmark aims at driving the farther as possible in one minute without colliding any other vehicles or static objects.
-        </p>
+    <div id='infotabs' class='wrapper webotsTabs webotsRobotWindowTabs'>
+      <input class="radio" id="one" name="group" type="radio" checked>
+      <input class="radio" id="two" name="group" type="radio">
+      <input class="radio" id="three" name="group" type="radio">
+      <div class="tabs">
+        <label class="tab" id="one-tab" for="one">Overview</label>
+        <label class="tab" id="two-tab" for="two">Metrics</label>
+        <label class="tab" id="three-tab" for="three">Instructions</label>
```
stefaniapedrazzi, line 15: I would rather keep specific ids describing the content (overview, metric, instructions, etc.) instead of using generic "one", "two", "three", etc. that doesn't tell anything about the content.

Same for panel ids, etc.

stefaniapedrazzi (changes requested):

projects/samples/robotbenchmark/viewpoint_control/plugins/robot_windows/viewpoint_control/viewpoint_control.html
```diff
@@ -2,45 +2,52 @@
...
-        <div style='text-align:center'>
-          <button id='record-button' disabled onclick='recordPerformance()'>Record</button>
+
+      <div class="panels">
+        <div class="panel" id="one-panel">
+          <p>This benchmark will teach you how to move the 3D viewpoint using the mouse.</p>
+          <p>The tab entitled <a class='tablink' onclick='$("#infotabs").tabs({active:1});'>Metrics</a> measures your progress in an exercise where
+            you have to place the viewpoint at a specific position and orientation. It also allows you to record your performance in
+            the robotbenchmark database (if you are logged-in).</p>
+          <p>The tab entitled <a class='tablink' onclick='$("#infotabs").tabs({active:2});'>Instructions</a> explains you how to use the mouse to
```
stefaniapedrazzi, line 24: These links to other tabs in the same document are no longer working.
They should be fixed or removed.

Same on  line 21 an 39.

stefaniapedrazzi (commented):
In `viewpoint_control.js` there is still a reference to the THREE library.
The code to evaluate the viewpoint position is currently only working when running it from the web interface and it doesn't seem we are interested in further supporting this benchmark.
But it would be good to clean it up and remove all the references to three.js.

BenjaminDeleze:
I think I addressed all the comments
  -  To use classes instead of ids was a bit tricky and I could not find another solution than introducing a bit of JS.
  -  I removed the viewpoint control benchmark.

stefaniapedrazzi:
I'm reviewing the changes now.
Could you please add an entry in the ChangeLog about the removal of the viewpoint_control benchmark?

stefaniapedrazzi (changes requested):
Thank you!

Images size is now updated correctly on window resize.
But even at maximum size, the legend of the plot in `humanoid_sprint.wbt` is unreadable.
Maybe it is possible to increase the maximum size of the robot window content.

BenjaminDeleze:
I totally removed the max-width, I do not think it is useful.
However, it was not sufficient as the legend of the plot is not readable either in robotbenchmark.net, so I also removed the predefined size of the image and it seems to display well now.

projects/samples/robotbenchmark/highway_driving/plugins/robot_windows/highway_driving/highway_driving.html
```diff
@@ -2,59 +2,79 @@
...
+          return;
+        height = newHeight;
+        let panels = document.getElementsByClassName('panel');
+        for (panel of panels) {
+          panel.style.top = height + 'px';
+        };
+      }
+      window.addEventListener('resize', adaptPositionOfText);
+      adaptPositionOfText();
+    </script>
```
stefaniapedrazzi, line 78: Just a question: why this was not needed before when using ids instead of classes in the CSS?
Shouldn't work exactly the same?

BenjaminDeleze: Because with ids we had a structure like:
```
<input></input>
<input></input>
<input></input>

<label class=tabs></label>
<label class=tabs></label>
<label class=tabs></label>

<div class=panel></panel>
<div class=panel></panel>
<div class=panel></panel>
```
Because we were using IDs, it was easy to connect one input to one particular div.
And also the element were displayed in order: the tabs on top, the panel below.

However when switching for classes the structures changed like this:
```
<input></input>
<label class=tabs></label>
<div class=panel></panel>

<input></input>
<label class=tabs></label>
<div class=panel></panel>

<input></input>
<label class=tabs></label>
<div class=panel></panel>
```

It is the case because as we don't have IDs anymore, it is way harder to connect one specific input to a specific panel. The simplest solution with only CSS I found was to connect the input to the nearest panel below it. But as you can see the structure is not the same without further changes it would display like:

```
<Overview>
text_of_panel, text_of_panel, text_of_panel, text_of_panel, text_of_panel, text_of_panel
text_of_panel, text_of_panel, text_of_panel, text_of_panel, text_of_panel, text_of_panel, 
text_of_panel, text_of_panel, text_of_panel, text_of_panel, text_of_panel, text_of_panel
<Metric>
<Instructions>
```

To solved that I made the position of the panels absolute. Problem: if we resize the window and the buttons are placed on several columns, we need to update the position of the panels to always be just below the lowest button. And that what the JS code does. 

I am not sure this is an optimal solution, but it is the simplest I could come up with without IDs and then scalable.

stefaniapedrazzi (dismissed):

BenjaminDeleze:
Can you approved it again please?
I had to fix a conflict in the changelog.

docs/reference/changelog-r2021.md
```diff
@@ -129,6 +129,8 @@ Released on June, Xth, 2021.
   - Cleanup
     - Deleted deprecated DifferentialWheels node ([#2749](https://github.com/cyberbotics/webots/pull/2749)).
     - Changed structure of the [projects/samples/howto]({{ url.github_tree }}/projects/samples/howto) directory, so each demonstration is in a dedicated directory ([#2639](https://github.com/cyberbotics/webots/pull/2639)).
+    - Remove the viewpoint_control benchmark that was not working in Webots ([#3192](https://github.com/cyberbotics/webots/pull/3192)).
+
```
stefaniapedrazzi, line 133:
```suggestion
```

stefaniapedrazzi (changes requested):
Yes, could you just please remove this empty line?

stefaniapedrazzi (approved):
Thank you

PR #3196, opened 2021-06-17
Various fixes

ShuffleWire:
**Description**
Running around in the code, I've found various things that might need to be fixed. I does not change (a lot) of code and should not impact anything.
I've gathered them in this PR, but I'm aware that it's not a proper PR, more a place to put my fixes together.
Each commit explain, if need be, the rational behind each modification.
Feel free to discard it if it's not relevant, or pick commit independently.

### DIFF
--- a/docs/reference/linearmotor.md
+++ b/docs/reference/linearmotor.md
@@ -23,6 +23,8 @@ It defaults to `"linear motor"`.
 - The `maxForce` field specifies both the upper limit and the default value for the motor *available force*.
 The *available force* is the force that is available to the motor to perform the requested motions.
 It is expressed in *newton* [N].
+When used in velocity mode, this maximum force is always applied (with the correct sign) until the speed is reached. 
+See : http://ode.org/wiki/index.php?title=Manual#Stops_and_motor_parameters
 The `wb_motor_set_available_force` function can be used to change the *available force* at run-time.
 The value of `maxForce` should always be zero or positive (the default value is 10 N).
-A small `maxForce` value may result in a motor being unable to move to the target position because of its weight or other external forces.
+A small `maxForce` value may result in a motor being unable to move to the target position because of its weight or other external forces. A motor can't go backward when used in speed mode.
--- a/docs/reference/linearmotor.md
+++ b/docs/reference/linearmotor.md
@@ -27,4 +27,5 @@ When used in velocity mode, this maximum force is always applied (with the corre
 See : http://ode.org/wiki/index.php?title=Manual#Stops_and_motor_parameters
 The `wb_motor_set_available_force` function can be used to change the *available force* at run-time.
 The value of `maxForce` should always be zero or positive (the default value is 10 N).
-A small `maxForce` value may result in a motor being unable to move to the target position because of its weight or other external forces. A motor can't go backward when used in speed mode.
+A small `maxForce` value may result in a motor being unable to move to the target position because of its weight or other external forces. 
+A motor can't go in the direction opposed to the requested direction when used in speed mode.
--- a/docs/reference/linearmotor.md
+++ b/docs/reference/linearmotor.md
@@ -20,11 +20,9 @@ A [LinearMotor](#linearmotor) node can be used to power a [SliderJoint](sliderjo
 This the name to which the `wb_robot_get_device` function refers.
 It defaults to `"linear motor"`.
 
-- The `maxForce` field specifies both the upper limit and the default value for the motor *available force*.
+- The `maxForce` field specifies both the upper limit and the default value for the motor *available force* and is expressed in *newton* [N].
 The *available force* is the force that is available to the motor to perform the requested motions.
-It is expressed in *newton* [N].
-When used in velocity mode, this maximum force is always applied (with the correct sign) until the speed is reached. 
-See : http://ode.org/wiki/index.php?title=Manual#Stops_and_motor_parameters
+In velocity control, this maximum force is always applied (with the correct sign) until the target speed is reached, more details are available [here](http://ode.org/wiki/index.php?title=Manual#Stops_and_motor_parameters).
 The `wb_motor_set_available_force` function can be used to change the *available force* at run-time.
 The value of `maxForce` should always be zero or positive (the default value is 10 N).
 A small `maxForce` value may result in a motor being unable to move to the target position because of its weight or other external forces. 
--- a/docs/reference/linearmotor.md
+++ b/docs/reference/linearmotor.md
@@ -25,4 +25,4 @@ The *available force* is the force that is available to the motor to perform the
 In velocity control, this maximum force is always applied (with the correct sign) until the target speed is reached, more details are available [here](http://ode.org/wiki/index.php?title=Manual#Stops_and_motor_parameters).
 The `wb_motor_set_available_force` function can be used to change the *available force* at run-time.
 The value of `maxForce` should always be zero or positive (the default value is 10 N).
-A small `maxForce` value may result in a motor being unable to move to the target position because of its weight or other external forces. 
+A small `maxForce` value may result in a motor being unable to move to the target position because of its weight or other external forces.
--- a/docs/reference/linearmotor.md
+++ b/docs/reference/linearmotor.md
@@ -22,7 +22,7 @@ It defaults to `"linear motor"`.
 
 - The `maxForce` field specifies both the upper limit and the default value for the motor *available force* and is expressed in *newton* [N].
 The *available force* is the force that is available to the motor to perform the requested motions.
-In velocity control, this maximum force is always applied (with the correct sign) until the target speed is reached, more details are available [here](http://ode.org/wiki/index.php?title=Manual#Stops_and_motor_parameters).
+When used in velocity control, this maximum force is always applied (with the correct sign) until the target speed is reached, more details are available [here](http://ode.org/wiki/index.php?title=Manual#Stops_and_motor_parameters).
 The `wb_motor_set_available_force` function can be used to change the *available force* at run-time.
 The value of `maxForce` should always be zero or positive (the default value is 10 N).
 A small `maxForce` value may result in a motor being unable to move to the target position because of its weight or other external forces.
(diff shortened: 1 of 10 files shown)


docs/reference/linearmotor.md
```diff
@@ -23,6 +23,8 @@ It defaults to `"linear motor"`.
 - The `maxForce` field specifies both the upper limit and the default value for the motor *available force*.
 The *available force* is the force that is available to the motor to perform the requested motions.
 It is expressed in *newton* [N].
+When used in velocity mode, this maximum force is always applied (with the correct sign) until the speed is reached. 
+See : http://ode.org/wiki/index.php?title=Manual#Stops_and_motor_parameters
 The `wb_motor_set_available_force` function can be used to change the *available force* at run-time.
 The value of `maxForce` should always be zero or positive (the default value is 10 N).
-A small `maxForce` value may result in a motor being unable to move to the target position because of its weight or other external forces.
+A small `maxForce` value may result in a motor being unable to move to the target position because of its weight or other external forces. A motor can't go backward when used in speed mode.
```
ad-daniel, line 30: I'm not sure I understand this change. In velocity control positive or negative target velocities can be provided, what do you mean it can't go backwards?
Also, for documentation we usually write one sentence per line to avoid formatting manually. Same thing for the one below

ad-daniel (commented):
Thank you for the contribution!

docs/reference/linearmotor.md
ShuffleWire: I mean that If when we request a movement in one direction, the resulting movement could not be in the opposite direction, whatever the force. For instance if we use request a vertical linear motor to extend (to push a mass to the top), even if the mass is crazy heavy, the motor could not retract. (which is actually not true for servomotor, when the external torque is too high, the servomotor go backward (or break...))
It's afaik, however, feel free to correct that :)

ShuffleWire: Made a change, see if it better for you

docs/reference/rotationalmotor.md
```diff
@@ -23,6 +23,9 @@ It defaults to `"rotational motor"`.
 - The `maxTorque` field specifies both the upper limit and the default value for the motor *available torque*.
 The *available torque* is the torque that is available to the motor to perform the requested motions.
 It is expressed in *newton meter* [N⋅m].
+When used in velocity mode, this maximum torque is always applied (with the correct sign) until the speed is reached. 
+See : http://ode.org/wiki/index.php?title=Manual#Stops_and_motor_parameters
```
ad-daniel, line 27:
```suggestion
- The `maxTorque` field specifies both the upper limit and the default value for the motor *available torque* and is expressed in *newton meter* [N⋅m].
The *available torque* is the torque that is available to the motor to perform the requested motions.
When used in velocity control, this maximum torque is always applied (with the correct sign) until the target speed is reached. More details are available [here](http://ode.org/wiki/index.php?title=Manual#Stops_and_motor_parameters).
```

docs/reference/linearmotor.md
```diff
@@ -23,6 +23,9 @@ It defaults to `"linear motor"`.
 - The `maxForce` field specifies both the upper limit and the default value for the motor *available force*.
 The *available force* is the force that is available to the motor to perform the requested motions.
 It is expressed in *newton* [N].
+When used in velocity mode, this maximum force is always applied (with the correct sign) until the speed is reached. 
+See : http://ode.org/wiki/index.php?title=Manual#Stops_and_motor_parameters
```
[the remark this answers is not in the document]
ad-daniel: I would re-organize the paragraph a bit, what do you think?

docs/reference/linearmotor.md
ad-daniel: I understand what you mean now, but I'm not entirely convinced. See this example for instance:

https://user-images.githubusercontent.com/44834743/122745358-34049480-d289-11eb-946b-f1ccce14b137.mp4

docs/reference/rotationalmotor.md
ShuffleWire: It's totally fine, completely forgot to use markdown formatting !

docs/reference/linearmotor.md
ShuffleWire: never meet this before ! 
so maybe remove this comment, better says anything than something wrong

ad-daniel: Yes maybe it's better to leave it out for now, but if you come across it again feel free to open an issue about it :)

docs/reference/linearmotor.md
```diff
@@ -20,9 +20,9 @@ A [LinearMotor](#linearmotor) node can be used to power a [SliderJoint](sliderjo
...
 
-- The `maxForce` field specifies both the upper limit and the default value for the motor *available force*.
+- The `maxForce` field specifies both the upper limit and the default value for the motor *available force* and is expressed in *newton* [N].
 The *available force* is the force that is available to the motor to perform the requested motions.
-It is expressed in *newton* [N].
+In velocity control, this maximum force is always applied (with the correct sign) until the target speed is reached, more details are available [here](http://ode.org/wiki/index.php?title=Manual#Stops_and_motor_parameters).
 The `wb_motor_set_available_force` function can be used to change the *available force* at run-time.
 The value of `maxForce` should always be zero or positive (the default value is 10 N).
-A small `maxForce` value may result in a motor being unable to move to the target position because of its weight or other external forces.
+A small `maxForce` value may result in a motor being unable to move to the target position because of its weight or other external forces.
```
ad-daniel, line 28:
```suggestion
A small `maxForce` value may result in a motor being unable to move to the target position because of its weight or other external forces.
```

ad-daniel: Last thing and then it looks good to me.

ShuffleWire:
regarding 4e3ea2a050d02e24beea5ae2e987823804b25213
not to do in this PR, but maybe move the Qstringlist in global (const ?) variable, to avoid rebuild it each time (or maybe static in the function). Maybe their is more places to make it, but I will do it later if I catch some.
Ultimately, I'm not a big fan of detecting of isDeviceTypeName and isSolidDeviceTypeName. i think it add another place to maintain, and prone to be not updated correctly. Could be we dynamic_cast, or a field in the object ?
Anyway, just a thought, not going to do I here

docs/reference/linearmotor.md
ShuffleWire: here we go.

stefaniapedrazzi:
> Ultimately, I'm not a big fan of detecting of isDeviceTypeName and isSolidDeviceTypeName. i think it add another place to maintain, and prone to be not updated correctly. Could be we dynamic_cast, or a field in the object ?

No. There are some use cases where no object is available but just the type name.
Do it is needed to have functions that mimic the dynamic cast functionality based on the type name.
Often there is a good reason behind the Webots code choices.
Please try to not propose random code improvements without a good functional or performance reason because they won't probably bring substantial benefits to the Webots simulator.

ShuffleWire:
> No. There are some use cases where no object is available but just the type name.
> So it is needed to have functions that mimic the dynamic cast functionality based on the type name.
> Often there is a good reason behind the Webots code choices.
> Please try to not propose random code improvements without a good functional or performance reason because they won't probably bring substantial benefits to the Webots simulator.

I'm fine with it, this is exactly why I've not tried to make more change in that direction. As stated, feel free to kick commit if they don't match Webots development direction. To be fair, I knew that this point was the only one that could be discussed. Don't want to make waste your time with it, however.
Will think twice in the future to push more beneficial ideas :)

docs/reference/rotationalmotor.md
```diff
@@ -20,9 +20,9 @@ A [RotationalMotor](#rotationalmotor) node can be used to power either a [HingeJ
 This the name to which the `wb_robot_get_device` function refer.
 It defaults to `"rotational motor"`.
 
-- The `maxTorque` field specifies both the upper limit and the default value for the motor *available torque*.
+- The `maxTorque` field specifies both the upper limit and the default value for the motor *available torque* and is expressed in *newton meter* [N⋅m]
 The *available torque* is the torque that is available to the motor to perform the requested motions.
-It is expressed in *newton meter* [N⋅m].
+When used in velocity mode, this maximum torque is always applied (with the correct sign) until the speed is reached, more details are available [here](http://ode.org/wiki/index.php?title=Manual#Stops_and_motor_parameters).
```
ad-daniel, line 25:
```suggestion
When used in velocity control, this maximum torque is always applied (with the correct sign) until the target speed is reached, more details are available [here](http://ode.org/wiki/index.php?title=Manual#Stops_and_motor_parameters).
```

docs/reference/rotationalmotor.md
```diff
@@ -20,9 +20,9 @@ A [RotationalMotor](#rotationalmotor) node can be used to power either a [HingeJ
 This the name to which the `wb_robot_get_device` function refer.
 It defaults to `"rotational motor"`.
 
-- The `maxTorque` field specifies both the upper limit and the default value for the motor *available torque*.
+- The `maxTorque` field specifies both the upper limit and the default value for the motor *available torque* and is expressed in *newton meter* [N⋅m]
```
ad-daniel, line 23:
```suggestion
- The `maxTorque` field specifies both the upper limit and the default value for the motor *available torque* and is expressed in *newton meter* [N⋅m].
```

ad-daniel (changes requested):

docs/reference/rotationalmotor.md
ShuffleWire: thanks !

ad-daniel:
Looks good for me, thank you! Just waiting for the test doc to pass.
(by the way, you can apply the suggested changes directly from github, it's easier to keep track of what has been done or not, particularly the small stuff like this).

ShuffleWire:
> Looks good for me, thank you! Just waiting for the test doc to pass.
> (by the way, you can apply the suggested changes directly from github, it's easier to keep track of what has been done or not, particularly the small stuff like this).

Thank a lot for your feedback :)
(yes, I did it sometimes, but the last one was already too old and gitlab was complaining :) )

ad-daniel (approved):
Thank you!

ShuffleWire:
Merge is blocked by a requested change that that is already fixed. could I do anything about it ?

PR #3197, opened 2021-06-17
Avoid Windows deleting webots-libcontroller header files

lukicdarkoo:


scripts/packaging/sync_controller_lib.sh
```diff
@@ -31,16 +31,16 @@ else
...
-# Prepare the structure
-rm -rf lib/${OSTYPE}
-mkdir -p lib/${OSTYPE}
-rm -rf include
-mkdir -p include
-
-# Copy files
+# Copy headers
 if [ "${OSTYPE}" != "msys" ]; then
+    rm -rf include
```
omichel, line 36:
```suggestion
    # don't copy the include files on Windows as they are the same as on other platforms, and they generate a huge diff due to Windows line endings which differ from Linux/macOS.
    rm -rf include
```

omichel (approved):
I would simply add a comment in the script so that we better understand the reason for this case.

omichel (approved):
Thank you.

Issue #3157, opened 2021-06-10
Display "name" field in tree


ShuffleWire:
**Is your feature request related to a problem? Please describe.**
When working on the tree with multiples solid, I always end up with this situation : 
![image](https://user-images.githubusercontent.com/83014015/121520814-a8bb1180-c9f3-11eb-8ed2-4ff5e9d3bf97.png)
It's very inconvient, and when I want to open a specific part, I often need to look manually for the good one (open and close each one, which is not ideal)

**Describe the solution you'd like**
Make the "name" field be displayed next to "Solid" in the tree (or anything which define a "name", basically a lot of thing)
Like with a GPS : 
![image](https://user-images.githubusercontent.com/83014015/121521485-6d6d1280-c9f4-11eb-9134-393c886c281f.png)

**Describe alternatives you've considered**
I could use a "DEF" keyword, but I don't consider to be the way to go, and additionally duplicate the name of the object
![image](https://user-images.githubusercontent.com/83014015/121521213-12d3b680-c9f4-11eb-955e-cf151ba8b606.png)

ShuffleWire:
https://github.com/cyberbotics/webots/pull/3198 is merged

PR #3198, opened 2021-06-17
Add the name of a solid next to it in tree

ShuffleWire:
Fix https://github.com/cyberbotics/webots/issues/3157

However, the fix is not the right way to do it, but I don't which one is the good one : 
- add "Solid" in either WbNodeUtilities::isDeviceTypeName or WbNodeUtilities::isSolidDeviceTypeName
- or remove the condition that this PR modify. In fact I don't understand why should we limit the display of the name to certains types. As long as the node have a "name" field, we should be good to display it.

### DIFF
--- a/src/webots/scene_tree/WbTreeItem.cpp
+++ b/src/webots/scene_tree/WbTreeItem.cpp
@@ -128,13 +128,9 @@ QString WbTreeItem::data() const {
   switch (mType) {
     case NODE: {
       QString fullName = mNode->fullName();
-      if (!fullName.startsWith("DEF ") && !fullName.startsWith("USE ")) {
-        if (mNode->nodeModelName() == "Solid" || WbNodeUtilities::isDeviceTypeName(mNode->nodeModelName())) {
-          WbSFString *name = mNode->findSFString("name");
-          if (name)
-            fullName += " \"" + name->value() + "\"";
-        }
-      }
+      WbSFString *name = mNode->findSFString("name");
+      if (name)
+        fullName += " \"" + name->value() + "\"";
       return fullName;
     }
     case FIELD: {
--- a/src/webots/scene_tree/WbTreeItem.cpp
+++ b/src/webots/scene_tree/WbTreeItem.cpp
@@ -128,9 +128,11 @@ QString WbTreeItem::data() const {
   switch (mType) {
     case NODE: {
       QString fullName = mNode->fullName();
-      WbSFString *name = mNode->findSFString("name");
-      if (name)
-        fullName += " \"" + name->value() + "\"";
+      if (!fullName.startsWith("DEF ") && !fullName.startsWith("USE ")) {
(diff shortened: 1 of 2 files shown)


omichel:
Thank you for this contribution.
I believe the second option you propose is the best: as long as the node has a "name" field (and no DEF name), we should use it to display it in the scene tree.

omichel:
It would be nice to also add a line in the change log about this.

ShuffleWire:
I'm not convinced if we should add the name for node which are def or use, however.

omichel:
USE/DEF name are displayed if specifed, this already gives a good indication on what the node is (if the USE/DEF name is well chosen). So, I would not display the "name" field in case the DEF/USE name is specified, as it would provide too much information which may be confusing.

ad-daniel:
Just a heads up, when the PR already has a reviewer assigned, re-request the review directly to that person after making the necessary changes. This way it's less likely to go unnoticed. Here's an example
rev

ShuffleWire:
> Just a heads up, when the PR already has a reviewer assigned, re-request the review directly to that person after making the necessary changes. This way it's less likely to go unnoticed. Here's an example

Thank for the advice :) 
However, afaik, I do think I can't to much more from my side : 
image
omichel is already requested, isn't ? and I don't think their are any pending requested change

src/webots/scene_tree/WbTreeItem.cpp
```diff
@@ -129,11 +129,9 @@ QString WbTreeItem::data() const {
     case NODE: {
       QString fullName = mNode->fullName();
       if (!fullName.startsWith("DEF ") && !fullName.startsWith("USE ")) {
-        if (WbNodeUtilities::isDeviceTypeName(mNode->nodeModelName())) {
-          WbSFString *name = mNode->findSFString("name");
-          if (name)
-            fullName += " \"" + name->value() + "\"";
-        }
+        WbSFString *name = mNode->findSFString("name");
```
omichel, line 132:
```suggestion
        const WbSFString *name = mNode->findSFString("name");
```

omichel (changes requested):
That looks very good to me.
Maybe we can just take advantage of this PR to make it const correct?

src/webots/scene_tree/WbTreeItem.cpp
ShuffleWire: of course !

omichel (approved):
Thank you.

ad-daniel:
> Thank for the advice :)
> However, afaik, I do think I can't to much more from my side :

rev
We generally handle the assignment of the reviewer, what I mean is after doing the changes and going out of draft, you can use this button to re-request the specific person for a second review.

ShuffleWire:
@ad-daniel ok so going in/out of draft is the trick. Will be trying that next time !

PR #3200, opened 2021-06-18
Fix field/context conversion from JavaScript to Lua

### LINKED ISSUE #3199: Colons are wrongly interpreted in inner Protos
**Describe the Bug**
If a proto uses another proto and specify it a textures url with a ":" in it, the interpreter will replace it with " =". Which will result in a invalid URL

**Steps to Reproduce**
1. Go to the develop branch
2. Open 'webots/projects/robots/mobsya/thymio/protos/Thymio2.proto'
3. Go to the line 316
4. Replace 'textures/thymio2_square_flare.png' with ' https://raw.githubusercontent.com/cyberbotics/webots/master/projects/robots/mobsya/thymio/protos/textures/thymio2_square_flare.png' 
5. Launch webots and open the 'obstacle_avoidance.wbt' world
6. See this error: ```WARNING: DEF THYMIO2 Thymio2 (PROTO) > Thymio2LED (PROTO) > Group > Transform > Shape > PBRAppearance > ImageTexture: 'https =//raw.githubusercontent.com/cyberbotics/webots/master/projects/robots/mobsya/thymio/protos/textures/thymio2_square_flare.png' not found.```

**Expected behavior**
The ':' should not be replaced in this situation

**Workaround**
Line 229 of WbProtoTemplateEngine.cpp, add a space after the colon.

ad-daniel:
**Addresses**
Fixes #3199.

**Description**
While Lua support is maintained, the formatting of fields and context has to be converted from JavaScript statements to Lua-friendly ones. Objects `{key: value}` are translated to tables `{key = value}`. If however the colon appears inside the value, it would also get replaced.

The proper way to handle it might be to use a regular expression like: `/(?<=\\w):\\s+(?=[['{\\d\\w-])/g` so that only the colons after the keys are replaced. 

The expression would specify a colon preceded by:

- word

followed by a space and:

- `[` : if value is a MFField
- `{`: if value is a Vector2/Vector3/SFNode
- `\\d`: if value is a SFFloat/SFInt
- `'`: if value is a string
- `-` if value is a negative SFFloat/SFInt.
- `\\w` if value is a word (true, false, undefined)

It works, but it almost feels overkill. The workaround of replacing `.replace(":", " =")` to `.replace(": ", " = ")` works so well that I'm tempted to just leave it like that.

Also, I took the opportunity to clean up how the the context strings are encoded, using `'` for string instead of `"`.

### DIFF
--- a/src/webots/vrml/WbProtoTemplateEngine.cpp
+++ b/src/webots/vrml/WbProtoTemplateEngine.cpp
@@ -62,25 +62,25 @@ bool WbProtoTemplateEngine::generate(const QString &logHeaderName, const QVector
   tags["fields"].chop(2);  // remove the last ",\n" if any
 
 #ifdef _WIN32
-  tags["context"] = QString("os: \"windows\", ");
+  tags["context"] = QString("os: 'windows', ");
 #endif
 #ifdef __linux__
-  tags["context"] = QString("os: \"linux\", ");
+  tags["context"] = QString("os: 'linux', ");
 #endif
 #ifdef __APPLE__
-  tags["context"] = QString("os: \"mac\", ");
+  tags["context"] = QString("os: 'mac', ");
 #endif
-  tags["context"] += QString("world: \"%1\", ").arg(worldPath);
-  tags["context"] += QString("proto: \"%1\", ").arg(protoPath);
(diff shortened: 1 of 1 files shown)


BenjaminDeleze (approved):
I think that we can go with the workaround as I cannot find a situation where it would fail.

PR #3202, opened 2021-06-21
Add Supervisor API to get field by index

stefaniapedrazzi:
Complete #3027 and synchronize with origin/develop branch:
* [x] minor fixes on the implementation
* [x] Add Matlab API
* [x] Add ROS API
* [x] Add documentation
* [x] Add tests


stefaniapedrazzi:
Only ROS test failed but seems unrelated to the latest changes.

docs/reference/supervisor.md
```diff
@@ -733,6 +757,8 @@ field = wb_supervisor_node_get_proto_field(node, 'field_name')
 | name | service/topic | data type | data type definition |
 | --- | --- | --- | --- |
 | `/supervisor/node/get_field` | `service` | `webots_ros::node_get_field` | `uint64 node`<br/>`string fieldName`<br/>`bool proto`<br/>`---`<br/>`uint64 field` |
+| `/supervisor/node/get_field_by_index` | `service` | `webots_ros::node_get_field_by_index` | `uint64 node`<br/>`int index`<br/>`bool proto`<br/>`---`<br/>`uint64 field` |
+| `/supervisor/node/get_number_of_fields` | `service` | `webots_ros::node_get_number_of_fields` | `uint64 node`<br/>`bool proto``---`<br/>`uint32 value` |
```
lukicdarkoo, line 761:
```suggestion
| `/supervisor/node/get_number_of_fields` | `service` | `webots_ros::node_get_number_of_fields` | `uint64 node`<br/>`bool proto`<br/>`---`<br/>`uint32 value` |
```

src/webots/nodes/utils/WbSupervisorUtilities.cpp
```diff
@@ -1133,6 +1149,37 @@ void WbSupervisorUtilities::handleMessage(QDataStream &stream) {
...
+      mFoundFieldCount = -1;
+      mFoundFieldIsInternal = false;
+
+      WbNode *const node = WbNode::findNode(node_id);
+      if (node) {
+        WbField *field = node->field(field_index, allowSearchInProto == 1);
+        if (field) {
+          WbMultipleValue *mv = dynamic_cast<WbMultipleValue *>(field->value());
+          mFoundFieldCount = mv ? mv->size() : -1;
+          mFoundFieldId = field_index;
```
lukicdarkoo, line 1170: Should we call it ID or index?

src/webots/nodes/utils/WbSupervisorUtilities.cpp
```diff
@@ -294,10 +294,12 @@ void WbSupervisorUtilities::initControllerRequests() {
   mFoundNodeParentUniqueId = -1;
   mFoundNodeIsProto = false;
   mFoundNodeIsProtoInternal = false;
+  mNodeFieldCount = -1;
   mFoundFieldId = -2;
```
lukicdarkoo, line 298: Not directly related to this PR, but this seems to be a bit cryptic, enum or similar would be helpful

lukicdarkoo (changes requested):
Thank you, it looks good, there are only a few minor comments.

Don't forget to add an entry to the changelog

src/webots/nodes/utils/WbSupervisorUtilities.cpp
stefaniapedrazzi: Index is the correct term for the public API: it represents the item at a given index in the node's list of field.
Fields objects don't have ids.
However the already existing code is (wrongly) referring to id also for fields (in `WbSupervisorUtilities.cpp` and `supervisor.c`).

In this PR I decided to not update/fix all the existing code variable and function names because it will cause many unrelated changes. But I can do it if you prefer.

src/webots/nodes/utils/WbSupervisorUtilities.cpp
stefaniapedrazzi: To me this is a normal behavior for variables meant to contain positive numbers like sizes or count:
`-1` means that it is not set, values `> 0` represent the found value, it is like NULL for pointers.
In any case, I don't see how you could use an enumeration to achieve the same result.
Note that any other solution would probably require additional variables.

lukicdarkoo: I was thinking something like this:
```cpp
enum { NOT_REQUESTED = -2, NOT_FOUND = -1 };

// ...

mFoundFieldId = NOT_REQUESTED;
```

stefaniapedrazzi: I'm not sure that mixing enum with int values is good.
With enumeration I would expect that all the possible values are listed in the enumeration and this won't be the case here.

If you want to add something like this, then I think you should not use an enumeration but simply define separate global variables for `NOT_REQUESTED` and `NOT_FOUND`.
However, this doesn't not only affect the two variables used in this PR but many others. So, if we want to improve this, it has to be done in a separate PR.

lukicdarkoo: > If you want to add something like this, then I think you should not use an enumeration but simply define separate global variables for `NOT_REQUESTED` and `NOT_FOUND`.

I agree, macros are probably a better idea.

> However, this doesn't not only affect the two variables used in this PR but many others. So, if we want to improve this, it has to be done in a separate PR.

👍

lukicdarkoo (approved):
Thank you

stefaniapedrazzi:
I will wait for the the approval of https://github.com/cyberbotics/webots_ros/pull/71 before merging this PR.

PR #3204, opened 2021-06-21
Fix Windows libcontroller determinism

lukicdarkoo:
This PR attempts to ensure Windows `libcontroller` binaries generated from the same source code are the same


omichel:
That's nice. However, I believe the workflow files should not be modified.

lukicdarkoo:
> That's nice. However, I believe the workflow files should not be modified.

Ups, sorry. I was running it a few times to confirm it works as expected.

omichel (approved):
Thank you.

PR #3206, opened 2021-06-21
fix matlab gitignore

BenjaminDeleze:
Regenerate the matlab controller .gitignore to include the `wb_supervisor_node_set_joint_position.m` file


lukicdarkoo (approved):
Thank you

stefaniapedrazzi:
Is the gitignore file automatically updated when running `mgenerate`?
Because it doesn't seem the case on my system.

Otherwise if other actions are needed we should also update the instructions here:
https://github.com/cyberbotics/webots/wiki/Adding-New-Node-and-API-Function#adding-a-new-api-function

BenjaminDeleze:
I was sure to have already updated the instructions, but it seems not. This is now fixed.

PR #3210, opened 2021-06-22
Don't create url for generic tabs in docs

ad-daniel:
**Description**
In certain situations when writing the documentaion the tab module can be useful to show good/bad practices, however by default all tab components are given an url which isn't desirable (and likely doesn't work if spaces are involved).

This PR:
-  only creates urls if the tab component is of type `tab-language`, `tab-os`
-  adds syntax highlighting for JavaScript.

### DIFF
--- a/docs/js/webots_documentation_loader.js
+++ b/docs/js/webots_documentation_loader.js
@@ -29,7 +29,7 @@ async function init() {
 
   await Promise.all(promises);
   script = document.createElement('script');
-  script.src = 'https://cyberbotics.com/wwi/R2021b/viewer.js';
+  script.src = '../resources/web/wwi/viewer.js';
   script.type = 'module';
   document.head.appendChild(script);
 }
--- a/docs/js/webots_documentation_loader.js
+++ b/docs/js/webots_documentation_loader.js
@@ -29,7 +29,7 @@ async function init() {
 
   await Promise.all(promises);
   script = document.createElement('script');
-  script.src = 'https://cyberbotics.com/wwi/R2021b/viewer.js';
+  script.src = 'https://cyberbotics.com/wwi/test_PR3156/viewer.js';
   script.type = 'module';
   document.head.appendChild(script);
 }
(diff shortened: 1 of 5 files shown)


omichel:
On this page: https://cyberbotics.com/doc/reference/javascript-procedural-proto?version=feature-generic-tab-module-in-docs, when I click on the "faster alternative" tab, the URL gets changed to https://cyberbotics.com/doc/reference/javascript-procedural-proto?version=feature-generic-tab-module-in-docs&tab-generic=faster%20alternative whereas it shouldn't (if I try to reload with the resulting URL, the corresponding tab is not selected).

ad-daniel:
I'm unsure about what's the right workflow here, as this branch is meant to be merged to another not yet merged into develop, am I supposed to overwrite the `viewer.js` in the server with my own? Being a minor change I don't expect it to break anything but it seems rather risky nonetheless.
Or by adding the `?version=` to the url it will automatically use my version of `viewer.js` as changed in this PR? If so then this mechanism might not be working given your results.

The way I tested it was to create a temporary test folder on the server and pointing `webots_documentation_loader` to that one, and then creating a local server (which by default will get the data from the server)

I re-added the test folder in case you want to test it that way.

omichel:
> Or by adding the ?version= to the url it will automatically use my version of viewer.js as changed in this PR?

Normally, by setting the `?version=` to the url, it should automatically use your version of `viewer.js`.

BenjaminDeleze:
> Normally, by setting the `?version=` to the url, it should automatically use your version of `viewer.js`.

No, `?version=` set the path to the resources hosted on github (.md, images,.x3d files...) but the .js files are hosted on the ftp server and there is only one version (R2021b) for all branches.

docs/js/webots_documentation_loader.js
```diff
@@ -29,7 +29,7 @@ async function init() {
 
   await Promise.all(promises);
   script = document.createElement('script');
-  script.src = 'https://cyberbotics.com/wwi/R2021b/viewer.js';
+  script.src = 'https://cyberbotics.com/wwi/test_PR3156/viewer.js';
```
omichel, line 32: This should be changed before the merge.

omichel (changes requested):
It looks very good.
Thank you!

omichel (approved):
Thank you.

PR #3221, opened 2021-06-24
Fix `test ros` on master

ad-daniel:
**Description**
Apply the same workflows to master branch as well, as done in develop in #3173

### DIFF
--- a/.github/workflows/test_suite_linux.yml
+++ b/.github/workflows/test_suite_linux.yml
@@ -56,7 +56,7 @@ jobs:
         export ROS_DISTRO=${{ matrix.ROS_DISTRO }}
         xvfb-run --auto-servernum make distrib -j4
     - name: Prepare Webots Controller Deployment
-      if: ${{ github.event_name == 'schedule' && matrix.os == 'ubuntu-18.04' }}
+      if: ${{ github.event_name == 'schedule' }}
       uses: webfactory/ssh-agent@v0.5.3
       with:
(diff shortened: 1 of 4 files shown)


ad-daniel:
It appears to be working, although I'm not sure why the previous version of the workflow (in master) restricts the key to only ubuntu 18.04 whereas in develop no distinction is made. I've changed it as it is in develop, but if somebody knows the reason that'd be better.

stefaniapedrazzi:
This is because master and develop branches are not synchronized.
The restriction on the master branch has been introduced by @lukicdarkoo recently (master branch only) in  #3180.

.github/workflows/test_suite_linux.yml
```diff
@@ -56,12 +56,12 @@ jobs:
         export ROS_DISTRO=${{ matrix.ROS_DISTRO }}
         xvfb-run --auto-servernum make distrib -j4
     - name: Prepare Webots Controller Deployment
-      if: ${{ github.event_name == 'schedule' && matrix.os == 'ubuntu-18.04' }}
+      if: ${{ github.event_name == 'schedule' }}
```
lukicdarkoo, line 59:
```suggestion
      if: ${{ github.event_name == 'schedule' && matrix.os == 'ubuntu-18.04' }}
```
Don't change these

lukicdarkoo (approved):
Thank you

Let's just wait for the tests

PR #3222, opened 2021-06-24
Resize labels on window resize

BenjaminDeleze:
Update the position and the size of the labels when resizing the window


omichel (approved):
Thank you.

PR #3226, opened 2021-06-24
Fix regeneration of a PROTO boundingObject

### LINKED ISSUE #3215: Crashes involving (but unrelated to) PROTO files
**Describe the Bug**
During the conversion of the Lua PROTO files, some crashes have been identified in some of them. As they're not regressions and that in all cases the PROTO loads fine at least once, it is unlikely that the problem is itself the PROTO file but rather something else.

Road-based PROTO:
~~- open `AddRoadLaneSegment.proto`~~
~~- attempt to delete a `RoadLine` from `lines`.~~
~~- webots freezes/crashes. Notice that doing so in the lowest level PROTO `Road.proto` appears to work fine, only in the derived ones the problem manifests. Might be related to the endless regeneration observed here #2994~~

`PipeBoundingObject.proto`: 
~~- create a world that uses it, for instance a solid with it.~~
~~- changing any of the parameters such as `radius` or `height` from the world file and reloading it works just fine.~~
~~- making the same changes from the interface makes it crash.~~

`SumoInterface.proto`:
- change the maximum number of vehicles to be smaller than the current number vehicles in the world, it crashes Webots.

ad-daniel:
**Fixes**
Fixes #3215 - second example.

**Description**
The MRE is a proto such as this:

```
#VRML_SIM R2021b utf8

PROTO Test [
  field SFFloat    height          2       # Defines the height of the pipe.
]
{
  %{ local a = fields.height.value; }%
  Group { # set of boxes
    children [
        Transform {
          children [
            Box {
              size %{= a }% %{= a }% %{= a }%
            }
          ]
        }
    ]
  }
}
``` 
When the user changes a parameter from the interface, it triggers a regeneration because this field is a template regenerator. As this PROTO is used in the `boundingObject`, and its parent is of type `Solid` it enters [this](https://github.com/cyberbotics/webots/blob/868b331bb7d29cc2ceed96233dd35c2752f1c84a/src/webots/nodes/utils/WbTemplateManager.cpp#L338-L340) condition. However, when it searches for the corresponding node in the parent fields (or rather, it requests its `nodeIndex`), it does so in the children field instead of the `boundingObject` field hence failing the assertion since it cannot find it there.


stefaniapedrazzi (changes requested):
The crash if the `PipeBoundingObject` PROTO is used with a parent basic `Solid` node is fixed.
But PROTO regeneration doesn't work correctly when the `PipeBoundingObject` PROTO is used as parameter of a parent PROTO (for example `Fence` or `Car`).
When inserting it (or after save and revert) it works correctly, but when changing the radius field errors/warnings are printed in the console and the bounding object is no longer generated.

stefaniapedrazzi:
If possible it would always be good to add a small test in the test_suite to prevent future regressions.

ad-daniel:
Yes, I'll be adding a test for this.
As for the other issue you brought up, as far as I can tell it isn't related to the fix in this PR as the same behavior occurs on 2021a.

The reason for the warning messages I kinda understand, there appears to be a strict enforcement taking place for `boundingObjects` (for instance, there cannot be a `Transform` within a `Transform`, even for base nodes) so while in the process of regenerating it probably triggers the messages (until the children is eventually added to it), as to why it doesn't accept the finalized node and instead deletes it I don't know.
Something similar happens if you add a `Transform { Transform { Box { }}}` boundingObject to a normal `Solid` (using base nodes) and load this world, it detects the 2 layers and deletes them, leaving only the first `Transform` giving the message `A child to the Transform placed in 'boundingObject' is expected.`.
The structure of this specific PROTO is `Group { Transform { Box {}}}`, which when doing this same experiment appears to be a valid `boundingObject` (i.e doesn't delete it), but I suspect it's the same mechanism that deletes it when changing one of its parameters.
Either way, maybe it should be addressed in a different PR.

stefaniapedrazzi:
In this PR or in another (it is the same), this should be fixed because it is a bug.
It is is good to have fixed the crash but the regeneration of a PROTO boundingObject is still buggy and `PipeBoundingObject` and similar PROTO nodes cannot be correctly regenerated.
At load they work correctly but no regeneration occurs.

ad-daniel:
The problem of the unrelated bug you mentioned stems from a wrongful validation of the newly regenerated node [here](https://github.com/cyberbotics/webots/blob/e4a6da3ac2c06c66790760cc2eca3ac20014c819/src/webots/nodes/utils/WbTemplateManager.cpp#L281), which results in its deletion. The reason it isn't validated is because despite the node in question being in a `boundingObject`, [this test](https://github.com/cyberbotics/webots/blob/e4a6da3ac2c06c66790760cc2eca3ac20014c819/src/webots/nodes/utils/WbTemplateManager.cpp#L223) returns false. The reason is that this specific case both the `boundingObject` and the node that uses it are PROTOs. As such, the pointer to the `boundingObject` the parent node points to can also be a `protoParameterNode`, which the `isInBoundingObject` doesn't consider. This also explains why using the `pipeBoundingObject` on a base node was working just fine.

I'll add a test for both.

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -929,12 +929,19 @@ bool WbNodeUtilities::isInBoundingObject(const WbNode *node) {
   const WbNode *const p = node->parentNode();
   if (p) {
     const WbSolid *const s = dynamic_cast<const WbSolid *>(p);
+    const WbFluid *const f = dynamic_cast<const WbFluid *>(p);
```
stefaniapedrazzi, line 932: Probably we can also improve the code by using directly `WbMatter` instead of both `WbSolid` and `WbFluid`.

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -929,12 +929,19 @@ bool WbNodeUtilities::isInBoundingObject(const WbNode *node) {
...
+      boundingObject = f->boundingObject();
 
-    const WbFluid *const f = dynamic_cast<const WbFluid *>(p);
-    if (f)
-      return f->boundingObject() == node;
+    while (boundingObject) {
+      if (boundingObject == node)
+        return true;
+      else
+        boundingObject = boundingObject->protoParameterNode();
```
stefaniapedrazzi, line 943: `else` is useless given the previous return
```suggestion
      boundingObject = boundingObject->protoParameterNode();
```

stefaniapedrazzi (changes requested):
Thank you.

I just have a couple of minor comments.

src/webots/nodes/utils/WbNodeUtilities.cpp
ad-daniel: Good point, it worked locally. Just waiting for the CI.
Also had to sync with develop because of a conflict.

stefaniapedrazzi (approved):
Thank you

Issue #2736, opened 2021-02-10
Lidar distances are somtimes too big if the ray hits an edge


Thomas-Feldmeier:
**Describe the Bug**
the lidar rays at the left/right edge of an obstacle sometimes measure a too big value, generating false data (see picture).
there are some blue points in empty space.

**Expected behavior**
It should be possible to use lidars without this problem. Maybe a parameter? (real lidars have less errors)

*Screenshots*

![LidarBug](https://user-images.githubusercontent.com/73528085/107527566-365b3180-6bb9-11eb-925b-9c4007f89a88.png)

![RobotinoTHU_1](https://user-images.githubusercontent.com/73528085/107667522-edc27780-6c8f-11eb-941d-c2d45687f2f8.png)

(implementation: the z-buffer distance of only one pixel at the center of the ray instead of averaging a lot of pixels?)

lukicdarkoo:
@mael25 There is a chance it is related to #37, maybe related to the spherical deformation

lukicdarkoo:
I have a theory for this one.

The point cloud is calculated from a depth camera that has the same horizontal resolution as the lidar:
https://github.com/cyberbotics/webots/blob/60d017e8f48caf1863f5d3a93f3937d3bc682075/src/webots/nodes/WbAbstractCamera.cpp#L497

Let's say we have a high-resolution depth camera that detects an edge of an object as follows:
`| wall | wall | wall | object | object | object |`

Now, try to reduce the resolution of the camera by x2:
`| wall | wall/object | object |`
Instead of encoding the edge in two pixels, it is now encoded in a single pixel. Which means that the edge has to be averaged.

To avoid the problem we can take high-resolution cameras and pick every nth pixel. But it is very expensive.

Another solution would be to tell OpenGL the way the edge pixels are encoded. I didn't check the part of the code that generates depth images, but I guess it is possible.

mael25:
I have some precisions on the issue after investigating:

- It occurs only if the `spherical` field of the `Lidar` is set to `true`.
- In this case, one or multiple sub-cameras are activated according to the needs (verical/horizontal `fieldOfView`). Their resolution is different than the one set for the `Lidar` but the output of the `merge_spherical.frag` shader meets the resolution of the `Lidar`.
[...]

PR #3230, opened 2021-06-25
Fix error in Lidar / RangeFinder measurement when hitting an edge

mael25:
**Description**
When hitting an edge, the Lidar / RangeFinder was returning values between the two possible values. This was because the values were sampled slightly in-between pixels of the camera.
This is solved by fetching the pixels directly by index instead of coordinates. We multiply the coordinates by the resolution, round it and we get the correct index.

**Related Issues**
This pull-request fixes issue #2736

lidar


resources/wren/shaders/merge_spherical.frag
```diff
@@ -110,20 +112,21 @@ void main() {
     coord.y *= pi_2 / fovY;
 
   vec2 faceCoord = vec2(0.5 * (1.0 - coord.x), 0.5 * (1.0 - coord.y));
```
lukicdarkoo, line 114: Do we need `faceCoord` and `coord` variables now?

lukicdarkoo (commented):
Nice!

resources/wren/shaders/merge_spherical.frag
mael25: `coord` yes as there are several operations on it, `faceCoord` could be indeed dropped by computing directly the index.

lukicdarkoo: Sorry, I missed it, it is good as is

omichel:
It seems the range finder tests are failing.
Shall we fix the tests?

mael25:
Here is a comparison of how the depth look with multiple layers before projection (obstacle is the ground and a wall, top view):
Develop:
fragcolor_develop
This PR:
fragcolor_new

mael25:
Comparison of the output of `texelFetch` (left) and `texture` (right)

comparison

mael25:
And after the spherical transform projection code, we get:
projected

mael25:
So it seems the new version with `texelFetch` on the left. although less smooth as there is no value taken between pixels, reflects the measurements actually taken by the camera.

However, it seems that there is a vertical offset (already existent) when we have multiple layers:
comparison2

mael25:
The `range_finder_spherical_horizontal_vertical` test passes locally, and for the `range_finder_horizontal_spherical_projection` test (which controls the distance of the lowest layer lidar points and the ground) to pass, we would need to adapt the allowed error: currently, the difference of depth measured for this layer at the center of the sub-camera and at the junction of the sub-camera should not exceed 0.01. 

Now, as shown above (https://github.com/cyberbotics/webots/pull/3230#issuecomment-869504991) the intersection circle with the ground that should appear is quite truncated so the error can go up to 0.1. Otherwise, we could assert instead than the ratio `sub_camera_center_depth/sub_camera_junction_depth` is bigger than 0.988

The vertical offset happening in both `develop` and this PR could be adressed in another PR...

mael25:
I made the following changes to the `RangeFinder` tests:

- In `range_finder_horizontal_spherical_projection`, the allowed error is 0.1 instead of 0.01.
- In `range_finder_spherical_horizontal_vertical`, if a pixel is `inf` and the one in the other `RangeFinder` is not, then the non-inf previous/next pixel (if existent) are considered for the assertion.

tests/api/controllers/range_finder_horizontal_spherical_projection/range_finder_horizontal_spherical_projection.c
```diff
@@ -27,7 +27,7 @@ int main(int argc, char **argv) {
     float sub_cameras_junction_depth = wb_range_finder_image_get_depth(image, width, x1, height - 1);
     float sub_camera_center_depth = wb_range_finder_image_get_depth(image, width, x2, height - 1);
     ts_assert_double_in_delta(
-      sub_cameras_junction_depth, sub_camera_center_depth, 0.01,
+      sub_cameras_junction_depth, sub_camera_center_depth, 0.1,
```
lukicdarkoo, line 30: Depth sensors (e.g. https://www.photoneo.com/products/phoxi-scan-xl/) have an accuracy of 0.5mm at 2-3 meters. Even cheap depth sensors (like RealSense D435) have an accuracy of up to 4cm. Therefore, we should be careful with these thresholds.

mael25:
The vertical offset error is now solved. It was due to the fact that `fovYCorrectionCoefficient` was not taken into account when computing the face coordinates in the `merge_spherical.frag` shader. 
Screenshot from 2021-06-30 16-07-02
`fovYCorrectionCoefficient` is used in `WbWrenCamera.cpp` to artificially increase the vertical field of view of the horizontal sub-cameras in case the vertical field of view requested by the user is just below 90deg (so just before the UP and DOWN sub-cameras are needed). This aimed to correct the error with the spherical projection.


Now that this bug is fixed, here's a comparison of the result between `develop` (top) and this PR (bottom).
tower6_develop
tower6_pr

tests/api/controllers/range_finder_horizontal_spherical_projection/range_finder_horizontal_spherical_projection.c
mael25: > Depth sensors (e.g. https://www.photoneo.com/products/phoxi-scan-xl/) have an accuracy of 0.5mm at 2-3 meters. Even cheap depth sensors (like RealSense D435) have an accuracy of up to 4cm. Therefore, we should be careful with these thresholds.

This so big difference was in fact due to the fact of those square corners as you can see on the first picture (left). This is not a regression, it is also present in `develop` (see second picture), but in `develop` there is an additional issue: there is an angle offset between the theoretical location of the lidar layers according to Webots interface and the actual place where the bottom layer is shown (1st picture, right).

error
Projection test: This PR (left), `develop` (right)
range_finder_horizontal_spherical_projection
Same configuration but with the `Lidar` a little closer to the ground (`develop`) to show the square corners.

Which means that to have a fair comparison, I changed again the test so that instead of looking at the 79th layer, we look at the 70th which is located above this square corners. This allows then to put the maximum difference between a sub-camera's center and junction depths to 4cm.

The issue with the square corners could be addressed in another pull request.

mael25:
@lukicdarkoo It seems that the `range_finder_spherical_horizontal_vertical` test (which compare the output of a 128x1 `RangeFinder` and a 90 deg rotated 1x128) does not pass on the CI because the behaviour with OpenGl is not exactly the same.

I try to allow up to two pixels of error but it still does not pass. As this test pass locally, I suggest that, after review, we merge this PR even though the test fails on the CI so that we can proceed with the release, and eventually correct the behaviour on the CI or adapt the test.

mael25:
I try one last modification of the test by increasing the height of the obstacle

mael25:
With Mesa 20.2.6, the `range_finder_spherical_horizontal_vertical` test passes and the vertical `RangeFinder` image is the same as the horizontal one.

However, it has sometimes trouble displaying the points from the lateral cameras (they are displayed only by moments). This is also the case on the `master` branch and has no effect on the test as the object to detect are in front of the `RangeFinder`.

mael25:
Adding the LIBGL_ALWAYS_SOFTWARE environment variable (`export LIBGL_ALWAYS_SOFTWARE=1`) enabled to reproduce the issue encountered in the CI.
It turned out the issue with the `range_finder_spherical_horizontal_vertical` CI test was due to the index being rescaled from a range of [0.1] to [0,resolution] instead of [0,resolution-1]

The slight change to the range made the other test (`range_finder_horizontal_spherical_projection`) fail. I had to slightly increase the allowed error (see https://github.com/cyberbotics/webots/pull/3230#discussion_r662071776)

omichel (approved):
Thank you.

PR #3234, opened 2021-06-28
Enhacement ros camera RecognitionObjects message

stefaniapedrazzi:
Addresses https://github.com/cyberbotics/webots_ros/issues/64:

Change the `ros` default controller so that it publish a single message including all the recognized objects instead of publishing one message per recognized object.


stefaniapedrazzi:
It would be good to also have a functional review to check for md5sum issues.
Locally I had issues with md5sum mismatch for the `RecognitionObjects.h` file so I applied the same fix as in #1641 and now it works on my system.

However, even if unrelated, I still have md5sum mismatch (ROS/catkin_ws vs Webots) issues for these services:
* `/robot/supervisor/node/get_pose`
* `/robot/supervisor/node/get_field_by_index`
So, it would be good to have a feedback when running the `complete_test` to check if it is just a local issue because I'm not sure these errors are detected by the automatic test.

projects/default/controllers/ros/headersGenerator.py
```diff
@@ -40,6 +40,7 @@ class HeadersGenerator:
         'geometry_msgs/Point': '4a842b65f413084dc2b10fb484ea7f17',
         'geometry_msgs/PointStamped': 'c63aecb41bfdfd6b7e1fac37c7cbe7bf',
         'geometry_msgs/Quaternion': 'a779879fadf0160734f906b8c19c7004',
+        'geometry_msgs/Transform': 'ac9eff44abf714214112b05d54a3cf9b',
```
stefaniapedrazzi, line 43: This fixes the md5sum mismatch for the `/supervisor/node/get_pose` service.

lukicdarkoo (approved):
It looks good, but I would delete the `uint32 count` field as it can be obtained as `objects.size()`

stefaniapedrazzi:
I removed the count field.
It should now be ready for review.

lukicdarkoo (approved):
Thank you

Liuxingyuming:
Sorry. I always get an empty array of object when using the topic "recognition_objects".  I'm wondering what could be the reason.
6ce60df0c48bda3fb262108d1f4c2a3
And this is my code:
`void camera_topCallback(const webots_ros::RecognitionObjects::ConstPtr &msg)
{
    camera_topRawvalue = *msg;
    ROS_INFO("Update camera_topRawvalue: numberofobj %d", msg->objects.size());
    //ROS_INFO("FIRST OBJ: %s", msg->objects[0].model.c_str());
}`

PR #3235, opened 2021-06-28
Fix `wb_supervisor_node_get_contact_points` caching

lukicdarkoo:
The `contact_points[descendants].last_update` was initialized to a random value, sometimes higher than 0, making the caching mechanism break.

Also, delete deprecated documentation:
https://www.cyberbotics.com/doc/reference/supervisor?version=fix-get-contact-points#wb_supervisor_node_get_contact_points

### DIFF
--- a/docs/reference/supervisor.md
+++ b/docs/reference/supervisor.md
@@ -1041,10 +1041,6 @@ The "[WEBOTS\_HOME/projects/samples/howto/center\_of\_mass/worlds/center\_of\_ma
 #### `wb_supervisor_node_enable_contact_point_tracking`
 #### `wb_supervisor_node_disable_contact_point_tracking`
 
-#### `wb_supervisor_node_get_contact_point`
-#### `wb_supervisor_node_get_contact_point_node`
-#### `wb_supervisor_node_get_number_of_contact_points`
-
 
 %tab-component "language"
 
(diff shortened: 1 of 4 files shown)


ad-daniel:
Looks good to me, the fix resolves the problem I was having.
For the docs, I don't know if there's other situations like this but maybe it would be better to do something like:
```
[
  ... API ...
]

description:
- describe available function 1 
- describe available function 2
- etc

deprecated functions:
- function 3 (replace by .... ): describe deprecated function 3
- function 4 (replace by .... ): describe deprecated function 4
```

just throwing an idea

Just waiting on the tests to see if something else broke because of it, thank you

docs/reference/supervisor.md
```diff
@@ -1164,13 +1164,17 @@ The `sampling_period` argument determines how often the contact point data shoul
 The `wb_supervisor_node_disable_contact_point_tracking` function disables contact point data tracking.
 
 The `wb_supervisor_node_get_contact_point` function returns the contact point with given index in the contact point list of the given `Solid`.
+**Deprecation warning:** This function is deprecated and the `wb_supervisor_node_get_contact_points` function should be used instead.
```
stefaniapedrazzi, line 1167: Deprecated functions should be removed from the reference manual.

lukicdarkoo: But the differential drive wasn't deleted (https://www.cyberbotics.com/doc/reference/differentialwheels), neither was `robot.get[DEVICE_TYPE]` (https://cyberbotics.com/doc/reference/robot?tab-language=python#wb_robot_get_device). I am not sure what is the best way to deal with deprecated functions.

omichel: I believe removing them from the documentation is one of the best way to discourage their use. So, I would go ahead a remove them from the documentation. Differential drive and `robot.get[DEVICE_TYPE]` should be deleted in a separate PR.

stefaniapedrazzi: We always deleted all the deprecation functions from the documentation.
In these two cases it is possible that we forgot to do it.

omichel: DifferentialWheels is already deleted from the develop branch. I will take care of deleting the python `robot.get[DEVICE_TYPE]` now.

ad-daniel (approved):
Thank you!

PR #3237, opened 2021-06-28
Fix a controller crash when changing the controller

### LINKED ISSUE #3231: A controller crashes after setting it for the first time
**Steps to Reproduce**
1. Open e.g. `distance_sensor.wbt`
2. Run it and reset
3. Change controller to `void`
4. Run and see the controller error:
    ```
    abort: robot.c:361: assertion failed: tag < robot.n_device
    WARNING: 'void' controller exited with status: 1.
    ```
5. If you play a bit with changing the controllers you will also get:
    ```
    WARNING: distance_sensor: The process crashed some time after starting successfully.
    WARNING: 'distance_sensor' controller crashed.
    ```

The error occurs almost every time (if not every time) one changes the controller. This is tested with R2021b, but I believe it is a very old bug.

lukicdarkoo:
Fixes #3231

In addition, this PR solves the Webots crash that can be produced as follows:
- open a world,
- pause and reset,
- set controller B,
- revert to controller A,
- run crashes Webots.

---

After the reset, the running controller (let's call it `A`) is stopped and added to the `mWaitingControllers` list. Once the controller parameter is switched from `A` to `B` the controller `B` is added to the `mWaitingControllers` as well (but the controller `A` is not deleted as it is not in the `mControllers` list). It makes both controllers be executed at the beginning of the simulation and making both of them crash.

This PR deletes the robot's controller from the `mWaitingControllers` and `mNewControllers` lists before adding a new one.

I am not sure whether I am breaking something though.


mael25:
This solves the issue, however Webots crashes with a segmentation fault in some cases when reloading the world.
For instance:

1. Open the world with controller A
2. Pause and Reset
3. Select controller B and Play
4. Reload world

lukicdarkoo:
@mael25 I could reproduce the crash only in a small percentage of cases, so I hope it is fixed now (I cannot reproduce the crash anymore).

mael25:
I still get the crash, but not each time...

mael25:
I still have sometimes a crash when I:

- Start the world with controller A
- Pause, Reset, switch to controller B, play
- Pause, Reset, switch back to controller A, play
- Reload -> segmentation fault

mael25 (approved):
It works now, thank you!

PR #3240, opened 2021-06-28
Fix mass computation when inserting a Physics node

### LINKED ISSUE #3009: Wrong warning about 'boundingObject' not defined
A user reported that a wrong warning is printed in the Webots console when adding a Physics node if a Mesh node is used in the boundingObject saying that the boundingObject is not defined.
This is not only reproducible with Mesh nodes but with other nodes too.

**Steps to Reproduce**
1. Open `propeller.wbt`
2. Select the `RED_HELICOPTER` Robot
3. Delete its `boundingObject` and `physics` nodes.
4. Add a Box node (or any node) in the `boundingObject` field.
5. Add a Physics node in the `physics` field.
6. The error message is printed in the console:
    ``WARNING: DEF RED_HELICOPTER Robot: Mass is invalid because 'boundingObject' is not defined. Using default mass 

Note that after saving and reverting no warning message is printed in the console.

stefaniapedrazzi:
Address #3009.
If:
1) `boundingObject` is defined when `physics` is NULL and
2) `pyhsics` node is added just after

the solid merger mass is not correctly set.
Not only the wrong message about `boundingObject` not defined is printed in the console, but other issues arise, for example the Solid mass in the field editor is not updated when modifying the bounding object.


BenjaminDeleze (dismissed):
Thank you!

stefaniapedrazzi:
I just fixed a typo

BenjaminDeleze (approved):
Oups, I missed it

Issue #3287, opened 2021-07-02
Missing textures


ad-daniel:
**Describe the Bug**
Getting:
`WARNING: RandomBuilding (PROTO) > DEF WALL Shape > PBRAppearance > ImageTexture: Cannot download https://raw.githubusercontent.com/cyberbotics/webots/R2021b/projects/objects/buildings/protos/textures/old_brick_wall.jpg: Error opening https://raw.githubusercontent.com/cyberbotics/webots/R2021b/projects/objects/buildings/protos/textures/old_brick_wall.jpg`

when opening `CH_Vens.wbt`.

ad-daniel:
To avoid opening an issue each time, here's another in `robotstadium_nao_vs_robotis-op2`: 
`
WARNING: RobotstadiumSoccerField (PROTO) > RobotstadiumGoal (PROTO) > Shape > DEF NET_APP PBRAppearance > ImageTexture: Cannot download https://raw.githubusercontent.com/cyberbotics/webots/R2021b/projects/objects/robotstadium/protos/textures/net.png: Error opening https://raw.githubusercontent.com/cyberbotics/webots/R2021b/projects/objects/robotstadium/protos/textures/net.png
`

PR #3314, opened 2021-07-02
Fix bad url for robostadium net

ad-daniel:
**Description**
Ditto. Fixes: https://github.com/cyberbotics/webots/issues/3287#issuecomment-872906344


lukicdarkoo (approved):
Thank you

PR #3317, opened 2021-07-02
Recommend for all users installation on Windows

omichel:
It is usually better to install Webots for all users (if the user has administrator privileges):
image

Moreover, this PR allows users to specify the type of installation also on the command line.


ad-daniel:
I'll test this when the new nightly gets built.

ad-daniel (approved):

PR #3323, opened 2021-07-05
Fix nao room demo

### LINKED ISSUE #3277: Nao room in guided tour is failing
The `nao_room.wbt` launched from the guided tour is failing with the following error (on Windows):
```
WARNING: Background: Texture dimension mismatch between leftUrl and rightUrl.
WARNING: "python.exe" was not found.
Webots requires Python version 3.9, 3.8, 3.7 or 2.7 (64 bit) from python.org in your current PATH.
To fix the problem, you should:
1. Check the Python command set in the Webots preferences.
2. Check the COMMAND set in the [python] section of the runtime.ini file of your controller program if any.
3. Fix your PATH environment variable to use the required Python 64 bit version (if available).
4. Install the required Python 64 bit version and ensure your PATH environment variable points to it.
```

omichel:
Fixes  #3277.

### DIFF
--- a/projects/robots/softbank/nao/worlds/nao_room.wbt
+++ b/projects/robots/softbank/nao/worlds/nao_room.wbt
@@ -139,7 +139,7 @@ WoodenChair {
 Nao {
   translation 0.997971 0.34 -1.74563
   rotation 0.57735126918836 0.57735126918836 0.5773482691869614 -2.094395307179586
-  controller "nao_demo_python"
+  controller "nao_demo"
   cameraWidth 320
   cameraHeight 240
 }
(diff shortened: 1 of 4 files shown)


BenjaminDeleze:
The background did not load correctly because of the following points:

- The background was using several time the same texture (.png)
- So it was cached by the intermediary cache of WbDownloader. Thus the texture was downloaded only once.
- Problem: WbBackground is using cubemap textures and not 2d textures. And Wren only has a cache for 2d textures. So it was not possible to get the textures from the Wren cache as their is no cache.

Solution:

- I went for the simplest solution: I introduced a boolean such that the downloader ignore the intermediary cache if the URL comes from a background.

BenjaminDeleze (approved):
Thank you!

Issue #3304, opened 2021-07-02
Using ROS instructions


mael25:
**Describe the Bug**
In the [Installing ROS and "webots_ros" Package](https://www.cyberbotics.com/doc/guide/tutorial-8-using-ros#installing-ros-and-webots_ros-package) instructions of the User Guide, at points 3, 5 and 6, the path given use `melodic`.

It might be good to highlight the fact that it has to be changed according to the ROS distribution installed and/or put `noetic` instead.

lukicdarkoo:
We should also delete `kinetic` as it is not supported anymore.

PR #3335, opened 2021-07-05
Update Using ROS instructions

mael25:
**Description**
In ROS instructions, set `noetic` as default and get rid of `kinetic`. (Issue #3304)


BenjaminDeleze (approved):
Thank you

Issue #3311, opened 2021-07-02
`thymio2_aseba` controller & robot window integration not working


ad-daniel:
**Describe the Bug**
In worlds `thymio2_ball.wbt` and `thymio2_pen.wbt`, the robot window doesn't allow to control the robot and therefore nothing happens.
By replacing the default controller with `thymio2_demo` both words become usable.

PR #3338, opened 2021-07-05
Replace default controller for `thymio2_pen.wbt` and `thymio2_ball.wbt`

ad-daniel:
**Description**
Fixes https://github.com/cyberbotics/webots/issues/3311.

The default controller `thymio2_aseba` is not integrated with the robot window therefore these two worlds cannot be played without external software. It is too premature to remove this external dependency altogether but this change at least makes the worlds playable.

The world `thymio2.wbt` already uses this controller.


mael25 (approved):
Thank you!

PR #3342, opened 2021-07-05
add information about cache size used

BenjaminDeleze:
**Description**
Add information about how many MB of the cache are actually filled in the preferences.

I am not sure about how to integrate this information


ad-daniel:
The implementation works, as to what is the best way of showing this information I honestly have no idea. 

- raw MB or percentage?
- on a label or directly on the button itself?  Like `Clear the cache (256 MB)`

This is how it is now

cache

what do you think @omichel?

src/webots/gui/WbPreferencesDialog.cpp
```diff
@@ -385,7 +390,7 @@ QWidget *WbPreferencesDialog::createNetworkTab() {
   QGridLayout *network = new QGridLayout(widget);
   QGroupBox *proxy = new QGroupBox(tr("Proxy"), this);
   proxy->setObjectName("networkGroupBox");
-  QGroupBox *cache = new QGroupBox(tr("Cache"), this);
+  QGroupBox *cache = new QGroupBox(tr("Disck cache"), this);
```
ad-daniel, line 393:
```suggestion
  QGroupBox *cache = new QGroupBox(tr("Disk cache"), this);
```

ad-daniel (changes requested):
Just a typo, looks good to me otherwise

ad-daniel (approved):
Thank you. Worth waiting for the tests to pass since I merged master into this branch.

Issue #3302, opened 2021-07-02
Resetting `supervisor.wbt`, the light doesn't revert to initial position


ad-daniel:
**Describe the Bug**
Steps:
- Load
- Run
- Reset

The light doesn't return to the initial position.

PR #3345, opened 2021-07-05
Fix pointlight reset

ad-daniel:
**Description**
Fixes https://github.com/cyberbotics/webots/issues/3302.

Contrary to other light nodes, the PointLight has a `location` field which isn't otherwise reset. I believe during a reset it is desirable for the light to return to whatever state it was previously set to (be it at `__init__` or another `stateid`).


ad-daniel:
Here's an quick and dirty example to test that the resetting now works [pointlight.zip](https://github.com/cyberbotics/webots/files/6768041/pointlight.zip)

omichel (approved):
Thank you.

PR #3348, opened 2021-07-06
Fix javascript construct

omichel:
The JavaScript "for of" construct doesn't work on Windows where the JS engine is older than the one on Linux and macOS. Therefore, we have to revert to an older construct to iterate on items.


BenjaminDeleze (approved):
Thank you!

PR #3349, opened 2021-07-06
Improve Urls handling

### LINKED ISSUE #3283: Momentary white textures when regenerating or resetting
**Describe the Bug**
To reproduce:
- open `mantis.wbt`
- reset.

Or:
- add arena
- change arena size.

In either case, after doing so the textures are white for like 1s before returning.
I imagine it's related to the fact that it has to check whether the server has a newer version of the textures. I don't know if it's something that can be avoided.

**System**
 - Operating System: Windows

BenjaminDeleze:
**Description**
Several improvements to webots when worlds use URLs

**Tasks**
  - [x] Pause the simulation until all the textures are ready (on reset for example), fix #3283 
  - [x] Change the default cache policy from `prefer network` to `prefer cache`.
  - [x] Modify the expiration date to be one day. (default = 5min).

### DIFF
--- a/src/webots/engine/WbSimulationWorld.cpp
+++ b/src/webots/engine/WbSimulationWorld.cpp
@@ -398,7 +398,8 @@ void WbSimulationWorld::reset(bool restartControllers) {
     }
   }
   updateRandomSeed();
-  WbSimulationState::instance()->resumeSimulation();
+  if (WbDownloader::progress() == 100)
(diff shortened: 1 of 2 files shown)


ad-daniel (approved):
Looks good to me. I've also tested reset from a supervisor and works as intended. Thank you.

omichel (approved):
Thank you.

Issue #3337, opened 2021-07-05
Cabinet: not possible to replace primaryAppearance


BenjaminDeleze:
**Describe the Bug**
If I delete the primaryAppearance of the cabinet Proto I cannot put another one instead.

**Steps to Reproduce**
1. Go to 'kitchen.wbt'
2. Select a cabinet
3. Delete a primaryAppearance
4. click to primaryAppearance to add a new one
5. The add node pop up but no node can be selected

**System**
 - Ubuntu 20

**Additional context**
secondaryAppearance is working fine

PR #3350, opened 2021-07-06
Fix cabinet `primaryAppearance`

ad-daniel:
**Description**
Fixes https://github.com/cyberbotics/webots/issues/3337.

The MRE of this bug is this PROTO
```
PROTO Cabinet [
  field SFNode   primaryAppearance   PaintedWood {}    # Defines the primary appearance.
  field SFNode   secondaryAppearance PaintedWood {}    # Defines the secondary appearance.
]
{
  Solid {
    children [
      CabinetDrawer{
         primaryAppearance IS primaryAppearance
         secondaryAppearance IS secondaryAppearance
      }
      CabinetDoor {
         primaryAppearance IS primaryAppearance
         secondaryAppearance IS secondaryAppearance
      }
    ]
  }
}
```
The reason for the bug is that although `CabinetDoor` has a `primaryAppearance` parameter, this is not used anywhere in the body of the PROTO. As such, when removing the default node and attempting to add a new one (in the parent PROTO `Cabinet`), the candidates must satisfy both the constraints imposed by `CabinerDrawer` (which uses this field) and `CabinetDoor` (which doesn't) and the result is that no node satisfies both.

The simplest solution is to remove this unnecessary field from `CabinetDoor`. No update to the documentation is necessary since this PROTO is hidden.

PS: At some point we should look into the `isAllowedToInsert` function, trying to debug this from the webots side was a nightmare. There has to be a cleaner way of checking node compatibility.

### DIFF
--- a/projects/objects/cabinet/protos/Cabinet.proto
+++ b/projects/objects/cabinet/protos/Cabinet.proto
@@ -206,7 +206,9 @@ PROTO Cabinet [
                   %< } >%
                   thickness IS innerThickness
                   handle IS handle
-                  primaryAppearance IS primaryAppearance
+                  %< if (data[0] === 'Drawer') { >%
+                    primaryAppearance IS primaryAppearance
+                  %< } >%
                   secondaryAppearance IS secondaryAppearance
                 }
             %< } else if (data[0] === 'Shelf') { >%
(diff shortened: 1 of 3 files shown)


mael25 (approved):
Congratulations on finding the bug!

Issue #3331, opened 2021-07-05
Very noticeable floor texture distorsions in `inverted_pendulum.wbt`


ad-daniel:
**Describe the Bug**
Ditto.

https://user-images.githubusercontent.com/44834743/124439189-662df000-dd79-11eb-870c-c1ee331396e0.mp4

Tested on Windows Nightly

EDIT: Probably view-point related as changing the viewpoint swaps which side has the distortions.

omichel:
I cannot reproduce this problem on my Windows machine in the development environment.

ad-daniel:
Same distortion appears on ubuntu 20 development environment.

https://user-images.githubusercontent.com/44834743/124453801-07bc3e00-dd88-11eb-9c6d-ab731fe1506d.mp4

BenjaminDeleze:
I cannot reproduce it on my ubuntu 20 on master. Do you make something special ?

ad-daniel:
No, just pulled latest master. Could it be AMD GPU related?

omichel:
> Could it be AMD GPU related?

Maybe.

PR #3351, opened 2021-07-06
"Fix" floor distorsion

ad-daniel:
**Description**
Fixes https://github.com/cyberbotics/webots/issues/3331.

Not exactly a fix, but I've noticed that choosing a size of 1000 or less and the issue virtually disappears. For the purpose of this world, even 100 would've largely sufficed.

The problem appears to be only affect `IndexedFaceSets` which are defined as a single quad (or 2 triangles) which are very long. I don't think it's very common to have a single 10 km-sided triangle. The only case I can think of is a 10 km straight `Road`. However after testing one, the issue doesn't show up at all even for a length of 100 km. Roads however are built as 8 triangles though, so even this case doesn't really fit the bill.

I've resized the floor so that the distortion doesn't appear, but I'm not sure whether it's practically useful to investigate it further given how niche a case it appears to be. I've put a `wontfix` label, but feel free to remove it.

https://user-images.githubusercontent.com/44834743/124601739-7b357c80-de68-11eb-8b29-66283f444043.mp4

### DIFF
--- a/projects/samples/howto/inverted_pendulum/worlds/inverted_pendulum.wbt
+++ b/projects/samples/howto/inverted_pendulum/worlds/inverted_pendulum.wbt
@@ -23,7 +23,7 @@ Viewpoint {
 TexturedBackground {
 }
 Floor {
-  size 2 10000
+  size 2 1000
   appearance PBRAppearance {
     baseColorMap ImageTexture {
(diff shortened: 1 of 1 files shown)


omichel (approved):
Great.

PR #3359, opened 2021-07-06
Fix nightly urls

### LINKED ISSUE #3353: sample world mir100.wbt sick_s300glass.dae cannot download
ubuntu 20.04, webots nightly build jul 5, sample world mir100:
`WARNING: Mir100 (PROTO) > SickS300 (PROTO) > Shape > Mesh: Cannot download https://raw.githubusercontent.com/cyberbotics/webots/R2021b/projects/devices/sick/protos/meshes/sick_s300_glass.dae: Error opening https://raw.githubusercontent.com/cyberbotics/webots/R2021b/projects/devices/sick/protos/meshes/sick_s300_glass.dae`

BenjaminDeleze:
**Description**
Fix #3353 

- [x] Fetch the tags beforehand to be sure to have them locally
- [x] Check if we are currently on a tag or not

### DIFF
--- a/scripts/packaging/Makefile
+++ b/scripts/packaging/Makefile
@@ -68,7 +68,8 @@ ISCC_FORMATTER = iscc_formatter.exe
 endif
 
 TARGET_URL = $(shell cat $(WEBOTS_HOME_PATH)/scripts/packaging/webots_version.txt)
-ifeq ($(shell git tag | grep $(TARGET_URL)),)
+TAGS = $(shell git fetch --all --tags)
+ifeq ($(shell git tag --points-at HEAD | grep $(TARGET_URL)),)
 TARGET_URL = $(shell cat $(WEBOTS_HOME_PATH)/resources/commit.txt)
 endif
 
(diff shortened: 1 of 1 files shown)


scripts/packaging/Makefile
```diff
@@ -68,7 +68,8 @@ ISCC_FORMATTER = iscc_formatter.exe
 endif
 
 TARGET_URL = $(shell cat $(WEBOTS_HOME_PATH)/scripts/packaging/webots_version.txt)
-ifeq ($(shell git tag | grep $(TARGET_URL)),)
+TAGS = $(shell git fetch --all --tags)
+ifeq ($(shell git tag --points-at HEAD | grep $(TARGET_URL)),)
```
omichel, line 72: I believe we have to use `:=` instead of `=` otherwise, since `TAGS` is never used, the `git fetch` will never be executed. See https://stackoverflow.com/a/6283363/810268
```suggestion
TAGS := $(shell git fetch --all --tags)
ifeq ($(shell git tag --points-at HEAD | grep $(TARGET_URL)),)
```
Also, in general for `$(shell ...)` commands, it's better to use `:=` (more efficient, less error prone).

omichel (changes requested):

omichel (approved):

PR #3360, opened 2021-07-06
Changed arm that respond to force

### LINKED ISSUE #3357: Wrong arm is responding to force in symmetry_matlab.wbt
**Describe the Bug**
In `symmetry_matlab.wbt`, its the left arm of the robot that respond to the applied force and the right one that mimics it while it is the opposite in `symmetry.wbt` as well as in the instructions

mael25:
Fixes #3357


omichel (approved):
Thank you.

PR #3362, opened 2021-07-07
Cleanup useless/broken jquery includes

omichel:
These jquery include files are not distributed any more, so including them from Webots was wrong.
Since they were not included in the distribution, removing them shouldn't break anything.


BenjaminDeleze (approved):
Thank you

PR #3365, opened 2021-07-07
Reset scroll when changing tab in Robot Window

### LINKED ISSUE #3358: Robot Window vertical scrollbar
**Describe the Bug**
If we scroll down in a tab within the Robot Window, and then go to another tab, the scrollbar position is not put back to the top. The main issue is that if the content to display in this second tab does not need to be extended, then blank content is added so that the scrollbar position is kept the same. Thus, the user has to scroll all the way up to see the graphs displayed.

**Steps to Reproduce**
1. Open `pioneer3dx.wbt` (or any other world where there is a lot of sensors in one of the tab of the Robot Window)
2. Open the `DistanceSensor` tab of the Robot WIndow
3. Scroll down
4. Switch to `PositionSensor` tab
5. Observe additionally added white space
6. If you scroll up to the top, then the vertical scrollbar disappear and everything looks good again.

**Expected behaviour**
This additional space is not added or the scrollbar position is automatically reset to the top.

mael25:
**Description**
Fixes #3358 
Reset the scroll position to zero when changing tab in Robot Window.

### DIFF
--- a/resources/projects/plugins/robot_windows/generic/window_manager.js
+++ b/resources/projects/plugins/robot_windows/generic/window_manager.js
@@ -37,6 +37,7 @@ function menuTabCallback(deviceType) {
   // Plots will be refreshed when the animation is over
   const canvas = new Canvas();
   canvas.clearCanvas();
+  window.scrollTo(0, 0);
 }
 
 function refreshSelectedTab() {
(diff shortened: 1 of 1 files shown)


omichel (approved):
Thank you.

PR #3367, opened 2021-07-07
Fixed restart of Webots on Windows

### LINKED ISSUE #3284: Webots doesn't restart when it is supposed to
On Windows, if you change the theme in the preferences, Webots will tell you that it needs to restart to apply the new theme. However, it will simply quit and will not restart automatically, the user has to manually restart it to see the new theme.

omichel:
Fixes #3284.


ad-daniel (approved):
Works now, thank you.

Issue #3292, opened 2021-07-02
Relocating the salamander example doesn't work


omichel:
This is because the skins folder is not copied.

ad-daniel:
> We should make the comments clearer about these two methods (copyExternalProject and copyProject) which are quite hard to understand what they do and more importantly why.

Also, it should be checked if the conclusions drawn https://github.com/cyberbotics/webots/pull/3368 are indeed valid and if not, add the copying of the skins folder also to `copyExternalProject` as it was done for `copyProject`.

stefaniapedrazzi:
Remaining issues addressed in #3398

PR #3368, opened 2021-07-07
Fix relocation of Skins

ad-daniel:
**Description**
Fixes https://github.com/cyberbotics/webots/issues/3292

Based on what I understood, `copyExternalProject` comes [into play](https://github.com/cyberbotics/webots/blob/8eeccb12a698624634b877da7d0d0c0c4e1df5f9/src/webots/editor/WbProjectRelocationDialog.cpp#L211-L212) only if `mExternalProjectPath` is defined, hence if [this](https://github.com/cyberbotics/webots/blob/8eeccb12a698624634b877da7d0d0c0c4e1df5f9/src/webots/editor/WbProjectRelocationDialog.cpp#L446-L454) point is reached. This is the case only if the skins folder would be inside the `protos` folder but according to `WbSkin`, the skins folder must be at the [root](https://github.com/cyberbotics/webots/blob/8eeccb12a698624634b877da7d0d0c0c4e1df5f9/src/webots/nodes/WbSkin.cpp#L405) (same level as worlds) hence it will never be the case.
As such, unless I misunderstood, it's sufficient to relocate this skins folder only in `copyProject` and not in `copyExternalProject`.

### DIFF
--- a/src/webots/editor/WbProjectRelocationDialog.cpp
+++ b/src/webots/editor/WbProjectRelocationDialog.cpp
@@ -312,6 +312,7 @@ int WbProjectRelocationDialog::copyProject() {
   //    controllers folder at the same level as the protos folder of the PROTO robot
   // 5) possibly the current controller file (even if corresponding to a PROTO robot)
   // 6) project libraries if a project controller or plugin is copied
+  // 7) copy skins folder if it exists
 
   const WbWorld *world = WbWorld::instance();
   const QString &worldFileBaseName = QFileInfo(world->fileName()).baseName();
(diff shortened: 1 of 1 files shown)


omichel (approved):
We should make the comments clearer about these two methods which are quite hard to understand what they do and why.

stefaniapedrazzi:
> Based on what I understood, copyExternalProject comes into play only if mExternalProjectPath is defined, hence if this point is reached. This is the case only if the skins folder would be inside the protos folder but according to WbSkin, the skins folder must be at the root (same level as worlds) hence it will never be the case.

Not really.
The `copyExternalProject` functions copies the resources used by PROTO models contained in the current world.
But these resources won't be located inside the PROTO `protos` folder but in the PROTO projects folder that corresponds to `protos/..`.

I will take care of updating the code and the comments.

PR #3371, opened 2021-07-07
Fixed deletion of robot window on reset (Linux and macOS)

omichel:
This fixes a wrong behavior of the robot window on Linux when resetting the `custom_robot_window_simple.wbt`: the series of numbers was wrong after the reset.
It also fixes the following warning occurring when resetting a simulation while a robot window is open on Linux and macOS:
```
Warning: Release of profile requested but WebEnginePage still not deleted. Expect troubles !
```


ad-daniel (approved):
Thank you.

PR #3373, opened 2021-07-07
Fix disabled proto test

### LINKED ISSUE #3243: Disabled test should be re-enabled
In #2787, the `tests/protos/modify_proto_template_field` was disabled because it was failing in the CI (but not on a local machine).
Given it was blocking the release and didn't harm the users, this was merged into the develop branch.
However, we should investigate why it failed and restore the test as soon as possible.

omichel:
Fixes #3243.
In addition to fixing #3243, this PR also displays the correct test name in the log. This will facilitate tests debugging in the future.


BenjaminDeleze (approved):
Thank you!

Issue #3312, opened 2021-07-02
Webots Guided Tour crash


mael25:
**Describe the Bug**
Webots crashes when we do the following steps:

**Steps to Reproduce**
1. Open the Guided Tour
2. Select one of the world
3. Cancel the loading
4. Select another world
5. Webots crashes

**System**

- Ubuntu 20.04 (Deb)

lukicdarkoo:
It crashes in wren:
```
libc.so.6!__GI_raise(int sig) (/build/glibc-YbNSs7/glibc-2.31/sysdeps/unix/sysv/linux/raise.c:50)
libc.so.6!__GI_abort() (/build/glibc-YbNSs7/glibc-2.31/stdlib/abort.c:79)
libc.so.6!__assert_fail_base(const char * fmt, const char * assertion, const char * file, unsigned int line, const char * function) (/build/glibc-YbNSs7/glibc-2.31/assert/assert.c:92)
libc.so.6!__GI___assert_fail(const char * assertion, const char * file, unsigned int line, const char * function) (/build/glibc-YbNSs7/glibc-2.31/assert/assert.c:101)
wren::ShaderProgram::prepareGl(wren::ShaderProgram * const this) (/home/lukic/webots/src/wren/ShaderProgram.cpp:216)
wren::ShaderProgram::setup(wren::ShaderProgram * const this) (/home/lukic/webots/src/wren/ShaderProgram.cpp:60)
wr_shader_program_setup(WrShaderProgram * program) (/home/lukic/webots/src/wren/ShaderProgram.cpp:334)
(anonymous namespace)::buildShader(WrShaderProgram * shader, const QFileInfo & vertexShader, const QFileInfo & fragmentShader) (/home/lukic/webots/src/webots/wren/WbWrenShaders.cpp:36)
WbWrenShaders::mergeSphericalShader() (/home/lukic/webots/src/webots/wren/WbWrenShaders.cpp:759)
WbWrenPostProcessingEffects::sphericalCameraMerge(float width, float height, int cameraCount, WrTextureInternalFormat textureFormat) (/home/lukic/webots/src/webots/wren/WbWrenPostProcessingEffects.cpp:142)
WbWrenCamera::setupSphericalPostProcessingEffect(WbWrenCamera * const this) (/home/lukic/webots/src/webots/wren/WbWrenCamera.cpp:840)
[...]

PR #3374, opened 2021-07-08
Fix incorrect cleaning of WbLights

ad-daniel:
**Description**
This issue can make the guided tour crash (https://github.com/cyberbotics/webots/issues/3312) but only on niche situations

The only way I was able to reproduce the crash was:
- open guided tour
- switch very quickly from `mantis.wbt` to `e-puck_line_demo.wbt` worlds while attempting to cancel the loading of either
- at some point, it crashes

Although each of these worlds only has one light source, and therefore the global class variable `cLights` should at most be of size 1, doing this process very quickly resulted in having bigger  `cLights` (size 2, 3, 5, ...). This is because this variable is appended in the `preFinalize` but only cleared if wren was initialized. In practice however this variable should be cleared regardless, if not, like in this situation what can happen is that pointers of previous worlds persist due to it being canceled before being finalized, and when we access this class variable the pointers are no longer valid resulting in the crash.


ad-daniel:
Other crashes related to the Guided Tour persist, although this issue also did, it wasn't the only source. As this is an issue of its own that goes beyond the guided tour, I'll create another PR for the other crashes.

omichel (approved):
Well spotted!

PR #3376, opened 2021-07-08
Fix Guided Tour world canceling and switch causing crash

ad-daniel:
**Description**
Fixes https://github.com/cyberbotics/webots/issues/3312.

Shouldn't be allowed to select a new world in the Guided Tour if still in the process of canceling the previous. We can't allow it because [WbApplication::loadWorld](https://github.com/cyberbotics/webots/blob/c4508c0c2c9b8d11cdb5d96c1fe093d812e6aa0b/src/webots/app/WbApplication.cpp#L258-L260) doesn't ensure the world was actually canceled when it receives the signal, it directly sets the flags to false. This is dealt with for normal worlds, but not from worlds selected through the Guided Tour.


mael25 (approved):
Thank you!

PR #3378, opened 2021-07-08
Improved test suite scripts

omichel:
This PR:
- improves the error reporting of the test suite.
- fixes PEP8 issues in python script
- disables the parser tests on Windows which were never working because the stderr and stdout redirection doesn't work.


ad-daniel (approved):
Thank you.

PR #3380, opened 2021-07-08
fix repo.txt when ssh

BenjaminDeleze:
**Description**
Correct the script that get the name of the repo when it is cloned using ssh


ad-daniel (approved):
It's working now, thank you.

Issue #3220, opened 2021-06-23
Lidar crash


omichel:
Some Lidar configuration cause Webots to crash (tested on the develop branch).
Open the [projects/samples/devices/worlds/lidar.wbt](https://github.com/cyberbotics/webots/blob/develop/projects/samples/devices/worlds/lidar.wbt) in Webots in pause mode and set the `Lidar.fieldOfView` to `1`. Save and revert &rarr; crash. If you attempt to open this world, it will crash when running it as well.

lukicdarkoo:
I cannot reproduce this issue, I tried even to be more extreme with the `Lidar.fieldOfView`, but it doesn't crash

omichel:
It definitely crashes on my Windows machine...

llessieux:
I cannot reproduce the crash but when I set the Lidar.fieldOfView to 1, I am losing all the rays. Not sure if that is expected or not ;)
I am using Windows too (11).

omichel:
Now, I cannot reproduce the crash from the master branch either (on Windows 10).

> but when I set the Lidar.fieldOfView to 1, I am losing all the rays. Not sure if that is expected or not ;)

This is not expected and I cannot produce this problem: I start the simulation and the value of the Lidar.fieldOfView is 1.57, I can run the simulation, set it to 1, and I still see the rays. Or I can stop the simulation, reload it, set the Lidar.fieldOfView to 1, save, revert and run and I can still see the rays. Under which circumstances do you see the rays disappearing?

llessieux:
Strange, I updated the source, rebuilt it and I can't reproduce that anymore.
Not sure what happened.

omichel:
I am closing this issue as the crash is not reproducible any more.

PR #3385, opened 2021-07-09
Fix memory problem with Lidar point cloud

### LINKED ISSUE #3384: Memory problem with Lidar point cloud
**Describe the Bug**
When using a `Lidar` with multiple layers, a few points are sometimes completely out of place, but with a length similar to the correct points.

These points in reality are overwritten by the depth image outputted by the camera, as most probably the memory space allocated to it is not big enough.

This causes some points of the first layers of the point cloud to alternate between a correct position and an out of place position when they are overwritten by the camera.

**Steps to Reproduce**
1. Open `sick_point_cloud.wbt`
2. See error

**Screenshots**

mael25:
**Description**
Fixes #3384

**Linked Pull Request**
#3032


omichel:
Probably related to #3220 as well.

src/webots/nodes/WbLidar.hpp
```diff
@@ -137,7 +137,7 @@ class WbLidar : public WbAbstractCamera {
   bool isRotating() const { return mActualType.startsWith('r', Qt::CaseInsensitive); }
   double verticalFieldOfView() const { return actualFieldOfView() * ((double)height() / (double)width()); }
 
-  WbLidarPoint *pointArray() { return (WbLidarPoint *)(lidarImage() + actualHorizontalResolution() * actualNumberOfLayers()); }
+  WbLidarPoint *pointArray() { return (WbLidarPoint *)(lidarImage() + width() * height()); }
```
omichel, line 140: That seems to fix #3220.

mael25: It seems that the handling of the image is currently done like this:

- The `AbstractCamera` writes the data (`width` x `height`) in the shared memory (corresponding pointer is returned by`lidarImage()` in `WbLidar.cpp`)
- We copy it in a temporary image (`mTemporaryImage`, also `width` x `height`)
- We overwrite the shared memory by putting this time only the layers needed (`actualHorizontalResolution` x `actualNumberOfLayers`) for the point cloud.
- We compute the coordinates for the WbLidarPoint of the point cloud and save them in the shared memory just after the image --> so at `lidarImage() + width * height`

mael25: The error when it was rotating is that the width of the image is not returned by `width() `but by `actualHorizontalResolution()`.

omichel (approved):
Thank you!

src/webots/nodes/WbLidar.hpp
```diff
@@ -131,13 +131,14 @@ class WbLidar : public WbAbstractCamera {
   void initializeImageSharedMemory() override;
 
   int size() const override {
-    return sizeof(float) * width() * height() + sizeof(WbLidarPoint) * actualHorizontalResolution() * actualNumberOfLayers();
+    return sizeof(float) * actualHorizontalResolution() * height() +
```
stefaniapedrazzi, line 134: This structure should contain first the lidar image followed by the point cloud information.
Are you sure that this new size computation really matches the content of the shared memory?

Because the previous formula seems more correct.

stefaniapedrazzi: Isn't that the WbRenderingDevice::width() value should be updated and it is not?

mael25: In fact the naming is a little bit misleading as `width()` is computed like this:
```
int WbLidar::width() const {
  if (isRotating())
    return ceil(actualHorizontalResolution() * (actualFieldOfView() / (2.0 * M_PI)));
  return actualHorizontalResolution();
}
```
So, in the case of a rotating lidar, it doesn't correspond anymore to the size of the lidar image, but to a smaller window.
Also, this line was recently changed in [#3032](https://github.com/cyberbotics/webots/commit/aaea21d899a03933eb1fc06d21754bbde805334e#diff-a8d078376f6e715ea6a984225234ddbbe25a58c20fdba74adad0ca623f3e4013L133) but it used to be `actualHorizontalResolution()` before.

src/controller/c/lidar.c
```diff
@@ -75,6 +76,7 @@ static void wb_lidar_new(WbDevice *d, unsigned int id, int w, int h, double fov,
 
   l = malloc(sizeof(Lidar));
   l->max_range = max_range;
+  l->height = h;
```
stefaniapedrazzi, line 79: This information is now duplicated.
It could be retrieved directly from the `AbstractCamera` struct (that contains the Lidar struct).

src/webots/nodes/WbLidar.hpp
stefaniapedrazzi: Ok. Thank you for the explanation.
I think that this class needs a deeper refactor because we should avoid misleading attribute names we should improve them.

src/controller/c/lidar.c
```diff
@@ -455,7 +458,7 @@ const WbLidarPoint *wb_lidar_get_point_cloud(WbDeviceTag tag) {
   const float *image = wb_lidar_get_range_image(tag);
   if (image == NULL)
     return NULL;
-  return (WbLidarPoint *)(image + l->number_of_layers * l->horizontal_resolution);
+  return (WbLidarPoint *)(image + l->height * l->horizontal_resolution);
```
stefaniapedrazzi, line 461: `l->height` corresponds to:
```
AbstractCamera *ac = lidar_get_abstract_camera_struct(tag);
ac->height
```

src/webots/nodes/WbLidar.hpp
mael25: However, I just saw that in case of rotating lidars, the initial image is not stored in the shared memory but only put in `mTemporaryImage`, thus we might be able to define the size to `sizeof(float) * actualHorizontalResolution() * actualNumberOfLayers()`, plus the space for the point cloud.

mael25:
I have updated the code of `WbLidar::writeAnswer()` a little bit so that when it calls `WbAbstractCamera::writeAnswer()`, the content of the `WbWrenCamera` is not put in the shared memory. This is done by setting `mImageChanged` to false before calling `WbAbstractCamera::writeAnswer()`.

Then, we still call `copyAllLayersToSharedMemory() `which will copy the content of the `WbWrenCamera` in a temporary image (`mTemporaryImage`), before putting only the needed layers in the shared memory.

This way, we avoid the first useless copy of the `WbWrenCamera`  content in the shared memory, which allow us to reduce the size of the shared memory as it will get only the layers we are interested in.

On the long term, we will try to get rid of the temporary image and take directly the needed layers from the buffer.

src/webots/nodes/WbLidar.cpp
```diff
@@ -235,13 +235,15 @@ void WbLidar::addConfigureToStream(QDataStream &stream, bool reconfigure) {
...
-    copyAllLayersToSharedMemory();
+  if (mImageChanged) {
+    mImageChanged = false;  // prevents the AbstractCamera from copying the whole content of the camera in the shared memory
+    WbAbstractCamera::writeAnswer(stream);
+    mSensor->resetPendingValue();
+    if (!isRotating() && mSensor->isEnabled())  // in case of rotating lidar, the copy is done during the step
+      copyAllLayersToSharedMemory();            // for non-rotating lidar, copy the layers needed in the shared memory
+  } else {
+    WbAbstractCamera::writeAnswer(stream);
+  }
```
omichel, line 246:
```suggestion
  } else
    WbAbstractCamera::writeAnswer(stream);
```

omichel (changes requested):

omichel (approved):

PR #3386, opened 2021-07-09
Fix scheduled actions

ad-daniel:
**Description**
Scheduled actions have been [failing](https://github.com/cyberbotics/webots/actions/workflows/test_suite_linux_develop.yml) due to the script not being included in the artifact.


omichel:
I don't believe it's the right way to fix this problem.
The `scripts/packaging` folder should not be included in the distribution as it is meant for creating the distribution.
Instead I believe we should fix the problem simply by adding the `update_urls.py` to the develop branch.

ad-daniel:
Since the release should occur soon, and that master will be merged into develop after that, it's bound to fail again only tonight (in theory) and then no longer appear, so not worth pushing this script to develop.

PR #3387, opened 2021-07-09
Improve python example

### LINKED ISSUE #3011: Setting PYTHONPATH in runtime.ini under Linux and macOS is ignored, overwritten
In the following runtime.ini:
```ini
[environment variables with paths]
PYTHONPATH=$(WEBOTS_HOME)/projects/samples/contest/robocup/controllers/referee
```
The `PYTHONPATH` environment variable specified here is ignored because it is overwritten by Webots with something like `$(WEBOTS_HOME)/lib/controller/python38`.

This bug affects only the Linux and macOS versions of Webots. On Windows, it works as expected.

It is affecting https://github.com/RoboCup-Humanoid-TC/webots/pull/81.

omichel:
Fixes #3011.
Actually, I cannot reproduce #3011 on Linux any more. Everything works as expected.
I believe the original issue description was wrong, mentioning the deprecated `[environment variable with relative paths]` instead of `[environment variables with paths]`.
I modified the python example to illustrate how to use `PYTHONPATH` in a `runtime.ini` file which is a common problem that Python controller developers are facing.

### DIFF
--- a/projects/languages/python/controllers/driver/common.py
+++ b/projects/languages/python/controllers/driver/common.py
@@ -12,5 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
 def common_print(caller):
     print('This module is common to both driver and slave controllers (called from ' + caller + ').')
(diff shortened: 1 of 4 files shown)


ad-daniel (approved):
Works both on linux and mac, thank you.

omichel:
Thank you for the review and tests.

PR #3388, opened 2021-07-09
Disable building of packages on develop

BenjaminDeleze:
**Description**
As develop and master have the same version name (R2021b) and we do not plan to add modification on develop until the end of FLU conversion, we can turn off the building of packages on develop.

Moreover having master and develop with the same version name is a bad idea as it causes conflict between the packages build on develop and those build on master during the upload to the release.

### DIFF
--- a/.github/workflows/test_suite_linux_develop.yml
+++ b/.github/workflows/test_suite_linux_develop.yml
@@ -2,7 +2,7 @@ name: Test Suite Linux (develop)
 
 on:
   schedule:
-    - cron:  '0 23 * * *'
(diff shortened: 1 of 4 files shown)


omichel (approved):

stefaniapedrazzi:
> Moreover having master and develop with the same version name is a bad idea as it causes conflict between the packages build on develop and those build on master during the upload to the release.

I think that this is the wrong approach.
I would have rather updated the version on the develop branch (we need to do it soon in any case) instead of this change that we will need to revert in one week or so.

PR #3390, opened 2021-07-09
Fixed P-Rob3 motor names

### LINKED ISSUE #3258: P-Rob3 device names
In `P-Rob3` PROTO models, the device names are not very self-explanatory but they use names likes  `1`, `2`, etc. (for joints), `2 sensor`, `2 sensor` (position sensors), etc.
Then for buttons the style changes and the number comes after: `button 1`, `button 2` (same for distance sensors)

Is there a reason for this choice?
Especially using numbers only like for the joints doesn't seem user-friendly and when seeing it in the web robot component I think the first impression is that there are some bugs.
The it would be nice to have some coherence in the device name format at least inside the same model.

omichel:
Fixes #3258.


ad-daniel (approved):
Thank you.

PR #3391, opened 2021-07-09
Fix bug sumo

BenjaminDeleze:
**Description**
Webots was crashing when trying to play the world with SUMO (like `highway overtake`).

The cause was that some `QNetworkReplys` where deleted locally but there entry was not removed/set to null in the intermediary cache.

### DIFF
--- a/src/webots/nodes/utils/WbDownloader.cpp
+++ b/src/webots/nodes/utils/WbDownloader.cpp
@@ -73,9 +73,9 @@ void WbDownloader::download(const QUrl &url) {
     if (!(mOffline == true && mCopy == false)) {
       mCopy = true;
       QNetworkReply *reply = gUrlCache[mUrl];
-      if (reply && !reply->isFinished()) {
+      if (reply && !reply->isFinished())
         connect(reply, &QNetworkReply::finished, this, &WbDownloader::finished, Qt::UniqueConnection);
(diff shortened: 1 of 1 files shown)


src/webots/nodes/utils/WbDownloader.cpp
```diff
@@ -70,11 +73,8 @@ void WbDownloader::download(const QUrl &url) {
...
       mCopy = true;
       QNetworkReply *reply = gUrlCache[mUrl];
-      if (reply) {
-        if (reply->isFinished())
-          finished();
-        else
-          connect(reply, &QNetworkReply::finished, this, &WbDownloader::finished, Qt::UniqueConnection);
+      if (reply && !reply->isFinished()) {
+        connect(reply, &QNetworkReply::finished, this, &WbDownloader::finished, Qt::UniqueConnection);
       } else
```
omichel, line 78: Coding style:
```suggestion
      if (reply && !reply->isFinished())
        connect(reply, &QNetworkReply::finished, this, &WbDownloader::finished, Qt::UniqueConnection);
      else
```

omichel (changes requested):
It looks good to me.
I only have a change request about the coding style.

omichel (approved):
Thank you.

PR #3393, opened 2021-07-12
Fix prob3 test

BenjaminDeleze:
**Description**
The P-Rob3 proto was changed, but not the corresponding reference file in the test suite, so the test suite was failing.


omichel (approved):
Thank you!

omichel:
The test failure on Windows is due to an update of MSYS2 which breaks the webots binary. I am investigating it now. Meanwhile, we can safely merge this PR.

Issue #3392, opened 2021-07-09
Incorrect resetting of BallJoint nodes


ad-daniel:
**Describe the Bug**
As reported on discord by FunkyPants, the resetting of `BallJoint` nodes appears to be incorrect:

To reproduce:
1. open [test.zip](https://github.com/cyberbotics/webots/files/6792383/test.zip) but don't run it.
2. keep resetting, the legs go everywhere at each reset

The model is fairly complex, so there might be more at play, but the simplest MRE I managed to find that still resets incorrectly is 
[ball_joint_bug.zip](https://github.com/cyberbotics/webots/files/6792389/ball_joint_bug.zip).

1. open world
2. reset. Since the simulation wasn't run, it shouldn't move but it does.

PR #3394, opened 2021-07-12
Fix `BallJoint` reset

ad-daniel:
**Description**
Fixes https://github.com/cyberbotics/webots/issues/3392.

### DIFF
--- a/docs/reference/changelog-r2021.md
+++ b/docs/reference/changelog-r2021.md
@@ -116,7 +116,7 @@ Released on July, 5th, 2021.
       - Fixed incorrect position retrieval immediately after resetting nested [Transform](transform.md) nodes ([#3051](https://github.com/cyberbotics/webots/issues/3051)).
       - Fixed reset of [Charger](charger.md) energy when the recharging [Robot](robot.md) battery is full ([#2879](https://github.com/cyberbotics/webots/pull/2879)).
       - Fixed [Camera](camera.md) image update in controllers after simulation reset ([#2725](https://github.com/cyberbotics/webots/pull/2725)).
-      - Fixed reset of simulations including [BallJoint](balljoint.md) nodes like the [Stewart Platform](https://github.com/cyberbotics/webots/blob/master/projects/samples/demos/worlds/stewart_platform.wbt) ([#2593](https://github.com/cyberbotics/webots/pull/2593)).
+      - Fixed reset of simulations including [BallJoint](balljoint.md) nodes like the [Stewart Platform](https://github.com/cyberbotics/webots/blob/master/projects/samples/demos/worlds/stewart_platform.wbt) ([#2593](https://github.com/cyberbotics/webots/pull/2593)) and ([#3394](https://github.com/cyberbotics/webots/pull/3394)).
       - Fixed step button status if simulation is reset from UI when the step button is disabled ([#2741](https://github.com/cyberbotics/webots/pull/2741)).
     - Fixed bug in the C++, Python and Java API where the [Robot.getDevice()](robot.md#wb_robot_get_device) methods were returning different objects when passing the same string argument for the device name ([#2957](https://github.com/cyberbotics/webots/pull/2957)).
     - Fixed the `robot_get_urdf` function to include leaf nodes in URDF ([#2803](https://github.com/cyberbotics/webots/pull/2803)).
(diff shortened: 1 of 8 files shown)


docs/reference/changelog-r2021.md
```diff
@@ -146,6 +146,7 @@ Released on July, 5th, 2021.
     - Fixed incorrect resetting of [PointLight](pointlight.md) nodes ([#3345](https://github.com/cyberbotics/webots/pull/3345)).
     - Fixed incorrect cleaning of [Light](light.md) node ([3374](https://github.com/cyberbotics/webots/pull/3374)).
     - Fixed crash provoked by canceling and switching world in the Guided Tour ([#3376](https://github.com/cyberbotics/webots/pull/3376)).
+    - Fixed incorrect reset of [BallJoint](balljoint.md) nodes ([#3394](https://github.com/cyberbotics/webots/pull/3394)).
```
stefaniapedrazzi, line 149: Note that there is already an entry about fixing reset of BallJoint (line 119).
Could please just add the link to the PR to the already existing entry instead of adding a new entry with almost duplicated text?

ad-daniel:
Although the fix is reasonable and indeed fixes the problem with the reset (both the MRE and the original word provided by the user), it breaks the `ball_joint_vs_hinge_joint.wbt` test. Specifically, the kinematic `BallJoint` doesn't behave like its dynamic counterpart anymore (more precisely, when motor2 is actuated it deviates).

This appears to have been an issue present since the implementation of the `BallJoint` [itself](https://github.com/cyberbotics/webots/commit/b2b37133799671903f8a70bdff0937a83bc64560), where this disparity between kinematic and dynamic was precisely fixed this way and in doing so broke the reset. The reason it never surfaced might be because the problem with the reset is only visible in specific conditions, for instance on `motor3.wbt` we don't notice it because of the symmetry.

Either way, the behavior shown in the related issue appears quite worse than the kinematic one, so I'm tempted to believe this approach is more correct and that it's the kinematic implementation that was and is wrong.

stefaniapedrazzi:
I think that if we are not able to fully fix the issue but we will introduce other problems, for the moment we should try to keep the same behavior as in the previous released version.
Then, we will have the time to work on a fully working solution for the next revision version.

But for users (all of them and not a specific one) it is worse if we break something that was working before.

src/webots/nodes/WbBallJoint.cpp
```diff
@@ -256,7 +256,6 @@ void WbBallJoint::updatePositions(double position, double position2, double posi
 }
 
 void WbBallJoint::updatePosition(double position) {
```
stefaniapedrazzi, line 258: It seems that there is no virtual base method for this function, if so then the override attribute could be removed.
Then, do we really need it for BallJoint and Hinge2Joint) or could it be simply deleted?

ad-daniel: It isn't really needed by the BallJoint and Hinge2 classes but they do serve a purpose in the base ones. They could be removed for the former two but it might be risky as well, all it takes is missing a "s" and inadvertently we call the parent method and might not be evident to debug. Wouldn't it be better instead to leave them and put an `assert(false)`?

omichel (approved):
It looks good to me.
I would suggest that we merge it now so that we can create the release candidate packages tonight and test tomorrow. If @stefaniapedrazzi has anything to add here, we should re-open a new PR tomorrow.

ad-daniel:
Actually, it turns out they can't even be removed. For instance, if we comment it out in the Hinge2Joint class, the test `move_hinge_2_joint_with_suspension` not longer works. It appears that when a supervisor sets the translation of the robot parent, this function ends up being called. It could be possible to track it down and call `updatePositionS` instead if it's a node of type `BallJoint` or `Hinge2`, but I'm not that convinced the result would be any cleaner.

tests/physics/controllers/ball_joint_reset/ball_joint_reset.c
```diff
@@ -0,0 +1,45 @@
+#include <webots/robot.h>
+#include <webots/supervisor.h>
+
+#include "../../../lib/ts_assertion.h"
+#include "../../../lib/ts_utils.h"
+
+int main(int argc, char **argv) {
+  ts_setup(argv[0]);
+
+  int time_step = wb_robot_get_basic_time_step();
```
omichel, line 10:
```suggestion
  const int time_step = wb_robot_get_basic_time_step();
```

omichel (changes requested):

omichel (approved):

PR #3397, opened 2021-07-12
Updated to version R2021c on develop

omichel:
Updated version on the develop branch, so that we can revert #3388.
This was done by running `scripts/new_version/new_version.sh`,


stefaniapedrazzi (dismissed):
Thank you.
The committed changes are all fine.

But the script returns a couple of replacement failures that should be fixed as well (here on or in another PR):
* update the `new_version.sh` script and remove the update of the `docs/dependencies.txt` that no longer depends on the version
* header of `tests/api/controllers/supervisor_import_robot_node/MyBot.wbo` is wrong and cannot be correctly updated to R2021c:
    ``#VRML_SIM R2021b utf8`` instead of ``#VRML_OBJ R2021b``

The other replacement failures messages are "correct":
* `projects/samples/tutorials/protos/FourWheelsRobot.proto` (doesn't have the header line)
* `projects/robots/mobsya/thymio/controllers/thymio2_aseba/aseba/aseba/clients/studio/plugins/ThymioVPL/UsageProfile.proto` (is not a Webots PROTO file)

stefaniapedrazzi (approved):
Thank you!

PR #3399, opened 2021-07-12
Reverts #3388

omichel:
Since we are increasing the version number on develop in #3397, we can now revert #3388 once #3397 is merged.


stefaniapedrazzi (approved):
Thank you

PR #3401, opened 2021-07-13
Enhancement lidar memory

mael25:
**Description**
The goal is to put the lidar layer pixels in the shared memory directly from the GPU buffer, without storing them before in a temporary image.

### DIFF
--- a/include/wren/frame_buffer.h
+++ b/include/wren/frame_buffer.h
@@ -30,7 +30,7 @@ void wr_frame_buffer_blit_to_screen(WrFrameBuffer *frame_buffer);
 
 /* The data written to 'data' will be in the format of output texture number 'index' */
 void wr_frame_buffer_copy_contents(WrFrameBuffer *frame_buffer, int index, void *data);
-void wr_frame_buffer_copy_row(WrFrameBuffer *frame_buffer, int index, int line, void *data);
+void wr_frame_buffer_copy_contents_part(WrFrameBuffer *frame_buffer, int index, void *data, int src, int size);
 void wr_frame_buffer_copy_pixel(WrFrameBuffer *frame_buffer, int index, int x, int y, void *data, bool flip_y);
 void wr_frame_buffer_copy_depth_pixel(WrFrameBuffer *frame_buffer, int x, int y, void *data, bool flip_y);
 
--- a/include/wren/frame_buffer.h
+++ b/include/wren/frame_buffer.h
@@ -29,7 +29,7 @@ void wr_frame_buffer_setup(WrFrameBuffer *frame_buffer);
 void wr_frame_buffer_blit_to_screen(WrFrameBuffer *frame_buffer);
 
 /* The data written to 'data' will be in the format of output texture number 'index' */
-void wr_frame_buffer_copy_contents(WrFrameBuffer *frame_buffer, int index, void *data, int offset = 0, int size = 0);
+void wr_frame_buffer_copy_contents(WrFrameBuffer *frame_buffer, int index, void *data, int offset, int size);
 void wr_frame_buffer_copy_pixel(WrFrameBuffer *frame_buffer, int index, int x, int y, void *data, bool flip_y);
 void wr_frame_buffer_copy_depth_pixel(WrFrameBuffer *frame_buffer, int x, int y, void *data, bool flip_y);
 
--- a/include/wren/frame_buffer.h
+++ b/include/wren/frame_buffer.h
@@ -30,6 +30,7 @@ void wr_frame_buffer_blit_to_screen(WrFrameBuffer *frame_buffer);
 
 /* The data written to 'data' will be in the format of output texture number 'index' */
 void wr_frame_buffer_copy_contents(WrFrameBuffer *frame_buffer, int index, void *data);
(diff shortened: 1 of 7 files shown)


mael25:
If the remianing tests pass, we might introduce this PR in the release.

omichel:
Did you observed any performance improvement?

src/wren/FrameBuffer.cpp
```diff
@@ -179,6 +179,27 @@ namespace wren {
...
+    const unsigned int currentPixelPackBuffer = glstate::boundPixelPackBuffer();
+    glstate::bindPixelPackBuffer(mOutputDrawBuffers[index].mGlNamePbo);
+
+    const Texture::GlFormatParams &params = drawBufferFormat(index);
+    const int sizeInBytes = params.mPixelSize * size;
+    const int srcInBytes = params.mPixelSize * src;
+
+    void *start = glMapBufferRange(GL_PIXEL_PACK_BUFFER, srcInBytes, sizeInBytes, GL_MAP_READ_BIT);
+    assert(start);
+    memcpy(data, start, sizeInBytes);
```
omichel, line 196: `memcpy` of images are evil (performance disaster). We should avoid them. I believe we could use `glGetBufferSubData` instead of `glMapBufferRange` to avoid this `memcpy`.

include/wren/frame_buffer.h
```diff
@@ -29,7 +29,7 @@ void wr_frame_buffer_setup(WrFrameBuffer *frame_buffer);
 void wr_frame_buffer_blit_to_screen(WrFrameBuffer *frame_buffer);
 
 /* The data written to 'data' will be in the format of output texture number 'index' */
-void wr_frame_buffer_copy_contents(WrFrameBuffer *frame_buffer, int index, void *data);
+void wr_frame_buffer_copy_contents(WrFrameBuffer *frame_buffer, int index, void *data, int offset = 0, int size = 0);
```
omichel, line 32: I am surprised this works. Normally, in pure C, we cannot define default parameters values for functions...
```suggestion
void wr_frame_buffer_copy_contents(WrFrameBuffer *frame_buffer, int index, void *data, int offset, int size);
```

src/wren/FrameBuffer.cpp
```diff
@@ -159,22 +159,20 @@ namespace wren {
     glstate::bindReadFrameBuffer(currentReadFrameBuffer);
   }
 
-  void FrameBuffer::copyContents(size_t index, void *data) {
+  void FrameBuffer::copyContents(size_t index, void *data, int offset, int size) {
     assert(index < mOutputDrawBuffers.size());
     assert(mOutputDrawBuffers[index].mGlNamePbo);
+    assert(src + size < mWidth * mHeight);
```
omichel, line 165: I am surprised this compiled...
```suggestion
    assert(offset + size < mWidth * mHeight);
```

omichel (changes requested):

src/wren/FrameBuffer.cpp
mael25: It seems that even putting assert(0) doesn't prevent the program to continue..

mael25:
Performances on lidar.wbt, with the following fields changed for the Lidar:

- horizontalResolution 2000
- verticalFieldOfview 1.5
- numberOfLayers 20


Results
| Branch | Avg speed factor | device:lidar(ms) | controller:lidar(ms)
| --- | --- | --- | --- |
| master | 1.58073 | 0.494 | 0.091
| this PR | 1.59912 | 0.493 | 0.089

So the performances seem to be similar...

mael25:
Measuring the time (real time, using `std::chrono::steady_clock::now()`) elapsed in `copyAllLayersToSharedMemory` (which includes the call to update the point cloud), averaged over 10'000 steps we get:

| Branch | **Mode fast, no rendering (ms)** | Normal (ms)
| --- | --- | --- 
| master | **3.635** | 4.472
| this PR | **4.373** | 4.536

So it seems it would be better to keep it as it is right now.

mael25:
When doing all the `glGetBufferSubData` calls directly from `FrameBuffer.cpp`, the performances do not get better: 4.386ms (fast mode, no rendering).

I've started another solution which would be to extract the layers in a second GPU buffer, so that we can then put in one call all the layers in the CPU.

mael25:
I have tested the performance change when using `glGetBufferSubData` instead of `memcpy` in the `copyContents` function of `FrameBuffer.cpp` using  `camera_motion_blur.wbt` with different cameras resolution.

Time elapsed in `writeAnswer` of `WbCamera.cpp` (which handles the copy of the image):

| Branch | Image 80x60 | Image 800x600 | Image 3000x2000
| --- | --- | --- | ---
| memcpy | 40us | 2.44ms | 29.36ms
| glGetBufferSubData | 42us | 2.46ms | 29.37ms

So it seems there is no improvement with this change...

mael25:
Performances in  `camera_motion_blur.wbt` (1000 steps, 3000x2000 camera resolution):

Elapsed time in:

`glMapBufferRange` / `memcpy`  / `glUnmapBuffer`: **29.35ms**
- glMapBufferRange: **26.94ms**
- memcpy: **2.38ms**

`glGetBufferSubData`: **29.347ms**

omichel:
Can we close this PR, or is it still relevant?

PR #3402, opened 2021-07-13
Fix mac pythonpath

### LINKED ISSUE #3285: Webots doesn't find the Python executable when running from Spotlight
The following works as expected:
```
% python3.8 --version
Python 3.8.5
% /Applications/Webots.app/webots /Applications/Webots.app/projects/languages/python/worlds/example.wbt
```

But when running the same world in Webots executed from the GUI Webots fails to find the Python executable.

**System**
 - Operating System: macOS

BenjaminDeleze:
**Description**
The python executable is not found on MacOS when launching Webots from the GUI (Spotlight, Finder, Dock). 

To fix that, we decided to hardcode the standard MacOS path to both python from python.org and python from homebrew in Webots.
It it does not work, a warning will be displayed, requesting the user to set the full path to his python executable in the preferences of webots.

Fix #3285.

### DIFF
--- a/src/webots/control/WbLanguageTools.cpp
+++ b/src/webots/control/WbLanguageTools.cpp
@@ -141,6 +141,7 @@ QString WbLanguageTools::pythonCommand(QString &shortVersion, const QString &com
   return pythonCommand;
 }
 
+#if defined __APPLE__ || defined __linux__
 const QString WbLanguageTools::checkIfPythonCommandExist(const QString &pythonCommand, QProcessEnvironment &env, bool log) {
   QString shortVersion;
   QProcess process;
@@ -159,6 +160,7 @@ const QString WbLanguageTools::checkIfPythonCommandExist(const QString &pythonCo
     shortVersion = QString(version[0][0]) + version[0][2];
   return shortVersion;
 }
+#endif
 
 #ifdef __APPLE__
 QString WbLanguageTools::findWorkingPythonPath(const QString &pythonVersion, QProcessEnvironment &env, bool log) {
--- a/src/webots/control/WbLanguageTools.cpp
+++ b/src/webots/control/WbLanguageTools.cpp
@@ -64,6 +64,10 @@ const QStringList WbLanguageTools::javaArguments() {
 QString WbLanguageTools::pythonCommand(QString &shortVersion, const QString &command, QProcessEnvironment &env) {
   QString pythonCommand = command;
   QString advice =
+#ifdef __APPLE__
+    "To fix the problem, you could try to insert the full path of your python distribution in "
+    "Webots->preferences->python command.\n";
+#else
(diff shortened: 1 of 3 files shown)


src/webots/control/WbLanguageTools.cpp
```diff
@@ -63,7 +63,11 @@ const QStringList WbLanguageTools::javaArguments() {
 
 QString WbLanguageTools::pythonCommand(QString &shortVersion, const QString &command, QProcessEnvironment &env) {
   QString pythonCommand = command;
-  const QString advice =
+  QString advice =
+#ifdef __APPLE__
+    "To fix the problem, you could try to insert the full path of your python distribution in "
```
omichel, line 68:
```suggestion
    "To fix the problem, you should set the full path of your python command in "
```

src/webots/control/WbLanguageTools.cpp
```diff
@@ -63,7 +63,11 @@ const QStringList WbLanguageTools::javaArguments() {
 
 QString WbLanguageTools::pythonCommand(QString &shortVersion, const QString &command, QProcessEnvironment &env) {
   QString pythonCommand = command;
-  const QString advice =
+  QString advice =
```
omichel, line 66: Can't we keep it const?
```suggestion
  const QString advice =
```

src/webots/control/WbLanguageTools.hpp
```diff
@@ -35,6 +35,9 @@ class WbLanguageTools {
 private:
   WbLanguageTools() {}
   ~WbLanguageTools() {}
+  static const QString checkIfPythonCommandExist(const QString &pythonCommand, QProcessEnvironment &env, bool log);
+  // cppcheck-suppress unusedPrivateFunction
+  static QString findWorkingPath(const QString &pythonVersion, QProcessEnvironment &env, bool log);
```
omichel, line 40: That method seems to be specific to python, so I would add "python" in its name:
```suggestion
  static QString findWorkingPythonPath(const QString &pythonVersion, QProcessEnvironment &env, bool log);
```

src/webots/control/WbLanguageTools.hpp
```diff
@@ -35,6 +35,9 @@ class WbLanguageTools {
 private:
   WbLanguageTools() {}
   ~WbLanguageTools() {}
+  static const QString checkIfPythonCommandExist(const QString &pythonCommand, QProcessEnvironment &env, bool log);
+  // cppcheck-suppress unusedPrivateFunction
```
omichel, line 39: Instead of `cppcheck-suppress unusedPrivateFunction`, we should declare and implement this function only in the macOS case (using `#ifdef`).

omichel (changes requested):
That looks good to me.
I have a couple of minor suggestions.

omichel (approved):
Thank you.

Issue #3406, opened 2021-07-14
HighwaySigns are broken


ad-daniel:
**Describe the Bug**
Highway signs appear to be broken, it doesn't look like it's a regression though.

![highway_overtake](https://user-images.githubusercontent.com/44834743/125583574-196e6dee-e179-4c91-abb1-a6a9999708c8.png)

EDIT: indeed, they were broken in 2021a already

PR #3407, opened 2021-07-14
Fix HighwaySign proto

ad-daniel:
**Description**
Fixes https://github.com/cyberbotics/webots/issues/3406.

highway_overtake_1

I've added it to the changelog since it was already broken in 2021a.


omichel (approved):
Thank you.

Issue #3400, opened 2021-07-13
Simulation Server leaking resources


lfelipesv:
I am testing the simulation server script and it seems the runWebotsinThread() function (https://github.com/cyberbotics/webots/blob/2f7f45f430746978ecc3dc399fc7dd5d4df362a2/resources/web/server/simulation_server.py#L264) is leaking resources (slowing down the server over time).

I think it might be related to using tornado and the asyncio event loop (https://github.com/cyberbotics/webots/blob/2f7f45f430746978ecc3dc399fc7dd5d4df362a2/resources/web/server/simulation_server.py#L307) inside a thread. Usually tornado does not seem thread safe.

If I comment L310 to L317 it seems to not create the zombie process. But I think it should be a way to close the websocket connection and the loop cleanly.

There are many lines related to closing the client.client_websocket:

https://github.com/cyberbotics/webots/blob/2f7f45f430746978ecc3dc399fc7dd5d4df362a2/resources/web/server/simulation_server.py#L142

But it does not seem to close it effectively.

Regards,

Felipe.

omichel:
I believe the main problem is that sometimes, some Webots processes are not quitting and remain active, consuming system resources. Maybe the script could be improved to monitor the active Webots processes (and also Webots controller processes) and kill those which have been inactive for a long period of time.

lfelipesv:
I did not have trouble with Webots processes not quitting recently Olivier. Checking the processes leaking resources they are directly associated with the simulation server script.

I think, however, that the problem can be not related to tornado or asyncio. I made some additional tests and it seems that the loop (iter(client.webots_process.stdout.readline, b'')) checking for the webots stdout is still running even after a client is disconnected.

https://github.com/cyberbotics/webots/blob/2f7f45f430746978ecc3dc399fc7dd5d4df362a2/resources/web/server/simulation_server.py#L310

Is there a way to stop this function and complete the thread when a client is disconnected?

The variable client.idle is used for monitoring but I don't understand the meaning of lines 316-317.

lfelipesv:
Oh, now thinking better, this termination had to be addressed by killing the webots process https://github.com/cyberbotics/webots/blob/2f7f45f430746978ecc3dc399fc7dd5d4df362a2/resources/web/server/simulation_server.py#L340 right?

omichel:
Not sure. Please try and report here any finding you may achieve.

lfelipesv:
[...]

PR #3412, opened 2021-07-14
solve thread hanging after webots process finishes

lfelipesv:
**Description**
This pull request address a bug that happens in the line https://github.com/cyberbotics/webots/blob/eca5ec4ce5774d175a75601a9ff396d4cb633a94/resources/web/server/simulation_server.py#L310 of the simulation server script.

*Bug behavior*: the thread that run webots hang after the user disconnect and the webots process is terminated. Because of that there exists proccesses using the server's memory (can be checked using htop - multiple processes appears linked to the simulation_server.py script) that leads to slow down and problems with the simulation server overtime.

*Solution proposed*: Check if the webots process in terminated in the blocking 'for' loop. If the webots process is None (meaning that the user disconnected and the webots process was terminated or killed) the loop breaks and the thread is completed.

**Related Issues**
This pull-request fixes issue #3400

**Tasks**
No task addressed.

**Documentation**
No need for change in the documentation.

**Screenshots**
No need for screenshots.

**Additional context**
Add any other context about the pull-request here.

### DIFF
--- a/docs/reference/changelog-r2021.md
+++ b/docs/reference/changelog-r2021.md
@@ -76,6 +76,7 @@ Released on July, 5th, 2021.
       - Exposed the `robot_description` ROS parameter (activated through the `--robot-description` flag) that contains the URDF of the robot.
     - Improved the URDF naming convention ([#2875](https://github.com/cyberbotics/webots/pull/2875)).
   - Bug fixes:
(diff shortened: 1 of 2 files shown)


omichel (approved):
Thank you.

Issue #3411, opened 2021-07-14
Pen drawing on Mac doesn't evaporates


BenjaminDeleze:
**Describe the Bug**
The first segment of line never evaporates in pen.wbt.

**Steps to Reproduce**
1. Go to 'devices/pen.wbt'
2. Run
3. See that the first segment (black) does not evaporate.

**Expected behavior**
The first segment should evaporate as the other ones

**Screenshots**
![pen](https://user-images.githubusercontent.com/25938827/125586365-6cff0a24-45a9-4b6d-84a8-556ccfae5acf.png)

**Additional context**
Tested on MacOs big sur
Could not reproduce on Ubuntu 20

ad-daniel:
![pen](https://user-images.githubusercontent.com/44834743/125588247-0aa936ea-7177-48cf-bd68-719ff2acb110.png)

Same behavior on windows, it gets lighter at the beginning but never actually vanishes.

BenjaminDeleze:
The issue is due to a too small ink_intensity (0.000008).
It is not reproducible on Ubuntu in pen.wbt as it does not return the same rand() suite, but if I force the ink_intensity to be that small I can reproduce the issue

PR #3416, opened 2021-07-14
Fix pen evaporation

ad-daniel:
**Description**
Fixes https://github.com/cyberbotics/webots/issues/3411.

Although the ink density is a floating point value in the API, the value was cast into an `unsigned char` when sending it to webots. For small values of the random seed, the first value of `rand()` was consistently very small which resulted in it being approximated to zero. On Webots side, there was no check to verify if a division by zero was occurring, hence this first segment of the trail was actually a NaN so it persisted despite the evaporation.

Mac:
pen

Windows:
pen_1

### DIFF
--- a/projects/samples/devices/controllers/pen/pen.c
+++ b/projects/samples/devices/controllers/pen/pen.c
@@ -49,6 +49,8 @@ int main() {
 
   wb_robot_init();
 
+  srand(123456789); // arbitrary value that starts in a pretty way
+
   /* get a handler to the distance sensors. */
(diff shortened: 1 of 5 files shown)


ad-daniel:
Since previously the value set by the controller and the value effectively applied differed (due to the casting of the double as `unsigned char` when sending it to webots), the values hard-coded in the test `distance_sensor_infra_red` returned by the distance sensor changed slightly, which is why due to the change this test also had to be adapted.

omichel (approved):
Thank you!

PR #3417, opened 2021-07-14
Adjust installation instructions for tarball

### LINKED ISSUE #3410: Incomplete instructions to install video creation packages on Ubuntu 18.04
Sample code provided in the Webots Guide  to install the video creation dependencies are incomplete and after running them the video creation still fails.

stefaniapedrazzi:
Fix #3410:

on Ubuntu 18.04, if Webots is installed using the tarball the previous instructions were not sufficient to successfully create and playback a movie.

It seems strange that even on Ubuntu 20.04 the given instructions will work as for sure `ffmpeg` is required.
But I will check if there are some differences.

### DIFF
--- a/docs/guide/installation-procedure.md
+++ b/docs/guide/installation-procedure.md
@@ -108,12 +108,13 @@ You will need to install *make* and *g++* to compile your own robot controllers.
 Other particular libraries could also be required to recompile some of the distributed binary files.
 The package names could slightly change on different releases and distributions.
 In this case an error message will be printed in the Webots console mentioning the missing dependency.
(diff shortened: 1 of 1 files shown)


stefaniapedrazzi:
I need the exact same packages also on Ubuntu 20.04.

ad-daniel (approved):
I don't recall having installed ffmpeg on my ubuntu 20 either, but it appears to be installed already regardless (due to python3-matplotlib apparently). Either way, works for me, thank you.

stefaniapedrazzi:
Installation instructions should be checked on clean systems.
It is also important to test the Webots tarball before installing the Webots debian package, otherwise all the dependencies will already be installed on the system.

PR #3426, opened 2021-07-14
Update tutorial-8-using-ros.md with new webots-ros dependency

stefaniapedrazzi:
`webots-ros` package now depends on a new package that needs to be manually installed (used by the new TIAGo example) to successfully compile the sources.


stefaniapedrazzi:
replaced by #3433

PR #3427, opened 2021-07-14
Removed path to node.js and cleaned-up

omichel:
This PR cleans-up the msys.sh script:
- remove the path to node binaries which are now removed from Webots.
- add path to Python, Java and MATLAB only if the corresponding environment variables are set.
- in case several python versions are installed, prefer python 3.8, then 3.9 and 3.7.

### DIFF
--- a/scripts/install/msys.sh
+++ b/scripts/install/msys.sh
@@ -1,12 +1,26 @@
 #!/bin/bash
 
-if [ $# -eq 0 ]
-then
+if [ $# -eq 0 ]; then
 export PATH=/mingw64/bin:/usr/bin:/c/WINDOWS/system32
 cd ~
 else
 WINDOWS_DRIVE=${1:1:1}:\\${1:3}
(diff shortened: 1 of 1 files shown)


lukicdarkoo (approved):
The `/c/WINDOWS/system32:/c/WINDOWS` path used to be after Python. Hopefully, it doesn't have side effects.

Otherwise, it looks good to me, thanks.

omichel:
> The `/c/WINDOWS/system32:/c/WINDOWS` path used to be after Python. Hopefully, it doesn't have side effects.

I did it on purpose as it seems cleaner. I didn't observed any side effect so far.

omichel:
> The `/c/WINDOWS/system32:/c/WINDOWS` path used to be after Python. Hopefully, it doesn't have side effects.

That is indeed causing a side effect: The top `Makefile` relies on the `find.exe` command which is now taken from `C:\WINDOWS\System32` instead of `/usr/bin`. Obvious this command is not the same and is causing an error: `FIND: Parameter format not correct`. I will fix it.

PR #3428, opened 2021-07-14
fix assets downloads on copy

### LINKED ISSUE #3419: Blocked assets download when copying a robot
Blocked assets download dialog is displayed on robot copy.

1. Open 'coupled_motors.wbt'
2. Copy the `GRIPPER` Robot
3. The loading world dialog opens saying it is downloading the assets, moreover it takes really long and don't even complete.

**Expected behavior**
If the robot is already loaded, there should not be any need to download again the assets on copy.
In any case, even if the copy should be necessary, the download should work and not be blocked.

BenjaminDeleze:
**Description**
When copying a object (non-Proto) with a DEF and a USE with urls inside, there was an infinite downloading popup.

This is because a node was deleted before it could finish checking that the URLs were in the cache.
When deleting a node with a WbDownloader, it is not enough to remove the downloads, we also need to decrement the number of total downloads.

Fix #3419

### DIFF
--- a/src/webots/nodes/utils/WbDownloader.cpp
+++ b/src/webots/nodes/utils/WbDownloader.cpp
@@ -57,6 +57,9 @@ WbDownloader::~WbDownloader() {
     if (gUrlCache.contains(mUrl))
       gUrlCache.remove(mUrl);
   }
+
+  if (!mFinished)
+    gCount--;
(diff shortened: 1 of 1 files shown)


omichel (approved):
Thanks.

Issue #3265, opened 2021-07-01
World clean-up


lukicdarkoo:
There are worlds that are not useful and some that don't use PBRAppearance. This PR proposes related changes. Note that this is not an exhaustive list, but only the initial one.

Worlds:
- `webots/projects/robots/neuronics/ipr/worlds/ipr_factory.wbt`: upgrade appearence
- `webots/projects/robots/neuronics/ipr/worlds/ipr_models.wbt`: delete
- `webots/projects/robots/neuronics/ipr/worlds/ipr_collaboration.wbt`: upgrade appearence
- `webots/projects/robots/neuronics/ipr/worlds/ipr_cube.wbt`: upgrade appearence
- `webots/projects/robots/k-team/koala/worlds/koala.wbt`: upgrade appearance
- `webots/projects/robots/robotis/darwin-op/worlds/visual_tracking.wbt`: delete (more/less copy of `webots/projects/robots/robotis/darwin-op/worlds/darwin-op.wbt`)
- `webots/projects/robots/epfl/biorob/worlds/salamander.wbt`: upgrade appearence
- `webots/projects/robots/epfl/biorob/worlds/ghostdog.wbt`: upgrade appearence

Robots:
- Khepera I: upgrade the appearance
- Khepera II: upgrade the appearance
- BioRob Yamor: upgrade the appearance
- PICAXE Microbot: upgrade the appearance
- HOAP-2: upgrade the appearance

ad-daniel:
Also `get_the_candies.wbt`, it says: `ERROR: 'Floor.proto': JavaScript error: The 'texture' field is deprecated, using the new 'appearance' field instead.`
It should be upgraded to PBRA though.

PR #3431, opened 2021-07-15
Fix `get_the_candies.wbt`

ad-daniel:
**Description**
It was reported [here](https://github.com/cyberbotics/webots/issues/3265), but contrary to the others that just need to be beautified, this one actually uses a deprecated field.


omichel (approved):
Thank you.

PR #3433, opened 2021-07-15
Add the MoveIt dependency to documentation

### LINKED ISSUE #3422: ROS test: compilation failure due to `moveit_ros_planning_interface`
`moveit_ros_planning_interface` is not automatically installed and `catkin_make` on Ubuntu 18.04 / melodic fails because of it.
I tested with a local copy of the `webots_ros` package, but the error is likely to occur also with the published package.

If this package cannot be installed automatically and needs to be installed manually, then we have to update the ROS tutorial accordingly:
https://cyberbotics.com/doc/guide/tutorial-8-using-ros?version=R2021b

lukicdarkoo:
Fixes #3422

### DIFF
--- a/docs/guide/tutorial-8-using-ros.md
+++ b/docs/guide/tutorial-8-using-ros.md
@@ -19,7 +19,7 @@ To install the latest version of ROS on Ubuntu use the following commands:
 sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros-latest.list'
 sudo curl -sSL https://raw.githubusercontent.com/ros/rosdistro/master/ros.key -o /usr/share/keyrings/ros-archive-keyring.gpg
 sudo apt-get update
-sudo apt-get install ros-noetic-desktop-full # takes time, get a coffee :)
+sudo apt-get install ros-noetic-desktop-full ros-noetic-moveit # takes time, get a coffee :)
 sudo apt-get install python3-rosdep
 sudo rosdep init
 rosdep update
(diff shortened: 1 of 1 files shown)


lukicdarkoo:
@stefaniapedrazzi I have just seen this one #3426 :/ We need full MoveIt installation because of MoveIt plugins

docs/guide/tutorial-8-using-ros.md
```diff
@@ -19,7 +19,7 @@ To install the latest version of ROS on Ubuntu use the following commands:
 sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros-latest.list'
 sudo curl -sSL https://raw.githubusercontent.com/ros/rosdistro/master/ros.key -o /usr/share/keyrings/ros-archive-keyring.gpg
 sudo apt-get update
-sudo apt-get install ros-noetic-desktop-full # takes time, get a coffee :)
+sudo apt-get install ros-noetic-desktop-full ros-noetic-moveit # takes time, get a coffee :)
```
stefaniapedrazzi, line 22: Just note that for compiling `webots_ros` it is enough to install `ros-noetic-moveit-ros-planning-interface`.

stefaniapedrazzi (approved):

docs/guide/tutorial-8-using-ros.md
stefaniapedrazzi: Additional packages may be required to run specific examples.
I think that in any case these specific packages should then be documented in the ROS tutorials page.

lukicdarkoo: Yes, but running the example may fail without the full installation.

stefaniapedrazzi: It is ok.
But note that other packages are also missing to run specific examples, like for example gmapping and probably also others.
So this snippet is not sufficient in any case.

PR #3436, opened 2021-07-15
Fix range finder with fully transparent objects

### LINKED ISSUE #3405: RangeFinder does see fully transparent objects
Sample world "range_finder". change DEF GREEN_BOX solid > shape >  appearance PBRAppearance > transparency to 1.

now the green box is fully transparent, but still visible by the RangeFinder.

delete PBRAppearance, add Appearance > material > transparency 1: same error. 

delete boundingObject: same error.

(https://www.cyberbotics.com/doc/reference/rangefinder#description:  Range-finder detects semi-transparent objects as if they were not transparent. An object can be semi-transparent either if its texture has an alpha channel, or if its Material transparency field is not equal to 1.)

webots --sysinfo
System: Linux 5.8.0-59-generic x86_64
Processor: Intel(R) Core(TM) i5-4590 CPU @ 3.30GHz
Number of cores: 4
OpenGL renderer: GeForce GTX 1050/PCIe/SSE2
OpenGL version: 4.6.0 NVIDIA 460.80

both webots_2021b_amd64_jul_7.deb (draft) and webots_2021a.deb had error. 
ubuntu 20.04

BenjaminDeleze:
**Description**
If a shape has a fully transparent (tranparency = 1) appearance, it should not be visible by the cameras.

**Related Issues**
Fix #3405

### DIFF
--- a/docs/reference/changelog-r2021.md
+++ b/docs/reference/changelog-r2021.md
@@ -150,6 +150,7 @@ Released on July, 5th, 2021.
     - Fixed crash provoked by canceling and switching world in the Guided Tour ([#3376](https://github.com/cyberbotics/webots/pull/3376)).
     - Fixed the path to the python command on macOS ([#3402](https://github.com/cyberbotics/webots/pull/3402)).
     - Fixed [HighwaySign](../guide/object-traffic.md#highwaysign) PROTO not displaying the texture ([#3407](https://github.com/cyberbotics/webots/pull/3407)).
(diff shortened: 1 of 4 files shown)


omichel:
I would add a line in the change log about this fix.

BenjaminDeleze:
Note that this fix works for camera, rangefinder and lidar, but not for laser pointer and distance sensor.

ad-daniel (approved):
Thank you, looks good to me.
As discussed, it might be worth opening a new issue regarding the other types of sensors.

Issue #3307, opened 2021-07-02
ROS libraries


mael25:
**Describe the Bug**
When launching the `e_puck_line.launch` ros example, I get the following error:

```
/snap/webots/x1/usr/share/webots/projects/default/controllers/ros/ros: error while loading shared libraries: libboost_system.so.1.65.1: cannot open shared object file: No such file or directory
WARNING: 'ros' controller exited with status: 127.
```

I had to manually add the ros libraries to the `LD_LIBRARY_PATH`:

export LD_LIBRARY_PATH=${WEBOTS_HOME}/lib/controller:`${WEBOTS_HOME}/projects/default/controllers/ros/lib/ros`

As I have not seen in the tutorial that it was needed to do so, I wonder if this is normal.

**System**
 - Ubuntu 20.04, Snap version of Webots

lukicdarkoo:
The missing `libboost_system.so.1.65.1` library was reported by a Discord user, so it seems to be a real problem. Did you use the ROS package from apt or you compiled it from develop branch?

mael25:
From apt

lukicdarkoo:
This works for me:
```
$ source /opt/ros/noetic/local_setup.bash
$ export WEBOTS_HOME=/home/lukic/Downloads/webots-R2021b-x86-64/webots
$ roslaunch webots_ros e_puck_line.launch

# the simulation works

$ echo $LD_LIBRARY_PATH
/opt/ros/noetic/lib:/opt/ros/noetic/lib/x86_64-linux-gnu
$ echo $PYTHONPATH
/opt/ros/noetic/lib/python3/dist-packages
$ echo $WEBOTS_HOME
/home/lukic/Downloads/webots-R2021b-x86-64/webots
```

lukicdarkoo:
I managed to reproduce it with Webots snap

PR #3438, opened 2021-07-15
Document the issue with snap and noetic

stefaniapedrazzi:
Address #3307:
document the existing issue with the snap package and ROS noetic.

### DIFF
--- a/docs/guide/tutorial-8-using-ros.md
+++ b/docs/guide/tutorial-8-using-ros.md
@@ -26,6 +26,12 @@ rosdep update
 sudo apt-get install ros-noetic-webots-ros
 ```
 
+If you installed Webots from the snap, you also need to setup the `LD_LIBARAY_PATH` variable:
+```sh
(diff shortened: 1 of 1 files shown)


stefaniapedrazzi:
@mael25 @lukicdarkoo  Could you please check if this is correct to make it work and if not update it?

lukicdarkoo:
Prepending didn't work for me, appending did, and only `${WEBOTS_HOME}/projects/default/controllers/ros/lib/ros` is needed. I created this in almost same time #3439

stefaniapedrazzi:
> Prepending didn't work for me, appending did, and only `${WEBOTS_HOME}/projects/default/controllers/ros/lib/ros` is needed. I created this at almost the same time #3439 :/

Ok fine. We can close this.
Could you just complete your branch adding code instead of text and also specifying the value of WEBOTS_HOME (it is is not straightforward with snap)?

PR #3442, opened 2021-07-15
Added a warning if a mesh cannot cast shadows

omichel:
This PR:
- [x] makes Webots display a warning if a `Shape` with `castShadows TRUE` contains a mesh geometry which is too large to cast shadows, e.g., the mesh has more than 21845 triangles (21845 = 65535 / 3).
- [ ] Fix all the proto files raising this warning.
- [ ] Fix all the world files raising this warning.


mael25:
Should the robotino3.x3d file be updated too?

mael25 (approved):
Thank you!

PR #3446, opened 2021-07-15
Optimization persistent frame buffer

mael25:
**Description**
The goal of this PR is to evaluate the gain in performance that would provide the use of a persistent buffer in `FrameBuffer.cpp`.
Persistent buffers are available since OpenGL 4.4 (they use `ARB_buffer_storage` extension), so this feature would be possible only for latest user configurations.

(PR not yet stable)

**Documentation**
Persistent buffer explained: https://www.bfilipek.com/2015/01/persistent-mapped-buffers-in-opengl.html
glBufferStorage: https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glBufferStorage.xhtml


BenjaminDeleze:
As this function is only available since OpenGL 4.4, it will no be compilable with Emscripten to be used in WrenJS.

If you use it only for the `wr_frame_buffer_copy_contents`, I think it is not a problem as we currently don't use this function in WebotsJS (at least for now). We could just make an exception to make Emscripten ignore this function.

However, I am not sure about the fact to introduce functionality that are available only for version above OpenGL 3.3.

mael25:
The persistent buffer is not fully stable yet but it brings encouraging results. Here are the performances measured on `mybot_camera.wbt` with a modified `Camera` resolution of 3000x2000. Performances are measured using `webots --log-performance`, `nvidia-smi` and `std::chrono::steady_clock::now()`. Elapsed time in function includes the buffer mapping/unmapping (if needed) as well as the copy of the frame from GPU to CPU.

| Measurement | Master | Optimization
| --- | --- | --- 
| Avg speed factor | 4.07x | _9.26x_*
| **Elapsed time in function** | **20.8ms** | **3.1ms**
| Webots Memory usage | 383/4041 MiB | 424/4041 MiB
| Volatile GPU-Util | 45%-65% | 100%

Using a persistent buffer seems to be really effective, the time needed to copy being drastically reduced. However, in fast mode, the usage of the GPU is 100%, which results in Ubuntu freezing regularly for a few seconds.  So this would be important to solve.

 _*Due to the 100% usage of GPU, this value is currently quite varying: it is often 7x, but can switch to 30x (I guess if there is no other process needing too much GPU resources at the same time)._

mael25:
After adding the synchronization object, the performances are similar to the current buffer. However, it might be interesting in the future to see if a performance gain can be obtained by changing the position of the  `glFenceSync` / `glClientWaitSync` calls.

PR #3415, opened 2021-07-14
Fixed issues with generic robot window

stefaniapedrazzi:
Address #3414:
Setting `LD_LIBRARY_PATH` is needed also for python controllers in order to correctly load the generic robot window.

The documentation was previously changed in #2940.

PR #3449, opened 2021-07-16
Fix robot window linking with an external Python controller

lukicdarkoo:
It should tackle this:
https://github.com/cyberbotics/webots/pull/3415#issuecomment-880673947

The test process:
```bash
make distrib
# move the tarbal to another location and extract it there
git clean -dfx
# test the external controller 
```


omichel:
Yes, you are right, my fix doesn't work with robot windows... So reverting it fully is the way to go.

omichel (approved):
Thank you.

PR #3456, opened 2021-07-16
Fix imageTextures urls format

### LINKED ISSUE #3451: Wrong urls export format in streaming viewer when building from sources.
**Describe the Bug**
Urls of `ImageTexture` are not exported correctly in streaming viewer mode when Webots is built from the sources.
The urls are in the form `webots://` whereas they should have the `raw.githubusercontent` format.

Note that the urls are exported correctly for background images, animations and for released version of Webots.

**Additional context**
WebotsJS will convert `webots://` urls to `raw.githubusercontent` ones pointing to the `release branch`, so this will be an issue only for images introduced after a release.

BenjaminDeleze:
**Description**
Fix #3451

### DIFF
--- a/src/webots/nodes/WbImageTexture.cpp
+++ b/src/webots/nodes/WbImageTexture.cpp
@@ -518,12 +518,11 @@ void WbImageTexture::exportNodeFields(WbVrmlWriter &writer) const {
   WbField urlFieldCopy(*findField("url", true));
   for (int i = 0; i < mUrl->size(); ++i) {
     if (mUrl->value()[i].indexOf("webots://") == 0) {
-      if (writer.isWritingToFile()) {
-        QString newUrl = mUrl->value()[i];
-        dynamic_cast<WbMFString *>(urlFieldCopy.value())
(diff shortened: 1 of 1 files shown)


stefaniapedrazzi (commented):
Note that `WbApplicationInfo::branch()` is not working correctly, at least in my development system, because of the wrong value stored in `resources/branch.txt`.
It seems that is an issue with the script or Makefile because even after recompiling from the root folder the `branch.txt` value is not updated.
With `make clean` is correctly removed but it is never recreated.

So in any case this fix won't solve my issues running animations and streaming on a development system.

stefaniapedrazzi:
The issue is caused because this script doesn't handle local branches detached from origin:
https://github.com/cyberbotics/webots/blob/master/scripts/get_git_info/get_git_info.sh#L6
So you need to create a branch to be able to properly run streaming an animations.

I don't fully understand this limitation and I think that this will cause us many issues with non working functionalities.

stefaniapedrazzi (approved):
Thank you.

Apart from the comment on extracting the current branch, the fix seems fine.

BenjaminDeleze:
I just checked and you are right. It was thought to work for normal branches and not for detached HEAD.
But I see no reason to keep the limitation on `HEAD`. I can work on a fix for that.

PR #3458, opened 2021-07-16
Fix default url branch

BenjaminDeleze:
**Description**
The branch should be `released` when the variable is an empty string


omichel (approved):
Thank you.

PR #3463, opened 2021-07-19
Update master version to R2021b revision 1

stefaniapedrazzi:
It is not planned to release R2021b revision 1 but to jump directly to R2021c.
However, I think it is better to set it to R2021b revision 1 until R2021c is ready so that it will be easier to release a bug fix version if we should have a urgent need for it.

### DIFF
--- a/scripts/new_version/new_version.sh
+++ b/scripts/new_version/new_version.sh
@@ -61,8 +61,6 @@ $CURRENT_DIR/new_version_file.sh "package:\\s'.*'" "package: '"$new_package"'" $
 if [ $new_version_year -ne $old_version_year ]; then
   $CURRENT_DIR/new_version_file.sh "year:\\s[0-9]\+" "year: "$new_version_year $WEBOTS_HOME/docs/js/showdown-extensions.js
 fi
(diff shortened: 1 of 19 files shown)


scripts/new_version/new_version.sh
```diff
@@ -81,5 +79,7 @@ then
 
   $CURRENT_DIR/new_version_file.sh "wwi\/$old_version_without_revision\/" "wwi\/$new_version_without_revision\/" $WEBOTS_HOME/docs/dependencies.txt
 
+  $CURRENT_DIR/new_version_file.sh $old_version_without_revision $new_version_without_revision $WEBOTS_HOME/projects/humans/c3d/plugins/robot_windows/c3d_viewer_window/c3d_viewer_window.html
```
lukicdarkoo, line 82: This seems to be an unrelated change, why is it changed?

stefaniapedrazzi: It is better to change the version of files used in the c3d robot window only on major released.
So it is related because if run the script without fixing this the c3d robot window file will get updated.

PR #3469, opened 2021-07-19
Fix branch.txt creation

BenjaminDeleze:
**Description**
Compute the branch name correctly if Webots is in a `HEAD detached` branch.

I wonder if we should add a default value to `branch.txt` and `repo.txt` in case the script fails to get them.
For example, in the case a user build Webots from the `source code.zip`package and so is not in a github repository and doesn't  already has raw.githubusercontent urls.

For the default repo, I would put `cyberbotics/webots`
For the default branch I do not really know between `master` and `released`

### DIFF
--- a/scripts/get_git_info/get_git_info.sh
+++ b/scripts/get_git_info/get_git_info.sh
@@ -3,7 +3,10 @@
 # Get the name of the working branch and write it to a file if correct
 branch=$(git branch 2>&1)
 
-if [[ $branch != fatal* ]] && [[ $branch != '* (HEAD'* ]]
+if [[ $branch == '* (HEAD'* ]]
+then
+	echo "$branch" | $(sed -n -e 's/^\* \(.*\)/\1/p' | cut -c 19- | rev | cut -c 2- | rev > $WEBOTS_HOME/resources/branch.txt)
+elif [[ $branch != fatal* ]]
 then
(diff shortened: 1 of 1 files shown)


stefaniapedrazzi:
Yes, I think it would be better to have default values.
Given that this should only be computed on development environment I think we could use `master` as default branch.

stefaniapedrazzi (changes requested):
Thank you.

In case a branch is detached from HEAD (for example ``origin/fix-cppcheck-errors``) I'm currently getting: ``origin/fix-cppcheck-errors`` in `branch.txt`.
But this won't work in URLs, it seems it should be just `fix-cppcheck-errors` instead.

BenjaminDeleze:
I did not think about that (not familiar with branch detached from HEAD).
Are there other situations than `origin/bran_name` and `commit_number` ?

stefaniapedrazzi:
> I did not think about that (not familiar with branch detached from HEAD).
> Are there other situations than `origin/bran_name` and `commit_number` ?

I'm not sure either, but we will cover most of the uses cases supporting these two kind of branch names.

stefaniapedrazzi:
Similar to the `commit_id` there is also the `tag`.

BenjaminDeleze:
> Similar to the `commit_id` there is also the `tag`.

Yes but it will work as the beginning and the end of the message are the same than the ones in the case of `commit_id`

stefaniapedrazzi (approved):
Thank you!

PR #3470, opened 2021-07-19
Add exception handling in publish_release.py script

stefaniapedrazzi:
Publishing a nightly build failed (https://github.com/cyberbotics/webots/runs/3071646906?check_suite_focus=true) with this error message:
```
  Creating release "Webots Nightly Build (14-7-2021)" with tag "nightly_14_7_2021" on commit "5569ee04f8b281e6f8077883657717f4e396f9f7"
  File "scripts/packaging/publish_release.py", line 103, in <module>
    prerelease=True)
  File "/usr/local/lib/python3.6/dist-packages/github/Repository.py", line 1009, in create_git_tag_and_release
    target_commitish=object,
  File "/usr/local/lib/python3.6/dist-packages/github/Repository.py", line 1061, in create_git_release
    "POST", f"{self.url}/releases", input=post_parameters
  File "/usr/local/lib/python3.6/dist-packages/github/Requester.py", line 355, in requestJsonAndCheck
    verb, url, parameters, headers, input, self.__customConnection(url)
  File "/usr/local/lib/python3.6/dist-packages/github/Requester.py", line 378, in __check
    raise self.__createException(status, responseHeaders, output)
github.GithubException.GithubException: 422 {"message": "Validation Failed", "errors": [{"resource": "Release", "code": "already_exists", "field": "tag_name"}], "documentation_url": "https://docs.github.com/rest/reference/repos#create-a-release"}
Error: Process completed with exit code 1.
```

The script already checks if a release already exists but in this particular case it was not sufficient.
Handling the exceptions raised when creating the tag and release we should be able to bypass this kind of errors and still upload the packages if the release effectively exists.

### DIFF
--- a/scripts/packaging/publish_release.py
+++ b/scripts/packaging/publish_release.py
@@ -101,21 +101,27 @@ if not releaseExists:
 
     if tagExists:
         print('Tag "%s" already exists.' % (tagName))
-        repo.create_git_release(tag=tagName,
-                                name=title,
-                                message=message,
-                                draft=draft,
-                                prerelease=True,
-                                target_commitish=options.commit)
+        try:
+            repo.create_git_release(tag=tagName,
+                                    name=title,
+                                    message=message,
+                                    draft=draft,
+                                    prerelease=True,
+                                    target_commitish=options.commit)
+        except GithubException as e:
+            print('Creation of release failed: ', e.data)
     else:
-        repo.create_git_tag_and_release(tag=tagName,
-                                        tag_message=title,
-                                        release_name=title,
-                                        release_message=message,
-                                        object=options.commit,
-                                        type='commit',
-                                        draft=draft,
-                                        prerelease=True)
+        try:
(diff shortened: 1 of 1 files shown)


lukicdarkoo (commented):
Looks good to me, but I am only concerned about the added try/except, maybe we shouldn't be silent about it?

stefaniapedrazzi:
> Looks good to me, but I am only concerned about the added try/except, maybe we shouldn't be silent about it?

Handling the exception and continue with the execution is exactly the objective of this change.
It is not clear why the tag/release creation failed and the error was saying that the release already exists.
So even in case of error we should continue with the script because it is likely that the package can  be uploaded despite the GitHub exception.

We are monitoring every day that all the packages are uploaded correctly and in case of error the error is still printed in the logs.
So the error is not silent and can still be retrieved.

lukicdarkoo (approved):
Thank you

PR #3478, opened 2021-07-21
Don't publish the libcontroller if it hasn't changed since yesterday

lukicdarkoo:
For example, Windows builds different binaries even though the source code hasn't changed. It generates unnecessary diff in the `webots-libcontroller` repository.

This PR makes the `sync_controller_lib.sh` script skip the execution if the libcontroller hasn't changed.


stefaniapedrazzi (approved):
Thank you

PR #3480, opened 2021-07-21
Clean code related to Ubuntu 16.04

### LINKED ISSUE #3424: Cleanup code related to Ubuntu 16.04
Some documentation related to Ubuntu 16.04 has been removed.
If we don't support any longer Ubuntu 16.04 all the related code has to be removed (for example in `webots_distro.c`).

mael25:
**Description**
Fixes #3424
Remove the code related to Ubuntu 16.04 that is no more needed

**Remaining Tasks**

- [x] https://github.com/cyberbotics/webots-dev/pull/7470 --> Fix not needed anymore.
- [x] ~~https://github.com/cyberbotics/webots/blob/master/projects/robots/robotis/darwin-op/remote_control/Makefile.robotis-op2 and https://github.com/cyberbotics/webots/blob/master/projects/robots/robotis/darwin-op/controllers/Makefile.robotis-op2 (update X11 library path)~~ --> These Makefile depends on the OS of the robot, not the computer (so must be kept as it is).
- [x] Add entry about this PR in the Changelog (R2021c or R2022a)

### DIFF
--- a/docs/local_exporter.py
+++ b/docs/local_exporter.py
@@ -47,7 +47,6 @@ def download(url, target_file_path):
             request = Request(url, headers={'User-Agent': 'Mozilla'})
             if platform.system() == 'Linux' and \
                hasattr(ssl, 'create_default_context'):
-                # On Ubuntu 16.04 there are issues with the certificates.
                 ctx = ssl.create_default_context()
                 ctx.check_hostname = False
                 ctx.verify_mode = ssl.CERT_NONE
--- a/docs/local_exporter.py
+++ b/docs/local_exporter.py
@@ -47,6 +47,7 @@ def download(url, target_file_path):
             request = Request(url, headers={'User-Agent': 'Mozilla'})
             if platform.system() == 'Linux' and \
                hasattr(ssl, 'create_default_context'):
+                # On Ubuntu 16.04 there are issues with the certificates.
                 ctx = ssl.create_default_context()
                 ctx.check_hostname = False
                 ctx.verify_mode = ssl.CERT_NONE
--- a/docs/reference/changelog-r2021.md
+++ b/docs/reference/changelog-r2021.md
@@ -1,7 +1,7 @@
 # Webots R2021 Change Log
 
 ## Webots R2021c
-Released on September, 30th, 2021.
+Released on XX, XXth, 2021.
 
   - Dependency Updates
     - **Stopped support for Ubuntu 16.04 ([#3480](https://github.com/cyberbotics/webots/pull/3480)).**
--- a/docs/reference/changelog-r2021.md
+++ b/docs/reference/changelog-r2021.md
@@ -1,5 +1,11 @@
 # Webots R2021 Change Log
 
+## Webots R2021c
+Released on September, 30th, 2021.
+
+  - Dependency Updates
+    - **Stopped support for Ubuntu 16.04 ([#3480](https://github.com/cyberbotics/webots/pull/3480)).**
+
 ## Webots R2021b
 Released on July, 16th, 2021.
 
@@ -150,7 +156,7 @@ Released on July, 16th, 2021.
     - Fixed crash provoked by canceling and switching world in the Guided Tour ([#3376](https://github.com/cyberbotics/webots/pull/3376)).
     - Fixed the path to the python command on macOS ([#3402](https://github.com/cyberbotics/webots/pull/3402)).
     - Fixed [HighwaySign](../guide/object-traffic.md#highwaysign) PROTO not displaying the texture ([#3407](https://github.com/cyberbotics/webots/pull/3407)).
-    - Fixed the transparency of [Material](material.md) and [PBRAppearance](pbrappearance.md) such that the object are no longer visible to cameras when completely transparent ([#3436](https://github.com/cyberbotics/webots/pull/3436)). 
+    - Fixed the transparency of [Material](material.md) and [PBRAppearance](pbrappearance.md) such that the object are no longer visible to cameras when completely transparent ([#3436](https://github.com/cyberbotics/webots/pull/3436)).
   - Cleanup
     - Deleted deprecated DifferentialWheels node ([#2749](https://github.com/cyberbotics/webots/pull/2749)).
     - Changed structure of the [projects/samples/howto]({{ url.github_tree }}/projects/samples/howto) directory, so each demonstration is in a dedicated directory ([#2639](https://github.com/cyberbotics/webots/pull/2639)).
--- a/scripts/packaging/webots_distro.c
+++ b/scripts/packaging/webots_distro.c
@@ -409,7 +409,7 @@ static char **expand_wildcard_filename(const char *big_buffer, int *n) {
 
 static void add_ros_dependencies(const char *path) {
   fprintf(fd, "mkdir -p %s/projects/default/controllers/ros/lib/ros\n", path);
-#if defined(WEBOTS_UBUNTU_18_04)
+#ifdef WEBOTS_UBUNTU_18_04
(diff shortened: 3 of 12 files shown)


mael25:
Regarding the search of an existing X11 library in  [Makefile.robotis-op2](https://github.com/cyberbotics/webots/blob/master/projects/robots/robotis/darwin-op/controllers/Makefile.robotis-op2#L38), we can remove the part concerning Ubuntu <=10 and could modify the part for Ubuntu >=12 to this:
```
ifneq ($(wildcard /usr/lib/($OSARCH)-linux-gnu/libX11.so.6),)
LIBX11_SOURCE = /usr/lib/($OSARCH)-linux-gnu/libX11.so.6
endif
```

scripts/packaging/webots_distro.c
```diff
@@ -409,28 +409,7 @@ static char **expand_wildcard_filename(const char *big_buffer, int *n) {
...
-  fprintf(fd, "cp /usr/lib/x86_64-linux-gnu/libboost_thread.so.1.58.0 %s/projects/default/controllers/ros/lib/ros\n", path);
-  fprintf(fd, "cp /usr/lib/x86_64-linux-gnu/libboost_chrono.so.1.58.0 %s/projects/default/controllers/ros/lib/ros\n", path);
-  fprintf(fd, "cp /usr/lib/x86_64-linux-gnu/libboost_filesystem.so.1.58.0 %s/projects/default/controllers/ros/lib/ros\n", path);
-  fprintf(fd, "cp /usr/lib/x86_64-linux-gnu/liblog4cxx.so.10 %s/projects/default/controllers/ros/lib/ros\n", path);
-  fprintf(fd, "cp /usr/lib/x86_64-linux-gnu/libboost_regex.so.1.58.0 %s/projects/default/controllers/ros/lib/ros\n", path);
-  fprintf(fd, "cp /usr/lib/x86_64-linux-gnu/libconsole_bridge.so.0.2 %s/projects/default/controllers/ros/lib/ros\n", path);
-  fprintf(fd, "cp /usr/lib/x86_64-linux-gnu/libapr-1.so.0 %s/projects/default/controllers/ros/lib/ros\n", path);
-  fprintf(fd, "cp /usr/lib/x86_64-linux-gnu/libaprutil-1.so.0 %s/projects/default/controllers/ros/lib/ros\n", path);
-#elif defined(WEBOTS_UBUNTU_18_04)
+#if defined(WEBOTS_UBUNTU_18_04)
```
stefaniapedrazzi, line 412: This could be simplified to 
```suggestion
#ifdef WEBOTS_UBUNTU_18_04
```

scripts/packaging/webots_distro.c
```diff
@@ -1257,13 +1211,7 @@ static void create_file(const char *name, int m) {
...
       // add the required libraries in order to avoid conflicts on other Linux distributions
       add_ros_dependencies("debian/usr/local/webots");
-#ifdef WEBOTS_UBUNTU_16_04
-      fprintf(fd, "cp /usr/lib/x86_64-linux-gnu/libraw.so.15 debian/usr/local/webots/lib/webots\n");
-      fprintf(fd, "cp /usr/lib/x86_64-linux-gnu/libvpx.so.3 debian/usr/local/webots/lib/webots\n");
-      fprintf(fd, "cp /usr/lib/x86_64-linux-gnu/libx264.so.148 debian/usr/local/webots/lib/webots\n");
-      fprintf(fd, "cp /usr/lib/x86_64-linux-gnu/libavcodec-ffmpeg.so.56 debian/usr/local/webots/lib/webots\n");
-      fprintf(fd, "cp /usr/lib/x86_64-linux-gnu/libpng12.so.0 debian/usr/local/webots/lib/webots\n");
-#elif defined(WEBOTS_UBUNTU_18_04)
+#if defined(WEBOTS_UBUNTU_18_04)
```
stefaniapedrazzi, line 1214: This could be simplified
```suggestion
#ifdef WEBOTS_UBUNTU_18_04
```

docs/local_exporter.py
```diff
@@ -47,7 +47,6 @@ def download(url, target_file_path):
             request = Request(url, headers={'User-Agent': 'Mozilla'})
             if platform.system() == 'Linux' and \
                hasattr(ssl, 'create_default_context'):
-                # On Ubuntu 16.04 there are issues with the certificates.
```
stefaniapedrazzi, line 50: I would keep this comment because it explains why on Linux the behavior is different.

stefaniapedrazzi:
> Regarding the search of an existing X11 library in [Makefile.robotis-op2](https://github.com/cyberbotics/webots/blob/master/projects/robots/robotis/darwin-op/controllers/Makefile.robotis-op2#L38), we can remove the part concerning Ubuntu <=10 and could modify the part for Ubuntu >=12 to this:
> 
> ```
> ifneq ($(wildcard /usr/lib/($OSARCH)-linux-gnu/libX11.so.6),)
> LIBX11_SOURCE = /usr/lib/($OSARCH)-linux-gnu/libX11.so.6
> endif
> ```

> Regarding the search of an existing X11 library in [Makefile.robotis-op2](https://github.com/cyberbotics/webots/blob/master/projects/robots/robotis/darwin-op/controllers/Makefile.robotis-op2#L38), we can remove the part concerning Ubuntu <=10 and could modify the part for Ubuntu >=12 to this:
> 
> ```
> ifneq ($(wildcard /usr/lib/($OSARCH)-linux-gnu/libX11.so.6),)
> LIBX11_SOURCE = /usr/lib/($OSARCH)-linux-gnu/libX11.so.6
> endif
> ```

I don't think we ever had issues with the architecture.
So maybe it is not even necessary to compute it and we could keep the  hard coded `i386-linux-gnu`.
But if it is easy to get it is ok to improve it.

stefaniapedrazzi:
Personally I would include this change in R2022a (that will be released in December) and not R2021c.
It we decide so, then it is fine to clean up the code now but we should wait until R2021c is released before merging.

stefaniapedrazzi:
Actually `Makefile.robotis-op2` refers to the OS installed on the robot.
So I would keep it as is unless we know exactly which OS is currently installed on the Robotis robots.

mael25:
I have also removed the double underscore fix for Ubuntu 16.04 on Unity (https://github.com/cyberbotics/webots-dev/pull/7470) as it is no longer needed. Indeed, it was applied only on Linux if the environment variable XDG_CURRENT_DESKTOP=="Unity". Since Ubuntu 16.10, this variable has an other format, i.e. "Unity:Unity7" (see https://bugs.launchpad.net/ubuntu/+source/unity-control-center/+bug/1629668), so the fix is not applied anyway. I have tested on 18.04 with Unity, and in my case the variable was "Unity:Unity7:ubuntu", so the fix was not applied and the underscores were looking fine (not doubled) in the Robot context menu.

mael25:
So the last step will be to add an entry in the Changelog about this PR before merging (in R2021c or R2022a).

stefaniapedrazzi:
Note that this PR should target the develop branch.

stefaniapedrazzi:
Second though: it seems fine to apply this to R2021c.

mael25:
I added an entry to the Changelog and changed the targeted branch, so this should be ready to merge.

docs/reference/changelog-r2021.md
```diff
@@ -1,5 +1,11 @@
 # Webots R2021 Change Log
 
+## Webots R2021c
+Released on September, 30th, 2021.
```
stefaniapedrazzi, line 4: I would not set a date for the moment
```suggestion
Released on XX, XXth, 2021.
```

stefaniapedrazzi (changes requested):
Thank you

stefaniapedrazzi (approved):
Thank you!

Issue #3476, opened 2021-07-20
Use actual boolean in Lidar::isRotating


ShuffleWire:
Currently, Fixed or Rotating type of a Lidar is determined using 
```
bool isRotating() const { return mActualType.startsWith('r', Qt::CaseInsensitive); }
```
However, it's not very efficient as this is called often.
Would not be better to actually store a bool in `mActualType` instead of a `QString`, and only use "Fixed/Rotating" in the display ?

Also, I do think that something like the "Spherical" tick-box could be nice.

stefaniapedrazzi:
This is fine, but note that this won't really speedup the simulation as performance gain is really minimal.
For optimizations we should first identify bottlenecks and then try to solve them instead of randomly search in the code.

lukicdarkoo:
@ShuffleWire The `WbLidar::updatePointCloud` method is probably the biggest bottleneck (by far) in the LiDAR implementation:
https://github.com/cyberbotics/webots/blob/c5637734a664733a585219c178eb753c0ff8f37b/src/webots/nodes/WbLidar.cpp#L354-L381
If you ever find a time, please try making a PR that optimizes it (see some ideas in https://github.com/cyberbotics/webots/issues/3046).

ShuffleWire:
> @ShuffleWire The `WbLidar::updatePointCloud` method is probably the biggest bottleneck (by far) in the LiDAR implementation:
> 
> If you ever find a time, please try making a PR that optimizes it (see some ideas in #3046).

I will see if I can do anything, but I couldn't promise much... I will take a look, still !

PR #3487, opened 2021-07-22
Cleanup Lidar isRotating

mael25:
**Description**
Optimization proposed by @ShuffleWire in #3476.

Replaced ` bool isRotating()` and `Qstring mActualType` by `bool mIsActuallyRotating`, so that the bool is computed once and not at each step (as `Lidar` type changes are applied only after save & reload).


ad-daniel (approved):
For the variable name, is there a reason you changed the name to `mIsActuallyRotating` instead of calling it `mIsRotating` similarly to the previous function's name?

Also, I believe we're back to the old method, meaning that unless it's an hotfix we should target develop instead

ad-daniel (commented):
For the variable name, is there a reason you changed the name to `mIsActuallyRotating` instead of calling it `mIsRotating` similarly to the previous function's name?

Also, I believe we're back to the old method, meaning that unless it's an hotfix we should target develop instead

mael25:
> For the variable name, is there a reason you changed the name to `mIsActuallyRotating` instead of calling it `mIsRotating` similarly to the previous function's name?
> 
> Also, I believe we're back to the old method, meaning that unless it's an hotfix we should target develop instead

In `WbLidar.hpp`, all the members variables that correspond to a field of the `Lidar` but which are computed during initialization and not affected by the user's changes in the Scene Tree have the 'actual' prefix. This is the case for the `Lidar` `type`  field since #2983.

Yes, we can target `develop`.

mael25:
Replaced by #3490. (base branch change doesn't work)

PR #3490, opened 2021-07-22
Cleanup Lidar isRotating

mael25:
**Description**
Optimization proposed by @ShuffleWire in #3476. Replaces #3487.

Replaced ` bool isRotating()` and `Qstring mActualType` by `bool mIsActuallyRotating`, so that the bool is computed once and not at each step (as `Lidar` type changes are applied only after save & reload).


ad-daniel (approved):
Thank you

PR #3506, opened 2021-07-26
Python in virtual env, MacOS

BenjaminDeleze:
The changes introduced in #3402 cause the pythonpath to always be the python of the system.

However, it is not possible to completely revert #3402, otherwise python controllers would not work if Webots is launched from the GUI.

So now if:

- Webots is launched from the GUI: use the hardcoded, system pythonpath.
- Webots is launched from a terminal: use the default python command.

### DIFF
--- a/src/webots/control/WbLanguageTools.cpp
+++ b/src/webots/control/WbLanguageTools.cpp
@@ -24,6 +24,8 @@
 #include <QtCore/QStringList>
 #include <QtCore/QTextStream>
 
+#include <cstdlib>
+
 #ifdef _WIN32
 static const QChar PATHS_SEPARATOR(';');
 #else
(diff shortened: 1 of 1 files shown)


omichel:
Did you checked what happens if webots is launched with the *open* command from a Terminal?

BenjaminDeleze:
Yes, I know that for development environment it does not work (but it wasn't working before either).
But I think that I checked for binary and it worked (I will double check)

BenjaminDeleze:
I verified the case with `open` and it works with Webots.app but not with development environment (because of an intermediary shell)

src/webots/control/WbLanguageTools.cpp
```diff
@@ -102,34 +104,42 @@ QString WbLanguageTools::pythonCommand(QString &shortVersion, const QString &com
...
-  } else {
-    shortVersion = checkIfPythonCommandExist(pythonCommand, env, true);
-    if (shortVersion.isEmpty())
-      pythonCommand = "!";
-  }
 
-  if (pythonCommand == "!")
-    WbLog::warning(QObject::tr("Python was not found.\n") + advice);
+    if (pythonCommand == "!")
+      WbLog::warning(QObject::tr("Python was not found.\n") + advice);
```
stefaniapedrazzi, line 141: You could keep this outside the if/else block and write it only once in stead of duplicating it for the "PWD" case.

stefaniapedrazzi (approved):
I don't have a macOS machine to test.
But I checked the code and seems fine.

I also checked on Linux and the virtual environment seems to work correctly.

stefaniapedrazzi (approved):
Thank you

PR #3510, opened 2021-07-27
Feature enu geometries

BenjaminDeleze:
**Description**
Convert the geometries to be z-aligned

**Tasks**
  - [x] Convert cylinder
  - [x] Convert capsule
  - [x] Convert cone
  - [x] Convert plane
  - [x] Convert elevation grid
  - [x] Change doc
  - [x] Verify webotsJS (if you want to test it, you have to compile wren with Emscripten and change index.html and WebStreaming.js to point to the local files)
  - [ ] Adapt the tests

**Additional detail**

There is a script that can convert automatically worlds (scripts/converter/convert_geometries.py).
But it has some known limitations:

- It cannot convert protos
- If there is a DEF geometry that is use in a bounding object and there is already a transform in the bounding object, we need to convert it manually. Because the script add an additional transform to the geometry and become the DEF node, so we would have two Transform in the bounding object, which is forbiden.
- You need to convert by hand the following parameters of ElevationGrid:
  -  zDimension -> yDimension
  -  zSpacing -> ySpacing
  - Add a translation on the upper transform of value -(yDimension -1) * ySpacing either on the negative y (if ENU) or positive z (NUE)
  - Inverse the lines of the height 
  - Rotation the elevation grid by (1 0 0 -1.57)
- If a plane is scaled by a transform, you have to invert y and z scale. You also need to rotate it by (1 0 0 -1.57) (even if it does not have a scale)
- Sometimes the resulting rotation is weird when squashing transforms


BenjaminDeleze:
The script is more or less working to convert world.
I converted the floor.proto and yamor.proto as they are needed in the tests.
I did the api tests until display_infrared but camera_revert, display_infrared and camera_recognition are still not working.

lukicdarkoo:
Note that the Box documentation should be changed as well, e.g.:
> On the front (+z), back (-z), right (+x), and left (-x)...

BenjaminDeleze:
Three tests are still not passing/correct:

- `api/worlds/camera_revert.wbt`
  - The camera detects only the green and the blue box.
  - If you change the color of one box to red, it will detect it as red. So the box is the problem and not the color on it.
  - Very strange because the box geometry was not modified.
  - The camera detects the red box if this box is moved toward the robot by 0.01m.
  - It could be a problem of precision in the rotations but I tried to go in this direction without success. 
- `api/worlds/camera_recognition.wbt` and `api/worlds/radar_frustrum.wbt `
  - The capsule and cylinder (maybe also plane, elevation grid and cone) are wrongly detected by cameras and radars  
  - I think that there is only one error for both these tests.
  - I guess that the error partially come from #3571.
  - Part of the problem is solved by reverting the line 316 of `WbObjectDetection.cpp` to ```const WbMatrix3 rotation = deviceInverseRotation * objectRotation;```
  - The remaining error is maybe in the calculation of the frustumPlanes but I did not have time to investigate there.
-The image on the frustum of the camera is wrongly rotated.  I corrected it some time ago, but it is broken again (I guess because of the new camera)

ad-daniel:
The last test that fails for me is `tests/protos/worlds/template_proto_follow_solid`:
- The "follow solid" feature in of itself works (do `alt+1` before starting the simulation)
- However, if you compare the evolution of `Viewpoint.position` of this branch with that of `feature-enu-plan`, it can be noticed that this one doesn't move at the same pace, hence why the test fails (they're different from the very first time step)

So far I've discovered:
- If you replace the `boundingObject` of `TemplateRobotSlotContainer` from a `Box { size 0.1 0.1 0.1 }` to a `Sphere { radius 0.05 }` in both branches, the evolution of `Viewpoint.position` matches. So I suspect it must have something to do with the Box.
- When comparing the contact point between the two branches at time zero we have:

a line for the current branch

a

a triangle for the reference branch. This latter begins to move straight away, whereas the other tilts into the arena and needs a couple of timesteps to stabilize before also moving smoothly.

In short I suspect an approximation somewhere results in a different `handleInitialCollisions` state, but I haven't found where.
In theory the test can be made to pass simply by changing the threshold, as the "follow solid" logic it aims to test does work, but if the collisions are indeed different, it might be worth investigating and fixing it.

tests/api/worlds/display.wbt
```diff
@@ -19,7 +19,7 @@ Robot {
   children [
     Transform {
       translation 0 0 -1.1
-      rotation 1 0 0 1.5707997
+      rotation 1 0 0 0.000003373064713
```
lukicdarkoo, line 22:
```suggestion
```

src/webots/nodes/WbCapsule.cpp
```diff
@@ -36,11 +36,8 @@
 #include <ode/ode.h>
 #include <cmath>
 
+#include <iostream>
```
lukicdarkoo, line 39:
```suggestion
```

src/webots/nodes/WbElevationGrid.cpp
```diff
@@ -46,12 +46,13 @@ void WbElevationGrid::init() {
   mData = NULL;
   mMinHeight = 0;
   mMaxHeight = 0;
+  mIs90DegreesRotated = true;
```
lukicdarkoo, line 49: Why is it rotated?

lukicdarkoo:
> `api/worlds/camera_revert.wbt`
> * The camera detects only the green and the blue box.
> * If you change the color of one box to red, it will detect it as red. So the box is the problem and not the color on it.
> * Very strange because the box geometry was not modified.
> * The camera detects the red box if this box is moved toward the robot by 0.01m.
> * It could be a problem of precision in the rotations but I tried to go in this direction without success.

I spent so much time looking for the bug in the .cpp files. Actually, the wheel rotation was 1.57 (instead of 1.5707), and when converted to -1.5707 it wasn't actually rotated for pi, but with 0.007*2 difference. The color detection filter is too strict and that small difference made the problem.

lukicdarkoo:
> The last test that fails for me is `tests/protos/worlds/template_proto_follow_solid`:

I strongly believe this one is due to approximations errors in the Plane-Box collision and/or Transform+Plane. The Plane is rotated so it had to be rotated back in the Floor PROTO. Inside of the floor, instead of just `Plane` now we have `Transform (rotated for -pi/2) > Plane`. The extra Transform should be deleted once the Floor PROTO is converted to FLU.

If you agree I will adapt the test.

src/webots/nodes/WbElevationGrid.cpp
omichel: It seems to be justified by a difference between ODE and VRML in https://github.com/cyberbotics/webots/blob/feature-enu-geometries/src/webots/nodes/WbGeometry.hpp#L171
However, given that we are leaving the VRML world with Y vertical, we could also afford to adopt the ODE standard for elevation grids, remove this rotation and adapt our documentation and examples accordingly.

omichel (dismissed):
So far, it looks good to me.

omichel:
I just ran the test suite on Windows from this branch and got the following 4 failures:
```
### api test
[...]
FAILURE with motion_loop: Unexpected target position (error = 0.000398559)
FAILURE with motion_regular: Unexpected target position (error = 0.000657562)
FAILURE with motion_reverse: Unexpected target position (error = 0.000658488)
[...]
FAILURE with robot_multiple_step: Distance sensor measurement after robot moved is wrong (received: 666.773853).
[...]
```
These tests are passing on the master branch.

omichel:
The very same test failures also happened in the CI tests.

tests/api/worlds/display_infrared.wbt
```diff
@@ -22,8 +22,7 @@ Robot {
       rotation 0 1 0 1.5708
       name "ds0-0"
       lookupTable [
-        0 0 0
-        2 2 0
+        0 0 0        2 2 0
```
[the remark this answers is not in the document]
lukicdarkoo: This is probably the script thing, but since we will probably convert it again it shouldn't be a big issue

omichel (approved):

PR #2750, opened 2021-02-12
Fixed handling of UTF-8 paths for world files opened from Windows GUI

omichel:
Fixes  #2744.

PR #3512, opened 2021-07-27
Fixed start of Webots from Windows cmd.exe

### LINKED ISSUE #3485: Windows: cannot launch Webots R2021b from CMD
A user reported on Discord that with R2021b it is no longer possible to start Webots from CMD and can't get any output with argument `--version` or `--help` either.

Double clicking on the executable `${WEBOTS_HOME}/msys64/mingw64/bin/webotsw.exe` works correctly.

I quickly checked and I get the same issue on my Windows (user without development system installed) after installing Webots R2021b for all the users.

Here is the related documentation:
https://www.cyberbotics.com/doc/guide/starting-webots#windows

omichel:
Fixes #3485. I also checked that this patch doesn't break the resolution of #2750.


lukicdarkoo (approved):
Thank you.

I confirm that the Windows package generated in CI works fine for me.

PR #3518, opened 2021-07-28
Connector to FLU

mael25:
**Description**
Switch the `Connector`  axes in order to meet FLU convention.
x-->y
y-->z
z-->x

**Notes**
Had to update `Yamor.proto` (which contains two `Connector`) as one of the API tests uses it.


BenjaminDeleze (approved):
Thank you!

tests/physics/worlds/connector_static_autolock.wbt
```diff
@@ -66,7 +67,7 @@ DEF SOLID Solid {
     }
     Connector {
       translation 0.05 0 0
-      rotation 0 1 0 1.5707963220000003
+      rotation 0.9999999999944944 -2.3464099999870814e-06 -2.3464099999870814e-06 -1.5707953071795862
```
omichel, line 70: Can't we have simpler values here, like:
```
rotation 1 0 0 -1.57079
```

PR #3520, opened 2021-07-28
Fix robotbenchmark robot windows

### LINKED ISSUE #3194: Issues in robotbenchmark robot windows when executed from Webots desktop
Info windows in robotbenchmark simulations have some issue when they are run from Webots desktop application.
Not sure if there is a simple way to fix them so that they works both online and on the desktop application.

* In the visual tracking simulation, vectors in formulas are not displayed correctly.
* links to documentation pages doesn't work.
    With a normal click nothing happens.
    With CTRL+click the documentation page opens instead of the robot windows and this should not happen.
    -> even with a single click it would be expected that they will open externally in a browser.
* Code block alignment is wrong (for example in pick place, inverted pendulum and others.

stefaniapedrazzi:
Fix #3194:
* [x] fix opening external links wit QtWebEngine
  * [x] check and fix QtWebKit (Windows)
* [x] fix using MathML with Chrome
* [x] fix code block indentation

### DIFF
--- a/projects/samples/robotbenchmark/humanoid_marathon/plugins/robot_windows/humanoid_marathon/humanoid_marathon.html
+++ b/projects/samples/robotbenchmark/humanoid_marathon/plugins/robot_windows/humanoid_marathon/humanoid_marathon.html
@@ -40,11 +40,11 @@
   # At the beginning, start walking on the spot.
   # After 45 timesteps, begin taking steps forward.
   while robot.step(timestep) != -1:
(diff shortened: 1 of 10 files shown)


stefaniapedrazzi:
I tested on Ubuntu and Windows and the issues are now fixed.

omichel (approved):
Thank you.

omichel:
Can we merge it now?

Issue #3523, opened 2021-07-29
mouse shift+move wrong direction if within rotation


Thomas-Feldmeier:
world with coordinate system ENU (maybe not required, have not tested it)

add a transform with a rotation, i used this one but i think any would do:
```
#VRML_SIM R2021b utf8
# rotates things which were build with NUE coordinate system (UP=y coordinate)
# to the ENU coordinate system (UP=z coordinate) if needed

PROTO ConvertCoord [
  field MFNode      children  []
]
{
  Transform {
%{ if context.coordinate_system == "ENU" then }%
    rotation 1 0 0 1.5708
%{ end }%
    children IS children
  }
}
```

now add some solid to the children of that transform (e.g. PottedTree)
press shift key and click on the solid to and move mouse to move the solid
now one direction works, but the other one not

maybe issue #158 is related with this one

mael25:
It is because when doing the Mouse+Shift click, it will change the `translation` field of the `Solid` and not the one of the `Transform` (regardless to the fact that they are defined within a Proto). Thus, although the translation of the `Solid` is modified as requested, the resulting movement appears strange because it is applied within the rotated frame of the `Transform`. You can observe it by right-clicking on the Proto --> Convert to Base Node(s).

Thomas-Feldmeier:
The Solid should move in the same direction of the mouse. As the coordinate system of the Solid can be rotated relative to the world coordinate system, the movement of the mouse has to be converted from the world coordinate system to the coordinate system of the Solid. In wb_supervisor_node_get_pose() such an transformation is done. Make a pose matrix with the mouse coordinate change between two updates as translation, rotation matrix should be identity matrix. Transform that matrix into the coordinate system of the Solid. Get the translation out of the new matrix, change solid position by this offset.

Thomas-Feldmeier:
maybe you need to invert the pose matrix [link](https://math.stackexchange.com/questions/152462/inverse-of-transformation-matrix)

Thomas-Feldmeier:
strg+mouse movement works correct (adding a force), that code could be used to fix shift+mouse movement.

Issue #158, opened 2019-02-06
Inverted direction when moving robot with SHIFT key in hoap2_sumo.wbt


stefaniapedrazzi:
**Describe the Bug**
In ``hoap2_sumo.wbt`` world, when moving the robot with SHIFT+mouse left button and drag, the robot is moved in the wrong direction w.r.t to the mouse movement: left instead of right and front instead of back.
For all the other robots and world the behavior is correct.

**Steps to Reproduce**
1. Go to 'hoap2_sumo.wbt'
2. move the Hoap2 robot with SHIFT+mouse left click+mouse drag

**Expected behavior**
The robot should be moved in the same direction as the mouse pointer.

**System**
 - Operating System: Linux Ubuntu 18.04
 - Graphics Card: NVIDIA GeForce GT 740

Note that there is also some strange behavior moving the robot in `robotis_op3_motion_player.wbt` with SHIFT+mouse, but this could be due to the size of the world.

fabienrohrer:
I cannot reproduce this bug on macOS or on my Ubuntu18.04 virtual machine.

DavidMansolino:
I can reproduce it on Ubuntu 16.04 native machine but not systematically most of the time it is working fine.

stefaniapedrazzi:
I can reproduce this systematically. But it depends on the view and on where I click with the mouse to start the rotation.

PR #3530, opened 2021-08-02
Hotfix shift+left drag

mael25:
**Description**
When doing Shift+Mouse(left) drag on a Solid, Webots currently assumes that it has no parent Transform and is a child of root. Therefore, if it does have a parent Transform, the Solid is wrongly moved (#3523).

### DIFF
--- a/src/webots/gui/WbDragTransformEvent.cpp
+++ b/src/webots/gui/WbDragTransformEvent.cpp
@@ -16,6 +16,7 @@
 
 #include "WbAbstractTransform.hpp"
 #include "WbEditCommand.hpp"
+#include "WbLog.hpp"
 #include "WbMathsUtilities.hpp"
 #include "WbSolid.hpp"
 #include "WbStandardPaths.hpp"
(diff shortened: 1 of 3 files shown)


stefaniapedrazzi:
Maybe this also solves #158

mael25:
> Maybe this also solves #158

It doesn't solve it yet, but I am going to try and fix this as well.

mael25:
@stefaniapedrazzi I can reproduce the #158 bug only if I drag the `Hoap` robot while having a `Viewpoint` orientation parallel to the drag plane (for instance front view), and if the destination of the drag is situated in the lower section of the world (y < 0)...

In these particular cases, it is difficult to determine the intersection between the mouse ray and the drag plane along the "depth" ( here z) axis. We could prevent the drag event to occur if the `Viewpoint` orientation is too much parallel to the drag plane and display a warning to the user.

Let me know if you can reproduce the bug in other situations.

stefaniapedrazzi:
> @stefaniapedrazzi I can reproduce the #158 bug only if I drag the `Hoap` robot while having a `Viewpoint` orientation parallel to the drag plane (for instance front view), and if the destination of the drag is situated in the lower section of the world (y < 0)...
> 
> In these particular cases, it is difficult to determine the intersection between the mouse ray and the drag plane along the "depth" (here z) axis. We could prevent the drag event to occur if the `Viewpoint` orientation is too much parallel to the drag plane and display a warning to the user.
> 
> Let me know if you can reproduce the bug in other situations.

Yes, the problematic case is when the Viewpoint plane is parallel to the drag plane (as in the initial Viewpoint pose of `hoap2_sumo.wbt`).
As you reported, the direction is wrong on the lower section of the world but the behavior is not optimal also in the upper section because it is very small.

If it is not possible to fix the behavior, s warning and/or preventing the move seems better that applying the wrong drag.

mael25:
Now it verifies that the horizontal plane is  visible enough when doing a shift+left drag event. For this I do not use the orientation of the Viewpoint, but instead the mouse ray direction. This is because the behavior depended on the location of the click within the object (especially for tall ones), as the resulting intersection with the drag horizontal plane is different.

Thus, it uses the scalar product between the mouse ray direction and the _up_ vector to move the object only if the angle ray-plane is larger than 15deg. Otherwise there is a warning.

src/webots/gui/WbDragTransformEvent.cpp
```diff
@@ -80,16 +83,41 @@ WbDragHorizontalEvent::~WbDragHorizontalEvent() {
...
-  mSelectedTransform->emitTranslationOrRotationChangedByUser();
+  if (!mIsInitializationDone) {
+    WbRay normalizedMouseRay = mMouseRay;
+    normalizedMouseRay.normalize();
+    // event occurs only if the mouse ray is not parallel to the horizontal drag plane
+    if (abs(normalizedMouseRay.direction().dot(mUpWorldVector)) > DRAG_HORIZONTAL_MIN_COS)
+      mIsMouseRayValid = true;
+    else {
+      mIsMouseRayValid = false;
+      WbLog::warning(tr("To drag this element, rotate first the view so that the horizontal plane is clearly visible."));
```
stefaniapedrazzi, line 94: Maybe this is better
```suggestion
      WbLog::warning(tr("To drag this element, first rotate the view so that the horizontal plane is clearly visible."));
```

stefaniapedrazzi (changes requested):
Thank you!

I tested and both cases are now working better. I just have a couple of minor comments.

mael25:
Thank you for the review, I have moved all the code executed only once to the constructor.

PR #3531, opened 2021-08-02
Documentation improve wb_supervisor_node_set_visibility

### LINKED ISSUE #3529: `wb_supervisor_node_set_visibility` also hides sibling nodes
In case of such a structure:
```
Solid {
  children [
    Shape A {
      ..
    }
    Shape B {
      ..
    }
  ]
}
```

Hiding the node A from a camera will also hide node B.
But if I add a `Transform` parent to the `Shape` that should be hidden then works correctly.

I will inspect if this is a limitation of the functionality, but in this case it should be clearly specified in the documentation that currently simply says that children are recursively hiddden too:
https://www.cyberbotics.com/doc/reference/supervisor#wb_supervisor_node_set_visibility

stefaniapedrazzi:
Fix #3529:
add additional specifications to the `wb_supervisor_node_set_visibility` functionality that depending on the passed node type also disables the sibling nodes.


ad-daniel:
Concerning [this](https://github.com/cyberbotics/webots/issues/3529#issuecomment-891135987) comment, will it be addressed in this PR or in a separate one?

stefaniapedrazzi:
Replaced by #3543.

PR #3532, opened 2021-08-03
Add note about coupled-motor usage in differentials

ShuffleWire:
Mechanical differentials drive, as described in https://en.wikipedia.org/wiki/Differential_(mechanical_device) could be very easily simulated using coupled motor. As I just figured out, even if the main role of it is to change speed to avoid slip, it also ensure that each wheel receive the **same** torque (so, for a regular car, each 2 (lets say front) wheel receive half the motor torque). That fact is often overlooked, seen as a drawback for many application (if a wheel slip, the other stop, cause no torque), but for simulation it actually awesome !
Been trying it on my project with a 4*4 configuration, and work really fine !

This is why that fact could be added in the doc.
However I do feel that my explanation is not very clear, English not been my mother-thong... feel free to update !

### DIFF
--- a/docs/reference/motor.md
+++ b/docs/reference/motor.md
@@ -230,7 +230,7 @@ Using `multiplier` here also makes sense, because some differentials don't split
 This parameter doesn't depend on the actual speed and characteristics of the wheels, it's only a mechanical setting.
 
 > **Note**: Although any among the coupled motors can be controlled, commands should be given to just one among them at any given time in order to avoid confusion or conflicts.
-For instance, it isn't possible to do Position Control for one motor and Velocity Control another at the same time.
+For instance, it is not possible to do Position Control for one motor and Velocity Control another at the same time.
 Whatever command is given to a motor, it is relayed to all of its siblings hence overwriting any prior settings imposed on them.
 In other words, only the last command given is the one actually being enforced across the coupling.
 
--- a/docs/reference/motor.md
+++ b/docs/reference/motor.md
@@ -221,8 +221,13 @@ Although each sibling motor receives the same command, what the motors actually 
 
 > **Note**: The motors are *logically* coupled together, not *mechanically*.
 If one of the motors is physically blocked, the others are in no way affected by it.
-This provide a very nice side-effect : when used in force-control mode, coupled-motor allows to easely simulate a mechanical differential. The role of a differential is to change the speed of the wheel relatively to each other. But also, it split *equally* the motor torque to each wheel. This is why, It's only needed to apply the same torque on multiple (coupled) motor, and the physic engine will figure out speeds. It works for regular car as well as for 4*4 vehicule, as long as they have 3 differencial (front, rear, and central).
-Using `multiplier` here also make sense, because some diferential don't split in half (sometime central differential split 40%-60% to get more torque to the rear wheels). But this parameter doesn't depend on the actual speed, environnement etc... of the wheels, it's only a mecanical setting.
+This provide a useful side-effect: when used in force-control mode, coupled-motor allows to easily simulate a mechanical differential.
+The role of a differential is to change the speed of the wheel relatively to each other.
+But also, it splits *equally* the motor torque to each wheel.
+This is why, it's only needed to apply the same torque on multiple (coupled) motor and the physic engine will adapt the speeds accordingly.
+It works for regular car as well as for a 4x4 vehicule, as long as they have 3 differential (front, rear, and central).
+Using `multiplier` here also make sense, because some differentials don't split in half: sometimes a central differential splits 40%-60% to get more torque to the rear wheels.
+This parameter doesn't depend on the actual speed and characteristics of the wheels, it's only a mecanical setting.
 
 > **Note**: Although any among the coupled motors can be controlled, commands should be given to just one among them at any given time in order to avoid confusion or conflicts.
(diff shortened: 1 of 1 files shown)


ad-daniel:
Yes you're right, it could be an interesting application. Perhaps instead of simply citing it in the docs, it might be better to instead create a sample world that showcases the differential functionality, and add it either to the [demos](https://cyberbotics.com/doc/guide/samples-demos) or [how-to](https://cyberbotics.com/doc/guide/samples-howto). The [devices](https://cyberbotics.com/doc/guide/samples-devices) samples already has an example for a gripper, but it's relatively simple (as it should be), so these other two options might be better suited for it. 

Just throwing the idea

ShuffleWire:
That would be a good way to create my first sample world, but I don't know I'll find time to do it immediately... I'll think about about it, and keep you updated :) Moreover I will need sometimes a bit of help to make this going in the right direction.

docs/reference/motor.md
```diff
@@ -221,6 +221,13 @@ Although each sibling motor receives the same command, what the motors actually
 
 > **Note**: The motors are *logically* coupled together, not *mechanically*.
 If one of the motors is physically blocked, the others are in no way affected by it.
+This provides a useful side-effect: when used in force-control mode, coupled-motor allows to easily simulate a mechanical differential.
```
ad-daniel, line 224:
```suggestion
This provides a useful side-effect: when used in force-control mode, coupled-motors allow to easily simulate a mechanical differential.
```

docs/reference/motor.md
```diff
@@ -221,6 +221,13 @@ Although each sibling motor receives the same command, what the motors actually
...
+The role of a differential is to change the speed of the wheel relatively to each other.
+But also, it splits *equally* the motor torque to each wheel.
+This is why it's only needed to apply the same torque on multiple (coupled) motors, and the physic engine will adapt the speeds accordingly.
+It works for a regular car as well as for a 4x4 vehicle, as long as they have 3 differentials (front, rear, and central).
+Using `multiplier` here also makes sense, because some differentials don't split in half: sometimes a central differential splits 40%-60% to get more torque to the rear wheels.
+This parameter doesn't depend on the actual speed and characteristics of the wheels, it's only a mechanical setting.
 
 > **Note**: Although any among the coupled motors can be controlled, commands should be given to just one among them at any given time in order to avoid confusion or conflicts.
 For instance, it isn't possible to do Position Control for one motor and Velocity Control another at the same time.
```
ad-daniel, line 233:
```suggestion
For instance, it is not possible to do Position Control for one motor and Velocity Control another at the same time.
```

docs/reference/motor.md
```diff
@@ -221,6 +221,13 @@ Although each sibling motor receives the same command, what the motors actually
 
 > **Note**: The motors are *logically* coupled together, not *mechanically*.
 If one of the motors is physically blocked, the others are in no way affected by it.
+This provides a useful side-effect: when used in force-control mode, coupled-motor allows to easily simulate a mechanical differential.
+The role of a differential is to change the speed of the wheel relatively to each other.
+But also, it splits *equally* the motor torque to each wheel.
+This is why it's only needed to apply the same torque on multiple (coupled) motors, and the physic engine will adapt the speeds accordingly.
+It works for a regular car as well as for a 4x4 vehicle, as long as they have 3 differentials (front, rear, and central).
+Using `multiplier` here also makes sense, because some differentials don't split in half: sometimes a central differential splits 40%-60% to get more torque to the rear wheels.
```
ad-daniel, line 229:
```suggestion
Using `multiplier` here also makes sense, because some differentials do not split in half: sometimes a central differential splits 40%-60% to get more torque to the rear wheels.
```

docs/reference/motor.md
```diff
@@ -221,6 +221,13 @@ Although each sibling motor receives the same command, what the motors actually
 
 > **Note**: The motors are *logically* coupled together, not *mechanically*.
 If one of the motors is physically blocked, the others are in no way affected by it.
+This provides a useful side-effect: when used in force-control mode, coupled-motor allows to easily simulate a mechanical differential.
+The role of a differential is to change the speed of the wheel relatively to each other.
+But also, it splits *equally* the motor torque to each wheel.
+This is why it's only needed to apply the same torque on multiple (coupled) motors, and the physic engine will adapt the speeds accordingly.
```
ad-daniel, line 227:
```suggestion
Therefore it suffices to apply the same torque on multiple (coupled) motors, and the physics engine will adapt the speeds accordingly.
```

docs/reference/motor.md
ad-daniel: Since we're removing contractions, might as well remove this one.

docs/reference/motor.md
ad-daniel: No contractions, ideally.

ad-daniel (changes requested):
Some minor changes.

ad-daniel (approved):

ad-daniel:
Thank you.

ShuffleWire:
@ad-daniel did you want me to merge it now, and I open a new PR for sample world when ready, or should I kept it open ?

ad-daniel:
I think we can merge this already, and concerning the demo/how-to you can open an issue about it, so that it doesn't get lost/forgotten if neither you nor us have the time to look into it at the moment.
These details added to the docs can always be modified in that PR, if needed.

Thank you

PR #3536, opened 2021-08-04
Add a figure for roll/tilt/pan axes

omichel:
This figure makes it clear what is roll, tilt/pitch and pan/yaw axes with respect to the X, Y and Z axes.
See https://cyberbotics.com/doc/reference/camera?version=add-roll-tilt-pan-figure


ad-daniel (approved):
I assume no attribution is needed?

Thank you.

omichel:
> I assume no attribution is needed?

I don't think so.

PR #3538, opened 2021-08-04
Convert the Viewpoint to FLU

lukicdarkoo:
**Documentation**
https://www.cyberbotics.com/doc/reference/viewpoint?version=feature-viewpoint-flu

**Tasks**:
- [x] Change the Viewpoint orientation to FLU
- [x] Fix mouse zoom
- [x] Fix mouse rotate
- [x] Fix mouse translate
- [x] Fix object translate
- [x] Fix object rotate
- [x] Fix apply force with mouse
- [x] Fix Billboard (only docs)
- [x] Fix `Change View > Front View`, ...
- [x] Fix Shift + move
- [x] Fix docs
- [x] Fix the coordinate system legend


src/webots/wren/WbCoordinateSystem.cpp
```diff
@@ -167,10 +167,11 @@ void WbCoordinateSystem::setVisible(bool b) {
 
 void WbCoordinateSystem::setOrientation(const WbQuaternion &quaternion) {
   float rotation[4];
-  WbRotation(quaternion).toFloatArray(rotation);
+  WbQuaternion adaptedQuaternion(WbQuaternion(0.5, -0.5, 0.5, 0.5) * quaternion);
```
lukicdarkoo, line 170: I am not sure whether there is a cleaner way

src/webots/nodes/WbViewpoint.cpp
```diff
@@ -1267,27 +1268,27 @@ bool WbViewpoint::moveViewpointToObject(WbBaseNode *node) {
...
 }
 
 void WbViewpoint::topView() {
-  orbitTo(WbVector3(0, 1, 0), WbRotation(1, 0, 0, -M_PI / 2));
+  orbitTo(WbVector3(0, 0, 1), WbRotation(-0.5773, 0.5773, 0.5773, 2.0944));
 }
 
 void WbViewpoint::bottomView() {
-  orbitTo(WbVector3(0, -1, 0), WbRotation(0, -0.707107, 0.707107, M_PI));
+  orbitTo(WbVector3(0, 0, -1), WbRotation(0.5773, -0.5773, 0.5773, 2.0944));
```
ad-daniel, line 1291: I think this one is wrong. When comparing it to how it behaves on master (using the `kid.wbt` world). On master, the red team is on the left side, in this PR it's on the right side.

src/webots/wren/WbCoordinateSystem.cpp
ad-daniel: Can't think of any either.

ad-daniel (commented):
Looks good, only have one doubt.

By the way, it's unrelated to this PR but is there a reason why the shortcuts to change view skip `alt + 3` and `alt + 5`, seems very counterintuitive.

src/webots/nodes/WbViewpoint.cpp
lukicdarkoo: Oh, you are right, it should be fixed now

src/webots/wren/WbCoordinateSystem.cpp
ad-daniel: Can't think of a better way either.

ad-daniel (approved):
Thank you.

lukicdarkoo:
> By the way, it's unrelated to this PR but is there a reason why the shortcuts to change view skip alt + 3 and alt + 5? seems very counterintuitive.

Agree, currently:
- ALT+5: Move Viewpoint to Object
- ALT+3: Seems to not be doing anything

resources/nodes/Viewpoint.wrl
```diff
@@ -2,7 +2,7 @@
 
 Viewpoint {
   vrmlField SFFloat                                                                 fieldOfView            0.785398  # (rad)
-  vrmlField SFRotation                                                              orientation            0 0 1 0
+  vrmlField SFRotation                                                              orientation            -0.5773 0.5773 0.5773 2.0944
```
omichel, line 5: That is not consistent with the documentation page.

omichel: This odd rotation value doesn't look very nice. Is it to make it oriented exactly as previously? Can't we afford a nicer rotation here?

lukicdarkoo: > Is it to make it oriented exactly as previously?

No, I didn't like how it was before: x-east, y-up, and z-south (EUS). Now it is x-east, y-north, and z-up (ENU).

> Can't we afford a nicer rotation here?

No. If we want to keep a nice default orientation we have to use this one. It is not an implementation workaround, but ENU (world) - FLU (camera) workaround.

omichel: As discussed, it might be better (although a bit arbitrary) to keep the 0 0 1 0 default orientation (as it is done in Gazebo and VRML97) to have the viewpoint from a human perspective rather than from a bird of prey perspective.

omichel: And also this saves us from having long floating point values in the default orientation.

omichel: But the nice default XY view from the top is also very convenient... So, let's keep it for now...

omichel: In any case, we should improve the documentation about the Viewpoint, as it seems quite unclear right now.

PR #3540, opened 2021-08-04
Implement more fluid object rotation

lukicdarkoo:
Sometimes, when the viewpoint is not aligned with the object's axis of rotation, it is hard to rotate the object. This new object rotation mechanism follows the one in Blender.

Let me know whether you like it (and I will proceed with the implementation)?


ShuffleWire:
That is so nice !

stefaniapedrazzi:
Shouldn't the new black line start from the center of the object (like in Blender) instead of the point when the user clicks?
Given that the rotation center is in any case the center of the object I think it would make more sense.

docs/reference/changelog-r2021.md
```diff
@@ -5,6 +5,7 @@ Released on XX, XXth, 2021.
 
   - Enhancements
     - Improved performance on [Lidar](lidar.md) point cloud generation ([#3499](https://github.com/cyberbotics/webots/pull/3499)).
+    - Improved UX when using the object rotation around axis feature ([#3540](https://github.com/cyberbotics/webots/pull/3540)).
```
stefaniapedrazzi, line 8: I'm not sure that users that will read the change log will understand what is UX.

src/webots/wren/WbTranslateRotateManipulator.cpp
```diff
@@ -263,6 +320,62 @@ void WbTranslateRotateManipulator::showNormal() {
...
+
+  // Double arrow
+  tail[1] += scale[1];
+  wr_transform_set_position(mRotationDoubleArrowTransform, tail + 3);
+
+  // Orientation
+  WbVector3 baseX, baseY, baseZ;
+  baseY = (end - begin).normalized();
+
+  // Check if the vector is parallel to the xy-plan of the camera (dot product close to zero)
```
stefaniapedrazzi, line 361:
```suggestion
  // Check if the vector is parallel to the xy-plane of the camera (dot product close to zero)
```

stefaniapedrazzi (changes requested):
Apart from the comment on the starting point of the new black arrow, the behavior seems fine.

lukicdarkoo:
> Shouldn't the new black line start from the center of the object (like in Blender) instead of the point when the user clicks?
> Given that the rotation center is in any case the center of the object I think it would make more sense.

That's a big question :) In Blender, there are no curved arrow widgets, so it is a bit different. I am afraid it would be unintuitive to click on the arrow widget and the center of rotation jumps to the object.

I will implement as you suggested and we can compare

lukicdarkoo:
@stefaniapedrazzi I definitely like more your suggestion. Let me know if it is better now

stefaniapedrazzi:
Thank you! I think it is better like this.
If you think that it is still not very intuitive we could also show the other two axis (short lines same as when showing the coordinate system) to better mark the center of the object.
Then if we have this new black arrow that is rotating we may evaluate if it still makes sense to also rotate the round arrow.

There is just one issue with this PR that is if the viewpoint is located on the negative side of the rotation axis, the resulting rotation is inverted with respect to the mouse drag and to the expected rotation.

lukicdarkoo:
> If you think that it is still not very intuitive we could also show the other two axis (short lines same as when showing the coordinate system) to better mark the center of the object.

Implemented

> Then if we have this new black arrow that is rotating we may evaluate if it still makes sense to also rotate the round arrow.

I believe that the arrows are still indicators, they clearly indicate the axis of rotation.

> There is just one issue with this PR that is if the viewpoint is located on the negative side of the rotation axis, the resulting rotation is inverted with respect to the mouse drag.

As soon as GitHub is fully functional I will push the update

stefaniapedrazzi:
Is this PR waiting for a new review?

lukicdarkoo:
> Is this PR waiting for a new review?
> I didn't receive any review request.

It is ready for review

stefaniapedrazzi (approved):
Thank you

PR #3543, opened 2021-08-05
Fix supervisor node set visibility

### LINKED ISSUE #3529: `wb_supervisor_node_set_visibility` also hides sibling nodes

stefaniapedrazzi:
Fix #3529:
if given node doesn't have a dedicated Wren node, then apply visibility to closest descendants with dedicated Wren node.

Only `WbBillboard`, `WbGeometry`, `WbTransform` and `WbMuscle` have a dedicated Wren node.
For all the other nodes we have to retrieve the descendant nodes:

Nodes overriding the function to retrieve closes descendant nodes with dedicated Wren node:
|Node | Function Overridden | Comments |
| ------ | --------- | -------- |
| BasicJoint | x  | |
| Billboard | x | |
| Geometry | x | |
| Group | x | |
| Motor | x | |
| Muscle  | x | Users can override the visibility properties in the main view switching the `Muscle.isVisible` flag |
| Propeller | x | Webots code overrides the visibility properties when enabling/disabling the slow/fast helix or is the fast helix is selected from the scene tree |
| Shape | x | |
| Slot | x | |
| SolidReference | x | |
| Transform | x | |
| Background | | This function cannot be used to hide background from cameras | 
| Fog | | This function cannot be used to hide fog from cameras |

Nodes that doesn't have any associated Wren nodes:
Appearance, Background, Brake, Color, ContactProperties, Coordinate, Damping, Focus, Fog, ImageTexture, ImmersionProperties, JointParameters, Lens, LensFlare, Light, Material, Normal, PBRAppearance, Physics, Recognition, TextureCoordinate, TextureTransform, Viewpoint, WorldInfo, Zoom

Remaining nodes are derived from other nodes already listed.

### DIFF
--- a/tests/api/controllers/supervisor_node_set_visibility/supervisor_node_set_visibility.c
+++ b/tests/api/controllers/supervisor_node_set_visibility/supervisor_node_set_visibility.c
@@ -31,17 +31,19 @@ int main(int argc, char **argv) {
   quick_assert_color(camera1, 31, 31, 0xFF0000, "Both cameras should see the red box at the beginning.\n");
 
   WbNodeRef camera_node = wb_supervisor_node_get_from_def("CAMERA0");
-  WbNodeRef box_node = wb_supervisor_node_get_from_def("BOX");
+  WbNodeRef camera1_node = wb_supervisor_node_get_from_def("CAMERA1");
+  WbNodeRef red_box_node = wb_supervisor_node_get_from_def("RED_BOX");
+  WbNodeRef green_box_node = wb_supervisor_node_get_from_def("GREEN_BOX");
 
-  wb_supervisor_node_set_visibility(box_node, camera_node, false);
+  wb_supervisor_node_set_visibility(red_box_node, camera_node, false);
 
   wb_robot_step(TIME_STEP);
 
-  quick_assert_color(camera0, 31, 31, 0x0000FF,
+  quick_assert_color(camera0, 31, 31, 0x00FF00,
                      "Camera 0 should not see the red box after it is been hide from this camera.\n");
   quick_assert_color(camera1, 31, 31, 0xFF0000, "Camera 1 should still see the red box after it was hidden from camera 0.\n");
 
-  wb_supervisor_node_set_visibility(box_node, camera_node, true);
+  wb_supervisor_node_set_visibility(red_box_node, camera_node, true);
 
   wb_robot_step(TIME_STEP);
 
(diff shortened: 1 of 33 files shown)


stefaniapedrazzi:
I fixed the set visibility function so that it is never applied to parent and sibling nodes.
However the `wb_supervisor_node_set_visibility` function has (same as previous version) some limitations that could only be overcome by adding additional mechanisms. These are listed in the table in PR description.
But I'm not sure it is worth adding these additional mechanisms now or wait until we have use cases that requires the resolution.

omichel:
I believe we should simply document these limitations in the reference manual for now as I believe they are not a problem for most use cases.

stefaniapedrazzi:
I has to fix conflicts in the ChangeLog.

tests/api/controllers/supervisor_node_set_visibility/supervisor_node_set_visibility.c
```diff
@@ -31,17 +31,19 @@ int main(int argc, char **argv) {
...
+  WbNodeRef green_box_node = wb_supervisor_node_get_from_def("GREEN_BOX");
 
-  wb_supervisor_node_set_visibility(box_node, camera_node, false);
+  wb_supervisor_node_set_visibility(red_box_node, camera_node, false);
 
   wb_robot_step(TIME_STEP);
 
-  quick_assert_color(camera0, 31, 31, 0x0000FF,
+  quick_assert_color(camera0, 31, 31, 0x00FF00,
                      "Camera 0 should not see the red box after it is been hide from this camera.\n");
```
ad-daniel, line 43:
```suggestion
                     "Camera 0 should not see the red box after it was hidden from this camera.\n");
```

tests/api/worlds/supervisor_node_set_visibility.wbt
```diff
@@ -37,12 +47,12 @@ DEF Test Robot {
     TestSuiteEmitter {
     }
     DEF CAMERA0 Camera {
-      translation 0 0.3 0
+      translation 0 0.84 5.50586e-07
```
ad-daniel, line 50:
```suggestion
      translation 0 0.84 0
```

tests/api/worlds/supervisor_node_set_visibility.wbt
```diff
@@ -37,12 +47,12 @@ DEF Test Robot {
...
       rotation 1 0 0 4.71239
       name "camera0"
     }
     DEF CAMERA1 Camera {
-      translation 0 0.3 0
+      translation 0 0.84 5.50586e-07
```
ad-daniel, line 55:
```suggestion
      translation 0 0.84 0
```

ad-daniel (changes requested):
Looks good, only very minor details

ad-daniel (dismissed):

ad-daniel:
Thank you.

lukicdarkoo (approved):

Issue #1014, opened 2019-10-17
Incomplete integrity tests of the dependencies


fabienrohrer:
Recently, we introduced MD5 checks for the Webots dependencies.

However there are other dependencies to check, mainly in the `projects` directory, such as sumo here:
https://github.com/cyberbotics/webots/blob/revision/projects/default/resources/Makefile

DavidMansolino:
Furthermore, it should be feasible to remove the temporary md5 file on Windows and Linux as it is already done on mac with something like this: 
```diff
diff --git a/dependencies/Makefile.linux b/dependencies/Makefile.linux
index 879c027..bd7dcf6 100644
--- a/dependencies/Makefile.linux
+++ b/dependencies/Makefile.linux
@@ -40,15 +40,13 @@ qt: $(WEBOTS_HOME)/lib/libQt5Core.so.$(QT_VERSION)
 
 $(WEBOTS_HOME)/lib/libQt5Core.so.$(QT_VERSION): $(WEBOTS_DEPENDENCY_PATH)/$(QT_PACKAGE)
        @echo "# uncompressing $(QT_PACKAGE)"
+       @if [ $(firstword $(shell md5sum $(QT_PACKAGE))) != "e004cc3d47fd359131fb931303214c9d" ]; then echo "MD5 checksum failed for $(QT_PACKAGE)"; exit 1; fi
        @tar xfm $(WEBOTS_DEPENDENCY_PATH)/$(QT_PACKAGE) -C $(WEBOTS_HOME)
 
 $(WEBOTS_DEPENDENCY_PATH)/$(QT_PACKAGE):
        @rm -f $(WEBOTS_DEPENDENCY_PATH)/$(QT_PACKAGE)
        @echo "# downloading $(QT_PACKAGE)"
        @wget -qq $(DEPENDENCIES_URL)/$(QT_PACKAGE) -P $(WEBOTS_DEPENDENCY_PATH)
-       @echo "e004cc3d47fd359131fb931303214c9d  $(QT_PACKAGE)" > $(QT_PACKAGE).md5
-       @md5sum -c --status $(QT_PACKAGE).md5 ; if [ $$? -ne 0 ]; then echo "MD5 checksum failed for $(QT_PACKAGE)"; exit 1; fi
-       @rm $(QT_PACKAGE).md5
        @touch $(WEBOTS_DEPENDENCY_PATH)/$(QT_PACKAGE)
```

PR #3551, opened 2021-08-09
Complete integrity tests for downloaded resources

stefaniapedrazzi:
Address #1014:

* [x] Linux
* [x] Windows
* [x] macOS


lukicdarkoo (approved):
Thank you

We should not forget to delete the `libssl_1.1.tar.xz` from the server.

stefaniapedrazzi:
> We should not forget to delete the `libssl_1.1.tar.xz` from the server.

On the server there are still all the old libraries (even for Ubuntu 14.04).
I will move all the old libraries in a new folder so that we know they are no longer used but they are still accessible if needed.

PR #3558, opened 2021-08-10
Hotfix inaccessible points in `getPointCloud` python API

ad-daniel:
**Description**
As reported by a user on discord, currently the data retrieved using `getPointCloud` python API is no longer accessible.

The change done in https://github.com/cyberbotics/webots/pull/3322 broke the API. Although at the time I'm sure it prevented the crash of the `enable_all_lidars.py` controller from happening ( in `sumo_interface_example.wbt`, mentioned here https://github.com/cyberbotics/webots/issues/3281), now even after reverting the change the controller doesn't appear to crash any longer. So it was either something else provoking it to begin with, or I don't know.

It can be tested using this world: [WebotsProject.zip](https://github.com/cyberbotics/webots/files/6961390/WebotsProject.zip)


BenjaminDeleze (approved):
Seems good to me

PR #3561, opened 2021-08-11
Fix ROS lidar crash

lukicdarkoo:
As reported on Discord, the ROS lidar crashes after it is enabled.

- dab7ac5ba41d: Fixes crash when the `--auto-publish` argument is used. It seems that the buffer is not available in the beginning, it might be related to the recent change of the lidar.
- 8aeab030c09: The sensor used to be enabled after it is added to the list. Since everything is happening in a callback (another thread) there was a warning saying the sensor is used before enabled (showing the error 1-2 times in the beginning).
- 999edb8389fd: Fixes the ros_control crash when a motor doesn't have an associated position sensor.


BenjaminDeleze (approved):
Thank you

PR #3562, opened 2021-08-11
Fix the Viewpoint documentation

lukicdarkoo:
Fixes https://github.com/cyberbotics/webots/pull/3538#discussion_r686682150

This default orientation is ENU friendly, it doesn't work nicely with NUE. Therefore, when creating a new project it is a bit weird but it should be fine once `NUE` is replaced in the `empty.wbt` world and the arena PROTO is converted.

### DIFF
--- a/docs/reference/viewpoint.md
+++ b/docs/reference/viewpoint.md
@@ -2,19 +2,19 @@
 
 ```
 Viewpoint {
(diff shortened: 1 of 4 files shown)


omichel (approved):
Thank you.

PR #3565, opened 2021-08-11
Increase scroll velocity when zooming

lukicdarkoo:
I found this speed more in line with the rotation/translation speed.


omichel:
I agree, it is better. It probably deserves a line in the change log.

lukicdarkoo:
> I agree, it is better. It probably deserves a line in the change log.

Added

omichel (approved):
Thank you.

PR #3568, opened 2021-08-12
Publish Python libController

lukicdarkoo:
We need this for the new Python interface in the ROS 2 package.

In addition:
- fix the Mac bash error,
- simplify the script, and
- make sure there are enough commits cloned so the `LAST_COMMIT_YESTERDAY` can be populated.

### DIFF
--- a/.github/workflows/test_suite_linux.yml
+++ b/.github/workflows/test_suite_linux.yml
@@ -57,12 +57,12 @@ jobs:
         export ROS_DISTRO=${{ matrix.ROS_DISTRO }}
         xvfb-run --auto-servernum make distrib -j4
     - name: Prepare Webots Controller Deployment
-      # if: ${{ github.event_name == 'schedule' && matrix.os == 'ubuntu-18.04' }}
+      if: ${{ github.event_name == 'schedule' && matrix.os == 'ubuntu-18.04' }}
       uses: webfactory/ssh-agent@v0.5.3
       with:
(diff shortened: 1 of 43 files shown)


omichel (approved):
How can we ensure that fetch-depth: 15 is enough? I guess this value is large enough to make it work in most cases, but if there are many recent commits this might not be sufficient. How did you determine this value?

lukicdarkoo:
> How can we ensure that fetch-depth: 15 is enough? I guess this value is large enough to make it work in most cases, but if there are many recent commits this might not be sufficient. How did you determine this value?

According to [this badge](https://img.shields.io/github/commit-activity/m/cyberbotics/webots.svg?color=%09%2346c018) we are making 40-50 commits to the master branch (probably similar for the develop branch). That makes 2-3 commits a day (we weekends excluded), so we can assume we make 5-10 commits in peaks. I put 15 just to be safe.

We can increase it to 30 or 50, it is "cheap", it will not make almost any impact on the clone time.

omichel:
> According to [this badge](https://img.shields.io/github/commit-activity/m/cyberbotics/webots.svg?color=%09%2346c018) we are making 40-50 commits to the master branch (probably similar for the develop branch). That makes 2-3 commits a day (with weekends excluded), so we can assume we make 5-10 commits in peaks. I put 15 just to be safe.
> 
> We can increase it to 30 or 50, it is "cheap", it will not make almost any impact on the clone time.

Thank you for the explanation. It seems fine to me. And even if we miss some commit one day, it is not super critical as it will get synchronized later when we push more commits.

PR #3576, opened 2021-08-16
Fixed cppcheck error on macOS

omichel:
It seems cppcheck was upgraded on GitHub Action macOS machines and is now detecting more errors.
This PR addresses them.


stefaniapedrazzi (approved):
Thank you. The fix is fine.
But it seems there is no need to check for `mTextureCoordinatesValid` inside the for loops for each triangles and vertices and the IF block could be moved outside the for loops.

stefaniapedrazzi (approved):
Thank you!

PR #3579, opened 2021-08-17
Cleanup remnants of offline doc

omichel:
Since the offline documentation was removed in R2021b, it is not needed any more to download any javascript dependencies during the build process.


lukicdarkoo (approved):
Thank you

PR #3580, opened 2021-08-18
Fix Camera.getImageArray example

stefaniapedrazzi:
Fix example of `Camera.getImageArray`.
* check if `image` is valid: if called before the first sampling period is expired the `image` value can be `None`
* fix `print` instruction not working with Python3
* fix indentation


lukicdarkoo (approved):
Thank you

PR #3588, opened 2021-08-20
Added a burnt parameter in Sassafras tree proto (climate change update)

omichel:
This is useful for the wildfire simulation and might be useful to others.


lukicdarkoo (approved):
Thank you

lukicdarkoo:
Or, it should go to `develop`?

omichel:
It's a minor enhancement in a proto file which doesn't affect the default behavior of the proto, so I believe it can go to the master branch.

PR #3589, opened 2021-08-20
Convert basic PROTOs to FLU

lukicdarkoo:
Improve the conversion scripts and begin converting basic PROTOs (PROTOs that don't contain other PROTOs)

**Tasks**
- Script improvements
  - [x] Convert `cord` in `Geometry`
  - [x] Detect whether a PROTO contains a JavaScript fragment
  - [x] Fix array parsing issues 
- PROTOs
  - [x] Kinect
  - [x] Floor
  - [ ] MultiSenseS21


lukicdarkoo:
@omichel Could you please check:
- number rounding in the `Geometry` nodes,
- formatting (I am putting all vertices in a single line), and
- comments (I am deleting all of them).

omichel:
- Regarding float conversion, it seems that for example:
`0 0 0 0 0.027 0` gets converted to `-0.0 0.0 0.0 -1e-07 -3e-07 0.027` which seems bad to me.
Can't we convert it to `0 0 0 0 0 0 0.027` instead?
Do we really need to make matrix multiplication?
Can't we simply swap Y/Z coordinates instead as it seems to be what the matrix multiplication does?
- I don't like too much deleting the comments as they are often useful, for example to explain which part of a device a shape is.
- Putting the vertices and indexes on a single line is fine (is already how Webots is saving files).

lukicdarkoo:
> Can't we simply swap Y/Z coordinates instead as it seems to be what the matrix multiplication does?

We can but it is not generic and it looks a bit hacky. Although it would produce better results.

> * Regarding float conversion, it seems that for example:
>   `0 0 0 0 0.027 0` gets converted to `-0.0 0.0 0.0 -1e-07 -3e-07 0.027` which seems bad to me.
>   Can't we convert it to `0 0 0 0 0 0 0.027` instead?

+1

lukicdarkoo:
I have just realized this is breaking many tests. I am reverting it to the draft mode until the tests are fixed

lukicdarkoo:
The tests are passing locally

BenjaminDeleze (changes requested):
Its seems like the axis of rotation of the `Velodyne HDL-64E` and `Velodyne HDL-32E` are not converted (to see, run the simulation, show robot window and override controller)

Issue #3513, opened 2021-07-27
empty folders "http:" etc. generated after converting PBR


Thomas-Feldmeier:
webots R2021b, ubuntu20.04

add shape > appearance > ThreadMetalPlate

right click on ThreadMetalPlate, convert to base nodes

now there is a "http:" folder generated in the same folder as the world file.
inside the "http:" folder is no file and only one folder called "raw.githubusercontent.com"
etc.

```
du https:
4	https:/raw.githubusercontent.com/cyberbotics/webots/R2021b/projects/appearances/protos/textures/thread_metal_plate
8	https:/raw.githubusercontent.com/cyberbotics/webots/R2021b/projects/appearances/protos/textures
12	https:/raw.githubusercontent.com/cyberbotics/webots/R2021b/projects/appearances/protos
16	https:/raw.githubusercontent.com/cyberbotics/webots/R2021b/projects/appearances
20	https:/raw.githubusercontent.com/cyberbotics/webots/R2021b/projects
24	https:/raw.githubusercontent.com/cyberbotics/webots/R2021b
28	https:/raw.githubusercontent.com/cyberbotics/webots
32	https:/raw.githubusercontent.com/cyberbotics
36	https:/raw.githubusercontent.com
40	https:
```
(du is a linux command to list the disk space of all subfolders)

PR #3591, opened 2021-08-23
Fix convert to base nodes for textures defined with URLs

BenjaminDeleze:
**Description**
When using the `convert to base node` option, textures and meshes defined with urls should not be saved locally.

**Related Issues**
This pull-request fixes issue #3513

### DIFF
--- a/src/webots/scene_tree/WbSceneTree.cpp
+++ b/src/webots/scene_tree/WbSceneTree.cpp
@@ -49,6 +49,7 @@
 #include "WbTreeItem.hpp"
 #include "WbTreeView.hpp"
 #include "WbUndoStack.hpp"
+#include "WbUrl.hpp"
 #include "WbValueEditor.hpp"
 #include "WbVariant.hpp"
 #include "WbViewpoint.hpp"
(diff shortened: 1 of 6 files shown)


src/webots/scene_tree/WbSceneTree.cpp
```diff
@@ -728,10 +728,12 @@ void WbSceneTree::convertProtoToBaseNode(bool rootOnly) {
     while (it.hasNext()) {
       it.next();
       const QString destination(WbProject::current()->worldsPath() + it.key());
-      const QFileInfo fileInfo(destination);
-      if (!QDir(fileInfo.absolutePath()).exists())
-        QDir().mkpath(fileInfo.absolutePath());
-      QFile::copy(it.value(), destination);
+      if (!(destination.contains("webots://") || destination.contains("http://") || destination.contains("https://"))) {
```
stefaniapedrazzi, line 731: I think it is better to use the functions in `WbUrl` class rather than hard coding again all the web prefixes.
This will make the code easier to maintain.

Moreover, it seems enough to check the `it.key()` prefix instead of the whole `destination` value.
So at least the check for HTTP and HTTPS could be replaced by the `WbUrl::isWeb(it.key())` function.

Then, it would also be better to also have a single function in `WbUrl` class that checks for URLs  starting with `webots://` (currently this prefix is hard coded and checked in 4 different classes).

BenjaminDeleze:
I think I addressed all comments.

PR #3592, opened 2021-08-23
Fix 'webots://' URLs not working in the CI tests

stefaniapedrazzi:
All the tests using `webots://` URLs passes the test suite even if texture is not found (and default not found texture is used).
Using the `https://` URLs instead works when tested locally.
This is now consistent with other tests that were already using the https URLs.

However, I'm not sure that this always works in the CI because in #3566 I still got wrong camera pixel values.

Not sure if it is better to convert also all the HTTPS URLs in the test suite and copy the corresponding textures in the `tests` folder too given that tests are passing.

### DIFF
--- a/tests/api/worlds/pen_mesh.wbt
+++ b/tests/api/worlds/pen_mesh.wbt
@@ -31,7 +31,7 @@ DEF BOARD Solid {
       }
       geometry Mesh {
         url [
-          "webots://projects/default/worlds/meshes/suzanne.obj"
+          "https://raw.githubusercontent.com/cyberbotics/webots/R2021b/projects/default/worlds/meshes/suzanne.obj"
         ]
       }
     }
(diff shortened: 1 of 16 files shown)


tests/api/worlds/pen_mesh.wbt
```diff
@@ -31,7 +31,7 @@ DEF BOARD Solid {
       }
       geometry Mesh {
         url [
-          "webots://projects/default/worlds/meshes/suzanne.obj"
+          "https://raw.githubusercontent.com/cyberbotics/webots/R2021b/projects/default/worlds/meshes/suzanne.obj"
```
BenjaminDeleze, line 34: I am not sure to understand why this change is needed.
Normally, `webots://` urls are not used in the CI but are converted as `https://raw.githubusercontent.com` ones.
And I think that it works, at least for meshes defined like that, because otherwise the `pen_mesh.wbt` test will not pass if the suzanne mesh is not found.

stefaniapedrazzi: I will inspect again but in #3566 this was not working.

stefaniapedrazzi:
I checked debugging the CI tests and I confirm that the textures are loaded correctly.
However it is not clear yet why the texture color is not correctly detected by the camera in the `tests/api/worlds/skin.wbt` test.
I will further investigate in #3683.

PR #3593, opened 2021-08-23
restore mIs90DegreesRotated init

BenjaminDeleze:
**Description**
Initialize `mIs90DegreesRotated` in the .cpp file (as before) instead of in the .hpp file


omichel (approved):
Thank you.

PR #3595, opened 2021-08-23
Fix missing initialization of new rotation handles nodes

stefaniapedrazzi:
Fix wren nodes initialization and cleanup introduced in #3540.
It seems there are two issues:
* if manipulator doesn't have rotations handles, the `mRotationLineTransform` and `mRotationDoubleArrowTransform` are not initialized to NULL and could cause Webots to crash when deleting them in `~WbTranslateRotateManipulator`.
* the material used to display the black line and double arrow is not deleted causing an assertion fault:
    ```webots-bin: Scene.cpp:99: void wren::Scene::reset(): Assertion `!PhongMaterial::cachedItemCount()' failed.```

This seems the cause of the test suite failure in  #3566.
Additional assertion and checks are added to prevent misusing these nodes.


lukicdarkoo (approved):
Thank you!

PR #3601, opened 2021-08-24
Add test billboard

BenjaminDeleze:
**Description**
Add a test case for the billboard

It was necessary to create a new folder for this tests because it needs to be launched with different options than the other ones (realtime mode, with rendering and withtout being minimized).


tests/with_rendering/controllers/billboard/billboard.c
```diff
@@ -0,0 +1,124 @@
...
+  ts_setup(argv[0]);
+
+  WbNodeRef node = wb_supervisor_node_get_from_def("VIEWPOINT");
+  ts_assert_pointer_not_null(node, "wb_supervisor_node_get_from_def(\"VIEWPOINT\") failed");
+  WbFieldRef position = wb_supervisor_node_get_field(node, "position");
+  ts_assert_pointer_not_null(node, "wb_supervisor_node_get_field(\"position\") failed");
+  WbFieldRef orientation = wb_supervisor_node_get_field(node, "orientation");
+  ts_assert_pointer_not_null(node, "wb_supervisor_node_get_field(\"orientation\") failed");
+
+  int time_step = wb_robot_get_basic_time_step();
```
lukicdarkoo, line 26:
```suggestion
  const int time_step = wb_robot_get_basic_time_step();
```

lukicdarkoo (approved):
Thank you!

PR #3605, opened 2021-08-24
Billboard test to fast mode

BenjaminDeleze:
**Description**
Change the mode of the billboard test to fast.

However, the saving of screenshots is not synchronous (from the supervisor at least) and so we cannot be sure that the images are saved when the test tries to access them. That is why it was necessary to increase the size of the waiting loop if we wanted to run in fast mode (the supervisor exits the loops as soon as it detects that the images are ready).

### DIFF
--- a/tests/test_suite.py
+++ b/tests/test_suite.py
@@ -216,12 +216,11 @@ finalMessage = 'Test suite complete'
 thread = threading.Thread(target=monitorOutputFile, args=[finalMessage])
 thread.start()
 
-webotsArguments = '--mode=fast --no-rendering --stdout --stderr --minimize --batch'
 webotsArguments = '--mode=fast --stdout --stderr --batch'
-webotsArgumentsNoRendering = webotsArguments + ' --no-rendering --minimize'
 if sys.platform != 'win32':
     webotsArguments += ' --no-sandbox'
-    webotsArgumentsRendering += ' --no-sandbox'
+webotsArgumentsNoRendering = webotsArguments + ' --no-rendering --minimize'
+
 
 for groupName in testGroups:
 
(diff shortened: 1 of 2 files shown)


tests/test_suite.py
```diff
@@ -217,7 +217,7 @@ def monitorOutputFile(finalMessage):
 thread.start()
 
 webotsArguments = '--mode=fast --no-rendering --stdout --stderr --minimize --batch'
-webotsArgumentsRendering = '--mode=realtime --stdout --stderr --batch'
+webotsArgumentsRendering = '--mode=fast --stdout --stderr --batch'
```
omichel, line 220: That would seem more logical to me:
```suggestion
webotsArguments = '--mode=fast --stdout --stderr --batch'
webotsArgumentsNoRendering = webotsArguments + ' --no-rendering --minimize'
```
And then we should adapt the code to `webotsArgumentsNoRendering`.

tests/with_rendering/controllers/billboard/billboard.c
```diff
@@ -55,7 +55,7 @@ int main(int argc, char **argv) {
   // as the test sometimes fails on Windows, probably because the system
   // needs some time to make the file available to other applications
   // after it was just created.
-  for (i = 0; i < 100; i++) {
+  for (i = 0; i < 10000; i++) {
     if (file_exists("image0.png") && file_exists("image1.png") && file_exists("image2.png"))
       break;
```
lukicdarkoo, line 60: I would rather use something like this:
```py
start_time = current_time()
while current_time() - start_time < 5 seconds:
   check_whether_the_images_are_created()
   some_small_pause()
```
That way we use the wall clock instead of the simulated time which makes the test more realistic and robust.

tests/with_rendering/controllers/billboard/billboard.c
```diff
@@ -55,11 +57,16 @@ int main(int argc, char **argv) {
...
+  time_t start_time = time(NULL);
+  while (time(NULL) - start_time < 5) {
     if (file_exists("image0.png") && file_exists("image1.png") && file_exists("image2.png"))
       break;
-
-    wb_robot_step(time_step);
+    else {
+      sleep(0.1);
+      wb_robot_step(time_step);
+    }
```
lukicdarkoo, line 69:
```suggestion
    sleep(0.1);
    wb_robot_step(time_step);
```
Slightly cleaner

lukicdarkoo: Do we need both, `sleep` and `wb_robot_step`?

BenjaminDeleze: Yes

lukicdarkoo (approved):
Thank you

PR #3610, opened 2021-08-25
FLU: Convert SolidBox

stefaniapedrazzi:
Convert `SolidBox` from LUF to FLU: adapt texture mapping to produce the same results as before.


BenjaminDeleze (approved):
Thank you!

PR #3609, opened 2021-08-25
FLU: Convert apartment structure PROTOS

stefaniapedrazzi:
* [x] (030) objects/geometries/protos/TexturedBox.proto 
* [x] (212) objects/apartment_structure/protos/Window.proto
* [x] (213) objects/apartment_structure/protos/DoorLever.proto
* [x] (214) objects/apartment_structure/protos/Door.proto (requires conversion of `TexturedParallelepiped` PROTO)
* [x] (215) objects/apartment_structure/protos/Wall.proto
* [x] (216) objects/apartment_structure/protos/GenericDoorAppearance.proto
* [x] (217) objects/apartment_structure/protos/DoorKnob.proto
* [x] (218) objects/apartment_structure/protos/WallPlug.proto
* [x] (219) objects/apartment_structure/protos/Radiator.proto
* [x] (220) objects/apartment_structure/protos/Ceiling.proto
* [x] (375) bounding_objects/protos/PipeBoundingObject.proto

PR #3613, opened 2021-08-26
FLU: solid and shapes objects

stefaniapedrazzi:
* [x] (137) objects/shapes/protos/TexturedBoxShape.proto (nothing to do)
* [x] (272) objects/solids/protos/SolidTorus.proto
* [x] (273) objects/solids/protos/SolidRoundedBox.proto
* [x] (275) objects/solids/protos/SolidPipe.proto
* [x] (376) bounding_objects/protos/TorusBoundingObject.proto

To be merged after #3609 .


BenjaminDeleze (changes requested):
Thank you!
Just a minor remark: The header of the `SolidPipe.proto` and its documentation need to be updated.

I am not sure about the texture mapping on the round parts of the `SolidRoundedBox` but I don't think it is so important

BenjaminDeleze (approved):
Thank you!

PR #3620, opened 2021-08-27
Docs flu geometires mapping

BenjaminDeleze:
**Description**

- [x] Update the docs with information concerning mapping of textures
- [x] Update the images of cone and cylinder to display +y instead of -y


omichel (dismissed):
Thank you.

docs/reference/plane.md
```diff
@@ -23,3 +23,5 @@ Note that a [Plane](#plane) node is in fact not really a plane: it's a half-spac
 Anything that is moving inside the half-space will be ejected out of it.
 This means that planes are only planes from the perspective of one side.
 If you want your plane to be reversed, rotate it by &pi; using a [Transform](transform.md) node.
+
+When a texture is applied to the plane, the texture is mapped onto the plane (seen from above) with the same orientation as if the image were displayed normally in 2D (along the *x* and *y* axis).
```
stefaniapedrazzi, line 27: Typo
```suggestion
When a texture is applied to the plane, the texture is mapped onto the plane (seen from above) with the same orientation as if the image was displayed normally in 2D (along the *x* and *y* axis).
```

docs/reference/elevationgrid.md
```diff
@@ -47,3 +47,5 @@ coordinate system
 ```
 
 The `thickness` field specifies the thickness of the bounding box which is added below the lowest point of the `height` field, to prevent objects from falling through very thin [ElevationGrid](#elevationgrid)s.
+
+When a texture is applied to the elevation grid, the texture is mapped onto the elevation grid (seen from above) with the same orientation as if the image were displayed normally in 2D (along the *x* and *y* axis).
```
stefaniapedrazzi, line 51: Typo
```suggestion
When a texture is applied to the elevation grid, the texture is mapped onto the elevation grid (seen from above) with the same orientation as if the image was displayed normally in 2D (along the *x* and *y* axis).
```

stefaniapedrazzi (approved):
Thank you

PR #3624, opened 2021-08-31
Avoid changing the `LD_LIBRARY_PATH` environment in the `.bashrc`

lukicdarkoo:
It makes other packages crash. For example, RViz loads the Webots Qt library instead of the system Qt library. A developer should be responsible for configuring the `LD_LIBRARY_PATH` environment variable.

### DIFF
--- a/scripts/install/bashrc.linux
+++ b/scripts/install/bashrc.linux
@@ -16,5 +16,4 @@ export WEBOTS_ALLOW_MODIFY_INSTALLATION=1                 # If defined, you are 
 # These environment variables are necessaries to compile and run Webots #
 #########################################################################
 
-export WEBOTS_HOME=$HOME/webots                                    # Defines the path to Webots home.
-# export LD_LIBRARY_PATH=$WEBOTS_HOME/lib/webots:$LD_LIBRARY_PATH  # Add the Webots libraries to the library path (useful when launching Webots directly without using the launcher).
+export WEBOTS_HOME=$HOME/webots  # Defines the path to Webots home.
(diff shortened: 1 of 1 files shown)


scripts/install/bashrc.linux
```diff
@@ -16,5 +16,5 @@ export WEBOTS_ALLOW_MODIFY_INSTALLATION=1                 # If defined, you are
 # These environment variables are necessaries to compile and run Webots #
 #########################################################################
 
-export WEBOTS_HOME=$HOME/webots                                  # Defines the path to Webots home.
-export LD_LIBRARY_PATH=$WEBOTS_HOME/lib/webots:$LD_LIBRARY_PATH  # Add the Webots libraries to the library path (useful when launching Webots directly without using the launcher).
+export WEBOTS_HOME=$HOME/webots                                    # Defines the path to Webots home.
+# export LD_LIBRARY_PATH=$WEBOTS_HOME/lib/webots:$LD_LIBRARY_PATH  # Add the Webots libraries to the library path (useful when launching Webots directly without using the launcher).
```
omichel, line 20: I believe we can even delete this line.
```suggestion
```

omichel (dismissed):

scripts/install/bashrc.linux
lukicdarkoo: I followed `JAVA_HOME` and `ROS_DISTRO` examples, but I would also rather delete them.

omichel: I would say it is good to keep the `JAVA_HOME` and `ROS_DISTRO` commented out as they are needed in case we want to build the corresponding libraries. It is not the case for the `LD_LIBRARY_PATH`.

omichel (approved):
Thank you.

PR #3626, opened 2021-09-01
FLU: convert objects/computers

ad-daniel:
**Description**
Ditto.


projects/objects/computers/protos/ComputerMouse.proto
```diff
@@ -13,63 +13,63 @@ PROTO ComputerMouse [
...
+DEF TOPMOUSE Shape {
```
lukicdarkoo, line 38: There is a bug in the script, the indentation doesn't work properly for certain nodes

lukicdarkoo (changes requested):
Thank you, it looks good, just check the indentation

projects/objects/computers/protos/ComputerMouse.proto
ad-daniel: Thanks, didn't notice. I checked the others as well and it appears to be the only case.

lukicdarkoo (approved):
Thank you

omichel:
I believe we can merge it as the test failures are not related with this PR.

PR #3641, opened 2021-09-03
Fix animation bug on Firefox

BenjaminDeleze:
**Description**
The animation were sometimes not loading correctly on Firefox due to some synchronicity issues. It is now fix.

I also changed the font of the timer to be monotype such that it will not "shake" anymore

### DIFF
--- a/resources/web/wwi/Animation.js
+++ b/resources/web/wwi/Animation.js
@@ -458,7 +458,7 @@ export default class Animation {
     document.getElementById('left-pane').appendChild(this._currentTime);
 
     const timeDivider = document.createElement('span');
-    timeDivider.innerHTML = '\\';
+    timeDivider.innerHTML = '\/';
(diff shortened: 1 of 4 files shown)


resources/web/wwi/Animation.js
```diff
@@ -458,7 +458,7 @@ export default class Animation {
     document.getElementById('left-pane').appendChild(this._currentTime);
 
     const timeDivider = document.createElement('span');
-    timeDivider.innerHTML = '\\';
+    timeDivider.innerHTML = '\/';
```
omichel, line 461: Isn't it working simply like that?
```suggestion
    timeDivider.innerHTML = '/';
```

BenjaminDeleze: yes

omichel (approved):
Thank you.

PR #3644, opened 2021-09-06
Fix pickable cone and cylinder

### LINKED ISSUE #3643: Cone and Cylinder not pickable after change in subdivision
**Describe the Bug**
After changing the number of subdivision of a cone or a cylinder, the shape cannot be selected anymore by the mouse.

Note that this bug does not appear with sphere

BenjaminDeleze:
**Description**
Fix #3643 

It seems like `setPickable` was missing in both `buildWrenMesh`functions


omichel (changes requested):
Thank you.
We should probably add a line in the change log for this fix.

omichel (approved):
Thank you.

PR #3654, opened 2021-09-08
Fix pickable local mesh

BenjaminDeleze:
**Description**
When adding a local mesh via the scenetree, the new mesh was not pickable.
This PR fix that.


ad-daniel (approved):
Thank you, tested both local file and `webots://` obj file

PR #3670, opened 2021-09-13
Fix broadcast restore state when ready

BenjaminDeleze:
**Description**
In the streaming viewer, if the simulation was playing and the streaming viewer was launched as `broadcast`, the simulation was paused after the loading instead of resuming.

#Fix https://github.com/cyberbotics/webots_ros2/issues/299


lukicdarkoo (approved):
It works nicely, thanks!

PR #3674, opened 2021-09-13
FLU: Ned and Microbot

BenjaminDeleze:
**Tasks**
Robots:
  - [x] Ned
  - [x] Simplifiy Ned bounding object (currently uses ifs)
  - [x] Microbot

Worlds:
  - [x] ned.wbt
  - [x] microbot_pe.wbt

Other: 
- [x] correct monitor.proto


omichel (changes requested):
It seems the ned robot is not any more able to pick and place the small blue box object. It is dropping it on the way...

BenjaminDeleze:
I modified a bit the initial position of the `ned` and it seems to work better now

omichel (approved):
Perfect. Thank you.

PR #3675, opened 2021-09-14
Change default coordinate system in `empty.wbt`

ad-daniel:
**Description**
Although from 2021b `ENU` is the default coordinate system (and is so defined in `WorldInfo.wrl`), when doing `File > New World` the default world loaded (`empty.wbt`) still uses `NEU`.


lukicdarkoo:
Uff, I am not sure whether this is a good idea. Since almost all PROTOs in R2021b are not FLU users will need to rotate almost all PROTOs, starting from basic stuff like the floor.

I would rather wait for R2021c

PR #3677, opened 2021-09-14
Measure of the impact of adding a Transform node

joachimhgg:
To estimate the impact of a tranform node, we lead multiple tests. The simulation is an item falling from a height of 3m until it hits a ball. It purposes to encounters bounding boxes. This simulation is repeated multiple times. The results are the average simulation time. We change the orientation of the vehicle using a transform node, so the first experiment is to measure the simulation time of this simulation for a world in NUE  (master/2021b version) and the same world but in ENU  (#3658 / 2021c version). As you can see bellow, these results are not relevant to show the impact of adding this node:
 
### Toyota, 2 extra transform

- 2021b:
 average on 1000 sim with ball on [0.9;0.6] 724.588ms

- 2021c:
 average on 1000 sim with ball on [0.9;0.6] 726ms

- Difference:
difference on 1000 sim: 1.412ms
*loss in performance: 0.19%*

### Citroen, 3 extra transform
- 2021b:
 average on 1000 sim with ball on [0.9;0.6] 722ms
 average on 10000 sim with ball on [0.9;0.6] 722ms
 average on 1000 sim with ball on [0;0.6]: 743.932ms ; with tesla bounding box: 747.128ms
  

- 2021c:
 average on 1000 sim with ball on [0.9;0.6] 735.584ms 
 average on 10000 sim with ball on [0.9;0.6] 735.602ms
 average on 1000 sim with ball on [0;0.6]: 742.02ms ; with tesla bounding box: 745.652ms

- Difference:
difference on 1000 sim with ball on [0.9;0.6]: 13.584ms
difference on 10000 sim with ball on [0.9;0.6]: 13.602ms
difference with ball on [0,0,0.6]: 1.9ms

**CCL: the location of the ball affects the simulation time.
loss in performance with ball on [0.9;0.6] : 1.85%**

### Tesla, 1 extra transform

- 2021b:
 average on 1000 simulations 724.92ms
 average on 10000 simulations 724.9036ms
 average on 1000 sim with ball on [0,0,0.6]: 723.988ms
 
- 2021c:
 average on 1000 simulations 725.148ms
 average on 10000 simulations 725.2032ms
 average on 1000 sim with ball on [0,0,0.6]: 723.988ms

- Difference:
difference on 1000 sim: 0.228ms
difference on 10000 sim: 0.2996ms
**loss in performance: 0.04%**

### OfficePhone, no transform

- 2021b:
average on 1000 simulations: 1.407116s

- 2021c:
average on 1000 simulations: 1.0715720000000002s

### Conclusion of the first experiment

Even if using the same world, controller and by only changing the protos orientation the results are not satisfying. For example, we observe a difference with the OfficePhone object which has no transform. We can conclude that the different version or the orientation of the protos affect the simulation time. And in general the performance of the version 2021c on this simulation is slightly slower.
The second experiment is to measure the performance by changing the number of transform node for the same proto on the same version 2021c.

### Citroen

- 9 extra transform for 10000 simulations: 720ms and 720ms. no difference
- 55 extra transform: 742.104ms
- 109 extra transform: 742.104ms
- 109 different extra transform (32 rdm translations): for 1000 sim: 742.104ms for 10000 sim: 742.0484ms

**Difference: 22.104ms
loss in performance for 55 or 109 extra transform: 2.98%
estimate loss in performance for 1 extra transform: [0;0.05]%**

### Tesla

- 9 extra transform for 10000 simulations: 0.712s and 0.712s,  no difference

## Conclusion

We can observe that adding extra transform nodes affects the performance, but it needs a lot of extra transform nodes (>50) to be visible, and so it seems that adding a few Transform nodes does not affect the performance for the performed simulation. We could then saving a lot of time by only rotating the whole proto to FLU, simplify the review and prevent mistakes of changing the meshes one by one.


lukicdarkoo:
Since the benchmark is relevant only for R2021b vs R2021c I wouldn't include it in the distribution. However, it is useful to have it, so maybe putting in `webots-projects`, or even better in a personal public repository.

I guess, for us, only the final results are important, like impact per number of added transforms.

omichel:
This PR is not meant to be merged. It is there to try to answer a question:

_Shall we allow us to add a Transform layer in some protos when converting them to FLU if it facilitates a lot the conversion, given the performance impact is apparently extremely low?_

omichel:
The benefits of adding a Transform node when it speeds-up the conversion development time are:
1. Faster conversion development time.
2. Less likely to introduce bugs.
3. A more in-depth conversion could always be done afterwards, case-by-case, seamlessly for the users.

The disadvantages are:
1. The execution/loading time may be very slightly slower.
2. The proto file doesn't look very nice and maybe tricky to maintain (like if we want to change the inertia matrix of a robot, we need to take into account the top Transform).

I believe it is more a strategic question than a technical/design question. If we can finish the FLU/ENU conversion earlier, we could focus on other strategic development topics. So, I would suggest that we allow us to add Transform nodes in case it speeds-up the conversion process significantly. However, in other cases when it is fairly easy to avoid adding a Transform node, then we should indeed avoid it.

PR #3678, opened 2021-09-14
Remove libusb dependency on Linux

omichel:
It seems the libusb library is not needed to link and execute Webots on Linux.
Moreover this library is not available on the JuMax HPC and it is a problem to compile Webots there.
So, I would suggest to remove it.


lukicdarkoo (approved):
Thank you.

Hopefully, we don't break anything

PR #3683, opened 2021-09-15
Test PR: check URL in CI system

stefaniapedrazzi:
Dummy PR just for checking the test_suite behavior in CI system.


stefaniapedrazzi:
Textures are retrieved correctly during the test.
The camera image size in the skin test is just 1x1 and the on the CI machine the computed pixel color is completely different than the one computed locally on my machine when using the `tagged_wall.jpg` image.
But increasing the image size and field of view the retrieved color gets more similar to the one I get locally.

PR #3691, opened 2021-09-16
FLU: Convert Turtlebot3

lukicdarkoo:
- [x] (213) projects/robots/robotis/turtlebot/worlds/turtlebot3_burger.wbt
- [x] (543) robots/robotis/turtlebot/protos/TurtleBot3Burger.proto


ad-daniel:
I'm having some texture issues and some objects missing compared to the master branch.

Missing board and its texture is on the board support
a
Missing battery holder
b
c
Some parts of the chassis have incorrect textures
d
Forgot to turn a screw
e

Do you also see it?

lukicdarkoo:
I think most of them come from the IndexFaceSet -> .obj conversion, I will revert it.

lukicdarkoo:
@ad-daniel I have just reverted the IndexFaceSet -> .obj conversion, it should be fine now

ad-daniel (changes requested):
Yes, much better using `IndexedFaceSet`.

There still appears to be an issue with the balljoint, after a couple of timesteps it dislocates itself but doesn't in 2021b.
a

Github doesn't let me see the changes made to the PROTO since there have been too many (due to the meshes), but by comparing the mass properties from the GUI the center of mass doesn't appear to be the same anymore, did you also correct the physics?

lukicdarkoo:
Corrected the ball.

> but by comparing the mass properties from the GUI the center of mass doesn't appear to be the same anymore
Should it be the same? Since the axes are changed

ad-daniel (approved):
Yes that's what I meant, making it consistent with the new axis.

The layout of the boxes in the arena is different from 2021b, but considering on 2021b the robot gets stuck after a couple of seconds this version might be better regardless.

Thank you.

PR #3696, opened 2021-09-17
Fix Nao v3.3 (21 DoF) hands

ad-daniel:
**Description**
The v3.3 (21 DOF) variant of the Nao proto fails, the regression was introduced during the conversion from Lua to JavaScript.

Before the fix:
b

After the fix:
a


lukicdarkoo:
Poor Nao 😢

lukicdarkoo (changes requested):
Thank you. This probably deserves a changelog entry

lukicdarkoo (approved):
Thank you!

PR #3699, opened 2021-09-17
Attempt to fix Window CI build

omichel:
It seems the openjdk chocolatey package was silently upgraded from version 16 to version 17 without notice. Hence the openjdk-16.0.2 folder was not existing any more, and that broke our build process. Since the openjdk16 package doesn't exist, I upgraded to openjdk17 with the hope it will be more stable than the plain openjdk package which may be upgraded again in the future to version 18 or so. I also updated the [wiki](https://github.com/cyberbotics/webots/wiki/Windows-Optional-Dependencies/_compare/65c2be9769423f9cd7325b4a212cd6f4f4f8f617).

### DIFF
--- a/.github/workflows/test_suite_windows.yml
+++ b/.github/workflows/test_suite_windows.yml
@@ -38,10 +38,10 @@ jobs:
       with:
          submodules: true
          fetch-depth: 15
-    - name: Install Visual Studio 10 and OpenJDK 16
+    - name: Install Visual Studio 10 and OpenJDK 17
       shell: powershell
       run: |
-        choco install --force openjdk
+        choco install openjdk17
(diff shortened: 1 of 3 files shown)


lukicdarkoo:
I am not sure whether this will affect users (maybe they will not be able to use the libController with the old version of Java). Therefore, adding a note in the changelog would be useful.

omichel:
This will not affect users as the build aims at Java 8 compatibility, see [here](https://github.com/cyberbotics/webots/blob/master/src/controller/java/Makefile#L166). So, the version of the JDK is not important.

lukicdarkoo (approved):
Thank you

PR #3703, opened 2021-09-21
Update CharacterSkin PROTO appearances

stefaniapedrazzi:
Use `PBRAppearance` instead of `Appearance` nodes in the `CharacterSkin` PROTO.


BenjaminDeleze (approved):
Thank you!

PR #3711, opened 2021-09-23
Fix log WebotsJS

BenjaminDeleze:
Restore log of stdout and stderr in the web console


omichel (approved):
It could be nice to have different colors for stdout/stderr in the console (and drop the stdout/stderr prefixes). But that's also OK like that.

PR #3717, opened 2021-09-26
Use stdout and stderr callbacks

aykborstelmann:
**Description**
I saw the unused `onstdout` and `onstderr` callbacks in the webots.js view component. Probably they were intended to be used. 

**Related Issues**
#3711

**Tasks**
Unfortunately, currently the output with color is not wrapped properly, I already asked on stackoverflow why this is like it is (see https://stackoverflow.com/questions/69336052/console-log-with-color-doesnt-wrap-first-line).
If there is a fix for that problem, I will push it, otherwise, I would probably don't use colors in the default implementation of `onstdout` or `onstderr`.

### DIFF
--- a/resources/web/wwi/Stream.js
+++ b/resources/web/wwi/Stream.js
@@ -55,9 +55,12 @@ export default class Stream {
     let data = event.data;
     if (data.startsWith('robot:') || data.startsWith('robot window:'))
       return 0; // We need to keep this condition, otherwise the robot window messages will be printed as errors.
-    else if (data.startsWith('stdout:') || data.startsWith('stderr:')) {
-      console.log(data);
+    else if (data.startsWith('stdout:')) {
+      this._view.onstdout(data.substring('stdout:'.length));
       return 0;
+    } else if (data.startsWith('stderr:')) {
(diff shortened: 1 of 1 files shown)


BenjaminDeleze (changes requested):
Thank you, it is a good idea to use these functions.

I could reproduce the line wrap bug (but only with Chrome) but I think that we need to keep a distinction between stderr and stdout. 
So either the bug with colors should be fixed or we should keep the prefixes.

aykborstelmann:
Yes, I will create a bug report at chrome, but this will take its time. I could also adjust the default event listeners to add the `stderr:` and `stdout:` prefix again, that way. The text given as parameters to them would be correct, so anyone could change them on their own.

Or what other solution do you prefer ?

BenjaminDeleze (approved):
I do not like the solution with console.warn and console.log because I am afraid that it would be confusing with warnings/messages thrown by the browser. 

After discussion with @omichel, it seems to us, that the bug you found is not critical and we can ignore it for now. 
So we can merge the PR as it is.

BenjaminDeleze:
I let you merge the PR :)

aykborstelmann:
Ok great 👍🏻

Issue #3577, opened 2021-08-16
ODE errors in Niryo Ned simulation


stefaniapedrazzi:
When running the `tests/manual_test/test_worlds_warnings.py` script I get these ODE errors for the `ned.wbt` simulation.

```
Testing: /home/stefi/webots/projects/robots/niryo/ned/worlds/ned.wbt
/home/stefi/webots/projects/robots/niryo/ned/worlds/ned.wbt

ODE Message 2: mass must be > 0 in dMassCheck() [ode/src/mass.cpp:48]

ODE Message 2: mass must be > 0 in dMassCheck() [ode/src/mass.cpp:48]

ODE Message 2: inertia must be positive definite in dMassCheck() [ode/src/mass.cpp:52]

ODE Message 2: inertia must be positive definite in dMassCheck() [ode/src/mass.cpp:52]
```

Another issue is that these errors are not printed in the Webots console.

lukicdarkoo:
I believe it is an ODE debug message and it looks similar to https://github.com/cyberbotics/webots-projects/issues/7

PR #3719, opened 2021-09-27
Fix mass calculation of Mesh nodes

lukicdarkoo:
Related to https://github.com/cyberbotics/webots-projects/issues/7 and it may also (fix) https://github.com/cyberbotics/webots/issues/3577.

We don't handle the mass property when used with Mesh nodes, this PR fixes it.


omichel (approved):
Well spotted. Thank you!

PR #3729, opened 2021-09-30
FLU: fix asymetric friction

BenjaminDeleze:
The `computeFrictionDirection` was wrongly converted in #3510 

I discovered that the cylinder asymmetric friction was wrong when trying to run robots with mecanum wheel such as SummitXL or Youbot.
With some more extensive inspection I believe that Box and Plane were wrong too.


omichel (approved):
Thank you. That looks good to me. Only the URDF test is failing, but is not related with this PR.

PR #3731, opened 2021-10-01
Fix tests_sources on Windows

omichel:
Attempt to fix the CI tests_sources on Windows.


omichel:
We should fix this on the master branch first.

PR #3733, opened 2021-10-01
FLU: kondo robots

BenjaminDeleze:
**Protos**
  - [x] Khr-2hv.proto
  - [ ] Khr-3hv.proto

**Worlds**

- [ ] anaglyph.wbt (wait for Create)
- [x] khr-2hv.wbt 

**Docs**
- [x] khr-2hv.md


BenjaminDeleze:
The name of the motors of Create were inverted. That is why there was a problem with the controller in `anaglyph.wbt`. I did not change the controller of the other worlds (`apartement.wbt` and `create.wbt`) because I think it was also wrong before: the robot was turning left when it was detecting a left obstacle. Now it turns right when a left obstacle is detected.

I also modified the `CreatedWall.proto`. I had introduced a error during the conversion.

Issue #3556, opened 2021-08-10
Sumo still crashing in `highway_overtake.wbt`


ad-daniel:
**Describe the Bug**
There appears to still exist problems with sumo. 

Steps to reproduce:
- Open  `highway_overtake.wbt`
- Let it run a lot. It might take 30 simulated minutes, sometimes 1 hour before it crashes.

It doesn't appear to be a regression, the crash fixed [here](https://github.com/cyberbotics/webots/pull/3391) happened almost immediately, so this one might have gone unnoticed.

The crash affects both ubuntu 20 and windows

llessieux:
Stack overflow for me and I didn't need to wait for 20 minutes.

llessieux:
caused by a NaN value being generated apparently and set to the angle section of a WbSFRotation object. No idea where yet.
This cause equality checks to fail all the time, triggering an infinite update loop causing the stack overflow.
I have found out that minimizing the windows make it happen faster too ;)

llessieux:
The cause is a floating point error it seems.
In WbAbstractTransform::rotate when the rotation angle is very small it seems that it can caused an invalid computation.
The WbRotation isValid is not checked (and it doesn't check for NaN on the angle anyway) so it propagates an invalid value and that causes havoc.

llessieux:
Funny I have driven on that road before :)

llessieux:
I'll propose a fix to avoid the crash but I can't really tell how wide an effect that can have.
I guess everybody can agree that propagating NaN values are not a good idea but what should be done in these case can be hard to decide.
The top call I can see is the WbSolid::prePhysicsStep. The rotate call is where it goes wrong.

omichel:
Maybe can you dig deeper to understand how this NaN value is computed? Is it the result of some specific mathematical operation?

llessieux:
Yes I should be able to get to the root of the issue. My guess is that the normalize of the angle is the issue but I will try to get the exact values causing the problem.

llessieux:
Invalid WbRotation created: 
v:0 6.0729265702774704e-05 0
Rotation :[
  0.9999999981559782 0 6.072926566544601e-05
  0 1 0
  -6.072926566544601e-05 0 0.9999999981559782
]

Intermediate Matrix WbMatrix3(mRotation->value()):[
  -0.9999999981559783 0 6.07292633151153e-05
  0 0.9999999999999994 0
  -6.07292633151153e-05 0 -0.9999999981559783
]

[...]

PR #3734, opened 2021-10-01
Fix crash in Highway_overtake.wbt #3556 due to acos getting out of va…

llessieux:
Fix crash in Highway_overtake.wbt #3556 due to acos getting out of valid range value causing a NaN valid to be generated.

This in turn is assigned to a WbRotation angle which triggers an infinite notification loop leading to a stack overflow.
A equality check with a NaN value always being false, the WbSFRotation keeps on emitting changed signals.

Add WbMathsUtilities::clampedAcos and replace acos calls with that when the code in place was not already checking the bounds.
Add WbMathsUtilities::clampedAsin too, and update code using asin to prevent  indefinite values being generated.

WbRotation::isValid has been updated to check against NaN value stored in the angle part too.

Both clampedAcos and clampedAsin functions are inline for performance reasons.


src/webots/maths/WbMathsUtilities.hpp
```diff
@@ -63,4 +67,25 @@ inline double WbMathsUtilities::normalizeAngle(double angle, double lastSpot = 0
     d -= 2.0 * M_PI;
   return d + lastSpot;
 }
+
+// Make sure that the value is within the valid range before calling acos
+// If not then the behavior is similar to a clamped value.
+inline double WbMathsUtilities::clampedAcos(double value) {
+  if (value >= 1.0)
+    return 0.0;
+  else if (value <= -1.0)
```
omichel, line 76: The `else` is not needed here.
```suggestion
  if (value <= -1.0)
```

src/webots/maths/WbMathsUtilities.hpp
```diff
@@ -63,4 +67,25 @@ inline double WbMathsUtilities::normalizeAngle(double angle, double lastSpot = 0
...
+    return 2.0 * M_PI;
+  return acos(value);
+}
+
+// Make sure that the value is within the valid range before calling asin
+// If not then the behavior is similar to a clamped value.
+inline double WbMathsUtilities::clampedAsin(double value) {
+  if (value >= 1.0)
+    return M_PI / 2;
+  else if (value <= -1.0)
```
omichel, line 86:
```suggestion
  if (value <= -1.0)
```

llessieux:
Something must be wrong with my Clang Format setup. It is complaining about an extra space but I can't figure out where...
Really strange. I made sure that all the files had been formatted before committing this time and yet...

src/webots/maths/WbRotation.hpp
```diff
@@ -81,7 +81,8 @@ class WbRotation {
   }
 
   // invalid only if |axis| == 0.0
-  bool isValid() const { return !(mX == 0.0 && mY == 0.0 && mZ == 0.0); }
+  // the mAngle == mAngle checks the NaN case.
+  bool isValid() const { return !(mX == 0.0 && mY == 0.0 && mZ == 0.0) && (mAngle == mAngle); }
```
omichel, line 85: Interesting, I use to use `isnan(mAngle)` instead, but it is fine that way (and requires no dependency on math.h)

omichel (changes requested):
That looks very good.
I will check the clang format problem.

src/webots/maths/WbRotation.hpp
llessieux: yes I saw that after having committed the code that is was used in other places when I did the review.
Do you want me to replace it?

omichel: I believe it is indeed slightly more readable (will save us the comment).

llessieux:
I should probably add something in the changelog I guess ;)

omichel:
> I should probably add something in the changelog I guess ;)

Yes, please.

ad-daniel:
I can confirm that I'm no longer having the crash on ubuntu 20 on this branch, thank you

docs/reference/changelog-r2021.md
```diff
@@ -18,6 +18,7 @@ Released on XX, XXth, 2021.
     - Fixed mass calculation of Mesh nodes ([#3719](https://github.com/cyberbotics/webots/pull/3719)).
     - Fixed regression where the v3.3 (21 DoF) variant of the [Nao](../guide/nao.md) PROTO had no hands ([#3696](https://github.com/cyberbotics/webots/pull/3696)).
     - Fixed laser and infra-red distance sensors hitting fully transparent objects ([#3726](https://github.com/cyberbotics/webots/pull/3726)).
+    - Fixed a crash caused by acos function being called with out of valid range value and leading to a stack overflow. ([#3734](https://github.com/cyberbotics/webots/pull/3734))
```
omichel, line 21:
```suggestion
    - Fixed a crash caused by acos function being called with an out-of-range value and leading to a stack overflow ([#3734](https://github.com/cyberbotics/webots/pull/3734)).
```

src/webots/maths/WbRotation.hpp
omichel, line 85:
```suggestion
  bool isValid() const { return !(mX == 0.0 && mY == 0.0 && mZ == 0.0) && !isnan(mAngle); }
```

omichel (changes requested):

omichel (approved):
Thank you!

llessieux:
Sorry it was dinner time here ;)
Merging now.

PR #3736, opened 2021-10-01
ENU/FLU: convert hexapod and soccer demo samples

stefaniapedrazzi:
Worlds:
* [x] (085) projects/samples/demos/worlds/hexapod.wbt
* [x] (086) projects/samples/demos/worlds/soccer.wbt


projects/samples/demos/worlds/hexapod.wbt
```diff
@@ -231,51 +230,51 @@ DEF HEXAPOD Robot {
...
           minPosition -0.7
           maxPosition 0.7
         }
       ]
       endPoint Solid {
-        translation 0.2 0 -0.05
+        translation 0.2 -0.05 0
         children [
           USE HIP_SHAPE
           DEF LEFT_FEMUR_TRANS Transform {
```
BenjaminDeleze, line 250: Shouldn't it be RIGHT_FEMUR_TRANS ?
Same remark for line 256.

If it is the case, the USEs have to be updated accordingly

BenjaminDeleze (changes requested):
Seems good. I have only one minor remark

projects/samples/demos/worlds/hexapod.wbt
stefaniapedrazzi: You are right.
I fixed some of the DEF names but not all of them

BenjaminDeleze (approved):
Thank you!

PR #3740, opened 2021-10-04
Add error message on connection timeout

aykborstelmann:
**Description**
When the simulation server is offline, we noticed that we don't get any errors message.

I fixed this issue in `server.js`, but I'm not sure if we should call `this._view.onquit()`, since it's also not called when the response from the session server starts with `Error:`: https://github.com/cyberbotics/webots/blob/53b60081c13e3b3f9dfee31fa4046c45dc870219/resources/web/wwi/Server.js#L31

What's your opinion on that @BenjaminDeleze ?

### DIFF
--- a/resources/web/streaming_viewer/setup_viewer.js
+++ b/resources/web/streaming_viewer/setup_viewer.js
@@ -46,6 +46,7 @@ function onDisconnect() {
   ipInput.disabled = false;
   modeSelect.disabled = false;
   broadcast.disabled = false;
+  connectButton.disabled = false;
 }
 
 function disconnect() {
--- a/resources/web/wwi/Server.js
+++ b/resources/web/wwi/Server.js
@@ -58,7 +58,7 @@ export default class Server {
 
   onError() {
     document.getElementById('webotsProgress').style.display = 'none';
-    this._view.onclose()
+    this._view.onquit()
   }
 
   onOpen(event) {
--- a/resources/web/wwi/Server.js
+++ b/resources/web/wwi/Server.js
@@ -29,7 +29,7 @@ export default class Server {
       .then(response => response.text())
       .then(function(data) {
         if (data.startsWith('Error:')) {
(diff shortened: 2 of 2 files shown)


aykborstelmann:
I'm also not sure if it is correct that this PR has master as it's base branch.

BenjaminDeleze:
Well spotted! 

Concerning the `this._view.onquit()`, I think it is a good idea to add a quit/reset in some ways.

I think we have a choice to make:

- Either we use something like here https://github.com/cyberbotics/webots/blob/53b60081c13e3b3f9dfee31fa4046c45dc870219/resources/web/wwi/Stream.js#L44
where it pop an alert with the error and then reload the web page.
- Or we can go with `this._view.onquit()` but it will demand more effort to the user if it has developed it own UI because, by default `onquit`, will reset the simulation but not the `connect` button for example as it is outside of WebotJS. But it is possible to add actions to the `view.onquit` function via the `disconnectCallback` of the StreamingViewer.js

In both case, I think we should add it here as well https://github.com/cyberbotics/webots/blob/53b60081c13e3b3f9dfee31fa4046c45dc870219/resources/web/wwi/Server.js#L31

Concerning the branch, as it is kind of a bug fix, I think Master is fine.

aykborstelmann:
I prefer the second solution, for our integration in the openroberta-lab, we don't want to reload the page. Is this fine, how I did this ?

BenjaminDeleze:
I think it is enough to replace the `this.view._onclose` by `this.view_onquit` in the function you created and then you should update the `setup.js` in `resources/web/streaming_viewer` by adding `connectButton.disabled = false;` at line 48.

The update of `setup.js` is to have a working example of how to modifiy the connect button throught the `onDisconnect` callback function.

omichel (approved):
Looks good to me.

aykborstelmann:
Okay then I'll merge it

PR #3741, opened 2021-10-04
Fix acos issues

llessieux:
Following the previous comments on issue  [#3556](https://github.com/cyberbotics/webots/issues/3556), I updated the previously added code with an assert if the input value is way out of range.
Here "way out of range" means greater by 0.0001.
I tested the highway_overtake in debug and it still works so the errors were always smaller than the epsilon specified.


omichel (dismissed):
Thank you!

llessieux:
You are welcome.

llessieux:
Do you want me to wait for ShuffleWire review too? Just in case ;)

omichel:
> Do you want me to wait for ShuffleWire review too? Just in case ;)

Yes, that would be nice.

ShuffleWire:
That sound good to me !
As a side note, it's rather surprising that you got value outside of the [-1;1] range, but numerical error are what they are...

ShuffleWire (dismissed):

ShuffleWire:
> > @omichel does assert() not already including NDEBUG stuff ?
> 
> Yes, it does, but here the `#ifdef` also cover the implementation of a constant which we want to avoid in the release version. Alternatively we may use a #define for this constant and remove the `#ifdef`/`#endif` statements.

Ok, I leave it like that, this is out-of-scope for this PR.

llessieux:
My guess is that there are quite a few other areas where values get out of range. 
I have seen some really strange behavior with cars spinning a lot after collisions and jumping when wheels touch the groun. I wouldn't be surprised to find that in these cases the forces generated are way above normal but then again it is just a guess right now.
Floating point computation especially when you use matrices thend to cascade a lot and at some points you will get non-normalizable matrices. I have had that in the past ;) Using double reduces the risk but it is still there.
Yes I wasn't sure what you would prefer for the epsilon, I wasn't keen on a global const or define and since I can't remember what is allowed or not in C++11 (working mainly in C++17 and just started on C++20 now ;), I decided to go for the safest way.
I am happy to update it with whatever you prefer ;) It is your code base and it won't take long to update.

omichel:
Maybe declaring:
```c++
static const double EPSILON = 0.0001;
```
at the global scope and removing the `#ifdef` and `#endif` around the `assert` would be fine.

llessieux:
I'll try that it might not trigger some warning because it is unused ;)

omichel:
> I'll try that it might not trigger some warning because it is unused ;)

If that happens, we could do this: 
```c++
#ifndef NDEBUG
static const double EPSILON = 0.0001;
#endif
```

llessieux:
Ok it worked fine with the static const. Just pushed the changes.

omichel (approved):
Great, thank you.

PR #3743, opened 2021-10-04
fix issue with python3X_path for Windows installation from sources

BenjaminHug8:
**Description**
Makefile of controller was failling because for Windows it tries to compile even if the `PYTHON3X_HOME` was not defined (for python 3.7, 3.8 or 3.9).

This PR add checks to avoid to compile if those paths are not defined.


omichel (approved):
Thank you!

PR #3744, opened 2021-10-04
Attempt to fix VM detection on Windows

### LINKED ISSUE #3738: Windows 11 with Hyper-V installed detected as VM.
When running Webots in Windows 11 with Hyper-V installed, the WbSysInfo::isVirtualMachine will return true even when Webots is running on the Host system.

This cause a rendering performance degradation due to excessive glFlush calls.

**Steps to Reproduce**
Run Webots with gdb on a Windows 11 machine with Hyper-V installed.
put breakpoint in isVirtualMachine and check the result.

**Expected behavior**
Webots correctly detects that it is not running in a VM.

**System**
 - Windows 11 x64 with Hyper-V installed
 - Nvidia RTX 3090 24GB.

Now what is strange is that the 31st bit of the ECX register in the result of the CPUID EAX=1 call is set.
This indicates the presence of an HyperVisor.
See [CPUID](https://en.wikipedia.org/wiki/CPUID#Calling_CPUID) information.

Now this might well be a Windows 11 ONLY issue.
I can't check that on Windows 10.

omichel:
Might fix #3738. Taken from https://stackoverflow.com/a/66974571/810268

### DIFF
--- a/src/webots/core/WbSysInfo.cpp
+++ b/src/webots/core/WbSysInfo.cpp
@@ -131,9 +131,8 @@ const QString &WbSysInfo::sysInfo() {
       QStringList list = kernel_version.split(rx, Qt::SkipEmptyParts);
       if (list.size() == 3) {
         int build_number = list[2].toInt();
-        if (build_number >= 2200) {
+        if (build_number >= 2200)
           version = "Windows 11";
-        }
       }
       sysInfo.append(version);
       break;
--- a/src/webots/core/WbSysInfo.cpp
+++ b/src/webots/core/WbSysInfo.cpp
@@ -126,12 +126,12 @@ const QString &WbSysInfo::sysInfo() {
     case QSysInfo::WV_WINDOWS10:  // Or Windows 11
     {
       QString version("Windows 10");
-      QString kernel_version = QSysInfo::kernelVersion();
+      QString kernelVersion = QSysInfo::kernelVersion();
       QRegExp rx("[.]");
-      QStringList list = kernel_version.split(rx, Qt::SkipEmptyParts);
+      QStringList list = kernelVersion.split(rx, Qt::SkipEmptyParts);
       if (list.size() == 3) {
-        int build_number = list[2].toInt();
-        if (build_number >= 2200)
+        const int buildNumber = list[2].toInt();
+        if (buildNumber >= 2200)
           version = "Windows 11";
       }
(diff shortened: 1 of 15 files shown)


llessieux:
I'll check that as soon as I can.

llessieux:
By the way, I have a fix for the Windows 11 detection too if you want. Same file.
It is just an extra build number check to switch between Windows 10 and 11.

src/webots/core/WbSysInfo.cpp
```diff
@@ -368,7 +368,7 @@ bool WbSysInfo::isVirtualMachine() {
   __get_cpuid(0x1, &eax, &ebx, &ecx, &edx);
   if (!(ecx & (1 << 31))) {
     virtualMachine = 1;
```
llessieux, line 368: I think that the virtualMachine should be set to 0 in this case.

llessieux (changes requested):

llessieux:
That's the change I have tested for the Windows 11 detection.

    case QSysInfo::WV_WINDOWS10:  // Or Windows 11
    {
      QString version("Windows 10");
      QString kernel_version = QSysInfo::kernelVersion();
      QRegExp rx("[.]");
      QStringList list = kernel_version.split(rx, Qt::SkipEmptyParts);
      if (list.size() == 3) {
        int build_number = list[2].toInt();
        if (build_number >= 2200) {
          version = "Windows 11";
        }
      }
      sysInfo.append(version);
      break;

omichel:
That looks good. May I ask you to commit it in there?

llessieux:
> 
> 
> That looks good. May I ask you to commit it in there?

Sure. But that will have to wait until tomorrow ;) Getting late here :)

Ok I can confirm that your fix seems to be working here.

llessieux:
Ok Windows 11 detection added.

omichel:
It doesn't work quite well as on my Windows 10 machine, it says "Windows 11". The list returned in `kernelVersion` is `("10", "0", "19043")` on my machine, which obviously leads to "Windows 11".

llessieux:
I think that there is a zero missing :) It should be 22000.

llessieux:
You can check it also by looking at the version number of NtosKernel.exe in C:\Windows\System32.

llessieux (approved):

PR #3747, opened 2021-10-05
Fix `StraightStairs` and tiago++ world

ad-daniel:
**Description**
The conversion of the `StraightStairs` PROTO was incorrect, resulting in wrong textures for the steps.

The staircase in tiago++ world also had a small mistake where the rail was not in the right position.

Before
aa
After
a

Before
c
After
d


omichel (approved):
Thank you.

PR #3748, opened 2021-10-06
Fix cppcheck2.6 errors

omichel:
It seems a new version of ccpcheck landed on the macOS CI machine, breaking the source tests there.
This PR attempts at fixing the code to be compliant with this new version of cppcheck.

It seems there are a few falses positive which I silenced by `cppcheck-suppress` statements.
The `cppcheck-suppress virtualCallInConstructor` are somehow false positive resulting from calls to `QtObject::connect` in constructors. This apparently safe to silence them as the corresponding methods are not called from the constructor, see https://stackoverflow.com/questions/53527950/is-it-safe-to-connect-a-signal-to-a-pure-virtual-slot-in-a-constructor-of-a-base


llessieux:
Strange about the need to remove the auto in favor of the real class in one case...
By the way, is there any reason as to why you are still on C++11 ?
Surely all the build tools are now capable of doing C++17 no ?

PR #3755, opened 2021-10-08
FLU: Fix rotating lidar

BenjaminDeleze:
The normal Lidar was converted correctly, but the rotating one was missing one rotation.


BenjaminDeleze:
The failing test is not related to this PR

Issue #3585, opened 2021-08-18
minPosition can be bigger than pi


Thomas-Feldmeier:
https://cyberbotics.com/doc/reference/motor

```
  SFFloat  minPosition       0        # (-inf, inf) or [-pi, pi]
  SFFloat  maxPosition       0        # (-inf, inf) or [-pi, pi]

```

should be

```
  SFFloat  minPosition       0        # (-inf, inf)
  SFFloat  maxPosition       0        # (-inf, inf)

```

E.g. in UR10e (rotational motor)  minPosition -6.28318530718

PR #3761, opened 2021-10-08
Fix more doc typos

omichel:


ShuffleWire:
Should we add : https://github.com/cyberbotics/webots/issues/3585 ?

PR #3791, opened 2021-10-18
FLU: Fix CoM test

BenjaminDeleze:
The test is using an E-Puck and was not modified when the E-Puck was converted to FLU


BenjaminDeleze:
I think so, it is `FAILURE with /home/runner/work/webots/webots/tests/api/worlds/extern.wbt: Timeout the results file has not been written quickly enough`.

I relaunched the tests to be sure

PR #3804, opened 2021-10-20
Update Info.plist

core-code:
according to your documentation you require 10.14 "Mojave", but the requirements are missing from the binary


core-code:
thanks @omichel !

PR #3806, opened 2021-10-21
Convert A4 portal to FLU

omichel:
I fixed the Protal.proto to be FLU and I also fixed the CardboardBox proto which was not FLU.
I tested the controller and it worked fine with the new world/proto.
I also added keyboard control to the controller to make it easy to test (for users who don't want to install the Picaxe software).


omichel:
> As we already converted some worlds with the cardboard box in it, maybe it is good to modify them as well. Or in another PR

I don't believe it is needed to change them, as the impact on existing simulations is very low (I checked a couple of them, like moose_demo.wbt and saw no problem).

PR #3856, opened 2021-11-02
Fix Windows builds due to OpenJDK update

omichel:
The OpenJDK package was updated yesterday on Chocolatey and a path change broke the Windows build.


ad-daniel (approved):
Thank you.

PR #3864, opened 2021-11-03
FLU: osm_importer

BenjaminDeleze:
**Description**
To be merged after #3768 

You can test the osm_importer by using the `.osm` files in `projects/vehicles/resources/osm_files`

The height interpolation was broken on master, so I fixed it. To test it you can try to generate a Webots world from `CH_Vens.osm` with the `--enable-3D` option.
It is quite long to get all the elevation data so you can use this file [elevation_good.txt](https://github.com/cyberbotics/webots/files/7467166/elevation_good.txt)
To use the file instead of downloading the point comment the lines 133-136 in `elevation.py` and add those lines just below 
```
text_file = open("elevation_good.txt", "r")
result = text_file.readlines()
result = [float(value) for value in result[0].split(', ')]
```

Instruction to use the osm_importer: https://www.cyberbotics.com/doc/automobile/openstreetmap-importer#how-to-use-the-importer


ad-daniel (approved):
Looks good, thank you.

PR #3867, opened 2021-11-05
Fix node deletion from supervisor not refreshing scene tree

### LINKED ISSUE #3863: supervisor does not remove nodes inside proto field MFNode
create a controller, proto and world file from the text below.
run the world, select the tree, press the - key (minus on numeric keypad is possible too).
now the tree is not visible in the graphics window, but it is still visible in the scene tree.
click on tree in the scene tree, press 'del' key to remove it.
webots will crash.

if the proto is simplified to
```js
{
  Group {
    children IS Test
  }
}
```
then webots will crash after pressing - key.

the bug does not appear if the selected object is not inside of an proto:
click on the proto, convert it to base nodes, no crash happens if the tree is deleted by pressing - key.

webots --sysinfo
System: Linux 5.11.0-38-generic x86_64
Processor: Intel(R) Core(TM) i5-4590 CPU @ 3.30GHz
Number of cores: 4
OpenGL renderer: GeForce GTX 1050/PCIe/SSE2
OpenGL version: 4.6.0 NVIDIA 460.91.03

webots --version
Webots version: R2021b

Ubuntu 20.04

TestCrash.cpp:
```c++
#include <webots/Supervisor.hpp>
#include <webots/Keyboard.hpp>

#include <stdio.h>

using namespace std;
using namespace webots;

  Supervisor *robot = new Supervisor();
  int timeStep = (int)robot->getBasicTimeStep();
  webots::Keyboard *keyboard = robot->getKeyboard();
  keyboard->enable(timeStep);
  while (robot->step(timeStep) != -1) {
    webots::Node* selected = robot->getSelected();
    int key = keyboard->getKey();
    if(key=='-' && selected) {
       selected->remove();
    }
  }
  delete robot;
  return 0;
}
```
-----------------------------

TestCrash.proto:
```js
#VRML_OBJ R2021b utf8

PROTO TestCrash [
  field MFNode     Test [
  ]
]
{
  Group {
    children [
      Group {
        children IS Test
      }
    ]
  }
}
```
--------------------------------
TestCrash.wbt:

```js
#VRML_SIM R2021b utf8
WorldInfo {
}
Viewpoint {
}
TexturedBackground {
}
TexturedBackgroundLight {
}
TestCrash {
  Test [
    PottedTree {
    }
  ]
}
Robot {
  controller "TestCrash"
  supervisor TRUE
}
```

ad-daniel:
**Description**
Fixes #3863.

Although node deletion using the "delete" button and node deletion from supervisor ultimately rely on a same call to `WbNodeOperations::deleteNode` in order to clear it, when deleting from supervisor the scene tree wasn't refreshed (i.e it didn't trigger `WbMultipleValue::itemRemoved` in `WbTreeItem`). The reason for the difference is that when deleting from keyboard the exposed scene tree node is actively selected [here](https://github.com/cyberbotics/webots/blob/d6a89d8f6ad9c6b8ac2f2dd822247082e2e1219a/src/webots/scene_tree/WbSceneTree.cpp#L426-L429) whereas when deleting from supervisor what is being referenced is the internal node to the PROTO.

The proposed fix climbs the ladder until a PROTO parameter is found (if any) before deleting.
Also, deletion of internal nodes from supervisor was restricted. Although it is not possible to select an internal node from the 3D view, one might get the nodeRef from methods like `getFromProtoDef`.

This world can be used for testing: [supervisor_remove.zip](https://github.com/cyberbotics/webots/files/7481758/supervisor_remove.zip)

### DIFF
--- a/src/webots/nodes/utils/WbSupervisorUtilities.cpp
+++ b/src/webots/nodes/utils/WbSupervisorUtilities.cpp
@@ -1525,6 +1525,12 @@ void WbSupervisorUtilities::handleMessage(QDataStream &stream) {
       unsigned int nodeId;
       stream >> nodeId;
       WbNode *node = WbNode::findNode(nodeId);
+
+      // as findNode might return the internal one, it's necessary to climb the ladder up to the protoParameterNode otherwise
+      // the scene tree will not be refreshed when deleting it
+      while (node && node->protoParameterNode() != NULL)
+        node = node->protoParameterNode();
+
       if (node) {
         if (node == mRobot)
           mShouldRemoveNode = true;
--- a/src/webots/nodes/utils/WbSupervisorUtilities.cpp
+++ b/src/webots/nodes/utils/WbSupervisorUtilities.cpp
@@ -1528,7 +1528,7 @@ void WbSupervisorUtilities::handleMessage(QDataStream &stream) {
 
       // as findNode might return the internal one, it's necessary to climb the ladder up to the protoParameterNode otherwise
       // the scene tree will not be refreshed when deleting it
-      while (node && node->protoParameterNode() != NULL)
+      while (node && node->protoParameterNode())
         node = node->protoParameterNode();
 
       if (!WbNodeUtilities::isVisible(node)) {
--- a/src/webots/nodes/utils/WbSupervisorUtilities.cpp
+++ b/src/webots/nodes/utils/WbSupervisorUtilities.cpp
@@ -1531,6 +1531,12 @@ void WbSupervisorUtilities::handleMessage(QDataStream &stream) {
       while (node && node->protoParameterNode() != NULL)
         node = node->protoParameterNode();
 
+      if (!WbNodeUtilities::isVisible(node)) {
+        mRobot->warn(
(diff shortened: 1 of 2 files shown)


src/webots/nodes/utils/WbSupervisorUtilities.cpp
```diff
@@ -1525,6 +1525,18 @@ void WbSupervisorUtilities::handleMessage(QDataStream &stream) {
       unsigned int nodeId;
       stream >> nodeId;
       WbNode *node = WbNode::findNode(nodeId);
+
+      // as findNode might return the internal one, it's necessary to climb the ladder up to the protoParameterNode otherwise
+      // the scene tree will not be refreshed when deleting it
+      while (node && node->protoParameterNode() != NULL)
```
omichel, line 1531: That seems superfluous:
```suggestion
      while (node && node->protoParameterNode())
```

omichel (changes requested):

Thomas-Feldmeier:
in supervisor_remove.zip, 

    key=='a'

should be

  key=='A'

for linux (ubuntu 20.04). pressing only the a key always returns 65 == 'A'

omichel (approved):
Thank you.

Thomas-Feldmeier:
Thank you.

Issue #3483, opened 2021-07-21
supervisor setSFInt32 crashes on own proto fields and lua regeneration


Thomas-Feldmeier:
webots R2021b, ubutun 20.04

Crash.proto:
```
#VRML_SIM R2021b utf8
PROTO Crash [
  field SFInt32 subdivision 1
]
{
DEF Crash Transform {
  children [
    Shape {
      geometry Sphere {
#        subdivision IS subdivision
        subdivision %{= fields.subdivision.value }%
      }
    }
    Robot {
     controller "Crash"
     supervisor TRUE
    }
  ]
}
}
```

Crash.cpp:
```
#include <webots/Supervisor.hpp>

int main(int argc, char **argv) {
  webots::Supervisor *robot = new webots::Supervisor();
  robot->getFromDef("Crash")->getField("subdivision")->setSFInt32(2);
  robot->step(32);   
}
```

add controller Crash, compile, add proto node Crash to an world, webots will crash
the controller writes field subdivision. this field is a parameter of lua -> lua regeneration (a change in parameter "subdivision" restarts the lua program generating a new proto file. the controller is part of this proto and would be restarted too. here the crash happens)
no crash if:
* subdivision IS subdivision   instead of   subdivision %{= fields.subdivision.value }%
* or no controller is inside of the node

%{ subdivision = fields.subdivision.value }% will crash too if there is a controller present

ad-daniel:
This is an odd one, I've made a few tests and:
- replacing the template language from Lua to JavaScript  it doesn't crash
```
#VRML_SIM R2021b utf8
# template language: javascript

PROTO Crash [
  field SFInt32 subdivision 1
]
{
DEF Crash Transform {
  children [
    Shape {
      geometry Sphere {
#        subdivision IS subdivision
        subdivision %<= fields.subdivision.value >%
      }
    }
    Robot {
     controller "Crash"
     supervisor TRUE
    }
  ]
}
}

```
- When using Lua and triggering the regeneration manually from the interface, it doesn't crash

The second test suggests that the issue might be related to having the controller running while regenerating the Robot that is using said controller, but it that were the case, it shouldn't work even in JavaScript but it does

PR #3869, opened 2021-11-08
Fix auto-regeneration crash

ad-daniel:
**Description**
Fixes: https://github.com/cyberbotics/webots/issues/3483

The process that leads to the crash is:
- Robot A is created when loading the world, being a supervisor, a connection is made between the signal `worldModified` and `handleWorldModificationFromSupervisor`.
- The controller uses a supervisor  `field_set_...` function to change an exposed parameter of the PROTO.
- When the controller is executed, during the execution of the [immediate messages](https://github.com/cyberbotics/webots/blob/1d3919e5212a1a9b2db580bcbe8a537771482506/src/webots/nodes/utils/WbSupervisorUtilities.cpp#L367-L384) a auto-regeneration is triggered, which results in the destruction of Robot A, and the creation of Robot B.
- After the regeneration, an emission of `worldModified` is triggered [here](https://github.com/cyberbotics/webots/blob/1d3919e5212a1a9b2db580bcbe8a537771482506/src/webots/nodes/utils/WbSupervisorUtilities.cpp#L383), however the sender of this emission no longer exists (Robot A) hence the crash

The proposed solution is to emit only if the sender robot (i.e the supervisor that invoked the `supervisor_set_...` function to begin with) still exists.

World that cause the crash: [lua_regen_crash.zip](https://github.com/cyberbotics/webots/files/7503479/lua_regen_crash.zip)

### DIFF
--- a/src/webots/nodes/utils/WbSupervisorUtilities.cpp
+++ b/src/webots/nodes/utils/WbSupervisorUtilities.cpp
@@ -381,8 +381,7 @@ void WbSupervisorUtilities::processImmediateMessages(bool blockRegeneration) {
     return;
 
   WbTemplateManager::instance()->blockRegeneration(false);
-  const QList<WbRobot *> &robots = WbWorld::instance()->robots();
-  if (robots.contains(mRobot))  // only emit if robot still exists (FieldSetRequest could have regenerated it)
+  if (WbWorld::instance()->robots().contains(mRobot))  // only emit if robot still exists (FieldSetRequest could have regenerated it)
     emit worldModified();
 }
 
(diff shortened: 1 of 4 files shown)


src/webots/nodes/utils/WbSupervisorUtilities.cpp
```diff
@@ -379,8 +379,11 @@ void WbSupervisorUtilities::processImmediateMessages(bool blockRegeneration) {
   mFieldSetRequests.clear();
   if (blockRegeneration)
     return;
+
   WbTemplateManager::instance()->blockRegeneration(false);
-  emit worldModified();
+  const QList<WbRobot *> &robots = WbWorld::instance()->robots();
+  if (robots.contains(mRobot))  // only emit if robot still exists (FieldSetRequest could have regenerated it)
```
omichel, line 385: Can we simplify it this way?
```suggestion
  if (WbWorld::instance()->robots().contains(mRobot))  // only emit if robot still exists (FieldSetRequest could have regenerated it)
```

omichel (commented):
Nice catch!

omichel (approved):
Thank you!

PR #3880, opened 2021-11-11
Fixed line endings

omichel:
A few files had Windows line ending (`CR` `LF`) or a mixture of Linux and Windows line endings. This PR fixes that to have only Linux line endings (`LF`).


ad-daniel:
I believe `driver.def` also has CRLF

ad-daniel (approved):
Thank you.

Issue #3917, opened 2021-11-16
Huge performance hit when using large ElevationGrid


ShuffleWire:
**Describe the Bug**
I'm using an ElevationGrid with large value (about 100 * 100 points). It work well (no issue to display that smoothly), but when I want to change an height, Webots take a will to update. No issue with smaller grid (ie the delay scale with the size of the grid).
I've been initially working with a supervisor to modify in real time the shape of the grid, and found the hard way that it cause crazy delay (if I try to update multiple point at once, Webots could take seconds to complete a step...).

**Steps to Reproduce**
1 - Open : 
<details>
<summary>World</summary>
<pre>#VRML_SIM R2021b utf8
WorldInfo {
  coordinateSystem "NUE"
}
Viewpoint {
  orientation -0.5846802669107566 -0.7637817352556263 -0.27347110702059674 1.6325835190324136
  position -11.007462585955334 19.873638859966295 5.441393012418699
}
TexturedBackground {
}
TexturedBackgroundLight {
}
Shape {
  geometry ElevationGrid {
    height [
      0, 0, 0, 0, 0, 0, 0, 0, 0, 0
    ]
    xDimension 100
    zDimension 100
  }
}
</pre>
</details>
2 - Change the first value of height (from 0 to 1, for instance)

3 - See that Webots take a while to update. (increase grid size if issue not clear enough). During that time one core is used at 100%.

**Expected behavior**
This modification should occur immediately.

**System**
 - Operating System: Debian 11
 - CPU: i9-8950HK
 - Webots master@7e71add8b994ec6f5c0ba3c1d9689f300c778e23 and same behaviour on develop@7fcfd87ec1fe4a9f77189c6754167b1aa9022eae

**Additional context**
After doing some timing, I found that the first value change cause a delay because of the call to `prepareGl();` from Mesh.cpp:35
However I can't find where is the bottleneck once the value have been changed once and we change it multiple time...
I've been profiling Webots, hence I provide the output, if anyone interested.
[output.zip](https://github.com/cyberbotics/webots/files/7544734/output.zip)
I've try to get to the heart of the issue (I was initially using a supervisor to get this behaviour, but that not actually needed), but at this point I'm not experienced enough with C++/Qt to dig more into that...

ad-daniel:
[...]

PR #3924, opened 2021-11-16
Fix performance degradation in `ElevationGrid`

ad-daniel:
**Description**
Fixes https://github.com/cyberbotics/webots/issues/3917.

- [x] **Open `CH_Vens.wbt`, select `height` field of elevation grid and click `reset to default value`. It takes forever.**
The issue here is related to how fields are reset. When resetting, the field is first emptied and then the default value is set. By removing items from the MF field one by one however results in the emission of the `changed` signal, everytime. What this means specifically for the `ElevationGrid` is that every time one element is removed, a call to `WbElevationGrid::updateHeight` is made which rebuilds the entire mesh. Do that for a 100x100 grid and things quickly get slow.
- [x] **Change of a single value in a big grid** 
- [x] **Degradation with other node (IndexedFaceSet)**


ShuffleWire:
What is the status on that ? I've just localy rebased on the master branch, and give it a try : the slowdown on CH_Vens is no more ! 
However, when I change one point at the time in the ElevationGrid I've provided you, there isn't any perceptible improvement :/ Scrolling up to change the value of the first value on the height lag a bit. It's not that fast as scrolling through value to change the x coordinate of an object, for instance. But for only 1 point, I would expect that to be significantly the same speed.
What else could I do to help you ? I'm forecasting to need this issue resolved in the near future, or at least know if I can rely on it.

ad-daniel:
Yes, only one of the issues has been addressed so far. I don't recall exactly, but if I remember right the other issue had something to do with shadows of the static mesh, essentially in [here](https://github.com/cyberbotics/webots/blob/master/src/wren/StaticMesh.cpp#L1817-L1916). More specifically I think the culprit was [this](https://github.com/cyberbotics/webots/blob/9c6469597a29f84dff5fcf05d3443d8fd38947f4/src/wren/StaticMesh.cpp#L1901) function, but I'm not 100% sure I remember what the problem was. I don't think it's a problem with the function itself, but rather that we shouldn't even reach it for elevation grids as we're not interested in having shadows (I think?) for the elevation mesh.

I'm not 100% on this, it's a bit murky at the moment, but I'll take a closer look to see if I remember exactly how far I went with tracking the issue down, but if you wish to investigate, I'd definitely start from this

ShuffleWire:
Ok, thank for the update !
TBF, getting into that code is a big step up, so I would have to spend a fair amount of time on it... Not sure if I can justify it... 
I will keep a eye on it, however. And work on it if I've time to spare.

omichel:
I finalized this PR, tested different cases of deleting of coordinates, and it seems pretty stable and fast in all the delete and edit operations I tested from the scene tree, including with the elevation grid and indexed face set provided by @ShuffleWire.

BenjaminDeleze (changes requested):
Updating value is very smooth.

However, when resetting value to default, "nothing" happens:

- The change in the scene tree is immediately visible but
- Nothing change in the world

In fact, I can modified the value of the `xDimension` for example and it will render with the old values of `height`.
The only way I found to reset the height in the world is to:
- Reset the default value
- Double click on `height` to insert a 0 value.

omichel:
The `changed` signal is indeed needed to update the 3D view. I restored it and it is working now as before.

BenjaminDeleze (approved):
Thank you!

PR #3926, opened 2021-11-17
Add warning when adding node after simulation has run

ad-daniel:
**Description**
Quite often it happens that the scene is modified only to discover towards the end of these modifications that the simulation wasn't at time zero. To warn against this sort of annoyance, to me this seems to be the less intrusive approach, compared to showing a dialog everything. Feedback is however welcome.

a

b


BenjaminDeleze (approved):
Seems good to me! Thank you.

PR #3930, opened 2021-11-17
Add `none` to physics plugin selector window

ad-daniel:
**Description**
Once set, it isn't possible to remove a physics plugin other than removing it from the world file.


omichel (approved):
Nice.

Issue #3915, opened 2021-11-15
wrong orientation of SUMO car when having pitch


joachimhgg:
**Describe the Bug**
 When a SUMO car is going up a slope oriented on pi, it switches direction. 

**Steps to Reproduce**
1. start the simulation of highway.wbt

**Measure**
Rotation given by SUMO to the car:
```
[-0.0, 0.0, -1.0, 2.890003531854632] **Flat terrain**
[-0.0, 0.0, -1.0, 2.9586541785522322]
[-0.0, 0.0, -1.0, 3.041949931865104]
[-0.04498478768056314, -0.0010825071104350933, -0.9989870855299835, 3.093522964396263] **slope**
[-0.04498481228943558, -0.00027879221857669644, -0.9989876320246328, 3.1292104114217336]
[0.04498481395497423, -6.063145936575398e-05, 0.9989876690116162, 3.138899742879164] **car switched direction**
[-0.04498481403766023, 0.0, -0.9989876708478423, 3.141592653589793]
[-0.07792093205630148, 0.0, -0.9969595419812568, 3.141592653589793]
[-0.07792093205630148, 0.0, -0.9969595419812568, 3.141592653589793]
[-0.07792093205630148, 0.0, -0.9969595419812568, 3.141592653589793]
[-0.08738838908878721, 0.0, -0.996174316800261, 3.141592653589793]
[-0.08738838908878721, 0.0, -0.996174316800261, 3.141592653589793]
[-0.08738838908878721, 0.0, -0.996174316800261, 3.141592653589793]
[0.08738837180561908, -0.0006289263444296139, 0.9961741197826894, 3.127254146061024]
[0.08738823356336733, -0.001886638223352062, 0.996172543905366, 3.0985862340869925]
[0.01549937691480403, -0.0005618995301711029, 0.999879719558393, 3.0691270228244134]
[0.015499375435341719, -0.0007117844123801999, 0.9998796241168556, 3.0498213211270055]
[0.015499373621294573, -0.0008606491438501741, 0.9998795070909287, 3.0306639346476563]
[0.0, 0.0, 1.0, 3.0310196468268304] **Flat terrain**
[0.0, 0.0, 1.0, 3.05361539197121]
[0.0, 0.0, 1.0, 3.0501803409844466]
[0.0, 0.0, 1.0, 3.1034945201328488]
[0.0, 0.0, 1.0, 3.112246294070845]
```

PR #3934, opened 2021-11-18
fix arccos boundaries

joachimhgg:
**Description**

This PR is related to these two issues #3556 and #3915 and the PR #3741. The bug came from the `clampedAcos` function in `WbMathsUtilities.hpp`
The boundaries of acos are [0,pi] for the values between [-1,1]. 
So the clamp should happen for `value` ]-1,1[ and take the value 0 or pi. However, for performance reason we can clamp between [-1,1].

Otherwise, The `NaN` issue can be reproduced by sending a rotation exactly equals to M_PI [0, 0, 1,3.14159265358979323].

In addition, I added a `ToString` function to `WbAxisAngle` Class and a default precision to the `toString` function of `WbRotation` Class.

### DIFF
--- a/src/webots/maths/WbMathsUtilities.hpp
+++ b/src/webots/maths/WbMathsUtilities.hpp
@@ -74,9 +74,9 @@ inline double WbMathsUtilities::normalizeAngle(double angle, double lastSpot = 0
 // If not then the behavior is similar to a clamped value.
 inline double WbMathsUtilities::clampedAcos(double value) {
   assert((fabs(value) < 1.0 + EPSILON) && "Value passed to clampedAcos out of range.");
-  if (value > 1.0)
+  if (value >= 1.0)
     return 0.0;
-  if (value < -1.0)
+  if (value <= -1.0)
     return M_PI;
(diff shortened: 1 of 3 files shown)


src/webots/maths/WbMathsUtilities.hpp
```diff
@@ -74,20 +74,20 @@ inline double WbMathsUtilities::normalizeAngle(double angle, double lastSpot = 0
 // If not then the behavior is similar to a clamped value.
 inline double WbMathsUtilities::clampedAcos(double value) {
   assert((fabs(value) < 1.0 + EPSILON) && "Value passed to clampedAcos out of range.");
-  if (value >= 1.0)
+  if (value > 1.0)
```
omichel, line 77: I would keep the `=` in the comparisons because it slightly speeds up the function by saving a call to the acos function.

omichel (changes requested):

omichel:
Please fix the clang-format failure.

ad-daniel (approved):
Thank you.

PR #3939, opened 2021-11-19
Fix rendering when switching from wireframe

### LINKED ISSUE #3938: Bad rendering when switching from wireframe mode to plain rendering
**Steps to Reproduce**
1. Load a world, for example `tiago++.wbt`
2. Put it in wireframe rendering
3. Reload the world
4. Switch to plain rendering
5. The world does not have correct lightning

**Screenshots**
tiago++

**System**
 - Operating System: Windows 10, Linux Ubuntu 20

**Additional context**
I tried to track the bug and found that by forcing `glPolygonMode(GL_FRONT_AND_BACK, polygonMode);` in https://github.com/cyberbotics/webots/blob/10f33c99863993f18071fb2e2862656fb8e4f5cb/src/wren/GlState.cpp#L360
to always be `GL_FILL` the bug does not appear. But I don't know what are the next steps to investigate.

BenjaminDeleze:
**Description**
The issue came from the brdf texture. This texture is only created once, so if we were starting webots in wireframe mode it was created only to be compatible with wireframe. Whereas if the brdf texture is create in plain rendering mode it is compatible for both mode.

To find this bug I used [RenderDoc](https://renderdoc.org/), it seems to be a very useful tool for digging into OpenGL issues.

**Related Issues**
Fix #3938 

Plain Rendering brdf          |  wireframe rendering brdf
:-------------------------:|:-------------------------:
good_brdf| wrong_brdf


omichel (approved):
Well spotted!

PR #3948, opened 2021-11-23
FLU: fix orthographic mode

BenjaminDeleze:
**Description**

- [x] Remove skybox rendering in orthographic mode
- [x] Remove shadows in orthographic mode

It seems like the orthographic mode was already converted. It could feel different depending on the `orthographicViewHeight`.
The `orthographicViewHeight` is a parameter that define the frustum height when in orthographic mode and that is modified when zooming in/out the scene. This parameter is automatically saved in the `.wbproj`. If we don't know that it is saved, the behviour can feel strange. For example if we:

1. Load the world and switch to orthographic mode.
2. Navigate the scene and zoom in/out.
3. Switch back to perspective mode and forget about orthographic.
4. Reload the world with the same viewpoint as in 1.
5. Switch to orthographic mode.

In 5. we will not see the same rendering as in 1. We will see the same viewpoint as in 1. __but__ with the zoom level of 2.


omichel (approved):
Thank you.

PR #3950, opened 2021-11-23
Fix crash when changing a world after the backward compatibility adapta…

### LINKED ISSUE #3946: Crash with the backward compatiblity with world R2021b open in Webots R2022a `feature-enu-plan`
**Describe the Bug**
If the following world created in `R2021b` is open with the `feature-enu-plan` branch (R2022a), Webots will crash 100% of the time.

**Steps to Reproduce**

1. Download this world [my_world_copy.zip](https://github.com/cyberbotics/webots/files/7581784/my_world_copy.zip)
2. Open it with the Webots `feature-enu-plan` branch (R2022a).
3. Try to open any other world or to create a new one.
4. Webots crashes.

**Additional context**
If the version of the world is changed from `R2021b` to `R2022a`, it will not crash anymore.

lukicdarkoo:
Fixes #3946


ad-daniel (dismissed):
Looks good to me.

It's unrelated to this PR but it appears that the backwards compatibility mechanism fails for this word, the flipping of the geometry in the bounding objects of the wheels is wrong (one of them is deleted entirely and the other isn't flipped). Might be worth opening an issue/investigating since it's a very simple robot

bug

lukicdarkoo:
I will fix it in this PR

lukicdarkoo:
I changed it, but it still has some problems - not all USE references are updated. If you save/load the world it looks fine.

ad-daniel:
a test appears to fail, is it because it isn't synched with the branch?

lukicdarkoo:
Synced with the parent branched and created an issue #3963

ad-daniel:
So do we need to revert the fix you attempted and just keep the fix for the crash or keep both?

lukicdarkoo:
> So do we need to revert the fix you attempted and just keep the fix for the crash or keep both?

There is just the crash fix. I will resolve the backward compatibility robustness in another PR as it turned out it is more complex than expected.

ad-daniel (approved):
Ah my bad, I must have mixed things up.
Thank you!

PR #3954, opened 2021-11-24
Enhancement-useNetconvert

joachimhgg:
**Description**

All our worlds but `Sumo_interface_example` are using useNetconvert as FALSE. The sumo_interface proto is not include by default in some of the worlds (`city_*`, `village_*`) so for the moment the user has to add the proto and put useNetconvert to FALSE. To use netConvert, the user has to setup a netConvert configuration file (sumo.netccfg) for which is not our "scenario tutorial". Indeed, a common way of implementing a SUMO interface is to create a net (so by using Netconvert outside webots).

Drawback: the user using the default value of useNetconvert will face to issue.


BenjaminDeleze (approved):
Thank you!

Issue #3945, opened 2021-11-22
Incorrect mapping with `Mesh` nodes


ad-daniel:
**Describe the Bug**
Importing 3d model from an obj file works correctly (it is converted to a `IndexedFaceSet`).
Loading the same obj into a Mesh node and applying the texture using a `ImageTexture` results in an incorrect mapping.

![testobj](https://user-images.githubusercontent.com/44834743/142843080-0d6b27d7-0f7e-4feb-aaf7-38b5a670ce1b.png)

If on the other hand if we do:
- take the `IndexedFaceSet` representation (save it into a separate PROTO file)
- convert it to obj using this script: https://github.com/cyberbotics/webots/blob/master/scripts/proto_splitter/proto2mesh.py
- load the obj into a `Mesh` node, the texture mapping works.

this might suggest that the issue is not with the Mesh node itself, but rather with the decoding/conversion that takes place. 
There's only like 2 examples in the library that use Mesh + Texture, but they appear to be working.

To reproduce: [testobj.zip](https://github.com/cyberbotics/webots/files/7580096/testobj.zip)

PR #3955, opened 2021-11-26
Correction of the texture mapping for the Mesh node

BenjaminHug8:
**Description**
The texture mapping with `Mesh` node was not working properly. This can be seen when using `.obj` files.

I just added an option to flip the `y` axis for the texture mapping as Wren might not be aligned with the convention of `OpenGl`.

`OpenGl` and `.obj` define the `y` axis up.

**Related Issues**
This pull-request fixes issue #3945

### DIFF
--- a/docs/reference/mesh.md
+++ b/docs/reference/mesh.md
@@ -11,14 +11,9 @@ Mesh {
 The [Mesh](#mesh) node represents a 3D shape imported from an external file created by a 3D modeling software.
 The [Mesh](#mesh) node can be used either as a graphical or as a collision detection primitive (in a boundingObject).
 Currently, the following formats are supported:
-  - [3D Studio mesh](https://wiki.fileformat.com/3d/3ds) files (.3ds).
-  - [Blender](https://www.blender.org/) files (.blend).
-  - [Biovision Hierarchy](https://en.wikipedia.org/wiki/Biovision_Hierarchy) files (.bvh).
   - [Collada](https://en.wikipedia.org/wiki/COLLADA) files (.dae).
-  - [Filmbox](https://en.wikipedia.org/wiki/FBX) files (.fbx).
   - [STL](https://en.wikipedia.org/wiki/STL_(file_format)) files (.stl).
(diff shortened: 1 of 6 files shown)


ad-daniel:
does this bugfix concern only `obj` files? By flipping it every time, don't we end up breaking the other [formats](https://cyberbotics.com/doc/reference/mesh) instead? Just wondering

BenjaminHug8:
> does this bugfix concern only `obj` files? By flipping it every time, don't we end up breaking the other [formats](https://cyberbotics.com/doc/reference/mesh) instead? Just wondering

Good question. As `.obj`, `.stl` and `.dae` are the three main formats used, we decided to only support these.
I will update this in the documentation and the code in another PR.

omichel:
> I will update this in the documentation and the code in another PR.

If related, the documentation should be update in the same PR.

BenjaminHug8:
> If related, the documentation should be update in the same PR.

I will do that in this PR then.

omichel:
The documentation was not updated in this PR. Can you please update it? Also the change log should include a line about this change,

BenjaminHug8:
> The documentation was not updated in this PR. Can you please update it? Also the change log should include a line about this change,

Yes of course!

docs/reference/mesh.md
```diff
@@ -11,14 +11,9 @@ Mesh {
 The [Mesh](#mesh) node represents a 3D shape imported from an external file created by a 3D modeling software.
 The [Mesh](#mesh) node can be used either as a graphical or as a collision detection primitive (in a boundingObject).
 Currently, the following formats are supported:
-  - [3D Studio mesh](https://wiki.fileformat.com/3d/3ds) files (.3ds).
-  - [Blender](https://www.blender.org/) files (.blend).
-  - [Biovision Hierarchy](https://en.wikipedia.org/wiki/Biovision_Hierarchy) files (.bvh).
```
ad-daniel, line 16: I don't think BVH should be dropped

BenjaminHug8:
@omichel may I ask you a review when you have time?

omichel (approved):
Thank you.

PR #3956, opened 2021-11-26
Add support for Collada references

lukicdarkoo:
Tasks:
- [x] Add ColladaShapes PROTO
- [x] Add a name flag to the Mesh node (`name` corresponds the `name` attribute in Collada files) 
  - [x] Test performances, Mesh vs IndexFaceSet (he version with Mesh nodes was way ahead of the IndexFacedSet version, so I didn't bother to quantify it)
- [x] Add documentation
- [x] Avoid circular dependencies
- [x] Add changelog
- [x] Fix crashes when choosing the file
- [x] Check if the sub-mesh exists

Docs:
- https://www.cyberbotics.com/doc/guide/object-shapes?version=feature-collada-mesh


omichel:
If too difficult to implement in Webots, we could also implement it as a PROTO that would instanciate a Group of Shape nodes with PBRAppearence and IndexedFaceSet.

lukicdarkoo:
> If too difficult to implement in Webots, we could also implement it as a PROTO that would instanciate a Group of Shape nodes with PBRAppearence and IndexedFaceSet.

It would certainly take a lot of time to implement such node in Webots. Creating a PROTO sounds like a good idea, but I am not sure what is a good way to generate PBRAppearence. Instead of IndexedFaceSet we can use Mesh and just pass .dae path to it

omichel:
> Instead of IndexedFaceSet we can use Mesh and just pass .dae path to it

This won't work as the Mesh node will merge all the submeshes together which is what we want to avoid. Or we should add an extra parameter to the Mesh node that indicates a submesh.

omichel:
To further clarify my idea, there are two things:
1) Add an "id" SFString parameter to the Mesh node that indicates a specific mesh inside a collada file (the "id" field will be ignored in case of OBJ or STL file).
2) Create a new ColladaShapes proto that will construct the collada shapes, including appearances and meshes, generating the following structure:
```js
Group {
  children [
    Shape {
      appearance PBRAppearance {
        # green appearance
      }
      geometry Mesh {
        url "collada.dae"
        id "1"
      }
    }
    Shape {
      appearance PBRAppearance {
        # red color
      }
      geometry Mesh {
        url "collada.dae"
        id "2"
      }
    }
    ...
  ]
}
```
Basically, the proto file will use Javascript code to open and load the content of the "collada.dae" file. It will extract the appearance properties and create the corresponding PBRAppearance node which will be associated with specific meshes inside the collada file.

Another option would be to create IndexedFaceSet nodes instead of Mesh nodes, but I fear this option would have a strong performance impact at loading time...

lukicdarkoo:
@omichel We have the prototype ready. I believe the solution is slightly better than the one we talked about, there is no code duplication, heavy lifting is done on the C++ side, and there are no many changes.

The Mesh optimization needs to be implemented (and current performance evaluated).

omichel:
That is a very nice idea to implement it this way.

lukicdarkoo:
@omichel I don't see a way to avoid circular dependency, please let me know if you have any idea.

omichel:
Can't we move `WbQjsCollada.cpp` into `nodes/utils`?

lukicdarkoo:
> Can't we move `WbQjsCollada.cpp` into `nodes/utils`?

I worked around it with a forward declaration. We typically forward declare only the class, but here I included a method as well.

src/webots/nodes/WbMesh.cpp
```diff
@@ -33,6 +33,7 @@
 
 void WbMesh::init() {
   mUrl = findMFString("url");
+  mName = findSFString("name");
```
ad-daniel, line 36: A connection like `connect(mName, &WbSFString::changed, this, &WbMesh::updateName);` should be added. We should probably also check that the name exists within the collada file.

docs/reference/changelog-r2021.md
```diff
@@ -11,6 +11,7 @@ Released on XX, XXth, 2021.
     - Increased the mouse wheel speed when zooming the 3D window ([#3565](https://github.com/cyberbotics/webots/pull/3565)).
     - Added speed vector output to GPS ([#3742](https://github.com/cyberbotics/webots/pull/3742)).
     - Added warning when attempting to add a node to an already started simulation ([#3926](https://github.com/cyberbotics/webots/pull/3926)).
+    - Added the [ColladaShapes](../guide/object-shapes.md#colladashapes) node that allows referencing Collada shapes and textures ([#3956](https://github.com/cyberbotics/webots/pull/3956)).
```
ad-daniel, line 14: This branch needs to be resynched with develop and the changelog entry should be added to `changelog-r2022.md` instead.

docs/reference/mesh.md
```diff
@@ -29,3 +30,6 @@ If the `url` value starts with `http://` or `https://`, Webots will get the file
 Otherwise, the file should be specified with a relative path.
 The same search algorithm as for [ImageTexture](imagetexture.md) is used (cf. [this section](imagetexture.md#search-rule-of-the-texture-path)).
 Absolute paths work as well, but they are not recommended because they are not portable across systems.
+
+The `name` field defines which sub-meshes are included.
```
ad-daniel, line 34: From the sub-meshes (plural) it leads to believe it's possible to specify more than one, but the parameter is of type `SFString`. Is it supposed to be possible to import say 2 meshes at once in the same ColladaShape or the actual options are: all of the meshes OR just one.

ad-daniel (changes requested):
Didn't test it yet, but got a few comments

omichel (changes requested):
I am not yet convinced that re-declaring the class is a good solution as I fear that this may cause memory leaks (because the actual object is not the one we believe it is, for example, I am not sure sizeof(WbQjsCollada) is the same when called from WbTemplateEngine.cpp and from WbQjsCollada.cpp). Also it introduces some code redundancy...

projects/shapes/protos/ColladaShapes.proto
```diff
@@ -0,0 +1,26 @@
...
+    }
+  >%
+
+Group {
+    children [
+        %< if (url !== '') { >%
+            %<= wbcollada.getVrmlFromFile(url) >%
+        %< } >%
+    ]
+}
```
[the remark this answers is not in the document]
lukicdarkoo: I used a style from the [WoodenBox](https://github.com/cyberbotics/webots/blob/master/projects/objects/factory/containers/protos/WoodenBox.proto) PROTO. Maybe we could reformat other PROTOs as well.

docs/reference/mesh.md
lukicdarkoo: It will include all sub-meshes with the given name (I am not sure if the same names are allowed). However, I was thinking to support the * (asterisk) wildcard, I am not sure if anyone will need it.

lukicdarkoo:
> I am not yet convinced that re-declaring the class is a good solution as I fear that this may cause memory leaks (because the actual object is not the one we believe it is, for example, I am not sure sizeof(WbQjsCollada) is the same when called from WbTemplateEngine.cpp and from WbQjsCollada.cpp). Also it introduces some code redundancy...

I don't keen on using it either as I am not aware of the side effects. As an alternative, we can put the header file in `core` and keep the implementation in `node/utils`. However, it will be also the first time we are separating .hpp/.cpp files.

docs/guide/object-shapes.md
```diff
@@ -1,5 +1,28 @@
 # Shapes
 
+## ColladaShapes
```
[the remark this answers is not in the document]
lukicdarkoo: I don't think it is needed as this is a PROTO, not a node

docs/guide/object-shapes.md
```diff
@@ -1,5 +1,28 @@
...
+
+The [ColladaShapes](#colladashapes) node represents a Collada shape (*.dae) imported from an external file.
+The [ColladaShapes](#colladashapes) node combines [Mesh](../reference/mesh.md) and [PBRAppearance](../reference/pbrappearance.md) nodes to preserve shapes and textures from Collada files. 
+The [ColladaShapes](#colladashapes) node can be used either as a graphical or as a collision detection primitive (in a boundingObject).
+If the file contains more than one mesh, the meshes will be merged into a single one.
```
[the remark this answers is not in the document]
lukicdarkoo: Maybe, it is more accurate `into a single one` -> `into a single shape`?

src/webots/nodes/WbMesh.cpp
lukicdarkoo: Good point, implemented

docs/guide/object-shapes.md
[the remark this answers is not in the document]
omichel: It's not a single shape, it's a Group of Shape nodes, so I would say:
"As Collada files may contain several meshes with material information, each mesh will be added as a Shape node inside a Group node."

docs/reference/mesh.md
omichel: I believe that if the `name` string is empty (default value), it should include all the meshes instead of none to be more user friendly. Therefore, the wildcard "*" would give the same behavior as the empty string, so maybe we can drop it for now and document only the behavior with the empty string. Otherwise, people will expect that strings like "*_arm" to work, e.g., match "left_arm" and "right_arm". In such a case I am not sure which material to use... The one of the right or left arm? So, I would postpone the implementation of the wildcards until we really need them.

omichel: I am not sure how Collada files are organized and if several meshes can use the same appearance, but it is the case, it would be more logical to merge all the meshes using the same appearance into the same IndexedFaceSet inside a Shape node. In that case the "name" parameter should refer to the appearance and not to the mesh... But I am not sure if appearance are named in Collada files...

projects/shapes/protos/ColladaShapes.proto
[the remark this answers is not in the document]
omichel: Yes, but we should do it in a separate PR.

omichel:
Can't we simply have some static method in WbQjsCollada to avoid creating a new singleton class like WbQjsColladaRelay?

lukicdarkoo:
> Can't we simply have some static method in WbQjsCollada to avoid creating a new singleton class like WbQjsColladaRelay?

Yes, perfect, thanks!

omichel (approved):
It looks very good now, thank you.

omichel (approved):
Great, thank you.

omichel:
@ad-daniel: we need your approval before we can merge this PR.

ad-daniel (approved):
Thank you.

PR #3960, opened 2021-11-29
Fix Radar sanity check and restrict field of view to 180°

ad-daniel:
**Description**
As far as I can tell, no 360°-radar sensors exist in the market currently. This feature is either achieved by combining 2 or 4 separate radars or by rotating it.
Currently, when setting a horizontal field of view to a value higher than PI, the detection breaks down because of [this](https://github.com/cyberbotics/webots/blob/1508457090e341e1bbadc81a7f019234dc9376ab/src/webots/nodes/utils/WbObjectDetection.cpp#L318-L333) snippet, instead of detecting objects in the frustum, it detects them in the slice it isn't supposed to (and occasionally certain random angles also give positive hits). This is because although in practice it was allowed, the detection logic behind the scenes wasn't meant for it:
- normals of the frustum planes are flipped
- bottom plane becomes the top plane behind the device, so distances should be adapted.

A more appropriate way of supporting a 360° radar in webots would be have 2 separate frustums, one for the front and another for the rear and do the detection in each separately, for now however I don't see the reason to do this change. Instead, the documentation and code was adapted to restrict the field of view to the actual range where it was supposed to be used in (180°).

Additionally, it seems that worlds containing radars with invalid settings (for example negative ranges) aren't actually detected/fixed at load time because the checks only occur if a value change is triggered, so this has been fixed.

### DIFF
--- a/src/webots/nodes/utils/WbObjectDetection.cpp
+++ b/src/webots/nodes/utils/WbObjectDetection.cpp
@@ -316,6 +316,8 @@ bool WbObjectDetection::computeBounds(const WbVector3 &devicePosition, const WbM
       }
     }
     // check distance between center and frustum planes
+    // Note: this sort of detection is only adapted for a field of view smaller than PI. If a larger FoV is desirable, the logic
+    // should be changed by having two separate frostums, more details here: https://github.com/cyberbotics/webots/pull/3960
     double distances[4];
     for (int j = 0; j < 4; ++j)
       distances[j] = frustumPlanes[j].distance(objectPosition);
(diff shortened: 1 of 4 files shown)


src/webots/nodes/utils/WbObjectDetection.cpp
```diff
@@ -316,6 +316,8 @@ bool WbObjectDetection::computeBounds(const WbVector3 &devicePosition, const WbM
       }
     }
     // check distance between center and frustum planes
+    // Note: this sort of detection is only adapted for a field of view smaller than PI. If a larger FoV is desirable, the logic
+    // should be changed by having two separate frostums, more details here: https://github.com/cyberbotics/webots/pull/3960
```
omichel, line 320:
```suggestion
    // should be changed by having two separate frustums, more details here: https://github.com/cyberbotics/webots/pull/3960
```

omichel (approved):
I assume there are no example in the distribution of Webots that rely on a field of view larger than PI.

ad-daniel:
No, the tests also only check on the front frustum.

PR #3801, opened 2021-10-20
New Python API

omichel:
Fixes #2385.

This PR aims at implementing a new python controller API for Webots replacing the previous one with the following advantages:

1. Don't depend on swig to generate the python library.
2. Add python type annotations.
3. More "pythonic" style: use of python getters and setters.
4. Use Python constructors to instantiate devices instead of the `Robot.get_device()` method.
5. Rely on libController instead of libCppController: simpler to maintain (especially data type conversions), less dependencies (like libstdc++) and more lightweight (no need to load libCppController, no overhead in calling C functions through C++ API).
6. Use of doc-string.
7. Various improvements of the API (including the supervisor API for managing nodes and fields).
8. We could get rid of the various lib/python37, lib/python38, lib/python39 folders and keep only one lib/python folder.
9. And probably more...

PR #3962, opened 2021-11-29
Fix lidar C API documentation

omichel:
As reported in https://github.com/cyberbotics/webots/pull/3801#issuecomment-980504699.


ad-daniel (approved):
Thank you.

PR #3965, opened 2021-11-30
Fix webots.js texture path

omichel:
Fixes the computation of the texture path which was missing and causing the loading of scenes with textures to fail on webots.cloud.

This is already updated on https://cyberbotics.com/wwi/R2022a/ and in  production on https://beta.webots.cloud


BenjaminDeleze (approved):
Thank you!

PR #3966, opened 2021-11-30
Disable css when not needed

BenjaminDeleze:
- [x] Disable the `animation.css` or `wwi.css` when no animation/stream/scene is opened.
- [x] Change general css rule targeting general `li` and `ul` to some specific `li` and `ul`


omichel (approved):
Thank you!

PR #3967, opened 2021-12-01
Improve web simulation

omichel:
The web simulation system currently has two shortcomings:

1. The URL scheme is unnecessarily complicated, we should drop the `webots://` URL in favor of a simple GitHub URL referring to a world file.
2. The simulation URL is not checked against an allowed/blocked list by the simulation server which basically expose the simulation server to any Webots simulation hosted on GitHub.

This PR addresses the following:
- [x] Remove remnants of Python 2 compatibility.
- [ ] Fix the URL scheme to make it simpler.
- [ ] Check the simulation URL against an allowed list.
- [ ] Remove the old `init` method used to load controllers in robotbenchmark.
- [x] Update the [documentation](https://cyberbotics.com/doc/guide/web-simulation?version=feature-improve-web-simulation).
- [x] Fixed [get_git_info.sh](https://github.com/cyberbotics/webots/blob/feature-enu-plan/scripts/get_git_info/get_git_info.sh) script which was not supporting URLs with `@` like `https://omichel@github.com/cyberbotics/webots.git`.


omichel:
Replaced by #3977 aiming at the develop branch.

PR #3968, opened 2021-12-01
Fix extrusion

### LINKED ISSUE #3656: Incorrect warping of in `Extrusion.proto`
**Describe the Bug**

1. Load world [extrusion_bug.zip](https://github.com/cyberbotics/webots/files/7131556/extrusion_bug.zip), it looks like:
a

2. Change last line of `spine` field from `0 20 0` to `0 20 5`. The extrusion warps as intended.
b

3. Set `z` coordinate back to 0.
4. Change last line of `spine` field from `0 20 0` to `5 20 0`. The warping direction is correct, but the short and long edges have been swapped.
c

Appears to be the case in 2021b already and therefore other PROTO like `Fence` and `CrashBarrier` are also affected.

### LINKED ISSUE #3665: Incorrect geometry generation of `CrashBarrier` (culprit is `Extrusion` technically)
**Describe the Bug**

To reproduce: 
1. Add a simple `CrashBarrier` to a world. Looks like this:
a
2. Change `path` field from `[0 0 0, 5 0 0]` to `[0 0 0, 5 1 0]`. The barries is expected to look similarly, but slanted.
b

I suspect it isn't the same bug as https://github.com/cyberbotics/webots/issues/3656 since I can't reproduce it by manually creating a paralleliped. This flipping only occurs if the profile of the extrusion is flat (i.e a surface instead of a volume). It appears to exist since 2020a at least, but since most worlds with `CrashBarriers` are flat, it never surfaced, but now that we're moving to ENU/FLU it's very evident indeed.

BenjaminDeleze:
**Description**
It seems like two cases were not implemented or incorrectly done:
- When the first spines were collinear.
- When all the extrusion was collinear.

**Related Issues**
Fix #3665 

**Tasks**
  - [x] Handle case when first spines are collinear.
  - [x] Handle case when all spines are collinear.
  - [x] Comment the proto to ease its readability. (Done by @joachimhgg )
  - [x] Update CrashBarrier.proto
  - [x] Correct orientation behaviour when there is only one orientation define. ` If orientation contain one value, it is applied at all spine points`
  - [x] fix protos containing collinear extrusion
  - [x]  check worlds containing collinear extrusion (hall.wbt, extrusion.wbt, highway_driving.wbt, CH_Morges.wbt, highway_overtake.wbt, village_center.wbt, village_realistic.wbt, village_winter.wbt, village.wbt)
  - [x] check osm_importer (barrier.py, river.py)
  - [x] check icon_studio

**Additional context**
It does not fix #3656 but I think that it may be the normal behaviour of extrusion as defined by VRML


joachimhgg:
[review_extrusion_protos.zip](https://github.com/cyberbotics/webots/files/7640745/review_extrusion_protos.zip)
World with all the protos which contain an Extrusion proto for review.

projects/objects/geometries/protos/Extrusion.proto
```diff
@@ -96,69 +105,83 @@ PROTO Extrusion [
...
           }
-          if (Y[i].x === 0 && Y[i].y === 0 && Y[i].z === 0) {
-            Y[i].x = 0;
-            Y[i].y = 0;
-            Y[i].z = 1;
-            console.info('Warning: orientation of the crossSection at the first spine point is undefined.');
+
+          if (Z[i].x === 0 && Z[i].y === 0 && Z[i].z === 0) { // if Z still colinear, it means that the entire spine is collinear.
+            // If the entire spine is collinear, the SCP is computed by finding the rotation of a vector along the positive Y-axis (v1)
+            //to the vector formed by the spine points (v2). The Y=0 plane is then rotated by this value.
```
ad-daniel, line 135:
```suggestion
            // to the vector formed by the spine points (v2). The Y=0 plane is then rotated by this value.
```

projects/objects/geometries/protos/Extrusion.proto
```diff
@@ -96,69 +105,83 @@ PROTO Extrusion [
...
+            // v2 is actually just our Y variable. }
+
+            //implementation taken from https://github.com/bradenmcd/openvrml/blob/45877f95a713cdfd8736a41073ed9a349db4c495/src/libopenvrml-gl/openvrml/gl/viewer.cpp#L2436
```
ad-daniel, line 138:
```suggestion
            // implementation taken from https://github.com/bradenmcd/openvrml/blob/45877f95a713cdfd8736a41073ed9a349db4c495/src/libopenvrml-gl/openvrml/gl/viewer.cpp#L2436
```

projects/objects/geometries/protos/Extrusion.proto
```diff
@@ -96,69 +105,83 @@ PROTO Extrusion [
...
-          Y[i].x = Y[i-1].x;
-          Y[i].y = Y[i-1].y;
-          Y[i].z = Y[i-1].z;
+      } else { // all other spines
+        Y[i].x = spine[i+1].x - spine[i-1].x;
+        Y[i].y = spine[i+1].y - spine[i-1].y;
+        Y[i].z = spine[i+1].z - spine[i-1].z;
+        Z[i] = wbvector3.cross(wbvector3.minus(spine[i+1], spine[i]), wbvector3.minus(spine[i-1], spine[i]));
+
+        //if colinear the value from the previous point is used instead.
```
ad-daniel, line 168:
```suggestion
        // if colinear the value from the previous point is used instead.
```

projects/objects/geometries/protos/Extrusion.proto
```diff
@@ -96,69 +105,83 @@ PROTO Extrusion [
...
+          Z[i].x = Z[i-1].x;
+          Z[i].y = Z[i-1].y;
+          Z[i].z = Z[i-1].z;
         }
       }
+
       if (i !== 0) {
-        if (wbvector3.dot(Y[i], Y[i-1]) < 0)
-          Y[i] = wbvector3.multiply(Y[i], -1);
+        //After determining the Z-axis, its dot product with the Z-axis of the previous spine point is computed. If this value is negative, the Z-axis is flipped (multiplied by -1).
```
ad-daniel, line 177:
```suggestion
        // after determining the Z-axis, its dot product with the Z-axis of the previous spine point is computed. If this value is negative, the Z-axis is flipped (multiplied by -1).
```

scripts/converter/convert_nue_to_enu_rub.py
```diff
@@ -335,8 +336,8 @@ def convert_nue_to_enu_world(filename, mode='all', objects_pi=[], objects_pi_2=[
         error_verbose, clean_verbose, warning_verbose = convert_nue_to_enu_world(
             filename, mode, objects_pi, objects_pi_2, objects_minus_pi_2)
         if error_verbose:
-            print('Conversion of \033[33m{}\033[m successfull but incomplete. \r\n■ Need to convert manually: {}'.format(
-                filename, error_verbose))
+            print('''Conversion of \033[33m{}\033[m successfull but incomplete. \r\n■ You may have the need to convert
```
ad-daniel, line 339:
```suggestion
            print('''Conversion of \033[33m{}\033[m successful but incomplete. \r\n■ You may need to convert
```

projects/objects/geometries/protos/Extrusion.proto
```diff
@@ -96,69 +105,83 @@ PROTO Extrusion [
...
+          Y[i].x = spine[1].x - spine[nbSpinePoints-2].x;
+          Y[i].y = spine[1].y - spine[nbSpinePoints-2].y;
+          Y[i].z = spine[1].z - spine[nbSpinePoints-2].z;
+          Z[i] = wbvector3.cross(wbvector3.minus(spine[1], spine[0]), wbvector3.minus(spine[nbSpinePoints-2], spine[0]));
+        } else if (i === 0) { // open spine and first spine
+          Y[i].x = spine[1].x - spine[0].x;
+          Y[i].y = spine[1].y - spine[0].y;
+          Y[i].z = spine[1].z - spine[0].z;
+
+          //The Z-axis used for the first spine point is the same as the Z-axis for spine[1] => Z[0] = Z[1]
```
ad-daniel, line 118:
```suggestion
          // the Z-axis used for the first spine point is the same as the Z-axis for spine[1] => Z[0] = Z[1]
```

ad-daniel (approved):
Looks good to me, thank you

ad-daniel (approved):

PR #3977, opened 2021-12-06
Improve web simulation

omichel:
The web simulation system currently has two shortcomings:

1. The URL scheme is unnecessarily complicated, we should drop the `webots://` URL in favor of a simple GitHub URL referring to a world file.
2. The simulation URL is not checked against an allowed/blocked list by the simulation server which basically expose the simulation server to any Webots simulation hosted on GitHub.

This PR addresses the following:
- [x] Remove remnants of Python 2 compatibility.
- [x] Fix the URL scheme to make it simpler.
- [x] Check the simulation URL against an allowed list.
- [x] Remove the old `init` method used to load controllers in robotbenchmark.
- [x] Update the [documentation](https://cyberbotics.com/doc/guide/web-simulation?version=feature-better-web-simulation).
- [x] Fix [get_git_info.sh](https://github.com/cyberbotics/webots/blob/feature-enu-plan/scripts/get_git_info/get_git_info.sh) script which was not supporting URLs with `@` like `https://omichel@github.com/cyberbotics/webots.git`.
- [x] ~~Fix the firejail command line options which were allowing a controller to modify files in another controller folder.~~
- [x] Use docker instead of firejail to containerize webots and controllers.
- [x] Removed SSL certificates options from the session and simulation servers (now obsolete since portRewrite is the preferred method).

I have setup a https://beta.webots.cloud/demo to use the simulation server from this PR.
You can add more simulations to test it (beware that the Webots version in the `.wbt` file should be R2022a, however in theory it should also work for another version for which a docker image is available on dockerhub, but I didn't tested it, if it works, it may take long to start the simulation the first time as it will download the corresponding docker image).


docs/guide/web-simulation.md
```diff
@@ -57,15 +55,15 @@ cd $WEBOTS_HOME/resources/web/server
 The session server keeps a track of the available simulation servers and assigns a connection to the most suitable simulation server (similar to a load balancer).
 A task of the simulation server is to start a Webots instance with the correct world.
 
-To show the user interface simply open the `$WEBOTS_HOME/resources/web/streaming_viewer/index.html` file in your browser.
+To show the user interface, simply open the `$WEBOTS_HOME/resources/web/streaming_viewer/index.html` file in your browser.
 In the user interface, find a `Connect to` field, and type for example:
 ```
 ws://localhost:1999/session?url=webots://github.com/cyberbotics/webots/branch/develop/projects/languages/python/worlds/example.wbt
 ```
```
BenjaminDeleze, line 62: Not updated

BenjaminDeleze (changes requested):
Seems good, just some remarks:

It worked well on the given examples on beta.webots.cloud. But I did not managed to make the quick start work.

It would also be good to better document the risk of not using docker and how to mitigates them (whitelist)

BenjaminDeleze (changes requested):
The documentation is clearer now!

When running the quick start example, I went to the monitoring page of the session server
local

If I click on the link of the simulation server, it redirect me to: `https://localhost/2000/monitor` which does not work. It should redirect to: `http://localhost:2000/monitor`.

Moreover, even so I can see my simulation server on `http://localhost:2000/monitor`, its load is still indicated as N/A in the session server monitoring page (as in the picture) and so I cannot open a simulation from index.html (`Session server error: No simulation server available at the moment.`)

omichel:
I re-tested and it works fine for me. Are you sure you are up-to-date with the latest changes? (`git fetch; git pull`)?

BenjaminDeleze (approved):
Retested this morning and it is all fine. 
Thank you!

PR #3978, opened 2021-12-06
Clarify `WbSkin` model supported formats

ad-daniel:
**Description**
As per [this](https://github.com/cyberbotics/webots/blob/733b1bb2f5df73a95f61bb7b2ab5b45b8f772f82/src/webots/nodes/WbSkin.cpp#L223), FBX is the only supported file type.


omichel (approved):
Thank you.

PR #3980, opened 2021-12-07
Fix objects after supervisor move

### LINKED ISSUE #2899: SoccerBall is moving after setting its position and resetting velocity

BenjaminDeleze:
**Description**
When an solid if moved by the supervisor, it sinks in the floor for one step after being move.

**Related Issues**
Fix #2899

**What is happening**

_Step x_
The supervisor send the following commands:
```
ball_translation_field.setSFVec3f([random_x, 0.0209205, random_z])
ball.resetPhysics()
```
_Step x+1_

- `resetPhysics` disable the body of the object.
- Start of world step.
  - The translation from the supervisor is applied.
  - Start of the physic step (WbSimulationCluster.cpp).
    - `dWorldStepAndSpaceCollide` (ode_MT.cpp).
      - `dWorldStep`: ode step function (moves objects).
      -  `odeNearCallback`: compute the collisions (WbSimulationCluster.cpp).
  - Start of the post physic step.
    - The body of the object is enable again.
    - `applyPhysicTransform`: the physics results are applied to the object (WbSolid.cpp)
  - End of post physic step.
- End of world step.

As the body is disable by `resetPhysics`, the collision of the object with the floor is not computed during one step.Thus, during the next step, gravity is applied to the object and it sinks in the floor.

**Issue not always reproducible**
The issue was not always reproducible (on Ubuntu in particular), after investigation it was due to this portion of code in `odeNearCallback`: 
```
  // if the geometry of one colliding object has changed this step, we need to enable the other body
  // (otherwise if the changed geometry is kinematic the other body will not move)
  if (odeGeomData2->lastChangeTime() > 0.0) {
    if (odeGeomData2->lastChangeTime() >= WbSimulationState::instance()->time()) {
      if (b1)
        dBodyEnable(b1);
    } else
      odeGeomData2->setLastChangeTime(0.0);
  }

  // no need to create contact joints for contacts between disable bodies and kinematic solids
  const bool b1Disabled = b1 && !dBodyIsEnabled(b1);
  const bool b2Disabled = b2 && !dBodyIsEnabled(b2);

  if ((s1->isKinematic() && b2Disabled) || (s2->isKinematic() && b1Disabled) || (b1Disabled && b2Disabled))
    return;
```

`odeGeomData2->lastChangeTime()` was not initialized in the constructor of `WbOdeGeomData` and so it was possible to have large values. Due to these values, the body was re-enable just before the check of either or not the body was disable.

**Why `resetPhysics` is necessary**
If we remove the call to `resetPhysics`, the body is not disable, thus the contact points will be computed. But it will still not work and even be worse. Because in ODE, we first calculate the new position of the object based on the previous contact points and then update the collisions. So the object will be translated directly in the floor (but with contact points) because gravity will be applied and the previous contact points would not correspond anymore to the position of the object

**Possibles fixes**
I see two possible ways of fixing it:

- `resetPhysics` normally also reset the forces applied to the object so there is maybe something done wrong there. I did not investigate this yet.
- Re-enable the body between `dWorldStep` and `odeNearCallback` as it was done when the initialization of `mLastChangeTime`was done wrong. But it seems not trivial to get it only in the right case.

**What need to be check**

- [ ] Verifiy that the fix works also for multithreading (bug was present in both single thread and multi thread but for simplicity reason I investigated only single thread)
- [ ] Check that nothing is broken by the fix

### DIFF
--- a/src/webots/engine/WbSimulationCluster.cpp
+++ b/src/webots/engine/WbSimulationCluster.cpp
@@ -45,6 +45,8 @@
 #include <cassert>
 #include <limits>
 
+#include <iostream>
+
 // "webots" where each character is replaced by its ascii hexadecimal number.
 const long long int WbSimulationCluster::WEBOTS_MAGIC_NUMBER = 0x7765626F7473LL;
 
@@ -100,6 +102,7 @@ void WbSimulationCluster::swapBuffer() {
 }
 
 void WbSimulationCluster::step() {
+  std::cout << "cluster step" << '\n';
   dSpaceUpdateFunction *spaceUpdateFunc = NULL;
   if (WbSolidDevice::hasDirtySensors())
     // update rays position after world step and before space collision detection
@@ -127,6 +130,7 @@ void WbSimulationCluster::step() {
   // qDebug() << "number of physics plugin contacts 2:" << dJointGroupGetCount(mContext->physicsPluginContactJointGroup2());
   // qDebug() << "number of contacts 1:" << dJointGroupGetCount(mContext->contactJointGroup1());
   // qDebug() << "number of contacts 2:" << dJointGroupGetCount(mContext->contactJointGroup2());
+  std::cout << "end of cluster step" << '\n';
 }
 
 dSpaceID WbSimulationCluster::space() const {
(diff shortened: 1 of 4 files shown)


whtqh:
***@***.***谢谢！王海涛

BenjaminDeleze:
Replaced by #4070

PR #3987, opened 2021-12-07
Fixes the poor trajectory of the robot in battery.wbt sample world

omichel:
The robot was getting stuck into the first charger.
This PR fixes its behavior so that it goes now through all the chargers.


omichel:

PR #3988, opened 2021-12-07
Fixed robot trajectory

omichel:
It is nice to show the robot both leave and re-enter the receiver/emitter area which wasn't the case with the current demo where it was just leaving it but never re-entering it. This PR makes that the robot first leaves and then re-enters the receiver/emitter area.


omichel:

PR #3993, opened 2021-12-07
Fixed trajectory of robot in pen.wbt demo

omichel:
The robot was getting stuck in the first corner...


ad-daniel:
Still gets stuck on the 3rd corner on ubuntu 20.

omichel:
> Still gets stuck on the 3rd corner on ubuntu 20.

It's better than the first one, but if you can propose a quick fix to improve that, I will happy to review it.

omichel:

PR #3998, opened 2021-12-07
Fixed ODE warning in track.wbt sample demo

omichel:
To avoid the ODE warning proposing to reduce the time step, the time step of this demo was reduced.


BenjaminDeleze (approved):
Thank you

PR #4001, opened 2021-12-08
Fix name clash with empty string or default name for the Solid node 

BenjaminHug8:
**Description**
In case siblings Solid nodes have the same name a warning will appear in the console of Webots.

This warning should not appear if `the name is an empty string or the default name` AND `the node is not a device`.


ad-daniel:
Won't this change clash with [this](https://cyberbotics.com/doc/reference/solid#unique-solid-name)  ?

BenjaminHug8:
> Won't this change clash with [this](https://cyberbotics.com/doc/reference/solid#unique-solid-name) ?

As discussed with @omichel I will close this PR as we need unique names for `Solid` nodes also to visualize center of mass, center of buoyancy and support polygons.

PR #3708, opened 2021-09-23
FLU: Convert Clearpath robots

lukicdarkoo:
- [x] (602) robots/clearpath/moose/protos/Moose.proto
- [x] (603) robots/clearpath/moose/protos/MooseWheel.proto
- [x] (604) robots/clearpath/pr2/protos/Pr2.proto
- [x] (236) projects/robots/clearpath/moose/worlds/moose_demo.wbt
- [x] (237) projects/robots/clearpath/pr2/worlds/pr2.wbt

PR #4004, opened 2021-12-08
Fix PR2 meshes

omichel:
The meshes of the PR2 robot were referring to local files which are not distributed.
This PR fixes this problem by using the `webots://` URL for meshes.
However, the textures are still messed-up... That might have been introduced in #3708.


lukicdarkoo (dismissed):
Thank you!


> However, the textures are still messed-up...

Before the meshes were IndexFaceSet and the whole model was like 10-20MB, so I really wanted to move it out of the Webots package. I used the tool to convert the IndexFaceSet to meshes, but I didn't observe any problems with meshes. Can you point me to the problematic textures?

omichel:
It seems that almost all the texture mappings are messed up:
pr2

BenjaminDeleze:
The PR-2 was still fine 20 days ago so I think that the problem comes from the changes introduced in https://github.com/cyberbotics/webots/pull/3955.

I will have a look at it.

BenjaminDeleze:
I just test to remove the fix and it is working again.

So I will modify the proto2mesh script and export again the files.

I am a bit afraid that the fix of #3955 broke other .obj files that was exported using the proto2mesh script before

BenjaminDeleze:
I checked the other worlds/PROTOS that use .obj and it seems PR2 was the only one with issues

lukicdarkoo (approved):

PR #4006, opened 2021-12-08
Fix tinkerbots_demo.wbt

### LINKED ISSUE #3983: tinkerbots_demo.wbt is broken
The conveyor belt is moving in the wrong direction, breaking the demo.

lukicdarkoo:
Fixes #3983


omichel (approved):
Thank you.
I also fixed a not very nice self-collision of the robot arm.

PR #4009, opened 2021-12-08
Wiki-enu-flu-conversion

joachimhgg:
**Description**
Wiki of the ENU/FLU conversion.
This PR is not intented to be merge as the doc will be added to the wiki. It is just to make it easier to review it. Images and links (to the scripts, etc) will be added in the wiki.

PR #4013, opened 2021-12-08
Avoid applying the backwards compability fix when USE/DEF is used

### LINKED ISSUE #3963: Backward compatibility loader doesn't work for a simple robot
More details:
https://github.com/cyberbotics/webots/pull/3950#pullrequestreview-814595020

World file:
[my_world_copy.zip](https://github.com/cyberbotics/webots/files/7581784/my_world_copy.zip)

lukicdarkoo:
Fixes #3963 (including the `sick.wbt` crash)

The USE/DEF mechanism makes a huge mess, I am not sure how to fix it. I am afraid of ending up in a rabbit hole and wasting too much time.

I will give it one short try, but I don't have a quick solution for e.g. `DEF Shape > Cylinder`. The idea is to convert it to `Transform > Shape > Cylinder`, convert all USE nodes to ordinary nodes and then convert them, but that doesn't work properly :/


joachimhgg (commented):
Thanks it is not crashing anymore. In addition, I think it would be better to have a WBmessageBox with an Ok button which indicates that the `.wbt` world file will be convert, that the user may have to change it manually, can save the changes to validate the conversion and can refer to the wiki (see #4009). 

I find too much repetitive information are shared in the console, could we resume them by a list of proto instead of one message for each? The strategies should be documented.

lukicdarkoo:
> I find too much repetitive information are shared in the console, could we resume them by a list of proto instead of one message for each? The strategies should be documented.

We mention from which PROTO/world file does it come from, but we have to list all of them because the user should know what is changed.

> In addition, I think it would be better to have a WBmessageBox with an Ok button which indicates that the .wbt world file will be convert, that the user may have to change it manually, can save the changes to validate the conversion and can refer to the wiki (see #4009).

It is a good idea, but what if the world is 2022a, but PROTOs inside are 2021b. In that case, there will be message boxes showing every time for each PROTO? I guess it is enough to print a warning in the console with the wiki link.

omichel:
Yes, we should find a good compromise between providing the right information without annoying too much the users.

joachimhgg (approved):
As discussed, we can create another PR with an updated dialog which will be launch at the first start of Webots to explain the ENU/FLU conversion. Thanks!

PR #4015, opened 2021-12-09
FLU wiki with focus on backward compability

lukicdarkoo:


lukicdarkoo:
@joachimhgg This is my suggested framework:
- The most important part is backward compatibility. Therefore, we stress it, give a clear changelog, and how to adapt worlds/PROTOs.
- Following FLU/ENU is optional and we should provide it as a recommendation. We give an introduction and point to tools that users can use to convert worlds and objects.

Let me know what do you think.

joachimhgg (approved):
I agree with you to focus on backward compatibility and add a recommendation section. I will implement your part with mine. Just a question remains about the scripts, we have one script for the proto, one for the nue to enu and one for both (#4012) but mostly tested on worlds. The main difference are that the first two scripts are adding a rotation until the last one are changing the rotation axis (e.g. `rotation 1 0 0 1.57` -> `0 -1 0 1.57` ). What do you think about it?

PR #4019, opened 2021-12-10
Fixed layout perspective of hinge joint with backlash demo

### LINKED ISSUE #3990: Bad window layout for hinge_joint_with_backlash.wbt
The default windows layout is bad in the hinge_joint_with_backlash.wbt sample device demo, we cannot see the blue part of the system:
image
We should probably reduce the size of the text editor window.

omichel:
Fixes #3990.
image


ad-daniel (approved):
Thank you.

PR #4020, opened 2021-12-10
Fix icon in menu linux mac

joachimhgg:
**Description**

- The icons in the menu bar were not showing up in Linux and Mac. I removed the lines causing the issue. I think it was an old fix but does not seems relevant anymore. Need to be test on Mac

- add a padding for the icon on the night theme.
- fix the `Yes` issue on Windows when (Tools/Preferences/change theme/ok)
- apply the same change for dusk
- load the webots_night.qss by default on linux


omichel:
On Windows, the window bars remains white. It can be turned to black (for dark themes) this way: https://envyen.com/posts/2021-10-24-QT-Windows-Dark-theme/

ad-daniel (dismissed):
Cool, thank you!

omichel:
As discussed yesterday, we should also get rid of the two useless menu items in the Edit menu, as one of them is causing some overlap problem.

src/webots/user_commands/WbActionManager.cpp
```diff
@@ -679,26 +676,12 @@ void WbActionManager::populateActions() {
   mActions[GO_TO_LINE] = action;
 
   action = new QAction(this);
-  action->setText(tr("&Toggle Line Comment"));
+  action->setText(tr("&Toggle Line Comment    "));
```
ad-daniel, line 679: Are these spaces necessary?

ad-daniel (changes requested):
Didn't even know these options existed

src/webots/user_commands/WbActionManager.cpp
joachimhgg: Yes, the padding between the text and the keyboard shortcut are miscalculated, and the easiest solution to prevent this without affecting the other items is to add space

omichel: In that case, we should add a comment about it.
And maybe exclude the spaces from the translation, e.g., `tr("&Toggle Line Comment") + "      ")`.

src/webots/gui/WbGuiApplication.cpp
```diff
@@ -540,8 +540,84 @@ void WbGuiApplication::loadInitialWorld() {
...
+#endif  // _WIN32
+
 void WbGuiApplication::udpateStyleSheet() {
-  QString themeToLoad = WbPreferences::instance()->value("General/theme", "webots_classic.qss").toString();
+  QString themeToLoad;
+#ifdef __linux__
+  themeToLoad = WbPreferences::instance()->value("General/theme", "webots_night.qss").toString();
+#else
+  themeToLoad = WbPreferences::instance()->value("General/theme", "webots_classic.qss").toString();
+#endif
```
omichel, line 619:
```suggestion
#ifdef __linux__
  const QString themeToLoad = WbPreferences::instance()->value("General/theme", "webots_night.qss").toString();
#else
  const QString themeToLoad = WbPreferences::instance()->value("General/theme", "webots_classic.qss").toString();
#endif
```

omichel (changes requested):

omichel (approved):

ad-daniel (approved):
The linux side of things works as intended, thank you.

PR #4023, opened 2021-12-13
Fix cylinder of linear_motor.wbt

BenjaminDeleze:
ditto


omichel (approved):
Well spotted! I added a couple of minor adjustments.

PR #4026, opened 2021-12-13
Fix backward compability crash for DEF nodes

### LINKED ISSUE #3963: Backward compatibility loader doesn't work for a simple robot

lukicdarkoo:
Fixes #3963

- Non-relevant change. I deleted `if (!WbWorld::instance()->isLoading())` as it seems that doesn't makes sense as there is an assert above.


BenjaminDeleze:
I tested a bit and I cannot reproduce the crashes anymore

joachimhgg (approved):
As Benjamin, I cannot reproduce the crash anymore. Thanks!

PR #4027, opened 2021-12-13
Fix ratslife crash

### LINKED ISSUE #3982: Crash of ratslife.wbt
**Describe the Bug**
`ratslife.wbt` crash when running the `contest_manager`

It seems to come from #3830.

BenjaminDeleze:
**Description**
The problem came from the fact that we were using the url to find the image in the `gImagesMap`. It is fine and handy to retrieve an image from a given url, but when we want to remove an image it does not work.

It does not work because the workflow goes as follow:

1. We/supervisor change the value of an url of a `imageTexture`.
2. The value of the url of the `imageTexture` is effectively changed.
3. A signal is emitted telling `imageTexture` that the url has changed.
4. `destroyWrenTexture` is called to clean the old image.
5. It was using the url to lookup the map. But the url was already the **new** url, so it was deleting the wrong image

Now it iterates through the images (an not through urls) when we want to delete an image.

Fix #3982


src/webots/nodes/WbImageTexture.cpp
```diff
@@ -272,16 +272,21 @@ void WbImageTexture::destroyWrenTexture() {
...
+    if (image && image == mImage) {
+      QString key = i.key();
+      const int instances = i.value().second - 1;
+      assert(instances >= 0);
+      if (instances == 0) {
+        delete mImage;
+        gImagesMap.remove(key);
+      } else {
+        gImagesMap[key] = std::make_pair(image, instances);
+      }
```
omichel, line 288: Coding style:
```suggestion
      } else
        gImagesMap[key] = std::make_pair(image, instances);
```

src/webots/nodes/WbImageTexture.cpp
```diff
@@ -272,16 +272,21 @@ void WbImageTexture::destroyWrenTexture() {
...
-      delete mImage;
-      gImagesMap.remove(url);
-    } else
-      gImagesMap[url] = std::make_pair(pair.first, instances);
+  QMapIterator<QString, std::pair<QImage *, int>> i(gImagesMap);
+  while (i.hasNext()) {
+    i.next();
+    QImage *image = i.value().first;
+    if (image && image == mImage) {
+      QString key = i.key();
```
omichel, line 280:
```suggestion
      const QString key = i.key();
```

src/webots/nodes/WbImageTexture.cpp
```diff
@@ -272,16 +272,21 @@ void WbImageTexture::destroyWrenTexture() {
...
-    assert(instances >= 0);
-    if (instances == 0) {
-      delete mImage;
-      gImagesMap.remove(url);
-    } else
-      gImagesMap[url] = std::make_pair(pair.first, instances);
+  QMapIterator<QString, std::pair<QImage *, int>> i(gImagesMap);
+  while (i.hasNext()) {
+    i.next();
+    QImage *image = i.value().first;
```
omichel, line 278:
```suggestion
    const QImage *image = i.value().first;
```

omichel (changes requested):

ad-daniel (approved):
Nice catch, no issues with the memory leak either. Thank you.

omichel (approved):
Looks good to me. Thank you.

PR #4031, opened 2021-12-14
Fix example worlds

BenjaminDeleze:
ditto


ad-daniel (approved):
Much better, thank you

Issue #4022, opened 2021-12-12
There may exist a memory leakage issue


FreshNing:
**Describe the Bug**
When I run the example in "${WEBOTS_HOME}\projects\samples\howto\supervisor_draw_trail", the memory consumption continuously increase.  Gradually，the simulation becomes slow and jerky. This problem is even more pronounced when I try to draw more trails. Additionally, referring to the example,  I utilize 'EvaluationGrid' to create a height map. However, it rendered slowly.

**Steps to Reproduce**
1. 'Simulation Reset'  would not release the increased memory
2. A function 'importMFNodeFromString()' was adopted
3. Large amounts of 'Field' modifications, like 'translation' 'rotation' 'height' ...
4. Other 'Node' operations, get Node from DEF and create children Node ...
5. Emitter and Receiver Nodes were used to make comunications between robot and supervisor robot 

**Expected behavior**
Find out the reasons and solutions

**System**
 - Operating System: Windows 10
 - Graphics Card: NVIDIA GeForce GTX 960M 2 GB
 - Webots Version: 2021b

**Additional context**
Wish for replies, thanks!

BenjaminDeleze:
I could reproduce the issue on ubuntu 20 with `supervisor_draw_trail.wbt`.

PR #4038, opened 2021-12-16
Fix-memory-leak-gl-buffer

joachimhgg:
**Description**

The buffer of the shadow coords was not release which led to memory leak.

**Related Issues**
This pull-request fixes issue #4022


src/wren/StaticMesh.cpp
```diff
@@ -1933,6 +1933,9 @@ namespace wren {
 
         if (mCacheData->mGlNameBufferUnwrappedTexCoords)
           glDeleteBuffers(1, &mCacheData->mGlNameBufferUnwrappedTexCoords);
+
+        if (mCacheData->mGlNameBufferShadowCoords)
+          glDeleteBuffers(1, &mCacheData->mGlNameBufferShadowCoords);
```
omichel, line 1938: I believe the test is not needed as this buffer is systematically created (similar to `mGlNameBufferIndices`):
```suggestion
        glDeleteBuffers(1, &mCacheData->mGlNameBufferShadowCoords);
```

omichel (changes requested):
Well spotted! 🍾

omichel (approved):

PR #4039, opened 2021-12-16
Fix-default-stylesheet

joachimhgg:
**Description**

Set default value of Stylesheet for each OS. Fix bug with linux stylesheets for the first launch.

clean your preferences before to test:

`python3 scripts/preferences_cleaner/preferences_cleaner.py`


src/webots/gui/WbGuiApplication.cpp
```diff
@@ -611,13 +611,8 @@ static void setDarkTitlebar(HWND hwnd) {
...
 void WbGuiApplication::udpateStyleSheet() {
-#ifdef __linux__
-  const QString themeToLoad = WbPreferences::instance()->value("General/theme", "webots_night.qss").toString();
-#else
-  const QString themeToLoad = WbPreferences::instance()->value("General/theme", "webots_classic.qss").toString();
-#endif
-
-  QFile qssFile(WbStandardPaths::resourcesPath() + themeToLoad);
+  mThemeLoaded = WbPreferences::instance()->value("General/theme").toString();
+  QFile qssFile(WbStandardPaths::resourcesPath() + mThemeLoaded);
```
BenjaminDeleze, line 615: Is it normal that we never close the `qssFile`?

BenjaminDeleze (commented):
Seems good, just a question.

src/webots/gui/WbGuiApplication.cpp
joachimhgg: `QFile` object is closed automatically upon its destructor, hence we do not need to call `close()` explicitly.

BenjaminDeleze (approved):

BenjaminDeleze:
Thank you

PR #4040, opened 2021-12-16
Fix revert core18

BenjaminDeleze:
- Revert to the dependencies of core18
- Add some missing ros libraries
- Fix sumo


omichel (approved):
Thank you.

PR #4055, opened 2021-12-20
Switch mSolid endPoint pointer to NULL when using static environnement

ShuffleWire:
I'm not sure about that, but I've encountered this while trying to fix another bug.
I'm not sure about the repercussion of it, so please give me any feedback :)

Without this, when switching to "static environment" in joint endpoint, the mSolid pointer stay with the same value. I don't think it's the expected behavior. But not sure if setting it to NULL is the way to go !

### DIFF
--- a/src/webots/nodes/WbSolidReference.cpp
+++ b/src/webots/nodes/WbSolidReference.cpp
@@ -22,7 +22,6 @@
 const QString WbSolidReference::STATIC_ENVIRONMENT = QString("<static environment>");
 
 void WbSolidReference::init() {
-  mSolid = NULL;
   mName = findSFString("solidName");
 }
 
@@ -68,7 +67,7 @@ void WbSolidReference::updateName() {
 }
 
 bool WbSolidReference::isClosedLoop() const {
-  if (!mSolid)
+  if (mSolid.isNull())
     return false;
 
   WbNode *parent = parentNode();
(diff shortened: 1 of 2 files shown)


src/webots/nodes/WbSolidReference.cpp
```diff
@@ -60,6 +60,8 @@ void WbSolidReference::updateName() {
   const bool linkToStaticEnvironment = name == STATIC_ENVIRONMENT;
   if (!linkToStaticEnvironment)
     mSolid = QPointer<WbSolid>(ts->findSolid(name, upperSolid()));
+  else
+    mSolid = NULL;
```
omichel, line 64: This will likely cause a crash on the second line. Rather I would write something like:
```suggestion
    mSolid.clear();
```
However, I am not sure the usage of `QPointer` in this class. I don't see why we need a `QPointer`, maybe we could try to revert to a normal pointer instead...

ShuffleWire: It do not crash thanks to the !linkToStaticEnvironment in the condition evaluated before will avoid evaluation of mSolid.isNull(). The NULL value is then nicely handle by the rest of webots, as far as my testing goes.
But I agree, clean() could be nicer, (i don't know anything about QPointer...).
May I lets you make the decision to keep it ?

omichel: I studied it and I found that the `QPointer` is needed here (because the corresponding object may be deleted meanwhile and we don't want to have a dangling point here), so we should keep it. Also `mSolid = NULL;` is fully equivalent to `mSolid.clear();`, however I would prefer the second syntax as it shows that `mSolid` is not a standard pointer, but a `QPointer`.

omichel (approved):
Thank you.

PR #4070, opened 2021-12-22
Fix physic reset bug

### LINKED ISSUE #2899: SoccerBall is moving after setting its position and resetting velocity

BenjaminDeleze:
**Description**
Fix #2899 by renabling the bodies just before the contactPoint calculation


BenjaminDeleze:
Current solution works only for single threaded worlds.

It also works for multi threaded worlds with more than two objects.

But when we have only the ball and the floor, the behaviour is the same as before.
Investigation so far:

- `odeNearCallback` is not even called in the case of the bug, this explains that the fix does not work.
- It is not called because an error is thrown before the call to `odeNearCallback` that make ode skip the collision detection part.
  - The error seems due to a bad calculation of a variable in `getVirtualAddress` that make this test fail:https://github.com/cyberbotics/webots/blob/5389e9d01de62f65ef7ed30f4759c375882663fb/src/ode/ode/src/ode_MT/clustering/cluster.cpp#L53

BenjaminDeleze:
I may have isolated the problem:

Because the following test fails, the collision computation is not done: https://github.com/cyberbotics/webots/blob/d9dada3ab5f47ebcee01f7cb269080465cc35dbe/src/ode/ode/src/ode_MT/ode_MT.cpp#L115

From there, there is two ways:

- Either we removed this test. It seems to solve the problem. The test suite succeeds and I did not remark any side effects. However, it is not elegant and I am afraid of unseen side effects.
- Or we try to solve the root of the problem.

**Why does the test fails?**

The test fails because this previous test fails: https://github.com/cyberbotics/webots/blob/5389e9d01de62f65ef7ed30f4759c375882663fb/src/ode/ode/src/ode_MT/clustering/cluster.cpp#L53

and it fails because the `hsAxis` variable is not up to date.
The current flow goes as follow:

- Step x:
  - Translate the ball
  - Do the physic step:
    - Check if `hsAxis` needs some update. It is not flagged as need update, so do nothing.
    - Try to compute the collisions.
      - As the `hsAxis` variable does not correspond anymore to the current position of the ball:
        - The collision detection is aborted.
        - The `hsAxis` is flagged as dirty.
- Step x + 1:
  - Do the physic step:
    - Check if `hsAxis` needs some update. It is flagged as so, so we update it.
    - Try and succeed to compute the collisions.

An interesting thing is that, if instead of translating the ball, I change the radius of it, the flow goes as follow:
- Step x:
  - Change the radius of the ball
  - Flag `hsAxis` as dirty.
  - Do the physic step:
    - Check if `hsAxis` needs some update and update it.
    - Try and succeed to compute the collisions.

The function responsible for flagging the `hsAxis` before the physic step is `applyToOdeData`.
When changing the radius, this function is called in `WbSphere.cpp`.
This function is also in `WbTransform.cpp` but it is not called in our case. Because this function is only called if the __transform__ or the __sphere__ that is moved/changed is a bounding object.
However, when changing the radius of the ball, the radius of the bounding sphere is also changed so it works, but when translating the object, only the translation of the ball *solid* is changed, not the one of the bounding object, so it is not called.

To make it work, we would need to transmit the information that the parent object to the bounding object such that it could call `applyToOdeData`.

BenjaminDeleze:
I found a working solution, but I think that this one is maybe suboptimal. I will continue to search a better one.

BenjaminDeleze:
I think I have a pretty good solution for the case with the supervisor.
But I could reproduce the bug in other situations, so let's make a point:

- Supervisor change in translation/rotation + `node.resetPhysics`: **Fixed** both with single and multi threading. (note that `node.resetPhysic` is needed for the objects to appear at the right position, otherwise it appears at the position defined in the supervisor + previous forces + gravity)
- Supervisor `saveState` then `loadState`: **bugs** even if the node did not move between the save and the load.
- Supervisor `saveState` then `loadState` + `node.resetPhysics`: **works with the fix**, but we do not want to use `node.resetPhysics` with `loadState`
- Move/Insert node in SceneTree: **bugs**
- Move node in GUI: **bugs**
- Reset: **bugs** (even if nothing moved)
- Reload: **works**

Regarding the reset, it is disturbing because you can have a bug by doing something like thing:

- Load a world with a ball on a floor
- Step: nothing move
- Reset
- Step: the ball sinks on the ground.

I thought that `handleInitialCollisions` (called at the loading of the world) was the solution, but it is also called in the reset and does not work.

BenjaminDeleze:
As this PR fixes the original bug, I propose to merge it and open an issues about the remaining bugs

src/ode/ode/src/ode_MT/ode_MT.cpp
```diff
@@ -572,3 +572,11 @@ void dGeomPlaneSetParams_MT (dGeomID g, dReal a, dReal b, dReal c, dReal d, dGeo
         clusterManager->currentCWAS->propagatePlaneParams(g, a, b, c, d, _geomPlaneSetParamsFunc);
     return _geomPlaneSetParamsFunc(g, a, b, c, d);
 }
+
+void dGeomOffset_MT (dxGeom *g) {
+  g->recomputeAABB();
+
+  if (clusterManager->currentCWAS) {
+      clusterManager->currentCWAS->propagateOffsetChange(g);
+  }
```
omichel, line 581:
```suggestion
  if (clusterManager->currentCWAS)
    clusterManager->currentCWAS->propagateOffsetChange(g);
```

omichel (changes requested):
We should add a line in the change log and open a new issue to list the remaining bugs to address.

omichel (approved):
Thank you.

PR #4073, opened 2021-12-22
Fixed small memory leaks

omichel:
[QNetworkReply](https://doc.qt.io/qt-5/qnetworkreply.html) objects need to be released using the `deleteLater()` method.


ad-daniel (approved):
Thank you!

Issue #4034, opened 2021-12-15
Scale parameter should be removed from Robot node


ad-daniel:
**Describe the Bug**

The `Robot` node inherits the scale parameter from the `Transform` node, using this parameter however has several issues/bugs:
- If a user sets the density in the `Physics`, it will scale but if it sets the mass, it doesn't.
- If the robot has sensors, like a distance sensor. After scaling it won't work well anymore. (detection point is incorrect, lookup table isn't respected, ...)
- Although it might be possible to "fix" all sensors to work with a smaller uniform scale, for non-uniform ones such as  `scale 0.5 1.5 0.3`, it would be incredibly hard to have the desired behavior for every sensor type

omichel:
I believe the `scale` parameter should be allowed only in the `Transform` node and in no other node, as it will cause various problems. Even in a simple `Solid` node, there is a problem if the `mass` is specified in the `Physics` node of the `Solid` as its value won't scale or won't be correct. Also, for the same reason, `Transform` nodes shouldn't contain any `Solid` node (or descendant). When needed, users should use an upper `Solid` node (instead of a Transform) to encapsulate other `Solid` nodes.

omichel:
An idea to make things clean would be to introduce a `Pose` node that would be basically like a `Transform`, but without the `scale` field. The `Solid` nodes would then inherit from `Pose` and a `Pose` node would be allowed to include solid children.

PR #4079, opened 2021-12-23
Add `Pose` node (like `Transform` node but without `scale`)

ad-daniel:
**Description**
Scaling of robots is problematic due to the presence of sensors and physics. The `Pose` node, from which `Transform`, `Solids` and `Robots` will inherit, is much like a `Transform` node but without the scale parameter. This latter will be exposed by the "new" `Transform` node, in order to maintain backwards compatibility.

**Related Issues**
Fixes https://github.com/cyberbotics/webots/issues/4034

**Tasks**
  - [x] add `Pose` node
    - [x] rename `Transform` to `Pose`
    - [x] rename `AbstractTransform` to `AbstractPose`
    - [x] rename node constants
    - [x] rename utilities like `findUpperTransform`, `findUppermostTransform`, ...
    - [x] rename variables, like `WbPose *transform` to `WbPose *pose`
  - [x] create `WbTransform` class that inherits from `WbPose`
  - [x] make `WbMatter` inherit from `WbPose`
  - [x] make `WbSkin` inherit from `WbTransform` instead of `WbAbstractTransform`
  - [x] remove `scale` from `Pose` (and all references to it in `Solid`, `Robot`, ... )
  - [ ] check `WbTrack` functionality
  - [ ] only allow `Shape` children in `Transform` nodes
  - [ ] `boundingObjects` should allow only Pose
  - [ ] x3d exporting of transform/pose
  - [ ] update docs
  - [ ] update node chart
  - [ ] script to automatically detect/change `Transform { Solid {}}` structures into `Pose { Solid {}}` (should be rare)


Justin-Fisher:
If it were possible, my inclination would be to instead move in the direction of letting `scale` work meaningfully with more things, rather than reducing the number of things that have a `scale` field.  E.g., it'd be nice to be able to make oblate spheroid bounding objects.  Or is it prohibitively difficult to get the physics to work right with those?

I did run into a lot of perplexment once when I had scaled a `Camera` node to make its visual appear the way I wanted, and discovered that this `scale` also squashed the camera's image!  That surprised me and took awhile to debug, but in the end I decided that was probably working-as-designed.  But this "fix" probably would have avoided my perplexment on that.

stefaniapedrazzi:
> make WbSkin inherit from WbTransform instead of WbAbstractTransform

Note that the `Skin`node should not have the `children` field, so if `Transform` does it then `Skin` cannot inherit from `Transform` and we still need an intermediate node like `WbAbstractTransform`.

ad-daniel:
mmh, that might be a bit of a problem. I assume the skin node does require the scale, correct?

stefaniapedrazzi:
> mmh, that might be a bit of a problem. I assume the skin node does require the scale, correct?

Yes.
The `WbAbstractTrasform` class has been introduced to avoid duplicating the translation/rotation/scale-related code in `WbTransform` and `WbSkin`.
Changing `WbSkin` so that inherits from the new `WbAbstractPose` would at least avoid duplication of code related to translation and rotation. It should be evaluated the size of the scale-related code.

ad-daniel:
> Changing WbSkin so that inherits from the new WbAbstractPose would at least avoid duplication of code related to translation and rotation. It should be evaluated the size of the scale-related code.

I see, I'm not entirely sure WbAbstractPose would even need to exist after the rework to be honest. Since the scale-related methods would be moved down to WbTransform, the only difference that remains between WbPose & WbAbstractPose is the children field. After discussing with Olivier, it might be cleaner to just have a Pose node that inherits from Group and consider the Skin its own thing. It would entail some duplication that's true, but frankly given that the `scale` now drop down at the Transform level, I'm not convinced the benefit of code-reutilization justifies adding an intermediate Frankenstein just for the Skin in order to have both a `scale` but not the `children` field, making the two coexist might turn out worse than duplicating translation/rotation. There's a lot of moving pieces though, so I'll investigate further

As it stands, it would look something like this, but like I mentioned there might be roadblocks:

```
WbGroup (children)
   |
   |
   |__ WbPose (add translation + rotation)
         |
         |__ WbSolid (add solid stuff) __ WbRobot
         |
         |__ WbTransform (add scale)

WbSkin
```

ad-daniel:
How do you suggest Skin to have scale but not children in this context (i.e. with the addition of the Pose node without scale)?

stefaniapedrazzi:
I would keep at least the WbAbstractPose class including all the translation/rotation code and use it as base class for the WbPose and WbSkin classes.

ad-daniel:
I'm fine with that, but that would still leave duplication for the scale, as I don't see another way of adding scale support without also getting children in the mix

ad-daniel:
continued here: https://github.com/cyberbotics/webots/pull/5978

PR #4088, opened 2022-01-04
Document rescaling functionality

ad-daniel:
**Description**
As far as I can tell, the re-scaling functionality is not currently documented. I've kept the wording generic enough (no mention of solids specifically) so that it will remain valid after https://github.com/cyberbotics/webots/pull/4079 is merged.

I've also taken the opportunity to update the translate/rotate manipulators so that the image is consistent with what it looks like in 2022a.

preview: https://cyberbotics.com/doc/guide/the-3d-window?version=enhancement-document-rescale


omichel (approved):
Great, thank you.

PR #4096, opened 2022-01-06
Allow the use of WebotsView in Safari

BenjaminDeleze:
Since version 15, Safari support webgl2. 
I tested if with https://www.cyberbotics.com/animations/soccer_example/soccer.html. It works more or less, there is sometimes some troubles to load the page, but once it is loaded it works well.
And with https://cyberbotics1.epfl.ch/open-roberta/ and I experimented no issue with this one.


omichel (approved):
🎉

Issue #4097, opened 2022-01-06
Rotating the object via GUI resets the rotation to 0 suddenly


Adman:
**Describe the Bug**
When I try to set the rotation values to `[0, 1, 0, 3.27249]` and then decrease the angle by clicking "down" arrow in GUI, the angle is reset to 0.

**Steps to Reproduce**
1. Open Webots with some world (you don't have to run the simulation).
2. Click on the object.
3. Go to `rotation` property in GUI.
4. Set `x=0 y=1, z=0, angle=3.27249`. Rotation type is `Axis-Angle`.
5. Click "down" arrow.

**Expected behavior**
The angle should decrease and robot rotate by a small margin.

**System**
 - Operating System: Linux Ubuntu 20.04
 - Tested with Webots R2022a and R2021b

PR #4098, opened 2022-01-07
Fix rotation spinner snap value

ad-daniel:
**Description**
Fixes https://github.com/cyberbotics/webots/issues/4097.

For values around ~3.27, the snapping mechanism associated to the rotation spinner drops to zero (because near these values the `intpart` is 0), hence while rotating using the spinner arrows occasionally it can snap to 0 instead of decreasing/increasing at a constant rate (decreasing from 6.4 it snaps to PI for instance).
The proposed solution maintains the snapping behavior (at multiples of PI) but also the constant increase/decrease.

PS: axis-angle fields are restricted to [-PI, PI], so in practice putting values above PI will wrap them around anyway but having it jump when using the spinner arrows doesn't make sense either way.


src/webots/scene_tree/WbFieldDoubleSpinBox.cpp
```diff
@@ -66,12 +66,12 @@ void WbFieldDoubleSpinBox::stepBy(int steps) {
       value += pow(10, -mDecimals) * steps;
       break;
     case RADIANS: {
-      // angle in radians: change by pi/24 and snap to pi/24 if close by 1%
+      // angle in radians: change by pi/24 and snap to the nearest multiple of PI
       value += M_PI * steps / 24.0;
       double intpart;
       double decpart = modf(fabs(value) / M_PI, &intpart);
```
BenjaminDeleze, line 72: Would it be possible to get rid of intpart, as it seems not used anymore?

BenjaminDeleze (commented):
Seems good, just a question

src/webots/scene_tree/WbFieldDoubleSpinBox.cpp
ad-daniel: indeed we can!

BenjaminDeleze (approved):
Thank you!

PR #4099, opened 2022-01-07
Remove `static` mention from docs

ad-daniel:
**Description**
Static is the new default for any PROTO, the tag has been removed a while back.


BenjaminDeleze (approved):
Thank you

Issue #3873, opened 2021-11-09
Improve `village_winter` crossroad textures


ad-daniel:
**Describe the Bug**

All the crossroads in `village_winter` have a different texture compared to the roads, which looks kinda odd. Would be nice to have something more consistent.

![aa](https://user-images.githubusercontent.com/44834743/140965492-a41f6b38-4ef7-4ae8-a09a-c6fb20e42b6c.png)

BenjaminDeleze:
It would be quite easy to replace them with `roundabout` protos.
The problem is that the sumo exporter does not have support for `roundabout` right now.  So we would not be able to regenerate the Sumo network.

Thomas-Feldmeier:
in village_winter.wbt, change Crossroad appearence to same as in Road
i changed the first Crossroad, it looks nice
![village_winter](https://user-images.githubusercontent.com/73528085/141085632-3ea29ed5-56c9-48b0-a7f2-06d0f5e8d021.png)
ad

ad-daniel:
Whatever change ends up being picked, it should be done in `feature-enu-plan` branch to avoid conflicts when merging

PR #4101, opened 2022-01-07
fix village winter texture

joachimhgg:
**Description**
Fix crossroad appearance in Village winter.

**Related Issues**
This pull-request fixes issue #3873

**Tasks**
  - [x] Change the PBRappareance of the crossroads by the Road_with_snow appearance
  - [x] Change textures files to have an "omnidirectional" road appearance. 
image

  
image
the result is not optimal for the roads, but it seems to be the more realistic PBR texture we could find:
image


BenjaminDeleze (commented):
Not, perfect, but way better indeed!
@omichel do you think an entry in the changelog is needed?

omichel:
> @omichel do you think an entry in the changelog is needed?

No, it's a detail.

BenjaminDeleze (approved):

PR #4104, opened 2022-01-10
Feature PROTO wizard

ad-daniel:
**Description**
Adds a Wizard for the creation of PROTO files.

https://user-images.githubusercontent.com/44834743/148840618-97dfe4ec-7db4-4f17-8d9c-d1644c128402.mp4

PS: it was necessary to change a test in the test suite because the function [WbVariant::toStringRepresentation](https://github.com/cyberbotics/webots/blob/85016b29383660d415ceffd7bd4c644357827c12/src/webots/vrml/WbVariant.cpp#L145-L173) did not export `SFNode` in a VRML compliant manner (i.e missing brackets). Although "toStringRepresentation" doesn't mention VRML per-se, given the rest of the function and where this is located, it seems clear to me this was the intent.
The missing of the brackets was a problem for the wizard since when attempting to print the default value of the exposed parameters of the PROTO for fields of type `SFNode`, the lack of brackets made it a non-valid PROTO.


src/webots/gui/WbNewProtoWizard.cpp
```diff
@@ -0,0 +1,412 @@
...
+    connect(selectAll, SIGNAL(stateChanged(int)), this, SLOT(updateCheckBox(int)));
+
+    foreach (const QString &name, fieldNames) {
+      mExposedFieldCheckBoxes.push_back(new QCheckBox(name));
+      layout->addWidget(mExposedFieldCheckBoxes.back());
+    }
+  } else {
+    QLabel *message = new QLabel("No fields.");
+    layout->addWidget(message);
+  }
```
omichel, line 380:
```suggestion
  } else
    layout->addWidget(new QLabel("No fields."));
```

omichel (changes requested):
Thank you.
This is very nice and useful.
The explanation are pretty clear.
I have just a few suggestions.

ad-daniel:
Added description field:
description

omichel:
The second page doesn't look great on my Windows machine:

image

omichel:
Also, I believe it could be a good idea to pre-select a number of field that should be exposed most of the time, as recommended in the [User Guide](https://cyberbotics.com/doc/reference/proto-design-guidelines#recommended-fields-for-simple-proto-nodes). So, if they exist, the "translation" and "rotation" field should be checked by default. Maybe also the "name" field and more arguably the "controller" field (not sure). What do you think?

ad-daniel:
Hopefully it is high enough now

ad-daniel:
Text should no longer be cut off.

omichel (approved):
I confirm it now displays nicely the fields on Windows. They are not cut any more.
Thank you.

PR #4105, opened 2022-01-10
Remove webots -> libcontroller notification of relayed commands in motor couplings

ad-daniel:
**Description**
Currently, in the context of coupled motors when Webots receives an API request (for instance setting of the target position), it relays this message to the other motors within the same coupling which results in each of them notifying libcontroller about their own updated state. This is necessary to ensure data symmetry between Webots and libcontroller, since for example when requesting data, the locally available one is usually provided (without asking Webots for it, hence it needs to be up to date). This is especially a problem with coupled motors since (currently) a single API call may affect multiple motors, not just the one associated with the tag from which the request stemmed.

This continuous exchange of information is however unnecessarily heavy and currently blocks the possibility of exposing the `robot_step_begin` and `robot_step_end` APIs. The proposed solution is to have libcontroller itself relay the information to other motors within its coupling.

### DIFF
--- a/src/controller/c/motor.c
+++ b/src/controller/c/motor.c
@@ -294,7 +294,6 @@ void wb_motor_set_velocity(WbDeviceTag tag, double velocity) {
       else
         fprintf(stderr, "Error: %s(): invalid sibling in coupling.\n", __FUNCTION__);
     }
-
   } else
     fprintf(stderr, "Error: %s(): invalid device tag.\n", __FUNCTION__);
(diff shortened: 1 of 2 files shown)


ad-daniel:
@ygoumaz I think this is done. I don't think the other demands of re-configuration done by Webots can be removed as they are triggered by a change of parameters from the UI. Let me know if something else is blocking/missing.

omichel (approved):

PR #4106, opened 2022-01-11
Remove `static` mention

ad-daniel:
**Description**
`static` is the default state for any new proto.


BenjaminDeleze (approved):
Thank you

PR #4117, opened 2022-01-14
Enhancement web viewer

BenjaminDeleze:
**Description**
Several improvements to WebotsView, you can test it on https://beta.webots.cloud

- [x] Update the documentation.

https://cyberbotics.com/doc/guide/web-simulation?version=enhancement-web-viewer
https://cyberbotics.com/doc/guide/web-animation?version=enhancement-web-viewer

*API*
- [x] Add the API function `getNode` that would return a webotsjs node. It is for example needed in order to retrieve the inital value of some fields to be able to reset them correctly without reloading the whole world.
- [x] Add the API function `updateNode` that allows the user to update a webotsjs field of a node, supported field so far are `rotation`, `translation` and `color` (base, diffuse, emissive). This function internally uses the `applyPose` function that is used to update the webotsJS based on update from webots desktop. It is easy to extend.
- [x] Add API functions to have a hook on both the stdout and stderr message sent by webots during simulation.
- [x] Expose the function `resize` that automatically resize the web-component.
- [x] Create the `hasScene` function to know if/when a scene is loaded.

*Fix*
- [x] Change default rotation in `Parser.js` from `0 1 0 0` to `0 0 1 0`.
- [x] Reduce the size of the setting button of animation that causing troubles in iframe.
- [x] Fixed the animations css.
- [x] Rename `revert` in `reload`

*Features*

- [x] Add an information panel about Webots (button "i") in the toolbar.
- [x] Add a toString method to WbVector3.js and WbVector4.js.
- [x] Harmonize toolbar across simulation/webots/scene
  - [x] Compile `animation.css`, `wwi.css` and `scene.css` in one single css file.
  - [x] Compile the creation of buttons/sliders/timers in one file.
  - [x] Reimplement the functionalites of the scene.
  - [x] Reimplement the functionalites of the animation .
  - [x] Reimplement the functionalites of the streaming viewer.
  - [x] Handle removal/hidding of the toolbar on view close
  - [x] Hide the toolbar when the mouse is off the panel or inactive.
  - [x] Move the possibility of disabling buttons by adding attributes to webots from `webots.js` to `WebotsView.js`.
  - [x] Disable the buttons/hide the toolbar when switching between worlds in streaming mode.
  - [x] Refresh the text in the information panel when we switch between worlds.
  - [x] Restore the broadcast mode.
  - [x] Restore mjpeg mode.
  - [x] Cleanup.


omichel (changes requested):
Maybe we could remove the Roboto fonts from webotsjs as they don't seem to be used by webotsjs.

omichel:
Just a small layout problem: the right margin of the information window seems too small (smaller than the left margin):
image

BenjaminDeleze:
It is now fixed

omichel (approved):
Perfect, thank you.

PR #4118, opened 2022-01-14
Fixed broken dependencies in libCppController

omichel:
The management of the dependencies was broken in the libCppController:
```bash
$ cd src/controller/cpp
$ make release
[...]
$ touch ../../../include/controller/cpp/webots/Robot.hpp
$ make release
make: Nothing to be done for 'release'.
```

This was due to the fact that the `build/release/Robot.d` dependency file was starting with:
```
Robot.o: Robot.cpp \
```
whereas it should start with:
```
build/release/Robot.o: Robot.cpp \
```
Hence the dependencies for `Robot.o` where silently ignored.

This PR fixes this bug.


ygoumaz (approved):
All right to me. I also checked the impact on the Python library and more particularly the test suite freezing issue when recompiling. It is working perfectly fine now.

PR #4137, opened 2022-01-20
Remove-webkit-webengine

joachimhgg:
**Description**

Remove `Webkit`, `WebEngine` and `WebChannel` dependencies. 

**Tasks**

keywords:
- [x] `WebEngine`
- [x] `Webkit`
- [x] `WbRobotWindowTransportLayer`
- [x] `qtwebchannel`
- [x] `QWebPage`
- [x] `WBWebPage`
- [x] `QT_WEB_INCLUDE`
- [x] `QWebSettings`
- [x] `local webots.js`
- [x] `ChannelTransport`

**Test on Windows**
- virtualHeadset

Test on Mac successful.


dependencies/Makefile.mac
```diff
@@ -33,21 +33,18 @@ cleanse: $(PACKAGES_CLEAN)
...
+	@rm -rf $(WEBOTS_DEPENDENCY_PATH)/$(QT_PACKAGE) $(WEBOTS_HOME)/Contents/Frameworks/Qt* $(WEBOTS_HOME_LIB)/qt $(WEBOTS_HOME)/include/qt $(WEBOTS_HOME)/bin/qt/lrelease $(WEBOTS_HOME)/bin/qt/lupdate $(WEBOTS_HOME)/bin/qt/moc
 
-qt: $(WEBOTS_HOME)/resources/web/local/qwebchannel.js
-
-$(WEBOTS_HOME)/resources/web/local/qwebchannel.js: $(WEBOTS_DEPENDENCY_PATH)/$(QT_PACKAGE)
-	@echo "# uncompressing $(QT_PACKAGE)"
-	@tar xfm $(QT_PACKAGE) -C $(WEBOTS_HOME)
+qt: $(WEBOTS_DEPENDENCY_PATH)/$(QT_PACKAGE)
 
 $(WEBOTS_DEPENDENCY_PATH)/$(QT_PACKAGE):
```
omichel, line 40: Simpler this way:
```suggestion
qt: $(WEBOTS_DEPENDENCY_PATH)/$(QT_PACKAGE)
```

scripts/packaging/update_all_worlds.py
```diff
@@ -43,4 +43,4 @@
 
 for i in range(len(worlds)):
     print('%d/%d: %s' % (i + 1, len(worlds), worlds[i]))
-    call([webotsFullPath, worlds[i], '--minimize', '--batch', '--no-sandbox', '--mode=pause', '--update-world'])
+    call([webotsFullPath, worlds[i], '--minimize', '--batch', '--mode=pause', '--update-world'])
```
ad-daniel, line 46: Was the `--no-sandbox` command removed or didn't exist to begin with? Either way there's still a reference to it in webots_ros [here](https://github.com/cyberbotics/webots_ros/blob/d0e2131164b5b18940a4548d93d0fc414d26020c/scripts/webots_launcher.py#L46)

ad-daniel (changes requested):
Nothing catches the eye, only a remaining reference. The branch compiles fine on ubuntu 20 and the robot windows open up correctly.

Running the test suite locally this test also failed:

`FAILURE with speaker: wb_speaker_is_speaking should return true when TTS is running.`

tried twice and happened both times, I don't know if it's related.

resources/web/local/qwebchannel.js
```diff
@@ -0,0 +1,448 @@
+/****************************************************************************
```
omichel, line 1: Shouldn't we remove this file? It seems useless now, isn't it?

scripts/packaging/update_all_worlds.py
joachimhgg: it has been removed since it was a command related to the WebEngine debugging. So it should also be removed in webot_ros. I am creating a PR.

resources/web/local/qwebchannel.js
joachimhgg: > Nothing catches the eye, only a remaining reference. The branch compiles fine on ubuntu 20 and the robot windows open up correctly.
> 
> Running the test suite locally this test also failed:
> 
> `FAILURE with speaker: wb_speaker_is_speaking should return true when TTS is running.`
> 
> tried twice and happened both times, I don't know if it's related.

It is related to openAL, #4047

joachimhgg:
> Nothing catches the eye, only a remaining reference. The branch compiles fine on ubuntu 20 and the robot windows open up correctly.
> 
> Running the test suite locally this test also failed:
> 
> `FAILURE with speaker: wb_speaker_is_speaking should return true when TTS is running.`
> 
> tried twice and happened both times, I don't know if it's related.


It is related to openAL, #4047.

resources/web/local/qwebchannel.js
joachimhgg: It is but the file is still include in the qt-5-12 folder package on the server. So each time we make Webots this file will be add to the `resources/web/local `. Because we will move to qt6, I suggest that we keep the file the time to move to qt6 because whatever it will be add when downloading the qt5 dependency.

omichel (approved):
It looks good to me now.

ad-daniel (approved):
Thank you, looks fine to me as well. 
I'll open an issue about the failing test, I just checked on develop and it doesn't work there either.

omichel (approved):

Issue #4127, opened 2022-01-18
Get an empty array of object with "recognition_objects" topic


BenjaminHug8:
Sorry. I always get an empty array of object when using the topic "recognition_objects".  I'm wondering what could be the reason.
![6ce60df0c48bda3fb262108d1f4c2a3](https://user-images.githubusercontent.com/76592592/149964981-4a857fd3-cc25-4a3e-9713-b4ac0f86e7bc.png)
And this is my code:
`void camera_topCallback(const webots_ros::RecognitionObjects::ConstPtr &msg)
{
    camera_topRawvalue = *msg;
    ROS_INFO("Update camera_topRawvalue: numberofobj %d", msg->objects.size());
    //ROS_INFO("FIRST OBJ: %s", msg->objects[0].model.c_str());
}`

_Originally posted by @Liuxingyuming in https://github.com/cyberbotics/webots/issues/3234#issuecomment-1015515193_

BenjaminHug8:
@Liuxingyuming I propose to continue to discuss about your issue here instead of inside a close pull request.

Liuxingyuming:
Okk, thank u!

Liuxingyuming:
I found that, when I echo the topic, the seq of the message is always 0. It seems that the message is not getting updated ？

BenjaminHug8:
Would it be possible to upload your package here so I can made some test on it?

Liuxingyuming:
I upload my project on github:

https://github.com/Liuxingyuming/Supermarket-Robot-Based-on-webots_ros

The ros code is in /src/webots_demo/src. Thanks.

BenjaminHug8:
I got the same issue locally, it seems that the message is not properly built.

epcm:
> I got the same issue locally, it seems that the message is not properly built.

So is there any temporary solution for this issue ?

BenjaminHug8:
I am working on it, once it will be ready you should be able to build Webots from sources or wait for a nightly build.

Liuxingyuming:
Thanks!

BenjaminHug8:
The fix has been merged in `master` branch of Webots.

You have now the choice between building Webots from sources or waiting for the next [nightly build](https://github.com/cyberbotics/webots/releases) in order to benefit of this fix.

I will close this issue as it should be solved now, but feel free to reopen it in case there is still something not working :)

PR #4139, opened 2022-01-20
md5sum exception only in headersGenerator

BenjaminHug8:
**Description**
The ROS `recognition objects` was returning an empty list of objects as it was still using old header of ROS `recognition objects` message.

These header files are needed as Webots create header files for messages and services on its own without `catkin_make`, thus md5sum can be wrong when using some messages from other packages.

I propose another solution to only copy past the md5sum of the ROS messages instead of the whole header file.

**Related Issues**
This pull-request fixes issue #4127 and is related to [this other issue](https://github.com/cyberbotics/webots_ros/issues/64).


BenjaminHug8:
@omichel should I launch other tests than `test ros` in the labels?

omichel:
> @omichel should I launch other tests than `test ros` in the labels?

I don't think so.

omichel (approved):
Thank you.

PR #4140, opened 2022-01-20
Fix world save API

ad-daniel:
**Description**
The function `supervisor_world_save` isn't behaving as documented/intended.
If no argument is provided (or if NULL is provided in case of C API), the world should be saved.
If an argument is provided but isn't valid (no `.wbt` extension ), a warning should be printed.
If a valid argument (filename + extension, or path + filename + extension), the world should be saved as such.
The PR restores this functionality.

~~PS: I don't know if the change of argument name is warranted. C api uses "filename" and all others "file". Technically the provided string can be a path, so "file" might make more sense but it's debatable. Either way should be consistent.~~


omichel (approved):
Thank you.

PR #4141, opened 2022-01-20
feature convert robot windows

joachimhgg:
**Description**
Convert robot windows of Webots robots to meet the new web browser robot window.

**Tasks**

- [x] vehicles
- [x] c3d
- [x] e-puck
- [x] thymio
- [x] custom_robot_window

robotBenchmark:
- [x] highway_driving
- [x] humanoid_sprint
- [x] inverted_pendulum
- [x] maze_runner
- [x] obstacle_advoidance
- [x] pick_and_place
- [x] pit_escape
- [x] robot_programming
- [x] sqare_path
- [x] visual_tracking
- [x] wall_following


ad-daniel (changes requested):
The robot window of the thymio2 is a bit too high up, with the labels ending up in what I assume is the header, can't it be centered vertically as well?
a
Similarly, the robotbenchmark ones don't appear to fill horizontally (on firefox), but isn't that big of a deal
b

On the other hand Is it necessary to have a tabs.css for each of them? At first glance they appear to be identical

joachimhgg:
> The robot window of the thymio2 is a bit too high up, with the labels ending up in what I assume is the header, can't it be centered vertically as well? a Similarly, the robotbenchmark ones don't appear to fill horizontally (on firefox), but isn't that big of a deal b
> 
> On the other hand Is it necessary to have a tabs.css for each of them? At first glance they appear to be identical

For Thymio, I removed the grey band. I added the `tabs.css` using `jsdelivr` instead of adding the tabs to all the `robot_windows` folders. For the robotbenchmark's robot windows, it was already the case before, it is because the parameter `width` is set to `90%` hence I imagine we can keep it like that.

PR #4145, opened 2022-01-24
Update Urls

BenjaminDeleze:
**Description**

- Update urls that were not updated by the script.
- Add a section in https://github.com/cyberbotics/webots/wiki/Release-Procedure to update urls that can only be updated once the tag is pushed.


omichel (approved):
Thank you.

PR #4147, opened 2022-01-24
Generalize update of Webots version in world file header

LucasWaelti:
**Description**
The [`convert_nue_to_enu_rub_to_flu.py`](https://github.com/cyberbotics/webots/blob/42fc2753ccda54b7747cf6e1c2d97820f609f7e6/scripts/converter/convert_nue_to_enu_rub_to_flu.py#L224) script only allows to update the version at the first line of the world or proto file if the version being updated is `R2021b`. 

For instance, the line `#VRML_SIM R2018b utf8` is not updated by the script and therefore requires to be modified by hand to `#VRML_SIM R2022a utf8`. 

This pull request should allow the script to update the Webots version in any case.


joachimhgg (approved):
Looks good to me, thanks for your contribution!

PR #4148, opened 2022-01-25
Fix node string display

ad-daniel:
**Description**
In https://github.com/cyberbotics/webots/pull/4104, in order to pre-fill the PROTO parameters the function `toStringRepresentation` was modified such that for SFNodes the `{}` is also included in the representation (just the name alone is not VRML compliant for a node, the brackets are needed). This function however didn't aim to be VRML compliant and is used for other things, most notably to display the names of nodes in the scene tree, which meant this change added curly brackets in the GUI as well.

Although the addition of curly brackets could have been done at the level of other functions (like `WbMultipleValue::toString`, `itemToString`), these too serve other purposes and changing it would have side-effects.

All things considered, the best option appears to be to add the brackets locally where it's needed.

### DIFF
--- a/src/webots/gui/WbNewProtoWizard.cpp
+++ b/src/webots/gui/WbNewProtoWizard.cpp
@@ -19,6 +19,7 @@
 #include "WbFileUtil.hpp"
 #include "WbLineEdit.hpp"
 #include "WbMessageBox.hpp"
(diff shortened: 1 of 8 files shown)


omichel (approved):
Cleaner this way indeed.

PR #4149, opened 2022-01-25
Fix mobile navigation

### LINKED ISSUE #4146: Web interface: Problems with touch devices
When using the webots simulation e. g. [here](https://cyberbotics1.epfl.ch/open-roberta/), touch events are not evaluated as expected.
- Moving the scene with one finger leads to disappearing of the whole scene
- Resizing with two fingers does not work. 

The only way to get the scene back is reconnecting.

**System**
 - Operating System: Android 7 (higher versions do not work either)
 - Browser: Chrome

BenjaminDeleze:
**Description**
Fix #4146 

I restored it close to what it is robotbenchmark, you can test it there: https://www.cyberbotics.com/animations/soccer_example/soccer.html

I advise you to switch in fullscreen mode to test on mobile.


omichel (approved):
It works pretty well now, even on my old phone...

PR #4155, opened 2022-01-26
Fix broken video link

omichel:
The NAO video was missing...
New version: https://cyberbotics.com/doc/guide/nao?version=fix-missing-video


ad-daniel (approved):
Thank you, wouldn't be a bad idea to improve the quality of these videos at some point

PR #4156, opened 2022-01-26
Advertise simulation servers

omichel:
This PR changes the simulation server so that it will notify a server (for example webots.cloud) about its availability to host public simulations.
The monitor page of the simulation was also improved.
Additionally, the `webots.js` API is improved: the URL passed to the connect method is now `https` instead of `wss` which is more straightforward.
The result can be seen at https://beta.webots.cloud/server


BenjaminDeleze (approved):
Seems all good for me

PR #4162, opened 2022-01-28
Fix documentation for lidar and rangeFinder about minRange and maxRange

BenjaminHug8:
**Description**
The `Field Summary` is not correct about the `maxRange` and `minRange`, saying that values out of bound will return INF instead of `maxRange` and `minRange`.


ad-daniel:
documentation changes should target the release branch

BenjaminHug8:
Closed in favor of another PR based on the correct branch.

PR #4163, opened 2022-01-28
fix Documentation max and min range for lidar and rangefinder

BenjaminHug8:
**Description**
The `Field Summary` is not correct about the `maxRange` and `minRange`, saying that values out of bound will return INF instead of `maxRange` and `minRange`.

### DIFF
--- a/docs/reference/lidar.md
+++ b/docs/reference/lidar.md
@@ -159,11 +159,11 @@ A typically good value for this field is to set it just big enough so that the s
 More information about the frustum is provided in the [frustum](camera.md#frustum) section of the [Camera](camera.md) node.
 
 - The `minRange` field defines the minimum range of the lidar, objects closer to the lidar than the minimum range are not detected (but still occlude other objects).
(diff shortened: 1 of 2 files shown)


ad-daniel:
I'm not convinced this is correct, according to the [shader](https://github.com/cyberbotics/webots/blob/master/resources/wren/shaders/encode_depth.frag) we actually return infinity if the depth is bigger than `maxRange` or if smaller than `minRange`.
I believe the change was necessary because otherwise there is no distinction between something that is actually at max-range and something that is beyond it. 

This is also what we expect in the range finder test, we actually expect [infinity](https://github.com/cyberbotics/webots/blob/b8d460f1d5c64b9c6500976a5115f9904b53be7d/tests/api/controllers/range_finder_checker/range_finder_checker.c#L69-L75).

The change was done for both the lidar and the range finder here: https://github.com/cyberbotics/webots/pull/2509

BenjaminHug8:
Closed as it seems lidar and range finder need more fixes.

PR #4167, opened 2022-01-28
Fix values greater than maxRange for the RangeFinder

BenjaminHug8:
**Description**
Values greater than `maxRange` for the `RangeFinder` should return `inf` but currently return `maxRange` value.


omichel (approved):
Well spotted!
I am wondering if the `isnan` tests are necessary in the `motion_blur.frag` file. Thanks to the fix in the `encode_depth.frag`, they shouldn't happen any more, isn't it?

BenjaminHug8:
The w component is used to normalize the coordinates of the vertices, it cannot be = 0.

omichel:
> The w component is used to normalize the coordinates of the vertices, it cannot be = 0.

In that case, we should remove the test and add a comment explaining why it cannot be 0.

omichel (approved):
Thank you.

PR #4171, opened 2022-01-31
Fix paperboat texture mapping

BenjaminDeleze:
**Description**
Since FLU the newspaper on the boat was not readable


omichel (approved):
Thank you.

PR #4176, opened 2022-02-02
Fix-robot-window

joachimhgg:
**Description**
Fix bug with update world tests. Fix override buttons for the generic window.


omichel (approved):
Thank you.

PR #4177, opened 2022-02-02
New HDR background: music hall.

omichel:
Add a new HDR background: `music_hall` to be used in the wrestling competition.


BenjaminDeleze (approved):
Thank you

It would be a good idea to convert the other background in `.jpg`. This would reduce their size and avoid to make an exception for the `music_hall` one.

PR #4178, opened 2022-02-02
Add warning for IS referring to unknown PROTO parameter

ad-daniel:
**Description**
Add a warning if the PROTO body references a PROTO parameter that doesn't exist.


src/webots/vrml/WbNode.cpp
```diff
@@ -935,9 +935,18 @@ void WbNode::readFields(WbTokenizer *tokenizer, const QString &worldPath) {
...
+          if (p->name() == alias) {
+            exists = true;
+            break;
+          }
+        }
+        if (exists) {
+          field->setAlias(alias);
+          copyAliasValue(field, alias);
+        } else
+          parsingWarn(tr("Field reference '%1' has no matching PROTO parameter.").arg(alias));
```
omichel, line 949: Maybe slightly more precise:
```suggestion
          parsingWarn(tr("Field IS reference '%1' has no matching PROTO parameter.").arg(alias));
```

omichel (changes requested):
Great.

omichel (approved):
Thank you.

PR #4180, opened 2022-02-02
Enhanced Nao PROTO

omichel:
The NAO proto was cleaned-up and improved.
This change breaks backward compatibility in the `version` and `color` / `customColor` fields.


ad-daniel (commented):
Only a minor change.

ad-daniel (approved):
Thank you!

PR #4182, opened 2022-02-02
Replaced cubic background PNG images with JPG images

omichel:
This should save a lot of download bandwidth.


BenjaminDeleze (approved):
Thank you

PR #4187, opened 2022-02-03
Line-ending test

ad-daniel:
**Description**
Added test that ensures line-ending conformity for the included assets.


ad-daniel:
The reason it was failing on Windows it that it seems git automatically converts to CRLF when cloning, which makes it fail the test. So for windows we need to specifically change git's config before the checkout action occurs.

tests/sources/test_line_ending.py
```diff
@@ -0,0 +1,48 @@
...
+
+class TestLineEnding(unittest.TestCase):
+    """Unit test of the line endings"""
+
+    def setUp(self):
+        """Get all files to be tested"""
+
+        self.files = []
+        for root_path, _, file_names in os.walk(os.environ['WEBOTS_HOME']):
+            for extension in ('*.proto', '*.wbt', '*.wbo', '*.forest'):
```
omichel, line 31: Why not adding more text files here? Like the *.c, *.h, *.cpp, *.hpp, *.py, *.java, *.m, *.md, etc.

omichel (approved):
Thank you.

omichel (approved):
Thank you.

PR #4190, opened 2022-02-03
Refresh docs for objects and fix unlinked PROTO

ad-daniel:
**Description**
 - regenerates the docs for objects in the `released` branch, some FLU related changes have not yet been updated.
 - removes unlinked fields (needs to be done in `released` since changing the proto header of objects requires regenerating the docs)


BenjaminDeleze (approved):
Thank you

PR #4194, opened 2022-02-04
Adds Franka Emika's panda robot

ad-daniel:
**Description**
Introduces Franka Emika's panda robot 

- [x] PROTO model (arm + hand)
- [x] demo world
- [x] demo controller
- [x] docs

preview: https://cyberbotics.com/doc/guide/panda?version=feature-franka-emika

panda wbt


omichel (changes requested):
Thank you.
It looks great.
We should add a line about it in the change log.
It seems the model takes long to load. Comparing the Niryo Ned robot, it is 5x heavier. Can't we reduce a little bit more the number of triangles?
Also, the compilation of the controller is not executed. It should be added in the robots/Makefile.
Finally, it seems the rotation centers of some joints are not all correctly positioned.

ad-daniel:
Yes, it was meant to be a temporary solution. As I've written here https://github.com/cyberbotics/webots/issues/4195 when this issue is addressed this robot can be simplified by using ColladaShapes (in the official repo, `.dae` files are provided for each of the seven links, so it would be the most natural way of doing it)

Until then, I've considered to convert the `IndexedFaceSets` to obj using `proto2mesh`, however each link is comprised of multiple sub-meshes (`PandaArmLink0` for instance has 12), hence I'm not super convinced having 40+ obj files is really faster than the current method. This is really a case where using `ColladaShape` makes sense, and if it's still heavy then, then those `.dae` can be decimated. Seems pointless to do things twice, then perhaps is best to wait for that to be fixed before merging this.

ad-daniel:
Bolted down
panda

omichel (approved):
Great. Thank you.

PR #4202, opened 2022-02-08
Fixed md5sum of libpico.zip on Windows

omichel:
I had to upload a new version of libpico.zip to be compliant with C++17 which is needed by Qt6.
This PR also fixes a compilation warning for lua which is back-ported from the develop branch.
Finally, it seems a new version of cppcheck on Windows was installed in the CI, so that many new cppcheck errors had to be fixed to pass the Windows cppcheck tests.


ad-daniel (approved):
Thank you!

Issue #4204, opened 2022-02-08
Some PROTO files are using IS reference with no matching PROTO parameter


BenjaminHug8:
**Describe the Bug**
A warning has been added to detect when the body of a PROTO use a `IS` statement without declaring the parameter in the head of the file.

One example is the mavic drone with the `battery` parameter (https://github.com/cyberbotics/webots/blob/master/projects/robots/dji/mavic/protos/Mavic2Pro.proto).

PR #4205, opened 2022-02-08
Fix unlinked field in `Mavic2Pro` and `PedestrianCrossing` PROTO

ad-daniel:
**Description**

Fixes https://github.com/cyberbotics/webots/issues/4204

I tried again to clean the proto cache and let webots rebuild it but no warnings are shown. Yet somehow the mavic does show a warning when opened, rightfully so. It means there might be more, but at this point I don't know how to find them other than opening all PROTO one by one.


BenjaminHug8 (changes requested):
As discussed together, it seems that the Tesla proto has also some warnings.

BenjaminHug8 (approved):
Thank you!

PR #4211, opened 2022-02-09
converted supervisor coordinates to display

ygoumaz:
**Description**
The matlab code was not converted to ENU coordinate system. The graph showing the robots position was wrong.


joachimhgg (approved):
thanks!

PR #4214, opened 2022-02-10
Fix-web-component

joachimhgg:
**Description**
regenerate robot web components. They were broken on develop because we changed the mountain background extension files.


BenjaminDeleze (approved):
Thank you!

PR #4216, opened 2022-02-10
Fix details on robot window

joachimhgg:
**Description**
Improve usage of the robot window.

**Tasks**
  - [x] remove double-click on PROTO robot
  - [x] automatically change port to allow multiple instances of webots to open multiple robot windows at the same time
  - [x] adapt simple custom robot window
  - [x] remove marge generic css
  - [x] fix slider on chrome


src/webots/gui/WbStreamingServer.cpp
```diff
@@ -78,10 +79,16 @@ void WbStreamingServer::setMainWindow(WbMainWindow *mainWindow) {
 }
 
 void WbStreamingServer::start(int port) {
+  mPort = port;
   try {
     create(port);
   } catch (const QString &e) {
     WbLog::error(tr("Error when creating the TCP streaming server on port %1: %2").arg(port).arg(e));
+    if ((WbPreferences::instance()->value("Streaming/port", 1234).toInt() + 10) > port) {
```
BenjaminDeleze, line 87: What is the use of this line?

BenjaminDeleze (changes requested):
- [x] I have some problems with multiple instances of webots:

- Open webots twice (with two different worlds) and try to open the robot windows
- First one is working fine (on 1234)
- Second one does not connect (open on 1234)

I got this error message but I think it is normal: `ERROR: Error when creating the TCP streaming server on port 1234: Cannot set the server in listen mode: The bound address is already in use
WARNING: Trying again with port 1235`

I think the problem comes from the `RobotWindow.js`: `    this.wsServer = decodeURI(getGETQueryValue('url', 'ws://localhost:1234/'));`
Try to update your branch with develop and see how I changed this line in `develop` and if it is possible to adapt it for this situation.
- [ ]  Would it be possible to change the css of generic.css to remove the horizontal scrollbar when it is not needed?

Otherwise it seems fine to me.

src/webots/gui/WbStreamingServer.cpp
joachimhgg: The +10 is to prevent the recursive function to try to many times (if the 10th next ports cannot be opened we should not continue to try).

joachimhgg:
> * [ ]  I have some problems with multiple instances of webots:
> 
>     * Open webots twice (with two different worlds) and try to open the robot windows
> 
>     * First one is working fine (on 1234)
> 
>     * Second one does not connect (open on 1234)
> 
> 
> I got this error message but I think it is normal: `ERROR: Error when creating the TCP streaming server on port 1234: Cannot set the server in listen mode: The bound address is already in use WARNING: Trying again with port 1235`
> 
> I think the problem comes from the `RobotWindow.js`: ` this.wsServer = decodeURI(getGETQueryValue('url', 'ws://localhost:1234/'));` Try to update your branch with develop and see how I changed this line in `develop` and if it is possible to adapt it for this situation.
> 
>     * [ ]   Would it be possible to change the css of generic.css to remove the horizontal scrollbar when it is not needed?
> 
> 
> Otherwise it seems fine to me.

It is the expected behavior. When opening a second instance, the `Streaming/port` preference is still `1234`. If the port cannot be opened (because already used but can be for other reasons), we displayed an error. Then, we alert the user that we will try with another port (it thus makes possible to open multiple instances from the same preferences). We will try the next 10 ports of the `Streaming/port` preference. I put 10 to allow the user to start 10 webots' instances from the same port preferences.

BenjaminDeleze:
> as the branch has already been updated, it is fine for me to use `decodeURL` with this default value.

My bad, I did not see that the branch was already up to date. Indeed using decodeURL is fine

joachimhgg:
the issue was indeed due to `decodeURI` introduce in #4208 . I changed it back to `'ws://' + window.location.host`. It does not seem to break your PR.

BenjaminDeleze (changes requested):

BenjaminDeleze (approved):
Thank you!

resources/web/wwi/RobotWindow.js
```diff
@@ -3,8 +3,7 @@ import {getGETQueryValue} from './request_methods.js';
 export default class RobotWindow {
   constructor(onready) {
     this.name = decodeURI(getGETQueryValue('name', 'undefined'));
-    this.wsServer = decodeURI(getGETQueryValue('url', 'ws://localhost:1234/'));
-
+    this.wsServer = window.location.href.substring(0, window.location.href.indexOf('robot_windows')).replace('https', 'wss').replace('http', 'ws');
```
omichel, line 6: We should be super careful with this string replacements as it introduces weaknesses in the code. For example a stronger version would be:
```js
this.wsServer = window.location.href.substring(0, window.location.href.indexOf('robot_windows')).replace('https://', 'wss://').replace('http://', 'ws://');
```
Note that in the current version the `replace('https', 'wss') is useless as it is covered by the `replace('http','ws')`...

Also the `indexOf('robot_windows') seems quite weak, can't we be more specific, e.g., `/robot_windows/`?

src/webots/gui/WbStreamingServer.hpp
```diff
@@ -44,6 +44,7 @@ class WbStreamingServer : public QObject {
   void sendToClients(const QString &message = "");
   void closeClient(const QString &clientID);
   bool getStreamStatus() { return mStream; }
+  int getPort() { return mPort; }
```
omichel, line 47: Coding style: C++ getters should be spelled `port()`, not `getPort()`.

resources/web/wwi/RobotWindow.js
BenjaminDeleze: _Also the indexOf('robot_windows') seems quite weak, can't we be more specific, e.g., /robot_windows/`?_

We can but we need to be sure to increment to also increment the index as we need to keep the first `/`

PR #3555, opened 2021-08-10
Conver Tracks to FLU

lukicdarkoo:

PR #4224, opened 2022-02-11
Fix bugs tracks

BenjaminDeleze:
Fix several issues related to tracks:
  - [x] During FLU conversion (#3555) a rotation was introduced in `WbTrack.wbt` instead of turning the geometries.
  - [x] The export in X3D comported two problems:
    - [x] It was not converted to FLU
    - [x] It was exported some node as `Transform USE=36` but the DEF node was a `Group`.
  - [x] The `USE` mechanism in `Parser.js` did not handle the case where the USE node had no defined ID.

### DIFF
--- a/tests/api/worlds/track.wbt
+++ b/tests/api/worlds/track.wbt
@@ -1,6 +1,7 @@
 #VRML_SIM R2022a utf8
 WorldInfo {
   basicTimeStep 8
+  coordinateSystem "NUE"
   contactProperties [
     ContactProperties {
       material1 "wheel material"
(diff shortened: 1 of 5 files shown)


tests/api/worlds/track.wbt
```diff
@@ -1,6 +1,7 @@
 #VRML_SIM R2022a utf8
 WorldInfo {
   basicTimeStep 8
+  coordinateSystem "NUE"
```
omichel, line 4: Do we really want this?

omichel (approved):

tests/api/worlds/track.wbt
BenjaminDeleze: Nearly all the tests are still `NUE`. I don't know if it is worth to convert them.

PR #4228, opened 2022-02-11
Fix missing irradiance warning message

ad-daniel:
**Description**
The warning messages for missing locally-defined irradiance file prints the url of the corresponding texture instead.


omichel (approved):
Well spotted!

PR #4230, opened 2022-02-14
Uploaded new libpico packages compatible with C++17

omichel:
Since Qt6 requires C++17, all the include files used by Webots should be C++17 compliant, including libpico, which wasn't.
So, I fixed the libpico package to be C++17 compliant and uploaded this new version.
Consequently the MD5 sums changed and need to be fixed.
#4202 fixed the MD5 sum for the Windows version.
This PR fixes the MD5 sums for the Linux and macOS versions.


BenjaminDeleze (approved):
Thank you

PR #4238, opened 2022-02-15
Replace QNetworkDiskCache with custom cache

ad-daniel:
**Description**
This PR introduces a different caching system for assets. Images, meshes, sounds (and eventually PROTO) are downloaded on the fly and saved in the `.cache/Cyberbotics/assets/` folder. The cache is opaque but deterministic, meaning each remote URL corresponds to a unique local file (identified by the SHA1 of the URL). As such, the assets folder can be:
- distributed as a zip file, allowing for offline usage of Webots and faster loading in webots.cloud
- the size of the distribution can be reduced further (PROTO will be defined as EXTERNPROTO references for their download)
- as the URLs refer to a specific release (defined by a tag), support for different versions of these assets (most notably PROTO) can be ensured.

Maximal cache size is defined but not strictly enforced (not without a restart). Meaning if the max size is defined to 512 MB, and loading a world results in this size being exceeded (for instance 600MB), it will require restarting Webots for the excess to be purged (by removing first the least used ones (oldest `lastRead`). 

Cached files are:
- [x] Backgrounds
- [x] Textures
- [x] Camera noise mask
- [x] Sounds
   - [x] Motor
   - [x] ContactProperties
   - [x] Speaker
- [x] Meshes 
- [x] Add test in test suite
   - [x] Textures
   - [x] Meshes
- [x] Cleanup/Testing


src/webots/core/WbNetwork.hpp
```diff
@@ -15,24 +15,38 @@
...
 private:
   static void cleanup();
   WbNetwork();
   ~WbNetwork();
+
+  void recomputeCacheSize();
+  static bool lastReadLessThan(QFileInfo &f1, QFileInfo &f2);
+
+  // cppcheck-suppress functionStatic
+  const QString urlToHash(const QString &url) const;
```
omichel, line 45: Why can't we do that?
```suggestion
  static const QString urlToHash(const QString &url) const;
```

src/webots/gui/WbPreferencesDialog.cpp
```diff
@@ -194,8 +194,16 @@ void WbPreferencesDialog::accept() {
   prefs->sync();
   if (changed)
     WbNetwork::instance()->setProxy();
-  if (!mCacheSize->text().isEmpty())
+
+  if (mCacheSize->text().toInt() < 512 || mCacheSize->text().isEmpty()) {
+    WbMessageBox::info(tr("For optimal performance, at least\n 512 MB of cache are necessary."), this);
```
omichel, line 199:
```suggestion
    WbMessageBox::info(tr("For optimal performance, at least\n512 MB of cache are necessary."), this);
```

src/webots/nodes/WbContactProperties.cpp
```diff
@@ -198,25 +210,25 @@ void WbContactProperties::updateSoftErp() {
...
+  // determine extension from url since for remotely defined assets the cached version doesn't retain this information
+  const QString extension = sound.mid(sound.lastIndexOf('.') + 1).toLower();
+
+  if (WbUrl::isWeb(sound)) {
+    assert(WbNetwork::instance()->isCached(sound));  // by this point, the asset should be cached
+    *clip = WbSoundEngine::sound(WbNetwork::instance()->get(sound), extension);
+  } else
+    *clip = WbSoundEngine::sound(WbUrl::computePath(this, name, sound), extension);
+
+  // TODO: need to delete?
```
omichel, line 231: This TODO should be addressed.

src/webots/nodes/WbContactProperties.cpp
```diff
@@ -20,6 +20,8 @@
 #include "WbUrl.hpp"
 #include "WbWorld.hpp"
 
+enum { BUMP, ROLL, SLIDE };
+
```
omichel, line 24: This enum seems unused.
```suggestion
```

src/webots/nodes/WbImageTexture.cpp
```diff
@@ -129,22 +131,22 @@ void WbImageTexture::postFinalize() {
...
+  const QString &url = mUrl->item(0);
+  const bool isWebAsset = WbUrl::isWeb(url);
+  if (isWebAsset && !WbNetwork::instance()->isCached(url))
     return false;
+
+  QString filePath;
+  if (isWebAsset)
+    filePath = WbNetwork::instance()->get(url);
+  else
+    filePath = path(true);
```
omichel, line 143:
```suggestion
  const QString filePath = isWebAsset ? WbNetwork::instance()->get(url) : path(true);
```

src/webots/nodes/WbMotor.cpp
```diff
@@ -35,6 +35,7 @@
 #include <ode/ode.h>
 
 #include <QtCore/QDataStream>
+#include <QtCore/QFile>
```
omichel, line 38: Seems useless:
```suggestion
```

src/webots/nodes/utils/WbDownloader.hpp
```diff
@@ -15,6 +15,8 @@
 #ifndef WB_DOWNLOADER_HPP
 #define WB_DOWNLOADER_HPP
 
+#include "WbNetwork.hpp"
+
```
omichel, line 19: That seems useless:
```suggestion
```

src/webots/nodes/utils/WbDownloader.cpp
```diff
@@ -15,16 +15,13 @@
 #include "WbDownloader.hpp"
 
 #include "WbApplication.hpp"
-#include "WbNetwork.hpp"
```
omichel, line 18: We should probably keep it here and not in the header file.

tests/api/controllers/cache_checker/cache_checker.c
```diff
@@ -0,0 +1,117 @@
...
+      "In test %d: wrong texture color in object %d, received color = (%d, %d, %d) but expected color = (%d, %d, %d)",
+      test_number, i, r, g, b, expected_color[0], expected_color[1], expected_color[2]);
+  }
+}
+
+void test_distance(int test_number, const double expected_distance) {
+  for (int i = 0; i < NB_MESHES; ++i) {
+    const double distance = wb_distance_sensor_get_value(distance_sensors[i]);
+    ts_assert_double_in_delta(distance, expected_distance, 1e-5,
+                              "In test %d: wrong mesh is applied (distance is %f instead of %f)", distance, expected_distance);
```
omichel, line 37:
```suggestion
                              "In test %d: wrong mesh %d is applied (distance is %f instead of %f)", test_number, i, distance, expected_distance);
```

omichel (changes requested):
This is a first code review.
It looks overall good to me.
Only the documentation should be updated in a different way.

ad-daniel:
Looks like removing the `md` files from `update_urls` makes it unnecessary to force the encoding

omichel (dismissed):
Since it is a quite big PR, it would be nice to have a second review.

src/webots/core/WbNetwork.cpp
```diff
@@ -93,6 +102,103 @@ void WbNetwork::setProxy() {
...
+  // cppcheck-suppress assertWithSideEffect
+  assert(isCached(url));  // the 'get' function should not be called unless we know that the file is cached
+
+  if (cacheMap.contains(url))
+    return cacheMap[url];
+
+  const QString location = mCacheDirectory + urlToHash(url);
+  cacheMap.insert(url, location);
+
+  return location;
```
BenjaminDeleze, line 130: Is it possible to reach this code ? 
The `assert(isCached(url))` return true only if `cacheMap` already contains the url or if the file exists on the disk and in the latest case, it inserts it in the `cacheMap`.
So if we pass the assert, the test at line 124 will always pass?

BenjaminDeleze (changes requested):
I managed to crash webots:

- Open `nao_room.wbt`
- Go in `Worldinfo` -> `contactProperties`
- Select one `ContactProperties` and change one of its urls to an invalid one
- Webots crash.

Error message: ```Warning: QIODevice::read (QFile, "https://raw.githubusercontent.com/cyberbotics/webots/R2022b/projects/default/worklds/sounds/slide.wav"): device not open
terminate called after throwing an instance of 'char const*'
./webots: line 96: 115567 Aborted                 (core dumped) "$webots_home/bin/webots-bin" "$@"```

src/webots/core/WbNetwork.cpp
ad-daniel: The assert is a sanity check when developing. If it's triggered it means we didn't implement the retrieval logic correctly, in other words we shouldn't have reached to point of calling the get function unless the file is in the local cache.

`cacheMap` is not a complete representation of the state of the cache, it's a representation that we build progressively as assets are requested/returned, and the information is lost when closing Webots. 
By keeping this internal representation allows to skip having to compute the hash from the url and checking if the file exists on the disk every time, which otherwise we would do even when the same url was requested before (or rather, in the current session).

omichel:
When clearing the cache from the Preferences window, Webots doesn't close the Preference window. When changing the cache size, it closes the Preferences window. This is not consistent. I believe changing the cache size should not close the Preferences window.

omichel:
Also, when opening the Preferences windows, doing not and pressing the "OK" button, we get a message window saying the cache size was updated. This should not happen.

BenjaminDeleze (changes requested):

BenjaminDeleze:
> I managed to crash webots:
> 
> * Open `nao_room.wbt`
> * Go in `Worldinfo` -> `contactProperties`
> * Select one `ContactProperties` and change one of its urls to an invalid one
> * Webots crash.
> 
> Error message: `Warning: QIODevice::read (QFile, "https://raw.githubusercontent.com/cyberbotics/webots/R2022b/projects/default/worklds/sounds/slide.wav"): device not open terminate called after throwing an instance of 'char const*' ./webots: line 96: 115567 Aborted (core dumped) "$webots_home/bin/webots-bin" "$@"`

It crashes also when changing the url of the sound of a motor

BenjaminDeleze (changes requested):
I found another crash (in a very rare edge case):
- Load a world.
- Clear the cash.
- Reload the world.

ad-daniel:
All issues should be addressed now, both with and without internet connection

BenjaminDeleze (dismissed):
Thank you!

I tried to make it bug in a lot of different ways but did not find any strange behaviors

src/webots/core/WbNetwork.cpp
```diff
@@ -13,16 +13,23 @@
...
 #include <QtCore/QStandardPaths>
 #include <QtNetwork/QNetworkAccessManager>
-#include <QtNetwork/QNetworkDiskCache>
 #include <QtNetwork/QNetworkProxy>
 
 static WbNetwork *gInstance = NULL;
 
+// cacheMap is an ephemeral (internal) representation of what is known about the cache at every session, as such it isn't
+// persistent nor is it ever complete. Its purpose is to speed up checking and retrieving previously referenced assets.
+static QMap<QString, QString> cacheMap;
```
omichel, line 31: We usually name these static variable with a `g` prefix.

src/webots/gui/WbPreferencesDialog.cpp
```diff
@@ -488,19 +497,23 @@ QWidget *WbPreferencesDialog::createNetworkTab() {
...
 
   // row 1
   QPushButton *clearCacheButton = new QPushButton(QString("Clear the cache"), this);
   connect(clearCacheButton, &QPushButton::pressed, this, &WbPreferencesDialog::clearCache);
-  layout->addWidget(new QLabel(tr("Amount of cache used : %1 MB.")
-                                 .arg(WbNetwork::instance()->networkAccessManager()->cache()->cacheSize() / (1024 * 1024)),
-                               this),
-                    1, 0);
+  layout->addWidget(
+    new QLabel(tr("Amount of cache used : %1 MB.").arg(WbNetwork::instance()->cacheSize() / (1024 * 1024)), this), 1, 0);
```
omichel, line 516:
```suggestion
    new QLabel(tr("Amount of cache used: %1 MB.").arg(WbNetwork::instance()->cacheSize() / (1024 * 1024)), this), 1, 0);
```

src/webots/nodes/WbCamera.cpp
```diff
@@ -156,17 +157,21 @@ WbCamera::~WbCamera() {
 
 void WbCamera::downloadAssets() {
   WbAbstractCamera::downloadAssets();
-  const QString &noiseMaskUrl = mNoiseMaskUrl->value();
-  if (!noiseMaskUrl.isEmpty()) {
+
+  QString noiseMaskUrl = mNoiseMaskUrl->value();
```
omichel, line 161: Why can't we keep it const ref?

src/webots/nodes/WbImageTexture.cpp
```diff
@@ -129,22 +132,17 @@ void WbImageTexture::postFinalize() {
...
+  const QString &url = mUrl->item(0);
+  const bool isWebAsset = WbUrl::isWeb(url);
+  if (isWebAsset && !WbNetwork::instance()->isCached(url))
     return false;
+
+  const QString filePath = isWebAsset ? WbNetwork::instance()->get(url) : path(true);
   QFile file(filePath);
-  if (!file.open(QIODevice::ReadOnly))
+  if (!file.open(QIODevice::ReadOnly)) {
+    warn(tr("Texture file could not be read: %1").arg(filePath));
```
omichel, line 143:
```suggestion
    warn(tr("Texture file could not be read: '%1'").arg(filePath));
```

src/webots/nodes/utils/WbDownloader.cpp
```diff
@@ -109,28 +92,16 @@ void WbDownloader::download(const QUrl &url) {
...
-    // The hack we use to detect if the expiration date of an asset is the one set by webots or is the automatic one works as
-    // follows: The automatic expiration date is an <http-date> and thus is in UTC format. The expiration date set by webots is
-    // in local format. So we just need to check the format of the expiration date to know if it needs to be updated or not.
-    if (metaData.expirationDate().toUTC().toString() == metaData.expirationDate().toString()) {
-      // increase expiration date to one day
-      metaData.setExpirationDate(QDateTime::currentDateTime().addDays(1));
-      WbNetwork::instance()->networkAccessManager()->cache()->updateMetaData(metaData);
+  // cache result
+  if (mNetworkReply && mNetworkReply->error()) {
+    mError = tr("Cannot download %1, error code: %2: %3")
```
omichel, line 97:
```suggestion
    mError = tr("Cannot download '%1', error code: %2: %3")
```

omichel (changes requested):
A few requests...

src/webots/nodes/WbMesh.cpp
```diff
@@ -122,32 +127,29 @@ void WbMesh::updateTriangleMesh(bool issueWarnings) {
...
-    } else
+    if (!WbNetwork::instance()->isCached(filePath)) {
+      if (mDownloader == NULL)  // never attempted to download it, try now
+        downloadAssets();
+      return;
+    }
+
+    QFile file(WbNetwork::instance()->get(filePath));
+    if (!file.open(QIODevice::ReadOnly)) {
+      warn(tr("Mesh file could not be read: %1").arg(filePath));
```
omichel, line 147:
```suggestion
      warn(tr("Mesh file could not be read: '%1'").arg(filePath));
```

src/webots/wren/WbWrenCamera.cpp
```diff
@@ -327,19 +328,24 @@ void WbWrenCamera::setRangeResolution(float resolution) {
...
+    if (!noiseMask.open(QIODevice::ReadOnly))
+      return tr("Cannot open noise mask file: '%1'").arg(noiseMaskUrl);
+
     QImage *image = new QImage();
-    QImageReader *imageReader = device ? new QImageReader(device) : new QImageReader(noiseMaskTexturePath);
+    QImageReader *imageReader = new QImageReader(noiseMaskUrl);
     if (!imageReader->read(image)) {
       delete image;
-      return tr("Cannot load %1: %2").arg(noiseMaskTexturePath).arg(imageReader->errorString());
+      return tr("Cannot load %1: %2").arg(noiseMaskUrl).arg(imageReader->errorString());
```
omichel, line 348:
```suggestion
      return tr("Cannot load '%1': %2").arg(noiseMaskUrl).arg(imageReader->errorString());
```

omichel (changes requested):

src/webots/nodes/WbCamera.cpp
ad-daniel: my bad, it can and should be

omichel (approved):
Thank you.

PR #4242, opened 2022-02-16
Track animation in webotsJS

BenjaminDeleze:
**Tasks**
  - [x] Animation of the Trackwheels
  - [x] Animation of the Track

**Additional context**
Concerning the Trackwheels, it had already a rotation attribute in Webots but it was an internal one. It was enough to turn this internal rotation into a hidden field for it to be automatically exported.

Concerning the track itself, it is more tricky because the elements of the track are not real nodes: they are created and handled directly by the track node. Following their evolution would be difficult and require to work in another way than what is already implemented (based on fields).
To bypass this problem, the movement of the track will be implemented directly in WebotsJS.

I had to force the export of the trackwheels to x3d, even if they do not have children because we need to know how many wheels there is to compute the path of the track segments.


BenjaminDeleze:
It needs to be done on develop as it changed the way `solid` are recognized. (`isSolid/isPropeller=true` => `type=solide/propeller`)

Issue #4240, opened 2022-02-16
Change of mesh url using supervisor not propagating in `boundingObject`


ad-daniel:
**Describe the Bug**
Changing the url of a `Mesh` in a solid (using the supervisor or manually in the scene tree) works as intended (both for PROTO/nested PROTO and Base Node), however if the shape containing this `Mesh` is also used as a bounding object, the latter doesn't appear to be updated when the change occurs. Or rather, it updates for base node but doesn't if within a PROTO.

**Steps to Reproduce**
1. Create a Solid such as:
```
Solid {
  children [
    DEF SHAPE_PROTO ProtoWithMesh {
      meshUrl [
        "webots://tests/api/worlds/meshes/cube_0.05m.obj"
      ]
    }
  ]
  name "solid1"
  boundingObject USE SHAPE_PROTO
}
``` 
where `ProtoWithMesh` is:
```
#VRML_SIM R2022b utf8

PROTO ProtoWithMesh
[
  field MFString meshUrl []
]
{
  Shape {
    appearance PBRAppearance {}
    geometry Mesh {
      url IS meshUrl
    }
  }
}
```
2. Manually, or using a supervisor, change the `meshUrl` at the level of the solid
3. Visually speaking, the change occurs correctly however the `boundingObject` does not update.

PR #4245, opened 2022-02-16
Fix notification of field change in a PROTO not propagating through DEF/USE

ad-daniel:
**Description**
Fixes https://github.com/cyberbotics/webots/issues/4240

The bug doesn't affect just meshes, any shape with a primitive as geometry doesn't propagate either if within a PROTO.

The bug lies in [this](https://github.com/cyberbotics/webots/blob/f114270206dc12f345fb86c67adce3fa6a93ec84/src/webots/vrml/WbNode.cpp#L753-L799) function. When a field changes, be it manually or from supervisor, this function is supposed to propagate the change to all [USE-nodes](https://github.com/cyberbotics/webots/blob/f114270206dc12f345fb86c67adce3fa6a93ec84/src/webots/vrml/WbNode.cpp#L779-L781) as well.
The functions that determine whether this notification has to occur (findSubFieldIndex and findSubField) however only search for a matching field (i.e a field with same name and type as the one that triggered the notification) only at the top level. Namely, in this case it searches only in the fields of the `Shape` node, which for the base node case entails checking `Shape`, `PBRAppearance` and `Mesh` where ultimately it finds it and everything works as intended. If the `Shape` is however defined within a PROTO, only the PROTO itself is checked, without checking if the trigger might have been in the appearance or geometry fields itself (i.e the nodes `PBRAppearance` and `Mesh` are not checked), hence why the notification doesn't occur.

As far as I can tell, these functions are not used elsewhere so it should be safe to do directly the change withing them, by forcing the search also in the sub nodes themselves.


omichel (approved):

ad-daniel:
I've also fixed the issue regarding the incorrect update of the mesh on the ODE side resulting from a change of url and added a sanity check for the url as well (a warning message is shown in case the resource cannot be found). This approach seemed the cleaner solution, it's pointless to even attempt to build a wren mesh (and everything that comes with it namely liberating the previous one from ODE, updating the mesh cache, ...) if we know the url is not valid to begin with.

It isn't necessary to add a test for this second aspect since it is implicitly being tested in https://github.com/cyberbotics/webots/pull/4238 (to test if the cache works this PR changes the mesh 3 times using a supervisor)

omichel (approved):
Well spotted!
Thank you.

PR #4248, opened 2022-02-18
Prevent failure in one job (within a matrix) to fail the other

ad-daniel:
**Description**
For debugging purposes it might be beneficial to know if a branch fails exclusively on ubuntu 18/20 or in both at the same time. Currently, the github actions are interrupted as soon as any of the parallel jobs within a matrix fails, this PR removes this condition.

I've taken the opportunity to update the version the `test world update` runs on, I don't see a reason why it should run on ubuntu 18.

### DIFF
--- a/.github/workflows/test_suite_linux.yml
+++ b/.github/workflows/test_suite_linux.yml
@@ -93,6 +93,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-18.04, ubuntu-20.04]
+        fail-fast: false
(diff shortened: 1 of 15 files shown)


omichel:
> I've taken the opportunity to update the version the `test world update` runs on, I don't see a reason why it should run on ubuntu 18.

Yes, we should move away from Ubuntu 18 as we will soon drop it and adopt Ubuntu 22.

omichel (approved):
Thank you.

PR #4249, opened 2022-02-18
Fixed compilation on Windows failing after gcc update

omichel:
Ditto


BenjaminDeleze (commented):
It works for me.
But isn't it strange to have `OPENAL_INCLUDE     =`  with nothing on the right of the equal ?

omichel:
> But isn't it strange to have `OPENAL_INCLUDE =` with nothing on the right of the equal ?

True, we can indeed remove it completely, undefined variables are by default empty in Makefile.

BenjaminDeleze (approved):
Thank you!

PR #4253, opened 2022-02-18
dockerfile-server-implementation

omichel:
**Description**

Pass the Webots tmp folder path as a build argument to use it in a `Dockerfile`.


BenjaminDeleze (approved):
Thank you

PR #4256, opened 2022-02-18
Add anchor rendering

ad-daniel:
**Description**
As it's always a pain not knowing where the anchor point is, this PR adds the rendering of it.
It scales a little with linescale but only to a certain point, otherwise it becomes useless.
It's set to render above everything else so it's always visible (if the joint axis are visible).
The color I've picked is darkish red, but I'm open to suggestions.

a
b
c
d


omichel (approved):
Looks good to me.

PR #4257, opened 2022-02-18
Forbid USE of TrackWheel

BenjaminDeleze:
**Description**
USE of TrackWheel are a bad practice, here is why:

- Define two tracks: T1 and T2.
- Define a TrackWheel in T1 with a DEF.
- Insert a USE of this TrackWheel in T2.
- Run T1 at speed v1 and T2 at speed v2.
- The USEd TrackWheel in T2 will go at speed v1 which is wrong.

So it is better to forbid the USE of TrackWheel.

(I took advantage of this PR to harmonize a bit the naming in `track.wbt`.)

Does it need to add an entry in the changelog?

### DIFF
--- a/docs/reference/changelog-r2022.md
+++ b/docs/reference/changelog-r2022.md
@@ -8,7 +8,7 @@ Released on XX XX, 2022.
     - Added a new HDR background: `music_hall` ([#4177](https://github.com/cyberbotics/webots/pull/4177)).
     - Replaced cubic background PNG images with more efficient JPG images ([#4182](https://github.com/cyberbotics/webots/pull/4182)).
     - Changed the way MATLAB is detected in the system using a new Webots preference ([#4233](https://github.com/cyberbotics/webots/pull/4233)).
-    - Forbid the USE of [TrackWheel](trackwheel.md) to avoid wrong behaviour ([#4257](https://github.com/cyberbotics/webots/pull/4257)).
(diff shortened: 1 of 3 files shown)


omichel (changes requested):
Yes, I believe it would be good to add a line about it in the change log, as some people may have proto or world files USing TrackWeel nodes and won't understand why they don't work any more.

docs/reference/changelog-r2022.md
```diff
@@ -8,6 +8,7 @@ Released on XX XX, 2022.
     - Added a new HDR background: `music_hall` ([#4177](https://github.com/cyberbotics/webots/pull/4177)).
     - Replaced cubic background PNG images with more efficient JPG images ([#4182](https://github.com/cyberbotics/webots/pull/4182)).
     - Changed the way MATLAB is detected in the system using a new Webots preference ([#4233](https://github.com/cyberbotics/webots/pull/4233)).
+    - Forbid the USE of [TrackWheel](trackwheel.md) to avoid wrong behaviour ([#4257](https://github.com/cyberbotics/webots/pull/4257)).
```
omichel, line 11: American English:
```suggestion
    - Forbid the USE of [TrackWheel](trackwheel.md) to avoid wrong behavior ([#4257](https://github.com/cyberbotics/webots/pull/4257)).
```

omichel (changes requested):

omichel (approved):
Thank you.

PR #4263, opened 2022-02-21
Sanitize urls in webotsjs

BenjaminDeleze:
- Refactor urls handling of background.
- Handle the case where there is several urls in the same variable (take only the first one).
- Handle the case where the url was not surrounded by quotes.


omichel (approved):
Looks good to me.

PR #4233, opened 2022-02-14
Add new matlab command preference

ygoumaz:
**Description**
Until now, Matlab is started by Webots by searching in the PATH of the system. This approach could cause problems, especially on MacOS where PATHs are managed differently between the terminal and the Finder (see issue #2608). Also, symlinks were needed on some operating systems.

To simplify the procedure for the user, this PR aims to create a General Webots preference to specify the absolute path to the Matlab executable. This way the user also has more control over which version of Matlab to use. When a Matlab controller is started, the specified executable is directly used as the start command.

**Related Issues**
This pull-request does not directly fix issue #2608, but brings an easier way to handle Matlab execution.

**Tasks**
  - [x] Add Matlab command preference in WbPreferences and WbPreferencesDialog.
  - [x] Start Matlab using the preference field from WbController.
  - [x] Update documentation.
  - [x] Update changelog.

**Documentation**
https://cyberbotics.com/doc/guide/using-matlab?version=fix-matlab-preference-path

**Screenshots**
New matlab command field in the preferences:
matlab_pref

PR #4265, opened 2022-02-21
modified behavior of default matlab command

ygoumaz:
**Description**
PR #4233 was adding a new MATLAB preference to indicate a custom MATLAB path. Its first behavior was to set the default MATLAB installation path in the preference when starting Webots for the first time.

This implementation is problematic in case the user installs MATLAB after Webots. 

This PR removes the default preference settings. Instead, when the user runs a MATLAB controller, Webots first checks the preference field. If it is empty, it will check the default installation path of MATLAB. If both paths are missing, an error message is displayed.

### DIFF
--- a/src/webots/control/WbLanguageTools.cpp
+++ b/src/webots/control/WbLanguageTools.cpp
@@ -211,6 +211,7 @@ QString WbLanguageTools::matlabCommand() {
   const QString matlabExecPath = "\\bin\\win64\\MATLAB.exe";
 #else  // __linux__
   const QString matlabPath = "/usr/local/MATLAB/";
+  // cppcheck-suppress unreadVariable
   const QString matlabExecPath = "/bin/matlab";
 #endif
   const QDir matlabDir(matlabPath);
@@ -227,7 +228,6 @@ QString WbLanguageTools::matlabCommand() {
 #endif
 
   return command;
-  // setDefault("General/matlabCommand", command);
 }
 
 const QStringList WbLanguageTools::matlabArguments() {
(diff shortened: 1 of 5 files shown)


src/webots/control/WbLanguageTools.cpp
```diff
@@ -194,6 +194,42 @@ const QStringList WbLanguageTools::pythonArguments() {
...
+QString WbLanguageTools::matlabCommand() {
+#ifdef __APPLE__
+  const QString matlabPath = "/Applications/";
+  const QString matlabAppWc = "MATLAB_R20???.app";
+  const QDir matlabDir(matlabPath);
+  const QStringList matlabVersions = matlabDir.entryList(QStringList() << matlabAppWc, QDir::Files, QDir::Name);
+  if (matlabVersions.isEmpty()) {
+    // setDefault("General/matlabCommand", "");
+    return "";
+  }
```
omichel, line 206:
```suggestion
  if (matlabVersions.isEmpty())
    return "";
```

src/webots/control/WbLanguageTools.cpp
```diff
@@ -194,6 +194,42 @@ const QStringList WbLanguageTools::pythonArguments() {
...
+#else  // __linux__
+  const QString matlabPath = "/usr/local/MATLAB/";
+  // cppcheck-suppress unreadVariable
+  const QString matlabExecPath = "/bin/matlab";
+#endif
+  const QDir matlabDir(matlabPath);
+  if (!matlabDir.exists()) {
+    // setDefault("General/matlabCommand", "");
+    return "";
+  }
```
omichel, line 221:
```suggestion
  if (!matlabDir.exists())
    return "";
```

omichel (changes requested):
That looks good to me.

omichel (approved):
Thank you.

PR #4269, opened 2022-02-22
Fix error make distrib

ad-daniel:
**Description**
When generating the proto cache, `make distrib` complains that [this proto](https://github.com/cyberbotics/webots/blob/develop/resources/templates/protos/template.proto) is invalid, which is normal since it's only the skeleton of one, not a fully fledged one.
As such, when searching protos recursively we should skip this directory. To my knowledge at the moment there is only this exception hence why I've added it directly in the function.


omichel:
I believe I already fixed this problem in  #4189 [here](https://github.com/cyberbotics/webots/pull/4189/commits/ae166c7915294e66c834e229fe20364c83fc1bd1).

ad-daniel:
We can close this then, your fix is nicer too

PR #4270, opened 2022-02-22
Fix url search in distributions

ad-daniel:
**Description**
`WbApplicationInfo::repo()` and `WbApplicationInfo::branch()` rely on text files that are not included in the distribution, hence if searching for the file locally fails, the search of its remote counterpart will fail as well since the url is malformed.

### DIFF
--- a/src/webots/nodes/utils/WbUrl.cpp
+++ b/src/webots/nodes/utils/WbUrl.cpp
@@ -122,9 +122,9 @@ QString WbUrl::computePath(const WbNode *node, const QString &field, const QStri
 
     if (isLocalUrl(url)) {
       QString newUrl(url);
-      WbVersion version = WbApplicationInfo::version();
+      const WbVersion version = WbApplicationInfo::version();
       // if it's an official release, use the tag (for example R2022b), if it's a nightly use the commit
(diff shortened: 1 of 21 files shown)


BenjaminDeleze (approved):
Thank you!

src/webots/nodes/utils/WbUrl.cpp
```diff
@@ -122,8 +122,10 @@ QString WbUrl::computePath(const WbNode *node, const QString &field, const QStri
 
     if (isLocalUrl(url)) {
       QString newUrl(url);
-      newUrl.replace("webots://", "https://raw.githubusercontent.com/" + WbApplicationInfo::repo() + "/" +
-                                    WbApplicationInfo::branch() + "/");
+      WbVersion version = WbApplicationInfo::version();
```
BenjaminDeleze, line 125:
```suggestion
      WbVersion version = WbApplicationInfo::version();
```

Would it be possible to make a const variable here ?

ad-daniel: Yes, both can actually

BenjaminDeleze (approved):
Thank you

Issue #4195, opened 2022-02-04
`ColladaShape` proto not usable in webots library


ad-daniel:
**Describe the Bug**
The `ColladaShape` PROTO currently isn't usable in webots library, meaning we can't have included worlds/objects that define their shape using this PROTO. If the url is provided as `webots://` or as `http://`, the existence check test [here](https://github.com/cyberbotics/webots/blob/df8ccec44920ca0eaa0137c835e2f77b9d8426ac/projects/shapes/protos/ColladaShapes.proto#L13) fails since this function only tests if a [local](https://github.com/cyberbotics/webots/blob/df8ccec44920ca0eaa0137c835e2f77b9d8426ac/src/webots/core/WbQjsFile.cpp#L22-L29) file exists, it doesn't sanitize the path. Even if we were to skip this check, the url is fed to Assimp ***as-is*** [here](https://github.com/cyberbotics/webots/blob/df8ccec44920ca0eaa0137c835e2f77b9d8426ac/src/webots/nodes/utils/WbNodeOperations.cpp#L490-L492), which if it starts with `webots://` or `http://` it won't know how to handle.

EDIT: when this issue has been addressed, the PROTO of the Panda robot can be simplified

PR #4285, opened 2022-02-28
Feature `CadShape` node

ad-daniel:
**Description**
To address this issue: https://github.com/cyberbotics/webots/issues/4195, this PR creates a full node (previously a PROTO) for the importing of collada objects

- [x] Add fields:
  - [x] url as MFString
  - [x] ccw
  - [x] castShadows
  - [x] isPickable
- [x] add segmentation and range finder visibility
- [x] add downloader
- [x] prevent it from being used in bounding object
- [x] replace pbr tokenizer init by direct init (and fix crash)
- [x] remove colladashape proto related code
- [x] remove model import and VRML export
- [x] add test 
- [x] docs https://cyberbotics.com/doc/reference/colladashape?version=feature-collada-shape-node
- [x] changelog (with backwards compatibility notice ColladaShapes -> ColladaShape)

Following the discussion here https://github.com/cyberbotics/urdf2webots/issues/168, the node should be made more generic and support both wavefront and collada files

- [x] add obj support
- [x] add test
- [x] update docs


src/webots/nodes/WbAbstractAppearance.cpp
```diff
@@ -37,6 +40,23 @@ WbAbstractAppearance::WbAbstractAppearance(const WbNode &other) : WbBaseNode(oth
...
+  mName = new WbSFString(QString(name.C_Str()));
+  mNameValue = mName->value();
+
+  mTextureTransform = new WbSFNode(NULL);
+}
+
+WbAbstractAppearance::~WbAbstractAppearance() {
+  delete mName;
+  if (mTextureTransform)
+    delete mTextureTransform;
```
omichel, line 57: It is OK to call delete with a NULL argument (it will simply do nothing):
```suggestion
  delete mTextureTransform;
```

src/webots/nodes/WbPbrAppearance.cpp
```diff
@@ -68,6 +70,77 @@ WbPbrAppearance::WbPbrAppearance(const WbNode &other) : WbAbstractAppearance(oth
...
+  WbAbstractAppearance("PBRAppearance", material) {
+  aiColor3D baseColor(1.0f, 1.0f, 1.0f);
+  material->Get(AI_MATKEY_COLOR_DIFFUSE, baseColor);
+  mBaseColor = new WbSFColor(baseColor[0], baseColor[1], baseColor[2]);
+
+  aiColor3D emissiveColor(0.0f, 0.0f, 0.0f);
+  material->Get(AI_MATKEY_COLOR_EMISSIVE, emissiveColor);
+  mEmissiveColor = new WbSFColor(emissiveColor[0], emissiveColor[1], emissiveColor[2]);
+
+  float opacity = 1.0f;
```
omichel, line 83:
```suggestion
  const float opacity = 1.0f;
```

src/webots/nodes/WbColladaShape.cpp
```diff
@@ -0,0 +1,508 @@
...
+const QString WbColladaShape::vrmlPbrAppearance(const aiMaterial *material) {
+  QString vrml;
+
+  aiColor3D baseColor(1.0f, 1.0f, 1.0f);
+  material->Get(AI_MATKEY_COLOR_DIFFUSE, baseColor);
+
+  aiColor3D emissiveColor(0.0f, 0.0f, 0.0f);
+  material->Get(AI_MATKEY_COLOR_EMISSIVE, emissiveColor);
+
+  float opacity = 1.0f;
```
omichel, line 411:
```suggestion
  const float opacity = 1.0f;
```

src/webots/nodes/WbPbrAppearance.cpp
```diff
@@ -76,6 +149,31 @@ WbPbrAppearance::~WbPbrAppearance() {
...
+    if (mRoughnessMap)
+      delete mRoughnessMap;
+    if (mMetalnessMap)
+      delete mMetalnessMap;
+    if (mNormalMap)
+      delete mNormalMap;
+    if (mOcclusionMap)
+      delete mOcclusionMap;
+    if (mEmissiveColorMap)
+      delete mEmissiveColorMap;
```
omichel, line 175: No need to test for NULL before deleting objects.

omichel (changes requested):
I believe the approach is pretty good.
Maybe it would help a little bit to add a comment to the shallow constructors so that we understand that these constructors should be used only in the special case where we want to build a shallow version of the node.

src/webots/vrml/WbSFColor.hpp
```diff
@@ -27,6 +27,7 @@ class WbSFColor : public WbSingleValue {
   Q_OBJECT
 
 public:
+  explicit WbSFColor(double r, double b, double g) : mValue(WbRgb(r, g, b)){};
```
omichel, line 30:
```suggestion
  explicit WbSFColor(double r, double b, double g) : mValue(WbRgb(r, g, b)) {}
```

src/webots/vrml/WbSFString.hpp
```diff
@@ -26,6 +26,7 @@ class WbSFString : public WbSingleValue {
   Q_OBJECT
 
 public:
+  explicit WbSFString(const QString &s) : mValue(s){};
```
omichel, line 29:
```suggestion
  explicit WbSFString(const QString &s) : mValue(s) {}
```

src/webots/nodes/WbColladaShape.cpp
```diff
@@ -0,0 +1,529 @@
...
+  deleteWrenObjects();
+
+  if (mUrl->size() == 0)
+    return;
+
+  Assimp::Importer importer;
+  importer.SetPropertyInteger(AI_CONFIG_PP_RVC_FLAGS, aiComponent_CAMERAS | aiComponent_LIGHTS | aiComponent_BONEWEIGHTS |
+                                                        aiComponent_ANIMATIONS | aiComponent_COLORS);
+
+  unsigned int flags = aiProcess_ValidateDataStructure | aiProcess_Triangulate | aiProcess_GenSmoothNormals |
```
omichel, line 199:
```suggestion
  const unsigned int flags = aiProcess_ValidateDataStructure | aiProcess_Triangulate | aiProcess_GenSmoothNormals |
```

src/webots/nodes/WbColladaShape.cpp
```diff
@@ -0,0 +1,529 @@
...
+      const aiMesh *mesh = scene->mMeshes[node->mMeshes[i]];
+
+      // compute absolute transform of this node from all the parents
+      const int vertices = mesh->mNumVertices;
+      const int faces = mesh->mNumFaces;
+      if (vertices < 3)  // silently ignore meshes with less than 3 vertices as they are invalid
+        continue;
+
+      if (vertices > 100000)
+        warn(tr("mesh '%1' has more than 100'000 vertices, it is recommended to reduce the number of vertices.")
```
omichel, line 273:
```suggestion
        warn(tr("Mesh '%1' has more than 100'000 vertices, it is recommended to reduce the number of vertices.")
```

omichel (changes requested):
Overall, it looks very good to me.
I will now test it.

omichel (commented):
There is no example in the distribution using ColladaShape.
It would be nice to either modify an existing proto or world to use ColladaShape or add a new example.
Or maybe it's coming in #4194?

src/webots/nodes/WbAbstractAppearance.cpp
```diff
@@ -37,6 +39,23 @@ WbAbstractAppearance::WbAbstractAppearance(const WbNode &other) : WbBaseNode(oth
   init();
 }
 
+WbAbstractAppearance::WbAbstractAppearance(const QString &modelName, const aiMaterial *material) :
+  WbBaseNode(modelName, material) {
+  aiString name("PBRAppearance");
```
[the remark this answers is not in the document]
ad-daniel: These can't be const since they are passed to assimp which is supposed to replace the contents (only if it finds it)

ad-daniel:
Basically there's 2 issues going on:
- On linux the cat model fails to open because assimp cannot decode the collada file (it contains empty tags such as `<subject/>` which it doesn't know how to handle). I believe msys installs assimp 5.2.3 whereas linux/mac install 5.0.1. I've upgraded to 5.2.3 on Linux as well and it seems the issue goes away (and save 70MB in the process, the compiled library is significantly smaller)
- The reason it looked like it wasn't working on windows is because the collada file contains `<transparency>0.0</transparency>` tag. In my implementation, to determine the transparency I check first for the presence of `AI_MATKEY_TRANSPARENCYFACTOR`, and if that fails I determine the transparency from the `AI_MATKEY_OPACITY` value, setting the transparency to `1 - opacity` (as far as I can tell, 90% of the times the latter is what is defined). The issue is that for reasons I don't know, that transparency in the collada file ends up being encoded as an opacity, so I ended up enforcing `1.0 - 0.0 = 1.0` as transparency, making the cat invisible. So I don't think there's an issue with the code, I don't see much else we can do other than report the behavior and hope that at some point it gets fixed, but until then some files might have the same behavior

ad-daniel:
The node is now generic and supports both collada and wavefront. A test was added that fails if all of a sudden the mapping changes for either format. It also ensures that a node added using `CadShape` compared to one built manually using `Shape { appearance PBRAppearance {} geometry Mesh {}}` (having the same roughness, basecolor, ... parameters of course) match.
A direct comparison between a `CadShape { url "something.dae" }` and `CadShape { url "same_something.obj" }` was not added on purpose since although similar they hardly ever would match in practice (since for instance obj defines the basecolor from a mat file, whereas collada does it differently).

Concerning the transparency, I've looked in the assimp codebase and it's a bit messy there as well, the reason is that different software programs apply the specifications of the various file formats in a very inconsistent/incomplete way, so assimp attempts to do its best while remaining as close as possible to the official specifications. In particular, I'm not even certain they consider it to be a bug since according to the code, depending on the situation, the transparency is flipped on purpose.
Since querying just the AI_MATKEY_OPACITY has worked well for years (for instance when importing from `3d model import` menu, or more recently with ColladaShapes PROTO), I've reverted my implementation to that approach instead (compared to the approach I used in this PR where I queried several MATKEYs related to transparency, not just AI_MATKEY_OPACITY).

Concerning the issues raised here https://github.com/cyberbotics/webots/issues/4447, the scaling was solved in a PR targetting master and the remaining issue about the inconsistency in the orientation (i.e., exporting as X-forward and on Webots not being X-forward) I will address it on a separate PR since it might entail updating all worlds that rely on the Mesh node.

This PR is currently blocking: https://github.com/cyberbotics/webots/pull/4194 and https://github.com/cyberbotics/urdf2webots/pull/169.

ad-daniel:
PS: there is no rush for the review.

docs/reference/cadshape.md
```diff
@@ -0,0 +1,41 @@
...
+```
+CadShape {
+  MFString url         []
+  SFBool   ccw         TRUE
+  SFBool   castShadows TRUE
+  SFBool   isPickable  TRUE
+}
+```
+
+The [CadShape](#cadshape) node renders a Collada object (.dae) or Wavefront (.obj) imported from an external file.
```
omichel, line 13:
```suggestion
The [CadShape](#cadshape) node renders a Collada (.dae) or Wavefront (.obj) object imported from an external file.
```

docs/reference/cadshape.md
```diff
@@ -0,0 +1,41 @@
...
+Both Collada and Wavefront files can include 3D geometries and appearances, be it materials or references to external textures, therefore the [CadShape](#cadshape) node builds both.
+If the file references external textures, these should be placed relative to the collada/wavefront file itself as indicated (i.e., in the location expressed within the .dae/.obj file), no additional search is performed for these textures.
+
```
omichel, line 16:
```suggestion
```

tests/rendering/worlds/cadshape_node.wbt
```diff
@@ -0,0 +1,344 @@
...
+      far 2
+    }
+    TestSuiteEmitter {
+    }
+  ]
+  controller "cadshape_checker"
+  supervisor TRUE
+}
+TestSuiteSupervisor {
+}
```
omichel, line 344:
```suggestion
}

```

omichel (changes requested):
Since this PR is quite big, it would be nice to have a second review.

BenjaminDeleze (changes requested):
Seems to works well! However, I have some remarks:

- `CadShape` is missing from `nodes-and-api-functions.md`
- While I had no problem to generate an animation/scene, I did not manage to make the streaming work. It seems to be a problem of textures path.
- Is it deliberate to remove the vrml export feature ? If it is the case, I think we can clean a lot of thing as (if I am not mistaken) the only exporter left is x3d.
- `Move viewpoint to object` does not work (if I remember correctly it could be because of the bounding sphere)

src/webots/nodes/WbPbrAppearance.cpp
```diff
@@ -595,3 +685,48 @@ bool WbPbrAppearance::exportNodeHeader(WbVrmlWriter &writer) const {
     return true;
   return WbAbstractAppearance::exportNodeHeader(writer);
 }
+
+void WbPbrAppearance::exportShallowNode(WbVrmlWriter &writer) const {
```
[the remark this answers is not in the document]
ad-daniel: I think it's better yes, because they have been instantiated "artificially" (i.e not from a tokenizer nor copied from an existing node) so most of the member variables of these nodes, for instance at the basenode level, are not defined. Hence although letting the "normal" writer do the export might technically be possible, it entails adding a ton of conditionals just about everywhere to prevent segmentation faults at every corner, which is not really a clean approach and definitely harder to maintain in the long term. Or at least, that was the rationale behind the choice

resources/web/wwi/nodes/WbCadShape.js
```diff
@@ -0,0 +1,75 @@
+import WbBaseNode from './WbBaseNode.js';
+import WbWorld from './WbWorld.js';
+import {getAnId} from './utils/utils.js';
+
+export default class WbCadShape extends WbBaseNode {
```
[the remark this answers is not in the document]
ad-daniel: The reason was that once webotsjs will support tessellation, it will be closer to a WbBaseNode than a WbGroup, but you're right, now it's definitely the opposite

resources/web/wwi/Parser.js
```diff
@@ -550,6 +553,33 @@ export default class Parser {
...
+    const use = this._checkUse(node, parentNode);
+    if (typeof use !== 'undefined')
+      return use;
+
+    const id = this._parseId(node);
+
+    let url = getNodeAttribute(node, 'url', '');
+    if (typeof url !== 'undefined')
+      url = url.split('"').filter(element => element)[0]; // filter removes empty elements
+    const ccw = parseFloat(getNodeAttribute(node, 'ccw', 'true'));
```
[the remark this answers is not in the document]
ad-daniel: You're right, it's how ccw was encoded for indexedfacesets before the cleanup the recent cleanup of webotsjs and I forgot to update it accordingly on this branch, I'll do it now

BenjaminDeleze:
When streaming a world with a `CadShape`, a folder with the textures of the `CadShape` is created. For example, when I stream the test world I got:
`$WEBOTS_HOME/textures/cube.jpg`

Maybe the `CadShape` images are put there instead of being in `localhost:1234/textures` and that is why the steaming viewer is not finding the textures. (Just a hypothesis)

src/webots/nodes/WbSphere.cpp
```diff
@@ -116,101 +116,20 @@ bool WbSphere::areSizeFieldsVisibleAndNotRegenerator() const {
 }
 
 void WbSphere::write(WbVrmlWriter &writer) const {
-  if (writer.isVrml() && mIco->value())
-    writeExport(writer);
-  else
-    WbGeometry::write(writer);
+  WbGeometry::write(writer);  // TOREM
```
omichel, line 119: What means  TOREM?

omichel (dismissed):

ad-daniel:
Apologies, I think there's more that can be cleaned up, I just forgot to put it back to draft

ad-daniel:
I think the PR can already be reviewed, the world update test will take a while but locally it came up clean, so fingers crossed should be fine. Linux test suite failed again due to `check_urdf` command missing, which appears to be happening a lot the last few days, or maybe it's just my impression.

I've cleaned up the obvious referenced to vrml exporter. Given that there's multiple writers still (urdf, webots), I don't think many of the `isX3d` can be removed  yet, or at least, not safely since we don't do much, to my knowledge, in terms of testing the exporters. So I've cleaned up the obvious cases only, with several exporters at play and little in terms of tests, it'd be a bit of a gamble to go further at this moment.

ad-daniel:
PS: I've also fixed an edge case when parsing .dae as when opening some of the collada files from the franka emika robot some of them triggered an assert.

src/webots/vrml/WbMFDouble.cpp
```diff
@@ -15,7 +15,7 @@
 #include "WbMFDouble.hpp"
 #include "WbToken.hpp"
 #include "WbTokenizer.hpp"
-#include "WbVrmlWriter.hpp"
+#include "WbWriter.hpp"
```
omichel, line 18: This one is likely useless:
```suggestion
```

src/webots/vrml/WbMFInt.cpp
```diff
@@ -15,7 +15,7 @@
 #include "WbMFInt.hpp"
 #include "WbToken.hpp"
 #include "WbTokenizer.hpp"
-#include "WbVrmlWriter.hpp"
+#include "WbWriter.hpp"
```
omichel, line 18:
```suggestion
```

src/webots/vrml/WbProtoList.cpp
```diff
@@ -21,7 +21,7 @@
 #include "WbProtoModel.hpp"
 #include "WbStandardPaths.hpp"
 #include "WbTokenizer.hpp"
-#include "WbVrmlWriter.hpp"
+#include "WbWriter.hpp"
```
omichel, line 24:
```suggestion
```

src/webots/vrml/WbProtoModel.cpp
```diff
@@ -27,7 +27,7 @@
 #include "WbToken.hpp"
 #include "WbTokenizer.hpp"
 #include "WbValue.hpp"
-#include "WbVrmlWriter.hpp"
+#include "WbWriter.hpp"
```
omichel, line 30:
```suggestion
```

src/webots/vrml/WbSingleValue.cpp
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 #include "WbSingleValue.hpp"
-#include "WbVrmlWriter.hpp"
+#include "WbWriter.hpp"
```
omichel, line 16:
```suggestion
```

omichel (changes requested):

BenjaminDeleze (approved):
Thank you

omichel (approved):

PR #4290, opened 2022-02-28
Fix update rate of webotsjs

BenjaminDeleze:
Webots was not sending an update to webotsjs at each timetep but according to the number of fps which is not what we want.

I put some `assert` to make sure that `currentTime > mLastUpdateTime` but I'm not sure if it's really necessary because the two variables (`currentTime` and `mLastUpdateTime`) are only used for the assert.

### DIFF
--- a/src/webots/engine/WbAnimationRecorder.cpp
+++ b/src/webots/engine/WbAnimationRecorder.cpp
@@ -29,7 +29,6 @@
 #include <QtCore/QFileInfo>
 #include <QtCore/QMutableListIterator>
 
-#include <cassert>
 // this function is used to round the transform position coordinates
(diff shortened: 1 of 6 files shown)


src/webots/engine/WbAnimationRecorder.cpp
```diff
@@ -29,6 +29,7 @@
 #include <QtCore/QFileInfo>
 #include <QtCore/QMutableListIterator>
 
+#include <cassert>
```
omichel, line 32:
```suggestion
```

omichel (changes requested):
Just a small clean-up.

omichel (approved):
Thank you.

PR #2077, opened 2020-08-05
Cleanup of the streaming server protocol

omichel:
Various cleanups and bug fixes for the streaming server:

- [x] Don't send useless 'world info' message in case of X3D streaming.
- [x] Fixed 'pause' message which was missing the time value.
- [x] Fixed 'robot window' messages.
- [x] Fixed bugs in the parsing the 'robot' line in multiple lines messages.
- [x] Support `x3d` and `mpjeg` modes in simulation server.

PR #4294, opened 2022-03-01
Fix sending of robot window names for x3d

BenjaminDeleze:
Partially revert #2077.
I could not find why it was removed.


omichel (approved):
I don't recall either. If you could not observe any side effect, that's probably fine.

PR #4295, opened 2022-03-02
Remove Ubuntu 18 from develop tests

omichel:
Since the develop branch is not compatible any more with Ubuntu 18.04 due to the upgrade to Qt6, we need to adapt the workflows accordingly.
Now the diff between the `test_suite_linux.yml` and `test_suite_linux_develop.yml` is so important that the workflow synchronization test is not very useful and more painful to maintain, so I disabled it.


BenjaminDeleze (approved):
Thank you

Issue #4313, opened 2022-03-03
Setting `materialIndex` to 0 in `.obj` mesh results in a crash


ad-daniel:
**Describe the Bug**
According to the docs this field only applies to collada files but apparently when used with `obj` it can crash regardless (only if value is set to 0, any other is fine).

**Steps to Reproduce**
1. Add a Shape with a Mesh node
2. Define an url for the mesh (an `.obj` file)
3. set `materialIndex = 0` 
4. Webots crashes

ad-daniel:
Changing the `name` field on a `obj` also results in a crash

PR #4314, opened 2022-03-03
Fix `name` and `materialIndex` causing crash on `.obj` meshes

ad-daniel:
**Description**
Fixes https://github.com/cyberbotics/webots/issues/4313

Although these fields should only apply to collada files, no distinction was made if the mesh happened to be an `obj`, resulting in crashes.

### DIFF
--- a/src/webots/nodes/WbMesh.cpp
+++ b/src/webots/nodes/WbMesh.cpp
@@ -308,8 +308,7 @@ void WbMesh::updateTriangleMesh(bool issueWarnings) {
 }
 
 uint64_t WbMesh::computeHash() const {
-  QString meshPathNameIndex = path();
-  meshPathNameIndex += (mIsCollada ? mName->value() + QString::number(mMaterialIndex->value()) : "");
+  const QString meshPathNameIndex = path() + (mIsCollada ? mName->value() + QString::number(mMaterialIndex->value()) : "");
   return WbTriangleMeshCache::sipHash13x(meshPathNameIndex.toUtf8().constData(), meshPathNameIndex.size());
 }
 
--- a/src/webots/nodes/WbMesh.cpp
+++ b/src/webots/nodes/WbMesh.cpp
@@ -523,7 +523,7 @@ void WbMesh::updateUrl() {
   if (path().isEmpty())
     return;
 
-  mIsCollada = path().mid(path().lastIndexOf('.') + 1).toLower() == "dae" ? true : false;
+  mIsCollada = (path().mid(path().lastIndexOf('.') + 1).toLower() == "dae");
 
   // we want to replace the windows backslash path separators (if any) with cross-platform forward slashes
   const int n = mUrl->size();
--- a/src/webots/nodes/WbMesh.cpp
+++ b/src/webots/nodes/WbMesh.cpp
@@ -35,6 +35,7 @@ void WbMesh::init() {
   mUrl = findMFString("url");
   mName = findSFString("name");
   mMaterialIndex = findSFInt("materialIndex");
+  mIsCollada = false;
   mResizeConstraint = WbWrenAbstractResizeManipulator::UNIFORM;
   mDownloader = NULL;
 }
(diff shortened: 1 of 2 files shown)


src/webots/nodes/WbMesh.cpp
```diff
@@ -304,8 +308,9 @@ void WbMesh::updateTriangleMesh(bool issueWarnings) {
 }
 
 uint64_t WbMesh::computeHash() const {
-  const QByteArray meshPathNameIndex = (path() + mName->value() + mMaterialIndex->value()).toUtf8();
-  return WbTriangleMeshCache::sipHash13x(meshPathNameIndex.constData(), meshPathNameIndex.size());
+  QString meshPathNameIndex = path();
+  meshPathNameIndex += (mIsCollada ? mName->value() + QString::number(mMaterialIndex->value()) : "");
+  return WbTriangleMeshCache::sipHash13x(meshPathNameIndex.toUtf8().constData(), meshPathNameIndex.size());
```
BenjaminHug8, line 313: Can we do this instead?

```suggestion
  const QByteArray meshHash = (path() + (mIsCollada ? mName->value() + mMaterialIndex->value()) : "").toUtf8();
  return WbTriangleMeshCache::sipHash13x(meshHash.constData(), meshHash.size());
```

src/webots/nodes/WbMesh.cpp
```diff
@@ -97,6 +98,9 @@ void WbMesh::createResizeManipulator() {
 }
 
 bool WbMesh::checkIfNameExists(const aiScene *scene, const QString &name) const {
+  if (!mIsCollada)
+    return false;
+
```
BenjaminHug8, line 103: Useless code, here is why:

`checkIfNameExists()` is called only if `mIsCollada` is `true` on line 161, so why check if it is false?

```suggestion
```

src/webots/nodes/WbMesh.cpp
```diff
@@ -518,6 +523,8 @@ void WbMesh::updateUrl() {
   if (path().isEmpty())
     return;
 
+  mIsCollada = path().mid(path().lastIndexOf('.') + 1).toLower() == "dae" ? true : false;
```
omichel, line 526: Simpler:
```suggestion
  mIsCollada = (path().mid(path().lastIndexOf('.') + 1).toLower() == "dae");
```

src/webots/nodes/WbMesh.cpp
```diff
@@ -304,8 +308,9 @@ void WbMesh::updateTriangleMesh(bool issueWarnings) {
 }
 
 uint64_t WbMesh::computeHash() const {
-  const QByteArray meshPathNameIndex = (path() + mName->value() + mMaterialIndex->value()).toUtf8();
-  return WbTriangleMeshCache::sipHash13x(meshPathNameIndex.constData(), meshPathNameIndex.size());
+  QString meshPathNameIndex = path();
+  meshPathNameIndex += (mIsCollada ? mName->value() + QString::number(mMaterialIndex->value()) : "");
```
omichel, line 312: Simpler:
```suggestion
  const QString meshPathNameIndex = path() + (mIsCollada ? mName->value() + QString::number(mMaterialIndex->value()) : "");
```

BenjaminHug8 (changes requested):
Thank you, overall it is a good check but I have minor suggestions.

EDIT:
@omichel has been faster than me, so maybe some of my suggestions are not anymore valid.

src/webots/nodes/WbMesh.cpp
BenjaminHug8: Another suggestion above is better that mine.

src/webots/nodes/WbMesh.cpp
ad-daniel: Good point, thank you!

BenjaminHug8 (approved):
All good for me, thanks!

ad-daniel:
they passed yes, I've just added a label and the checks restarted

omichel (approved):

PR #4317, opened 2022-03-04
Docker in docker implementation

joachimhgg:
## Docker-in-docker implementation

**State of the project**

Testing project location: `https://github.com/cyberbotics/syllabus/blob/main/test/sensors`

when putting the tag docker to false in config, the sensor's controller is able to start in its docker.
command example return by `WbController.cpp`: 
`docker run --network none -v /tmp/webots_13_837357824:/tmp/webots_13_837357824 -e WEBOTS_SERVER=/tmp/webots_13_837357824 -e WEBOTS_ROBOT_ID=496 sha256:faabb422f74a54602577f356eaf70736dcc53dc0824fa84260c5dc8c218d4aab`

To run a docker in docker we need:
- Docker installation on the docker via [this Dockerfile](https://github.com/cyberbotics/syllabus/blob/main/test/sensors/Dockerfile) and a bash script wrapDocker (does not seem needed?). The installation of docker is [obsolete](https://github.com/jpetazzo/dind) so should be taken carefully. Even so, this solution install docker successfully. Another solution, maybe cleaner, would have been to build the webots image from the [dind image](https://hub.docker.com/_/docker/). We should add this image of Webots+docker to dockerHub.

- As soon as the docker image of the project is built, you can get the image id (`docker ps`), go inside the docker with `docker exec -it 140675378930304_webots_1 /bin/bash` and run (inside the docker), for example `docker version` to confirm the installation of docker in the docker. Then, when starting the sensor simulation, you can run `docker images` to see that docker built correctly the controller's docker.

- As mention in this [useful reading](https://itnext.io/docker-in-docker-521958d34efd), to access to the images of the project's docker, you need to share the same daemon host by adding the volume `/var/run/docker.sock:/var/run/docker.sock`. To do so, I introduced a new docker-compose.yml called `docker-compose-competition.yml` that you can find in this PR. To select it, you need to add the keyword `DockerCompose:competition` in the `webots.yml`.

**Result**

The docker image of the controller is built, but when docker run is failing due to:
`socket connect() failed for /tmp/webots_13_837357824, errno=111`
Maybe because the `/tmp/webots_13_837357824` is root in the controller's docker, because of wrong docker run arguments, or maybe because of open port issues.
For the latter, I passed the Webots's port to the controller's docker, but it did not work.


omichel:
I believe docker-in-docker is not a good idea as it pose a [number of problems](https://stackoverflow.com/a/33003273/810268) which we want to avoid:
1. It does not cooperate well with Linux Security Modules (LSM).
2. It creates a mismatch in file systems that creates problems for the containers created inside parent containers.

Instead, I believe it would be better to modify Webots running in its own docker to ask the simulation server to start controllers as extern controllers in sibling dockers. Basically, the simulation server would act as a proxy for Webots to launch controllers in dockers.

omichel:
I believe docker-in-docker is not a good idea as it pose a [number of problems](https://stackoverflow.com/a/33003273/810268) which we want to avoid:
1. It does not cooperate well with Linux Security Modules (LSM).
2. It creates a mismatch in file systems that creates problems for the containers created inside parent containers.

Instead, I believe it would be better to modify Webots running in its own docker to ask the simulation server to start controllers as extern controllers in sibling dockers. Basically, the simulation server would act as a proxy for Webots to launch controllers in dockers.

Issue #4312, opened 2022-03-03
Setting invalid mesh (using a remote url) crashes Webots


ad-daniel:
**Describe the Bug**
Setting an invalid remote mesh (ex: `http://asd`) in **debug mode** results in [this](https://github.com/cyberbotics/webots/blob/4acda886887a9a436b857d01f4b0c405adddfdcd/src/webots/nodes/utils/WbDownloader.cpp#L103) assert being triggered. In release-mode it doesn't crash, but still if the flag triggers it means a regression was introduced.

**Steps to Reproduce**
1. Add a Shape with a Mesh geometry
2. Set `http://asd` as url of the mesh
3. Webots crashes

I think it's related to something I've been dealing with lately, which is that whenever a change is done to a url field, the corresponding slot (`updateUrl`) is called twice instead of just once. I think the second call is what triggers the assert but I'm not sure yet.

ad-daniel:
It's not just the `url` field, `name`, `ccw` and `materialIndex` also trigger twice (both on master and develop)

ad-daniel:
The multiple firings is due to the fact that in my simulation the mesh was also used as bounding object, hence when any of these fields triggers for the DEF node, it also triggers for the USE ones. Removing the bounding object didn't solve the primary issue of the assert being triggered, so the multiple firings wasn't the cause.

ad-daniel:
The issue stems [here](https://github.com/cyberbotics/webots/blob/54593fe6e37c46973d060bf08d76b9dd248bba9e/src/webots/nodes/utils/WbDownloader.cpp#L114-L122).
Basically what happens is:
- when the url changes the download starts and finishes immediately
- since the url is invalid the condition above kicks in, which triggers a new download (which since `mOffline` is set to true will download from the cache instead from the net)
- the second time the download fails again because the file isn't available on the cache either, but we don't enter anymore in this condition anymore since the flag `mOffline` is set to true after the first failure
- since no early abort occurs this time, the rest of the `finished` function is executed hence triggering the assert since `mNetworkReply` is no longer NULL after the first attempt.

On one side we don't want to endlessly attempt to download, but on the other side it seems unavoidable that this assert will trigger for an invalid url, which begs the question if we should keep it. The presence of the assert is justified, and if we just ignore it we introduce a memory leak (since we overwrite `mNetworkReply` right after the assert). According to the docs:

[...]

PR #4318, opened 2022-03-04
Fix download of invalid url

ad-daniel:
**Description**
Fixes https://github.com/cyberbotics/webots/issues/4312 (more details about the issue available here).

As far as I can see, this is the cleanest approach (I could think of) to keep the assert while preventing a memory leak. Basically we trigger the second download only after the reply of the first has been properly destroyed. I'm open to better solutions though, if there's better ways.


omichel (approved):
That looks good to me.
Thank you.

PR #4334, opened 2022-03-11
Enhancement toolbar min size

BenjaminDeleze:
Add a basic mechanism to ensure that the toolbar looks fine on resize

### DIFF
--- a/resources/web/wwi/Toolbar.js
+++ b/resources/web/wwi/Toolbar.js
@@ -10,7 +10,7 @@ export default class Toolbar {
     this._view = view;
     this.type = type;
     this.parentNode = parentNode;
-    this.size = 0;
+    this.width = 0;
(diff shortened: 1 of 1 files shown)


resources/web/wwi/Toolbar.js
```diff
@@ -10,13 +10,17 @@ export default class Toolbar {
     this._view = view;
     this.type = type;
     this.parentNode = parentNode;
+    this.size = 0;
```
omichel, line 13: Using `width` instead of `size` would be more precise, isn't it?
```suggestion
    this.width= 0;
```

omichel: Or even better `minWidth`?

omichel (approved):
Thank you.

PR #4339, opened 2022-03-14
Async close of dockers

BenjaminDeleze:
Close the docker in a asynchronous way in order to not block the new connections.

as this branch is derived from the `docker-in-docker` branch, it also introduce some changes made in the `docker-in-docker`branch:

- fix a typo
- Refactor the way we use the webots.yml to allow to select a docker-compose (for now we only have the default docker-compose and the theia one, but for example with the competition it is likely to have a new one)

I think it is a good idea to merge this changes now as it is likely that we will never merge `docker-in-docker`


omichel (approved):
Thank you.

omichel (approved):

PR #4340, opened 2022-03-14
Remove unnecessary loop in ColladaShape

ad-daniel:
**Description**
As far as I can tell, this loop was used only for debugging purposes. Currently therefore the code extracts from the material property the same parameters multiple times for no apparent reason


src/webots/nodes/utils/WbNodeOperations.cpp
```diff
@@ -355,30 +355,26 @@ static void addModelNode(QString &stream, const aiNode *node, const aiScene *sce
...
-        // Uncomment this part to print all the properties of this material
-        // qDebug() << propertyName << property->mData << property->mSemantic << property->mIndex << property->mDataLength
-        //          << property->mType;
-      }
+      float values[3];
+      float value;
+      unsigned int count = 3;
+      if (aiGetMaterialFloatArray(material, AI_MATKEY_COLOR_DIFFUSE, values, &count) == AI_SUCCESS && count == 3)
+        baseColor = WbVector3(values[0], values[1], values[2]);
+      count = 3;
```
omichel, line 363: That's odd: why do we need to set this value to 3 twice?

PR #4344, opened 2022-03-15
Launch controllers in sibling dockers

omichel:
This PR aims at changing Webots so that when it runs a world containing extern controllers, the simulation server will launch these extern controllers, possibly in a docker. The stdout/stderr of the robot controllers is sent to Webots instead of landing in the simulation server console, so that it can be displayed online.

It seems that using memory mapped files instead of shared memory would make it easier to share volumes across dockers. Also, the performance of memory mapped files is similar to the one of shared memory (or even better) according to https://github.com/goldsborough/ipc-bench. I used https://github.com/omichel/mmap to test the performance on a native Linux versus running the client and the server in two different dockers.

I implemented and removed a TCP interface between the libController and Webots. This was not useful for the purpose of this PR (I found a simpler solution). Nevertheless it might be useful when we want to support controllers running on remote machines. Therefore I kept the code for this TCP interface (libController side) here: https://github.com/omichel/tcp_client

- [x] changed constant values for WB_STDOUT and WB_STDERR to be compliant with POSIX values for stdout/stderr.
- [x] redirect stdout/stderr to the Webots console for `<extern>` controllers if the WEBOTS_STDOUT_REDIRECT and WEBOTS_STDERR_REDIRECT environment variables are set.
- [x] Webots prints messages on stdout to ask the simulation server to start-up controllers.
- [x] Support for multiple instances of Webots running on the same host machine in different dockers.
- [x] Start-up of robot controllers in sibling dockers.
- [x] Support for pipe across dockers (to make simple robot controllers work, e.g., e-puck default controller).
- [x] Isolate pipes for each controller.
- [x] Replace shared memory segments with memory mapped files (to allow isolation for each controller).
- [x] Removed the obsolete SSL and edit controller features from the streaming servers.
- [x] Split the monitor option from the --stream option.
- [x] Removed the ssh connection and edit controller functionality from the streaming server.
- [x] Ensure that disconnected `<extern>` controllers can re-connect.
- [x] Ensure a controller can be changed in the scene tree from regular to `<extern>`.
- [x] Ensure a controller can be changed in the scene tree from `<extern>` to regular.
- [x] Ensure a controller can be changed in the scene tree from `<extern>` to `<none>`.
- [x] Ensure a controller can be changed in the scene tree from `<none>` to `<extern>`.
- [x] Support for memory mapped files across dockers (to support camera, lidar and range-finder devices).
- [x] Test on Linux.
- [x] Test on Windows.
- [x] Test on macOS.
- [x] Code clean-up.
- [x] Update [documentation for extern controllers](https://cyberbotics.com/doc/guide/running-extern-robot-controllers?version=feature-launch-controllers-in-sibling-dockers).
- [x] Update change log.

Follow-up issues:
- clean-up libController: rename `g_pipe` to `pipe` and find better names for `image` and `g_image`
- split the WbTcpServer class into a WbTcpServer and WbStreamingServer classes.

### DIFF
--- a/src/controller/c/robot.c
+++ b/src/controller/c/robot.c
@@ -1184,11 +1184,9 @@ static char *compute_socket_filename() {
       filenames[count++] = strdup(de->d_name);
     }
     closedir(dr);
-    if (count == 0) {
-      fprintf(stderr, "Webots instance %d has no extern controller robots.\n", number);
-      exit(EXIT_FAILURE);
-    }
-    if (count > 1) {  // more than one extern controller in the current instance of Webots
+    if (count == 0)
+      socket_filename = NULL;
+    else if (count > 1) {  // more than one extern controller in the current instance of Webots
       fprintf(stderr, "Webots instance %d has several extern controller robots.\n", number);
       fprintf(stderr, "Please set the WEBOTS_CONTROLLER_URL environment variable to select one among:\n");
       for (int i = 0; i < count; i++)


src/webots/gui/WbMultimediaStreamingServer.cpp
```diff
@@ -31,11 +31,14 @@
 #include <QtGui/QMouseEvent>
 #include <QtWebSockets/QWebSocket>
 
+#include <iostream>
+
+using namespace std;
+
```
BenjaminDeleze, line 37: Is it really needed?

src/webots/gui/WbX3dStreamingServer.cpp
```diff
@@ -29,9 +29,11 @@
 #include <QtCore/QFileInfo>
 #include <QtWebSockets/QWebSocket>
 
-WbX3dStreamingServer::WbX3dStreamingServer(bool monitorActivity, bool disableTextStreams, bool ssl, bool controllerEdit) :
-  WbStreamingServer(monitorActivity, disableTextStreams, ssl, controllerEdit, true),
-  mX3dWorldGenerationTime(-1.0) {
+#include <iostream>
+
+using namespace std;
+
```
BenjaminDeleze, line 35:
```suggestion
```

Can be removed if `cerr` is revert to `WbLog`

src/webots/gui/WbX3dStreamingServer.cpp
```diff
@@ -47,11 +49,11 @@ void WbX3dStreamingServer::start(int port) {
     try {
       WbAnimationRecorder::instance()->initFromStreamingServer();
     } catch (const QString &e) {
-      WbLog::error(tr("Error when initializing the animation recorder: %1").arg(e));
+      cerr << tr("Error when initializing the animation recorder: %1").arg(e).toUtf8().constData() << endl;
```
BenjaminDeleze, line 52: It is maybe better to keep WbLog

src/webots/launcher/webots-linux.sh
```diff
@@ -99,8 +91,4 @@ wait ${webots_pid}
 
 webots_return_code=$?
 
-# clean-up tmp folder and pipe files in case webots crashed without clean-up
-rm -rf $WEBOTS_TMP_PATH
-rm -f ${TMPDIR}/webots_${webots_pid}_*
```
BenjaminDeleze, line 104: Why is it not needed anymore?

BenjaminDeleze (changes requested):
I finished reading through the code. Here are some minors comments.

The `docker` folder is missing in the Makefile of `projects/samples/camera`.

I think it is not very important but when launching the `launcher_without_docker.py` script, if `camera` is not found, the script exits but Webots not.

In the `README.md` of `howto/docker` it could be good to add a section describing what should be installed (docker-compose with pip for example) to run the scripts.

I will begin testing now.

BenjaminDeleze (commented):
I have found some minors bugs that I am not sure if they are worth fixing:
Lets take a simulation with two NAOs called NAO and NAO1 that have extern protos. I launch one instance of Webots.
- Bug 1:
  - Lets start both controllers without the `WEBOTS_CONTROLLER_URL`, both controllers will connect on the same NAO (correct behavior)
  - Then disconnect the second one and define the `WEBOTS_CONTROLLER_URL` to NAO1.
  - Relaunch the second controller. It will connect to NAO1.
  - Problem: the simulation will still not start (`synchronization` is true for both NAO) but the two controllers are connected and running.
  - It seems that when connecting two controllers to the same robot, the first one is replaced by the second, but the script continue to run like normally in the terminal (or IDE,..)
- Bug 2:
  - Lets rename NAO1 to NAO2, save the simulation and reload.
  - Try to connect both controllers.
  - The first one will connect right but not the second one (it tries to connect to NAO1).
  - Problem: if I change NAO2 to NAO1 in the scene tree it still does not work.
  - I had to change the controllers as follow for the update to be made: `<extern>` -> `<none>` (or any other controllers) -> `<extern>`
- Bug 3:
  - We only need one NAO for this one. Ensure, synchronization is set to `true`.
  - Connect the extern controller.
  - Disconnect the extern controller.
  - Change the synchronization to `false`.
  - The simulation will not resume.
  - Note that the opposite will work: begin with `false` and change it to `true` will pause the simulation.
- Bug 4:
  - Launch two instances of Webots. The first will get port 1234 and the second will output `Error when creating the TCP streaming server on port 1234: Cannot set the server in listen mode: The bound address is already in use, trying again with port 1235
` - However even so it writes that it will try with 1235, it will still listen to 1234 (`INFO: "NAO" extern controller: waiting for connection on ipc://1234/NAO`)

src/webots/gui/WbMultimediaStreamingServer.cpp
omichel: No indeed, that's a remnant of debugging...

projects/samples/howto/docker/launcher.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+if sys.platform != 'linux':
+    sys.exit('This script runs only on Linux, not on ' + sys.platform)
+
+controllers = {
+  "MyBot": "controllers/camera/camera"
+}
+port = 1234
+docker_image = 'cyberbotics/webots:R2022b-1'
```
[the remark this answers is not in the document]
omichel: I set it to R2022b even though it doesn't exist yet, but I added a comment and that is ready for the release already.

src/webots/launcher/webots-linux.sh
omichel: Because the temporary folder is not any more based on the PID and the script doesn't know which port number Webots uses. Moreover this clean-up was not strictly needed since Webots does the clean-up pretty well on quit and on start-up.

src/webots/control/WbControlledWorld.cpp
```diff
@@ -443,74 +328,96 @@ void WbControlledWorld::updateCurrentRobotController() {
...
+    mNewControllers.append(controller);
   connect(controller, &WbController::hasTerminatedByItself, this, &WbControlledWorld::deleteController);
+  assert(controllerInOnlyOneList(controller));
 }
 
 void WbControlledWorld::handleRobotRemoval(WbBaseNode *node) {
-  WbRobot *robot = static_cast<WbRobot *>(node);
-  assert(robot);
-  mRobotsWaitingExternController.removeAll(robot);
+}
```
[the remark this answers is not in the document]
omichel: No, I will clean this up...

src/webots/gui/WbX3dStreamingServer.cpp
ygoumaz: Is there a reason to use std::cerr instead of WbLog ?

ygoumaz (changes requested):
Thank you! I just have a few minor suggestions for the code and documentation. I still need to test the behavior in Webots.

src/webots/control/WbControlledWorld.cpp
```diff
@@ -25,90 +25,48 @@
...
-  qputenv("WEBOTS_SERVER", mServer->fullServerName().toUtf8());
-  QFile file(WbStandardPaths::webotsTmpPath() + "WEBOTS_SERVER");
-  if (file.open(QIODevice::WriteOnly)) {
-    QTextStream stream(&file);
-    stream << mServer->fullServerName().toUtf8() << '\n';
-    file.close();
-  }
   foreach (WbRobot *const robot, robots()) {
     connect(robot, &WbRobot::startControllerRequest, this, &WbControlledWorld::startController);
     connect(robot, &WbRobot::isBeingDestroyed, this, &WbControlledWorld::handleRobotRemoval);
```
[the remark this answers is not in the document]
omichel: No, I just removed it.

src/webots/control/WbControlledWorld.cpp
```diff
@@ -25,90 +25,48 @@
...
+    delete mExternControllers.takeFirst();
 }
 
 void WbControlledWorld::setUpControllerForNewRobot(WbRobot *robot) {
-  if (!robot)
-    return;
+  assert(robot);
 
   connect(robot, &WbRobot::startControllerRequest, this, &WbControlledWorld::startController);
   connect(robot, &WbRobot::isBeingDestroyed, this, &WbControlledWorld::handleRobotRemoval);
```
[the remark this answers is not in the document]
omichel: No, this was just fixed.

docs/guide/running-extern-robot-controllers.md
```diff
@@ -163,19 +163,25 @@ In this case, you simply need to set the `controller` field of this robot to `<e
...
-If that simulation has more than one extern controller, you may also set the `WEBOTS_ROBOT_NAME` environment variable to specify the robot to which your controller should connect.
+### Multiple Concurrent Simulations and Multiple Extern Robot Controllers
 
-> **Note**: the environment variables can be set inside the controller program, before calling the `wb_robot_init()` function.
+If you are running multiple simulations simultaneously on the same machine, and each simulation has several robots that you want to control from extern controllers, then you need to indicate to each controller to which instance of Webots and to which robot it should try to connect.
+This can be achieved by setting the `WEBOTS_CONTROLLER_URL` environment variable to the following value: `ipc://<port>/<robot_name>` where `<port>` is the TCP port (defined in the `--port` command line option) of the target Webots instance and `<robot_name>` is the name of the robot to which you want to connect your controller.
+
+### Notes about the WEBOTS\_CONTROLLER\_URL Environment Variable
```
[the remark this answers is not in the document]
omichel: I fixed this behavior which seems odd, so we don't need to document it that way.

projects/samples/howto/docker/controllers/camera/Makefile
```diff
@@ -0,0 +1,25 @@
+# Copyright 1996-2022 Cyberbotics Ltd.
```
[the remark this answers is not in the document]
omichel: Fixed.

src/controller/c/system.c
```diff
@@ -84,7 +84,7 @@ const char *wbu_system_short_path(const char *path) {
 #endif
 }
 
-static const char *wbu_system_tmpdir() {
+const char *wbu_system_tmpdir() {
```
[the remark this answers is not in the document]
omichel: I turned `wbu_system_webots_tmp_path` into `wbu_system_webots_instance_path` and changed the environment variable accordingly. See 6c16045379ef7bee5b9f2dddf6e052c5ffcc8519.

projects/samples/howto/docker/simulation/docker-compose-webots.yml
```diff
@@ -0,0 +1,25 @@
...
+  webots:
+    image: simulation
+    build:
+      context: .
+      dockerfile: Dockerfile
+      args:
+        - WEBOTS_DEFAULT_IMAGE=$IMAGE
+    command: /usr/local/webots/webots --port=$PORT --mode=realtime --heartbeat=5000 --extern-urls --batch $WORLD
+    environment:
+      - DISPLAY=:0
```
BenjaminDeleze, line 17:
```suggestion
      - DISPLAY
```
I had to make this change to make it work on my computer as I have `DISPLAY=1` by default.
It would be good if someone with `DISPLAY=0` could test the fix to see if it still works for them.

BenjaminDeleze (changes requested):

projects/samples/howto/docker/simulation/docker-compose-webots.yml
omichel: I just tested and I confirm this works. My `DISPLAY` environment variable is set to `:0`.

omichel:
Thank you for the in depth review. I believe I fixed all the issues you reported, except for Bug 2 and Bug 3 reported by @BenjaminDeleze because I believe they are minor and fixing them would make the code way more complex. Also ideally we should handle the case where the name of a robot is changed while it is already connected, which is even trickier to address. In both cases, it is fairly easy to understand for the user that changing the name/synchronization of a robot in a running simulation requires a revert to handle the case properly on the controller side.

src/controller/c/robot.c
```diff
@@ -1170,22 +1166,22 @@ static char *compute_socket_filename() {
       free(folder);
       return NULL;
     }
-    if (count > 1)  // sort folders (robot names) in alphabetical order
-      qsort(filenames, count, sizeof(char *), strcmp_sort);
-    for (int i = 0; i < count; i++) {  // keep only the first extern controller
+    if (count > 1) {  // more than one extern controller in the current instance of Webots
+      fprintf(stderr, "Webots instance %d has several extern controller robots.\n", number);
+      fprintf(stderr, "Please set the WEBOTS_CONTROLLER_URL environment variable to select one among:\n");
```
ygoumaz, line 1171: I have spotted a bug when displaying the possible robots to which the user can connect. If we open multiple worlds with different robot names successively, folder are created but never deleted. Therefore, the list contains a lot of names which are not available in the current world. 

Also the current behavior doesn't check if the robot is waiting for an extern controller, therefore all robots are considered as "waiting for extern controller". It makes it impossible to work on worlds with multiple robots like `/projects/samples/devices/worlds/gps` for example.

ygoumaz (changes requested):

src/controller/c/robot.c
omichel: Well spotted! I believe I fixed this problem.

BenjaminDeleze (changes requested):
I discovered some additional/new bugs:

Bug 5: 
I have a crash when:
1. Open `spot.wbt`
2. Play
3. Reset
4. Crash with: `WARNING: Spot (PROTO) > DEF HEAD_SHAPES Transform > Camera: Cannot allocate memory mapped file for camera image.`

Note: in the case of a crash, the `/tmp/webots-PORT` folder is note removed.

Bug 6:
I don't know if it is a bug or not but when there is only one controller: `WEBOTS_CONTROLLER_URLS=NAO` and `WEBOTS_CONTROLLER_URLS=ipc://1234/NAO` is working but `WEBOTS_CONTROLLER_URLS=ipc://NAO` is not.

Bug 7:
When a robot is renamed or removed, its folder is not removed (it seems like the files in the folder are deleted but not the folder itself.

Bug 8:
1. Open Webots with two robots with extern controllers (NAO and NAO1).
2. Open a terminal and go to the the external controller (`nao_demo`) for example.
3. Define `WEBOTS_CONTROLLER_URL` to `NAO`
4. Launch the controller
5. Close the controller
6. Define `WEBOTS_CONTROLLER_URL` to `NAO1`
7. Launch the controller
8. Webots crashes

omichel:
Thanks for the review. Bugs 5, 7 and 8 should be fixed. Bug 6 is not a bug, but I added an error message and improved the documentation to make it clear.

BenjaminDeleze (approved):
Thank you :tada:

ygoumaz (approved):
Thank you!

omichel:
ready for git merge

PR #4348, opened 2022-03-16
Removed reference to deprecated example

omichel:
A user reported that the `ros_python.py` link was broken.
See https://cyberbotics.com/doc/guide/using-ros?version=fix-reference-to-deprecated-example


omichel:
OK, then they are probably not deprecated... We just need to fix the link then...

omichel:
I just fixed the link.

ad-daniel (approved):
Thank you

Issue #4341, opened 2022-03-14
Overlay layout in 3d view is inconsistent between reloads


ad-daniel:
**Describe the Bug**

In theory layout information is loaded from the .wbproj, so between reloads there should not be a difference.
In practice it appears that sometimes this is not the case for cameras overlays, where at times they're small and other times they're full screen in the 3D view. It doesn't happen every time so it's unclear what triggers the change of behavior but it appears to be a regression compared to the previous release where I don't recall it happening.

PR #4350, opened 2022-03-16
Fix perspective save and camera overlays

ad-daniel:
**Description**
Fixes https://github.com/cyberbotics/webots/issues/4341 and an issue with overlays where what is shown in the menu doesn't correspond to what's selected.

- [x] `WEBOTS_DISABLE_SAVE_SCREEN_PERSPECTIVE_ON_CLOSE` behavior

If the variable is not defined or is false:
  - any change to the layout/overlays persists across reloads and restarts, except where expressly undesired (while making movies) 

If the variable is true:
- any change to the layout/overlays is ignored
- if the user saves, the disabler is ignored and we do indeed save the perspective.

- [x] camera overlays (`Overlays > Camera devices` not representative of the overlays being active)

It is unclear to me why in the `selectedRobot` method we arbitrarily take the first robot as selection by default (if a node isn't selected). This has the following side-effects:
- is counter-intuitive, because the robot picked isn't actually selected as the name entails (i.e by the user) but chosen instead by default
- if there's multiple robots, none of them is selected, only if 1 exists in the world.
- despite this choice being made on paper, in practice when the world is loaded the robot is neither selected in the scene tree nor is it highlighted in the 3D view. So it's selected under the hood, but there's no visible proof of it.
- It breaks the `Overlays` menu, because since a robot is selected under the hood we create the `Camera devices` sub-menu that shows that all the cameras are being visualized when in fact they are not. This menu should be populated only if we actually select a robot and the only fields that should be "checked" are those that are actually visible.

It seems reasonable to assume that choice was made for a reason, so removing it might have side-effects of its own but I haven't found any so far.

### DIFF
--- a/docs/guide/the-user-interface.md
+++ b/docs/guide/the-user-interface.md
@@ -207,7 +207,7 @@ If the light sensor device is disabled or the first measurement is not available
   - The **Show Pen Painting Rays** allows you to display, or to hide, the rays in which the pen devices paint.
 These rays are drawn as violet lines if painting is enabled, otherwise as gray lines.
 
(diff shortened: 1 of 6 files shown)


omichel (changes requested):
We should also fix the documentation accordingly: https://cyberbotics.com/doc/guide/the-user-interface

ad-daniel:
Both the issue with the overlay expanding when restricted on both sides and the overlay growing in size outside the visible 3d view should be addressed now.

omichel (approved):
It works perfectly now.
Thank you.

PR #4352, opened 2022-03-17
Feature publish asset cache zip

ad-daniel:
**Description**

Given the recent change to the caching system, a snapshot of the assets can also be provided. 
This is beneficial for offline use, for webots.cloud and dockers in general


ad-daniel:
Should target develop, not master

PR #4353, opened 2022-03-17
Publish asset cache

ad-daniel:
**Description**

Given the recent change to the caching system, a snapshot of the assets can also be provided. 
This is beneficial for offline use, for webots.cloud and dockers in general


BenjaminDeleze (commented):
Seems good to me, but we should somehow provide instructions on how to use it.

BenjaminDeleze (approved):
Thank you

omichel:
I am wondering whether we should name this file to `assets-R2022b.zip` as when we will release it for both the stable and development versions, we will likely need two different asset files.

PR #4361, opened 2022-03-18
Fixed annoying warning

### LINKED ISSUE #4356: Error: wb_supervisor_field_get_count() can only be used with multiple fields (MF).
Webots Supervisor docs say:  
> The wb_supervisor_field_get_count function returns the number of items of a multiple field (MF) passed as an argument to this function. If a single field (SF) or NULL is passed as an argument to this function, it returns -1. Hence, this function can also be used to test if a field is MF (like WB_MF_INT32) or SF (like WB_SF_BOOL).

If you do call `wb_supervisor_field_get_count()` on an SF field, it gives the expected value -1, but also prints a warning: "Error: wb_supervisor_field_get_count() can only be used with multiple fields (MF)."

According to the documentation, returning -1 is the intended useful behavior here, so there should be no need for the warning.  Or, if we aren't supposed to use this on SF fields, then the documentation shouldn't suggest this as legitimate usage.

omichel:
Fixes #4356.
This function should not display a warning for SF fields as it may be used to determine if a field is SF or MF. See https://cyberbotics.com/doc/reference/supervisor#wb_supervisor_field_get_count

### DIFF
--- a/src/controller/c/supervisor.c
+++ b/src/controller/c/supervisor.c
@@ -2835,13 +2835,10 @@ WbFieldType wb_supervisor_field_get_type(WbFieldRef field) {
 
 int wb_supervisor_field_get_count(WbFieldRef field) {
   if (!check_field(field, __FUNCTION__, WB_NO_FIELD, false, NULL, false, false))
-    return false;
-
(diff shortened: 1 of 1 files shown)


ad-daniel (approved):
Thank you.

PR #4364, opened 2022-03-19
Allow multiple extra projects. Add WEBOTS_EXTRA_PROJECT_PATH env variable

BarceloChristian:
**Description**
***Enhancement proposal***
Allows to list and import multiple extra projects defined in the `WEBOTS_EXTRA_PROJECT_PATH` environment variable.
The preferences option `Extra project path` is also enabled with the same functionality. Both can define multiple paths by separating them with `:` (or `;` on Windows)
This will enable the usage of Webots in projects using robots defined in multiple places.

**Additional context**
- Defining each robot in a different place (being packages) is a common pattern when working with ROS, where each robot might have its own package with its definition, the launch files required to spawn the robot, its controller, and so on.
- On the same line, `webots_ros` can be improved to set `WEBOTS_EXTRA_PROJECT_PATH` by using rospack together with the `<export/>` tags in the `package.xml`
- Even more, (and still in `webots_ros`) an spawn (ROS)service to import on runtime the desired robots can be included, easing the migration from other popular simulators where this is a common pattern, allowing the environments to be dynamically generated on runtime (at least in terms of number/position of robots).

### DIFF
--- a/docs/guide/preferences.md
+++ b/docs/guide/preferences.md
@@ -24,10 +24,12 @@ The default value is `python`.
 It should work on most systems assuming that `python` is installed and available from the command line.
 On some systems, it may be useful to set it to `python3.7` for example if you want to launch the controllers with this specific version of Python.
 Bear in mind that this value may be overridden by the content of a `runtime.ini` file of a Python controller that may redefine a specific Python command to launch that controller.
-- The **Extra projects path** defines the paths to user folders similar to the `WEBOTS_HOME/projects` folder.
-These user folders should contain projects resources that can be used in the current project (such as PROTO nodes, controllers, textures, etc.). Multiple user folders can be added by separating them with `:` (or `;` on Windows).
+- The **Extra project path** defines the paths to user folders similar to the `WEBOTS_HOME/projects` folder.
+These user folders should contain projects resources that can be used in the current project (such as PROTO nodes, controllers, textures, etc.).
+Multiple user folders can be added by separating them with `:` (or `;` on Windows).
 They may contain multiple sub-folders, each one associated to one sub-project (which should respect the [Standard File Hierarchy of a Project](the-standard-file-hierarchy-of-a-project.md)).
-Alternatively, the environment variable `WEBOTS_EXTRA_PROJECT_PATH` can be used. If the `Extra projects path` parameter and the `WEBOTS_EXTRA_PROJECT_PATH` environment variable are set, then both will be considered.
+Alternatively, the environment variable `WEBOTS_EXTRA_PROJECT_PATH` can be used.
+If the `Extra project path` parameter and the `WEBOTS_EXTRA_PROJECT_PATH` environment variable are set, then both will be considered.
 These paths have the priority over the other search paths.
 - The **Warnings: Display save warning only for scene tree edit** checkbox prevents Webots from displaying any warning dialog window when you quit, reload or load a new world after the current world was modified by either changing the viewpoint, dragging, rotating, applying a force or torque to an object, or modifying the world from a controller.
 It will however still display a warning if the world was modified from the scene tree.
--- a/docs/guide/preferences.md
+++ b/docs/guide/preferences.md
@@ -24,11 +24,11 @@ The default value is `python`.
 It should work on most systems assuming that `python` is installed and available from the command line.
 On some systems, it may be useful to set it to `python3.7` for example if you want to launch the controllers with this specific version of Python.
 Bear in mind that this value may be overridden by the content of a `runtime.ini` file of a Python controller that may redefine a specific Python command to launch that controller.
-- The **Extra projects path** defines the path to a user folder similar to the `WEBOTS_HOME/projects` folder.
-This user folder should contain projects resources that can be used in the current project (such as PROTO nodes, controllers, textures, etc.).
-It may contain multiple sub-folders, each one associated to one sub-project (which should respect the [Standard File Hierarchy of a Project](the-standard-file-hierarchy-of-a-project.md)).
-It has the priority over the other search paths.
-This folder may also contain a `default` project that can be used to define generic controllers, textures, sounds, etc.
+- The **Extra projects path** defines the paths to user folders similar to the `WEBOTS_HOME/projects` folder.
+These user folders should contain projects resources that can be used in the current project (such as PROTO nodes, controllers, textures, etc.). Multiple user folders can be added by separating them with `:` (or `;` on Windows).
+They may contain multiple sub-folders, each one associated to one sub-project (which should respect the [Standard File Hierarchy of a Project](the-standard-file-hierarchy-of-a-project.md)).
+Alternatively, the environment variable `WEBOTS_EXTRA_PROJECT_PATH` can be used. If the `Extra projects path` parameter and the `WEBOTS_EXTRA_PROJECT_PATH` environment variable are set, then both will be considered.
+These paths have the priority over the other search paths.
 - The **Warnings: Display save warning only for scene tree edit** checkbox prevents Webots from displaying any warning dialog window when you quit, reload or load a new world after the current world was modified by either changing the viewpoint, dragging, rotating, applying a force or torque to an object, or modifying the world from a controller.
 It will however still display a warning if the world was modified from the scene tree.
 - The **Telemetry: Send technical data to Webots developpers** checkbox allows Webots to send anonymous technical data to Webots developpers in order to help improving the software.
--- a/docs/reference/changelog-r2022.md
+++ b/docs/reference/changelog-r2022.md
@@ -10,7 +10,7 @@ Released on XX, XXth, 2022.
     - Added Wizard for the creation of PROTO files ([#4104](https://github.com/cyberbotics/webots/pull/4104)).
     - Added two new Robot API functions, `wb_robot_step_begin` and `wb_robot_step_end`, to optimize computer intensive controllers ([#4107](https://github.com/cyberbotics/webots/pull/4107)).
     - Changed the behavior of `wb_robot_wwi_receive_text` to iterate through the received messages buffer ([#4336](https://github.com/cyberbotics/webots/pull/4336)).
-    - Added the ability to define multiple `Extra Project Paths` through the Preferences menu, and an alternative method by setting the `WEBOTS_EXTRA_PROJECT_PATH` environment variable. ([#4364](https://github.com/cyberbotics/webots/pull/4364)).
+    - Added the ability to define multiple `Extra Project` paths through the Preferences menu, and an alternative method by setting the `WEBOTS_EXTRA_PROJECT_PATH` environment variable. ([#4364](https://github.com/cyberbotics/webots/pull/4364)). **Previously set Extra project paths should be re-set in the preferences menu**
   - New Objects:
     - Added some objects on the hospital theme: hospital bed, drip stand, medicine bottle, hand sanitizer, curtain, photo frame, flower pot, emergency exit sign and a fabric appearance ([#4166](https://github.com/cyberbotics/webots/pull/4166)).
     - Extended the CardboardBox to become a container and added a cardboard appearance ([#4359](https://github.com/cyberbotics/webots/pull/4359)).
--- a/docs/reference/changelog-r2022.md
+++ b/docs/reference/changelog-r2022.md
@@ -10,6 +10,7 @@ Released on XX, XXth, 2022.
     - Added Wizard for the creation of PROTO files ([#4104](https://github.com/cyberbotics/webots/pull/4104)).
     - Added two new Robot API functions, `wb_robot_step_begin` and `wb_robot_step_end`, to optimize computer intensive controllers ([#4107](https://github.com/cyberbotics/webots/pull/4107)).
     - Changed the behavior of `wb_robot_wwi_receive_text` to iterate through the received messages buffer ([#4336](https://github.com/cyberbotics/webots/pull/4336)).
+    - Added the ability to define multiple `Extra Project Paths` through the Preferences menu, and an alternative method by setting the `WEBOTS_EXTRA_PROJECT_PATH` environment variable. ([#4364](https://github.com/cyberbotics/webots/pull/4364)).
   - New Objects:
     - Added some objects on the hospital theme: hospital bed, drip stand, medicine bottle, hand sanitizer, curtain, photo frame, flower pot, emergency exit sign and a fabric appearance ([#4166](https://github.com/cyberbotics/webots/pull/4166)).
     - Extended the CardboardBox to become a container and added a cardboard appearance ([#4359](https://github.com/cyberbotics/webots/pull/4359)).
--- a/src/webots/core/WbProject.cpp
+++ b/src/webots/core/WbProject.cpp
@@ -39,7 +39,7 @@ static QString gPreviousPath = QString();
 static WbProject *gCurrentProject = NULL;
 static WbProject *gSystemProject = NULL;
 static WbProject *gDefaultProject = NULL;
-static WbProject *gExtraDefaultProject = NULL;
+static QList<WbProject *> *gExtraProjects = NULL;
 
 void WbProject::cleanupCurrentProject() {
   delete gCurrentProject;
@@ -47,10 +47,6 @@ void WbProject::cleanupCurrentProject() {
 
 void WbProject::cleanupDefaultProject() {
   delete gDefaultProject;
-}
-
-void WbProject::cleanupExtraDefaultProject() {
-  delete gExtraDefaultProject;
 }
 
 void WbProject::cleanupSystemProject() {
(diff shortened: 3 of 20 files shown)


ad-daniel (changes requested):
Hi, thank you for the contribution!
Overall I agree that it could be a good idea to provide this option, however it seems to me there's a bit of an overlap with the currently provided method of defining additional (extra) proto locations through the preferences window. Therefore I think the two could be somewhat merged together:
- I believe the current method (i.e. using the preferences menu) only accepts a single path, this could be extended by allowing to provide multiple ":" separated paths.
- When checking if extra paths are defined (again through the preferences), at the same time we can also check if the `WEBOTS_PROTO_PATH` (which perhaps should be renamed `WEBOTS_EXTRA_PROTO` for consistency) environmental variable is defined and consider these paths as well. In short, instead of creating ad-hoc code to handle it, we can re-use the same. If both the preferences and the environmental variable are defined, then we just consider both at the same time.

Also as a reminder, unless it's strictly necessary, since this change provides a new feature the PR should target the develop branch instead of master. Whatever the case, I think the change also warrants a mention in the changelog and the documentation should be updated accordingly.
If you encounter any difficulty don't hesitate to let us know.

BarceloChristian:
> Hi, thank you for the contribution! Overall I agree that it could be a good idea to provide this option, however it seems to me there's a bit of an overlap with the currently provided method of defining additional (extra) proto locations through the preferences window. Therefore I think the two could be somewhat merged together:
> 
> * I believe the current method (i.e. using the preferences menu) only accepts a single path, this could be extended by allowing to provide multiple ":" separated paths.
> * When checking if extra paths are defined (again through the preferences), at the same time we can also check if the `WEBOTS_PROTO_PATH` (which perhaps should be renamed `WEBOTS_EXTRA_PROTO` for consistency) environmental variable is defined and consider these paths as well. In short, instead of creating ad-hoc code to handle it, we can re-use the same. If both the preferences and the environmental variable are defined, then we just consider both at the same time.
> 
> Also as a reminder, unless it's strictly necessary, since this change provides a new feature the PR should target the develop branch instead of master. Whatever the case, I think the change also warrants a mention in the changelog and the documentation should be updated accordingly. If you encounter any difficulty don't hesitate to let us know.

Thanks for your review and quick reply, I'm glad to see you agreed on the change.
I addressed the first round of changes to reuse the sections of the preferences menu. Then, if you don't mind, I have a couple of questions:
- You mentioned the preferences menu could receive multiple paths, shall I move forward with that suggestion too? Please take into account that, at least for now (I can create a different ticket for that and work on it later) this change allows you to include ONLY proto files from different paths, you cannot use it to import complete projects as the preferences menu does.
  - If so, in the case of environment variables `:` is widely used, but maybe for the preferences menu we can use `,` or `;`. Let me know what you think about it and I'll move forward with the implementation asap.
- About the Changelog, even though `Webots R2022a Revision 1` was not released yet, the Changelog for R2022b is being already written. Where should I add this PR?
- Besides [this](https://cyberbotics.com/doc/guide/preferences) piece of documentation where I can add the usage of the environment variable as a compliment of the preferences menu, do you think there's any other valuable place to mention it?

ad-daniel:
So basically there's a couple of things to keep in mind.
- As you said, the "Extra projects path" is meant for more than just PROTO, so the locations defined in that field should respect the hierarchy of a project (i.e PROTO file should be in a folder called protos, worlds in worlds, and so forth). The "issue" with your approach of accepting only PROTO is that PROTOs could be, and often are, Robots and therefore come with an associated controller which in your proposition would have to be put elsewhere entirely, so it ends up being a bit messy. So the idea is to have the environmental variable and the preference setting be effectively two methods of achieving the same thing.
This does however enforce the same constraints of following the project hierarchy both on the preference option and the environmental variable (which can be called `WEBOTS_EXTRA_PROJECT_PATH` for extra clarity). Code-wise, we just append the contents of one to the other, remove duplicates if any, and explore each path when looking for PROTOs.
- About the changelog yes you're right, internally we separate the changes made on the master branch (Webots R2022a Revision 1) with those made on the develop branch (Webots R2022b). It's not guaranteed that a revision will be published, and if we don't end up doing one, we'll simply merge the two changelogs together when releasing R2022b. Until then, we keep the distinction. Since your changes target the develop branch, it should be mentioned in the R2022b section)
- Concerning the separator, since Windows uses `;` and Linux/mac use `:` we should probably keep it that way. So based on the OS it should look for these symbols.
- For the documentation yes, it should be sufficient to mention there that the environmental variable is a possible alternative along with some details (the fact that both are considered, the separator, etc)

omichel:
> - Concerning the separator, since Windows uses `;` and Linux/mac use `:` we should probably keep it that way. So based on the OS it should look for these symbols.

We should indeed rely on [QDir::listSeparator()](https://doc.qt.io/qt-6/qdir.html#listSeparator).

BarceloChristian:
@ad-daniel Following your suggestions, now the `extraDefaultProject()` is simply `extraProjects()` and returns a `QList` with all the available extra projects. Then, it's used all along the code to find PROTOs, worlds, controllers, and so on.
Both the `WEBOTS_EXTRA_PROJECT_PATH` environment variable and the `Extra project path` preference allow multiple projects by separating them with the OS default separator (defined by [QDir::listSeparator()](https://doc.qt.io/qt-6/qdir.html#listSeparator) as @omichel suggested). Duplicated paths are discarded.
Changelog and documentation are updated.

docs/reference/changelog-r2022.md
```diff
@@ -10,6 +10,7 @@ Released on XX, XXth, 2022.
     - Added Wizard for the creation of PROTO files ([#4104](https://github.com/cyberbotics/webots/pull/4104)).
     - Added two new Robot API functions, `wb_robot_step_begin` and `wb_robot_step_end`, to optimize computer intensive controllers ([#4107](https://github.com/cyberbotics/webots/pull/4107)).
     - Changed the behavior of `wb_robot_wwi_receive_text` to iterate through the received messages buffer ([#4336](https://github.com/cyberbotics/webots/pull/4336)).
+    - Added the ability to define multiple `Extra Project Paths` through the Preferences menu, and an alternative method by setting the `WEBOTS_EXTRA_PROJECT_PATH` environment variable. ([#4364](https://github.com/cyberbotics/webots/pull/4364)).
```
ad-daniel, line 13:
```suggestion
    - Added the ability to define multiple `Extra Project` paths through the Preferences menu, and an alternative method by setting the `WEBOTS_EXTRA_PROJECT_PATH` environment variable ([#4364](https://github.com/cyberbotics/webots/pull/4364)).
```

src/webots/core/WbProject.cpp
```diff
@@ -75,14 +71,38 @@ WbProject *WbProject::defaultProject() {
...
+  if (gExtraProjects == NULL) {
+    gExtraProjects = new QList<WbProject *>();
+    // collect extra project paths in a QSet to avoid duplicate entries
+    QSet<QString> projectPaths;
+
+    if (!WbPreferences::instance()->value("General/extraProjectsPath").toString().isEmpty()) {
+      foreach (const QString &path, WbPreferences::instance()
+                                      ->value("General/extraProjectsPath")
+                                      .toString()
+                                      .split(QDir::listSeparator(), Qt::SkipEmptyParts))
```
ad-daniel, line 84:
```suggestion
    if (!WbPreferences::instance()->value("General/extraProjectPath").toString().isEmpty()) {
      foreach (const QString &path, WbPreferences::instance()
                                      ->value("General/extraProjectPath")
                                      .toString()
                                      .split(QDir::listSeparator(), Qt::SkipEmptyParts))
```

ad-daniel: I've discussed internally and we decided on keeping it singular, like with the environmental variable `WEBOTS_EXTRA_PROJECT_PATH`, no "s".
So we should probably maintain consistency by keeping it singular everywhere.
(might need to do a wide search in case it's referenced elsewhere other than WbPreferences)

src/webots/scene_tree/WbAddNodeDialog.cpp
```diff
@@ -229,10 +229,14 @@ void WbAddNodeDialog::updateItemInfo() {
         mInfoText->setPlainText(
           tr("This folder lists all suitable node that were defined (using DEF) above the current line of the Scene Tree."));
         break;
-      case PROTO_EXTRA:
-        mInfoText->setPlainText(tr("This folder lists all suitable PROTO nodes from the extra projects located in: '%1'.")
-                                  .arg(WbPreferences::instance()->value("General/extraProjectsPath").toString()));
-        break;
+      case PROTO_EXTRA: {
+        QString title("This folder lists all suitable PROTO nodes from the preferences Extra project paths and "
+                      "the 'WEBOTS_EXTRA_PROJECT_PATH' environment variable.\n");
```
ad-daniel, line 234:
```suggestion
                      "the 'WEBOTS_EXTRA_PROJECT_PATH' environment variable:\n");
```

docs/guide/preferences.md
```diff
@@ -24,11 +24,11 @@ The default value is `python`.
 It should work on most systems assuming that `python` is installed and available from the command line.
 On some systems, it may be useful to set it to `python3.7` for example if you want to launch the controllers with this specific version of Python.
 Bear in mind that this value may be overridden by the content of a `runtime.ini` file of a Python controller that may redefine a specific Python command to launch that controller.
-- The **Extra projects path** defines the path to a user folder similar to the `WEBOTS_HOME/projects` folder.
-This user folder should contain projects resources that can be used in the current project (such as PROTO nodes, controllers, textures, etc.).
-It may contain multiple sub-folders, each one associated to one sub-project (which should respect the [Standard File Hierarchy of a Project](the-standard-file-hierarchy-of-a-project.md)).
-It has the priority over the other search paths.
-This folder may also contain a `default` project that can be used to define generic controllers, textures, sounds, etc.
+- The **Extra projects path** defines the paths to user folders similar to the `WEBOTS_HOME/projects` folder.
```
ad-daniel, line 27:
```suggestion
- The **Extra project path** defines the paths to user folders similar to the `WEBOTS_HOME/projects` folder.
```

ad-daniel: see other comment

src/webots/core/WbProject.cpp
```diff
@@ -39,7 +39,7 @@ static QString gPreviousPath = QString();
 static WbProject *gCurrentProject = NULL;
 static WbProject *gSystemProject = NULL;
 static WbProject *gDefaultProject = NULL;
-static WbProject *gExtraDefaultProject = NULL;
+static QList<WbProject *> *gExtraProjects = NULL;
```
ad-daniel, line 42: this one makes sense to keep it plural since it's a list

ad-daniel (changes requested):
Great work, it works very nicely!
I've left a few comments, mostly about naming/style

src/webots/scene_tree/WbAddNodeDialog.cpp
BarceloChristian: What about leaving the `.` if there's no extra project path defined, and the `:` if there's any? I wouldn't like to leave an open `:` if the list is empty.

src/webots/core/WbProject.cpp
BarceloChristian: It is referenced in WbPreferencesDialog and, correct me if I'm wrong, the translations should be modified too.
This could be a breaking change for somebody migrating from 2022a, is there any extra documentation you want me to add for this?

src/webots/scene_tree/WbAddNodeDialog.cpp
ad-daniel: The "extra project" section is visible in the add node dialog window only if it is non-empty, so there will always be something to show

BarceloChristian: That's a really good point, will address this last comment and upload the changes then

src/webots/core/WbProject.cpp
ad-daniel: There are contributors that sometimes take it upon themselves to do the translations for other languages, but most of them are incomplete, and in practice, rarely used. As for ourselves, we only aim to have the English version up to date as we don't have the resource nor knowledge to ensure all others are properly translated. By doing changes to the english strings, the others are flagged as needing adjustment on its own.

I don't think this change breaks backwards compatibility as it extends functionality without removing anything that was there before, so we're fine on that front (unless I missed something, are you referring to something in particular?)
Mentioning the added feature in the changelog as you did is sufficient.

ad-daniel: Ah I see what you meant for the translations, then yes it needs to be updated as well.

For the backwards compatibility you're right, it does technically require some adjustment by the user so it should be mentioned in bold in the changelog. You can see an example of this [here](https://cyberbotics.com/doc/reference/changelog-r2021). It should be sufficient to notify the user that if they had any extra project paths they should re-set the field in the preferences.

docs/guide/preferences.md
```diff
@@ -24,11 +24,11 @@ The default value is `python`.
...
+These user folders should contain projects resources that can be used in the current project (such as PROTO nodes, controllers, textures, etc.). Multiple user folders can be added by separating them with `:` (or `;` on Windows).
```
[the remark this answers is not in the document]
BarceloChristian: Done

docs/guide/preferences.md
```diff
@@ -24,11 +24,11 @@ The default value is `python`.
...
+They may contain multiple sub-folders, each one associated to one sub-project (which should respect the [Standard File Hierarchy of a Project](the-standard-file-hierarchy-of-a-project.md)).
+Alternatively, the environment variable `WEBOTS_EXTRA_PROJECT_PATH` can be used. If the `Extra projects path` parameter and the `WEBOTS_EXTRA_PROJECT_PATH` environment variable are set, then both will be considered.
```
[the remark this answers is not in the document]
BarceloChristian: Done

src/webots/vrml/WbProtoList.cpp
```diff
@@ -110,8 +111,9 @@ void WbProtoList::updateProjectsProtoCache() {
 void WbProtoList::updateExtraProtoCache() {
   gExtraProtoCache.clear();
   QFileInfoList protosInfo;
-  if (!WbPreferences::instance()->value("General/extraProjectsPath").toString().isEmpty())
-    findProtosRecursively(WbPreferences::instance()->value("General/extraProjectsPath").toString(), protosInfo);
+
+  foreach (const WbProject *project, *WbProject::extraProjects()) { findProtosRecursively(project->path(), protosInfo); }
```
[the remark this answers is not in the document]
BarceloChristian: My bad, second time you mention it.
Done!

docs/reference/changelog-r2022.md
BarceloChristian: Done

src/webots/core/WbProject.cpp
BarceloChristian: Changed naming, translation files, and changelog

ad-daniel (approved):
Looks good to me, thank you for the contribution!
Github actions is currently having issues, so will need to wait for that to be fixed, also @omichel  might want to take a look as well

docs/guide/preferences.md
```diff
@@ -24,11 +24,13 @@ The default value is `python`.
...
-It may contain multiple sub-folders, each one associated to one sub-project (which should respect the [Standard File Hierarchy of a Project](the-standard-file-hierarchy-of-a-project.md)).
-It has the priority over the other search paths.
-This folder may also contain a `default` project that can be used to define generic controllers, textures, sounds, etc.
+- The **Extra project path** defines the paths to user folders similar to the `WEBOTS_HOME/projects` folder.
+These user folders should contain projects resources that can be used in the current project (such as PROTO nodes, controllers, textures, etc.).
+Multiple user folders can be added by separating them with `:` (or `;` on Windows).
+They may contain multiple sub-folders, each one associated to one sub-project (which should respect the [Standard File Hierarchy of a Project](the-standard-file-hierarchy-of-a-project.md)).
+Alternatively, the environment variable `WEBOTS_EXTRA_PROJECT_PATH` can be used.
+If the `Extra project path` parameter and the `WEBOTS_EXTRA_PROJECT_PATH` environment variable are set, then both will be considered.
```
omichel, line 32: I just have one comment here: it would be nice to specify in the documentation which one has the priority. If we have two different PROTO files both named `MyProto.proto` and one of them is located in a path specified in the `WEBOTS_EXTRA_PROJECT_PATH` environment variable while the other is located in a path specified in the `Extra project path` preference settings, which one is going to be used by Webots?

omichel (commented):
It looks very good to me, thank you.
I just have one request regarding the documentation.

docs/guide/preferences.md
BarceloChristian: Oh, I didn't realize this comment before! I can push a hotfix for the documentation if you want.
Basically, it gives priority to whatever is set in the Preferences first, then the environment variables.
If the same PROTO is defined multiple times in the extra paths, only the first occurrence will be available, the rest of them are going to be disabled.

Issue #4363, opened 2022-03-18
`wb_supervisor_node_get_field_by_index` not working


Justin-Fisher:
`wb_supervisor_node_get_field_by_index` sometimes returns an incorrect field.  

**Short version:** at least in some cases, if I use `wb_supervisor_node_get_field_by_index` to get the field with index `a` from the node with id `b`, and then try to use it to get the field with index `b` from the node with id `a`, it wrongly returns that first field again, rather than correctly returning a different field.  So probably you somewhere have variables for a node id and a field index swapped, probably in a case where you're trying to check if a field had already been accessed so that you could return the same field handler again.  Hopefully that will help you to debug it!?  

**Longer version:** (Heavily edited now that I have a better sense of what's happening.)

Here's some bare-bones Python code that consistently reproduces this for me, in a world with the controller's robot having node `id 8`.  While the commented-out line remains commented out, or even if its index is changed from 8 (the `FPS` field) to any other index, then the final line correctly prints that, in the `self` robot, the field with index 1 is a `rotation` field'; whereas if that line is uncommented with index 8, so it accesses the `FPS` field of `WorldInfo` (node id 1) first, then, for some reason, the next-to-last line mistakenly returns that same pointer for the FPS field again, and hence the final line incorrectly reports that the robot's rotation field is named `FPS`.  

This wire-crossing is apparently confusing Node id 8 Field index 1 with Node id 1 Field index 8.  Similarly, another wire-crossing I noticed (between the robot's `children` field and a contact-properties `softCFM` field) is apparently between Node id 8 Field index 3 and Node id 3 Field index 8.  So probably you somewhere have variables for a node id and a field index swapped, probably in a case where you're trying to check if a field had already been accessed.  

```python
    from ctypes import CDLL, c_void_p, c_char_p
    class Ref(c_void_p): pass
    wb = CDLL(r"D:\Program Files\Webots\lib\controller\Controller")
    wb.wb_robot_init()
    wb.wb_supervisor_node_get_from_id.restype = Ref
    wb.wb_supervisor_node_get_self.restype = Ref
    wb.wb_supervisor_node_get_field_by_index.restype = Ref
    wb.wb_supervisor_field_get_name.restype = c_char_p
    world_info = wb.wb_supervisor_node_get_from_id(1)
    self = wb.wb_supervisor_node_get_self()
    # FPS_field = wb.wb_supervisor_node_get_field_by_index(world_info, 8) # uncommenting this fubars getting the rotation field
[...]

PR #4366, opened 2022-03-21
fix get_field_by_index (#4363)

Justin-Fisher:
*I*HAVE*NOT*TESTED*THIS*!!! wb_supervisor_node_get_field_by_index was calling find_field_by_id with the node id and index arguments reversed, causing it to return the wrong field if the one with the numbers reversed had been accessed. Similarly for the proto version.


omichel:
Unfortunately, it doesn't work and some CI tests are not passing anymore with this change...

PR #4370, opened 2022-03-21
Fix stdin redirection

### LINKED ISSUE #4369: Wrong `freopen` flag for stdin under windows
In `src/webots/gui/main.cpp` on [Line 47](https://github.com/cyberbotics/webots/blob/f5ae03cbfb535553ab7870963a3f3479e913c2a8/src/webots/gui/main.cpp#L47) It says:
```
   (void)freopen("CONIN$", "w", stdin);
```
I believe this should be a `"r"` flag for reading. Since I don't have a windows machine, I can't actually test this, but many entries on the internet give a hint that this a bug in webots.

omichel:
Fixes #4369.


ad-daniel (approved):
Thank you

Issue #4362, opened 2022-03-18
ROS Contact Point interface doesn't follow documentation; docs also not very descriptive


nathanieltagg:
**Describe the Bug**
ROS controller does not present contact points as documentation lists

**Steps to Reproduce**
Start webots with a ROS-controlled robot.
$ rosservice list | grep contact
yields these services:
/MyRobot/supervisor/node/disable_contact_points_tracking
/MyRobot/supervisor/node/enable_contact_point_tracking
/MyRobot/supervisor/node/get_contact_point
/MyRobot/supervisor/node/get_contact_point_node
/MyRobot/supervisor/node/get_number_of_contact_points

**Expected behavior**
The documentation says there should be a service /get_contact_points
 https://cyberbotics.com/doc/reference/supervisor?tab-language=ros#wb_supervisor_node_get_contact_points
The currently advertised service "get_contact_point" seems to be able to access only a single point at a time.

It's not obvious how to use contact points.  For instance, asking for get_number_of_contact_points on my robot node, with recusion on, seems to crash the ros controller.

**System**
R2022a

ad-daniel:
You're right, in theory those API functions have been deprecated [here](https://github.com/cyberbotics/webots/pull/3162) and the change to `get_contact_pointS` introduced with `R2021b`. Not sure why they're still mentioned, either way the documented services are the correct one. I'll look into it.

PR #4371, opened 2022-03-21
Remove advertising of deprecated functions and cleanup

ad-daniel:
**Description**

Fixes https://github.com/cyberbotics/webots/issues/4362

Since the functions `get_number_of_contact_points`, `get_contact_point` and `get_contact_point_node` have been deprecated in favor of `get_contact_points`, the former ones should not be advertised whereas the latter should.

This PR also cleans up a typo in the virtual reality service

### DIFF
--- a/projects/default/controllers/ros/headersGenerator.py
+++ b/projects/default/controllers/ros/headersGenerator.py
@@ -35,7 +35,7 @@ class HeadersGenerator:
 
     # these md5 sums are copied from the corresponding header files in 'include'
     # directory
-    predifinedMD5 = {
+    predefinedMD5 = {
         'Header': '2176decaecbce78abc3b96ef049fabed',
         'geometry_msgs/Point': '4a842b65f413084dc2b10fb484ea7f17',
         'geometry_msgs/PointStamped': 'c63aecb41bfdfd6b7e1fac37c7cbe7bf',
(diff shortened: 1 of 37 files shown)


ad-daniel:
The issue with the md5 stemmed from the fact that when computing it we should ignore the arrays, however when doing the replacement with the custom message (which also was absent), it was done incorrectly.

The reason it surfaced only now is that in RosSupervisor, we mixed the names for the servers [here](https://github.com/cyberbotics/webots/blob/master/projects/default/controllers/ros/RosSupervisor.cpp#L81) (note the variable is called `mNodeGetContactPointServer` instead of `mNodeGetContactPoint>>s<<Server`). There used to be a `get_contact_point` service so it didn't raise any error.

projects/default/controllers/ros/headersGenerator.py
```diff
@@ -72,13 +73,13 @@ def ros_md5sum(self, srv):
...
                 l = l.split('#')[0].strip()  # strip comments
-                if "geometry_msgs/" in l or "std_msgs/" in l or "sensor_msgs/" in l or "Header" in l:
-                    # replace message by its corresponding md5
+                if "geometry_msgs/" in l or "std_msgs/" in l or "sensor_msgs/" in l or "Header" in l or 'webots_ros/' in l:
                     message = l.split(' ')[0]
-                    # the fact that it is an array is not releavant for the computation of the MD5
-                    message = message.replace("[]", "")
-                    if message in self.predifinedMD5:
-                        l = l.replace(message, self.predifinedMD5[message])
+                    # replace message by its corresponding md5
```
omichel, line 78: Consistency with the following comment:
```suggestion
                    # replace message by its corresponding MD5
```

omichel (commented):
Thank you. I just have one minor comment.

omichel (approved):
Thank you.

PR #4372, opened 2022-03-21
Fixed stdout/stderr redirection when all Webots console are closed

### LINKED ISSUE #4368: No Stdout output when no webots console is open
**Describe the Bug**
If I run webots with `--stdout --stderr` I only receive output in bash if I have a console open in webots (which shows nothing).
If I start webots with any console open, I also don't receive anyoutput in bash. It starts outputting as soon I open a console in webots.

**Context**
This was actually run in the RoboCup version of webots, but I am strongly assuming that the same issue exists in webots.

omichel:
Fixes #4368.

### DIFF
--- a/docs/reference/changelog-r2022.md
+++ b/docs/reference/changelog-r2022.md
@@ -12,6 +12,7 @@ Released on XX XX, 2022.
   - Cleanup
     - Removed `wb_robot_get_type` API function as it no longer serves a purpose ([#4125](https://github.com/cyberbotics/webots/pull/4125)).
   - Bug fixes
+    - Fixed redirection of stdout/stderr to the terminal when no Webots console is open ([#4372](https://github.com/cyberbotics/webots/pull/4372)). 
     - Fixed the URDF exportation of [SolidReference](solidreference.md) nodes ([#4102](https://github.com/cyberbotics/webots/pull/4102)).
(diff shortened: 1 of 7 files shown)


SGSSGene (dismissed):
I have cherry-picked the commits onto the RoboCup-fork and it worked as expected. Solving my issue, Thank you!

ad-daniel:
Insofar as printing the messages when the console is closed it does work as intended, but if a console is open, messages are printed both in the webots console and in the terminal. I don't think this is intended, after all one of the benefits that `--stdout --stderr` provided (to me at least) was that by only outputting to the terminal you could gain significant performance (especially if you print a lot of info). So this would be lost

omichel:
> Insofar as printing the messages when the console is closed it does work as intended, but if a console is open, messages are printed both in the webots console and in the terminal. I don't think this is intended, after all one of the benefits that `--stdout --stderr` provided (to me at least) was that by only outputting to the terminal you could gain significant performance (especially if you print a lot of info). So this would be lost

The behavior you describe was indeed wanted. I believe that if a user don't want to see messages in the Webots console and gain performance, she has several options:
1. close all the consoles in Webots.
2. filter out all the messages that display in the Webots consoles (or allow only a few important messages, like errors or warnings).

It is simple for the users to understand that the `--stdout` and `--stderr` option won't affect the Webots consoles, but only the Webots streams (stdout and stderr).

ad-daniel (approved):
Thank you

PR #4383, opened 2022-03-23
Fix scheduled tests and nightly header

ad-daniel:
**Description**

The failed test revealed a tagging issue with the generation of the nightly.
For reference, the latest master commit is `d45037c` and the latest develop commit is `365d318`. The reason the cache test failed is that the urls of the distribution uploaded as artifact are correct (i.e `365d318`), but when updating the urls to run the test suite (which are not included in the distribution), the assets were tagged with `d45037c`, and in master of course the texture used for the test doesn't exist.

Fundamentally the issue is that the nightly builds and scheduled tests are run from the master branch workflows, so any reference of `${{ github.sha }}` will reference the `d45037c`.
To see this is in fact the case, when checking the headers of the latest nightlies they both mention `d45037c`

a
b

Given this, I'm not entirely sure why the urls themselves in the distributions end up being correct, but I suspect it's because somehow it gets the right hash here: https://github.com/cyberbotics/webots/blob/d45037c97eeaf46622e2973259e4d0ec582c8567/scripts/packaging/Makefile#L66-L70

Either way, a possible fix is to get the hash from the latest commit, that way whether it's run from the master or develop workflow doesn't really matter.

### DIFF
--- a/.github/workflows/test_suite_linux.yml
+++ b/.github/workflows/test_suite_linux.yml
@@ -49,7 +49,7 @@ jobs:
         sudo scripts/install/linux_optional_compilation_dependencies.sh
     - name: Set Commit SHA in Version
       if: ${{ github.event_name == 'schedule' }}
-      run: python scripts/packaging/set_commit_and_date_in_version.py ${{ github.sha }}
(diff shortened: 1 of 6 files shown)


omichel (approved):
That seems good to me. Thanks.

PR #4386, opened 2022-03-23
Add parameter name to pit

BenjaminDeleze:
Without this parameter, inserting several pits in a single worlds will trigger warnings


omichel (approved):
Thank you.

PR #4392, opened 2022-03-23
Fixed e-puck robot window

omichel:
The e-puck robot window was broken on the develop branch and corresponding nightly builds.
Also the code in WbStreamingServer.cpp needed some clean-up.

### DIFF
--- a/src/webots/gui/WbStreamingServer.cpp
+++ b/src/webots/gui/WbStreamingServer.cpp
@@ -190,39 +190,31 @@ void WbStreamingServer::onNewTcpConnection() {
 void WbStreamingServer::onNewTcpData() {
   QTcpSocket *socket = qobject_cast<QTcpSocket *>(sender());
 
-  const QString &line(socket->peek(1024));  // Peek the request header to determine the requested url.
+  const QString &line(socket->peek(8 * 1024));  // Peek the request header to determine the requested url.
   QStringList tokens = QString(line).split(QRegularExpression("[ \r\n][ \r\n]*"));
-  if (tokens[0] == "GET") {
-    const QString &requestedUrl(tokens[1].replace(QRegularExpression("^/"), ""));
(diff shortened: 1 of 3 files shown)


src/webots/gui/WbStreamingServer.cpp
```diff
@@ -190,54 +190,44 @@ void WbStreamingServer::onNewTcpConnection() {
...
+  const QString requestedUrl = completeUrl.left(completeUrl.lastIndexOf('?'));
+  QString filePath;
+  static const QStringList streamerFiles = {"index.html", "setup_viewer.js", "style.css", "webots_icon.png"};
+  if (requestedUrl == "streaming")  // shortcut
+    filePath = WbStandardPaths::resourcesWebPath() + "streaming_viewer/index.html";
+  else if (streamerFiles.contains(requestedUrl))
+    filePath = WbStandardPaths::resourcesWebPath() + "streaming_viewer/" + requestedUrl;
+  else if (requestedUrl.startsWith("robot_windows/"))
+    filePath = WbProject::current()->pluginsPath() + requestedUrl;
+  else if (requestedUrl.startsWith("resources/projects/plugins/robot_windows/generic/") || requestedUrl.endsWith(".js") ||
```
BenjaminDeleze, line 220: This seems wrong, at least when I try to open a generic robot window, it does not find it.
But it bugs only in the development environment.

BenjaminDeleze (changes requested):
Seems good, only one remark:

src/webots/gui/WbStreamingServer.cpp
omichel: I believe I fixed it in a simple way, without breaking any backwards compatibility.
However, I will open a new issue about URLs for robot windows and robot controllers, that should behave like URL for images, sounds and meshes, with either relative or absolute references as discussed.

BenjaminDeleze (approved):
Thank you

PR #4393, opened 2022-03-23
Better asset archive name and cache bugfixes

ad-daniel:
**Description**

Improve asset zip file name to reflect branch.

This PR also fixes:
- warning when a contact property is present but no url is provided.
- use full hash when doing the conversion `webots://` to `https://`
- due to the presence of procedural PROTO, the previous method of generating the cache archive missed several files

And adds:
- Cache completeness check in the nightly build (when running `update world`)

### DIFF
--- a/scripts/packaging/generate_asset_cache.py
+++ b/scripts/packaging/generate_asset_cache.py
@@ -23,7 +23,11 @@ import hashlib
 
 EXTENSIONS = [
     'jpg', 'JPG', 'png', 'PNG', 'jpeg', 'JPEG', 'stl', 'STL', 'dae', 'DAE', 'obj', 'OBJ', 'mp3', 'MP3', 'wav', 'WAV',
-    'hdr', 'HDR'
+    'hdr', 'HDR', 'fbx', 'FBX'
+]
+
+SKIPPED_DIRECTORIES = [
+    '/icon/', '/icons/', '/robot_windows/'
 ]
 
 # ensure WEBOTS_HOME is set and tag was provided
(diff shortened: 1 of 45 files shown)


scripts/packaging/generate_asset_cache.py
```diff
@@ -17,11 +17,19 @@
...
 import os
 from pathlib import Path
-import re
 import sys
 import shutil
 import hashlib
 
+EXTENSIONS = [
+    'jpg', 'JPG', 'png', 'PNG', 'jpeg', 'JPEG', 'stl', 'STL', 'dae', 'DAE', 'obj', 'OBJ', 'mp3', 'MP3', 'wav', 'WAV',
+    'hdr', 'HDR', 'fbx', 'FBX'
```
omichel, line 26: I believe we can remove all the uppercase versions as we don't have and won't have uppercase extensions in the Webots distribution. Also, I believe we can remove the `.jpeg` which we also don't and won't have.

ad-daniel:
I've added a check on world update which compares the amount of assets in the cache before and after running the test in order to ensure that the cache is indeed being used (and complete).

I've tried again generate a distribution locally and to use it offline and it appears to be working as intended. To test it locally requires to run this:

```
python scripts/packaging/set_commit_and_date_in_version.py $(git log -1 --format='%H')
make -j12 distrib
```
and make sure the WEBOTS_HOME points to the right directory.

From the CI it seems to be working as intended, but I can double check with the next nightly the offline behavior once more.

BenjaminDeleze (approved):
Thank you!

PR #4394, opened 2022-03-25
Add support for track animation in WebotsJS

BenjaminDeleze:
**Tasks**
  - [x] Animation of the Trackwheels
  - [x] Animation of the Track

**Additional context**
Concerning the Trackwheels, it had already a rotation attribute in Webots but it was an internal one. It was enough to turn this internal rotation into a hidden field for it to be automatically exported.

Concerning the track itself, it is more tricky because the elements of the track are not real nodes: they are created and handled directly by the track node. Following their evolution would be difficult and require to work in another way than what is already implemented (based on fields).
To bypass this problem, the movement of the track will be implemented directly in WebotsJS.

I had to force the export of the trackwheels to x3d, even if they do not have children because we need to know how many wheels there is to compute the path of the track segments.

You can test it here: https://cyberbotics.com/animations/track/track.html


resources/web/streaming_viewer/index.html
```diff
@@ -23,7 +23,7 @@ <h1>Webots streaming viewer</h1>
       </p>
     </header>
     <webots-view style="height:calc(100% - 120px);"></webots-view>
-    <script type="module" src="https://cyberbotics.com/wwi/R2022b/WebotsView.js"></script>
+    <script type="module" src="../wwi/WebotsView.js"></script>
```
omichel, line 26: Shouldn't we keep an absolute reference here instead of a relative one?

src/webots/nodes/WbTrack.cpp
```diff
@@ -908,8 +885,7 @@ void WbTrack::exportAnimatedGeometriesMesh(WbVrmlWriter &writer) const {
 
   if (writer.isX3d()) {
     writer << "<Transform ";
-    writer << "translation='" << position << "' ";
-    writer << "rotation='" << rotation << "'>";
+    writer << " type='animatedgeometry'>";
```
omichel, line 888: There are two consecutive spaces, here, it would be simpler to write a single line instead, like:
```c++
    write << "<Transform type='animatedgeometry'>";
```

omichel, line 888: Also I am not sure why we write 'animatedgeometry' here whereas I would have written 'track', but track seems to be used elsewhere as well. Not sure what is the difference...

omichel (changes requested):
Overall it looks very good to me. I just have a couple of suggestions and questions.

resources/web/streaming_viewer/index.html
BenjaminDeleze: yes, I forgot to change it back

src/webots/nodes/WbTrack.cpp
BenjaminDeleze: `type=track` is for the whole node `track` whereas `type=animatedgeometry` is for the track belt element (that is called `animatedgeometry` in Webots.

omichel: As discussed, I would suggest to use "role" instead of "type" for this kind of information (and to update the ImageTexture code accordingly to be consistent, and maybe others). However, this could be done in a separate PR.

BenjaminDeleze:
@omichel, as discussed I introduced the distinction between `type` and `role`

omichel (changes requested):

BenjaminDeleze:
Please do not merge it. 

I want to adapt the animation of CodeInstitute first and generate a new docker image for the server

PR #4402, opened 2022-03-28
Fix crash when adding animatedGeometries when no trackwheels are there

BenjaminDeleze:
ditto


src/webots/nodes/WbTrack.cpp
```diff
@@ -468,6 +468,9 @@ void WbTrack::updateAnimatedGeometriesAfterFinalization(WbBaseNode *node) {
 void WbTrack::updateAnimatedGeometries() {
   clearAnimatedGeometries();
 
+  if (mWheelsList.size() <= 0)
```
omichel, line 471: The size() of a QVector (alias QList) cannot be negative:
```suggestion
  if (mWheelsList.isEmpty())
```

omichel (approved):

PR #4423, opened 2022-04-01
Add css file for web scene/animation/simulation

ThomasOliverKimble:
**Description**
After seeing the Robocup matches embedded on their website, it was decided that it was a good idea to export a `CSS` file alongside the `HTML` file to allow users to quickly have a nice looking web animation or scene.
Also modified the streaming viewer `HTML` and `CSS` to match (see picture below, ignore the lack of materials in the scene).
I used the colors from [cyberbotics.com](https://cyberbotics.com/) rather than [webots.cloud](https://beta.webots.cloud/) but this can be changed.

**Tasks**
  - [ ] Review code 
  - [ ] Agree on the design (colors, sizing, text, etc.)
  - [ ] Confirm documentation modifications.

**Screenshots**
web-scene
web-animation
streaming-server-screenshot


BenjaminDeleze (changes requested):
Seems good to me, however I have some suggestions:

- I would enable the css by default, otherwise they will do it.
- Thus I would put only the image with css styling in the doc.
- I am not totally convinced by the colors of `connect/disconnect` button, even so I think that coloring the button is a nice idea. Maybe the red should be the same as the one of the Webots logo.
- In animations/scenes the page has an vertical overflow. I think that it should not.
- In animations/scenes, I am not convinced about the align-left for the text of the footer.

ThomasOliverKimble:
> Seems good to me, however I have some suggestions:
> 
> * I would enable the css by default, otherwise they will do it.
> * Thus I would put only the image with css styling in the doc.
> * I am not totally convinced by the colors of `connect/disconnect` button, even so I think that coloring the button is a nice idea. Maybe the red should be the same as the one of the Webots logo.
> * In animations/scenes the page has an vertical overflow. I think that it should not.
> * In animations/scenes, I am not convinced about the align-left for the text of the footer.

Modifications have been applied:
* css enabled by default -> only css image in doc
* I agree on the colored buttons -> they are now always blue
* I have removed vertical overflow on animations/scenes, now there is only a scrollbar in the `Description` section if it is too long
* footer is now aligned left

docs/guide/web-animation.md
```diff
@@ -22,7 +22,9 @@ Webots will ask to playback the resulting file in the default Web browser (from
 
 For more details about the HTML export dialog please refer to [this section](web-scene.md#how-to-export-a-web-scene).
 
-**Note**: The `X3D` file, the `JSON` animation file and the required textures are exported in the same directory as the target `HTML` file.
+**Note**: The `CSS` file, the `X3D` file, the `JSON` animation file and the required textures are exported in the same directory as the target `HTML` file.
+
+**Note**: The `CSS` file is meant be used as a styling guide but can be modified/removed.
```
BenjaminDeleze, line 27:
```suggestion
**Note**: The `CSS` file is meant to be used as a styling guide but can be modified/removed.
```

docs/guide/web-scene.md
```diff
@@ -26,14 +26,16 @@ The 3D navigation in the player is possible using the mouse or the touch screen,
 Select the `File / Export HTML5 Scene...` menu item and choose the target `HTML` file in the pop-up dialog.
 When the export is completed, Webots will ask to playback the resulting file in the default Web browser.
 
-**Note**: The `X3D` file and the required textures are exported in the same directory as the target `HTML` file.
+**Note**: The `CSS` file, the `X3D` file and the required textures are exported in the same directory as the target `HTML` file.
+
+**Note**: The `CSS` file is meant be used as a styling guide but can be modified/removed.
```
BenjaminDeleze, line 31:
```suggestion
**Note**: The `CSS` file is meant to be used as a styling guide but can be modified/removed.
```

resources/web/streaming_viewer/index.html
```diff
@@ -11,18 +11,31 @@
...
+        </div>
+        <div class="grid-item"><p>Broadcast:</p></div>
+        <div class="grid-item">
+          <input id="broadcast" type="checkbox"/>
+        </div>
+      </div>
     </header>
-    <webots-view style="height:calc(100% - 120px);padding:20px;"></webots-view>
+    <div class="webots-view-container" style="margin: 0; height: calc(100vh - 176px); display:flex; justify-content: center;" >
+      <webots-view style="width:100%; display:block; position: relative"></webots-view>
```
BenjaminDeleze, line 37: I am not convinced of using inline CSS, if we also have an external CSS file

resources/web/templates/x3d_playback.html
```diff
@@ -4,19 +4,22 @@
...
-    <h1 style="height:5%">%title%</h1>
-    <webots-view style="height:80%; display:block;" data-scene=%x3dName% data-animation=%jsonName%></webots-view>
-    <p style="height:5%">If the animation is not loading correctly, please refer to the <a style='color: #5DADE2;' target:'_blank' href='https://cyberbotics.com/doc/guide/web-scene#remarks-on-the-used-technologies-and-their-limitations'>User Guide</a>
-      it could be that your browser prevents local files CORS requests.</p>
-    <p style="height:5%">%description%</p>
+    <div class="title-container">
+      <h1>Web %type%: <strong>%title%</strong></h1>
+    </div>
+    <div class="webots-view-container" style="margin:0; display:flex; justify-content: center;" >
+      <webots-view style="width:100%; display:block; position: relative" data-scene=%x3dName% data-animation=%jsonName%></webots-view>
```
BenjaminDeleze, line 15: Same remark concerning the inline css

BenjaminDeleze (changes requested):
Very good, juste some minors remarks

BenjaminDeleze (approved):
Thank you

PR #4428, opened 2022-04-04
Cleanup pbr export

BenjaminDeleze:
- For PBRAppearance, now export only PBR (and not also simple Appearance).
- Export `ccw` as boolean and not 1/0 to match consistency.
- Export version of Webots to the .x3d
- Remove unused parameters

### DIFF
--- a/resources/web/wwi/Parser.js
+++ b/resources/web/wwi/Parser.js
@@ -493,7 +493,7 @@ export default class Parser {
         if (typeof geometry !== 'undefined')
           continue;
       }
-      console.log('Parser: error with node: ' + child.tagName + '. Either the node is unknown or the same shape contains several appearances/geometry.');
+      console.log('Parser: error with node: ' + child.tagName + '. Either the node is unknown or the same shape contains several appearances/geometries.');
     }
 
     const shape = new WbShape(id, castShadows, isPickable, geometry, appearance);
--- a/resources/web/wwi/Parser.js
+++ b/resources/web/wwi/Parser.js
@@ -471,29 +471,19 @@ export default class Parser {
     let geometry;
     let appearance;
 
-    for (let i = 0; i < node.childNodes.length; i++) {
+    for (let i = node.childNodes.length - 1; i >= 0; i--) {
       const child = node.childNodes[i];
       if (typeof child.tagName === 'undefined')
         continue;
 
       if (typeof appearance === 'undefined') {
-        if (child.tagName === 'Appearance') {
-          if (isBoundingObject)
-            continue;
-          // If a sibling PBRAppearance is detected, prefer it.
-          let pbrAppearanceChild = false;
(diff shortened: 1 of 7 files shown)


resources/web/wwi/Parser.js
```diff
@@ -503,13 +493,9 @@ export default class Parser {
         if (typeof geometry !== 'undefined')
           continue;
       }
-
-      console.log('X3dLoader: Unknown node: ' + child.tagName);
+      console.log('Parser: error with node: ' + child.tagName + '. Either the node is unknown or the same shape contains several appearances/geometry.');
```
ad-daniel, line 496:
```suggestion
      console.log('Parser: error with node: ' + child.tagName + '. Either the node is unknown or the same shape contains several appearances/geometries.');
```

src/webots/nodes/WbPbrAppearance.hpp
```diff
@@ -72,11 +72,15 @@ class WbPbrAppearance : public WbAbstractAppearance {
...
-  void exportNodeFooter(WbVrmlWriter &writer) const override;
+  void exportNodeFields(WbVrmlWriter &writer) const override;
   const QString &vrmlName() const override {
     static const QString name("Appearance");
     return name;
   }
+  const QString &x3dName() const override {
+    static const QString name("PBRAppearance");
+    return name;
+  }
```
ad-daniel, line 83:
```suggestion
  const QString &x3dName() const override { return nodeModelName(); }
```

resources/web/wwi/Parser.js
```diff
@@ -471,29 +471,19 @@ export default class Parser {
     let geometry;
     let appearance;
 
-    for (let i = 0; i < node.childNodes.length; i++) {
+    for (let i = node.childNodes.length - 1; i >= 0; i--) {
```
ad-daniel, line 474: more of a question, I've noticed you've reversed the order here. So the order in which children are provided matters? I must have picked the right one by chance with ColladaShape then

ad-daniel (changes requested):
I've tested a few exports and it appears to be behave correctly, thank you

resources/web/wwi/Parser.js
BenjaminDeleze: No the order does normally not matter.

I reversed the order because the old animation were build like:
```
<Shape>
  <Appearance> </Appearance>
  <PBRAppearance> </PBRAppearance>
  <Geometry></Geometry>
</Ahape>
```

If someone try to play an old animation, it will work and by reverting the order it will select the `PBRAppearance` and ignore the `Appearance` as the parser now only care for the first Appearance it encounters.

omichel: In that case, we should add a comment in the code about it otherwise, we won't understand what it was done this way.

ad-daniel (approved):
Thank you

PR #4434, opened 2022-04-06
Fix multiple initialization of generic robot window resulting in bugs

BenjaminDeleze:
ditto


omichel:
Why does a robot window receive several configure messages?

BenjaminDeleze:
In a normal usage, it does normally not receive several configure messages, but this could happen for example:

- if several tabs open the same robot windows.
- if a robot window is opened and refreshed before is has received the config message.

omichel (changes requested):
I would suggest to add a comment in the code to explain why we need to handle this case.

omichel (approved):

PR #4437, opened 2022-04-07
Fix min size toolbar

BenjaminDeleze:
As the toolbar elements are now larger, the min size must be increased accordingly


ThomasOliverKimble (approved):
My bad, forgot to update these... Looks good.

PR #4445, opened 2022-04-11
Enhancement animation loading

BenjaminDeleze:
After seeing the animation generated by Robocup (https://robocup-hlvs.de/GAME_K-GD6-3/public_logs/K-GD6-3.html) , I improved the loading time of the animation by:

- starting the loading of JSON ealier.
- Then I generated some keyframes (think about it like snapshots of the simulation) in order to increase the response time when jumping forward or backward in the replay. The keyframes are generated during the loading.
  - I measured an overhead of 300ms on the Robocup animation for a total loading time of about 16 seconds. It seems acceptable to me knowing that this animation has a JSON file of 400MB.
- I also reordered some loops to minimize the number of iterations
- Finally I made some cleanup.

To test and compare with the original: https://cyberbotics.com/animations/robocup/K-GD6-3.html

### DIFF
--- a/resources/web/wwi/Animation.js
+++ b/resources/web/wwi/Animation.js
@@ -89,7 +89,7 @@ export default class Animation {
               allPoses.set(poses[j].id, currentIdFields);
             }
           }
-        } else { // Check the previous keyFrames to get missing updates
+        } else { // Check the previous keyFrame to get missing updates
           const poses = this._keyFrames.get(i - 1).poses;
           for (let element of poses) {
             const id = element[0];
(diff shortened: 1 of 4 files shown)


resources/web/wwi/Animation.js
```diff
@@ -89,75 +157,123 @@ export default class Animation {
...
+          previousPoseStep = (closestKeyFrame + 1) * this.keyFrameStepSize;
+          if (this._keyFrames.size > 0)
+            previousStepIsAKeyFrame = true;
+        }
+
+        const completeIds = new Set();
+        const appliedFieldsByIds = new Map();
+        const appliedLabelsIds = new Set();
+
+        // We do not want to include the previousPoseStep in the loop as its updates are in the keyFrame. However, we need to include it if their is no keyFrames or if it is the step 0 as their is no keyFrame for it
```
omichel, line 178: their -> there

resources/web/wwi/Animation.js
```diff
@@ -89,75 +157,123 @@ export default class Animation {
...
+        if (previousStepIsAKeyFrame || previousPoseStep !== 0)
+          previousPoseStep++;
+
+        for (let i = this.step; i >= previousPoseStep; i--) { // Iterate through each step until the nearest keyFrame is reached or all necessary updates have been applied. Go in decreasing order to minize the number of step.
```
omichel, line 182: steps

omichel (changes requested):
That looks good to me.

omichel:
It works very well! However, when testing on both Firefox and Chrome, there is a waiting time after the scene is display and before the buttons of the GUI are displayed. What is webotsjs doing at this point? It would be nice to display a progress message here, maybe, otherwise the user may believe that it is just displaying a static picture...

BenjaminDeleze:
> It works very well! However, when testing on both Firefox and Chrome, there is a waiting time after the scene is displayed and before the buttons of the GUI are displayed. What is webotsjs doing at this point? It would be nice to display a progress message here, maybe, otherwise the user may believe that it is just displaying a static picture...

It is when the scene has finished to load, but not the json.
Indeed it would be nice, but as @ThomasOliverKimble is reworking the progress bar, I thought it was maybe better to wait until the new version is out.

resources/web/wwi/Animation.js
```diff
@@ -89,75 +157,123 @@ export default class Animation {
...
+          previousPoseStep = (closestKeyFrame + 1) * this.keyFrameStepSize;
+          if (this._keyFrames.size > 0)
+            previousStepIsAKeyFrame = true;
+        }
+
+        const completeIds = new Set();
+        const appliedFieldsByIds = new Map();
+        const appliedLabelsIds = new Set();
+
+        // We do not want to include the previousPoseStep in the loop as its updates are in the keyFrame. However, we need to include it if there is no keyFrames or if it is the step 0 as their is no keyFrame for it
```
omichel, line 178:
```suggestion
        // We do not want to include the previousPoseStep in the loop as its updates are in the keyFrame. However, we need to include it if there is no keyFrames or if it is the step 0 as there is no keyFrame for it
```

omichel (changes requested):

omichel:
Not sure why the mac source tests are failing...

resources/web/wwi/Animation.js
BenjaminDeleze: oups

omichel (approved):
Thank you.

PR #4446, opened 2022-04-11
Fix streaming run button

ThomasOliverKimble:
**Description**
The run button was not working with the new toolbar design. It is now fixed, and the play and pause button have been modified accordingly (design and animations).

**Screenshots**
pause_run
play_pause
play_run

ps. Ignore the broken world, was using a broken version of webots...

### DIFF
--- a/resources/web/streaming_viewer/index.html
+++ b/resources/web/streaming_viewer/index.html
@@ -23,7 +23,7 @@
       </p>
     </header>
     <webots-view style="height:calc(100% - 120px);"></webots-view>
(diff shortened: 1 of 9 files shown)


BenjaminDeleze:
It works fine, but maybe the animations should be a little be faster. What do you think?

ThomasOliverKimble:
> It works fine, but maybe the animations should be a little be faster. What do you think?

I agree, faster animations feel more responsive.

BenjaminDeleze (approved):
Thank you

Issue #4447, opened 2022-04-12
Incongruence between different importing mechanisms


ad-daniel:
**Describe the Bug**

The following objet was created with blender (note the axis)
![blender](https://user-images.githubusercontent.com/44834743/162920392-0662698b-906a-42f8-b33b-bc69c97d896f.png)

The big sphere is supposed to be the front facing side, so the object was exported as ***X-foward, Z-up*** both in Collada and Wavefront formats. 

In webots, when importing these models using the "3D model importer" (`file > import 3d model`) we obtain this:
![import](https://user-images.githubusercontent.com/44834743/162920926-03603dd7-aa80-4e27-9d2a-571e3d79e62a.png)

Note they are oriented differently, (the collada shape appears to have an additional `1 0 0 1.5708` rotation) and that even if we were to remove this extra rotation, neither is actually facing X forward as in blender.

When importing the same objects in the Mesh node instead, we obtain this

![mesh](https://user-images.githubusercontent.com/44834743/162921839-b092a2ff-7fc8-4118-8ffc-8fb2eeae9eed.png)

Now they have the same orientation, but the Collada one is significantly bigger. Once again, neither is X-forward.

The result should be the same regardless of the importing mechanism being employed.

PR #4172, opened 2022-02-01
Fix ColladaShapes colors and orientations

BenjaminHug8:
**Description**
The `ColladaShapes` was not handling correctly the orientation of the meshes + it was not correctly taking into account the fact that a mesh can have multiple colors, resulting in a mix of the colors.

PR #4448, opened 2022-04-12
Fix collada orientation

ad-daniel:
**Description**

As surfaced in https://github.com/assimp/assimp/issues/849, importing in assimp non-obj files results in them being rotated 90°. Although the fix proposed in https://github.com/cyberbotics/webots/pull/4172 is technically correct, it ends up overwriting the root transform and therefore any translation & scaling information already present at the root level would end up being overwritten by the fix, which resulted in the behavior shown in the last image of https://github.com/cyberbotics/webots/issues/4447 (coordinate system not centered and scaling issue)

A simpler approach which doesn't suffer from this issue is to simply rotate the root transform by the desired amount.

From this issue also emerged the fact that when loading mesh nodes, the call to `WbTriangleMeshGeometry::preFinalize` builds the mesh prior to having determined whether it's collada or wavefront (which previously only occurred in the function `updateUrl` that comes after), hence `mIsCollada` needs to be defined both before this call and in `updateUrl`) for it to behave correctly.

### DIFF
--- a/src/webots/nodes/WbMesh.cpp
+++ b/src/webots/nodes/WbMesh.cpp
@@ -167,7 +167,8 @@ void WbMesh::updateTriangleMesh(bool issueWarnings) {
 
   // Assimp fix for up_axis, adapted from https://github.com/assimp/assimp/issues/849
   if (mIsCollada)  // rotate around X by 90° to swap Y and Z axis
-    scene->mRootNode->mTransformation *= aiMatrix4x4(1, 0, 0, 0, 0, 0, -1, 0, 0, 1, 0, 0, 0, 0, 0, 1);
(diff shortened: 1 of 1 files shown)


ad-daniel:
To test this, one can open either of  [these](https://github.com/cyberbotics/webots/files/8487453/test.zip) blender files and export them as collada/wavefront (both using the same convention, for instance X-forward, Z-up) and then import the resulting files into webots using a Mesh node.

Note: in blender, when exporting collada you need to check the "Apply" case. I think this is what has been messing with me.

BenjaminDeleze (approved):
oh the `apply` is vicious! I missed it too.

Seems good otherwise

Issue #4441, opened 2022-04-07
Ros controller crashes when requesting node information from device with a space in the name


nathanieltagg:
**Describe the Bug**
Querying the ROS controller for /RobotName/supervisor/get_from_device crashes the ROS controller if the device name has a space in it.

**Steps to Reproduce**
1. Load a simulation with the MiR robot and the default ROS controller
2. `rosservice call /Mir100/imu_data_accelerometer/get_model 1`
The controller crashes.

**Additional context**
The problem is obviously that the ROS controller advertises devices using the Ros::fixedNameString to remove ros-unfriendly characters.  This is handled fine when making /<controller>/<device>/... calls, but the /supervisor simply uses the un-translated name.

This can be fixed in several ways:
- Disallow spaces in device names when using the ros controller (e.g. generate a warning or error message in Ros.cpp when setting up the device table)
- Have the RosSupervisor class run through all available devices, checking for a match using the fixedNameString
- Change the call so that RosDevice handles this kind of request instead of the Supervisor.
 
In addition, RosSupervisor.cpp should include a check for a null device pointer around line 469 to avoid crashes.

Thanks,
Nathaniel

ad-daniel:
Isn't it the other way around? It crashes if you `get_from_device` while providing as argument the advertised (i.e. cleaned) name whereas it works when passing the actual name. For instance when calling
```
rosservice call /Mir100/supervisor/get_from_device "imu_data compass"
```
it checks the string `"imu_data compass"` against the `mDeviceList` list [here](https://github.com/cyberbotics/webots/blob/734324b084d8e4b52b37e8de0fd82beee1344f21/projects/default/controllers/ros/Ros.cpp#L234), which are instances of Device objects (technically RosDevice which wraps Device) and [results](https://github.com/cyberbotics/webots/blob/734324b084d8e4b52b37e8de0fd82beee1344f21/src/controller/cpp/Device.cpp#L25-L27) in a call of `wb_robot_get_device` which will find `"imu_data compass"` just fine.

The issue appears to be the fact that if we provide the sanitized name (which is also what we [get](https://github.com/cyberbotics/webots/blob/734324b084d8e4b52b37e8de0fd82beee1344f21/projects/default/controllers/ros/Ros.cpp#L450) when calling `/robot/get_device_list`) we do indeed get a crash of the controller since the device with such name does not exist

nathanieltagg:
[...]

PR #4449, opened 2022-04-13
Fix crash with unsanitized device names

ad-daniel:
**Description**
Fixes https://github.com/cyberbotics/webots/issues/4441

When advertising devices, their name is sanitized (i.e., spaces/dashes replaced by underscores), so for instance a `imu accelerometer` can be enabled with `rosservice call /MyRobot/imu_accelerometer/enable`. 

However, when we call for instance `/MyRobot/robot/get_device_list` a sanitized list of the devices was proposed to the user (in this case `imu_accelerometer` was returned), which if they were subsequently queried, for example to find the noderef of a particular device, resulted in a crash since none could be found with such a name. Since most devices used in the Webots library are unsanitized, I think it makes more sense for them to be accessible in an unsanitized way, otherwise the users need to manually edit the worlds.

### DIFF
--- a/projects/default/controllers/ros/Ros.cpp
+++ b/projects/default/controllers/ros/Ros.cpp
@@ -278,11 +278,11 @@ void Ros::fixName() {
 
   mRobotName = mRobot->getName();
   mRobotName += '_' + webotsPID + '_' + webotsHostname;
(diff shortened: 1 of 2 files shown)


BenjaminDeleze (approved):
Thank you

nathanieltagg:
Thanks for this!

PR #4452, opened 2022-04-14
Fix nightly issues

ad-daniel:
**Description**

As far as I can tell from the debug messages, the job that creates the release is always unable to include the distribution files itself created. As tonight's messages have shown, the binaries are indeed available so that's not the problem, therefore I suspect the issue lies with the fact that between [creating the release](https://github.com/cyberbotics/webots/blob/9c87a9d4f7332e764e6bb01947a761444bc488b1/scripts/packaging/publish_release.py#L95) on github and [requesting the list of all releases](https://github.com/cyberbotics/webots/blob/9c87a9d4f7332e764e6bb01947a761444bc488b1/scripts/packaging/publish_release.py#L127) available not enough time passes and therefore no upload occurs since the newly created release is not present (as far as the script is concerned).

I've added an error message and a delay to validate if indeed this is the case, but I suspect so since it's consistent with what we've seen so far

### DIFF
--- a/.github/workflows/test_suite_linux.yml
+++ b/.github/workflows/test_suite_linux.yml
@@ -57,7 +57,6 @@ jobs:
         export LIBGL_ALWAYS_SOFTWARE=true
         export ROS_DISTRO=${{ matrix.ROS_DISTRO }}
         xvfb-run --auto-servernum make distrib -j4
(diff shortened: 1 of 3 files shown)


BenjaminDeleze (approved):
Thank you

PR #4455, opened 2022-04-15
Fix macos build

ad-daniel:
**Description**

https://github.com/cyberbotics/webots/pull/4452 appears to have fixed the issues with the publishing of nightly packages, however macos failed because it appears to use a lower version of python


omichel (approved):
Thank you.

PR #4457, opened 2022-04-19
Handle BO in webotsJS as in webots

BenjaminDeleze:
Before this PR, the bounding objects were exported as indexedLineSet to x3d, inside `<switch>`. This PR changes the following things:

- [x] Bounding objects are now exported as geometric primitives (`box`, `sphere`, ...).
- [x] Improve the sanitization of input based on what is done in Webots desktop.
- [x] Creation of the `WbSolid.js` class.
- [x] Fixing some bugs:
  - [x] Weird behaviour of elevation grid when there is not enough value in `height` 
  - [x] Elevation grid thickness was not exported.
  - [x] Selection was not working correctly with object of id `-1`.
  - [x] LineSetShader was not exporting the normal value (was causing error in the web).
     - It was working fine on Firefox and desktop but not at all on Chrome.
  - [x] Gtao apply on the sphere (both Webots and webotsJS)
  - [x] `isInBoundingObject` was used both for an attribute and a function 
- [x] Support of USE in bounding object in webotsJS.


BenjaminDeleze:
While trying to fix an issue with the export of DEF/USE bounding object I encounter a strange situation.
It seems that `Group` or `Transform` bounding objects that contains directly a `Geometry` (whiteout a `Shape`) causes trouble with the DEF/USE mechanism.
The `Group` or `Transform` does not appear in the proposition when I try to add a USE bounding object.
However if I defined it directly in the world/PROTO file it works well (as this is the case for the Aibo robot for example).


Illustration:

```
Solid {
  boundingObject DEF TEST Group {
    children [
      Sphere {
        radius 0.1
      }
    ]
  }
}
Solid {
  translation 0 0.2 0
  name "solid(1)"
  boundingObject USE TEST
}
```

This works well.
But if I delete the bounding object of `solid(1)` in Webots, I cannot add it again.

@omichel is this the normal behaviour?

omichel:
> @omichel is this the normal behaviour?

No, `USE TEST` should appear in the list of nodes that could be inserted as a bounding object. So, that is a bug.

BenjaminDeleze:
> > @omichel is this the normal behaviour?
> 
> No, `USE TEST` should appear in the list of nodes that could be inserted as a bounding object. So, that is a bug.

Ok, I propose to open another issue as it does not concern this PR directly

resources/web/wwi/nodes/WbElevationGrid.js
```diff
@@ -99,4 +104,38 @@ export default class WbElevationGrid extends WbGeometry {
...
+    if (newYDimension)
+      this.yDimension = newYDimension;
+
+    const newYSpacing = resetIfNonPositive(this.ySpacing, 1.0);
+    if (newYSpacing)
+      this.ySpacing = newYSpacing;
+
+    this._checkHeight();
+
+    return !newTickness && !newXDimension && !newXSpacing && !newYDimension && !newYSpacing;
```
omichel, line 131: That seems odd as the default values for `newXDimension` is 0 and !0 is true while the default value for newXSpacing is 1 and !1 is false. See [here](https://stackoverflow.com/a/11550718). So I suspect there might be bug hidden here. Instead, we should probably test this way:
```suggestion
    return newTickness === false && newXDimension === false && newXSpacing === false && newYDimension === false && newYSpacing === false;
```
The same issue may arise in other implementations of the `_sanitizeFields()` method.

resources/web/wwi/nodes/utils/WbFieldChecker.js
```diff
@@ -0,0 +1,38 @@
+function resetIfNonPositive(value, defaultValue) {
+  if (value <= 0) {
+    console.error('Invalid ' + value + ' changed to ' + defaultValue + '. The value should be positive.');
```
omichel, line 3: I would display a warning here, not an error, as it is not fatal.
```suggestion
    console.warn('Invalid ' + value + ' changed to ' + defaultValue + '. The value should be positive.');
```
The same applies in similar functions below.

src/webots/nodes/WbGroup.cpp
```diff
@@ -436,13 +436,19 @@ void WbGroup::readHiddenKinematicParameter(WbField *field) {
 void WbGroup::exportBoundingObjectToX3D(WbVrmlWriter &writer) const {
   assert(writer.isX3d());
 
-  writer << "<Group>";
+  if (isUseNode() && defNode())
+    writer << "<" << x3dName()
+           << " role='boundingObject' USE=\'n" + QString::number(defNode()->uniqueId()) + "\'></" + x3dName() + ">";
```
omichel, line 441: Can we simplify this way?
```suggestion
           << " role='boundingObject' USE=\'n" + QString::number(defNode()->uniqueId()) + "\' />";
```
This would also apply in other classes.

omichel (changes requested):
I am super convinced by the need of the WbSolid class, as it doesn't correspond to a X3D object unlike other classes. So that makes a special case for the solid node.

BenjaminDeleze:
> I am super convinced by the need of the WbSolid class, as it doesn't correspond to a X3D object unlike other classes. So that makes a special case for the solid node.

I guess you wanted to say "not" super convinced.

The solid class is not strictly needed but it makes the code way more cleaner because you do not have to check: is this node a transform? if yes, then is this transform a solid? and also because it allows us to clean up the `WbTransform` class from the code that is not needed by transforms. And also it makes it more readable and simpler to understand in my opinion.
Moreover it is not the only class that does not correspond to X3D objects like `WbTrack`, `WbTrackWheel`, `WbAbstractAppearance`, `WbBaseNode`, `WbGeometry`.

It goes in the direction of having WebotsJS corresponds more to Webots.

For me the question is more to know if we really want to continue to export `solid`, `track`, `trackwheel` and `propeller` as transform or if we export them as <Solid>, <Track>, <Propeller>,...

omichel:
> I guess you wanted to say "not" super convinced.

Yes, sorry.

> The solid class is not strictly needed but it makes the code way more cleaner because you do not have to check: is this node a transform? if yes, then is this transform a solid? and also because it allows us to clean up the `WbTransform` class from the code that is not needed by transforms. And also it makes it more readable and simpler to understand in my opinion. Moreover it is not the only class that does not correspond to X3D objects like `WbTrack`, `WbTrackWheel`, `WbAbstractAppearance`, `WbBaseNode`, `WbGeometry`.
> 
> It goes in the direction of having WebotsJS corresponds more to Webots.
> 
> For me the question is more to know if we really want to continue to export `solid`, `track`, `trackwheel` and `propeller` as transform or if we export them as `<Solid>`, `<Track>`, `<Propeller>`, ...

Yes, I agree with you.

Let's keep the Solid class for now, in later we could consider exporting these nodes as you suggest. It seems to be a good idea.

src/webots/nodes/WbGroup.cpp
BenjaminDeleze: I tested it and yes we can.
I looked up at some documentation and found it is a common thing in XML (empty tag) but I found no mention for X3D, but as we are moving away from it, I think we can use empty tags.

omichel (approved):
Thank you.

BenjaminDeleze:
Please do not merge. I have to adapt the existing animations and create a new docker before merging.

PR #4468, opened 2022-04-21
Fix orientation dae

BenjaminDeleze:
It seems that the orientation of all the dae were broken some time ago


ygoumaz (approved):
It seems good to me. Thank you!

PR #4472, opened 2022-04-22
Sanitize html input and improve robustness of CSS

BenjaminDeleze:
Concerning the sanitization, https://developer.mozilla.org/en-US/docs/Web/API/HTML_Sanitizer_API seems promising but it still experimental and not well supported by browers.

I had to improve the robustness of the CSS due to the change introduced in #4423


resources/web/wwi/Parser.js
```diff
@@ -1249,4 +1249,10 @@ function rotateHDR(image, rotate) {
   return rotatedbits;
 }
 
+function sanitizeHTML(text) {
```
omichel, line 1252: Shouldn't we make it private, as it is used only from within this class?
```suggestion
function _sanitizeHTML(text) {
```

omichel (approved):
Thank you.

PR #4473, opened 2022-04-22
Fix fluid orientation

### LINKED ISSUE #4463: Immersion plane upside down
When using a `Cylinder` (or other geometries like `Capsule` and `Sphere`) in a `Fluid`, we get the following warning:
`WARNING: Fluid > USE CylinderGeom > Cylinder: The normal to this geometry's immersion plane is not opposed to the gravity vector. This may yield unexpected behaviors when immersing solids. (Please consult the Reference Manual for the definition of immersion planes.)`

And indeed the behaviour is not correct.
To test it:

- Open `salamander.wbt`
- Replace the bounding object by a `Cylinder`
- Save and reload
- Observe that the salamander is not swimming correctly
- Turn the Cylinder upside down (Z down)
- Save and reload
- It works well.

It seems that during the FLU conversion, the warning and the Fluid physics were converted such that the geometries in the bounding object of the Fluid should be Z down.

I checked in R2021b and it was this opposite (Y up).

So I think that we should change that to work as before (Z up)

BenjaminDeleze:
Fix #4463

### DIFF
--- a/projects/samples/geometries/worlds/floating_geometries.wbt
+++ b/projects/samples/geometries/worlds/floating_geometries.wbt
@@ -21,10 +21,8 @@ DEF SWIMMING_POOL SolidPipe {
 }
 DEF STILL_WATER Fluid {
   translation 0 0 0.2
-  rotation 1 0 0 1.5708
   children [
-    DEF S Transform {
-      rotation 1 0 0 1.5708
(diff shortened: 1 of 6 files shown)


omichel (approved):
Looks good to me.

Issue #4128, opened 2022-01-18
Python 3.10 support


SGSSGene:
Currently, webots only works with python 3.9 (or lower). It would be nice to have python 3.10 support.

SGSSGene:
Very cool, thank you. For reference, I have patched a few things, so I can run the Robocup-HumanoidTC branch.
My changes are too chaotic for a PR, but maybe helpful for the next person working on it.
https://github.com/SGSSGene/webots/commit/dcee03c99bb785fa4cb961e0f738875c40fff91b

omichel:
That looks like a very good start. Could you please go ahead and create a draft PR which we will review and possibly complete? I just sent you an invite to become a member of the Webots committers, this should facilitate your contribution.

omichel:
Note: @Justin-Fisher is currently developing a new Python API which should be fully independent from the Python version (for Python >= 3.5). See details at #3801. So, it may not be necessary in the long term to maintain manual support for each version of python...

Issue #1524, opened 2020-04-06
The process crashed some time after starting successfully


abylikhsanov:
**Describe the Bug**
After trying to go through the tutorial for creating the controller [here](https://cyberbotics.com/doc/guide/tutorial-1-your-first-simulation-in-webots?tab-language=python) I get a warning:

```
WARNING: e_p: The process crashed some time after starting successfully.
WARNING: 'e_p' controller crashed.
```

**Steps to Reproduce**
Follow the tutorial for creating a new controller and use the same code in C++ or Python. The bug seems to happen when creating a `Robot` object.

**System**
 - Operating System: [Linux Ubuntu 18.04]
 - Graphics Card: [NVIDIA GeForce RTX 2070 8 GB.]

DavidMansolino:
Can you share the code you are using (just to make 100% sure it is correct according to the tutorial) ?
You are using the 2020a-revision1 version of Webots right?

abylikhsanov:
Here is the code:

```
#include <webots/Robot.hpp>

// Added a new include file
#include <webots/Motor.hpp>

#define TIME_STEP 64

// All the webots classes are defined in the "webots" namespace
using namespace webots;

int main(int argc, char **argv) {
 Robot *robot = new Robot();

 // get the motor devices
 Motor *leftMotor = robot->getMotor("left wheel motor");
 Motor *rightMotor = robot->getMotor("right wheel motor");
 // set the target position of the motors
 leftMotor->setPosition(10.0);
 rightMotor->setPosition(10.0);

 while (robot->step(TIME_STEP) != -1);

 delete robot;

 return 0;
```

I installed the binaries (via `apt`) and I do remember it was 2020 version but do not remember letter or revision (I will check tomorrow in the office) but should be the latest release if I installed the binaries.

DavidMansolino:
I can indeed reproduce the crash (at least with the Python version of the controller).

abylikhsanov:
Hi, yes the version is `R2020a rev 1`

DavidMansolino:
I can indeed reproduce it with ``R2020a rev 1`` (but not with my development environment), I am currently installing a nightly build to see if this happens too.

DavidMansolino:
I can confirm it is crashing with the nightly releases too.

DavidMansolino:
The crash happens only with Python3.x (it is working fine with Python 2.7), and here is the stack on Mac:
```
Termination Signal:    Segmentation fault: 11
[...]

PR #4474, opened 2022-04-24
changes to compile with python3.10

ijnek:
**Description**
These changes are required to get webots working with Python3.10. The change from ``[0-9]`` to ``[0-9]*`` in the sed command is necessary to set ``PYTHON_VERSION`` to ``3.10`` and not ``3.1`` like it currently does.

While I see https://github.com/cyberbotics/webots/pull/3801 is in progress, this is necessary for a short term solution for Ubuntu Jammy since it uses Python3.10 by default.

**Related Issues**
Resolves: #4128

**Additional Notes**

``src/webots/control/WbLanguageTools.cpp`` contains a lot of references to python versions, but I have not made changes to that file.


ad-daniel (changes requested):
Thank you for the contribution! I think some adjustments are necessary also in `WbLanguageTools.cpp` and the version bump should be mentioned in the changelog

omichel (changes requested):

src/controller/python/.gitignore
```diff
@@ -1,2 +1,4 @@
 /controller??.o
+/controller???.0
```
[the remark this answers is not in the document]
ijnek: Oh my :facepalm:

ijnek:
@ad-daniel and @omichel thanks for the review and your work with the project.
I managed to get around to look at ``WbLanguageTools.cpp`` and make the appropriate changes.

Small note, I noticed that inside ``#elif __APPLE__``, python3.8 used to be preferred over python3.9 for some reason. Was that intentional?

ad-daniel:
I believe it was intentional, there were reported problems with higher versions of python on macOS, such as: https://github.com/cyberbotics/webots/issues/1524

PS: for mentioning the change in the changelog, it should be added to this file in a similar fashion: https://github.com/cyberbotics/webots/blob/master/docs/reference/changelog-r2022.md

ijnek:
From #1524, it seems that this is an issue affecting all Python versions >=3.8, which doesn't explain why 3.8 comes before 3.9 - they seem to be equally as problematic. 

Forgot about updating the Changelog, I'll do that now.

src/webots/control/WbLanguageTools.cpp
```diff
@@ -107,15 +107,20 @@ QString WbLanguageTools::pythonCommand(QString &shortVersion, const QString &com
...
-        }
+      }
+      if (pythonCommand == "!") {
+        pythonCommand = findWorkingPythonPath("3.8", env, false);
+        shortVersion = "38";
+      }
+      if (pythonCommand == "!") {
+        pythonCommand = findWorkingPythonPath("3.7", env, true);
+        shortVersion = "37";
+      }
```
omichel, line 123:
```suggestion
```

ad-daniel:
> From #1524, it seems that this is an issue affecting all Python versions >=3.8, which doesn't explain why 3.8 comes before 3.9 - they seem to be equally as problematic.

You're right, and if I have to make a guess it's likely because 3.9 was too recent/untested at that point. Either way I don't see a reason to keep it like this now, if higher versions are available we should target those and fix any emerging issues accordingly.

omichel:
There are some clang-format issues, could you please fix them?

ijnek:
> There are some [clang-format issues](https://github.com/cyberbotics/webots/runs/6157582840?check_suite_focus=true), could you please fix them?

Should be fixed up now

src/webots/control/WbLanguageTools.cpp
```diff
@@ -107,15 +108,19 @@ QString WbLanguageTools::pythonCommand(QString &shortVersion, const QString &com
     shortVersion = checkIfPythonCommandExist(pythonCommand, env, true);
   else {
     if (pythonCommand == "python" || pythonCommand == "python3") {
-      pythonCommand = findWorkingPythonPath("3.8", env, false);
-      shortVersion = "38";
+      pythonCommand = findWorkingPythonPath("3.10", env, false);
+      shortVersion = "310";
```
omichel, line 112: We should not assign the `shortVersion` unless we know the python command is valid (e.g., `!= "!"`) otherwise, it will be assigned several times (see my suggestion below).

omichel (changes requested):
Thank you, we are almost there.
I still have a couple of minor suggestions.

src/webots/control/WbLanguageTools.cpp
ijnek: This change sounds good, done.

omichel (approved):
Great. Thank you!

ad-daniel (approved):
Thank you!

ijnek:
@omichel and @ad-daniel thanks for both of your reviews

BenjaminDeleze:
It seems like this PR broke the package creation: https://github.com/cyberbotics/webots/runs/6173005367?check_suite_focus=true

ijnek:
It does seem related to this PR, but I'm not sure why it didn't fail in this PR...?

ad-daniel:
By default only minimal tests are performed, need to add the `test suite` label for the full test to be performed, or like in this case, for the distribution to be built. I did however run it locally beforehand, but I forgot in my machine I have 3.10 from custom PPA so it worked fine. Not a big issue, it can be addressed it in a separate PR. Basically the issue is that 3.10 is not available from official repos in ubuntu 18/20 so it fails in the CI

PR #4475, opened 2022-04-25
Fix list of USE

### LINKED ISSUE #4458: USE of group/transform with geometries
It seems that `Group` or `Transform` bounding objects that contains directly a `Geometry` (whiteout a `Shape`) causes trouble with the DEF/USE mechanism.
The `Group` or `Transform` does not appear in the propositions when I try to add a USE bounding object.
However if I defined it directly in the world/PROTO file it works well (as this is the case for the Aibo robot for example).

Illustration:
```
Solid {
  boundingObject DEF TEST Group {
    children [
      Sphere {
        radius 0.1
      }
    ]
  }
}
Solid {
  translation 0 0.2 0
  name "solid(1)"
  boundingObject USE TEST
}
```
This works well.
But if I delete the bounding object of solid(1) in Webots, I cannot add it again.

BenjaminDeleze:
Fix #4458 

The check was sometimes believing that we were trying to insert a DEF node inside a non-bounding object whereas it was in fact a bounding object.


ad-daniel (changes requested):
Appears to work nicely, thank you!

ad-daniel (approved):
Thank you

PR #4478, opened 2022-04-26
Fix build due to missing python 3.10

ad-daniel:
**Description**
Ditto.

EDIT: Took the opportunity to install python from official website rather than our own and to use the macos-11 instead of 10.15 in the CI.


omichel (approved):
Maybe we could also upgrade to macos-11 in test_sources.yml?

omichel (approved):
Thank you.

PR #4480, opened 2022-04-26
Upgrade checkout and attempt fixing discord bot

ad-daniel:
**Description**

Apparently the bot failed again, this time it couldn't push because supposedly its branch was behind the remote one, which doesn't make much sense since it's a fresh clone. There has been an update to the github action "checkout" recently, I haven't found any similar issue on their github but shouldn't hurt to upgrade either. On top of it the PR forces the branch to be recreated. 

It theory it should be sufficient to make a pull before creating the commit, but it really shouldn't be necessary to do so, hence it seems the best approach (or at least, the only I can think of, but I'm open to ideas)

### DIFF
--- a/.github/workflows/discord_updater.yml
+++ b/.github/workflows/discord_updater.yml
@@ -8,10 +8,10 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
       with:
         fetch-depth: 0
-        ref: released 
+        ref: released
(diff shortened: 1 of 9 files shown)


ad-daniel:
Since the bot failed again with the same safe.directory error, and that [checkout@v3](https://github.com/actions/checkout/blob/main/CHANGELOG.md) supposedly addresses it, it might be sufficient to upgrade it to also fix the bot

omichel (approved):
Thank you.

omichel:
🤞🏻

PR #4485, opened 2022-04-28
Don't generate css if exporting metafile

ad-daniel:
**Description**
When launching webots with `--enable-x3d-meta-file-export` (typically to re-generated web component) we also generate the css which is not really necessary for that task, and it needs to be manually filtered out before committing.

I've also renamed a function as it makes more sense given what it does.


BenjaminDeleze (approved):
Thank you!

Issue #4488, opened 2022-04-29
Issues when converting to base node


ad-daniel:
**Describe the Bug**
When converting from PROTO to basenode, some issues occur:

1. Panda: `CadShape` urls are empty
2. Scara demo, a number of `ERROR: Duplicate field value..` are printed

PR #4491, opened 2022-04-29
Fix basenode conversion

ad-daniel:
**Description**

Fixes: https://github.com/cyberbotics/webots/issues/4488

- [x] Panda
- [x] Scara


omichel (dismissed):
Thank you.

ad-daniel:
Concerning the warning when converting a PointLight from PROTO to basenode, it stemmed from the fact that some fields were written twice, one time by `WbLight::exportNodeFields` and then manually by
```
findField("radius", true)->write(writer);
...
```

The change however broke the streaming of PointLights, and the reason is that the [exporting](https://github.com/cyberbotics/webots/blob/f1391adf0842d8d488324505bf224edb99d0a78e/src/webots/nodes/WbLight.cpp#L228-L239) at the level of WbLight propagated down to `WbBaseNode::exportNodeFields` only for webots writers, not x3d (probably why the the manual writing was added to WbPointLight in the first place).

In practice, it seems that just letting WbBaseNode do the exporting is sufficient as it exports all fields correctly, both webots and x3d.

BenjaminDeleze (approved):
Seems good to me

PR #4493, opened 2022-04-29
Fix labels in webotsJS

### LINKED ISSUE #3279: Color boxes are painted over text
The replay from html of some games has an appearance where the text seems to be *hidden* behind the colored boxes of the layout (with transparency involved). The text should always be written above the boxes.

Visual example of the bug:
robocup_layout_behind

See original issue [here](https://github.com/RoboCup-Humanoid-TC/webots/issues/277).

BenjaminDeleze:
fix #3279


ad-daniel (approved):
Thank you

PR #4495, opened 2022-05-02
Removed definition of fields inherited from the Transform node

omichel:
As reported by @Justin-Fisher in https://github.com/cyberbotics/webots/pull/3801#issuecomment-1112004052


stefaniapedrazzi (approved):
Thank you

PR #4496, opened 2022-05-02
Cleanup floating windows

ThomasOliverKimble:
**Description**

Quick cleanup:
* Make the `interact` function more understandable (`id` -> `direction`) as discussed
* Force cursor and remove pointer events type during interactions (this helps with general usability, for example the window can no longer be closed when being dragged)


omichel (approved):
Thank you.

Issue #1875, opened 2020-07-06
x3d streaming issues with template regenerated PROTO nodes 


stefaniapedrazzi:
x3d streaming of template regenerated PROTO nodes only partially works.
I'm testing with a PROTO node with a slot template parameter  and changing the appearance of the slot descendant node.

But only the first change is correctly visualized in the web interface.
Subsequent changes are not taken into account.

Steps to reproduce with the following example:
1. start Webots with `--stream` and connect with the streaming viewer
2. move the Transform node in the slot parameter -> cube moves in the web interface too
3. trigger Container PROTO regeneration by changing the Parquetry type
4. move again the Transform node
-> the node doesn't move anymore

Example:
World file:
```
WorldInfo {
}
Viewpoint {
  orientation -0.8941063369853983 0.4330888056765547 0.1140523763900121 0.5728726641946077
  position 0.5636279862677915 1.250913320255655 1.9416637974495892
}
Background {
  skyColor [
    0.15 0.45 1
  ]
}
Container {
  slot [
    Transform {
      translation 0 0.14 0
      children [
        Shape {
          appearance Parquetry {
            type "chequered"
          }
          geometry Box {
            size 0.1 0.1 0.1
          }
        }
      ]
    }
  ]
}
```

Container.proto:
```
#VRML_SIM R2020b utf8
# license: Copyright Cyberbotics Ltd. Licensed for use only with Webots.
# license url: https://cyberbotics.com/webots_assets_license
# tags: static
# Configurable standard flat floor.

PROTO Container [
  field MFNode slot      []

]
{
  %{
    io.stderr:write("'slot' template field.\n")
  }%
  Solid {
    children IS slot
  }
}
```

PR #4497, opened 2022-05-02
WebotsJS: Fix PROTO regeneration and update bugs

BenjaminDeleze:
**Related Issues**
This pull-request fixes issue #1875 

**Tasks**
  - [x] Fix crash when replacing an appearance.
  - [x] Fix deleted item ID when inside PROTO.
  - [x] Fix update after PROTO regeneration.

Both the crash and the problem of updates after PROTO regeneration happened because `createWrenObject` was called twice on some objects.

Some objects were not deleted in WebotsJS because the id sent along the `delete:` was not matching the id sent at the world creation.

### DIFF
--- a/resources/web/wwi/Parser.js
+++ b/resources/web/wwi/Parser.js
@@ -715,8 +715,6 @@ export default class Parser {
       geometry = this._parseElevationGrid(node, id);
     else if (node.tagName === 'PointSet')
       geometry = this._parsePointSet(node, id);
-    else
-      console.error('Not a recognized geometry: ' + node.tagName);
 
     if (typeof parentId !== 'undefined' && typeof geometry !== 'undefined')
       geometry.parent = parentId;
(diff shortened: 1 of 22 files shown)


stefaniapedrazzi:
Changes seem fine.
I'm now testing them but probably I will complete the tests tomorrow morning.

stefaniapedrazzi (approved):
Thank you!
I confirm the original bug is fixed.

I noted that changes to the geometries (size, radius, height, etc.) are not propagated to the web scene (same behavior as in the develop branch)
I don't remember what is the status of this.

BenjaminDeleze:
> I noted that changes to the geometries (size, radius, height, etc.) are not propagated to the web scene (same behavior as in the develop branch)
> I don't remember what is the status of this.

This was not implemented in the previous version and is not implemented in this one either.
Only update of position, orientation and colors are supported as they are the most frequent ones.
However, it would be quite quick and easy to add support for other changes and I can do it if we need some specific parameters.

PR #4498, opened 2022-05-02
Clean-up creation of QProcess in the WbController class

omichel:
This PR cleans-up the `WbController` class.
The `QProcess` doesn't need to be created in case of extern controller, so it shouldn't be created in the constructor of `WbController`, but only in the `start` method of this class which is not called in case of extern controller.

This PR aims at the develop branch because I will need it in #4344 and I see little interest seeing this back-ported to the master branch.


BenjaminDeleze (approved):
Seems good to me

PR #4499, opened 2022-05-02
Fix bug when clicking on a CadShape

BenjaminDeleze:
ditto


ad-daniel (approved):
Thank you, but I have a doubt concerning the behavior itself. When you select once it's supposed to show all bounding objects and if you click two times it's supposed to only show the "selected" one, correct? Currently when clicking twice it shows the bounding object of a number of segments, not just one (tested on a scene). Is it because we're currently cheating and treating CadShape as a group of Shapes? Meaning it will disappear once we implement the tessellator on webotsjs?

BenjaminDeleze:
> Currently when clicking twice it shows the bounding object of a number of segments, not just one (tested on a scene)

It is normal, it will show the BO of the selected solid AND its descendants

PR #4503, opened 2022-05-03
Add a function to change the ambiant occlusion level

BenjaminDeleze:
ditto

Example: https://cyberbotics.com/animations/panda2/panda.html


ad-daniel (approved):
Thank you, much better!

PR #4505, opened 2022-05-03
Fix warning message during loading

BenjaminDeleze:
- [x] Fix `onerror` to display the error message only when the function is called
- [x] Add a new error message when the response status is `404`

### DIFF
--- a/resources/web/wwi/X3dScene.js
+++ b/resources/web/wwi/X3dScene.js
@@ -111,9 +111,15 @@ export default class X3dScene {
         const loader = new Parser(prefix);
         await loader.parse(xmlhttp.responseText, renderer);
         onLoad();
+      } else if (xmlhttp.status === 404) {
+        if (document.getElementById('webotsProgressMessage'))
+          document.getElementById('webotsProgressMessage').innerHTML = 'File not found';
       }
     };
-    xmlhttp.onerror = document.getElementById('webotsProgressMessage').innerHTML = 'File not found.';
+    xmlhttp.onerror = () => {
+      if (document.getElementById('webotsProgressMessage'))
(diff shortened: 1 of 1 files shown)


resources/web/wwi/X3dScene.js
```diff
@@ -111,9 +111,15 @@ export default class X3dScene {
...
         onLoad();
+      } else if (xmlhttp.status === 404) {
+        if (document.getElementById('webotsProgressMessage'))
+          document.getElementById('webotsProgressMessage').innerHTML = 'File not found';
       }
     };
-    xmlhttp.onerror = document.getElementById('webotsProgressMessage').innerHTML = 'File not found.';
+    xmlhttp.onerror = () => {
+      if (document.getElementById('webotsProgressMessage'))
+        document.getElementById('webotsProgressMessage').innerHTML = 'An error occured during the loading.';
```
omichel, line 121: Don't we have an idea about this error? Is there an error string somewhere. Otherwise we should display:
```suggestion
        document.getElementById('webotsProgressMessage').innerHTML = 'An unknown error occurred during the loading.';
```

Note: occurred has two "r".

resources/web/wwi/X3dScene.js
```diff
@@ -111,9 +111,15 @@ export default class X3dScene {
         const loader = new Parser(prefix);
         await loader.parse(xmlhttp.responseText, renderer);
         onLoad();
+      } else if (xmlhttp.status === 404) {
+        if (document.getElementById('webotsProgressMessage'))
+          document.getElementById('webotsProgressMessage').innerHTML = 'File not found';
```
omichel, line 116: Can we display which file is not found? That would be more helpful.

omichel (approved):
Thank you.

PR #4507, opened 2022-05-04
Fix streaming when lights are involved

ad-daniel:
**Description**

It looks like the cleanup done in https://github.com/cyberbotics/webots/pull/4491 didn't go far enough, there's duplicate writers for the other lights as well


stefaniapedrazzi (approved):
I confirm the streaming is now working again.

PR #4512, opened 2022-05-05
Fix sphere position in the documentation viewer

BenjaminDeleze:
- [ ] Fix position when a `position sensor` is selected (e.g. in scara:`base_arm_position` or panda: `panda_joint5_sensor`).
- [ ] Fix position when a `motor` is selected (e.g in scara:`base_arm_motor` or `panda: panda_joint5`).
- [ ] Fix position when a `led` is selected (e.g in scara: `epson_led`).


BenjaminDeleze:
I regenerated the `.x3d` to include the patch of #4517 and to also correct the rotation of the viewpoint for Tiagos

ad-daniel:
Overall it works quite well, the only "deterioration" compared to before I noticed is with sliders, for instance the fingers of the youbot (or any other similar gripper):

released: https://cyberbotics.com/doc/guide/youbot
this branch: https://cyberbotics.com/doc/guide/youbot?version=fix-viewer-sphere-position

is it because the SliderJoint doesn't actually have an anchor?

BenjaminDeleze:
> is it because the SliderJoint doesn't actually have an anchor?

Yes I think.

There is two solutions:

- Either I put it in the parent node (as it is now) and it will look kind of wrong but not too much.
- Or I put it in the children node (like before) and it will look right some time (e.g Youbot) and completely wrong in other (e.g Ned)

BenjaminDeleze:
I fixed the Linear Motors and the Ned

Issue #4508, opened 2022-05-04
Obj Cadshape does not import mtl when mesh is an URL


BenjaminDeleze:
The bug occurs when having an web url (like `https://raw.githubusercontent.com/cyberbotics/webots/develop/tests/rendering/worlds/meshes/colored_cube.obj`) in a CadShape. 

It seems that the obj file is loaded correctly but not the mtl file as the mesh will appear but with a grey texture.

No errors are printed in the console.

![cadshape_node](https://user-images.githubusercontent.com/25938827/166663109-8be0bd03-8577-4b68-a1dc-7aae162d89c3.png)

Note: The dae works fine

PR #4513, opened 2022-05-05
Fix loading of remote wavefront files in `CadShape` (that reference material files)

ad-daniel:
**Description**

Fixes: https://github.com/cyberbotics/webots/issues/4508

Collada files embed material information within the dae file itself, and may also reference external textures. When loading these files, the structure generated by assimp contains these references and shallow `ImageTexture` nodes can be instantiated on the fly after the fact, while in the process of building the meshes/appearances. In other words, when these ImageTexture nodes are instantiated, we can download and load the textures by manufacturing their url based on the url of the "dae" file itself.

In wavefront files however things are not so simple because the material properties are themselves external (and the materials in turn can reference external textures themselves), and by the time we feed the obj file to assimp, the corresponding material file needs to be accessible or the resulting assimp structure will be incomplete. This is not a problem for locally defined assets since everything is available anyway, but for remote urls it is an issue since we need to download the `.obj` file before even knowing it requires any `.mtl` file.
As such, the only solution is to:
- download the obj asset
- open it and download all referenced materials
- before feeding the obj  to assimp, replace all occurrences of `.mtl` within it with the path that points to the cached material file

Wavefront files can also reference external textures within the `.mtl` file, but the node is capable of retrieving them just fine since it can manufacture their url from the url of the mtl file itself (as it does for collada).

- [x] proof of concept
- [x] cleanup/robustness


src/webots/nodes/WbCadShape.cpp
```diff
@@ -165,10 +226,83 @@ void WbCadShape::updateUrl() {
...
+
+    QStringList lines = content.split('\n', Qt::SkipEmptyParts);
+    foreach (QString line, lines) {
+      const QString cleanLine = line.trimmed();
+      if (!cleanLine.startsWith("mtllib"))
+        continue;
+
+      materials = cleanLine.split(' ', Qt::SkipEmptyParts);
+      materials.removeFirst();  // first is "mtllib"
+      break;                    // only one occurrence of mtllib is allowed, so as soon as one is found the search can stop
```
BenjaminDeleze, line 296: Is this a restriction that you want to introduce?
Because it is possible to have several `.mtl` in a `.obj` (https://en.wikipedia.org/wiki/Wavefront_.obj_file#Reference_materials)
I tested to have a `.obj` with two `mtllib` call with different `.mtl` file and I could load it in Blender and also in Webots (with local file)

src/webots/nodes/WbCadShape.cpp
```diff
@@ -165,10 +226,83 @@ void WbCadShape::updateUrl() {
...
+bool WbCadShape::areMaterialAssetsAvailable(const QString &url) {
+  QStringList rawMaterials = objMaterialList(url);  // note: 'dae' files will generate an empty list
+  foreach (QString material, rawMaterials) {
+    if (!WbNetwork::instance()->isCached(generateMaterialUrl(material, url)))
+      return false;
+  }
+  return true;
+}
+
+void WbCadShape::generateMaterialMap(const QString &url) {
```
BenjaminDeleze, line 261: It seems like this function is not used?

BenjaminDeleze (changes requested):
Works well, I only have two remarks

src/webots/nodes/WbCadShape.cpp
ad-daniel: Then I must have misunderstood the [specs](http://paulbourke.net/dataformats/obj/):

> When you assign a material library using the Model program, only one map library per .obj file is allowed. You can assign multiple libraries using a text editor.

Something like `mtllib filename1.mtl filename2.mtl` is supported already. You tested something like:
```
mtllib filename1.mtl
mtllib filename2.mtl
```
and blender didn't complain? For instance if both files define the same materials but they are of different colors, which will it pick?

Either way I can add support for that second case, in the off-chance somebody has it

src/webots/nodes/WbCadShape.cpp
ad-daniel: Ops, indeed forgot to remove it after refactoring

src/webots/nodes/WbCadShape.cpp
BenjaminDeleze: and blender didn't complain?
Exactly and Webots (so assimp) handle it well as well (locally).
If you want to test it: https://github.com/BenjaminDeleze/sandbox/tree/master/meshes
I did not test to define two materials with the same name

ad-daniel: Then we should support it, thanks!

ad-daniel:
The other fix is related to a buggy behavior when dealing with remote urls. Under certain conditions (clear cache > set remote url > set invalid remote url > set valid remote url), despite putting a valid url at the end, it didn't download it (WbMesh was affected by it too)

BenjaminDeleze (approved):
Thank you!

PR #4517, opened 2022-05-06
Restore export of shadows

BenjaminDeleze:
Fix a bug introduced in #4491 
As `castShadow` is a `field` and not a `vrmlField` it is not exported by the `WbNode` x3d exporter so we need to specify it.


ad-daniel (approved):
Thank you

PR #4518, opened 2022-05-06
Remote extern controllers through TCP

ygoumaz:
**Description**
Currently, extern controllers can be used on a local machine with Webots. With multiple Webots instances, an environment variable WEBOTS_PID containing the PID of the Webots process allows to differentiate which instance we want to target. Same for the robots: with multiple robots with extern controllers, an environment variable WEBOTS_ROBOT_NAME can be set to target a specific robot.

The goal of this PR is to extend the extern controllers to remote communication. Webots could start on a machine and a controller on another machine. A new environment variable WEBOTS_CONTROLLER_URL replaces the two old ones to specify if we are on a local or remote machine and which robot we target.

URL format is as follows: `<protocol>://<machine_ip>:<webots_port>/<robot_name>`. Three protocols should be available:
- `ipc` for local extern controllers (already implemented in PR #4344).
- `tcp` for remote extern controllers (this PR implementation).
- `udp` will be implemented in a future PR.

The webots_port allows to differentiate Webots instances, while the robot_name allows to target a specific robot.

**Tasks**
- [x] Add new TCP client API on libController side
- [x] Check for WEBOTS_CONTROLLER_URL variable at robot initialization
- [x] New remote and local scheduler in libController
- [x] Get controller connection via TCP on Webots side
- [x] Check for targeted robot and assign TCP socket to corresponding controller
- [x] Update WbController behavior in case of remote connection
- [x] Create new WbDataStream class for request transfer between Webots and libController
- [x] Serialize and send AbstractCamera data in chunks to remote controller through TCP
- [x] Extend to other rendering devices (camera, lidar and range finder)
- [x] Receive serialized data from AbstractCamera devices on controller side
- [x] Allow controller re-connection if closed on controller side
- [x] Allow automatic disconnection of controller when closing Webots
- [x] Add client whitelist on Webots side
- [x] Perform tests on 2 distinct machines
- [x] Documentation / changelog
- [x] Final tests / Clean-up

**Documentation**
https://cyberbotics.com/doc/guide/running-extern-robot-controllers?version=feature-enable-remote-extern-controllers

**Additional context**
This PR continues the work accomplished in PR #4344, where the local extern controller are implemented using the WEBOTS_CONTROLLER_URL variable and the `ipc` protocol.


src/webots/nodes/utils/WbDataStream.cpp
```diff
@@ -0,0 +1,75 @@
...
+}
+
+WbDataStream &WbDataStream::operator<<(qint8 n) {
+  return (WbDataStream &)append(n);
+}
+
+WbDataStream &WbDataStream::operator<<(qint16 n) {
+  for (int i = 0; i != sizeof(n); ++i) {
+    append((char)((n & (0xFF << (i * 8))) >> (i * 8)));
+  }
```
omichel, line 30:
```suggestion
  for (int i = 0; i != sizeof(n); ++i)
    append((char)((n & (0xFF << (i * 8))) >> (i * 8)));
```

src/webots/nodes/utils/WbDataStream.cpp
```diff
@@ -0,0 +1,75 @@
...
+  return *this;
+}
+
+WbDataStream &WbDataStream::operator<<(qint32 n) {
+  for (int i = 0; i != sizeof(n); ++i) {
+    append((char)((n & (0xFF << (i * 8))) >> (i * 8)));
+  }
```
omichel, line 37:
```suggestion
  for (int i = 0; i != sizeof(n); ++i)
    append((char)((n & (0xFF << (i * 8))) >> (i * 8)));
```

src/webots/nodes/utils/WbDataStream.cpp
```diff
@@ -0,0 +1,75 @@
...
+  QDataStream ds((QByteArray) * this);
+  ds.setByteOrder(QDataStream::LittleEndian);
+  ds.device()->seek(0);
+  ds >> nb_chunks;
+
+  WbDataStream new_nb_chunks;
+  unsigned short new_nb_chunks_value = nb_chunks + n;
+  new_nb_chunks << new_nb_chunks_value;
+  replace(0, (int)sizeof(unsigned short), new_nb_chunks);
+}
```
omichel, line 75:
```suggestion
}

```

omichel (changes requested):
Good job. This is a first code review. I still need to test it.

src/controller/c/tcp_client.h
```diff
@@ -27,18 +27,14 @@
...
-
-typedef struct _TcpClient TcpClient;
-
-TcpClient *tcp_client_new(const char *host, int port);
-int tcp_client_open();
-int tcp_client_connect(TcpClient *c, const char *host, int port);
-bool tcp_client_send(TcpClient *c, const char *buffer, int size);
-int tcp_client_receive(TcpClient *c, char *buffer, int size);
-void tcp_client_close(TcpClient *c);
+typedef int TcpClient;
```
omichel, line 30: I believe it makes the code less opaque to remove this `typedef` and use `int` instead for the file descriptor of the socket, which is something very standard.
```suggestion
```

src/controller/c/tcp_client.c
```diff
@@ -109,18 +103,17 @@ bool tcp_client_send(TcpClient *c, const char *buffer, int size) {
...
+int tcp_client_receive(TcpClient c, char *buffer, int size) {
+  return recv(c, buffer, size, 0);
 }
 
-void tcp_client_close(TcpClient *c) {
+void tcp_client_close(TcpClient c) {
 #ifdef _WIN32
-  closesocket(c->fd);
+  closesocket(c);
   free(c);
```
omichel, line 113: That seems wrong and is causing a compilation warning:
```suggestion
```

omichel (approved):
Thank you.
I tested it on Windows and it works nicely.
I would be nice however to have a second review as this PR is quite big.

BenjaminDeleze (changes requested):
Seems good! only some minors things:
It is strange to me that if we have two robots the behavior is different depending if the connection is remote or not:
Local: if name is not specified, refuse the connection
Remote: if name is not specified, connect to the first one

otherwise I ran most of the guided tour worlds without any problems!

ygoumaz:
The behavior is indeed not consistent with local extern controllers. I have updated the code to refuse connection if the robot-name is unset and multiple extern controllers are defined in the simulation. The documentation has also been updated.

docs/guide/running-extern-robot-controllers.md
```diff
@@ -170,17 +172,28 @@ You can repeat this for the other controllers, e.g., set a different value to th
...
+`<port>` is the TCP port (defined in the `--port` command line option) of the Webots instance to which you want to connect your controller.
+Finally, `<robot_name>` is the name of the robot to which you want to connect your controller.
+Note that the URL path `/<robot_name>` can be left blank and the controller will connect to the only robot with an `<extern>` controller.
+
 ### Notes about the WEBOTS\_CONTROLLER\_URL Environment Variable
 
-If the `WEBOTS_CONTROLLER_URL` is not set, the controller will connect to the only extern robot of a Webots instance.
+`WEBOTS_CONTROLLER_URL` can be left unset and the controller will connect to the only extern robot of a local Webots instance.
+If this instance has several extern robots, Webots will refuse the connection and the controller will quit, displaying an error.
+For remote connections, it is mandatory to specify the complete URL for the controller to find the Webots instance, but the `/<robot_name` URL path can be left blank if only one robot has an `<extern>` controller.
```
omichel, line 196:
```suggestion
For remote connections, it is mandatory to specify the complete URL for the controller to find the Webots instance, but the `/<robot_name>` URL path can be left blank if only one robot has an `<extern>` controller.
```

omichel (dismissed):
Thank you.

BenjaminDeleze (dismissed):
Thank you

omichel (approved):
Looks good to me.

BenjaminDeleze (approved):
Thank you

PR #4519, opened 2022-05-06
Enhancement assimp webotsjs

BenjaminDeleze:
The goal of this PR is to support `Mesh` and `ColladaShape` in WebotsJS as in Webots.
For now, Webots converts the `WbMesh`/`WbColladaShape` in `WbIndexedFaceSet`.

- [x] Compile Assimp with Emscripten and build some wrappers to be able to use it in JS (made thanks to https://github.com/cyberbotics/assimpjs).
- [x] Modify the export of
  - [x]  `WbMesh`
  - [x]   `WbColladaShape`
- [x] In WebotsJS, implement the parsing and rendering of:
  - [x]  `WbMesh`
  - [x] `WbColladaShape`
- [x] Regenerate docs.
- [x] Rename `WbPBRAppearance` in `WbPbrAppearance` to stick with coding style.
- [x] Modify assimp compilation instructions: https://github.com/cyberbotics/webots/wiki/Assimp-compilation
- [x] Put the function to load images in a new file: `image_loader.js` to avoid circular dependencies between `Parser.js` and `WbCadShape.js`


ad-daniel (changes requested):
Thank you, overall works nicely, the issues I've encountered are:

- the pickable state appears no to be active on the webotsjs side. If you load the car or the elf I sent you last time it seems it's impossible to rotate around them
- if the mesh file contains spaces, ex: `/home/daniel/Desktop/3D Models/cads/car/car.dae` it doesn't appear to be able to load it on the webotsjs side. Maybe it's due to the filtering when splitting the url?
- is the stream supposed to respond to field changes? When I change the url for instance, isn't it supposed to load the new object? If I cut and paste the node it does so, but not if I just change the url.
- Lastly, just a question: I noticed that for the docs of the robotino3 you copied the `head_white_cover.obj` file, can't we refer to the one in the protos instead? Otherwise we'll end up doubling all meshes

resources/web/wwi/dependencies/assimpjs.js
```diff
@@ -0,0 +1,21 @@
+
+var assimpjs = (() => {
+  var _scriptDir = typeof document !== 'undefined' && document.currentScript ? document.currentScript.src : undefined;
+  if (typeof __filename !== 'undefined') _scriptDir = _scriptDir || __filename;
+  return (
+function(assimpjs) {
```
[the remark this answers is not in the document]
BenjaminDeleze: yes, it is a "glue" file generated automatically by Emscripten to load the `.wasm`

BenjaminDeleze:
> * is the stream supposed to respond to field changes? When I change the url for instance, isn't it supposed to load the new object? If I cut and paste the node it does so, but not if I just change the url.

I don't think so as it was not the case before. (only some fields supports updates: translation, rotation, some colors and lights)

resources/web/wwi/nodes/WbCadShape.js
```diff
@@ -1,75 +1,324 @@
...
+    else if (properties.get(10))
+      occlusionMap = this._createImageTexture(properties.get(10));
+
+    let emissiveColorMap;
+    if (properties.get(14))
+      emissiveColorMap = this._createImageTexture(properties.get(14));
+    else if (properties.get(4))
+      emissiveColorMap = this._createImageTexture(properties.get(4));
+
+    return new WbPbrAppearance(getAnId(), baseColor, baseColorMap, transparency, roughness, roughnessMap, metalness, metalnessMap, iblStrength, normalMap, normalMapFactor, occlusionMap, occlusionMapStrength, emissiveColor, emissiveColorMap, emissiveIntensity, undefined);
```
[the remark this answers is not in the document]
BenjaminDeleze: Until now not (at least in practice). After some discussion, it was decided to put it at 128 (as we do for other languages), but I will adapt all our JS code base to this rule in another PR.

src/webots/nodes/utils/WbUrl.cpp
```diff
@@ -195,8 +200,11 @@ QString WbUrl::exportTexture(const WbNode *node, const QString &url, const QStri
           return newRelativePath;
         }
       }
+      if (isTexture)
+        node->warn(QObject::tr("Texture export fails, because too much textures are sharing the same name: %1.").arg(url));
+      else
+        node->warn(QObject::tr("Mesh export fails, because too much meshes are sharing the same name: %1.").arg(url));
```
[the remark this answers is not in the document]
BenjaminDeleze: I tested without and it is necessary

BenjaminDeleze:
> * the pickable state appears no to be active on the webotsjs side. If you load the car or the elf I sent you last time it seems it's impossible to rotate around them

I think you discovered a more general bug here. It seems that webotsJS does not handle well the pickable state when we are too far from an object. It is more visible with the car and elf because the are big, so we put the viewpoint further. I will handle it in another issue as it is not related to this PR. Well spotted!

>* if the mesh file contains spaces, ex: /home/daniel/Desktop/3D Models/cads/car/car.dae it doesn't appear to be able to load it on the webotsjs side. Maybe it's due to the filtering when splitting the url?

It appeared to be a problem not only with `CadShape` but also with more general mesh and images. I believe I fixed the problem.

>* Lastly, just a question: I noticed that for the docs of the robotino3 you copied the head_white_cover.obj file, can't we refer to the one in the protos instead? Otherwise we'll end up doubling all meshesnoticed that for the docs of the robotino3 you copied the head_white_cover.obj file, can't we refer to the one in the protos instead? Otherwise we'll end up doubling all meshes

It was a problem of Robotino that has an wrongly defined url inside its Proto, it is now fixed

ad-daniel (approved):
Looks good, thank you

src/webots/nodes/WbCadShape.cpp
```diff
@@ -578,107 +583,59 @@ const WbVector3 WbCadShape::absoluteScale() const {
...
+  for (int i = 0; i < mUrl->size(); ++i) {
+    if (WbUrl::isLocalUrl(mUrl->value()[i])) {
+      QString newUrl = mUrl->value()[i];
+      dynamic_cast<WbMFString *>(urlFieldCopy.value())
+        ->setItem(i, newUrl.replace("webots://", "https://raw.githubusercontent.com/" + WbApplicationInfo::repo() + "/" +
+                                                   WbApplicationInfo::branch() + "/"));
+    } else if (WbUrl::isWeb(mUrl->value()[i]))
+      continue;
+    else {
+      QString meshPath(WbUrl::computePath(this, "url", mUrl, i));
```
omichel, line 605:
```suggestion
      const QString meshPath(WbUrl::computePath(this, "url", mUrl, i));
```

src/webots/nodes/WbMesh.cpp
```diff
@@ -387,3 +389,46 @@ void WbMesh::updateMaterialIndex() {
...
+      QString newUrl = mUrl->value()[i];
+      dynamic_cast<WbMFString *>(urlFieldCopy.value())
+        ->setItem(i, newUrl.replace("webots://", "https://raw.githubusercontent.com/" + WbApplicationInfo::repo() + "/" +
+                                                   WbApplicationInfo::branch() + "/"));
+    } else if (WbUrl::isWeb(mUrl->value()[i]))
+      continue;
+    else {
+      QString meshPath(WbUrl::computePath(this, "url", mUrl, i));
+      if (writer.isWritingToFile()) {
+        QString newUrl = WbUrl::exportMesh(this, mUrl, i, writer);
```
omichel, line 414:
```suggestion
        const QString newUrl = WbUrl::exportMesh(this, mUrl, i, writer);
```

src/webots/nodes/WbMesh.cpp
```diff
@@ -387,3 +389,46 @@ void WbMesh::updateMaterialIndex() {
...
+
+  findField("ccw", true)->write(writer);
+  findField("materialIndex", -1)->write(writer);
+  if (!mName->value().isEmpty()) {
+    QString name_dirty = mName->value();
+    name_dirty.replace("\'", "&apos;", Qt::CaseInsensitive);
+    name_dirty.replace("\"", "&quot;", Qt::CaseInsensitive);
+    name_dirty.replace(">", "&gt;", Qt::CaseInsensitive);
+    name_dirty.replace("<", "&lt;", Qt::CaseInsensitive);
+    writer << " name='" << name_dirty.replace("&", "&amp;", Qt::CaseInsensitive) << "'";
```
omichel, line 432: We should use camelCase for variables:
```suggestion
    QString dirtyName = mName->value();
    dirtyName.replace("\'", "&apos;", Qt::CaseInsensitive);
    dirtyName.replace("\"", "&quot;", Qt::CaseInsensitive);
    dirtyName.replace(">", "&gt;", Qt::CaseInsensitive);
    dirtyName.replace("<", "&lt;", Qt::CaseInsensitive);
    writer << " name='" << dirtyName.replace("&", "&amp;", Qt::CaseInsensitive) << "'";
```

src/webots/vrml/WbSFColor.hpp
```diff
@@ -27,7 +27,7 @@ class WbSFColor : public WbSingleValue {
   Q_OBJECT
 
 public:
-  explicit WbSFColor(double r, double b, double g) : mValue(WbRgb(r, g, b)) {}
+  explicit WbSFColor(double r, double g, double b) : mValue(WbRgb(r, g, b)) {}
```
omichel, line 30: 🤣

omichel (changes requested):

omichel (approved):
Thank you.

PR #4526, opened 2022-05-11
Improve Supervisor unlimited power description

stefaniapedrazzi:
We regularly receive questions about using the Supervisor to access device measurements from other robots.
I will try to add some additional specifications and change misleading text.


ad-daniel (approved):
Thank you

PR #4536, opened 2022-05-16
New protos

BenjaminDeleze:
- [x] Add a gaz canister proto
- [x] Change the flower pot protos to a smaller model


ad-daniel (changes requested):
Since for the physics of the canister we expose just a boolean, we should ensure the default mass being applied is appropriate. Is it the case already?
Other than that it looks good to me

BenjaminDeleze:
> Since for the physics of the canister we expose just a boolean, we should ensure the default mass being applied is appropriate. Is it the case already? Other than that it looks good to me

Well spotted, 289 kg is maybe a bit too big.

BenjaminDeleze:
I adjusted the size of the bottle slightly to better match a real bottle so that I could assign a correct mass to it.

ad-daniel (approved):
Thank you

PR #4502, opened 2022-05-03
Add support for Ubuntu 22.04

stefaniapedrazzi:
Add support of the latest Ubuntu 22.04.

* [x] Remove specific code for automatically installing optional dependencies and creating packages on Ubuntu 18.04.
* [x] Remove code to compile Python3.6 API (not available by default on Ubuntu 22.04)
* [x] On Ubuntu 22.04, download OpenSSL dependencies from Cyberbotics server
* [x] Create and upload new sumo package for Ubuntu 20.04
* [x] Update instructions to create packages on Ubuntu 22.04
* [x] Update Ubuntu 20.04 package creation to make them work on Ubuntu 22.04 too

Notes:
* ROS is currently not officially supported on Ubuntu 22.04 and it is not known if it will be in the future given that *ROS noetic* is the last release.

Errors fixed:
* ``Warning: Incompatible version of OpenSSL (built with OpenSSL 1.x, runtime version is >= 3.x)``
   -> Default SSL library is version 3 but we Qt6 is still built with version 1.
   We need to re-add the download of default Ubuntu 20.04 system OpenSSL libraries.
* Sumo interface: changed default version of `libgdal` and `libproj`

Tests:
* run Ubuntu 20.04 packages:
    * [x] Debian package
    * [x] tarball -> works after including openssl libraries and additional sumo libraries in the packages.
* local package creation
    * [x] tarball

PR #4548, opened 2022-05-23
Improve deprecation flags and double initialization

stefaniapedrazzi:
Address comments in #4502.

md5-related functions are deprecated in OpenSSL 3.
For the moment it seems that the new version of OpenSSL is only used on macOS and Ubuntu 22.04 so we need to ignore the warnings only on these platforms.


omichel (approved):
Thank you.

PR #4551, opened 2022-05-24
Adds EXTERNPROTO

ad-daniel:
**Description**

This PR externalizes PROTO nodes in world files, as it's a necessary step in order to later on be able to load PROTO files directly on the web.

Basic principles:
- all PROTO nodes used by a world need to be declared using  the syntax: `EXTERNPROTO "url/or/path.proto"` in the world file.
- inserted nodes (add-node) are automatically declared.
- PROTO that might be spawned during the simulation need to be declared as Ephemeral prior to it, the declaration is done by using the interface accordingly. Ephemeral PROTO are essentially EXTERNPROTO as any other (and saved as such), the only difference is just that at load time no instance of these nodes exists in the node tree.

The PR includes:
- a backwards compatibility mechanism ensures that worlds prior to R2022b can be loaded just the same. When the world is saved, the declarations are added.
- an export button was added to the add-node window. It allows to create a local copy of a PROTO (typically needed if you wish to have a local copy of an official webots PROTO since they're not longer shipped).
- PROTO and icons are cached (and both are added to the asset_cache zip)

To generate a local distribution:
```
python scripts/packaging/set_commit_and_date_in_version.py $(git log -1 --format='%H')
make -j12 distrib
install and unset WEBOTS_HOME
```
Suggestions for improvements (codewise or usability wise) are most welcome


ad-daniel:
I think the PR can already be reviewed. The update world test failed because a single asset (difference should be 21 not 22) is missing from the cache, it might take some time to figure out which and it doesn't really matter all that much

ad-daniel:
nevermind, the `generate_proto_list.py` script on windows is choking for some reason

src/webots/gui/main.cpp
```diff
@@ -213,5 +213,5 @@ int main(int argc, char *argv[]) {
 #endif
 #endif
 
-  return app.exec();
+  app.exec();
 }
```
omichel, line 217: It is strange we declare the main function as int and it doesn't return an exit status.

src/webots/nodes/utils/WbNodeOperations.cpp
```diff
@@ -170,6 +171,14 @@ WbNodeOperations::OperationResult WbNodeOperations::importNode(WbNode *parentNod
   bool isNodeRegenerated = false;
   int nodeIndex = itemIndex;
   foreach (WbNode *node, nodes) {
+    if (node->isProtoInstance()) {
+      // ensure the node was declared as EXTERNPROTO prior to import it
+      if (!WbProtoManager::instance()->isDeclaredExternProto(node->modelName())) {
+        WbLog::error(tr("In order to import the PROTO '%1', first it must be declared in the Ephemeral EXTERNPROTO list.")
```
omichel, line 177:
```suggestion
        WbLog::error(tr("In order to import the PROTO '%1', first it must be declared in the EXTERNPROTO list of the world file.")
```

src/webots/vrml/WbProtoModel.hpp
```diff
@@ -38,7 +38,7 @@ class WbProtoModel : public QObject {
 public:
   // create
   WbProtoModel(WbTokenizer *tokenizer, const QString &worldPath, const QString &fileName = QString(),
-               QStringList baseTypeList = QStringList());
+               const QString &externPath = QString(), QStringList baseTypeList = QStringList());
```
omichel, line 41:
```suggestion
               const QString &externPath = "", QStringList baseTypeList = QStringList());
```

Makefile
```diff
@@ -53,9 +53,9 @@ space := $(null) $(null)
 WEBOTS_HOME_PATH=$(subst $(space),\ ,$(strip $(subst \,/,$(WEBOTS_HOME))))
 include $(WEBOTS_HOME_PATH)/resources/Makefile.os.include
 
-.PHONY: clean cleanse debug distrib release webots_dependencies webots_target clean-docs docs
+.PHONY: clean cleanse debug distrib release webots_dependencies webots_target clean-docs docs proto_list
 
-release debug profile: docs webots_target
+release debug profile: docs webots_target proto_list
```
omichel, line 58: We consider the proto list is a dependency of Webots, e.g., Webots needs it to run, we should  move `proto_list` as a `webots_target` dependency on line 104.
```suggestion
release debug profile: docs webots_target
```

src/webots/vrml/WbParser.cpp
```diff
@@ -18,12 +18,14 @@
...
 #include "WbNodeModel.hpp"
-#include "WbProtoList.hpp"
+#include "WbProtoManager.hpp"
 #include "WbProtoModel.hpp"
 #include "WbProtoTemplateEngine.hpp"
 #include "WbToken.hpp"
 #include "WbTokenizer.hpp"
 
+#include <QtCore/QDir>
+#include <QtCore/QFileInfo>
```
omichel, line 28: It seems these include files are not needed.
```suggestion
```

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -0,0 +1,619 @@
...
+
+  WbParser parser(&tokenizer);
+  if (!parser.parseProtoInterface(worldPath))
+    return NULL;
+
+  tokenizer.rewind();
+  while (tokenizer.peekWord() == "EXTERNPROTO")  // consume all EXTERNPROTO tokens, if any
+    parser.skipExternProto();
+
+  bool prevInstantiateMode = WbNode::instantiateMode();
```
omichel, line 80:
```suggestion
  const bool prevInstantiateMode = WbNode::instantiateMode();
```

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -0,0 +1,619 @@
...
+    return model;
+  } catch (...) {
+    WbNode::setInstantiateMode(prevInstantiateMode);
+    return NULL;
+  }
+}
+
+void WbProtoManager::readModel(WbTokenizer *tokenizer, const QString &worldPath) {
+  WbProtoModel *model = NULL;
+  bool prevInstantiateMode = WbNode::instantiateMode();
```
omichel, line 94:
```suggestion
  const bool prevInstantiateMode = WbNode::instantiateMode();
```

omichel (changes requested):
I completed the source code review. I will now test the functionality.

omichel:
I create a package for Windows and tested it.
When opening the ure.wbt demo from the guided tour I got the following errors:
```
WARNING: UR5e (PROTO) > HingeJoint > Solid > Shape > PBRAppearance > ImageTexture: Cannot download 'https://raw.githubusercontent.com/cyberbotics/webots/714643ac2d882f53717dd9c3a13f1c55136e77d6/projects/robots/universal_robots/protos/UR5e/textures/normal_0.jpg', error code: 5: Operation canceled
WARNING: UR5e (PROTO) > HingeJoint > Solid > Shape > PBRAppearance > ImageTexture: Cannot download 'https://raw.githubusercontent.com/cyberbotics/webots/714643ac2d882f53717dd9c3a13f1c55136e77d6/projects/robots/universal_robots/protos/UR5e/textures/occlusion_0.jpg', error code: 5: Operation canceled
WARNING: UR5e (PROTO) > HingeJoint > Solid > HingeJoint > Solid > Shape > PBRAppearance > ImageTexture: Cannot download 'https://raw.githubusercontent.com/cyberbotics/webots/714643ac2d882f53717dd9c3a13f1c55136e77d6/projects/robots/universal_robots/protos/UR5e/textures/normal_1.jpg', error code: 5: Operation canceled
...
```
And the robot was not displayed.
I tried to reload it and it worked out...

src/webots/nodes/utils/WbNodeOperations.cpp
ad-daniel: Since this message is displayed when we attempt to import a non-declared node (using a supervisor), referring to "Ephemeral EXTERNPROTO" might be more clear for the user what it needs to do (since there's a button named just that on the interface)

omichel: Yes, I agree.

stefaniapedrazzi:
I'm also testing a little bit on Linux.

I have a couple of comments about the new "Ephemeral EXTERNPROTO" button functionality

1) Maybe the "Ephemeral EXTERNPROTO" button appearance could be improved.
At the moment, especially when using the dark themes I have the feeling it is quite hard to understand that it is a button and not a label.

    Maybe simply adding some padding left and right and setting a border color (in the dark themes) could improve the appearance.
Screenshot from 2022-06-16 09-56-03
Screenshot from 2022-06-16 09-56-32


2) Why did you choose to display the ephemeral EXTERNPROTO list in the field editor instead of opening a new dialog window?
This is causing some issues with the field editor itself because clicking on the button doesn't reset the selection, but there is no other ways than changing the current selection to restore the node content in the field editor.

ad-daniel:
Sorry I didn't see your post before:

> 1. Maybe the "Ephemeral EXTERNPROTO" button appearance could be improved.

Good idea, in the dark themes the buttons are indeed flat.

> 2. Why did you choose to display the ephemeral EXTERNPROTO list in the field editor instead of opening a new dialog window?

@omichel suggested it this way. The idea was that it was more intuitive for the user as changes to EXTERNPROTO pertain to the world as other fields do. The initial choice was even for it to show up as an item itself in the scene tree rather than a button but that proved problematic.
Concerning the issue you mention, I don't think I understood. What's the issue with the field editor? Prior to opening the panel I disable the existing field editor, is that it?

omichel:
>  Why not adding a field to the WorldInfo node?

That would be inconsistent with the content of the world file as the EXTERNPROTO statement should be ahead of the any node definition.

Clearing the current selection seems to be the best solution to me.

ad-daniel:
Does this convince you more:

a
b

Adding to the scene tree is problematic because contrary to all other items it doesn't have neither a field nor a node associated, so it ends up being even more of a hack. The only alternative would be to have  a second tree above this one with a single element in it, but that doesn't sound great either

docs/guide/the-scene-tree.md
```diff
@@ -41,8 +41,23 @@ This option is only displayed for Geometry nodes and nodes derived from [Transfo
...
+
+%figure "Ephemeral EXTERNPROTO"
+
+![ephemeral_externproto.png](images/ephemeral_externproto.png)
+
+%end
+
+Every world file must declare in its header the PROTO assets it relies upon using the `EXTERNPROTO` keyword, and providing an url or path towards each PROTO used.
+This list is normally compiled automatically by Webots when a world is saved.
+If however additional PROTO might be used during the execution of the simulation, typically by having a supervisor node inserting it into the world, then the user must declare these PROTO.
```
omichel, line 61:
```suggestion
If however additional PROTOs might be used during the execution of the simulation, typically by having a supervisor node inserting it into the world, then the user must declare these PROTOs.
```

omichel (approved):
Looks good to me.

PR #4561, opened 2022-05-25
Add a forklift proto

BenjaminDeleze:
It was sometime hard to "fill" the industrial environments because most of the existing factory objects are kind of small.


omichel (approved):
Thank you.

PR #4569, opened 2022-05-31
Fix support polygon access out of bounds

### LINKED ISSUE #4280: Automatic exit of webots on ubuntu
Hi, 
I was trying to use webots simulator to train and simulate quadruped robot by EPFL. While training, I activated optional render of polygon support. In the next run when I tried to accelerate the simulation, webots simulator crashed automatically with the following line on the console.
```/usr/local/bin/webots: line 90: 5289 Segmentation fault (core dumped) "$webots_home/bin/webots-bin" "$@"```

Upon close examination of the recent changes, I disable the rendering of support polygon and now webots runs smoothly.  I can share my work if needed.

stefaniapedrazzi:
Fix #4280.
If `numberOfContactPoints` value greater than `mSupportPolygon` size, `mSupportPolygon` was often accessed outside his bounds.
If `numberOfContactPoints` is lower than 32 (default size of the QVarLengthArray `mSupportPolygon`) nothing bad occurs, otherwise we may overwrite other WbSolid structures like the `mJointChildren` vector.


BenjaminDeleze (approved):
Seems good to me!

PR #4572, opened 2022-05-31
Provide patch instructions for Ubuntu-22.04

omichel:
https://cyberbotics.com/doc/guide/installation-procedure?version=doc-ubuntu22-instructions#installing-the-debian-package-directly


ad-daniel (approved):
Thank you

stefaniapedrazzi:
We should not forgot to remove this on the develop branch.

Also note that if you run Webots 2022a on Ubuntu 22.04 the sumo simulations won't work.

stefaniapedrazzi:
I also forgot to mention that Python simulation of R2022a don't work out of the box in Ubuntu 22.04 because R2022a has no support for Python 3.10 (the only Python version available on Ubuntu 22.04).

But if we will remove this note in develop, it is ok to do not specify it.

PR #4582, opened 2022-06-02
Fix local IDE

BenjaminDeleze:
With remote servers we have (for now) urls of the form: `https://server_address.com/2505`
but when testing on a local server it is something like: `http://localhost:2500`
So we have to handle both `/` and `:` separator

### DIFF
--- a/resources/web/wwi/FloatingIde.js
+++ b/resources/web/wwi/FloatingIde.js
@@ -7,9 +7,14 @@ export default class FloatingIDE extends FloatingWindow {
     this.headerText.innerHTML = 'Source Code Editor';
 
     this.frame.id = this.name + '-robot-window';
-    const elements = url.split('/').filter(element => element);
-    const portString = elements[elements.length - 1];
+    let elements = url.split('/').filter(element => element);
+    let portString = elements[elements.length - 1];
+    if (Number.isNaN(parseFloat(portString))) {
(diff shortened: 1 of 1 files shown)


resources/web/wwi/FloatingIde.js
```diff
@@ -7,9 +7,14 @@ export default class FloatingIDE extends FloatingWindow {
...
     this.frame.id = this.name + '-robot-window';
-    const elements = url.split('/').filter(element => element);
-    const portString = elements[elements.length - 1];
+    let elements = url.split('/').filter(element => element);
+    let portString = elements[elements.length - 1];
+    if (Number.isNaN(parseFloat(portString))) {
+      elements = url.split(':').filter(element => element);
+      portString = elements[elements.length - 1];
+    }
     const port = parseFloat(portString) + 500;
```
omichel, line 16: Shouldn't we use `parseInt` instead of `parseFloat`?
```suggestion
    if (Number.isNaN(parseInt(portString))) {
      elements = url.split(':').filter(element => element);
      portString = elements[elements.length - 1];
    }
    const port = parseInt(portString) + 500;
```

BenjaminDeleze: Indeed it seems there is not reason to use parseFloat

omichel (approved):
Thank you.

PR #4589, opened 2022-06-03
Fix compilation on Windows CI

omichel:
It seems the buggy `-Wdangling-pointer` warning was introduced in [gcc 12.1.0](https://gcc.gnu.org/onlinedocs/gcc-12.1.0/gcc/Warning-Options.html) as compared to [gcc 11.3.0](https://gcc.gnu.org/onlinedocs/gcc-11.3.0/gcc/Warning-Options.html#Warning-Options) which is not yet present on the Windows CI machines. I prefer to make a strict test on gcc 12.1.0 to be able to detect when we can get rid of these pragmas after a gcc update.

### DIFF
--- a/src/webots/nodes/WbConnector.cpp
+++ b/src/webots/nodes/WbConnector.cpp
@@ -372,9 +372,9 @@ void WbConnector::snapOrigins(WbConnector *other) {
       h[i] /= 2.0;
   }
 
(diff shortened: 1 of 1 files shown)


ygoumaz (approved):
This still prevents the warning on dangling pointers to show up and doesn't raise and additional warning on other platforms for unknown pragma. Works fine on Linux and Windows.

PR #4594, opened 2022-06-03
Fix mac bundle

omichel:
This PR aims at fixing the Webots.app bundle on macOS which is displayed as "damaged" when it contains a universal binary for webots (arm64 & x86_64).

1. Universal apps needs to be signed, therefore I added the `codesign` step for every binary.
2. In order to sign a macOS bundle, the bundle has to respect the Apple bundle folder hierarchy which was wrong, so I fixed it as well.
3. I upgraded to Python 3.10.5.
4. While testing I also found and fixed a bug preventing the detection of the most recent installed version of MATLAB.


ygoumaz (commented):
I have some minor suggestions/questions on the code. I tested the compilation and distribution on Linux and it works fine. I still need to test on MacOS.

scripts/packaging/mac_distro.py
```diff
@@ -159,19 +186,31 @@ def create_webots_bundle(self):
 
         print('Done.\n')
 
-    def make_dir(self, directory):
+    def make_dir(self, directory=''):
```
stefaniapedrazzi, line 189: Why did you add a default empty value for the directory?
If directory is empty, then this function does nothing and should probably not be called.

omichel: I used it before to create the 'Contents' folder, but I don't use it any more, so I just removed it.

omichel:
Note: you can test that a binary file, e.g. `webots` contains both architectures with `lipo -archs webots`. It also works with `.dylib` dynamic libraries and `.a` static libraries.

ygoumaz (commented):
The Matlab issue seems to be completely fixed. 

On the other side, I noticed a few additional wrong behaviors with the compilation on MacOS:

- `make clean` doesn't clean the lib/controller/matlab folder correctly.
- It is impossible to call `make release` or `make distrib` multiple time in a row, as some directories already exists.
- It is impossible to make a distribution. The process crashes at step "checking RPATH system" with the following errors:
Dependency error:
  - File: ./resources/lua/modules/gd/gd.dylib
  - Dependency: @rpath/lib/webots/liblua.dylib

  Dependency error:
  - File: ./resources/lua/modules/gd/gd.dylib
  - Dependency: @rpath/lib/webots/libgd.3.dylib

omichel:
OK, I believe I fixed all the issues you reported.

omichel:
🎉

PR #4601, opened 2022-06-07
Fixed gcc compilation failure

omichel:
It seems the Windows CI machine upgraded gcc to version 12.1.0 thus causing the compilation to [fail](https://github.com/cyberbotics/webots/runs/6764615371?check_suite_focus=true). This PR back-ports the fix already present in the develop branch (#4589) to fix the problem.

### DIFF
--- a/src/webots/nodes/WbConnector.cpp
+++ b/src/webots/nodes/WbConnector.cpp
@@ -371,12 +371,17 @@ void WbConnector::snapOrigins(WbConnector *other) {
     for (int i = 0; i < 3; ++i)
       h[i] /= 2.0;
   }
-
+// gcc 12.1.0 is raising a false positive warning here about dangling pointers
+#pragma GCC diagnostic push
+#if __GNUC__ == 12 && __GNUC_MINOR__ == 1 && __GNUC_PATCHLEVEL__ == 0
+#pragma GCC diagnostic ignored "-Wdangling-pointer"
+#endif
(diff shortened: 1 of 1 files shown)


src/webots/nodes/WbConnector.cpp
```diff
@@ -371,12 +371,17 @@ void WbConnector::snapOrigins(WbConnector *other) {
     for (int i = 0; i < 3; ++i)
       h[i] /= 2.0;
   }
-
+// gcc 12.1.0 is raising a false positive warning here about dangling pointers
+#pragma GCC diagnostic push
+#if __GNUC__ == 12 && __GNUC_MINOR__ == 1 && __GNUC_PATCHLEVEL__ == 0
+#pragma GCC diagnostic ignored "-Wdangling-pointer"
+#endif
```
ygoumaz, line 378: The line above is not deleted in the develop branch.
```suggestion

// gcc 12.1.0 is raising a false positive warning here about dangling pointers
#pragma GCC diagnostic push
#if __GNUC__ == 12 && __GNUC_MINOR__ == 1 && __GNUC_PATCHLEVEL__ == 0
#pragma GCC diagnostic ignored "-Wdangling-pointer"
#endif
```

ygoumaz (commented):
Warnings are successfully removed on Windows and Linux. Just a minor modification suggestion.

PR #4604, opened 2022-06-07
Fix path to appdmg.json file

stefaniapedrazzi:
Fix remove of `appdmg.json` linking to the wrong path.
After correctly removing it, the `appdmg.json` should no longer appear in the release page.

https://github.com/cyberbotics/webots/releases/tag/nightly_6_6_2022


BenjaminDeleze (dismissed):
Thank you

BenjaminDeleze (approved):

PR #4605, opened 2022-06-07
Fix loading of broken image to be non-blocking

BenjaminDeleze:
ditto

### DIFF
--- a/resources/web/wwi/image_loader.js
+++ b/resources/web/wwi/image_loader.js
@@ -87,7 +87,7 @@ function _loadImage(src) {
     img.onerror = () => {
       if (typeof img.failed === 'undefined') {
         console.error('Error in loading: ' + src);
-        img.src = 'https://cyberbotics.com/wwi/images/missing_texture.pn';
+        img.src = 'https://cyberbotics.com/wwi/images/missing_texture.png';
         img.failed = true;
       } else
         // Fail to get the missing texture image, exit to avoid an infinite loop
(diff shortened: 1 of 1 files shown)


resources/web/wwi/image_loader.js
```diff
@@ -83,7 +84,15 @@ function _loadImage(src) {
     img.onload = () => {
       resolve(img);
     };
-    img.onerror = () => console.error('Error in loading: ' + src);
+    img.onerror = () => {
+      if (typeof img.failed === 'undefined') {
+        console.error('Error in loading: ' + src);
+        img.src = 'https://cyberbotics.com/wwi/images/missing_texture.pn';
```
omichel, line 90:
```suggestion
        img.src = 'https://cyberbotics.com/wwi/images/missing_texture.png';
```

resources/web/wwi/image_loader.js
```diff
@@ -83,7 +84,15 @@ function _loadImage(src) {
...
+        img.failed = true;
+      } else
+        // Fail to get the missing texture image, exit to avoid an infinite loop
```
omichel, line 93: This should never happen unless the `missing_texture.png` file is missing, right?
If so, we should probably print an error message in the console here.

omichel (changes requested):

resources/web/wwi/image_loader.js
BenjaminDeleze: It could happen if there is a internet cut-off right at the wrong moment. 
Highly unlikely but not impossible.

A message is automatically printed in the console when WebotsJS fail to download an URL

Issue #4602, opened 2022-06-07
Suggestions to improve UI for exporting animations and scenes locally


stefaniapedrazzi:
In the develop branch, it is very difficult to find how to save a scene or animation locally.
It is OK to try to promote as much as possible the upload on webots.cloud.
But at the moment it is not at all self-explanatory what will be done when pressing the "upload" buttons when the "Save as local files" option is checked.

I think that we could apply some minor improvements to make it more user friendly.

Here are some suggestions for discussion:
1. I think that if the naming the menu item simply "Share" instead of "Share on webots.cloud" would be less confusing in case a user wants to save locally only. Otherwise she might never open the dialog to search for the option.
2. Similarly, I personally would change the dialog title from "Share your simulation online" to "Share your simulation"
3. The word "upload" in the button text is misleading when the "save as local files" option is checked.
   What about using radio buttons and change, something like the one in the picture?
    ![Screenshot from 2022-06-07 10-48-12](https://user-images.githubusercontent.com/5910449/172338192-1683a843-6171-4abb-abf7-dbd6cbae51f0.png)
4. Otherwise if we should decide to keep the save as local file as an exceptional behavior enabled by a checkbox, I think we should at least change the label from "Save as local files" to "Save as local files only"

----
Additionally, the documentation has not been updated accordingly to the changes to the UI and still refers to the old menu items.
These are at least three pages that should be updated:
* https://www.cyberbotics.com/doc/guide/web-scene?version=develop#how-to-export-a-web-scene
* https://www.cyberbotics.com/doc/guide/web-animation?version=develop#how-to-export-a-web-animation
* https://www.cyberbotics.com/doc/guide/the-user-interface?version=develop#file-menu

BenjaminDeleze:
I will implement these changes

PR #4606, opened 2022-06-07
Improve export dialog

BenjaminDeleze:
Adopt the improvements discussed in #4602 

- [x] Code
- [x] CSS
- [x] Docs


src/webots/gui/WbShareWindow.cpp
```diff
@@ -19,6 +19,7 @@
 #include <QtGui/QDesktopServices>
 #include <QtWidgets/QCheckBox>
```
stefaniapedrazzi, line 20: It seems that this include is useless now.

docs/guide/the-user-interface.md
```diff
@@ -76,12 +76,9 @@ Checking the video caption option will display the acceleration value in the top
...
 
-- The **Export HTML5 Scene...** item allows you to export the current world as an interactive 3D ".html" file.
-You can get more information about this topic in [this section](web-scene.md).
-
-- ![](images/animation-button.png =26x26) The **Make HTML5 Animation...** item allows you to record a simulation as a 3D animation and publish it on a HTML5 web page.
-Once the animation recording is started, this item is changed to `Stop HTML5 Animation...` and can be used to stop the animation recording.
-You can get more information about this topic in [this section](web-animation.md).
+- ![](images/share-button.png =26x26) The **Share...** item allows you to export the current world as an interactive 3D scene (more information about this topic in [this section](web-scene.md)), or to record a simulation as a 3D animation and publish it on a HTML5 web page.
+Once the animation recording is started, this item is changed to `Stop HTML5 Animation recording.` and can be used to stop the animation recording.
+You can get more information about animations in [this section](web-animation.md).
```
stefaniapedrazzi, line 81: We should also describe the two different possibilities (save locally and upload to webots.cloud).
Moreover we mention webots.cloud, but we should also describe somewhere what is webots.cloud.

docs/guide/the-user-interface.md
```diff
@@ -76,12 +76,9 @@ Checking the video caption option will display the acceleration value in the top
...
 %end
 
-- The **Export HTML5 Scene...** item allows you to export the current world as an interactive 3D ".html" file.
-You can get more information about this topic in [this section](web-scene.md).
-
-- ![](images/animation-button.png =26x26) The **Make HTML5 Animation...** item allows you to record a simulation as a 3D animation and publish it on a HTML5 web page.
-Once the animation recording is started, this item is changed to `Stop HTML5 Animation...` and can be used to stop the animation recording.
-You can get more information about this topic in [this section](web-animation.md).
+- ![](images/share-button.png =26x26) The **Share...** item allows you to export the current world as an interactive 3D scene (more information about this topic in [this section](web-scene.md)), or to record a simulation as a 3D animation and publish it on a HTML5 web page.
+Once the animation recording is started, this item is changed to `Stop HTML5 Animation recording.` and can be used to stop the animation recording.
```
stefaniapedrazzi, line 80: i would change this sentence given that the previous one is about both saving a scene and an animation.
Maybe something like this could be better:
```suggestion
If the animation recording is started, this item is changed to `Stop HTML5 Animation recording.` and can be used to stop the animation recording.
```

src/webots/gui/WbShareWindow.cpp
```diff
@@ -33,41 +34,38 @@ WbShareWindow::WbShareWindow(QWidget *parent) : QDialog(parent) {
...
   QGridLayout *layout = new QGridLayout(this);
 
-  QLabel *labelIntro = new QLabel(this);
-  labelIntro->setOpenExternalLinks(true);
-  labelIntro->setText(tr("<html><head/><body><p>Publish your simulation on <a href=\"https://%1/\"><span "
-                         "style=\" text-decoration: underline; color:#5dade2;\">%1</span></a>.</p></body></html>")
-                        .arg(uploadUrl));
-  layout->addWidget(labelIntro, 3, 0, 1, 2);
+  QRadioButton *boxButtons[2];
+  boxButtons[0] = new QRadioButton("Upload to " + uploadUrl, this);
```
stefaniapedrazzi, line 42: The only thing that is worse than the previous version is that now the webots.cloud is a simple text and no longer a rich text with link to URL.
But it is seems complicated to enable rich HTML text in radio label, so it's fine.

stefaniapedrazzi (changes requested):
Thank you. It looks much better to me now.

I just have a couple of comments.

1. wouldn't be better to add a minimum width for the dialog box now that it is very small?

2. I think that we should add a section in the documentation about webots.cloud. But this can be done in a separate PR.

docs/guide/web-animation.md
```diff
@@ -17,11 +17,10 @@ The [figure below](#screenshot-of-a-web-animation-page-generated-by-webots) show
 
 ### How to Export a Web Animation
 
-Select the `File / Make HTML5 Animation...` menu item and choose the target `HTML` file.
+Select the `Share...` menu item and choose if you want to upload it to [Webots.cloud](https://webots.cloud) or to save it locally, then click the `Record and share animation` button to start the recording.
```
stefaniapedrazzi, line 20: Shouldn't be with lower case:
```suggestion
Select the `Share...` menu item and choose if you want to upload it to [webots.cloud](https://webots.cloud) or to save it locally, then click the `Record and share animation` button to start the recording.
```

I also checked and in the code and documentation of the preferences the URL is still `beta.webots.cloud` and not `webots.cloud`

BenjaminDeleze: Yes, but it will be replaced by webots.cloud in the coming weeks, so I'm not sure it's worth changing it there and risk forgetting to change it again.

docs/guide/the-user-interface.md
BenjaminDeleze: I added a sentence about saving locally or uploading to webots.cloud.

As you suggested, I think it would be a could idea to make some documentation about webots.cloud, but in another PR, and then we can put a link here pointing to this documentation.

BenjaminDeleze:
This is the default (and minimum) size of the dialog box on my computer. 
Is it smaller on your computer?
Do you think it should be bigger?

size

stefaniapedrazzi:
Yes, this is the same I get on my computer.
Personally I would make it a little bit bigger.

docs/guide/the-user-interface.md
```diff
@@ -76,12 +76,10 @@ Checking the video caption option will display the acceleration value in the top
...
 %end
 
-- The **Export HTML5 Scene...** item allows you to export the current world as an interactive 3D ".html" file.
-You can get more information about this topic in [this section](web-scene.md).
-
-- ![](images/animation-button.png =26x26) The **Make HTML5 Animation...** item allows you to record a simulation as a 3D animation and publish it on a HTML5 web page.
-Once the animation recording is started, this item is changed to `Stop HTML5 Animation...` and can be used to stop the animation recording.
-You can get more information about this topic in [this section](web-animation.md).
+- ![](images/share-button.png =26x26) The **Share...** item allows you to export the current world as an interactive 3D scene (more information about this topic in [this section](web-scene.md)), or to record a simulation as a 3D animation and publish it on a HTML5 web page.
+For both options, it is possible to save the result locally or to upload it automatically to webots.cloud where it can be easily shared and seen by other people.
```
stefaniapedrazzi, line 80:
```suggestion
For both options, it is possible to save the result locally or to upload it automatically to [webots.cloud](https://webots.cloud) where it can be easily shared and seen by other people.
```

stefaniapedrazzi (commented):
Thank you for the changes.
Here are just a couple of reference to webots.cloud that should probably also be updated.

stefaniapedrazzi (approved):
Thank you!

PR #4608, opened 2022-06-08
Fix urdf export crashes

ad-daniel:
**Description**

using the `./webots conver /path/to/a/proto` utility currently generates several crashes


omichel (approved):
Thank you.

Issue #4003, opened 2021-12-08
Incorrect texture mapping of the salamander


BenjaminDeleze:
The texture mapping of the salamander in `animated_skin.wbt` is different

BenjaminDeleze:
seems to have been broken somewhere in https://github.com/cyberbotics/webots/pull/3566

BenjaminDeleze:
The bad mapping have appeared during the translation from ogre format to .fbx.

It is not critical because the salamander still have a texture, but it is less pretty (especially around the head). As I don't know how to fix it (I tried manually with blender and tried to find a ogre exporter), I will let it for later

stefaniapedrazzi:
The different mapping was made on purpose, so the fact that it looks different is not an "issue" but the issue is if it looks bad.
What do you mean by "bad mapping"?
Could you point exactly which part has a bad mapping?
We should aim at having a nice looking mapping but not exactly the same as before.
In any case, I can rework it and make it looks better.

It is very complicated to have the exact same old Ogre mapping as before and it is much better to create a new one with Blender as it was done in https://github.com/cyberbotics/webots/pull/3566 (and also when creating the Ogre mapping).
To way to fix is to manually adjust the map in Blender.

BenjaminDeleze:
I think that the mapping around the head looks strange. I said "bad mapping", because I had the impression that the previous mapping had a "specific mapping" for the head (small scales that radiate and get bigger) whereas now it seems weird with big scales.

Before:
![animated_skin_before](https://user-images.githubusercontent.com/25938827/166205380-505788f5-2925-4a6b-81ad-3ca8abb03d3c.png)
Now
![animated_skin_now](https://user-images.githubusercontent.com/25938827/166205389-8778d27a-6d90-42df-9984-edb95e61a47f.png)

But I am not sure if it is worth to rework it as it looks okay as long as we do not zoom in too much.

PR #4610, opened 2022-06-08
Improve texture mapping of salamander head

stefaniapedrazzi:
Address #4003.

Improve the texture mapping by reducing the size of the skin patterns on the head.

Here is a screenshot with the new appearance:
animated_skin


BenjaminDeleze (approved):
Thank you! It definitely looks better

PR #4615, opened 2022-06-08
Fix cadshape with .mtl spaced name

BenjaminDeleze:
**Description**
if we had an `.obj` with:
```
mtllib Cat 3D Model.mtl
```
CadShape believe there was 3 `.mtl` files. So I had to modifiy a bit how it was split

### DIFF
--- a/src/webots/nodes/WbCadShape.cpp
+++ b/src/webots/nodes/WbCadShape.cpp
@@ -285,8 +285,8 @@ QStringList WbCadShape::objMaterialList(const QString &url) const {
       if (!cleanLine.startsWith("mtllib"))
         continue;
 
-      cleanLine = cleanLine.replace("mtllib", "");
-      materials << cleanLine.split(' ', Qt::SkipEmptyParts);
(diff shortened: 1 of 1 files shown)


src/webots/nodes/WbCadShape.cpp
```diff
@@ -285,8 +285,10 @@ QStringList WbCadShape::objMaterialList(const QString &url) const {
       if (!cleanLine.startsWith("mtllib"))
         continue;
 
-      cleanLine = cleanLine.replace("mtllib", "");
-      materials << cleanLine.split(' ', Qt::SkipEmptyParts);
+      cleanLine = cleanLine.replace("mtllib ", "");
```
omichel, line 288: I would feel safer with this. If the line contains an extra space at the end, that would break the rest of the code.
```suggestion
      cleanLine = cleanLine.replace("mtllib ", "").trimmed();
```

omichel (changes requested):
I believe we should also handle the case where the filenames are in double quotes, see https://developer.blender.org/T67266

PR #4617, opened 2022-06-08
Remove i686 code (libController 32 bits on Windows)

omichel:
This code was used for building controller programs linked against libraries that are available only for i686 architecture. This was the case for the Nao simulator SDK libraries and some old versions of MATLAB, Python, Java, etc. Hopefully, this is not needed any more.

Thanks to this PR, the size of the installation setup will shrink from 453 to 341 MB.


BenjaminDeleze (approved):
Thank you

PR #4619, opened 2022-06-09
Fix segmentation color change

### LINKED ISSUE #4618: Setting recognition colors does not apply until world is reloaded
When setting a `recognitionColors` for a solid, the color does not appear in the camera frame (with recognition node, segmentation enabled, and both `recognitionEnable` and `enableRecognitionSegmentation` set from the script), while the bounding box immediately appears.
If I save the scene and then reload the world, then the segmentation color is properly displayed as well.

I would like to automatically set the `recognitionColors` for a bunch of objects from a script when starting the simulation. I know the code to do this is working because I can see that the field is properly populated and, again, if I save the scene and reload the world, I can see the color displayed properly. 
However, the issue appears even if I set the field manually before starting the simulation: am I missing some initialization step that occurs when restarting, or should this work already?
Thank you!

stefaniapedrazzi:
Fix #4618: when `Solid.recognitionColors` changes, the descendant meshes (WbGeometry, WbCadShape and WbSkin) needs to be notified to update their segmentation color.


BenjaminDeleze (dismissed):
Thank you!

stefaniapedrazzi:
I had to fix some conflicts in the changelog.

BenjaminDeleze:
It seems like the test suite windows had a random failure

BenjaminDeleze (approved):
Thank you

PR #4102, opened 2022-01-08
Avoid exporting SolidReference to URDF

lukicdarkoo:
This PR fixes https://github.com/cyberbotics/webots_ros2/issues/365.

According to https://github.com/cyberbotics/urdf2webots/issues/31, URDF doesn't support closed-loop kinematics. Therefore, we simply skip converting SolidReference nodes to URDF.

**Tasks**
  - [x]  Fix
  - [x] Add changelog

PR #4620, opened 2022-06-10
Closed-loop kinematics detection in webots convert URDF

### LINKED ISSUE #4332: Webots dies if URDF export is called on parallel robot
**Describe the Bug**
If the `getUrdf()` method is called on a robot that has parallel kinematics, webots dies with the following error

```
terminate called after throwing an instance of 'std::bad_alloc'
what():  std::bad_alloc
```
Of course, it is impossible to export an URDF for such a robot, since URDF does not support it. It would be just nice to throw a better error message to point the user to it (took me a moment to realize what the problem is).
For many robots it would still be possible to provide an URDF if passive joints are ignored (in the given example this would be possible). But this is more complicated and the user can remove the passive joints manually. So I think an error message would be enough.

I'm not sure how it comes to this error, but I guess that the algorithm gets caught in some loop.

**Steps to Reproduce**
1. Load GankenKun robot from https://cloud.robocup.org/s/WRX6METzXRPb8YX This one has parallel kinematics in the legs.
2. Run getURDF()
3. See that webots crashes with the error above

**Expected behavior**
Throw an error with a text like "Can not provide URDF for this robot since it has parallel kinematics. This is not supported by URDF. Depended on you robot, you might be able to get an URDF if you remove passive joints."

**Screenshots**
Picture of the parallel kinematics in the legs of the exemplary robot. 
gankenkun_parallel

**System**
 - Graphics Card: NVIDIA GeForce GTX 1060 6GB

ygoumaz:
Fixes #4332.

The crash was already fixed in PR #4102. Therefore robot.getUrdf() is now displaying a warning if the robot contains closed-loop kinematics.

However, this is not the case when using `webots convert` command line for an URDF conversion. The parsing of the PROTO isn't updating the solid children and solid end point references. Therefore, the closed-loop is not detected and no warning is displayed. This PR adds the references update, as well as the warnings for a better user experience.

### DIFF
--- a/src/webots/gui/WbSingleTaskApplication.cpp
+++ b/src/webots/gui/WbSingleTaskApplication.cpp
@@ -134,7 +134,11 @@ void WbSingleTaskApplication::convertProto() const {
   WbNode *node = WbNode::regenerateProtoInstanceFromParameters(model, fields, true, "");
   for (WbNode *subNode : node->subNodes(true)) {
     if (type == "URDF" && dynamic_cast<WbSolidReference *>(subNode))
-      cout << tr("Warning: Exporting a Joint node with a SolidReference endpoint (%1) to URDF is not supported.").arg(static_cast<WbSolidReference *>(subNode)->name()).toUtf8().constData() << endl;
+      cout << tr("Warning: Exporting a Joint node with a SolidReference endpoint (%1) to URDF is not supported.")
+                .arg(static_cast<WbSolidReference *>(subNode)->name())
+                .toUtf8()
+                .constData()
+           << endl;
     if (dynamic_cast<WbSolid *>(subNode))
       static_cast<WbSolid *>(subNode)->updateChildren();
     if (dynamic_cast<WbBasicJoint *>(subNode)) {
--- a/src/webots/gui/WbSingleTaskApplication.cpp
+++ b/src/webots/gui/WbSingleTaskApplication.cpp
@@ -20,6 +20,8 @@
 #include "WbProtoCachedInfo.hpp"
 #include "WbProtoList.hpp"
 #include "WbProtoModel.hpp"
+#include "WbSolid.hpp"
+#include "WbSolidReference.hpp"
 #include "WbSoundEngine.hpp"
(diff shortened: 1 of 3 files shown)


src/webots/gui/WbSingleTaskApplication.cpp
```diff
@@ -130,9 +132,18 @@ void WbSingleTaskApplication::convertProto() const {
   // Generate a node structure
   WbNode::setInstantiateMode(true);
   WbNode *node = WbNode::regenerateProtoInstanceFromParameters(model, fields, true, "");
-  for (WbNode *subNode : node->subNodes(true))
-    if (dynamic_cast<WbBasicJoint *>(subNode))
+  for (WbNode *subNode : node->subNodes(true)) {
+    if (type == QString("URDF") && dynamic_cast<WbSolidReference *>(subNode))
```
omichel, line 136: That is simpler and should work as well:
```suggestion
    if (type == "URDF" && dynamic_cast<WbSolidReference *>(subNode))
```

src/webots/gui/WbSingleTaskApplication.cpp
```diff
@@ -130,9 +132,18 @@ void WbSingleTaskApplication::convertProto() const {
...
+      cout << tr("Warning: Exporting a Joint node with a SolidReference endpoint (").toUtf8().constData()
+           << static_cast<WbSolidReference *>(subNode)->name().toStdString()
+           << tr(") to URDF is not supported.").toUtf8().constData() << endl;
```
omichel, line 139: It is better not to split a `tr` string as they are supposed to be translated at some point:
(caution: not tested, not clang-formatted)
```suggestion
      cout << tr("Warning: Exporting a Joint node with a SolidReference endpoint (%1) to URDF is not supported.").arg(static_cast<WbSolidReference *>(subNode)->name()).toUtf8().constData() << endl;
```

omichel (changes requested):
Well spotted! Just a couple of details...

omichel (approved):
Thank you.

PR #4634, opened 2022-06-13
Fix busy TCP port message

### LINKED ISSUE #4322: Open port warning
These warnings are displayed when opening two instances of Webots.
It is not clear in the message why Webots try to open a streaming server (it is to be able to display robot windows).

```
ERROR: Error when creating the TCP streaming server on port 1234: Cannot set the server in listen mode: The bound address is already in use
WARNING: Trying again with port 1235
```

We should improve the message and also refactor the code to open a streaming server only when the user open the first robot window or launch Webots with `--stream`

omichel:
Fixes #4322.

The error message is now hopefully more helpful:
```
WARNING: Could not listen to port 1234. Cannot set the server in listen mode: The bound address is already in use. Using port 1235 instead. This could be caused by another running instance of Webots. You should use the '--port' option to start several instances of Webots.
```


omichel:
The message was improved, explaining what Webots listen to on this TCP/IP port:
```
WARNING: Could not listen to extern controllers and robot windows on port 1234. Cannot set the server in listen mode: The bound address is already in use. Using port 1235 instead. This could be caused by another running instance of Webots. You should use the '--port' option to start several instances of Webots.
```

BenjaminDeleze (approved):
Thank you

PR #4638, opened 2022-06-14
Fixed missing Linux compilation dependency

omichel:
Following the [wiki instructions](https://github.com/cyberbotics/webots/wiki/Linux-installation/), I was unable to compile Webots on Ubuntu 22.04 because of a missing compilation dependency: `python-is-python3`. This PR fixes the problem.


stefaniapedrazzi:
We should just decide if Python is now mandatory or not for compiling Webots.
If yes, then this change is fine, we should remove it from `linux_optional_compilation_dependencies.sh` and make sure that the installation of Python works on all the platforms (it seems there were issue reported recently on Windows).
If not, we should instead fix the Makefile so that it doesn't break the compilation if python is not found.

omichel:
It turns out that python-is-python3 seems to be needed at least for the compilation of wrenjs and the vehicle library, but if it's missing it displays an error message instead of nicely skipping the compilation of these components. However, python3 is installed on all the system under which we compile Webots. So, I would suggest that python3 be used by default everywhere instead of python and that python3 is a requirement for compiling Webots. This means we should change several Makefile to use python3 instead of python and remove the python-is-python3 package installation.

omichel:
I tested this branch on both Windows and Linux and compilation went fine.

BenjaminDeleze (approved):
Seems good to me

PR #4626, opened 2022-06-13
Fix Skin bones orientations when using Webots nodes

stefaniapedrazzi:
Fix #4611: bones were wrongly displayed if Webots Solid nodes were used as in the `animated_skin.wbt` simultation.
* compute bones orientation based on current and parent node position
* add representation of tail bones with same orientation and scale as parent bone (as also done by Blender)
* avoid code duplication by moving additional instructions to `WbSkin::createBoneRepresentation`

PR #4639, opened 2022-06-14
Fix test suite in debug mode

### LINKED ISSUE #4636: Asstertion failure with extern controllers in test suite
When running the `extern` test from the test_suite locally on Ubuntu 22.04 with Webots compiled in debug mode, I get this error:
```
webots-bin: control/WbControlledWorld.cpp:414: void WbControlledWorld::externConnection(WbController*, bool): Assertion `mControllers.count(controller) + mTerminatingControllers.count(controller) == 1' failed.
./webots: line 88: 133826 Aborted                 (core dumped) "$webots_home/bin/webots-bin" "$@"
```

ygoumaz:
Fix #4636.

The test suite is failing when Webots is compiled in debug mode because of 2 tests:

- extern: After an external controller terminates by itself, an assertion checks whether the controller is contained in the mController or mTerminatingController list. This is often not the case, as the controller may not be processing requests before being terminated. This is the only situation where a controller may not be in a list at all.
- skin: The orientation correction introduced in #4626 for skin bones does not take the singular case of zero offset into account.


src/webots/nodes/WbSkin.cpp
```diff
@@ -728,7 +728,8 @@ bool WbSkin::createSkeletonFromWebotsNodes() {
       const WbVector3 unit(0, 0, 1);
       const WbVector3 &norm = offset.normalized();
       const WbVector3 &axis = unit.cross(norm).normalized();
-      const float boneOrientation[4] = {(float)unit.angle(norm), (float)axis[0], (float)axis[1], (float)axis[2]};
+      const float angle = offset.length() ? (float)unit.angle(norm) : 0;
```
stefaniapedrazzi, line 731: I think it would be better to move the check in the `WbVector3::angle` function directory to avoid similar issues in the future.
Moreover we do not need to compute the sqrt but  `WbVector3::length2` could be checked instead.

stefaniapedrazzi (changes requested):
Thank you!

stefaniapedrazzi (commented):
Thank you.

At first sight, I would say that it is not very efficient to compute the length2 twice per method and it probably would be better to store it in a variable and reuse it later.
Especially because these are functions that will be called during the simulation step.
However, I didn't benchmark it to be sure what is the performance difference.

src/webots/maths/WbVector2.hpp
```diff
@@ -127,8 +127,14 @@ class WbVector2 {
   double distance2(const WbVector2 &v) const { return (*this - v).length2(); }
 
   // returns a unit vector with the same direction: / length
-  void normalize() { *this /= length(); }
-  WbVector2 normalized() const { return *this / length(); }
+  void normalize() {
+    double l = length();
+    l ? *this /= l : *this;
```
omichel, line 132:
```suggestion
    return l ? *this /= l : *this;
```

omichel (approved):
Thank you.

PR #4641, opened 2022-06-14
Fixed warning on Ubuntu 22.04

omichel:
Starting Webots on Ubuntu 22.04 is raising the following warning in the console:
```
Info: Warning: Ignoring WAYLAND_DISPLAY on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
```
This PR fixes it.


stefaniapedrazzi (approved):
Thank you!

On my local system the windowing system was set by default to x11. Not sure way, but I will change it back to default wayland.

PR #4643, opened 2022-06-14
Enhancement reduce assets size

BenjaminDeleze:
**Description**
Reduce the size of images above 4 MB by either reducing their resolution and/or the quality.

The gain is from 852.3 MB (756 MB compressed) to 639.4 MB (642.7 MB compressed)


omichel (approved):
Thank you.

PR #4646, opened 2022-06-14
Enhancement share reset and thumbnail robustness

ThomasOliverKimble:
**Description**

- Improves usability of animation and scene sharing by pausing when the dialog is displayed and resuming afterwards (+ code cleanup)
- Gives a more clear indication of what is happening on simulation reset during animation recording with an info log message .
- Improves thumbnail robustness by only resizing back to original size once screenshot is written thanks to a new signal.
- Add checkbox to preferences to disable thumbnails if wanted


omichel (approved):
Looks good to me.

ThomasOliverKimble:
I have added an option to the preferences to disable thumbnails. I have also done some changes server side to allow uploads without thumbnails, so it should work all of the time now. Please let me know if you find any other bugs.

omichel (approved):
All good.

PR #4656, opened 2022-06-15
Use assimp-5.2.3 on Windows

### LINKED ISSUE #4653: The skin and animated skin examples are broken on Windows
This was likely caused by an update of assimp...
animated_skin
skin

omichel:
Fixes #4653.

Since assimp 5.2.4 introduced a bug breaking up the skeletons / skins in Webots, this PR reverts to version 5.2.3 which works well.
On the long term, it might be a good solution to have the same version of assimp on Linux, macOS and Windows, regardless of MSYS2 updates.


ygoumaz (approved):
Thank you!

PR #4658, opened 2022-06-16
Fix crash on save

### LINKED ISSUE #4652: Crash on save
**Describe the Bug**
Webots crashes when saving

**Steps to Reproduce**
1. Open this example world:
```
#VRML_SIM R2022b utf8
WorldInfo {
}
Viewpoint {
  orientation 0.13709495266901983 -0.11218389057731509 -0.9841848142740347 4.497584235534405
  position 0.6804287750772149 -3.592400836888023 1.4935760550098043
}
Solid {
  children [
    Shape {
      geometry Box {
        size 0.1 0.1 0.1
      }
    }
  ]
}

```
2. Cut the shape
3. Paste the shape out of the Solid
4. Save
5. Webots crashes

**System**
 - Ubuntu 20.04, develop branch

BenjaminDeleze:
Fix #4652

### DIFF
--- a/src/webots/control/WbLanguageTools.cpp
+++ b/src/webots/control/WbLanguageTools.cpp
@@ -97,6 +97,8 @@ QString WbLanguageTools::pythonCommand(QString &shortVersion, const QString &com
     pythonCommand = "!";
   } else
     shortVersion = QString(version[0][0]) + version[0][2];
+  if (version[0][3] != '.')
+    shortVersion += version[0][3];  // handle versions 310, 311, 321, etc.
 #elif __APPLE__
   if (std::getenv("PWD"))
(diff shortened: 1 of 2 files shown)


BenjaminDeleze:
I agree but I was afraid that by pointing directly to the signal in the code, one modification in the code would be enough to shift the lines and so not point to the right line anymore.

stefaniapedrazzi:
> I agree but I was afraid that by pointing directly to the signal in the code, one modification in the code would be enough to shift the lines and so not point to the right line anymore.

In fact you should not point to a specific line, but just mention the method/signal that is triggered and described the problem.
The connect statement can then be easily found searching in the source code.

stefaniapedrazzi (approved):
Thank you!

PR #4659, opened 2022-06-16
Fix animation time start

ThomasOliverKimble:
**Description**
Timer now starts at zero when animation is recorded.


BenjaminDeleze (approved):
Thank you!

PR #4661, opened 2022-06-16
Fix redundant copy of libode.so

### LINKED ISSUE #4650: Redundant copy of libode.so
When building Webots develop branch from sources after a `make cleanse`, the ODE library `libode.so` is compiled and copied to `WEBOTS_HOME/lib/webots`. This should only be the case if `libode.so` is not already located in `WEBOTS_HOME/lib/webots`. If multiple `make`s are done consecutively, libode.so is always copied.
```
# copying libode.so to $WEBOTS_HOME/lib/webots/libode.so
```
**Steps to Reproduce**
1. Install Webots development envrionment (develop branch) or `make cleanse` in an already existing one.
2. `make` -> ode library is compiled, linked and copied.
3. `make` again -> ode library is copied once again but it should not.

It seems that this is an old issue that was highlighted by #4543.

One solution could be to directly compile libode.so in the correct directory instead of making an additional copy.

ygoumaz:
Fix #4650 

This PR is changing the target directory when linking the libode.so shared library. This way no copy is needed anymore.


omichel (changes requested):
I checked the code and tested it on Windows and it works very well.
I would maybe rename the new variable from `LIB_BUILD_DIR` to `TARGET_LIB_DIR` which seems more understandable to me.

omichel (approved):
Thank you.

PR #4663, opened 2022-06-16
Enhance thumbnail upload robustness

### LINKED ISSUE #4654: Thumbnail not always uploading to cloud
**Describe the Bug**
Sometimes when uploading to `webots.cloud`: `share`->`upload to webots.cloud`->`Record and export animation` or `Export scene`, the thumbnail is not saved in time for the upload and therefore is ignored and not uploaded to the server.

ThomasOliverKimble:
**Description**
Fixes #4654: Checks for files 10 times every 100ms, if they don't exist, don't try and upload them.

### DIFF
--- a/src/webots/gui/WbMainWindow.cpp
+++ b/src/webots/gui/WbMainWindow.cpp
@@ -1609,11 +1609,11 @@ void WbMainWindow::upload() {
   if (filenames.isEmpty())  // add empty texture
     filenames.append("");
 
-  if (mUploadType == 'A' && uploadFileExists("cloud_export.json"))
+  if (QFileInfo(WbStandardPaths::webotsTmpPath() + "cloud_export.json").exists() && mUploadType == 'A')
     filenames << "cloud_export.json";
-  if (uploadFileExists("cloud_export.x3d"))
-    filenames << "cloud_export.x3d";
-  if (WbPreferences::instance()->value("General/thumbnail").toBool() && uploadFileExists("cloud_export.jpg"))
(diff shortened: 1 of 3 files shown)


src/webots/gui/WbMainWindow.cpp
```diff
@@ -1747,6 +1747,21 @@ void WbMainWindow::uploadStatus() {
     WbMessageBox::critical(tr("Upload failed: Upload status could not be modified."));
 }
 
+bool WbMainWindow::uploadFileExists(QString filename) {
+  int maxIterations = 10;
+  while (!QFileInfo(WbStandardPaths::webotsTmpPath() + filename).exists() && maxIterations) {
+    QTime delay = QTime::currentTime().addMSecs(100);
+    while (QTime::currentTime() < delay)
+      QCoreApplication::processEvents(QEventLoop::AllEvents);
```
omichel, line 1755: Calling `QCoreApplication::processEvents` in an active loop is probably super computer intensive. I believe we should avoid this:
```suggestion
    QCoreApplication::processEvents(QEventLoop::AllEvents, 100);
```

omichel (changes requested):

src/webots/gui/WbMainWindow.cpp
ThomasOliverKimble: I have changed this for a QThread::msleep(100) instead.

omichel (approved):
Better this way indeed.
Thank you.

omichel (approved):

PR #4669, opened 2022-06-16
Remove timeout

ad-daniel:
The timeout added with the new caching system sometimes interferes with the loading (especially big worlds with slow internet connections). I added it because if the user went offline mid-download it prevented the load from getting stuck, but as it stands this approach appears to have problems of its own.
Pushing this temporary solution for now so that nightlies aren't affected by it


ygoumaz (approved):
Thank you!

PR #4670, opened 2022-06-17
fix button width for mac

ThomasOliverKimble:
**Description**
When uploading a scene/animation to webots.cloud (`share`->`upload to webots.cloud`->`Export scene` or `Record and export animation`), the `Open in Browser` button is cropped on macOS. This fixes it.


omichel (approved):
Thank you.

Issue #4668, opened 2022-06-16
R2022a Nightly build broken on windows


BenjaminDeleze:
The last nightly builds for windows (R2022a) do not manage to load `.jpg` (both remote and local)

What have been tested so far:

- R2022b is working
- R2022a is working on ubuntu 20 (both `.deb` and `.snap`)
- R2022a is working when built from the sources
- I saw no errors in the logs of the github action

Could it be something related with `libjpeg` or `openjpeg`?

BenjaminDeleze:
It was decided to drop the nightly build of R2022a for windows, as R2022b is working and will be released in the next couple of weeks.

PR #4672, opened 2022-06-17
Remove R2022a windows nigthly

BenjaminDeleze:
As the release will happen quite soon, it is not worth investing time trying to fix the R2022a package for windows
(for more information check #4668)


BenjaminDeleze:
I also removed the R2022a.exe from the existing nigthly build

ygoumaz (approved):
Thank you!

omichel (approved):
Thank you.
Note: we should not forget to revert this when we release R2022b, maybe open an issue with milestone R2022b-rev1.

PR #4674, opened 2022-06-17
Fixed index of red cylinder on contact point demo

omichel:
Fixes a minor problem in a demo controller.


ygoumaz (approved):
Well seen, thank you!

PR #4676, opened 2022-06-17
Fix download OpenSSL package only on Ubuntu 22.04

ygoumaz:
The support of Ubuntu 22.04 (#4502) added the download of OpenSSL dependencies. It should only be downloaded if the system is Ubuntu 22.04 which is currently not the case in the corresponding Makefile.


stefaniapedrazzi (changes requested):
Thank you

scripts/packaging/webots_distro.py
```diff
@@ -53,6 +53,7 @@
 
 # sanity check: the following command will display an error if webots fails to start
 status = os.system(f'bash -c "{webots_command} --sysinfo"')
+print(status)
```
stefaniapedrazzi, line 56: This debug print statement should be removed before merging.
Are you still debugging the issue on Windows?

dependencies/Makefile.linux
```diff
@@ -15,11 +15,12 @@ PICO_PACKAGE = libpico.tar.bz2
 LUA_PACKAGE = lua-5.2.3.tar.gz
 OPENSSL_PACKAGE = libssl_1.1.tar.xz
 
-PACKAGES = qt open-al ois pico lua-gd lua assimp open-ssl
+PACKAGES = qt open-al ois pico lua-gd lua assimp
 PACKAGES_CLEAN = $(addsuffix -clean, $(PACKAGES))
 
 ifeq ($(UBUNTU_VERSION), 22.04)
 	PACKAGES += open-ssl
+	PACKAGES_CLEAN += open-ssl-clean
```
stefaniapedrazzi, line 23: Is this really needed?
Makefile uses lazy evaluation.
I tested locally and without this works fine, otherwise the target is added twice.

dependencies/Makefile.linux
```diff
@@ -15,11 +15,12 @@ PICO_PACKAGE = libpico.tar.bz2
 LUA_PACKAGE = lua-5.2.3.tar.gz
 OPENSSL_PACKAGE = libssl_1.1.tar.xz
 
-PACKAGES = qt open-al ois pico lua-gd lua assimp open-ssl
+PACKAGES = qt open-al ois pico lua-gd lua assimp
```
stefaniapedrazzi, line 18: `lua-gd` target has been removed in another PR but not from the list of targets.
```suggestion
PACKAGES = qt open-al ois pico lua assimp
```

scripts/packaging/webots_distro.py
omichel: > Are you still debugging the issue on Windows?

No, this was fixed in  #4682.

dependencies/Makefile.linux
ygoumaz: You are right, this is not needed with lazy evaluation.

dependencies/Makefile.linux
ygoumaz: Well spotted!

stefaniapedrazzi (approved):
Thank you!

PR #4679, opened 2022-06-17
Fixed highway driving robotbenchmark

### LINKED ISSUE #4648: highway_driving_benchmark controller crashes
**Describe the Bug**
The controller crash with the messages
```
INFO: sensors_visualization: refusing connection attempt from another extern controller.
WARNING: 'highway_driving_benchmark' controller exited with status: 1
```

This bugs seems to have been introduced in #4344

omichel:
Fixes #4648.

This simulation had two problems:
1. Two robots had the same default name and it was not detected because one robot is the children of another robot. I will open a new issue about this case which is currently not very well handled for extern robot controllers.
2. The ephemeral EXTERPROTO declaration for SumoInterface was missing in the world file.


ad-daniel (approved):
Thank you

PR #4682, opened 2022-06-17
Fixed PEP8 in generate_proto_list.py

omichel:
generate_proto_list.py was not respecting FLAKE8 W605, see https://stackoverflow.com/questions/69983472/how-to-satisfy-flake8-w605-invalid-escape-sequence-and-string-format-in

This PR also makes that `generate_proto_list.py` will do nothing if no PROTO file changed, thus making the global `make` run faster.

Finally, it disable the `webots --sysinfo` on the Windows platform because it crashes on the CI machine due to the lack of audio card.


omichel:
I removed the exit in case of failure on Windows. I would propose to merge this PR as soon as possible to allow the nightly builds on Windows.

omichel:
I allowed myself to merge this PR, so that the Windows nightly build will be created and we could test them on Monday.

PR #4686, opened 2022-06-20
Remove note about running old versions of Webots

stefaniapedrazzi:
This note has been added for R2022a, but it is incomplete given that it doesn't specify for example that Python won't work out of the box and you need to install a different versions, etc.

When R2022b is released we should promote the use of the latest version instead.


omichel (approved):
I agree. Thank you.

PR #4688, opened 2022-06-20
Fix nightly tests

ad-daniel:
**Description**
Due to recent changes world update currently fails, and I've taken the opportunity to do some cleanup (update_urls script called twice)


ad-daniel:
The branch isn't synchronized, but the tests passed (don't want to wait another 1h30)

PR #4689, opened 2022-06-20
Fix scheduled tests

ad-daniel:
**Description**
The failing test suite and world update is likely due to the fact that the scheduled nightly tests are actually run from the master branch's version of the workflow. As such, the test_suite_develop.yml workflow in the master branch needs to reflect the changes that have been made in the workflows on the develop side of things, otherwise it won't work. The fix should address both issues at once.

Assuming this PR https://github.com/cyberbotics/webots/pull/4688 passes without problems, it means that on the develop side we don't have issues

### DIFF
--- a/.github/workflows/test_suite_linux_develop.yml
+++ b/.github/workflows/test_suite_linux_develop.yml
@@ -101,12 +101,18 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: 3.9
+    - name: Get branch name
+      id: branch-name
+      uses: tj-actions/branch-names@v5.1
     - name: Test Suite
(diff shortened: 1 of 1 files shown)


ad-daniel:
The tests on the develop side are working as intended: https://github.com/cyberbotics/webots/pull/4688 (technically it failed because of the recent change in highway_driving, but it's unrelated: https://github.com/cyberbotics/webots/runs/6963975483?check_suite_focus=true, it will be fixed in that PR)
So the issue indeed appears to be on the master side, hence the need for this change. This PR can be reviewed

omichel (approved):
Thank you.

Issue #4687, opened 2022-06-20
Wrong scene tree selection reset on world save


stefaniapedrazzi:
When saving a world, the selection in the scene tree is reset, but the selection in the 3D view is not.
This is wrong and the selection should not change on save.

This is a regression bug introduced in the develop branch (R2022b). On R2022a it works correctly.

omichel:
Fixed in #4690.

PR #4690, opened 2022-06-20
Fix object selection issues

stefaniapedrazzi:
Address #4687 and fix inconsistencies with selection behavior:
* [x] on save the selection in the scene tree is wrongly reset (but not in the 3D scene)
    -> do not modify selection save
* [x] when clicking on the `Ephemeral EXTERNPROTO` button only the selectionin the scene tree was reset but not the selection in the 3D scene
    -> now it also resets the selection in the 3D scene

### DIFF
--- a/.github/workflows/test_suite_linux_develop.yml
+++ b/.github/workflows/test_suite_linux_develop.yml
@@ -4,7 +4,7 @@ on:
   schedule:
     - cron:  '0 23 * * *'
 
-concurrency: 
+concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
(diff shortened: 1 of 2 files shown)


stefaniapedrazzi:
The issue with selection on save has been introduced in #4593 and #4658 to prevent crashes when saving the thumbnail and preventing showing the selection when saving the thumbnail.
It is fine to clear the selection, but after saving the thumbnail we should restore the selection.

stefaniapedrazzi:
Finally, I simplified the code so that it is no longer necessary to change the selection but all the optional renderings, handles, etc. are temporary hidden only from the viewpoint camera.

ThomasOliverKimble (approved):
This is a much more elegant solution. Thank you!

PR #4691, opened 2022-06-20
Make webots run in the Windows CI machine

omichel:
Investigating why `webots --sysinfo` crashes in the Windows CI machines.


omichel:
It turns out that the crash is due to OpenGL initialization, not OpenAL.

Issue #4692, opened 2022-06-20
icon_studio script no longer works with EXTERNPROTO


stefaniapedrazzi:
The `icon_studio` script doesn't work with EXTERNPROTO because it tries to import objects that are not defined as "Ephemeral EXTERNPROTO" and this is forbidden.

PR #4693, opened 2022-06-20
Fix icon studio

ad-daniel:
**Description**
Fixes https://github.com/cyberbotics/webots/issues/4692

- [x] cleanup warnings (unconnected PROTO fields)
- [x] fix incorrect objects
- [x] only allow the user to remove ephemerals through the GUI (node removal should not otherwise attempting to: `spawn object A -> delete A -> spawn object A` won't work because although it would declared for the first spawn, it won't for the second due to the removal)


omichel (approved):
Thank you.

PR #4698, opened 2022-06-21
Fix Qt plugin path on Windows

omichel:
While testing @carikaser experienced a failure at start-up: Webots complained it could not find the Qt plugins and refused to launch. This failure is difficult to reproduce, especially on our development machines (even with a different user account), because it seems that Qt is somehow caching the path to the Qt plugins and manages to find them even though they are not correctly specified by Webots.

This PR fixes the problem (tested on a new PC on which Webots was never installed).

The previous code assumed that `WEBOTS_HOME` was properly defined, which is not the case unless the user explicitly did so.


ygoumaz (approved):
Thank you!

PR #4697, opened 2022-06-21
Fix nightlies on master side

ad-daniel:
**Description**

More debugging regarding the develop nightlies failing if the test is triggered from master

PR #4699, opened 2022-06-21
Fix nightlies on develop side

ad-daniel:
**Description**

Temporary branch on develop side to manually debug: https://github.com/cyberbotics/webots/pull/4697


ad-daniel:
As can be seen here: https://github.com/cyberbotics/webots/pull/4697, the tests appear to finally work so by doing these changes on develop and those in master the nightlies should work. In practice it seems that ever since the cache change the develop scheduled runs never worked in practice since no changes were made on the master side of things to account for it. How it never complained is beyond me.

omichel (approved):
Thank you.

Issue #4694, opened 2022-06-20
Movie creation fail on Ubuntu 22.04 (tarball)


omichel:
`ffmpeg` fails complaining it cannot load `libaom.so.0`. However, this library doesn't existing on Ubuntu 22.04 (was upgraded to `libaom.so.3`). Removing `$WEBOTS_HOME/lib/webots/libavcodec.so.58` fixes the problem. So, I assume there are libraries in the `$WEBOTS_HOME/lib/webots/` folder that are incompatible with Ubuntu 22.04 and break `ffmpeg`. We should probably remove all the `ffmpeg` related libraries from there.

stefaniapedrazzi:
None of the ffmpeg related libraries should be distributed within the tarball.
This is probably a bug introduced in #4549.

ffmpg should be isntalled manually:
https://www.cyberbotics.com/doc/guide/installation-procedure#installing-the-tarball-package

PR #4700, opened 2022-06-21
Cleanup Linux package dependencies

stefaniapedrazzi:
Address #4694.


ad-daniel:
by the way, in the linux workflows there is currently a libzip related hotfix, it might or might not be relevant anymore

stefaniapedrazzi:
Robotis OP2 robot window now opens correctly from both the Debian and tarball package but it cannot be compiled from the tarball.

omichel (approved):
Thank you.

PR #4701, opened 2022-06-21
Fixed font warning on macOS, better default monospaced font

omichel:
This PR fixes a warning showing-up on macOS since the introduction of the EXTERNPROTO.
It also defaults to a better looking font for the text editor and console.


ad-daniel (approved):
You learn something new every day, thank you

PR #4705, opened 2022-06-21
Fix URLs reverted before creating the snap package

stefaniapedrazzi:
URLs were reverted before including the files in the snap package.


stefaniapedrazzi:
I just started the tests but I don't have the time to wait for the results.
If they pass, this PR is ready for review and could be merged today so that we might have working snap packages tomorrow to test.

ad-daniel (approved):
macOS build appears to be failing, but it doesn't seem to be related to any of this (likely cloudflare), as such we I think we can merge it so that tonight the snap packages are built

PR #4706, opened 2022-06-21
Remove webots_ros from sample worlds

ygoumaz:
Worlds from the `webots_ros` submodule are located in the `resources` directory and were added to the sample worlds in #3835. This is not really useful and a bit confusing as these worlds can't work alone. Instead, they should be launched using `roslaunch` directly from the package folder.


omichel (approved):
Thank you.

PR #4707, opened 2022-06-21
Improve Using-ROS tutorial documentation

ygoumaz:
Installing the `webots_ros` package was a bit unclear and not well-structured. The documentation now also mentions installation from sources which should be used to get the last modifications to the `webots_ros` repository.


omichel (approved):
Thank you.

PR #4710, opened 2022-06-22
Fixed crazyflie controller compilation from upper Makefile

omichel:
The crazyflie controller was not compiled from the global Makefile.


ygoumaz (approved):
Thank you. The compilation is now correctly performed from the global Makefile.

PR #4713, opened 2022-06-22
Improved extern controllers documentation

omichel:
Improves documentation to avoid a bad use case experienced by @carikaser.


carikaser (approved):
Everything seems fine to me. Thank you for that clarification.

PR #4723, opened 2022-06-22
Fix test-ros

ygoumaz:
test-ros is not working for the following reasons:
- [x] Webots runtime dependencies are not compiled which leads to a crash from the Webots node directly at the start.
- [x] Python 3.6 is not supported by Webots anymore, therefore the controller is crashing at startup.
- [x] The keyboard test is waiting for an input to continue. It should be ignored in the CI to prevent the test from being blocked indefinitely. (See [#120](https://github.com/cyberbotics/webots_ros/pull/120))

Also, the test is currently not showing any logs when Webots is running, which removes a guarantee that tests are correctly executed. 
- [x] This PR also adds ROS_INFO messages to the console during the `complete_test`.


omichel (approved):
Thank you.

PR #4733, opened 2022-06-23
Fix macOS bundle

omichel:
Fixed macOS bundle:
- [x] Removed symlink in the Webots.app folder.
- [x] Removed launcher from the bundle.
- [x] Fixed missing symlinks in Qt frameworks.
- [x] Signed the bundle before creating the DMG file.
- [x] Fixed paths in runtime.ini files.
- [x] Deleted obsolete runtime.ini files (many were adding paths for the native e-puck or automobile robot windows which are now HTML robot windows).


ThomasOliverKimble (approved):
Looks good to me. Modified controllers are working (on MacOS and Linux, have not tested on Windows), and the .dmg package opens Webots and seems to work well even when downloaded from github.

ps. The Darwin-OP and automobile robot windows do not seem to be working but they didn't work before this PR on mac either, they can be addressed in a separate PR

PR #4737, opened 2022-06-23
Add new workflow `test webots build` label

stefaniapedrazzi:
Add new label to be used by the GitHub Action workflow to compile Webots without generating the packages.

This could maybe speed up the tests.
What do you think? Could this be useful?

If yes, I can complete the implementation for all the workflow files.


omichel:
Yes, that would be useful.
Meanwhile, we should also have a wiki page that explains the actions fired by the tags.

stefaniapedrazzi:
Here is a new page describing the labels:
https://github.com/cyberbotics/webots/wiki/CI-Test-Labels
Also linked from the Git Worklflow page (https://github.com/cyberbotics/webots/wiki/Git-workflow)

I will add it to the wiki home page when this PR is accepted.

stefaniapedrazzi:
`make webots_target -j4` instruction works fine on linux and windows in this PR but not on macOS.
I would need to test on the macOS machine to better debug the issue.

.github/workflows/test_suite_linux_develop.yml
```diff
@@ -12,6 +12,25 @@ defaults:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-20.04, ubuntu-22.04]
+    runs-on: ${{ matrix.os }}
+    steps:
+    - uses: actions/checkout@v3
+      with:
+         submodules: true
+         fetch-depth: 15
```
BenjaminDeleze, line 26:
```suggestion
      with:
         submodules: true
         fetch-depth: 15
         ref: develop
```

.github/workflows/test_suite_mac_develop.yml
```diff
@@ -4,11 +4,27 @@ on:
...
+    runs-on: ${{ matrix.os }}
+    steps:
+    - uses: actions/checkout@v3
+      with:
+         submodules: true
+         fetch-depth: 15
+         ref: develop
+    - name: Build Webots
+      run: |
+        export PATH=/usr/local/bin/:$PATH
```
BenjaminDeleze, line 26: Why do we need to add this to the PATH only for `develop` ?

BenjaminDeleze (changes requested):
Seems good, only some small remarks

PR #4107, opened 2022-01-11
Add `robot_step_begin()` and `robot_step_end()` API functions for controller optimization

ygoumaz:
**Description**
This PR aims to add 2 new API functions in addition to the existing `robot_step()`. The two functions are called `step_begin()` and `step_end()` and are currently hidden from the user. The first one prepares all the useful data for the simulation to perform a step and sends them to Webots. The second one retrieves the data (sensors, etc.) once the rendering by Webots is finished.

Exposing these two functions to the user allows him to separate the use of the `robot_step()` in two parts. All the code executed in between is parallelized with the Webots process. This is particularly useful if both codes (controller and Webots) have a particularly long execution time. The faster of the two processes will simply wait for the second to finish. Currently, with `robot_step()`, both are done sequentially. The implementation of this parallelization adds a new controller optimization option to the user.

**Tasks**
- [x] Add in `/include/controller/c/webots/`.
- [x] Add in `/src/controller/c/` (don't forget to include a "bad device tag" error message if the function takes a WbDeviceTag).
- [x] Add in `/src/controller/c/Controller.def`.
- [x] Add in `/src/webots/nodes/`.
- [x] Add in `/include/controller/cpp/webots/`.
- [x] Add in `/src/controller/cpp/`.
- [x] Add in `/src/controller/java/`, i.e. in Makefile and controller.i.
- [x] Add in `/src/controller/python/` i.e. in Makefile and controller.i.
- [x] Add in `/src/controller/matlab/mgenerate.py`, possibly also in `lib/controller/matlab/` and run `mgenerate.py` with the `-update` option.
- [x] Add in `/src/webots/core/WbLanguage.cpp`.
- [x] Add in `/docs/reference/'device_name'.md`.
- [x] Add a sample usage demo in `/projects/sample/`.
- [x] Add a new API test in /tests/api (new world and controller).

**Documentation**
https://cyberbotics.com/doc/reference/robot?version=enhancement-new-step-func-exports

**Additional context**
In the case of the OPTIMA project, Webots is executed on a machine without GPU, leading to slow camera rendering. On the other side, the robot controller is also very slow, as it includes the forward propagation of a convolutional neural network. Using the two new API functions exposed by this PR, the two processes can be parallelized and much more optimal results can be achieved.

PR #4739, opened 2022-06-23
Fix warnings wb_robot_step() from ros controllers

ygoumaz:
Fixes https://github.com/cyberbotics/webots_ros/issues/108.

`wb_robot_step_begin()` and` wb_robot_step_end()`, introduced in #4107, are displaying a warning if not used in the correct order. These warnings are incorrectly triggered by asynchronous ROS controllers. This PR disables warnings when a ROS controller is connected as these API functions are not available in ROS.


omichel:
Instead of silencing this warning, I believe it is better to make the `wb_robot_step()` function thread-safe and protect it entirely with a mutex, so that if called from different threads, it will wait until the first thread unlock the mutex at the end of the `wb_robot_step()` function before proceeding with the one from the second thread.

src/controller/c/robot_private.h
```diff
@@ -48,6 +48,8 @@ const char *robot_get_device_name(WbDeviceTag tag);
 const char *robot_get_device_model(WbDeviceTag tag);
 void robot_mutex_lock_step();
 void robot_mutex_unlock_step();
+int robot_step_begin(int duration);
+int robot_step_end();
```
omichel, line 52: This is not needed as these functions are used only in the robot module and should be static:
```suggestion
```

src/controller/c/robot.c
```diff
@@ -933,19 +923,44 @@ int wb_robot_step_end() {
...
+  robot_mutex_lock_step();
+  int e = robot_step_begin(duration);
+  robot_mutex_unlock_step();
+
+  return e;
+}
+
+int wb_robot_step_end() {
+  robot_mutex_lock_step();
+  int e = robot_step_end();
```
omichel, line 942:
```suggestion
  const int e = robot_step_end();
```

src/controller/c/robot.c
```diff
@@ -933,19 +923,44 @@ int wb_robot_step_end() {
...
 
+int wb_robot_step_begin(int duration) {
+  if (stdout_read != -1 || stderr_read != -1) {
+    fflush(NULL);  // we need to flush the pipes
+    stream_pipe_read(stdout_read, &(robot.console_stdout));
+    stream_pipe_read(stderr_read, &(robot.console_stderr));
+  }
+
+  robot_mutex_lock_step();
+  int e = robot_step_begin(duration);
```
omichel, line 934:
```suggestion
  const int e = robot_step_begin(duration);
```

src/controller/c/robot.c
```diff
@@ -852,7 +852,7 @@ void wbr_robot_battery_sensor_set_value(double value) {
     robot.battery_value = value;
 }
 
-int wb_robot_step_begin(int duration) {
+int robot_step_begin(int duration) {
```
omichel, line 855:
```suggestion
static int robot_step_begin(int duration) {
```

omichel (changes requested):

omichel (approved):

PR #4714, opened 2022-06-22
Attempt to fix macOS CI failure

omichel:

PR #4741, opened 2022-06-24
Fix macOS workflow on master

omichel:
Backport #4714 on the master branch so that it is executed on schedule and the nightly builds for macOS should be working again.


ygoumaz (approved):
Thank you!

PR #4746, opened 2022-06-24
Fix nightlies develop side

ad-daniel:
**Description**

The first upload job is not executed for the "scheduled" event, so the assets need to be uploaded in the second


omichel (approved):
Thank you.

PR #4747, opened 2022-06-24
Fix test world update

ad-daniel:
**Description**

The first upload job is not executed for the "scheduled" event, so the assets need to be uploaded in the second upload task, hence why the world upload worked for the manual trigger but not with a [scheduled run](https://github.com/cyberbotics/webots/runs/7033478162?check_suite_focus=true)


omichel (approved):
Thank you.

Issue #4720, opened 2022-06-22
EXTERNPROTO listed in add node dialog on linux snap


stefaniapedrazzi:
On Webots installed with the snap package on Ubuntu 22.04, all the EXTERNPROTO are listed (with non-humand-readable names) in the add node dialog:
![Screenshot from 2022-06-22 14-56-16](https://user-images.githubusercontent.com/5910449/175034325-e3cef7ff-ce12-4410-9877-e4aa276ff868.png)

When selecting them, errors are printed in the console.
It is possible to insert them in the scene, even if then textures are not found.

PR #4751, opened 2022-06-24
Fix "current world" list in AddNode

ad-daniel:
**Description**

Fixes: https://github.com/cyberbotics/webots/issues/4720

The "current world" category is supposed to list the PROTO that are referenced in the current world file (i.e. all EXTERNPROTO mentioned in the wbt). If some or all of these url are not local (as is the case for an official distribution), then its name must be inferred by a reverse lookup of the cached file (since the cached file's name is the hash). Also, being an official PROTO means we can skip the info generation step

### DIFF
--- a/src/webots/scene_tree/WbAddNodeDialog.cpp
+++ b/src/webots/scene_tree/WbAddNodeDialog.cpp
@@ -618,6 +618,7 @@ int WbAddNodeDialog::addProtosFromProtoList(QTreeWidgetItem *parentItem, int typ
 
     // populate tree
     if (flattenHierarchy) {
+      printf("%s -> %s\n", it.key().toUtf8().constData(), info->url().toUtf8().constData());
       QTreeWidgetItem *item =
         new QTreeWidgetItem(QStringList() << QString("%1 (%2)").arg(nodeName).arg(baseType) << info->url());
       parentItem->addChild(item);
--- a/src/webots/vrml/WbProtoManager.cpp
+++ b/src/webots/vrml/WbProtoManager.cpp
@@ -325,9 +325,16 @@ void WbProtoManager::generateProtoInfoMap(int category, bool regenerate) {
   const QStringList protos = listProtoInDirectory(category);
   const QDateTime lastGenerationTime = mProtoInfoGenerationTime.value(category);
   foreach (const QString &protoPath, protos) {
-    const QString protoName = QFileInfo(protoPath).baseName();
+    QString protoName;
+    const bool isCachedProto = protoPath.startsWith(WbNetwork::instance()->cacheDirectory());
+    if (isCachedProto)  // cached file, infer name from reverse lookup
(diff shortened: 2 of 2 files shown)


ad-daniel:
Note: the issue doesn't just happen in snap, it can be reproduced by:
```
python3 scripts/packaging/update_urls.py develop
python3 scripts/packaging/generate_proto_list.py develop
```

stefaniapedrazzi:
The text under the section "PROTO nodes (Current World File)" is now human readable and the info are available.

However, it is confusing that there are two different entries for the same PROTO one for the current world and one in the webots projects folder.
It is also not clear what the behavior will be if the two instances will point to different files.
Previously in the Add node dialog, if multiple PROTO (current project and webots projects) had the same name, the item in the webots projects section was grayed out so that it was clear that the one in the current project section was used.

Note that if the URL from the current file and webots projects differs, two different EXTERNPROTO declarations are added in the file causing an error after save and reload.

ad-daniel:
What I meant is if a proto is defined locally, it may appear both in the "current world" category and in the "current project" category, and similar stuff for the others. So the overlap is inevitable and if we try to remove the overlaps by greying things out, in most cases I'm not sure there will anything left to show and trying to differentiate things with categories loses its purpose.

When inserting the 2nd `RectangleArena` the ambiguity should (and I [think](https://github.com/cyberbotics/webots/blob/7df3cada6ba017e569e143f8bf5cb5088dd6a631/src/webots/vrml/WbProtoManager.cpp#L567-L571) is) detected and prevent you from doing so. When loading a "malformed" world, I'm not sure if it does catch it, but I'll double check. In principle ambiguous situations should not be possible.

stefaniapedrazzi:
> What I meant is if a proto is defined locally, it may appear both in the "current world" category and in the "current project" category, and similar stuff for the others. So the overlap is inevitable and if we try to remove the overlaps by greying things out, in most cases I'm not sure there will anything left to show and trying to differentiate things with categories loses its purpose.

If it is the same PROTO, then it is fine to not gray it out and show it in both "current world" and "current project".
However, if default PROTO from the Webots projects is used, I don't see the point in showing it under the "current project" section.

> When inserting the 2nd RectangleArena the ambiguity should (and I [think](https://github.com/cyberbotics/webots/blob/7df3cada6ba017e569e143f8bf5cb5088dd6a631/src/webots/vrml/WbProtoManager.cpp#L567-L571) is) detected and prevent you from doing so. When loading a "malformed" world, I'm not sure if it does catch it, but I'll double check. In principle ambiguous situations should not be possible.

No, it is not detected and the file resulted having two different EXTERNPROTO declarations for the same world.

stefaniapedrazzi:
> If it is the same PROTO, then it is fine to not gray it out and show it in both "current world" and "current project".

Actually, I think we should just define some priorities: for example if it is local and extern, show it only under the "current local proto" section.

Having the same PROTO listed under multiple sections is not very user-friendly and brings some confusion.

Issue #4755, opened 2022-06-27
Adding PROTO nodes in Windows nightly build does not work


carikaser:
Impossible to create PROTO nodes on Windows. The PROTO nodes categories are gone and Webots cannot find the PROTO nodes when trying to add them to a world.

PR #4760, opened 2022-06-27
Fix windows nightly

ad-daniel:
**Description**
Fixes https://github.com/cyberbotics/webots/issues/4755

On windows nightlies, the proto-list.xml url were generated as:
`<url>D:\a\webots\webots\projects\appearances\protos\Asphalt.proto</url>`
instead of a `https://raw.github..` hence the reported issue.

The culprit appears to be [this line](https://github.com/cyberbotics/webots/blob/7df3cada6ba017e569e143f8bf5cb5088dd6a631/scripts/packaging/generate_proto_list.py#L198) where we are supposed to replace the absolute path of the PROTO assets with something like `prefix/path/to/asset`. Since windows might be using backslashes instead of forward slashes, the replace command `replace(WEBOTS_HOME + '/', prefix)` would fail (since WEBOTS_HOME itself might be: `c:\this\and\that` therefore the string `c:\this\and\that/` would not be found and the raw path would be kept and distributed

### DIFF
--- a/.github/workflows/test_suite_windows.yml
+++ b/.github/workflows/test_suite_windows.yml
@@ -3,8 +3,8 @@ name: Test Suite Windows
 on:
   pull_request:
     types: [opened, synchronize, reopened, labeled, unlabeled]
-  schedule:
-    - cron:  '0 23 * * *'
+  # schedule:
+  #   - cron:  '0 23 * * *'
   push:
     tags:
       - nightly_*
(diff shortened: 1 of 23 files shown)


ad-daniel:
Locally it seems to do the trick, I'll wait for the tests to pass before merging but it should be fairly safe

omichel (approved):
Thank you.

PR #4748, opened 2022-06-24
Fixed hang after copy/paste of robots on Linux

omichel:
Fixes #4716.

PR #4765, opened 2022-06-27
Fixed camera reset

omichel:
There was a bug introduced in #4748 causing a robot camera to crash Webots when resetting a simulation.
This PR fixes this crash.


ygoumaz (approved):
Works fine again! Thanks.

PR #4774, opened 2022-06-28
Add Panda splash-image

ygoumaz:
Add the new Panda splash-image and adapt other splash-image to match the 580x580 px resolution.


omichel (approved):
Nice. Thank you.

PR #4775, opened 2022-06-28
Improve error message

omichel:
This error message is bit weird:
```
Error: wb_supervisor_field_get_mf_node() called with an out-of-bound index: 0 (should be between 0 and -1).
```
This PR propose to replace it with:
```
Error: wb_supervisor_field_get_mf_node() called on an empty list.
```


ygoumaz (approved):
Thank you!

PR #4776, opened 2022-06-28
Hotfix applying PROTO changes on reload

### LINKED ISSUE #4773: Local PROTO changes not applied on world reload
It seems that on the develop branch, the local PROTO models are not read again from file when a world is (re-)loaded.
Thus, changes to PROTO files are not applied to the current world until Webots is restarted.

It seems fine to keep and do not reload all the EXTERN PROTOs, but at world load we should clean the list of local PROTOs.

stefaniapedrazzi:
Fixes #4773.
If a PROTO model file or an EXTERNPROTO declaration in world file is changed, we need to reload the corresponding PROTO models from files when reloading the world otherwise the change are never applied to the world until Webots is restarted.


stefaniapedrazzi:
Two cache tests were no longer working after this bug fix:
* `web_proto_with_texture.wbt`: use a PROTO with hard coded wed URL for texture (develop branch)
* `local_proto_with_texture.wbt`:
    This world  (and `ShapeWithLocalTexture.proto`) was not using local files but Webots remote (web) default files were used.
    I changed the URLs and used relative path to local files.

ad-daniel:
FYI, `local` in these context is an url of type `webots://` specifically. It's what `WbUrl::isLocalUrl` calls them. Essentially there's 4 ways PROTO can be defined, and we need to test all 4:
- web: `https://...`
- local: `webots://...`
- relative: `../..`
- absolute: `/../../` 

In practice yes, a `webots://` url is turned into an absolute one (or we could turn it into a relative one), but when in the webots development environment, they're provided as `webots://` so a test for this specific case is needed

ad-daniel:
note: the `update_url.py` should not apply to the `cache` tests, this category is handled exclusively by `setup_environment`. I don't know if with the recent changes this is still the case, but there should be an exception in `update_url`

stefaniapedrazzi:
I will check again.
But currently it works correctly locally, but when testing in the CI the URLs expected to be local are translated to default Webots web URLs ( using the same tag as the release one).

I don't think anything changed. This is the expected behavior when using a released package as in the tests.
It is not the `update_url.py` scripts that translated the `webots://` links to web ones but Webots itself.

So if we want to test local URL we have to explicitly use local URLs.

stefaniapedrazzi:
> In practice yes, a webots:// url is turned into an absolute one (or we could turn it into a relative one), but when in the webots development environment, they're provided as webots:// so a test for this specific case is needed

As it is, the local test will never work in the CI: the world will always load a  `ShapeWithWebTexture.proto` PROTO with invalid URL.

Also note that relative path are never tested.
I can change it and mix the two.

ad-daniel:
`relative_proto_with_texture.wbt` should be testing the relative ones: https://github.com/cyberbotics/webots/blob/develop/tests/cache/worlds/relative_proto_with_texture.wbt

But yes, the `local_proto_with_texture.wbt` cannot work in the CI, it currently does work because `update_url` changes it to a remote one. It's a tests specific to the webots development environment, a user should never encounter a world that has `webots://` urls (not at the root, nested protos will have them but a remote url is manufactured on the spot based on the parent proto). In short, we may disable it in the CI and keep it for when the test suite is run locally.

stefaniapedrazzi:
Note that the local test works in the CI only because the current version is buggy and the correct Proto is not loaded.

ad-daniel:
All the cache tests that involve local urls (`webots://`) cannot really work in the CI because a webots:// url references locally available proto which in practice aren't distributed and therefore not available in the CI. The CI worked so far because models with the same name were already present in memory which shouldn't be and was fixed in this PR and partially because the `update_urls` script changed these local references to remote ones.

In theory we can make the test "pass" by changing the urls, but what we would be testing isn't what the test is supposed to check so having it run on the CI this way is misleading.

The test is still rather important because it reflects the behavior we can expect in the webots development environment, so at least locally it should be run. As such, I've done some modifications:
- I've moved all local-url related tests to the same world
- this world (`local_test_with_texture.wbt`) is skipped in the CI (later we should perhaps add an additional job that just runs the cache test from the development environment so that it gets tested consistently
- changing of urls in the `cache` directory is done exclusively by `cache_environment.py` and the directory is skipped by `update_url`

A second review of this PR would be welcome

ad-daniel:
This PR should be prioritized as others depend on it

ThomasOliverKimble:
I am getting errors with `tests/cache/backwards_compatibility.wbt`.
It seems as because `TestSuiteEmitter` and `TestSuiteSupervisor` are not official PROTOS, the backwards compatibility mechanism gives the following errors:
```
ERROR: PROTO 'TestSuiteEmitter' is not a known Webots PROTO. The backwards compatibility mechanism may fail.
ERROR: PROTO 'TestSuiteSupervisor' is not a known Webots PROTO. The backwards compatibility mechanism may fail.
```

(also minor detail, the pallets have the same name, so we could change that as well)

omichel (approved):
Thank you.

PR #4782, opened 2022-06-28
Added template.proto to distributions

### LINKED ISSUE #4729: Broken PROTO Wizard on nightly snap package
**Describe the Bug**
Creating a new PROTO with the PROTO Wizard does not work and gives the error: "Some directories or files could not be created."

**Steps to Reproduce**
Install 20-06-2022 nightly .snap file, start webots, create new Proto via the Proto Wizard, configure, finish

**Webots Version**
Snap package nightly build 20-06-2022

ygoumaz:
Fixes #4729.

The PROTO wizard is not working in distributed packages because `template.proto ` is not distributed. The PR also removes the .cache file verification and copy when creating a distribution.

### DIFF
--- a/scripts/packaging/generic_distro.py
+++ b/scripts/packaging/generic_distro.py
@@ -124,12 +124,6 @@ class WebotsPackage(ABC):
         if not os.access(local_dir_path, os.F_OK):
             print_error_message_and_exit(f"Missing dir: {dir_name}")
 
-    def compute_md5_of_file(self, filename):
-        with open(filename, 'rb') as file_to_check:
-            # read contents of the file and compute md5sum
-            data = file_to_check.read()
-            return hashlib.md5(data).hexdigest()
-
     @abstractmethod
     def make_dir(self, directory):
         pass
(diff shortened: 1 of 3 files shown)


scripts/packaging/generic_distro.py
```diff
@@ -185,26 +185,9 @@ def add_file(self, file_path):
             self.package_files.append(os.path.relpath(world_project_file, self.webots_home))
 
         if file_path.endswith('.proto') and str(os.path.sep + 'protos' + os.path.sep) in file_path:
-            # compute MD5 fo PROTO file
             if not os.path.exists(absolute_path):
                 print_error_message_and_exit(f"Missing file: {absolute_path}")
 
-            md5Value = self.compute_md5_of_file(absolute_path)
```
ad-daniel, line 192: Since it no longer appears to be necessary, I think we can also clean up the `compute_md5_of_file` function

ad-daniel (changes requested):
Looks good, just a detail

ad-daniel (approved):
Thank you

PR #4787, opened 2022-06-28
Fixed relocation of projects

### LINKED ISSUE #4772: Controller not found after relocation
**Describe the Bug**

In the latest nightly (tar), relocating the `highway_driving` sample prints an additional warning concerning the controller. Since this warning was not present in this issue: https://github.com/cyberbotics/webots/issues/4726, I suspect the issue was recently added.

```
ERROR: JavaScript error: could not open file: /home/daniel/Downloads/highway_driving/worlds/forest/highway_overtake/1.forest.
WARNING: DEF WEBOTS_VEHICLE0 LincolnMKZ (PROTO): The 'automobile' robot window library has not been found.
WARNING: DEF WEBOTS_VEHICLE0 LincolnMKZ (PROTO): The controller directory has not been found, searched the following locations:
/home/daniel/Downloads/highway_driving/controllers/highway_driving/
/home/daniel/Downloads/webots-R2022b-x86-64/webots/projects/vehicles/protos/lincoln/controllers/highway_driving/
/home/daniel/Downloads/webots-R2022b-x86-64/webots/projects/default/controllers/highway_driving/
/home/daniel/Downloads/webots-R2022b-x86-64/webots/resources/projects/controllers/highway_driving/
```

omichel:
Fixes  #4772.

This PR fixes the missing copies of controller files as well as SUMO files.

### DIFF
--- a/src/webots/editor/WbProjectRelocationDialog.cpp
+++ b/src/webots/editor/WbProjectRelocationDialog.cpp
@@ -324,7 +324,7 @@ int WbProjectRelocationDialog::copyWorldFiles() {
   } else
     setStatus(tr("Impossible to read file '%1'").arg(world->fileName()));
 
-  // copy net directory if any
+  // copy SUMO net directory if any
   QString fileName = world->fileName();
   const QString netDir = fileName.replace(".wbt", "_net");
   if (QDir().exists(netDir))
--- a/src/webots/editor/WbProjectRelocationDialog.cpp
+++ b/src/webots/editor/WbProjectRelocationDialog.cpp
@@ -324,7 +324,7 @@ int WbProjectRelocationDialog::copyWorldFiles() {
   } else
     setStatus(tr("Impossible to read file '%1'").arg(world->fileName()));
 
-  // copy net file if any
+  // copy net directory if any
   QString fileName = world->fileName();
   const QString netDir = fileName.replace(".wbt", "_net");
   if (QDir().exists(netDir))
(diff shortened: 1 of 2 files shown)


src/webots/editor/WbProjectRelocationDialog.cpp
```diff
@@ -224,67 +224,29 @@ void WbProjectRelocationDialog::copy() {
...
-  if (copyProtoProject) {
-    // copy all PROTO textures
-    result += WbFileUtil::copyDir(projectPath + "protos/textures", mTargetPath + "/protos/textures", true, true, true);
-    result += WbFileUtil::copyDir(projectPath + "protos/meshes", mTargetPath + "/protos/meshes", true, true, true);
-    result += WbFileUtil::copyDir(projectPath + "protos/skins", mTargetPath + "/protos/skins", true, true, true);
-  }
-
-  bool projectLibrariesCopied = false;
+  // copy all local resources
+  result += WbFileUtil::copyDir(projectPath + "protos", mTargetPath + "/protos", true, false, true);
```
ad-daniel, line 231: More of a question really, does the project relocation action trigger only for our own folders? If so can't we skip copying the protos folder? It should be empty no? (well, technically it shouldn't be there to begin with after the cleanup of the distribution)

ad-daniel (changes requested):
Works well, just a question

src/webots/editor/WbProjectRelocationDialog.cpp
omichel: Yes, good point. I can remove this copy of this `protos` folder indeed.

omichel: Finally, I would prefer to keep it like that for the following reasons:
1. At some point we may include a protos folder in our example to demonstrate how to have a local PROTO.
2. Some people may fork Webots and add their own projects which may include a protos folder.
3. Some people may distribute some project folder which installs without write rights, and thus users would need to relocate it, including possibly a protos folder.

ad-daniel (approved):
Thank you

PR #4795, opened 2022-06-29
Fix crash in triangle mesh cache

### LINKED ISSUE #4450: Crash when selecting field in scene tree
**Describe the Bug**
Webots crashes when attempting to select a url field.

**Steps to Reproduce**
1. Open the following world
```
#VRML_SIM R2022a utf8
WorldInfo {
}
Viewpoint {
}
Background {
  skyColor [
    0.15 0.45 1
  ]
}
Transform {
  children [
    Shape {
      geometry Mesh {
      }
    }
  ]
}
Transform {
  children [
    Shape {
      geometry Mesh {
      }
    }
  ]
}

```
3. Expand the first transform (only this one), and add a random mesh to the Mesh node 
4. Expand the second Transform, when you attempt to select the `url` field Webots crashes.

Note: it hardly crashes on the first, primarily on the second transform

omichel:
Fixes #4450.

The life cycle of the `WbTriangleMesh` object wasn't correct: it is created and deleted in `WbTriangleMeshGeometry`, but also deleted in `WbTriangleMeshCache` without updating the pointer in `WbTriangleMeshGeometry`. Hence after deleting the `WbTriangleMesh` in `WbTriangleMeshCache`, `WbTriangleMeshGeometry.mTriangleMesh` was pointing to some memory that was released. Therefore the `WbTriangleMesh` pointed to by `WbTriangleMeshGeometry` was containing corrupted data.

I believe there is no need to delete this `WbTriangleMesh` object in `WbTriangleMeshCache`, as it doesn't belong to `WbTriangleMeshCache`.


ad-daniel (approved):
Thank you

omichel:
I believe this is not the right, as there are still some memory leaks and dangling pointers. I am working on a better fix...

omichel:
The problem was related to empty meshes that were marked as "invalid" and hence could not be cached properly. This fix marks empty meshes as valid, so that they can be cached properly.

ad-daniel (approved):
Thank you

PR #4796, opened 2022-06-29
Fix empty url

ad-daniel:
**Description**

Setting a empty `""` url used to show a warning message and it also interrupted the asset loading process (since there's nothing load). Neither happens currently


omichel (approved):
Thank you.

PR #4800, opened 2022-06-29
Fix robot window mutex lock

ygoumaz:
#4739 introduced a fully thread safe `wb_robot_step` function which is locked from start to end. The problem is that robot windows (called from the step function) are also trying to call API functions, causing a double recursive mutex lock from the same thread. 

The PR solves the problem by allowing recursive lock of the mutex by the same thread on Linux and macOS, which was already possible on Windows. Also, the mutex lock and unlock functions are renamed for clarity.


omichel (approved):
Looks good to me.
Thank you.

PR #4789, opened 2022-06-29
Add 'clean-urls' target to main Makefile

stefaniapedrazzi:
Add target to reset all the assets URLs to the format `webots://`.

PR #4801, opened 2022-06-29
Debug test-suite CI wrong url

ad-daniel:
**Description**

As it emerged on another PR, during the test suite some urls seem to target R2022b which isn't released yet, so more info is needed to know why and where this happens.


ad-daniel:
The tests were failing because in https://github.com/cyberbotics/webots/pull/4789 two exceptions were added to the update_urls script, which meant that the EXTERNPROTO references weren't being converted either. The exception now only applies to the body, not the declarations

Issue #4814, opened 2022-06-30
OSM importer and SUMO exporter broken URLs


ThomasOliverKimble:
**Describe the Bug**
OSM importer and SUMO exporter produce broken PROTO URLs, probably due to the fact that EXTERN PROTOS are not defined in the created world. In the test described below, the `village.wbt` is created, and we get the following errors:

`ERROR: Could not open file: '/Applications/Webots.app/https:/raw.githubusercontent.com/cyberbotics/webots/d0149b322711129b0aa1e241f50515c5dd43badf/projects/objects/trees/protos/SimpleTree.proto'.`

The script that writes the world can be found here: [$WEBOTS_HOME/https://github.com/cyberbotics/webots/blob/master/resources/osm_importer/utils/misc_utils.py](https://github.com/cyberbotics/webots/blob/master/resources/osm_importer/utils/misc_utils.py).
It does not take EXTERN PROTO into account.

**Test Performed to Reproduce Bug**
1. Install the python dependencies as explained in https://www.cyberbotics.com/doc/automobile/openstreetmap-importer#dependencies

2. Convert the OSM file:

For Linux:
mkdir -p $HOME/my_project/resources/osm_files
mkdir -p $HOME/my_project/resources/OSM_importer
cp $WEBOTS_HOME/projects/vehicles/importer_demo.sh $HOME/my_project/importer_demo.sh
cp $WEBOTS_HOME/resources/osm_importer/config.ini $HOME/my_project/resources/OSM_importer/config.ini
cp $WEBOTS_HOME/projects/vehicles/resources/osm_files/village.osm $HOME/my_project/resources/osm_files/village.osm
cd $HOME/my_project
chmod u+x ./importer_demo.sh
./importer_demo.sh village

For macOS:
mkdir -p $HOME/my_project/resources/osm_files
mkdir -p $HOME/my_project/resources/OSM_importer
cp $WEBOTS_HOME/Contents/projects/vehicles/importer_demo.sh $HOME/my_project/importer_demo.sh
cp $WEBOTS_HOME/Contents/Resources/osm_importer/config.ini $HOME/my_project/resources/OSM_importer/config.ini
cp $WEBOTS_HOME/Contents/projects/vehicles/resources/osm_files/village.osm $HOME/my_project/resources/osm_files/village.osm
cd $HOME/my_project
chmod u+x ./importer_demo.sh
./importer_demo.sh village

3. Open the generated world file

PR #4815, opened 2022-06-30
Fix osm world creator

ad-daniel:
**Description**

Fixes: https://github.com/cyberbotics/webots/issues/4814

Currently doesn't work because R2022b is not available yet so the sub-proto cannot be retrieved recursively. Can be tested using `develop` instead.

Requires https://github.com/cyberbotics/webots/pull/4792 to be merged first to fix the second issue mentioned


omichel (approved):
Thank you.

PR #4825, opened 2022-07-04
Update remaining urls to R2022b

BenjaminDeleze:
Note that these urls will not be valid until the tag `R2022b` is created,


omichel (approved):
Thank you.

PR #4826, opened 2022-07-04
Improve field editor visibility

ygoumaz:
The field editor is currently hidden by default in many sample worlds and it is difficult for the user to find a way to unhide it to edit fields in the scene tree.

This PR ensures that when an item is selected in the scene tree, the field editor is visible. If no field is selected and the field editor shows an empty gray frame, it is automatically collapsed. The user can always collapse the editor manually to get a better visibility of the scene tree.


BenjaminDeleze (approved):
I played a bit with it and it seems to work well!

PR #4828, opened 2022-07-04
Fixed font in ephemeral proto warning

### LINKED ISSUE #4823: Text of EPHERMAL PROTO not really readable
As you can see on the picture, the description text before the `Insert new` button is cropped and the no slider is visible (even so we can scroll the text).

I have the same problem in both night and classic theme
display_bug

omichel:
Fixes #4823:

image

(Tested on Windows and Linux with various themes)


BenjaminDeleze (approved):
Thank you!

PR #4842, opened 2022-07-05
Fixed MATLAB on macOS

omichel:
Fixes the execution of MATLAB controllers on macOS.


BenjaminDeleze (approved):
Thank you

BenjaminDeleze (approved):
Indeed seems better!

PR #4846, opened 2022-07-05
Fix guided tour crash in auto mode

omichel:
In addition to fixing a random crash with the auto mode of the guided tour, this also leaves the current running mode (pause, real-time or fast) across all the guided tour worlds, which is more user friendly.


ygoumaz (approved):
I reduced the transparency of the guided tour window for a better user experience. Feel free to check if this works for you. Otherwise, the tour seems to work fine now.

PR #4847, opened 2022-07-05
Fix inverted sound

### LINKED ISSUE #4840: Speaker left and right are inverted
**Describe the Bug**
The sound from the left and right of the speaker is inverted.
This can be seen by opening the speaker.wbt world in the devices tour.
Observed on macOS, Linux (snap) and Windows.

ThomasOliverKimble:
**Description**
Fixes #4840. Does not seem like the most elegant fix, but it works.
Not sure if the problem was caused by FLU conversion: I tried switching the x, y and z values for position, sound and direction to match the conversion but it didn't work. 
Any other suggestions are appreciated!


omichel (approved):
Nice 1-character fix!

PR #4851, opened 2022-07-06
Fix open controller in text editor

ygoumaz:
The `Edit` button to open a controller in the text editor is not working for PROTOs in distributions because of their changed URLs (raw.githubusercontent.com/...).


ad-daniel (approved):
Appears to do the job, thank you

PR #4855, opened 2022-07-06
Add script for automatic declaration of EXTERNPROTO

ad-daniel:
**Description**

Adds script that converts a project prior to R2022b to R2022b.

To test:
1. install an official binary
2. copy a random robot from the distributed ones to the desktop
3. for all worlds and PROTO in the project: remove any existing EXTERNPROTO references, and change the version from `R2022b` to `R2022a`.
4. download the script at the root of the project (same level as `worlds`, and `protos` folders)
5. run it with: `python3 declare_externproto.py /path/to/webots/installation`
6. load the world using webots


omichel (dismissed):
Thank you.

Issue #4852, opened 2022-07-06
Worlds with warnings


BenjaminDeleze:
I reran the `test_worlds_warnings.wbt` with a longer timeout and on the nightly tar.gz and it says that the following worlds have warnings:

- projects/vehicles/worlds/CH_Morges.wbt
- projects/vehicles/worlds/village_center.wbt
- projects/samples/howto/url/worlds/url.wbt
- projects/samples/howto/rolling_friction/worlds/rolling_friction.wbt
- projects/robots/adept/pioneer3/worlds/pioneer3at.wbt
- projects/robots/adept/pioneer3/worlds/pioneer3dx_collision_avoidance.wbt

PR #4856, opened 2022-07-06
Fix world warnings

ad-daniel:
**Description**

Fixes: https://github.com/cyberbotics/webots/issues/4852

If we allow `update_urls` to revert the urls in the body of this world, it will always create a diff. The proper solution would be to add an exception on top of the existing exception, but isn't worth it. This fix doesn't alter what the sample world is supposed to show.

Fixed two controller programs from the sample/howto folder that weren't compiled.


omichel (approved):
Thank you.

PR #4860, opened 2022-07-06
Rework findModel logic and fix several bugs

ad-daniel:
**Description**

Well, turned out to be a rework of almost everything. Fundamentally the issues were:
1. Only PROTO declared in the world were downloaded on load. As such, if you had a local PROTO that depended on a remote one and this PROTO wasn't used by the world (i.e., not cached), when opening the add-node dialog it was impossible to generate the node info from it.
   - The solution is that prior to opening the dialog (idem for proto wizard dialog and ephemeral PROTO dialog), all the dependencies must be downloaded.
   - As such, the first time you open the dialogs, it might be slightly slower
2. PROTO declarations were enforced when retrieving the assets (i.e., when building the PROTO tree) but not necessarily used when loading the world itself. The closest PROTO ancestor was used to manufacture the url of a PROTO child.
   - Now references to the parent file are propagated while building the node structure, and the location is read directly from there.
3. Due to how the urls were manufactured, if you had a local PROTO like `
```
PROTO MyProto [
]
{
  Tesla {
  }
}
```
Where Tesla references local textures (`webots://`, since the PROTO is an official one and therefore cached).
When manufacturing the url for the textures, the closest ancestor found was MyProto, not Tesla, which being a local file meant that it assumed the textures were also local, which they weren't.

Additionally, several other bugs were solved:
- when opening the addnode dialog multiple times, the `Current World` category was visible, then not, then visible, ...
- replaced the backwards compatibility mechanism: if the world is < R2022b, an error message links to the procedure. Since Webots can't write PROTO, the previous approach did only half of the job and if local PROTO were present, it resulted in many error messages, which wasn't very pleasant for the user. Now based on the local and official PROTO, the missing declarations are injected which allows the world to be loaded, but only when the user follows the procedure the message will disappear.
- Ephemeral PROTO list not refreshing correctly
- some items of proto-list.xml were parsed and loaded incorrectly


ygoumaz:
I have assigned two reviewers but I will definitely test the PR on my side too.

src/webots/nodes/utils/WbUrl.cpp
```diff
@@ -302,3 +218,18 @@ const QString WbUrl::computeLocalAssetUrl(const WbNode *node, QString url) {
...
     return WbUrl::computePath(node, "url", url, false);
 }
+
+const QString WbUrl::computePrefix(const QString &rawUrl) {
+  QString url = rawUrl;
+  if (url.startsWith(WbNetwork::instance()->cacheDirectory()))
+    url = WbNetwork::instance()->getUrlFromEphemeralCache(url);
+
+  if (isWeb(url)) {
+    QRegularExpression re("(https://raw.githubusercontent.com/cyberbotics/webots/[a-zA-Z0-9\\_\\-\\+]+/)");
```
stefaniapedrazzi, line 228: This is a general comment and doesn't have to be addressed in this branch.
However this regular expression is repeated 4 times in the code and the rawgit URL is hard coded about 10 times.
(same for the `webots://`).
This is not very good and hardcoded  text and common manipulations of URLs should be defined in a single place (for example here in WbUrl class).

src/webots/scene_tree/WbInsertExternProtoDialog.cpp
```diff
@@ -133,7 +144,12 @@ void WbInsertExternProtoDialog::accept() {
   }
 
   // the addition must be declared as EXTERNPROTO so that it is added to the world file when saving
+  if (mPath.startsWith(WbStandardPaths::webotsHomePath()))
+    mPath = mPath.replace(WbStandardPaths::webotsHomePath(), "webots://");
+  if (mPath.startsWith(WbProject::current()->protosPath()))
+    mPath = QDir(WbProject::current()->worldsPath()).relativeFilePath(mPath);
   WbProtoManager::instance()->declareExternProto(mProto, mPath, true);
+  WbLog::info(tr("You need to save the world for the declaration to be taken into consideration."));
```
stefaniapedrazzi, line 152: I'm not sure if a user reading this message in the console will understand what declaration it refers to.
Probably it would be better to add it in the text. For example
```suggestion
  WbLog::info(tr("You need to save the world for the EXTERNPROTO declaration to be taken into consideration."));
```

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -105,85 +107,116 @@ void WbProtoManager::readModel(WbTokenizer *tokenizer, const QString &worldPath)
...
+    if (WbUrl::isWeb(parentFile)) {
+      QRegularExpression re("(https://raw.githubusercontent.com/cyberbotics/webots/[a-zA-Z0-9\\_\\-\\+]+/)");
+      QRegularExpressionMatch match = re.match(parentFile);
+      if (match.hasMatch()) {
+        // inject the prefix based on that of the parent
+        modelPath = protoDeclaration.replace("webots://", match.captured(0));
+        // if the PROTO tree was built correctly, by definition the child must be cached already too
+        assert(WbNetwork::instance()->isCached(modelPath));
+      } else {
+        WbLog::error(tr("The cascaded url inferring mechanism is supported only for official webots assets."));
```
stefaniapedrazzi, line 168:
```suggestion
        WbLog::error(tr("The cascaded URL inferring mechanism is supported only for official Webots assets."));
```

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -105,85 +107,116 @@ void WbProtoManager::readModel(WbTokenizer *tokenizer, const QString &worldPath)
...
+        return NULL;
+      }
+    } else
+      modelPath = WbUrl::computePath(protoDeclaration);
+  } else
+    modelPath = WbUrl::computePath(protoDeclaration);
```
stefaniapedrazzi, line 174: This could also maybe be simplified by
```suggestion
    }
  }
  if (modelPath.isEmpty())
    modelPath = WbUrl::computePath(protoDeclaration);
```

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -105,85 +107,116 @@ void WbProtoManager::readModel(WbTokenizer *tokenizer, const QString &worldPath)
...
+
+  // determine prefix and disk location from modelPath
+  const QString modelDiskPath = WbUrl::isWeb(modelPath) ? WbNetwork::instance()->get(modelPath) : modelPath;
+  const QString prefix = WbUrl::computePrefix(modelPath);  // used to retrieve remote assets (replaces webots:// in the body)
 
-    WbProtoModel *model = readModel(QFileInfo(url).absoluteFilePath(), worldPath, url, baseTypeList);
+  if (QFileInfo(modelDiskPath).exists() && !modelPath.isEmpty()) {
```
stefaniapedrazzi, line 180: I would switch the conditions and list first the quicker to check to 
```suggestion
  if (!modelPath.isEmpty() && QFileInfo(modelDiskPath).exists()) {
```

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -105,85 +107,116 @@ void WbProtoManager::readModel(WbTokenizer *tokenizer, const QString &worldPath)
...
   return NULL;
 }
 
+QString WbProtoManager::findExternProtoDeclarationInFile(const QString &url, const QString &modelName) {
+  if (url.isEmpty())
+    return QString();
+
+  QString path = url;
+  if (WbUrl::isWeb(path))
+    path = WbNetwork::instance()->get(path);
```
stefaniapedrazzi, line 198: This could be also done in one line:
```suggestion
  const QString path = WbUrl::isWeb(path) ? WbNetwork::instance()->get(url) : url;
```

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -105,85 +107,116 @@ void WbProtoManager::readModel(WbTokenizer *tokenizer, const QString &worldPath)
...
+
+  QFile file(path);
+  if (!file.open(QIODevice::ReadOnly)) {
+    WbLog::error(tr("Could not find declarations because file '%1' is not readable.").arg(url));
```
stefaniapedrazzi, line 202: This error is not very clear. What "declarations" are not found?

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -753,3 +823,36 @@ void WbProtoManager::cleanup() {
...
+    }
+
+    if (WbUrl::isLocalUrl(url)) {
+      url = QDir::cleanPath(WbStandardPaths::webotsHomePath() + url.mid(9));
+      if (QFileInfo(url).exists())
+        return url;
+    }
+
+    return QString();
+  }
```
stefaniapedrazzi, line 855: This is redundant
```suggestion
  }
```

stefaniapedrazzi (changes requested):
I started checking the code and I will now test the functionality

BenjaminDeleze:
I tested on Ubuntu 22 in the development environment and everything seems ok.
Now testing on a distrib

stefaniapedrazzi:
Would not be possible to specify where the EXTERNPROTO declarations are missing?
Because in complex world having this message without any reference to a file is not very useful:
``ERROR: PROTO declarations are missing. Please adapt your project to R2022b following these instructions: https://github.com/cyberbotics/webots/wiki/How-to-adapt-your-world-or-PROTO-to-Webots-R2022b``
For example in my case, everything (worlds, PROTO files) but a single PROTO file was not correct.

ThomasOliverKimble:
After performing some tests, so far I have noticed two errors:

1. Copy paste PROTO out of context warning:
    To reproduce:
        - Download webots from source and make while on this PR's branch
        - Create new project with the wizard
        - Add a PROTO via the + button (beer bottle for example)
        - Copy and Paste PROTO with keyboard (ctrl + c, ctrl + v)
    I get the following warning: "ERROR: PROTO declarations are missing. Please adapt your project to R2022b following these instructions: https://github.com/cyberbotics/webots/wiki/How-to-adapt-your-world-or-PROTO-to-Webots-R2022b"

2. Make distrib local URLs
    To reproduce:
        - Download webots from source
        - run the following line as from webots home: python scripts/packaging/set_commit_and_date_in_version.py $(git log -1 --format='%H') (found in test_suite_linux,yml)
        - Create new project with the wizard
        - Add a PROTO via the + button (thymio for example)
        - Save world
        - View PROTO source file in a text or code editor
    Externproto URLs are still of type "webots://", but world file contains remote paths.

ad-daniel:
> Because in complex world having this message without any reference to a file is not very useful:
ERROR: PROTO declarations are missing. Please adapt your project to R2022b following these instructions: https://github.com/cyberbotics/webots/wiki/How-to-adapt-your-world-or-PROTO-to-Webots-R2022b

Technically yes, but every missing declaration would print a similar message which on a somewhat complex world becomes unmanageable. Fundamentally the problem I see is that at the beginning, a user might not know how to fix it even if he knew where it was missing (he might be able to declare a local PROTO, but for a remote one he'd need to read the wiki regardless to know how to find the url of an official one). The point of the script was to make the process more effortless. Not sure what would be a good middle ground.

BenjaminDeleze:
> 1. Copy paste PROTO out of context warning:
>    To reproduce:
>    - Download webots from source and make while on this PR's branch
>    - Create new project with the wizard
>    - Add a PROTO via the + button (beer bottle for example)
>    - Copy and Paste PROTO with keyboard (ctrl + c, ctrl + v)
>    I get the following warning: "ERROR: PROTO declarations are missing. Please adapt your project to R2022b following these instructions: https://github.com/cyberbotics/webots/wiki/How-to-adapt-your-world-or-PROTO-to-Webots-R2022b"

I could reproduce this only if I do not save between the `add` and the `copy/paste`
And I found another issue:

- Build webots in dev environment
- Create a new project (with the wizard)
- Create a new proto (with the wizard), we will call it TEST
- Add TEST as an ephemeral EXTERNPROTO
- Save the world
- Add TEST to the world
- Remove it from the ephemal list
- Save
- Try to copy/paste TEST => `ERROR: In order to import the PROTO 'test15', first it must be declared in the Ephemeral EXTERNPROTO list and the world must be saved.`
- Close Webots and reopen it: `ERROR: In order to import the PROTO 'test15', first it must be declared in the Ephemeral EXTERNPROTO list and the world must be saved.`

Strange enough I do not have the warning when reloading the world


Also it is normal that when adding a PROTO to a world, before saving, the PROTO is also added and visible in the ephemeral list?

stefaniapedrazzi:
Here are other issues that I found (still testing):
1. In `projects/samples/howto/texture_change` the stone texture is still not found.
    It seems like the controller is compiled before modifying the URL and thus everything works correctly only after relocating and recompiling the controller. 
    This error is not strictly related with this PR, but the `webots://` URL is not recognized by Webots in this PR.
    
2. When inserting in the list of Ephemeral EXTERNPROTO a PROTO already declared in the file nothing happens.
    This behavior seems buggy because it is not clear why this PROTO has not been added to the list.
    I would expect it to be possible to add it to the ephemeral list even if it is already declared, otherwise it should be grayed out in the list of PROTO.
    If not, then at least a message in the console should be printed explaining the current PROTO is already declared in the world file.

3. When testing adding a node from a supervisor controller, if you add the new node as ephemeral, run the controller, save the world but cancel the saving when asked to choose for a new name (simulation has run):
  -> the new node PROTO declaration is removed from the ephemeral list but not added to the world

I have many crashes on world reload when testing the tarball on Ubuntu 2022b. I will test then again in the development environment.

stefaniapedrazzi:
Here are some other issues:
4. I used the `declare_externproto.py` script to convert the R2022a `vehicles` project but the `city.wbt` is still showing some errors. For tiago++ project that is simpler it worked correctly.

5. The "Export" functionality from the Add Node dialog should not allow users to modify the installation directory by exporting PROTOs to it. Instead if the current project is located in the installation directory, before "exporting" the relocation project dialog should be popped up.

6. This is not an issues, but the remote URLs are sometimes inconsistent: in general the rawgit URL contains the tag (in case Webots is compiled from a branch) however, for example after exporting a PROTO from the add node dialog the URL are using the Webots version instead. It seems better to use the branch/commit tag instead of the version tag that may not exist.

ad-daniel:
>     1. In `projects/samples/howto/texture_change` the stone texture is still not found.
Yes, the distributed controller file (`.c`) has remote urls, but the binary likely didn't. It doesn't seem related to this PR, nothing being distributed should have `webots://` urls as no assets are included in the distribution, so it's normal it doesn't find it locally.
 
>     2. When inserting in the list of Ephemeral EXTERNPROTO a PROTO already declared in the file nothing happens.

Yes this was changed this morning, but not yet committed. Currently the dialog doesn't show nodes that are already declared. The reason is it doesn't make sense for it to be considered as ephemeral since by definition they are declared already and therefore not ephemeral.

>     3. When testing adding a node from a supervisor controller, if you add the new node as ephemeral, run the controller, save the world but cancel the saving when asked to choose for a new name (simulation has run):
>        -> the new node PROTO declaration is removed from the ephemeral list but not added to the world

I'm not sure to understand, does the controller need to spawn the node prior to attempt to saving? Can't reproduce it  

> I have many crashes on world reload when testing the tarball on Ubuntu 2022b. I will test then again in the development environment.

On the tar for ubuntu 20 didn't have any for a while, can you try to replace the binaries of the tar with binaries compile in debug mode to see if any assert triggers? Or a core dump? Can't do much if I can't reproduce it unfortunately
Also, how are you cleaning the cache? You should use the option in the preferences, if you delete manually the `.cache` folder while webots is open it will likely crash

stefaniapedrazzi:
>  3. When testing adding a node from a supervisor controller, if you add the new node as ephemeral, run the controller, save the world but cancel the saving when asked to choose for a new name (simulation has run):
       -> the new node PROTO declaration is removed from the ephemeral list but not added to the world

> I'm not sure to understand, does the controller need to spawn the node prior to attempt to saving? Can't reproduce it

No.
The problem is that the PROTO declaration is removed from the ephemeral list too early.
So if the user cancel the saving action, the PROTO is removed from the ephemeral list but not added to the world file.
The result is that to make it work again you have to readd the node in the ephemeral list manually.

> Yes, the distributed controller file (.c) has remote urls, but the binary likely didn't. It doesn't seem related to this PR, nothing being distributed should have webots:// urls as no assets are included in the distribution, so it's normal it doesn't find it locally.

Ok.
But it should be fixed.
The controller source is correctly update and doesn't containg any "webots://", but the binary is not recompiled so is still using the "webots://" URL.

> Yes this was changed this morning, but not yet committed. Currently the dialog doesn't show nodes that are already declared. The reason is it doesn't make sense for it to be considered as ephemeral since by definition they are in the world already and therefore not ephemeral.

Yes, I agree.
But the current behavior is confusing. I will wait to test the new version that you say it is not yet committed.

stefaniapedrazzi:
> On the tar for ubuntu 20 didn't have any for a while, can you try to replace the binaries of the tar with binaries compile in debug mode to see if any assert triggers? Or a core dump? Can't do much if I can't reproduce it unfortunately
Also, how are you cleaning the cache? You should use the option in the preferences, if you delete manually the .cache folder while webots is open it will likely crash

I will now test again just for the crashes and try to reproduce them. No action is required on your side, it was just to mention that in general Webots didn't seem to be very robust at the moment.

ygoumaz:
I am currently testing in the tarball distribution. I have found two issues:

- In supervisor.wbt (samples/devices), I try to import a Fridge PROTO during the simulation execution (instead of the Sphere). Even though a warning tells me to save the world after adding the Fridge PROTO to the ephemeral EXTERNPROTO list, the supervisor is able to spawn it right away without any reset, reload, saving or any other manipulation.
- If you try to convert this same imported Fridge to Base Node (right-click in scene tree), Webots crashes. If you do it while the simulation is paused, it will crash the next time you press play.

ygoumaz:
In addition to my last comment, a PROTO (imported from the Add Node window) converted to Base Node is deleted from the scene tree when resetting the simulation. I don't know if this is normal behavior or a bug related to this PR.

BenjaminDeleze:
> In addition to my last comment, a PROTO (imported from the Add Node window) converted to Base Node is deleted from the scene tree when resetting the simulation. I don't know if this is normal behavior or a bug related to this PR.

I think it is normal behavior

BenjaminDeleze:
Some worlds do not run because of missing extern protos delcaration:
- c3d_viewer.wbt
- highway_driving.wbt

BenjaminDeleze:
MacOS:

- 7) I am not sure it is related to this PR but I had several time the pop up `downloading asset` blocked for ever at 0% (even so the object/textures where visible in the world)
- 8) I got a bug when doing:
  - Create new project
  - Add a `Lincoln MKZ`
  - Convert Root to base node
  - Error VehicleWheel has not been declared as EXTERNPROTO
- 9) View proto: it is not possible to modify directly the file, but if I click on a world and type on the keyboard, it is possible to accept suggestions and thus to modify the file
- 10) When creating PROTO with the PROTO wizard it does not get the subproto (for example if I make a PROTO from a cabinet, it will declare only the cabinet proto and neither the `PaintedWood` nor the `CabinetHandle`

ygoumaz:
> * 7. I am not sure it is related to this PR but I had several time the pop up `downloading asset` blocked for ever at 0% (even so the object/textures where visible in the world)

I encountered this bug too on Linux, but could not reproduce it systematically.

ygoumaz:
> * In supervisor.wbt (samples/devices), I try to import a Fridge PROTO during the simulation execution (instead of the Sphere). Even though a warning tells me to save the world after adding the Fridge PROTO to the ephemeral EXTERNPROTO list, the supervisor is able to spawn it right away without any reset, reload, saving or any other manipulation.

I found an additional bug with the import of ephemeral EXTERNPROTOs. In the same scenario, I try to import the Fridge with the supervisor, but this time I remove the Fridge from the list of ephemeral EXTERNPROTOs. As expected, I get the warning in the console: 
```
ERROR: In order to import the PROTO 'Fridge', first it must be declared in the Ephemeral EXTERNPROTO list and the world must be saved.
```

But from there, the scene tree seems completely frozen. The text editor doesn't open correctly anymore and it isn't possible to double click on a field to expand it anymore.

ad-daniel:
To summarize things a bit, so far:

- [x] specify where EXTERNPROTO declarations are missing @ad-daniel 
- [x] copy-paste shows warning @ad-daniel 
  - [ ] cut-paste shows warning (see note below, likely related)
- [ ] benjamin's ephemeral shenanigans (note: not solvable without adding an extra flag, just ephemeral/not insufficient)
- [ ] texture_change controller wrong url @stefaniapedrazzi 
- [x] ephemeral EXTERNPROTO declaration of an already inserted PROTO @ad-daniel 
- [ ] cancelling the save when importing a node using supervisor
- [x] declare_externproto.py script improvements @ad-daniel 
- [x] export functionality should relocate if the destination is invalid @ad-daniel 
- [ ] inconsistent remote urls @ad-daniel 
- [ ] spawning of declared ephemeral (but world not saves after declaration) should be prohibited
- [ ] base-node conversion of imported node crashes webots
- [x] some worlds miss EXTERNPROTO declarations @ad-daniel 
- [ ] 0% download pop up
- [ ] conversion to ROOT base-node missing declarations
- [ ] "View PROTO" read-only restrictions can be bypassed
- [x] PROTO, created using the wizard, that has other PROTO as default value in the parameters aren't declared @ad-daniel 
- [ ] Text editor unresponsive after ephemeral EXTERNPROTO manipulations

I can't address everything myself, or at least not quickly, any help is appreciated. Flag any entry you intend to work on so we avoid overlaps

ygoumaz:
I have spotted an additional bug with the PROTO wizard, which may be more related to its implementation than to this PR.

The issue is the following. When creating a new PROTO with the wizard from the Cabinet PROTO (all fields selected), several fields are not completely generated. For example the rowsHeights field is given this way:

`field MFFloat rowsHeights [`, missing an additional `]`.

stefaniapedrazzi:
> spawning of declared ephemeral (but world not saved after declaration) should be prohibited 

Are you sure that we want to prevent it?
This seems a user-friendly functionality. To me it is normal that modifications that are not saved won't work after reload.

stefaniapedrazzi:
I even think that if it is not the case already, we will need to add a Supervisor function to add ephemeral EXTERNPROTO to the world in the next version of Webots .

ad-daniel:
> Ok. I agree that it is a matter of consistency and inserting from a Supervisor is the same as inserting from the scene tree. But why then a user can insert a new PROTO not defined as ephemeral from the add node dialog without first adding it to the list and then save and reload the file? It is exactly the same case.

The practical answer is because we can take care of downloading the PROTO when the user accepts the insertion, but you're right it's inconsistent.
Ideally we would do the same for node imports, when the user spawns something we could parse the string, extract the  PROTO nodes, download them, then do the actual node creation. The issue is: where do we download it from? Is it the locally available "Altino" I want to spawn or the official webots one? What if I want to spawn a PROTO that is only remotely accessible (but not one of ours)? Without a declaration being available somewhere, we can't make this choice.

ad-daniel:
Possible states for declarations: INSTANTIATED, EPHEMERAL, 

At load time:
- populate mExternProto with all the declarations
- flag all visible nodes as INSTANTIATED
- flag everything else as EPHEMERAL

Insert from add-node:
- add declaration with state INSTANTIATED

Insert from supervisor:
- if no declarations are available -> error
- if an EPHEMERAL declaration is available -> change to INSTANTIATED

Paste:
- no changes to states nor to the declarations

Cut/Delete/Supervisor remove:
- no changes to states nor to the declarations

Ephemeral Panel:
- EPHEMERALs can be added or removed

On World save:
- Go through the list of INSTANTIATED nodes, filter out those for which an instance no longer exists, then write to file the rest
- write to file all EPHEMERALs

In short, only a world save should remove declarations (well, and manually using the ephemeral panel)

As for the inconsistency between spawned nodes and added nodes being treated differently, I have no good solution at this moment.

stefaniapedrazzi:
In general seems fine to me.
But there will be some possible issue with supervisor controller:
1. world declares an ephemeral node A that will be added by the controller
2. supervisor controller inserts node A -> node is no longer EPHEMERAL but INSTANTIATED
3. world is reloaded/reset and saved: node A is no longer instantiated in the list and will be removed
4. After reloading and running again the world, the supervisor controller won't work because node A is no longer declared as EXTERNPROTO.

Filtering out nodes is tricky.
The only way it will work will be to be able to specify in the EXTERNPROTO declaration if it was manually added and lock it so that the cleanup method won't delete the manually added Ephemeral EXTERNPROTO but only the ones added through the Add node dialog.

ygoumaz:
On Windows:

- Following test about _edit remote PROTO_: 
  1. Open complex sample world (village_realistic.wbt in my case)
  2. Right-click a proto (BusSimple.proto in my case), two options should be available: `view PROTO source` and `edit PROTO source`
  3. Click `edit PROTO source`, accept relocation, and SAVE the world.
    3.1 Open the world file with text editor, ensure the relocated PROTO is declared locally.

  The EXTERNPROTO URL is indeed modified, but is given as absolute path. I believe it should be relative to the local project.

- [All systems] Importing a remote PROTO (from a sample project) to a current local project triggers warning because the original controller/robot_window/etc. can't be found. It may be more user-friendly to update these fields to <generic>.

BenjaminDeleze:
This one is perhaps a bit far-fetched but I can make Webots crash by doing:

- Modifiy the url of a Proto (let's say Bed.proto) in the proto-list.xml to something that does not exist
- Open webots, click the add button
- Add a Bed.

If I just click on the proto, it says that it cannot download the image from the server, but if I actually add it, it crashes

BenjaminDeleze:
Is it normal that hidden protos are listed in the `Insert New` Ephemeral Externproto list?

ad-daniel:
> Is it normal that hidden protos are listed in the `Insert New` Ephemeral Externproto list?

Since they don't show up in the add-node dialog, I assume it shouldn't be allowed to be imported either. I'll fix that in this PR as well: https://github.com/cyberbotics/webots/pull/4899

stefaniapedrazzi:
I'm getting an assertion failure in this branch:
1. open `connector_static_autolock.wbt` (tests/physics)
2. Select the field `DEF SOLID Solid > Connect > boundingObject`
3. click the "Add Node" button
-> Assertion Failure

<details>
<summary>Stack trace</summary>
<pre>
```(gdb) bt
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=140737252543808) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=140737252543808) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=140737252543808, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007ffff4cc1476 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007ffff4ca77f3 in __GI_abort () at ./stdlib/abort.c:79
#5  0x00007ffff4ca771b in __assert_fail_base
    (fmt=0x7ffff4e5c150 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=0x555556060bf0 "mUrl.endsWith(\".proto\", Qt::CaseInsensitive)", file=0x555556060ade "vrml/WbProtoModel.cpp", line=96, function=<optimized out>) at ./assert/assert.c:92
#6  0x00007ffff4cb8e96 in __GI___assert_fail
    (assertion=0x555556060bf0 "mUrl.endsWith(\".proto\", Qt::CaseInsensitive)", file=0x555556060ade "vrml/WbProtoModel.cpp", line=96, function=0x555556060a78 "WbProtoModel::WbProtoModel(WbTokenizer*, const QString&, const QString&, const QString&, QStringList)") at ./assert/assert.c:101
#7  0x0000555555cf0e85 in WbProtoModel::WbProtoModel(WbTokenizer*, QString const&, QString const&, QString const&, QList<QString>)
    (this=0x55555d6753c0, tokenizer=0x7fffffffb720, worldPath=..., url=..., prefix=..., baseTypeList=...) at vrml/WbProtoModel.cpp:96
#8  0x0000555555ce3cf9 in WbProtoManager::generateInfoFromProtoFile(QString const&) (this=0x555556909e30, protoFileName=...) at vrml/WbProtoManager.cpp:681
#9  0x0000555555ce284f in WbProtoManager::generateProtoInfoMap(int, bool) (this=0x555556909e30, category=10002, regenerate=true) at vrml/WbProtoManager.cpp:540
#10 0x0000555555a6d28a in WbAddNodeDialog::addProtosFromProtoList(QTreeWidgetItem*, int, QRegularExpression const&, bool)
    (this=0x7fffffffca50, parentItem=0x55555d9b9aa0, type=10002, regexp=..., regenerate=true) at scene_tree/WbAddNodeDialog.cpp:587
#11 0x0000555555a6ca33 in WbAddNodeDialog::buildTree() (this=0x7fffffffca50) at scene_tree/WbAddNodeDialog.cpp:542
```
</pre>
</details>

```
webots-bin: vrml/WbProtoModel.cpp:96: WbProtoModel::WbProtoModel(WbTokenizer*, const QString&, const QString&, const QString&, QStringList): Assertion `mUrl.endsWith(".proto", Qt::CaseInsensitive)' failed.
```
mUrl value is `/home/stefi/.cache/Cyberbotics/Webots/assets/066f4f4590ff1262091bd2c78d5ef682fad673b4`.

ad-daniel:
The obvious fix is to get the url from the cache path with a reverse lookup, but in practice that shouldn't happen in the first place, so the error is likely in the way the WbProtoModel constructor is reached in this specific case. I'll look into it.

ygoumaz:
> * If you try to convert this same imported Fridge to Base Node (right-click in scene tree), Webots crashes. If you do it while the simulation is paused, it will crash the next time you press play.

I still can reproduce the crash when converting an imported node to base node during execution (tarball distribution and sources).

stefaniapedrazzi:
I have a crash when opening the "Add Node" dialog when following the manual test `tests/manual_tests/worlds/insertion_in_nested_parameter.wbt`:

1. open tests/manual_tests/worlds/insertion_in_nested_parameter.wbt`
2. select RobotWithSlot > extensions > SolidSlot > slot
3. open the Add Node dialog

```
0x00007ffff7bc3c02 in QString::arg(QString const&, int, QChar) const () from /home/stefi/develop/webots/lib/webots/libQt6Core.so.6
(gdb) bt
#0  0x00007ffff7bc3c02 in QString::arg(QString const&, int, QChar) const ()
    at /home/stefi/develop/webots/lib/webots/libQt6Core.so.6
#1  0x0000555555a6eda6 in WbAddNodeDialog::addProtosFromProtoList(QTreeWidgetItem*, int, QRegularExpression const&, bool)
    (this=0x7fffffffca30, parentItem=0x55555d876750, type=10002, regexp=..., regenerate=true) at scene_tree/WbAddNodeDialog.cpp:659
#2  0x0000555555a6da33 in WbAddNodeDialog::buildTree() (this=0x7fffffffca30)
    at scene_tree/WbAddNodeDialog.cpp:542
```
```
(gdb) frame 1
#1  0x0000555555a6eda6 in WbAddNodeDialog::addProtosFromProtoList (this=0x7fffffffca30, parentItem=0x55555d876750, type=10002, regexp=..., 
    regenerate=true) at scene_tree/WbAddNodeDialog.cpp:659
659	      new QTreeWidgetItem(QStringList() << QString("%1 (%2)").arg(protoName).arg(info->baseType()) << info->url());
(gdb) print protoName
$1 = {d = {d = 0x55555d871e50, ptr = 0x55555d871e60 u"4d3c59b17edd7555e46e12088b41cd2113c5fc5c", size = 40}, static _empty = 0 u'\000'}
(gdb) print info
$2 = (const WbProtoInfo *) 0x0
```

stefaniapedrazzi:
It is still possible to cut and paste on the editor for read-only PROTO files.
I will take care of fixing it.

ad-daniel:
The crash on the manual test should be addressed.

ad-daniel:
> When opening R2022a worlds (like city.wbt) I get no warnings.

I do get them, did you per chance use the script already to convert them to R2022b? I believe those were the worlds you used previously to test it.

> Currently no error is displayed if a PROTO doesn't declare an internal PROTO as EXTERNPROTO, but the declaration is added in the world.

I'll look into it now.

stefaniapedrazzi:
> I do get them, did you per chance use the script already to convert them to R2022b? I believe those were the worlds you used previously to test it.

No, I relocated the R2022a project (worlds and controllers folders but not the ) from a clean Webots R2022a installation before trying to opening it with R2022b and modify it.
I also checked the R2022a file and the world file doesn't contain any EXTERNPROTO definition.
But the default R2022b PROTO files are automatically linked (without throwing any error in the console).

stefaniapedrazzi:
I'm now getting an assertion failure in this branch:
```
webots-bin: vrml/WbProtoManager.hpp:78: WbProtoInfo::WbProtoInfo(const QString&, const QString&, const QString&, const QString&, const QString&, const QString&, const QString&, const QStringList&, const QStringList&, bool): Assertion `mParameters.size() == mParameterNames.size()' failed.
```

For `SickLdMrs.proto` the `type`and `angularResolution` are not listed in `mParameterNames`.
These are SFString parameters with field value restrictions.

stefaniapedrazzi:
It seems that PROTOs are loaded too early when instantiating the WbMainWindow and WbSceneTree and the WEBOTS_SAFE_MODE is broken.
I'm getting the error with the PROTO used in the `city.wbt` world, but I cannot load any other world using the WEBOTS_SAFE_MODE` or running `./webots projects/samples/devices/worlds/accelerometer.wbt` because Webots always tries to load the `SickLdMrs.proto` that it is not used in the `empty.wbt` or `accelerometer.wbt` world.

ad-daniel:
Yes, in development mode the assert is quite brutal, preventing you from opening webots at all. But unless the issue that triggers the assert is addressed immediately, it will show up later in harder to find situations with 100% certainty, often resulting in crashes that are hard to pin-point (just clicking add-node, or while creating a PROTO with the wizard, etc). So it's brutal, but at the same time it's the best moment to fix it is immediately.

I agree however that a cleanse (or even a clean) should remove it to ensure it gets regenerated. However it means that we need to remember to clean when switching to a branch that contains changes to PROTOs, to the script that generates the xml or the webots parser. If we don't clean, it might not get regenerated, and this situation may happen again

stefaniapedrazzi:
> The crash on the manual test should be addressed.

I confirm the two crashes are now fixed.

BenjaminDeleze:
> * ~0% download pop up~ unable to be reproduced consistently

I made some experiments on the Mac with both low and high internet speed,... and was not able to reproduce it a single time

BenjaminDeleze:
> * The `highway_overtake.wbt` world is still missing some EXTERNPROTO declaration when played. I will handle it.

After investigation, it is a problem of a PROTO being both instanciated in the world and need as importable. It will be fix by the refactor

ad-daniel:
Since this PR is becoming unmanageable and that all changes done to it went through secondary PRs, I think it's best if we merge this into master, launch a new release candidate and do the extensive tests on that one rather than this PR (otherwise we end up doing the tests twice). If issues persist, we should be able to address them today. We can of course delay the merge if you @stefaniapedrazzi or @omichel want to do another quick pass at the syntax/sources

stefaniapedrazzi:
I think that we should review the PR before merging.
Otherwise after merging it to master this will never be done.

src/webots/nodes/utils/WbUrl.cpp
```diff
@@ -302,3 +219,92 @@ const QString WbUrl::computeLocalAssetUrl(const WbNode *node, QString url) {
...
+    "https://raw.githubusercontent.com/cyberbotics/webots/" + (commit.isEmpty() ? version.toString() : commit) + "/";
+  return url;
+}
+
+QString WbUrl::combinePaths(const QString &rawUrl, const QString &rawParentUrl) {
+  // use cross-platform forward slashes
+  QString url = rawUrl;
+  url = url.replace("\\", "/");
+  QString parentUrl = rawParentUrl;
+  parentUrl = parentUrl.replace("\\", "/");
```
omichel, line 257:
```suggestion
  parentUrl.replace("\\", "/");
```

src/webots/nodes/utils/WbUrl.cpp
```diff
@@ -89,128 +90,44 @@ QString WbUrl::computePath(const WbNode *node, const QString &field, const WbMFS
...
+  QStringList searchPaths;
+  if (!relativeTo.isEmpty())
+    searchPaths << relativeTo;
+  searchPaths << WbProject::current()->protosPath();
+  searchPaths << WbProject::current()->worldsPath();
+
+  foreach (const QString &path, searchPaths) {
+    if (QDir::isRelativePath(url)) {
+      QString protosPath = QDir(path).absoluteFilePath(url);
+      protosPath = QDir::cleanPath(protosPath);
```
omichel, line 128:
```suggestion
      const QString protosPath = QDir::cleanPath(QDir(path).absoluteFilePath(url));
```

src/webots/nodes/utils/WbUrl.cpp
```diff
@@ -302,3 +219,92 @@ const QString WbUrl::computeLocalAssetUrl(const WbNode *node, QString url) {
     // when streaming a release (or nightly), "webots://" urls must be inferred
     return WbUrl::computePath(node, "url", url, false);
 }
+
+const QString WbUrl::computePrefix(const QString &rawUrl) {
+  QString url = rawUrl;
+  if (url.startsWith(WbNetwork::instance()->cacheDirectory()))
+    url = WbNetwork::instance()->getUrlFromEphemeralCache(url);
```
omichel, line 226:
```suggestion
  const QString url = rawUrl.startsWith(WbNetwork::instance()->cacheDirectory()) ? WbNetwork::instance()->getUrlFromEphemeralCache(rawUrl) : rawUrl;
```

src/webots/vrml/WbNode.cpp
```diff
@@ -1099,8 +1103,7 @@ QStringList WbNode::listTextureFiles() const {
       QString protoPath;
       if (proto) {
         WbProtoModel *protoModel = proto->proto();
-        QFileInfo fileInfo(protoModel->fileName());
-        protoPath = fileInfo.path() + "/";
+        protoPath = protoModel->path();
       }
```
omichel, line 1107:
```suggestion
```

src/webots/vrml/WbNode.cpp
```diff
@@ -1099,8 +1103,7 @@ QStringList WbNode::listTextureFiles() const {
       QString protoPath;
       if (proto) {
         WbProtoModel *protoModel = proto->proto();
```
omichel, line 1105:
```suggestion
      if (proto)
        protoPath = proto->proto()->path();
```

src/webots/vrml/WbNode.cpp
```diff
@@ -1215,6 +1218,66 @@ void WbNode::exportNodeContents(WbWriter &writer) const {
...
+  if (!isProtoInstance())
+    return;
+
+  addExternProtoFromFile(mProto);
+}
+
+void WbNode::addExternProtoFromFile(const WbProtoModel *proto) const {
+  QString path = proto->url();
+  if (WbUrl::isWeb(path) && WbNetwork::instance()->isCached(path))
+    path = WbNetwork::instance()->get(path);
```
omichel, line 1231:
```suggestion
  const QString path = (WbUrl::isWeb(proto->url()) && WbNetwork::instance()->isCached(proto->url())) ? WbNetwork::instance()->get(proto->url() : proto->url();
```

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -105,85 +107,130 @@ void WbProtoManager::readModel(WbTokenizer *tokenizer, const QString &worldPath)
...
+  if (url.isEmpty())
+    return QString();
+
+  QFile file(WbUrl::isWeb(url) ? WbNetwork::instance()->get(url) : url);
+  if (!file.open(QIODevice::ReadOnly)) {
+    WbLog::error(tr("Could not search for EXTERNPROTO declarations in '%1' because the file is not readable.").arg(url));
+    return QString();
+  }
+  QString identifier = modelName;
+  identifier = identifier.replace("+", "\\+").replace("-", "\\-").replace("_", "\\_");
```
omichel, line 220:
```suggestion
  identifier.replace("+", "\\+").replace("-", "\\-").replace("_", "\\_");
```

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -198,99 +245,135 @@ QString WbProtoManager::findModelPath(const QString &modelName) const {
...
+      protoNodeList.insert(proto, url);
+
+    if (!WbUrl::isWeb(url)) {
+      // open the PROTO file and extract all PROTO nodes it refrerences by brute force (comparing against the known nodes)
+      QFile file(url);
+      if (file.open(QIODevice::ReadOnly)) {
+        const QByteArray &contents = file.readAll();
+        foreach (const QString &item, knownProto) {
+          QString identifier = item;
+          identifier = identifier.replace("+", "\\+").replace("-", "\\-").replace("_", "\\_");
```
omichel, line 312:
```suggestion
          identifier.replace("+", "\\+").replace("-", "\\-").replace("_", "\\_");
```

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -661,20 +760,16 @@ void WbProtoManager::exportProto(const QString &path, int category) {
...
     input.close();
 
     // in webots development environment use 'webots://', in a distribution use the version
-    if (WbApplicationInfo::branch().isEmpty()) {
-      const WbVersion &version = WbApplicationInfo::version();
-      const QString &reference = version.commit().isEmpty() ? version.toString() : version.commit();
-      contents = contents.replace("webots://", "https://raw.githubusercontent.com/cyberbotics/webots/" + reference + "/");
-    }
+    if (WbApplicationInfo::branch().isEmpty())
+      contents = contents.replace("webots://", WbUrl::remoteWebotsAssetPrefix());
```
omichel, line 772:
```suggestion
      contents.replace("webots://", WbUrl::remoteWebotsAssetPrefix());
```

src/webots/vrml/WbProtoModel.cpp
```diff
@@ -449,17 +448,18 @@ WbFieldModel *WbProtoModel::findFieldModel(const QString &fieldName) const {
...
-    QString path = mPath;
-    if (mPath.startsWith("https://"))
-      path = path.replace(QRegularExpression("https://raw.githubusercontent.com/cyberbotics/webots/[a-zA-Z0-9\\_\\-\\+]+/"),
-                          WbStandardPaths::webotsHomePath());
+  QString protoPath = path();
+
+  if (!protoPath.isEmpty()) {
+    if (WbUrl::isWeb(protoPath))
+      protoPath =
+        protoPath.replace(QRegularExpression(WbUrl::remoteWebotsAssetRegex(false)), WbStandardPaths::webotsHomePath());
```
omichel, line 456:
```suggestion
      protoPath.replace(QRegularExpression(WbUrl::remoteWebotsAssetRegex(false)), WbStandardPaths::webotsHomePath());
```

omichel (changes requested):

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -753,3 +872,51 @@ void WbProtoManager::cleanup() {
   mSessionProto.clear();
   mExternProto.clear();
 }
+
+QString WbProtoManager::injectDeclarationByBackwardsCompatibility(const QString &modelName) {
+  // check if it's the current project
+  QStringList projectProto = listProtoInCategory(PROTO_PROJECT);
```
stefaniapedrazzi, line 882:
```suggestion
  const QStringList projectProto = listProtoInCategory(PROTO_PROJECT);
```

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -753,3 +872,51 @@ void WbProtoManager::cleanup() {
...
+  foreach (const QString &proto, projectProto) {
+    if (proto.contains(modelName + ".proto", Qt::CaseInsensitive))
+      return QFileInfo(proto).absoluteFilePath();
+  }
+  // check if it's in the EXTRA projects
+  QStringList extraProto = listProtoInCategory(PROTO_EXTRA);
```
stefaniapedrazzi, line 888:
```suggestion
  const QStringList extraProto = listProtoInCategory(PROTO_EXTRA);
```

stefaniapedrazzi (changes requested):
here are some comments to the source code

src/webots/scene_tree/WbAddNodeDialog.cpp
```diff
@@ -14,15 +14,15 @@
 
 #include "WbAddNodeDialog.hpp"
 
-#include <WbDownloader.hpp>
-#include <WbNetwork.hpp>
+#include "../editor/WbProjectRelocationDialog.hpp"
```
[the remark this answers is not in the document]
stefaniapedrazzi: The include rules are defined in the Webots Makefile, you should modify it there, checking that it adding the `editor` module to the `scene_tree` module doesn't cause any circular dependency.
https://github.com/cyberbotics/webots/blob/master/src/webots/Makefile#L46

src/webots/vrml/WbProtoTreeItem.cpp
```diff
@@ -52,17 +56,20 @@ void WbProtoTreeItem::parseItem() {
...
-      const QString subProtoUrl = WbUrl::generateExternProtoPath(subProto, mUrl);
+      const bool isImportable = !match.captured(1).isEmpty();
+      const QString subProto = match.captured(2);
+      const QString subProtoUrl = WbUrl::combinePaths(subProto, mUrl);
+      if (subProtoUrl.isEmpty())
+        continue;
 
-      if (!subProtoUrl.endsWith(".proto")) {
+      if (!subProtoUrl.endsWith(".proto", Qt::CaseInsensitive)) {
         mError << QString(tr("Malformed EXTERNPROTO url. The url should end with '.proto'."));
```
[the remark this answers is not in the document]
stefaniapedrazzi: Same wrong case for `URL` is also used at line 80

src/webots/core/WbApplicationInfo.cpp
```diff
@@ -53,10 +54,23 @@ const QString &WbApplicationInfo::repo() {
...
   return repoName;
 }
 
+const QString &WbApplicationInfo::commit() {
+  static QString commit;
+  static bool firstCall = true;
+  if (firstCall) {
+    // included only in locally created distributions and nightlies, official distributions don't
+    const QString commitString("resources/commit.txt");
+    commit = getInfoFromFile(commitString);
```
stefaniapedrazzi, line 68:
```suggestion
    commit = getInfoFromFile("resources/commit.txt");
```

src/webots/engine/WbSimulationWorld.cpp
```diff
@@ -28,6 +28,7 @@
 #include "WbPerformanceLog.hpp"
 #include "WbPhysicsPlugin.hpp"
 #include "WbPreferences.hpp"
+#include "WbProtoManager.hpp"
```
stefaniapedrazzi, line 31: This is not needed
```suggestion
```

src/webots/gui/WbMainWindow.cpp
```diff
@@ -1990,6 +1992,17 @@ void WbMainWindow::newPhysicsPlugin() {
 }
 
 void WbMainWindow::newProto() {
+  // should not generate the wizard until all the dependencies are available, double re-entry is necessary
+  const QAction *const caller = qobject_cast<QAction *>(sender());
+  if (caller) {  // if the function is reached by the GUI menu
```
stefaniapedrazzi, line 1997:
```suggestion
  if (qobject_cast<QAction *>(sender())) {  // if the function is reached by the GUI menu
```

stefaniapedrazzi (commented):
I missed a couple of files in the previuos review

ad-daniel:
> The test suite is failing. I executed `make cleanse` before recompiling. I'm still checking and I will list all the failures:

Yes I've noticed as well and I've fixed it already here: https://github.com/cyberbotics/webots/pull/4860/commits/d94e927b522a69a297e123c6cf6b4c83f9e4a41e

Basically in this test we purposefully attempt to load an unknown node, which triggered one of the warning messages that is supposed to tell you how you need to declare it. That's however possible only if the node is among the known ones.

ad-daniel:
I believe I've addressed all the raised points, let me know if I missed something

src/webots/Makefile
```diff
@@ -43,7 +43,7 @@ WB_CONTROL_INCLUDE       = $(QT_CORE_INCLUDE) $(QT_NETWORK_INCLUDE) -Icore -Ieng
 # dependency of NODES on QT_GUI is necessary in WbImageTexture to retrieve the size of a texture before loading it in Wren
 WB_NODES_INCLUDE         = $(QT_CORE_INCLUDE) $(QT_NETWORK_INCLUDE) $(QT_GUI_INCLUDE) $(ODE_INCLUDE) $(CONTROLLER_INCLUDE) $(OIS_INCLUDE) $(FREETYPE_INCLUDE) $(WREN_INCLUDE) $(STB_INCLUDE) $(ASSIMP_INCLUDE) -Iapp -Icore -isystem external/siphash -Imaths -Inodes -Inodes/utils -Iode -Iplugins -Isound -Iutil -Ivrml
 WB_APP_INCLUDE           = $(QT_CORE_INCLUDE) $(QT_NETWORK_INCLUDE) $(ODE_INCLUDE) $(WREN_INCLUDE) -Icontrol -Icore -Ieditor -Iengine -Imaths -Inodes -Inodes/utils -Iplugins -Iscene_tree -Ivrml
-WB_SCENE_TREE_INCLUDE    = $(QT_CORE_INCLUDE) $(QT_GUI_INCLUDE) $(QT_WIDGETS_INCLUDE) $(ASSIMP_INCLUDE) -isystem external/siphash -Iapp -Icore -Iwidgets -Imaths -Inodes -Inodes/utils -Iode -Iuser_commands -Ivrml
+WB_SCENE_TREE_INCLUDE    = $(QT_CORE_INCLUDE) $(QT_GUI_INCLUDE) $(QT_WIDGETS_INCLUDE) $(ASSIMP_INCLUDE) $(WB_EDITOR_INCLUDE) -isystem external/siphash -Iapp -Icore -Iwidgets -Imaths -Inodes -Inodes/utils -Iode -Iuser_commands -Ivrml -Ieditor
```
stefaniapedrazzi, line 46: It should not be needed to add `WB_EDITOR_INCLUDE` (you are duplicating many includes this way).
Modules are listed by alphabetical order.
```suggestion
WB_SCENE_TREE_INCLUDE    = $(QT_CORE_INCLUDE) $(QT_GUI_INCLUDE) $(QT_WIDGETS_INCLUDE) $(ASSIMP_INCLUDE) -isystem external/siphash -Iapp -Icore -Ieditor -Imaths -Inodes -Inodes/utils -Iode -Iuser_commands -Ivrml -Iwidgets
```

src/webots/editor/WbProjectRelocationDialog.cpp
```diff
@@ -402,24 +402,24 @@ bool WbProjectRelocationDialog::validateLocation(QWidget *parent, QString &filen
...
 
-  // change filename parameter: get relative filename with respect to previous project path
+  // get relative filename with respect to previous project path
   QString absolutePath;
   if (mExternalProtoProjectPath.isEmpty())
     absolutePath = current->path();
   else
     absolutePath = mExternalProtoProjectPath;
-  filename = QDir(absolutePath).relativeFilePath(filename) + (filename.endsWith("/") ? "/" : "");
+  QString newFilename = QDir(absolutePath).relativeFilePath(filename) + (filename.endsWith("/") ? "/" : "");
```
stefaniapedrazzi, line 411: Previously this function was adjusting the filename and returning it to the caller.
Now this behavior changed and the caller will have a wrong and outdated filename.
Is this really wanted?

src/webots/nodes/utils/WbUrl.cpp
```diff
@@ -295,10 +211,103 @@ bool WbUrl::isLocalUrl(const QString &url) {
...
+
+const QString &WbUrl::remoteWebotsAssetPrefix() {
+  static QString url;
+  static bool firstCall = true;
+  if (firstCall) {
+    // if it's an official release, use the tag (for example R2022b), if it's a nightly or local distribution use the commit
+    url = "https://raw.githubusercontent.com/cyberbotics/webots/" +
+          (WbApplicationInfo::commit().isEmpty() ? WbApplicationInfo::version().toString() : WbApplicationInfo::commit()) + "/";
+    firstCall = false;
+  }
```
stefaniapedrazzi, line 250: This could also be simplified
```suggestion
  static QString url;
  if (url.isEmpty())
    // if it's an official release, use the tag (for example R2022b), if it's a nightly or local distribution use the commit
    url = "https://raw.githubusercontent.com/cyberbotics/webots/" +
          (WbApplicationInfo::commit().isEmpty() ? WbApplicationInfo::version().toString() : WbApplicationInfo::commit()) + "/";
```

src/webots/scene_tree/WbAddNodeDialog.cpp
```diff
@@ -759,6 +752,10 @@ int WbAddNodeDialog::selectionType() {
 }
 
 void WbAddNodeDialog::exportProto() {
+  QString destination = WbProject::current()->protosPath();
+  if (!WbProjectRelocationDialog::validateLocation(this, destination))
```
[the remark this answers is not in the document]
stefaniapedrazzi: Is there a reason why we need the `destination` variable that is used only once?

src/webots/vrml/WbNode.cpp
```diff
@@ -1215,6 +1216,64 @@ void WbNode::exportNodeContents(WbWriter &writer) const {
...
+    parsingWarn(tr("File '%1' is not readable.").arg(path));
+    return;
+  }
+
+  QString ancestorName;
+  if (proto->isDerived())
+    ancestorName = proto->ancestorProtoName();
+
+  // check if the root file references external PROTO
+  QRegularExpression re("^\\s*EXTERNPROTO\\s+\"(.*\\.proto)\"", QRegularExpression::MultilineOption);
```
stefaniapedrazzi, line 1242:
```suggestion
  const QRegularExpression re("^\\s*EXTERNPROTO\\s+\"(.*\\.proto)\"", QRegularExpression::MultilineOption);
```

src/webots/vrml/WbProtoTreeItem.cpp
```diff
@@ -14,20 +14,24 @@
 
 #include "WbProtoTreeItem.hpp"
 
+#include "WbApplicationInfo.hpp"
 #include "WbDownloader.hpp"
 #include "WbNetwork.hpp"
+#include "WbProject.hpp"
 #include "WbStandardPaths.hpp"
 #include "WbUrl.hpp"
+#include "WbVersion.hpp"
```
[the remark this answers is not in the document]
stefaniapedrazzi: This comment doesn't seem to have been addressed.

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -105,85 +107,134 @@ void WbProtoManager::readModel(WbTokenizer *tokenizer, const QString &worldPath)
   model->ref();
 }
 
-WbProtoModel *WbProtoManager::findModel(const QString &modelName, const QString &worldPath, QStringList baseTypeList) {
+WbProtoModel *WbProtoManager::findModel(const QString &modelName, const QString &worldPath, const QString &parentFilePath,
+                                        QStringList baseTypeList) {
```
stefaniapedrazzi, line 111: We should avoid passing argument by value
```suggestion
                                        const QStringList &baseTypeList) {
```

src/webots/vrml/WbProtoManager.hpp
```diff
@@ -135,20 +141,24 @@ class WbProtoManager : public QObject {
   // searches for proto model according to:
   // 1. First in the session list (i.e., PROTO discovered by navigating the PROTO tree stemming from a world file
   // 2. If the first fails, it searches among the known official Webots PROTO (proto-list.xml)
-  WbProtoModel *findModel(const QString &modelName, const QString &worldPath, QStringList baseTypeList = QStringList());
+  WbProtoModel *findModel(const QString &modelName, const QString &worldPath, const QString &parentFilePath,
+                          QStringList baseTypeList = QStringList());
 
-  WbProtoModel *readModel(const QString &fileName, const QString &worldPath, const QString &externUrl = QString(),
+  WbProtoModel *readModel(const QString &url, const QString &worldPath, const QString &prefix = QString(),
                           QStringList baseTypeList = QStringList()) const;
```
stefaniapedrazzi, line 148: This is an issue already merged in the master branch but we should avoid passing objects by value
```suggestion
                          const QStringList &baseTypeList = QStringList()) const;
```

src/webots/vrml/WbProtoManager.hpp
```diff
@@ -135,20 +141,24 @@ class WbProtoManager : public QObject {
   // searches for proto model according to:
   // 1. First in the session list (i.e., PROTO discovered by navigating the PROTO tree stemming from a world file
   // 2. If the first fails, it searches among the known official Webots PROTO (proto-list.xml)
-  WbProtoModel *findModel(const QString &modelName, const QString &worldPath, QStringList baseTypeList = QStringList());
+  WbProtoModel *findModel(const QString &modelName, const QString &worldPath, const QString &parentFilePath,
+                          QStringList baseTypeList = QStringList());
```
stefaniapedrazzi, line 145: we should avoid passing objects by value
```suggestion
                          const QStringList &baseTypeList = QStringList());
```

stefaniapedrazzi (changes requested):
A couple of comments seems to not have been addressed.
And I found a couple more.

Just for note, in code comments it doesn't matter the case we use for "proto", and "url" but we should use the correct one for messages and documentation.

src/webots/vrml/WbProtoManager.hpp
ad-daniel: I don't believe we can since the value is changed and propagated further down, unless we remove the default value from the declaration

src/webots/editor/WbProjectRelocationDialog.cpp
ad-daniel: `validateLocation` was changed to address this: https://github.com/cyberbotics/webots/pull/4860#discussion_r928663714
but as you pointed out this isn't a good solution either as it changes the behavior.
Then I don't think we can avoid having the intermediary `destination` variable, even if unused

stefaniapedrazzi: Yes, if it is needed then it is fine to have the `destination` variable.

src/webots/vrml/WbProtoManager.hpp
ad-daniel: It can't be const, and to pass by reference we do indeed need to remove the default value, which adds a whole lot of hanging variables, I'm 50/50 on whether passing by reference really brings us much in terms for performance/cleanliness in this case: https://github.com/cyberbotics/webots/pull/4860/commits/9d58bca3ac6aa85ea56878fd31e3ce709d228991
What do you think?

src/webots/scene_tree/WbAddNodeDialog.cpp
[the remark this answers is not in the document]
ad-daniel: As mentioned below, due to `validateLocation` possibly modifying the destination, in this case it's necessary

src/webots/vrml/WbProtoManager.hpp
stefaniapedrazzi: The default value can be set even if you pass the argument by reference.
What doesn't work?

ad-daniel: I'm getting: 

```
cannot bind non-const lvalue reference of type ‘QStringList&’ {aka ‘QList<QString>&’} to an rvalue of type ‘QStringList’ {aka ‘QList<QString>’}
  145 |                           QStringList &baseTypeList = QStringList());
```

stefaniapedrazzi: You should also add the `const

ad-daniel: Ah nevermind, I've read wrong, I thought you suggested WbProtoModel should receive it by reference, but that one can't because it appends items to the list. `readmodel` and `findmodel` indeed can

stefaniapedrazzi (approved):
Thank you!

I will make more functional tests with the release candidate.

omichel (approved):
Thank you.

src/webots/app/WbApplication.cpp
```diff
@@ -209,6 +195,10 @@ void WbApplication::loadWorld(QString worldName, bool reloading, bool isLoadingA
 
   worldName = QDir::cleanPath(worldName);
 
+  WbTokenizer tokenizer;
+  tokenizer.tokenize(worldName);
+  WbParser parser(&tokenizer);
```
stefaniapedrazzi, line 200: What is the purpose of calling the tokenizer and parser here?

Note that `tokenizer.tokenize(worldName);` is called again at line 224.
And `parser` is used only later at line 241.

ad-daniel: In the previous backwards compatibility mechanism the tokenizer was used to retrieve the list of undeclared nodes which required moving it to the beginning of the function (since we enter it twice). As this is no longer the case, it can be returned where it was if it's causing issues

stefaniapedrazzi: Yes, this is wrong. I will revert it.

PR #4862, opened 2022-07-07
Add changelog for 2023

BenjaminDeleze:
ditto

### DIFF
--- a/docs/reference/changelog-r2023.md
+++ b/docs/reference/changelog-r2023.md
@@ -1,7 +1,7 @@
 # Webots R2023 Change Log
 
 ## Webots R2023a
-Released on XX, XXth, 2023.
+Released on December, 12th, 2022.
   - New Robots
   - New Objects
   - New Features
--- a/docs/reference/changelog-r2023.md
+++ b/docs/reference/changelog-r2023.md
@@ -1,0 +1,10 @@
+# Webots R2023 Change Log
+
+## Webots R2023a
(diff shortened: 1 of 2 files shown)


docs/reference/changelog-r2023.md
```diff
@@ -0,0 +1,10 @@
+# Webots R2023 Change Log
+
+## Webots R2023a
+Released on XX, XXth, 2023.
+  - New Robots
+  - New Objects
+  - New Features
+  - Enhancements
```
omichel, line 8: I believe this order is slightly better (from the most important to the less important):
```suggestion
  - New Features
  - Enhancements
  - New Robots
  - New Objects
```

docs/reference/changelog-r2023.md
```diff
@@ -0,0 +1,10 @@
+# Webots R2023 Change Log
+
+## Webots R2023a
+Released on XX, XXth, 2023.
```
omichel, line 4: Let's target a release date (we may change it if needed):
```suggestion
Released on December, 12th, 2022.
```

omichel (changes requested):

docs/reference/changelog-r2023.md
BenjaminDeleze: I kept the order that we have in the current changelog, so we may also want to change it before the release

omichel (approved):

PR #4863, opened 2022-07-07
Fix path of edited PROTO

ad-daniel:
**Description**

Ditto. It was adding 2 forward slashes


omichel:
`QDir::absolutePath()` and `QDir::relativeFilePath()` never return a string with a final "/" while `QDir::absoluteFilePath()` returns a string with a final "/" if its argument has a final "/".

ygoumaz (dismissed):
I have tested the relocation on multiple different projects and it seems to work fine with the assertion suggested by @omichel in [his comment](https://github.com/cyberbotics/webots/pull/4863/files#r915518752).

PR #4871, opened 2022-07-11
add .yaml extension

ThomasOliverKimble:
**Description**
The simulation server only parses through the `.yml` extension, when it should also support `.yaml`


BenjaminDeleze (approved):
Seems good! thank you

omichel:
In case both `.yml` and `.yalm` exist, it will send twice the ide message. We should probably add a break statement to prevent that.

PR #4878, opened 2022-07-12
Fix invalid retrieve of non-top ancestor Robot

stefaniapedrazzi:
Fix retrieving ancestor Robot node.
The previous version was returning an invalid pointer when casting the top node to `WbRobot` independently from the node type.
For example it was failing in this case:
```
Group {
  children [
    Robot {
      children [
        Speaker { }
      ]
    }
  ]
}
```

### DIFF
--- a/src/webots/gui/WbMainWindow.cpp
+++ b/src/webots/gui/WbMainWindow.cpp
@@ -2475,8 +2475,10 @@ void WbMainWindow::prepareNodeRegeneration(WbNode *node) {
       QStringList perspective = factory->windowPerspective(device);
       if (perspective.isEmpty())
         perspective = device->perspective();
-      const WbRobot *robot = dynamic_cast<const WbRobot *>(WbNodeUtilities::findTopNode(node));
-      mTemporaryProtoPerspectives.insert(robot->name() + "\n" + device->name(), perspective);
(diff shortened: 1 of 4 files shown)


src/webots/gui/WbMainWindow.cpp
```diff
@@ -2475,8 +2475,10 @@ void WbMainWindow::prepareNodeRegeneration(WbNode *node) {
       QStringList perspective = factory->windowPerspective(device);
       if (perspective.isEmpty())
         perspective = device->perspective();
-      const WbRobot *robot = dynamic_cast<const WbRobot *>(WbNodeUtilities::findTopNode(node));
-      mTemporaryProtoPerspectives.insert(robot->name() + "\n" + device->name(), perspective);
+      const WbRobot *robot = WbNodeUtilities::findRobotAncestor(device);
+      assert(robot);
+      if (robot)
```
ygoumaz, line 2480: Is it useful to check the robot variable here, as we already check it with the assertion above?

stefaniapedrazzi: Yes. The assertion is only executed in debug mode.
If we don't want that Webots crashes when running in release mode we need the check.

ygoumaz (approved):

PR #4885, opened 2022-07-13
Fix recompilation of controllers before creating packages

### LINKED ISSUE #4884: "texture_change" sample controller not recompiled after changing URLs
It seems that the controller in the `texture_change` sample simulation is not recompiled after changing the URLs in the source file just before creating the package (tested with tarball on Ubuntu 22.04).

As a result the sample doesn't work because the stone texture is not found.
To make it work, you have to relocate the project and recompile the controller.

stefaniapedrazzi:
Fix #4884:
controllers were recompiled before converting the URLs.


stefaniapedrazzi:
just note that the packages built from this branch are available at the end of this page:
https://github.com/cyberbotics/webots/actions/runs/2663339602

ygoumaz (approved):
Thank you! I have tested to make a distribution locally and also tested the CI generated one. Both seem to be complete and work well.

PR #4889, opened 2022-07-14
Fix editing cached PROTO sources

stefaniapedrazzi:
As reported in https://github.com/cyberbotics/webots/pull/4860#issuecomment-1183342479, it was still possible to edit read-only cached PROTO sources.
These issues are fixed in this PR:
* [x] disable auto-completer if file is set read-only
* [x] skip keyboard events other than copy, cut and paste to prevent editing the file
* [x] after "Save as" action, restore the `WbTextBuffer` state: enable writing and restore auto-completer


BenjaminDeleze (approved):
Thank you!

PR #4892, opened 2022-07-14
Fix updating the selection from the scene tree

### LINKED ISSUE #4891: Incorrect context menu after running supervisor.wbt
The context menu of PROTO (and other nodes) is wrong after running the `supervisor.wbt` simulation and only default actions are displayed but not Robot and PROTO specific actions.

This is due to issues in the `WbSceneTree::updateSelection()` function that wrongly detect that the selection change was triggered by a Supervisor controller action and not a user action.

stefaniapedrazzi:
Fix #4891.
This issue is also reproducible in the `master` branch, but to avoid conflicts can be fixed directly in this branch.


ad-daniel (approved):
Thank you

PR #4893, opened 2022-07-14
Fix early WbExternProtoEditor update

stefaniapedrazzi:
The WbExternProtoEditor content was previously saved as soon as the "SAVE" action was triggered but it is too early because the user could still abort the saving (from the dialog boxes that will be shown for warning about the simulation status or asking for the filename).

In this PR, instead of connecting directly to the "SAVE" action signal, the `WbExternProtoEditor::updateContents` is called just after the world is saved.


ad-daniel (approved):
Thank you!

PR #4898, opened 2022-07-14
Use generic controller in PROTO wizard and favor relative urls rather absolute when editing a proto

ad-daniel:
**Description**

- When attempting to EDIT a remote PROTO (after relocation), and saving the world, the EXTERNPROTO declaration should be relative (to the world) rather than absolute
- In the PROTO wizard, controller fields are set to `"<generic>"` by default since the Wizard doesn't relocate controllers.
- I've changed the `generate_proto_list.py` so that the PROTO headers (ex: `field   SFColor         colorOverride       1 1 1` is saved as `field SFColor colorOverride 1 1 1`), as we can't enforce good formatting anyway there's no reason to keep all those spaces
- Fix PROTO Wizard creating empty EXTERNPROTO declarations in `Road.proto`. The reason is that PROTO parameters may contain additional `{` which contrary to all other instances of this symbol doesn't entail it's preceded by a node, for example `field        MFNode{RoadLine {}}  lines                     [ RoadLine { } ] `. In this case, MFNode is a node type, not an actual node.


ad-daniel:
The PR can now be reviewed

src/webots/gui/WbNewProtoWizard.cpp
```diff
@@ -132,7 +132,10 @@ void WbNewProtoWizard::accept() {
         const QStringList parameters = info->parameters();
         for (int i = 0; i < parameters.size(); ++i) {
           if (mExposedFieldCheckBoxes[i + 1]->isChecked()) {
-            interface += "  " + parameters[i] + "\n";
+            if (parameterNames[i] == "controller")
```
ygoumaz, line 135: We could extend it to the robot windows as well.

ygoumaz (commented):
Thank you! We could also convert robot windows to generic, in addition to controllers. We may also wait for #4900 to be merged to see if the changes to `generate_proto_list.py` still work fine.

ad-daniel:
The PR can now be reviewed, the merge conflict was resolved

ygoumaz (commented):
Every changes work fine, thank you! However, I am wondering if we should even extend the `<generic>` conversion to remote PROTOs that are imported locally when editing them in Webots (right-click on PROTO + Edit PROTO Source + Copy message). After save and reload, the controller and windows are obviously not found as they were not copied to the local project.

ad-daniel:
I've looked into it a bit and although it's possible, some parts are somewhat tricky. Basically you need to:
- read the PROTO contents prior to exporting it, find any mention of "controller" or "window" (which depending on whether or not the field is exposed in the PROTO header might be tricky since the regex might find false positives or fail altogether if JavaScript is involved)
- extract the currently defined value for these fields
- check if the referenced controller/window is actually available locally (if a full relocation occurred, we don't need to change it to generic controller for example)
- if none is found, overwrite the field with "generic"

Although it's possible, I'm not sure it's worth delaying the release for it now. As it stands I think the warning should be self-explanatory: if you exported the PROTO and not its controller, it makes sense that it cannot find it and the warning should allow you to know what is the problem. I suggest we open an issue and address it after the release

ygoumaz (approved):
I will open an issue about the general `<generic>` conversion. Thank you!

PR #4900, opened 2022-07-15
Fix missing interface informations in PROTOs

ygoumaz:
As exposed in[ #4860 (comment)](https://github.com/cyberbotics/webots/pull/4860#issuecomment-1184078213), some fields are not completely captured when generating the PROTO list. In consequence, using some PROTOs (e.g. Cabinet) in the PROTO wizard as base node will create an incomplete resulting PROTO.

### DIFF
--- a/scripts/packaging/generate_proto_list.py
+++ b/scripts/packaging/generate_proto_list.py
@@ -78,7 +78,9 @@ class ProtoInfo:
                 self.description += clean_line.strip() + '\\n'
 
     def parse_parameters(self):
-        for match in re.findall(r'(?<=\s\s)((?:field|vrmlField)\s+(?:\w+|(?:\{.*\}))+[\s\S]*?(?=\s\s+field\s|\s\s+vrmlField\s|\s\s+hiddenField\s|\s\s+hidden\s|\s\s+deprecatedField\s|\s\s+unconnectedField\s|\n\n|\]\s*\{))', self.contents):
+        for match in re.findall(r'(?<=\s\s)((?:field|vrmlField)\s+(?:\w+|(?:\{.*\}))+[\s\S]*?(?=\s\s+field\s|\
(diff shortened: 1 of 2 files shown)


scripts/packaging/generate_proto_list.py
```diff
@@ -78,7 +78,7 @@ def parse_header(self):
                 self.description += clean_line.strip() + '\\n'
 
     def parse_parameters(self):
-        for match in re.findall(r'(?<=\s\s)((?:field|vrmlField)\s+(?:\w+|(?:\{.*\}))+\s+[^\n\#]+)', self.contents):
+        for match in re.findall(r'(?<=\s\s)((?:field|vrmlField)\s+(?:\w+|(?:\{.*\}))+[\s\S]*?(?=\s\sfield|\s\svrmlField|\s\shiddenField|\s\shidden|\s\sdeprecatedField|\s\sunconnectedField|\]\n\{))', self.contents):
```
ad-daniel, line 81:
```suggestion
        for match in re.findall(r'(?<=\s\s)((?:field|vrmlField)\s+(?:\w+|(?:\{.*\}))+[\s\S]*?(?=\s\sfield|\s\svrmlField|\s\shiddenField|\s\shidden|\s\sdeprecatedField|\s\sunconnectedField|\]\s*\{))', self.contents):
```

ad-daniel (changes requested):
Unfortunately when I run it in debug mode the famous assert triggers: `Assertion mParameters.size() == mParameterNames.size()' failed.` which means that the parsing on webots or python side is wrong. Printing the url, it seems the culprit is `Mirror.proto`

ygoumaz:
In addition to Mirror.proto, the new regex in the python script captures a new field in Nao.proto which was not parsed correctly on Webots side. Both PROTOs are now completely captured and the assertion never fails for any proto.

PR #4903, opened 2022-07-15
Improve robustness

ad-daniel:
**Description**

This PR changes:
- tests of the extension should be case-insensitive
- the way urls were combined worked well, but the robustness of the solution was limited. The important change done is at this spot: https://github.com/cyberbotics/webots/blob/be46d7978b81d84fcc3fa9cc9382bbaa8b21429e/src/webots/nodes/utils/WbUrl.cpp#L294-L305

A good way of testing the change to combinePaths is to add

```
  if (parent == NULL) {
    combinePaths("../../asd.proto",
                 "https://raw.githubusercontent.com/cyberbotics/webots/projects/devices/sick/protos/SickS300.proto");
   // etc
  }
```

to the constructor of WbProtoTreeItem (or somewhere else where are guaranteed to pass) and print the results for different combinations (webots://, absolute, remote)

### DIFF
--- a/src/webots/core/WbProject.cpp
+++ b/src/webots/core/WbProject.cpp
@@ -161,7 +161,7 @@ QString WbProject::computeBestPathForSaveAs(const QString &fileName) {
 }
 
 WbProject::WbProject(const QString &path) {
-  if (path.endsWith(".wbt")) {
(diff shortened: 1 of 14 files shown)


BenjaminDeleze (approved):
Seems to work well, thank you!

PR #4907, opened 2022-07-18
Update "Transform to" rules

stefaniapedrazzi:
Update "Transform to" rules:
* update rules assuming that solid nodes should have a solid parent
* remove outdated `WbNodeUtitilies::hasSolidChildren` function and replace it by `WbNodeUtilities::hasADeviceDescendant()`  and `WbNodeUtilities::hasAJointDescendant()` depending on the destination type 
* cache costly repeated searches for device descendants (other descendant search are executed only twice)
* ensure that only nodes with a solid ancestor can be transformed to `Connector` (*)

Additionally:
* update the insert rules to prevent inserting a `Connector` node without a parent `Solid` node or as top node (*)
* update paste rules preventing to paste a device in `Robot` node not located at the top level
* ignore `Connector` devices in `WbNodeUtilities::hasADeviceDescendant()` function given that is is only used to detect if a `Robot` ancestor is needed (not the case for the `Connector`)

(*) The `WbConnector` code assumes an upper Solid node exists.

### DIFF
--- a/src/webots/nodes/utils/WbNodeUtilities.cpp
+++ b/src/webots/nodes/utils/WbNodeUtilities.cpp
@@ -1752,7 +1752,7 @@ bool WbNodeUtilities::isAllowedToInsert(const WbField *const field, const QStrin
 }
 
 WbNodeUtilities::Answer WbNodeUtilities::isSuitableForTransform(const WbNode *const srcNode, const QString &destModelName,
-                                                                bool *hasDeviceChildren) {
+                                                                int *hasDeviceDescendantFlag) {
   const QString &srcModelName = srcNode->nodeModelName();
 
   // cannot transform into same type
@@ -1786,12 +1786,15 @@ WbNodeUtilities::Answer WbNodeUtilities::isSuitableForTransform(const WbNode *co
 
   if (destModelName == "Group" || destModelName == "Transform") {
     if (isSolidTypeName(srcModelName)) {
-      if (!hasDeviceChildren) {
-        // cache value
-        hasDeviceChildren = new bool[1];
-        *hasDeviceChildren = hasADeviceDescendant(srcNode, true);
+      bool hasDevices;
+      if (hasDeviceDescendantFlag && hasDeviceDescendantFlag[0] >= 0) {  // read cached value
+        hasDevices = hasDeviceDescendantFlag[0] == 1;
+      } else {
+        hasDevices = hasADeviceDescendant(srcNode, true);
+        if (hasDeviceDescendantFlag)
+          hasDeviceDescendantFlag[0] = hasDevices ? 1 : 0;
       }
-      if (hasDeviceChildren[0] || hasAJointDescendant(srcNode))
+      if ((hasDevices && !hasARobotAncestor(srcNode)) || (hasAJointDescendant(srcNode) && !findUpperSolid())
         return UNSUITABLE;
       return !findUpperSolid(srcNode) && hasADeviceDescendant(srcNode, false) ? UNSUITABLE : LOOSING_INFO;
     }
@@ -1801,15 +1804,18 @@ WbNodeUtilities::Answer WbNodeUtilities::isSuitableForTransform(const WbNode *co
 
   if (isRobotTypeName(srcModelName)) {
     if (destModelName == "Solid" || destModelName == "Charger" || destModelName == "Connector") {
-      if (!hasDeviceChildren) {
-        // cache value
-        hasDeviceChildren = new bool[1];
-        *hasDeviceChildren = hasADeviceDescendant(srcNode, true);
+      bool hasDevices;
+      if (hasDeviceDescendantFlag && hasDeviceDescendantFlag[0] >= 0)  // read cached value
+        hasDevices = hasDeviceDescendantFlag[0] == 1;
+      else {
+        hasDevices = hasADeviceDescendant(srcNode, true);
+        if (hasDeviceDescendantFlag)
+          hasDeviceDescendantFlag[0] = hasDevices ? 1 : 0;
       }
       if (destModelName == "Solid" || destModelName == "Charger")
-        return hasDeviceChildren[0] ? UNSUITABLE : LOOSING_INFO;
+        return hasDevices ? UNSUITABLE : LOOSING_INFO;
       if (destModelName == "Connector") {
-        return (hasDeviceChildren[0] || !findUpperSolid(srcNode)) ? UNSUITABLE : LOOSING_INFO;
+        return (hasDevices || !findUpperSolid(srcNode)) ? UNSUITABLE : LOOSING_INFO;
       }
     }
 
--- a/src/webots/nodes/utils/WbNodeUtilities.cpp
+++ b/src/webots/nodes/utils/WbNodeUtilities.cpp
@@ -636,9 +636,7 @@ namespace {
     return false;
   }
 
-  bool isSolidNode(WbBaseNode *node) {
-    return dynamic_cast<WbSolid *>(node);
-  }
+  bool isSolidNode(WbBaseNode *node) { return dynamic_cast<WbSolid *>(node); }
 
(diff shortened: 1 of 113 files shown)


stefaniapedrazzi:
I didn't check if it is a regression, but on the master branch it is possible to transform a top Robot to a device.
So we need to fix this for the release.

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -1785,72 +1769,69 @@ WbNodeUtilities::Answer WbNodeUtilities::isSuitableForTransform(const WbNode *co
...
-    if (destModelName == "Group" || destModelName == "Transform") {
-      if (!hasSolidChildren(srcNode))
-        return LOOSING_INFO;
-
-      return UNSUITABLE;
+  if (destModelName == "Group" || destModelName == "Transform") {
+    if (isSolidTypeName(srcModelName)) {
+      if (!hasDeviceChildren) {
+        // cache value
+        hasDeviceChildren = new bool[1];
```
omichel, line 1791: There is a memory leak here: `hasDeviceChildren` will never be free.

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -1785,72 +1769,69 @@ WbNodeUtilities::Answer WbNodeUtilities::isSuitableForTransform(const WbNode *co
...
-          return SUITABLE;
+    return UNSUITABLE;
+  }
 
-        return UNSUITABLE;
+  if (isRobotTypeName(srcModelName)) {
+    if (destModelName == "Solid" || destModelName == "Charger" || destModelName == "Connector") {
+      if (!hasDeviceChildren) {
+        // cache value
+        hasDeviceChildren = new bool[1];
```
omichel, line 1806: Memory leak here as well.

src/webots/scene_tree/WbSceneTree.cpp
```diff
@@ -620,7 +620,10 @@ void WbSceneTree::transform(const QString &modelName) {
   assert(dynamic_cast<WbGroup *>(currentNode));
 
   // check if loosing information
-  if (WbNodeUtilities::isSuitableForTransform(currentNode, modelName) == WbNodeUtilities::LOOSING_INFO) {
+  bool *hasDeviceChildren = NULL;
+  const WbNodeUtilities::Answer answer = WbNodeUtilities::isSuitableForTransform(currentNode, modelName, hasDeviceChildren);
+  delete hasDeviceChildren;
```
omichel, line 625: This deletes a NULL pointer (useless).

src/webots/user_commands/WbContextMenuGenerator.cpp
```diff
@@ -50,13 +50,17 @@ namespace WbContextMenuGenerator {
...
+      // cache intensive searches results
+      bool *hasDeviceChildren = NULL;
       // find all nodes suitable for transform
       foreach (const QString &modelName, basicModels) {
-        WbNodeUtilities::Answer answer = WbNodeUtilities::isSuitableForTransform(selectedNode, modelName);
+        const WbNodeUtilities::Answer answer =
+          WbNodeUtilities::isSuitableForTransform(selectedNode, modelName, hasDeviceChildren);
         if (answer != WbNodeUtilities::UNSUITABLE) {
           suitableModels << modelName;
         }
```
omichel, line 61:
```suggestion
        if (answer != WbNodeUtilities::UNSUITABLE)
          suitableModels << modelName;
```

src/webots/user_commands/WbContextMenuGenerator.cpp
```diff
@@ -50,13 +50,17 @@ namespace WbContextMenuGenerator {
...
       }
+      delete hasDeviceChildren;
```
omichel, line 63: Useless.

omichel (changes requested):

src/webots/nodes/utils/WbNodeUtilities.cpp
stefaniapedrazzi: No, there is no memory leak current in the code.
The hasDeviceChildren pointer is deleted where the `WbNodeUtilities::isSuitableForTransform` is called.

src/webots/scene_tree/WbSceneTree.cpp
stefaniapedrazzi: This is needed to avoid memory leak.

src/webots/nodes/utils/WbNodeUtilities.cpp
omichel: I am pretty sure there is a memory leak here.
The `hasDeviceChildren` pointer is not passed by reference, hence it is not modified outside the scope of this method, thus the memory leak. You can check the value you are deleting afterwards, it will always be NULL.

stefaniapedrazzi: You are right.
Instead of fixing and passing the pointer by reference so that the value can be modified, I simplified the code and passed a reference to an int.
It is not as clean as using a pointer for storing the three states (not computer, true and false) but the code is simpler and less error-prone.

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -1785,72 +1769,81 @@ WbNodeUtilities::Answer WbNodeUtilities::isSuitableForTransform(const WbNode *co
...
-        return LOOSING_INFO;
+  if (destModelName == "Group" || destModelName == "Transform") {
+    if (isSolidTypeName(srcModelName)) {
+      bool hasDevices;
+      if (hasDeviceDescendantFlag && hasDeviceDescendantFlag[0] >= 0) {  // read cached value
+        hasDevices = hasDeviceDescendantFlag[0] == 1;
+      } else {
+        hasDevices = hasADeviceDescendant(srcNode, true);
+        if (hasDeviceDescendantFlag)
+          hasDeviceDescendantFlag[0] = hasDevices ? 1 : 0;
```
omichel, line 1795:
```suggestion
          *hasDeviceDescendantFlag = hasDevices ? 1 : 0;
```

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -1785,72 +1769,81 @@ WbNodeUtilities::Answer WbNodeUtilities::isSuitableForTransform(const WbNode *co
...
-          return SUITABLE;
+    return UNSUITABLE;
+  }
 
-        return UNSUITABLE;
+  if (isRobotTypeName(srcModelName)) {
+    if (destModelName == "Solid" || destModelName == "Charger" || destModelName == "Connector") {
+      bool hasDevices;
+      if (hasDeviceDescendantFlag && hasDeviceDescendantFlag[0] >= 0)  // read cached value
+        hasDevices = hasDeviceDescendantFlag[0] == 1;
```
omichel, line 1815:
```suggestion
      if (hasDeviceDescendantFlag && *hasDeviceDescendantFlag >= 0)  // read cached value
        hasDevices = *hasDeviceDescendantFlag == 1;
```

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -1785,72 +1769,81 @@ WbNodeUtilities::Answer WbNodeUtilities::isSuitableForTransform(const WbNode *co
...
+      else {
+        hasDevices = hasADeviceDescendant(srcNode, true);
+        if (hasDeviceDescendantFlag)
+          hasDeviceDescendantFlag[0] = hasDevices ? 1 : 0;
```
omichel, line 1819:
```suggestion
          *hasDeviceDescendantFlag = hasDevices ? 1 : 0;
```

omichel (changes requested):

stefaniapedrazzi:
Windows tests failures are not related to the changes in this PR.

omichel (approved):
Thank you.

Issue #4818, opened 2022-07-01
Improvements to the test suite


ad-daniel:
**Describe the Bug**

- tests in the cache directory should be run one-by-one (i.e. relaunching webots every time and clearing the cache in-between), adding this possibility might require significantly refactor `test_suite.py` and the `test_suite_supervisor.py`
- `local_proto_with_texture.wbt` (or more in general the cache tests) should be run in the CI in the **development environment**
- robot window still broken for ubuntu 22 in the CI

PR #4908, opened 2022-07-18
Improve tests: restart Webots after each "cache" test

stefaniapedrazzi:
Fix first issue in  #4818.


tests/test_suite.py
```diff
@@ -226,76 +214,37 @@ def monitorOutputFile(finalMessage):
...
-        os.makedirs(protosTargetDirectory)
-    for protoFileName in protoFileNames:
-        shutil.copyfile(protosSourceDirectory + protoFileName,
-                        protosTargetDirectory + os.sep + protoFileName)
-    worldsCount = generateWorldsList(groupName, worldsFilename)
-    firstSimulation = findFirstWorldFilename(worldsFilename)
-    if not os.path.exists(firstSimulation):
-        continue
-
+    indexFilename = testsFolderPath + groupName + os.sep + 'worlds_index.txt'
```
omichel, line 230: Why not using [os.path.join](https://docs.python.org/3/library/os.path.html#os.path.join) which [seems to be a better practice](https://stackoverflow.com/a/16789790/810268) than os.sep? (this could apply to other places in the code)
```suggestion
    indexFilename = os.path.join(testsFolderPath, groupName, 'worlds_index.txt')
```

omichel (approved):
Looks good to me.
I just have a comment regarding the use of `os.sep`.

stefaniapedrazzi:
> Looks good to me. I just have a comment regarding the use of `os.sep`.

I didn't modify the old code, but just moved it.
I will pass through the whole file and fix it.

stefaniapedrazzi:
I will fix all the uses of the `os.sep` instead of `os.path.join`  in the other Pythons scripts in a different PR.
Here I just fixed the ones in the test suite (not test sources) related to this PR changes.

omichel (approved):
Thank you.

omichel:
Note: chocolatey is currently under maintenance. We can safely ignore the failing Windows tests.

PR #4909, opened 2022-07-18
Add flag to use mingw python

Kreijstal:
Basically adds a flag to windows make files so that it can compile with
mingw python as well as downloading python version externally
This is for convenience and also reduce the non interactive dependencies if you want to package it later

**Description**
So it's really hard to use the mingw's version of python with webots, in this pull request I edit the makefiles in such a way that it retains the current behaviour but given a flag it uses mingw's version of python instead of using python home


omichel (approved):
Thank you for this nice contribution.
I will also document it on the [wiki page](https://github.com/cyberbotics/webots/wiki/Windows-Optional-Dependencies/#python).

Note: you will be able to merge this PR once you accept my GitHub invite to become a member of The Cyberbotics Committers group.

omichel:
Thank you!
See my wiki addition [here](https://github.com/cyberbotics/webots/wiki/Windows-Optional-Dependencies/_compare/f0190196b3265d671530ed831252c30ecbcd12a8...8ed9ba01ff7405bfc4372834d16e6714ad4ba613).

PR #4913, opened 2022-07-19
Disable application edit actions for read-only text files

stefaniapedrazzi:
Complete #4889 (addressing WbTextBuffer only) by preventing and disabling the application actions:
undo, redo, cut, paste, save and replace actions on read-only text buffer.


BenjaminDeleze (approved):
Thank you

PR #4914, opened 2022-07-19
Fixed wrong assert in WbContactSound.cpp

omichel:
Ditto


ad-daniel:
In theory it has been addressed in the other branch already, but personally I've removed the check entirely. The reason is it doesn't seem to be wrong for `contactProperties` to be NULL, there's plenty of worlds that don't necessarily need to define some contact properties and therefore the assert would trigger quite often for no reason.

omichel:
All right.

PR #4915, opened 2022-07-19
Removed useless test after assert

omichel:
Since we have an assert on `robot`, we can assume that from this point `robot` will never be `NULL`. Thus, the test is superfluous.


ad-daniel (approved):
Thank you

PR #4920, opened 2022-07-20
Improve error messages for worlds and protos prior to R2022b

ygoumaz:
As mentioned in [#4860 (comment)](https://github.com/cyberbotics/webots/pull/4860#issuecomment-1187669542), old worlds don't display the complete list of missing EXTERNPROTO declarations. This PR also makes the distinction between old PROTO files (prior to R2022b) and new ones to show correct warnings.

Error messages are updated to help the user add the correct declarations in his files.

In addition, the condition on known models needs to be changed to trigger in the correct situation.


ad-daniel (dismissed):

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -132,29 +132,31 @@ WbProtoModel *WbProtoManager::findModel(const QString &modelName, const QString
...
+    const QString backwardsCompatibilityMessage =
+      tr("Please adapt your project to R2022b following these instructions: "
+         "https://github.com/cyberbotics/webots/wiki/How-to-adapt-your-world-or-PROTO-to-Webots-R2022b");
+    const QString outdatedProtoMessage =
+      tr("'%1' must be converted because EXTERNPROTO declarations are missing.").arg(parentFilePath);
+    const QString errorMessage =
+      tr("Missing declaration for '%1': Add 'EXTERNPROTO \"%2\"' in '%3'.")
+        .arg(modelName)
+        .arg(protoDeclaration.isEmpty() ? mWebotsProtoList.value(modelName)->url() : protoDeclaration)
+        .arg(parentFilePath);
```
ad-daniel, line 144: I think we can lower the scope of the messages to where they are actually used, no need to define them all when we only show some of them

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -254,6 +259,12 @@ QMap<QString, QString> WbProtoManager::undeclaredProtoNodes(const QString &filen
     return protoNodeList;
   // fill queue with nodes referenced by the world file
   queue << parser.protoNodeList();
+  const QString backwardsCompatibilityMessage =
+    tr("Please adapt your project to R2022b following these instructions: "
+       "https://github.com/cyberbotics/webots/wiki/How-to-adapt-your-world-or-PROTO-to-Webots-R2022b");
+
+  displayMissingDeclarations(backwardsCompatibilityMessage);
```
ad-daniel, line 266: Since here there's just one message, I think we can pass it directly without creating an intermediary variable

ad-daniel (commented):
Looks good, only some details. I'll do more extensive checks when https://github.com/cyberbotics/webots/pull/4930 is merged into this PR

ad-daniel:
From what I've been able to test, the behavior appears to be correct. The general approach is to display the wiki link if R2022a world/protos are detected whereas to display which declarations are missing if the world/protos are from R2022b onward (the reason being, you don't want to have the users do it manually if it involves a multitude of changes). The only case which is only partially handled is if the missing declaration of a PROTO does exist and is on the world side, but although a bit strange it's not entirely wrong either, as the same 2 objects can't have different urls anyway. Whatever the case, if this hanging declaration has no instances, as soon as the world is saved the declaration is purged and from that point what and where it's missing will be shown. It's not ideal but isn't that common an occurrence either, you kinda need to chase for it. The "proper" solution involves removing the check that searches for a declaration in the mExternProto list in findModel function. Doing so solves the issue, but opens up a new one, namely that when you:
1. Insert a node with add-node
2. Don't save the world (so the declaration for the added node exists only in mExternProto list which hasn't been saved yet)
3. Cut the node you just added and attempt to paste it
4. When pasting, findModel has no way of knowing where to find this node, currently it infers it's declaration by searching for one in the mExternProto list.

In short, if the mExternProto check is removed to solve the other "oddity", a solution needs to be found to handle this case which as it stands I don't see a better way.

I think we can merge this PR, and if @stefaniapedrazzi thinks the issues she found have a been addressed, we can merge the other PR into master and a release candidate can be built

omichel:
> In short, if the mExternProto check is removed to solve the other "oddity", a solution needs to be found to handle this case which as it stands I don't see a better way.

One solution could be to include the EXTERNPROTO url in the cut buffer, so that when we paste a cut PROTO, we can also re-insert the EXTERNPROTO url if needed.

ad-daniel:
The issue isn't that a declaration (url to the node location) is missing during the cut/paste action, but rather that since findmodel MUST search among the `mExternProto` list for the declaration of inserted nodes (since they aren't saved yet, they're only available there), in some situations such as the one mentioned by Stefania, by allowing the search among ALL the mExternProto declarations (not just the inserted ones), situations that should raise a warning silently don't. In other worlds, it's a matter of priority. Since a match is found in `mExternProto` too early, the declaration doesn't need to be injected by backwards compatibility and therefore a warning isn't shown. The solution is to restrict the search among item  in the `mExternProto` list only to inserted nodes.

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -812,6 +827,16 @@ void WbProtoManager::updateExternProto(const QString &protoName, const QString &
   assert(false);  // should not be requesting to change something that doesn't exist
 }
 
+QString WbProtoManager::cleanupExternProtoPath(const QString &url) {
+  QString path = url;
+  if (path.startsWith(WbStandardPaths::webotsHomePath()))
+    path = path.replace(WbStandardPaths::webotsHomePath(), "webots://");
```
omichel, line 833:
```suggestion
    path.replace(WbStandardPaths::webotsHomePath(), "webots://");
```

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -875,3 +903,20 @@ QString WbProtoManager::injectDeclarationByBackwardsCompatibility(const QString
...
 }
+
+void WbProtoManager::displayMissingDeclarations(QString message) {
+  if (!mUniqueErrorMessages.contains(message)) {
+    mUniqueErrorMessages << message;
+    WbLog::error(message);
+  }
+}
+
+WbVersion WbProtoManager::checkProtoVersion(QString protoUrl, bool *foundProtoVersion) {
```
omichel, line 914:
```suggestion
WbVersion WbProtoManager::checkProtoVersion(const QString &protoUrl, bool *foundProtoVersion) {
```

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -121,40 +121,44 @@ WbProtoModel *WbProtoManager::findModel(const QString &modelName, const QString
...
+      displayMissingDeclarations(errorMessage);
     }
+    if (protoDeclaration.isEmpty())
+      return NULL;
   }
 
   // a PROTO declaration is provided, enforce it
   QString modelPath;  // how the PROTO is referenced
-  if (WbUrl::isWeb(protoDeclaration)) {
+  if (WbUrl::isWeb(protoDeclaration) && WbNetwork::instance()->isCached(modelPath)) {
```
omichel, line 160:
```suggestion
  if (WbUrl::isWeb(protoDeclaration) && WbNetwork::instance()->isCached(modelPath))
```

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -121,40 +121,44 @@ WbProtoModel *WbProtoManager::findModel(const QString &modelName, const QString
...
     modelPath = protoDeclaration;
-    assert(WbNetwork::instance()->isCached(modelPath));
   } else if (WbUrl::isLocalUrl(protoDeclaration)) {
```
omichel, line 162:
```suggestion
  else if (WbUrl::isLocalUrl(protoDeclaration)) {
```

omichel (changes requested):

omichel (approved):

PR #4923, opened 2022-07-20
Fixed execution of homebrew python controllers on x86_64

### LINKED ISSUE #4919: Controller crash with Python from homebrew
I can run controllers normally with Python installed in:
```
/Library/Frameworks/Python.framework/Versions/3.9/bin/python3.9
or 
/usr/local/bin/python3.9
```

but the controllers crash when trying with:
```
/usr/local/Cellar/python@3.8/3.8.6_2/python3.9
or 
/usr/local/opt/python@3.8/libexec/bin/python
```

omichel:
Fixes #4919.

It seems the execution of homebrew python controllers was broken after introducing the support for M1.

- The Makefile was not rebuilding the `_brew` version of the controllers.
- The `PYTHONPATH` was set to brew only if executing brew python from the `/usr/local/Cellar` folder. I also added the `/usr/local/opt` folder which is a common way to refer to the brew version of python as well.

Note: the M1 brew python is not supported because brew doesn't provide universal binaries and it would be complicated to support it (we would need to build universal binaries ourselves for brew which is not a trivial task).


BenjaminDeleze (approved):
Thank you

PR #4501, opened 2022-05-03
Cleanup Robot window and controller specification

ThomasOliverKimble:
**Description**
- Since we want to be able to specify that a robot has no robot window, we should change the specification of the robot `window` in the `Robot` node.
- To make it more consistent with other parameters, we should also change the `controller` fields and the `WorldInfo` node.
- Some other minor bugs and behavior modifications concerning the robot windows have been addressed as well.
- Some modifications were needed to be done in WebotsJS for correct behavior.

**Related Issues**
This pull-request fixes issue #4429.

**Tasks**

Webots:
  - [x] `Robot.controller "void"` &rarr; `"<generic>"` (default value: run the legacy "void" controller, to be renamed to "generic") .
  - [x] `Robot.controller ""` &rarr; `"<none>"` (run no controller).
  - [x] `Robot.controller "<extern>"` &rarr; unchanged (run an extern controller).
  - [x] `Robot.window ""` &rarr; `"<generic>"` (default value: use the generic robot window).
  - [x] `Robot.window "<none>"` (no robot window).
  - [x] `Robot.remote_control ""` &rarr; `"<none>"` (default value: no remote control library).
  - [x] `WorldInfo.physics ""` &rarr; `"<none>"` (default value: no physics plugin).
  - [x] `WorldInfo.window ""` &rarr; `"<none>"` (default value: no world info window).
  - [x] Update `.proto`, `.wbt` and `.md` files.
  - [x] Update documentation.
 
 WebotsJS:
  - [x] Update web view to not show `<none>` robot windows.
  - [x] Update web view to use `<generic>` window by default.

**Documentation**
Modifies all documentation as well as `.proto` and `.wbt` files mentioning the new specifications.

**Screenshots**
controller_fields
robot_window_fields
world_info

PR #4924, opened 2022-07-20
Add missing window fields to protos

ThomasOliverKimble:
**Description**
One of the goals of the specifications update for robot windows in #4501 was to be able to select `<none>` for robot windows if we did not want them to show up (in benchmarks for example). However, most robots were missing the declaration of the `window` field, so it was impossible to select or add `<none>`.

PROTOs have now been updated.


omichel (approved):
Looks good to me.
Thank you.

PR #4926, opened 2022-07-20
Update SoundEngine's viewpoint after regeneration

ad-daniel:
**Description**
Ditto.

### DIFF
--- a/src/webots/sound/WbSoundEngine.cpp
+++ b/src/webots/sound/WbSoundEngine.cpp
@@ -50,7 +50,6 @@
 #include <cassert>
 
 static WbWorld *gWorld = NULL;
-static WbViewpoint *gViewpoint = NULL;
 static bool gOpenAL = false;
 static bool gMute = true;
 static int gVolume = 80;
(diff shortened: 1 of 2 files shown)


src/webots/sound/WbSoundEngine.cpp
```diff
@@ -128,20 +127,24 @@ bool WbSoundEngine::openAL() {
 void WbSoundEngine::setWorld(WbWorld *world) {
   if (world) {
     gWorld = world;
-    gViewpoint = world->viewpoint();
-    if (gViewpoint)
-      QObject::connect(gViewpoint, &WbViewpoint::cameraParametersChanged, &WbSoundEngine::updateListener);
+    if (gWorld && gWorld->viewpoint())
```
omichel, line 130: That should be sufficient.
```suggestion
    if (gWorld)
```

omichel (changes requested):

src/webots/sound/WbSoundEngine.cpp
ad-daniel: Indeed, doesn't appear to be necessary

omichel (approved):
Thank you.

ad-daniel:
Ah right, didn't notice `world` was already checked above

omichel (approved):

PR #4935, opened 2022-07-21
Remove useless DLLs in the Windows package

omichel:
`/mingw64/bin/libjasper-4.dll` is now missing on MSYS2 and hence it was preventing the Windows distribution from building.
Other DLLs are actually not used any more by Webots, so this is some clean-up.
I tested the resulting package, including movie creation and Darwin-op robot window and everything works smoothly without these DLLs.


BenjaminDeleze (approved):
Thank you

PR #4936, opened 2022-07-21
Improve Python scripts

### LINKED ISSUE #1428: Remove 'global' Variables from Python Code

stefaniapedrazzi:
* Replace direct `os.sep` usage in Python scripts by built-in `os.path.join` and `os.path.normpath` function.
* Remove global variables  (fix #1428 for the webots repo)


omichel (approved):
Looks good to me.
Thank you.

PR #4937, opened 2022-07-21
Fix `generate_objects_doc.py`

stefaniapedrazzi:
Currently the tag name used in `generate_objects_doc.py` doesn't match the one generated in the `proto-list.xml` file and the script to generate the objects documentation is broken.


BenjaminDeleze (approved):
I thought I fixed it some time ago, but it was only on develop :(

Thank you!

Issue #4942, opened 2022-07-25
c3d controller broken for Python >= 3.9


stefaniapedrazzi:
Controller crashes with the following error:
```
  File "/home/stefi/webots/projects/humans/c3d/controllers/c3d_viewer/c3d.py", line 329, in _as_array
    elems.fromstring(self.bytes)
AttributeError: 'array.array' object has no attribute 'fromstring'
```

This method has been removed in Python 3.9 (was deprecated at 3.2)

PR #4943, opened 2022-07-25
Fix c3d controller

stefaniapedrazzi:
Address #4942:
`array.fromstring` function (deprecated in Python 3.2 and removed in Python 3.9) is renamed `array.frombytes`:
https://docs.python.org/3/library/array.html#array.array.frombytes


omichel (approved):
Thank you.

PR #4947, opened 2022-07-26
Debug world-update cache difference

ad-daniel:
**Description**

Ditto.


ad-daniel:
Since the action `world update` passes on this PR but failed last night, it isn't easy to debug. Merging this PR and letting it run tonight might give some clues

omichel (approved):

ad-daniel:
The world update test passed tonight without any change, must have been a fluke. Perhaps it's due to the fact that yesterday we had both a nightly and a release candidate, maybe it downloaded and did the latter. Either way I don't think it's a bad idea to leave the additional prints in the workflow

Issue #4954, opened 2022-07-26
Exporting from Add Node dialog overwrite local file without warning


stefaniapedrazzi:
Clicking on the "Export" button on the Add Node dialog will overwrite the local file (if it exists) without any warning.

PR #4956, opened 2022-07-26
Show warning if file to be exported exists already

ad-daniel:
**Description**

Fixes: https://github.com/cyberbotics/webots/issues/4954


src/webots/Makefile
```diff
@@ -33,7 +33,7 @@ WB_MATHS_INCLUDE         = $(QT_CORE_INCLUDE)
 WB_CORE_INCLUDE          = $(QT_CORE_INCLUDE) $(QT_NETWORK_INCLUDE) $(QT_GUI_INCLUDE) $(QT_QML_INCLUDE) $(LUA_INCLUDE) -Iqml -Iexternal/compilation_timestamp
 WB_UTIL_INCLUDE          = $(QT_CORE_INCLUDE) $(ODE_INCLUDE) -Isound -Icore
 WB_WREN_INCLUDE          = $(QT_CORE_INCLUDE) $(QT_GUI_INCLUDE) $(QT_OPENGL_INCLUDE) $(WREN_INCLUDE) -Icore -Imaths -Ivrml -Iutil
-WB_VRML_INCLUDE          = $(QT_CORE_INCLUDE) $(WB_APP_INCLUDE) $(WB_NODES_INCLUDE) -Icore -Imaths -Ivrml
+WB_VRML_INCLUDE          = $(QT_CORE_INCLUDE) -Iapp -Icore -Imaths -Inodes/utils -Ivrml
```
stefaniapedrazzi, line 36: Here there is a bad circular dependency between `vrml` and (`app`, `nodes/utils`) module.
`vrml` should not depend on the `app` or `nodes` module.

It doesn't seem to be added in this PR, however it should be fixed and it would be better to already avoid using any classes/namespace from `app` and `nodes` in the `vrml` module.

ad-daniel: If I remember right, I think I added those because of WbProtoManager, since after the change it's needed in a whole lot of places where the previous class (WbProtoList) wasn't necessary. I think this isn't a problem at the moment since this class is a singleton (probably why it compiles), but it might indeed be sufficient to move it somewhere else (dunno where) so as to remove possible future issues.

ad-daniel: Actually the issue is that WbProtoManager and WbProtoTreeItem (`vrml`) depend on WbUrl (`nodes/utils`) for all url manipulation stuff, and since WbProtoManager is supposed to trigger the definite load of a world (2nd load), it also depends on WbApplication (`app`). Not sure what is the better way of untangling these dependencies

stefaniapedrazzi: One option would be to move `WbUrl` from `node/utils` to another module.
If it could be moved to `core` then it could be used from the `vrml` module without issues.
To achieve it, it should be needed to move the `WbUrl::orderedSearchPaths` function elsewhere but currently it is only used in the `WbExtendedStringEditor` class so doesn't seem a problem.

omichel: This was introduced in the huge PR #4551.

omichel: Fixing this circular dependency is not trivial. It requires a re-factoring of several classes with many changes in methods, parameters, etc. I believe this is out of the scope of this PR as it wasn't introduced here and would make this PR difficult to review. So, I will open an issue about it and it should be resolved in a different PR.

stefaniapedrazzi: I agree.
But I would avoid adding additional dependencies in this PR that should be fixed again later.

omichel (approved):
Thank you.

Issue #4964, opened 2022-07-27
Confusion between PROTOs that should be distinguished


ygoumaz:
If 2 PROTOs (1 local and 1 remote) have the same name, adding one of them to the list of IMPORTABLE EXTERNPROTOs will always spawn the remote one.

**Steps to Reproduce**
1. Create a new PROTO with the PROTO wizard from any object (e.g. Cabinet). Give it the same name as an official Webots PROTO, but different than the PROTO you have created it from (e.g. Nao).
2. Open the IMPORTABLE EXTERNPROTO list editor.
3. Click on `Insert new`.
4. Select the local PROTO you have created in "current project".
5. Spawn this PROTO with a supervisor.
6. The official Nao will be imported, instead of the Cabinet named Nao.

Moreover, after step 4, if the world is saved, the Nao is declared in the world file as IMPORTABLE EXTERNPROTO but with the remote URL and not the local one.

This behavior is probably 'normal' right now, as we don't allow/handle multiple PROTOs with same names. But #4953 raises a similar problem and PROTOs should be stored according to their URL instead of their names to allow such behavior.

omichel:
Can we close this one, as it seems it was fixed in #4967?

ad-daniel:
Indeed

PR #4967, opened 2022-07-27
Fix same PROTO having different URLs

ad-daniel:
**Description**

Fixes https://github.com/cyberbotics/webots/issues/4964
Fixes https://github.com/cyberbotics/webots/issues/4953 (point 1)

If different URL for equally named PROTO are provided at different levels in the hierarchy, the existing logic would return the first available (existing) model by name, silently ignoring the declaration.


ygoumaz (changes requested):
I have not checked the code yet, only tested the behavior. It seems to work as intended in most of the cases, however I have encountered some specific situations where the behavior is not correct. I spotted 2 of them that you can reproduce with the following steps:

1. Create a new project.
2. Create a new PROTO with the wizard. Choose an object (e.g. a Fridge) as reference but name it Nao.
3. Open the IMPORTABLE EXTERNPROTO panel and click on `Insert new`.
4. Choose the Nao(Solid) you just created from the Fridge.
5. Save the world. The URL is relative and correct in the world file.
6. Now, by clicking on the `Insert new` button, no Nao is available, which is correct. With the `Add Node` button, only the local one can be added, which I believe is coherent and correct too.
7. Now remove the Nao from the list of IMPORTABLE EXTERNPROTOs and save the world again.
8. Click on `Add Node`. **It is impossible to add the official Nao(Robot), which should be possible.** 
9. For the second issue, click now on `Insert new` and select the official Nao(robot) and save the world. The URL of the declaration in the world file is in `webots://` format which is correct.
10. Now remove the Nao from the IMPORTABLE EXTERNPROTO list, click on `Insert New` and add the local Nao(Solid) PROTO. 
11. Save the world again. Now the URL in the world file is absolute. **It should be relative (../protos) as in step 5.**

ygoumaz:
Here is the current status of this PR:

- As discussed, `Add node` and `Insert new` only display/gray out PROTOs that can't be added. Also cut/paste is now using a specific externProtoCutBuffer to store the value of the last cut PROTO, because the mExternProto list is purged at each delete in the scene tree and not only at save anymore. Moreover, if the user tries to Add a node with the same name as the one stored in the buffer, a warning will be triggered. Following bugs must still be fixed:
- Two PROTOs have the same name (local and remove Ball for example) and the user adds one of them to the world with `Add Node`, let's say the remote one. Then the user cuts this particular PROTO from the scene tree (it was the only and last Ball of the world). Then he clicks on `Add Node` and inserts the local Ball, it will display the warning, spawn the local Ball and (theoretically) clear the buffer. But now pasting is still possible and the world now contains two Ball instances, one remote and one local, which is forbidden.
- In the following project [bug.zip](https://github.com/cyberbotics/webots/files/9211343/bug.zip), it is impossible to add a RectangleArena with `Add Node`, because the URL is in `raw.github.com` format instead of `webots://`. This won't happen for users, so it isn't theoretically a bug.
- Intensive tests should be performed to check all possible cases including combinations of the following operations : cut/paste, saving, reloading, check the IMPORTABLE and Add Node lists to see if correct elements are forbidden.

omichel:
I am closing this PR as it has conflicts with #4978. All the changes introduced here were synchronized in #4978.

PR #4970, opened 2022-07-27
Fix project relocation of `skin_animated_humans`

### LINKED ISSUE #4965: skin_animated_humans.wbt not working after relocation
The `motions_files` folder in the `skin_animated_humans` project folder is not copied when relocating, so the simulation cannot run after relocating the project.

stefaniapedrazzi:
Fix #4965: rename the motions folder so that it is copied during project relocation.


omichel (approved):
Thank you.

Issue #4952, opened 2022-07-26
EXTERNPROTO declaration is parsed even if added in comment


stefaniapedrazzi:
It seems that even if the EXTERNPROTO declaration is added in a comment, Webots parses and processes it.

Here is a sample world file:
```
#VRML_SIM R2022b utf8
# EXTERNPROTO "../protos/Ball.proto"

EXTERNPROTO "https://raw.githubusercontent.com/cyberbotics/webots/R2022b/projects/objects/floors/protos/RectangleArena.proto"

WorldInfo {
}
Viewpoint {
}
RectangleArena {
}
```

Error in the console:
```
ERROR: '/home/stefi/Desktop/test_project1/worlds/test2.wbt': Failed to load due to invalid token(s).
```

ygoumaz:
I have investigated the problem and it seems that the EXTERNPROTOs in comments are not processed, which is correct. The error comes from the fact that there is a comment directly after the first line `#VRML_SIM R2022b utf8` without any space. I believe a space is required here for the header to be valid and the PROTO to be parsed correctly.

stefaniapedrazzi:
I will check again.
Because Webots was crashing even if I was commenting the EXTERNPROTO causing the issue.

stefaniapedrazzi:
I confirm the issue is just the space.

However, even in this case Webots should throw an error and load the default world instead of the current behavior:
on the VM it freezes, on my development system it seems to partially load something but crashes as soon as I click on the step button.

stefaniapedrazzi:
Additionally there is no reason why a new line is needed between the header and the comment.
Also note that this is valid in VRML.

omichel:
I assume we can now close this issue.

PR #4973, opened 2022-07-27
Fix handling tokenizer errors

stefaniapedrazzi:
As reported in https://github.com/cyberbotics/webots/issues/4952#issuecomment-1196507366, errors in the worlds are not handled correctly.

Here we parse the world without checking if the tokenizer returned errors and the tokenizer is run multiple times for the same world.


omichel (approved):
Thank you.

PR #4974, opened 2022-07-27
Fix impossible relocation of controllers outside of local project in distributions

### LINKED ISSUE #4961: Relocation not possible when saving controllers not located in the project directory
In sample worlds with controllers that are not local to the project, like `apartment.wbt` or `custom_robot_window.wbt`, trying to edit and save these controller triggers the warning message but doesn't offer the possibility to relocate the project. However, relocating is possible when saving the world or when saving controllers of some other worlds (like `create.wbt` or `ratslife.wbt`) because they are local to the project.

**Steps to Reproduce**
1. Open `apartment.wbt` or `custom_robot_window.wbt`.
2. Edit the controllers of the e-puck.
3. Save or compile or clean the controller. The warning pops up but no relocation is possible.
4. Now, open `create.wbt` or `ratslife.wbt` and follow the same procedure. The relocation will be possible.

ygoumaz:
Fixes #4961.

In a sample world, when trying to save a controller related to a PROTO (defined with a remote URL in the world file) belonging to another Webots project folder, the path to the controller contains the remote URL and Webots is therefore unable to locate it. Webots does not propose to relocate the project in this case.


ad-daniel (approved):
Thank you

Issue #4960, opened 2022-07-26
Resource path should be unique


omichel:
Webots R2022b introduces many changes in the way resources (textures, sounds, cad models, etc.) are accessed.
It support the following URL schemes: `webots://`, `https://` and local file (either relative or absolute).
We should ensure that each resource is specified uniquely (the U in URL).
This means that we should not search several paths for a resource, but only one and fail if the file is not found at this place.
Currently Webots apparently still explores a number of folders looking for resources.
This should be fixed.

PR #4978, opened 2022-07-27
Replace blind search for asset locations with context-based search and add handle same PROTO having different urls

ad-daniel:
**Description**

Fixes: https://github.com/cyberbotics/webots/issues/4960

When a relative url is provided for an asset, rather than searching for it in multiple places, it should be retrieved based on the context:
- URLs in a world file are resolved relative to the world file itself
- ~~URLs in the header of a PROTO are resolved relative to the world, if an uninterrupted chain of IS exposes this parameter at the world level, or relative to the closest PROTO if it doesn't.~~
- URLs in the header of a PROTO must be relative to the PROTO that sets the parameter
   - if an uninterrupted chain of "IS" statements is such that the field is visible from the scene tree, then the url must be relative to the world
   - if the visibility chain of the parameter breaks at some point in the nestedness (i.e, the parameter is exposed for a few ancestors, but not all the way through to the world), the url must be relative to the topmost PROTO **_that sets this url internally_**. It may happen that one or multiple PROTO just relay the url, each setting the field from what they themself receive, so the url doesn't change across any of them but the PROTO from which this info is obtained may vary wildly. Reason why it's necessary to chase down which among them all actually sets the field, so that we know relative to which among the many PROTO involved we need to compute the url.
- URLs in the body of a PROTO are resolved relative to the PROTO itself
- If a supervisor sets the url, then by definition the field involved is visible and therefore the url provided to the API must be relative to the world file.

It sounds more complicated than it actually is, in summary the criteria used is: if the field that references this asset is visible from the scene tree then it must be provided relative to the world, if it isn't visible, then it must be relative to the PROTO that sets it.


src/webots/scene_tree/WbAddNodeDialog.cpp
```diff
@@ -713,6 +715,22 @@ void WbAddNodeDialog::accept() {
     return;
   }
 
+  const WbExternProto *cutBuffer = WbProtoManager::instance()->externProtoCutBuffer();
+  const QString protoName =
+    QUrl(mTree->selectedItems().at(0)->text(FILE_NAME)).fileName().replace(".proto", "", Qt::CaseInsensitive);
+  if (cutBuffer && cutBuffer->name() == protoName && !mRetrievalTriggered) {
+    QMessageBox::StandardButton cutBufferWarningDialog =
+      WbMessageBox::warning("A PROTO with the same name as the one you are about to insert is contained in the clipboard. Do "
```
omichel, line 723:
```suggestion
      WbMessageBox::warning("A PROTO node with the same name as the one you are about to insert is contained in the clipboard. Do "
```

src/webots/scene_tree/WbNodeEditor.cpp
```diff
@@ -78,10 +82,20 @@ WbNodeEditor::WbNodeEditor(QWidget *parent) :
...
           &WbSelection::showResizeManipulatorFromSceneTree, Qt::UniqueConnection);
 }
 
+void WbNodeEditor::printUrl() {
+  if (!mNode->isProtoInstance())
+    return;
+
+  WbLog::info(tr("EXTERNPROTO URL of '%1': '%2'")
+                .arg(mNode->modelName())
+                .arg(WbProtoManager::instance()->externProtoDeclaration(mNode->modelName(), true)));
```
omichel, line 96:
```suggestion
  WbLog::info(tr("EXTERNPROTO \"%2\"")
                .arg(WbProtoManager::instance()->externProtoDeclaration(mNode->modelName(), true)));
```

omichel (commented):
It looks good to me. I have only a couple of suggestions.

omichel (approved):
Thank you.

PR #4986, opened 2022-07-28
Fix mime type for models

BenjaminDeleze:
I found the correct mime types here: https://www.iana.org/assignments/media-types/media-types.xhtml


ad-daniel (dismissed):
Thank you

omichel:
Shouldn't we add also `model/stl` as this is also supported by Webots?

ad-daniel (approved):
Thank you

Issue #4955, opened 2022-07-26
Generated PROTO Node file should be read-only in Text Editor


stefaniapedrazzi:
It would be better to also open the generated PROTO node in read-only mode in the Text Editor.

PR #4987, opened 2022-07-29
Fixed generated proto temporary filename

omichel:
Generating a PROTO named `E-puck.Tiviy.proto` seems a bit strange and misleading, instead I believe it is better to generate a file with same name as the original PROTO, but with a different extension, e.g., `.generated_proto`, inside a `generated_protos` folder in the Webots temporary folder.
Also, we don't need a unique path name because the generated PROTO files are in any case temporary and won't affect the behavior of Webots.

This PR also simplifies the code.


BenjaminDeleze (commented):
Seems good, be it think it would be better if the generated protos would be read only in the text editor

omichel:
> Seems good, be I think it would be better if the generated protos would be read only in the text editor.

We already have an issue about it: #4955, but I am not convinced it is more useful or harmful for the users.

BenjaminDeleze (approved):
Thank you

PR #4988, opened 2022-07-29
Fixed generation of relative URLs

omichel:
When modifying the URL to a world resource from the scene tree GUI, the file selector was generating sometimes a relative URL, sometimes an absolute URL. This PR fixes this behavior by generating always a relative URL (relative to the world file).

Moreover, it simplifies the code and removes the search in PROTO search paths which is useless and complicated.


ad-daniel (approved):
Works nicely now. I'll test more extensively when this gets merged into the other PR about urls to ensure there's no clashes between the two. Thank you

Issue #4946, opened 2022-07-26
Webots window blocked and black at start


stefaniapedrazzi:
When lot of `IMPORTABLE EXTERNPROTO` are declared in a file Webots takes very long to start and load a world.
While loading, the Webots window is completely black (if it is open at start), it is blocked and there is no dialog message showing the loading progress with the option to cancel the load that it is very handy for big simulations.

Steps to reproduce:
1. open the `scripts/icon_studio/worlds/icon_studio.wbt`
-> after a while I even gen the message that `Webots` is not responding and I'm asked if I want to "Force Quit" and as I user I would think that Webots is buggy and it freezed

I tested on Ubuntu 22.04.

Loading all the models takes long, but we should at least show and update the loading progress bar and add the option to cancel the load.

stefaniapedrazzi:
This also occurs with other complex worlds but takes less time (for example the `mavic_2_pro.wbt`).
1. Loading window shows text `Parsing
2. Loading window shows text  Downloading assets`
3. Loading window becomes black and unresponsive
4. Loading window shows text  `Finalizing nodes`

omichel:
I believe we should not load the importable externproto models until they are actually imported. That would fix the issue and also be better in particular with a use case where you set a large number of importable externproto but use randomly only a couple of them during a simulation run.

ad-daniel:
What do you intend by "loading"? IMPORTABLE EXTERNPROTO nodes aren't created until they are spawned, they are on the other hand downloaded when parsing the world, which was the intended behavior since we don't want to halt the simulation, wait for things to download, and proceed only then to the next timestep. Moreover, PROTO flagged as "IMPORTABLE EXTERNPROTO" can also be instantiated at load-time (i.e., IMPORTABLE EXTERNPROTO supersedes simple EXTERNPROTO to avoid having 2 separate declarations), as such a-priori we cannot know which among the IMPORTABLE EXTERNPROTO we'll need "only later".

Icon studio has ~300+ PROTO at level 0, which in turn depend in many more sub-proto and each of those depends on tons of textures. Although when run locally there's no "download" per-se taking place (since everything is local), we still need to open each of them, apply several regex to extract the declarations and so on, which I assume is what is taking so long.

Also note there's 2 stages of "Download assets", one is for the PROTO (taking place before the loading of the world), one is for the other assets (textures, sounds, ..), which is done during the loading. In-between the two I assume it's creating the nodes.

omichel:
[...]

PR #4991, opened 2022-08-02
Reduce size of Mavic2Pro PROTO file

stefaniapedrazzi:
As discussed in https://github.com/cyberbotics/webots/issues/4946#issuecomment-1195152165, all the `IndexedFaceSet` of the `Mavic2Pro` model as been moved to extern OBJ files.
The resulting PROTO file has been reduced from 5.6 MB to 23.7 kB.


ad-daniel (approved):
Thank you

PR #4992, opened 2022-08-02
Speedup loading worlds

stefaniapedrazzi:
As reported in #4946, loading complex worlds is very slow.
The bottleneck is currently that the EXTERNPROTO Editor is updated continuously on load for each of the declared EXTERNPROTO.
This is useless and it is sufficient to update the editor only once at the end of the load.

 I also removed a duplicated replace of the webots home path.


stefaniapedrazzi:
Issues with the loading dialog black and unresponsive will be addressed in a separate PR.

ad-daniel (approved):
Nice catch, thank you

PR #4993, opened 2022-08-02
Ameliorate load progress status

ad-daniel:
**Description**
Fixes: https://github.com/cyberbotics/webots/issues/4946

Although the finalization is comprised of multiple heavy steps, in practice the progress bar only tracked the progress of one of them which for some heavy worlds resulted in the dialog to either be black (until the steps with tracked progress were reached) or stuck at 0%.

There is still a small period of time where the progress bar is black before the "parsing nodes" phase, but it doesn't appear to be consistent nor reproducible. This might be related to the regex, but until we find an egregious example that allows to identify where it gets stuck, it's not easy to identify the source. Also, simplifying the PROTO might already do the trick.

### DIFF
--- a/src/webots/nodes/WbGroup.cpp
+++ b/src/webots/nodes/WbGroup.cpp
@@ -67,7 +67,7 @@ void WbGroup::preFinalize() {
   WbBaseNode::preFinalize();
 
   if (isWorldRoot()) {
-    emit worldLoadingStatusHasChanged(tr("Pre-finalize nodes"));
+    emit worldLoadingStatusHasChanged(tr("Pre-finalizing nodes"));
     mLoadProgress = 0;
   }
 
--- a/src/webots/nodes/WbGroup.cpp
+++ b/src/webots/nodes/WbGroup.cpp
@@ -195,7 +195,7 @@ void WbGroup::createWrenObjects() {
   WbBaseNode::createWrenObjects();
 
   if (isWorldRoot())
-    emit worldLoadingStatusHasChanged(tr("Create WREN objects"));
+    emit worldLoadingStatusHasChanged(tr("Creating WREN objects"));
 
   WbMFNode::Iterator it(*mChildren);
   while (it.hasNext()) {
--- a/src/webots/nodes/WbGroup.cpp
+++ b/src/webots/nodes/WbGroup.cpp
@@ -179,7 +179,7 @@ void WbGroup::createOdeObjects() {
   WbBaseNode::createOdeObjects();
 
   if (isWorldRoot())
-    emit worldLoadingStatusHasChanged(tr("Create ODE objects"));
+    emit worldLoadingStatusHasChanged(tr("Creating ODE objects"));
 
   WbMFNode::Iterator it(*mChildren);
   while (it.hasNext()) {
--- a/src/webots/nodes/WbGroup.cpp
+++ b/src/webots/nodes/WbGroup.cpp
@@ -82,6 +82,8 @@ void WbGroup::preFinalize() {
     }
     n->preFinalize();
     emit childFinalizationHasProgressed(mLoadProgress * 100 / (4 * childCount()));
+    if (mFinalizationCanceled)
+      return;
   }
 }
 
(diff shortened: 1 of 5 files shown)


src/webots/nodes/WbGroup.cpp
```diff
@@ -66,21 +66,33 @@ void WbGroup::downloadAssets() {
 void WbGroup::preFinalize() {
   WbBaseNode::preFinalize();
 
+  if (isWorldRoot()) {
+    emit worldLoadingStatusHasChanged(tr("Pre-finalize nodes"));
```
omichel, line 70:
```suggestion
    emit worldLoadingStatusHasChanged(tr("Pre-finalizing nodes"));
```

src/webots/nodes/WbGroup.cpp
```diff
@@ -66,21 +66,33 @@ void WbGroup::downloadAssets() {
...
+    if (mFinalizationCanceled)
+      return;
   }
 }
 
 void WbGroup::postFinalize() {
   WbBaseNode::postFinalize();
 
+  if (isWorldRoot())
+    emit worldLoadingStatusHasChanged(tr("Post-finalize nodes"));
```
omichel, line 94:
```suggestion
    emit worldLoadingStatusHasChanged(tr("Post-finalizing nodes"));
```

src/webots/nodes/WbGroup.cpp
```diff
@@ -162,20 +178,30 @@ int WbGroup::nodeIndex(WbNode *child) const {
 void WbGroup::createOdeObjects() {
   WbBaseNode::createOdeObjects();
 
+  if (isWorldRoot())
+    emit worldLoadingStatusHasChanged(tr("Create ODE objects"));
```
omichel, line 182:
```suggestion
    emit worldLoadingStatusHasChanged(tr("Creating ODE objects"));
```

omichel (changes requested):

src/webots/nodes/WbGroup.cpp
```diff
@@ -162,20 +178,30 @@ int WbGroup::nodeIndex(WbNode *child) const {
...
+    if (mFinalizationCanceled)
+      return;
+  }
 }
 
 void WbGroup::createWrenObjects() {
   WbBaseNode::createWrenObjects();
 
+  if (isWorldRoot())
+    emit worldLoadingStatusHasChanged(tr("Create WREN objects"));
```
omichel, line 198:
```suggestion
    emit worldLoadingStatusHasChanged(tr("Creating WREN objects"));
```

src/webots/nodes/WbGroup.cpp
```diff
@@ -66,21 +66,33 @@ void WbGroup::downloadAssets() {
...
   while (it.hasNext()) {
+    mLoadProgress++;
     WbBaseNode *const n = static_cast<WbBaseNode *>(it.next());
     if (!mHasNoSolidAncestor) {
       WbGroup *group = dynamic_cast<WbGroup *>(n);
       if (group)
         group->mHasNoSolidAncestor = false;
     }
     n->preFinalize();
+    emit childFinalizationHasProgressed(mLoadProgress * 100 / (4 * childCount()));
```
omichel, line 84: Did you consider adding a call to `QCoreApplication::processEvents(QEventLoop::AllEvents);` here to ensure the GUI is actually refreshed at this point?

ad-daniel: I don't believe it's necessary since
`childFinalizationHasProgressed` triggers `WbApplication::setWorldLoadingProgress` and in turn this signal 
[triggers](https://github.com/cyberbotics/webots/blob/2b385b0fb2a3bafc800eb70096199d03bcebefed/src/webots/gui/WbMainWindow.cpp#L145) `WbMainWindow::setWorldLoadingProgress` which processes the events:
https://github.com/cyberbotics/webots/blob/2b385b0fb2a3bafc800eb70096199d03bcebefed/src/webots/gui/WbMainWindow.cpp#L2380-L2385

omichel (approved):
Thank you.

PR #4995, opened 2022-08-03
Better Windows alias for atom

omichel:
Typing "atom" to launch atom from the MSYS2 console is more consistent with Linux and macOS than having to type "edit".


ad-daniel (approved):
Thank you

PR #4999, opened 2022-08-04
Fix turtlebot meshes

omichel:
IndexedFaceSet to meshes.


ad-daniel (approved):
Looks good, thank you.
By the way, I've realized we probably need to re-generate the x3d after these optimizations are done, as I think the x3d used by the docs still reference the geometries using indexedfacesets

PR #5000, opened 2022-08-04
P-Rob3 meshes

omichel:
IndexedFaceSet to meshes


ad-daniel (changes requested):
Looks good, however in both PROTO the meshes have been defined relatively rather than with "webots://". Although this is supported, when it comes time to generating the x3ds a copy of each mesh will be generated for the docs rather than re-use the existing ones

ad-daniel (approved):
As discussed, with a script all the urls will be translated to relative ones, in a separate PR, if they're immediately below (i.e. only if it's something like `./path` but not if it requires stuff like `../../../stuff`)

Issue #4905, opened 2022-07-18
Conversion to `<generic>` controllers and robot windows for remote PROTOs imported locally


ygoumaz:
As discussed in #4898, remote PROTOs that are imported locally when editing them in Webots (right-click on PROTO + Edit PROTO Source + Copy/Import message) are not coming with their respective controllers and robot windows, which means that, after save and reload, they are not found as they were not copied to the local project.

Currently, relocating a project usually copies the controllers inside the new folder. Also, #4898 updated the PROTO wizard to generate `<generic>` fields for windows and controllers from template PROTOs. 

Therefore, it would be more consistent and user-friendly to also update these fields to `<generic>` when importing remote PROTOs to the local project.

PR #5003, opened 2022-08-04
Use relative URLs in PROTO files

omichel:
I ran this script in the `WEBOTS_HOME/projects` folder to perform the conversion automatically:

```python
import glob

for file in glob.glob('**/*.proto', recursive=True):
    unix_file = file.replace('\\', '/')
    search = '"webots://projects/' + unix_file[:unix_file.rfind('/') + 1]
    with open(file, 'r') as f:
        data = f.read()
    new_data = data.replace(search, '"')
    if new_data != data:
        print(unix_file)
        with open(file, 'w', newline='\n') as f:
            f.write(new_data)
```


BenjaminDeleze (commented):
It does not work at least for the following protos:

- Forest.proto
- BunchOfSunFlowers.proto
- Sofa.proto
- Book.proto
- CeilingLight.proto

I think it is because of the new rule introduced in #4978 : 

- If the field that references this asset is visible from the scene tree then it must be provided relative to the world. However in the protos the default texture is defined relative to the proto

omichel:
I am fixing this: reverting to `webots://` URLs in the header part of the PROTO files.

omichel:
I used this script to revert the `webots://...` URLs in the header of PROTO files and fix the problem reported:

```python
import glob
import os

for file in glob.glob('**/*.proto', recursive=True):
    unix_file = file.replace('\\', '/')
    url = '"webots://projects/' + unix_file[:unix_file.rfind('/') + 1]
    fd = open(file, 'r')
    output = open('tmp.proto', 'w', newline='\n')
    for line in fd:
        if 'field MFString' in line and '"webots://' not in line and '[]' not in line and ('.jpg"' in line or '.png"' in line):
            print(unix_file)
            print(line)
            new_line = line.replace('"', url, 1)
            print(new_line)
        else:
            new_line = line
        output.write(new_line)
    fd.close()
    output.close()
    os.remove(file)
    os.rename('tmp.proto', file)
```

stefaniapedrazzi:
When remote PROTO models are edited and copied to the local project, relative URLs are no longer working and
How do handle it?
What is the advantage of using relative URLs?

projects/vehicles/protos/generic/TruckTrailer.proto
```diff
@@ -5,17 +5,17 @@
...
+EXTERNPROTO "TruckWheel.proto"
 
 PROTO TruckTrailer [
   field SFString   name             "truck trailer"
   field SFFloat    length           11.15             # Defines the total length of the trailer.
   field SFFloat    kingPinDistance  2.4               # Defines the distance (along the Z direction) between the front of the trailer and the rotation axis.
   field SFFloat    mass             5000              # Defines the mass of the trailer (wheels not included).
   field MFFloat    wheelAxles       [ 7.93, 9.03, 10.13 ]
-  field SFNode     appearance       PBRAppearance { metalness 0 roughness 0.4 baseColorMap ImageTexture { url [ "webots://projects/vehicles/protos/generic/textures/truck_trailer_webots.jpg" ] } }
+  field SFNode     appearance       PBRAppearance { metalness 0 roughness 0.4 baseColorMap ImageTexture { url [ "textures/truck_trailer_webots.jpg" ] } }
```
BenjaminDeleze, line 18:
```suggestion
  field SFNode     appearance       PBRAppearance { metalness 0 roughness 0.4 baseColorMap ImageTexture { url [ "webots://projects/vehicles/protos/generic/textures/truck_trailer_webots.jpg" ] } }
```

projects/vehicles/protos/generic/TruckTrailerMesh.proto
```diff
@@ -4,7 +4,7 @@
 # tags: hidden
 
 PROTO TruckTrailerMesh [
-  field SFNode    appearance   PBRAppearance { metalness 0 roughness 0.4 baseColorMap ImageTexture { url [ "webots://projects/vehicles/protos/generic/textures/truck_trailer_webots.jpg" ] } }
+  field SFNode    appearance   PBRAppearance { metalness 0 roughness 0.4 baseColorMap ImageTexture { url [ "textures/truck_trailer_webots.jpg" ] } }
```
BenjaminDeleze, line 7:
```suggestion
  field SFNode    appearance   PBRAppearance { metalness 0 roughness 0.4 baseColorMap ImageTexture { url [ "webots://projects/vehicles/protos/generic/textures/truck_trailer_webots.jpg" ] } }

```

projects/vehicles/protos/generic/TruckTrailerSimple.proto
```diff
@@ -3,13 +3,13 @@
...
-EXTERNPROTO "webots://projects/vehicles/protos/generic/TruckRearLight.proto"
-EXTERNPROTO "webots://projects/vehicles/protos/generic/TruckTrailerMesh.proto"
+EXTERNPROTO "TruckWheel.proto"
+EXTERNPROTO "TruckRearLight.proto"
+EXTERNPROTO "TruckTrailerMesh.proto"
 
 PROTO TruckTrailerSimple [
   field SFString   name                 "truck trailer"
-  field SFNode     appearance           PBRAppearance { metalness 0 roughness 0.4 baseColorMap ImageTexture { url [ "webots://projects/vehicles/protos/generic/textures/truck_trailer_webots.jpg" ] } }
+  field SFNode     appearance           PBRAppearance { metalness 0 roughness 0.4 baseColorMap ImageTexture { url [ "textures/truck_trailer_webots.jpg" ] } }
```
BenjaminDeleze, line 12:
```suggestion
  field SFNode     appearance           PBRAppearance { metalness 0 roughness 0.4 baseColorMap ImageTexture { url [ "webots://projects/vehicles/protos/generic/textures/truck_trailer_webots.jpg" ] } }
```

BenjaminDeleze (changes requested):

omichel:
> When remote PROTO models are edited and copied to the local project, relative URLs are no longer working and the assets are not found. How do you handle it?

We should probably download the resources corresponding to the relative URL, in addition of the PROTO...

> What is the advantage of using relative URLs?

It is simpler for the user to set them when creating a PROTO file.

stefaniapedrazzi:
> It is simpler for the user to set them when creating a PROTO file.

Yes, I agree. For custom PROTOs is easier to use relative URLs.
But I think that it is easier for users if the default PROTOs we distribute use remote URLs.
Otherwise users will have lot issues when copying and moving files between projects.
I think that the remote paths are simplifying a lot assets dependencies, so I'm not sure that the changes in this PR are really good for users.

stefaniapedrazzi:
> We should probably download the resources corresponding to the relative URL, in addition of the PROTO...

If we still have to copy the assets to local projects, we loose all the advantages to use remote URLs.

projects/objects/bedroom/protos/Bed.proto
```diff
@@ -13,7 +13,7 @@ PROTO Bed [
   field SFNode     frameAppearance    PaintedWood { colorOverride 0.11 0.11 0.11 }                                                      # Defines the appearance of the frame.
   field SFNode     mattressAppearance PBRAppearance { roughness 1 metalness 0 }                                                         # Defines the appearance of the mattress.
   field SFNode     pillowAppearance   PBRAppearance { roughness 1 metalness 0 }                                                         # Defines the appearance of the pillows.
-  field SFNode     blanketAppearance  PBRAppearance { baseColorMap ImageTexture { url "webots://projects/objects/bedroom/protos/textures/duvet.jpg" } roughness 1 metalness 0 }  # Defines the appearance of the blanket.
+  field SFNode     blanketAppearance  PBRAppearance { baseColorMap ImageTexture { url "textures/duvet.jpg" } roughness 1 metalness 0 }  # Defines the appearance of the blanket.
```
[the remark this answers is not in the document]
omichel: I cannot apply your suggestion... Can you try?

omichel:
>> We should probably download the resources corresponding to the relative URL, in addition of the PROTO...

> If we still have to copy the assets to local projects, we loose all the advantages to use remote URLs.

You are right, downloading the assets is not a good idea. Instead we should convert the relative URLs to `https::` URLs (as we do for the `webots://` URLs.

projects/objects/bedroom/protos/Bed.proto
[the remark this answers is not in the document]
BenjaminDeleze: I cannot, I get `This diff has recently been updated`

projects/vehicles/protos/generic/TruckTrailerMesh.proto
```diff
@@ -5,6 +5,7 @@
 
 PROTO TruckTrailerMesh [
   field SFNode    appearance   PBRAppearance { metalness 0 roughness 0.4 baseColorMap ImageTexture { url [ "webots://projects/vehicles/protos/generic/textures/truck_trailer_webots.jpg" ] } }
+
```
BenjaminDeleze, line 8:
```suggestion
```

BenjaminDeleze (changes requested):
I loaded the whole guided tour and some other worlds and could not find any error.

However the mechanism to convert relative urls to `https` when editing the proto is not implemented and thus editing protos is broken.

omichel:
> I loaded the whole guided tour and some other worlds and could not find any error.
> 
> However the mechanism to convert relative urls to `https` when editing the proto is not implemented and thus editing protos is broken.

I will be taking care of this in a new PR.

BenjaminDeleze:
> I will be taking care of this in a new PR.

Is it a good practice as it will introduce a bug in master?

omichel:
> Is it a good practice as it will introduce a bug in master?

I mean we won't merge this PR until the other PR fixing the issue is merged, so that the master branch won't be broken.

projects/objects/traffic/protos/HighwayPole.proto
```diff
@@ -22,6 +22,7 @@ PROTO HighwayPole [
   field SFFloat       curveRadius           0.4                                                                                                             # Defines the curvature radius of the pole.
   field MFNode        rightHorizontalSigns  [ HighwaySign { name "vertical sign" } ]                                                                        # Defines the horizontal signs on the right side of the pole.
   field MFNode        rightVerticalSigns    [ HighwaySign { name "horizontal sign" height 2.1 length 3.2 texture "webots://projects/objects/traffic/protos/textures/highway_sign_la_ciotat.jpg" } ]  # Defines the vertical signs on the right side of the pole.
+
```
BenjaminDeleze, line 25:
```suggestion
```

BenjaminDeleze:
Is it possible to fix the conflicts such that I can test with the fix of #5051 ?

omichel:
I just fixed the problem.

However, it remains a little problem with the controller path which is not found. I believe one way to fix it would be to compute the controller path to find the controller inside the Webots installation folder. However, this solution is not nice because it would either set the controller as a long relative path that would break if the project is moved or as a an absolute path which is not portable. An alternative to fixing the controller path would be to copy the controller from the Webots installation directory to the local project, but this seems a bit overkill as the user initially just wants to use the PROTO, not necessarily copy its default controller.

So, I see two options:

- change the controller to "<generic>".
- ask the user if she want the default robot controller to be copied in its project or if she prefers the controller to set to "<generic>".
- Do nothing and let the user fix the controller path.

I would tend to prefer the first one as it simplifies the user experience and does what it is supposed to do (e.g., download a PROTO in the local project, not a controller). The default controller is anyhow used when running the provided demo worlds.

omichel:
We also have the same problem for the robot `window` and the `controllerArgs` parameters. I propose to set the `window` to `<generic>` and `controllerArgs` to `[]`.

BenjaminDeleze:
I think that even in the previous version the controller path was broken (so solution 3).

But I would say that solution 1 is nicer.

Solution 2 is nice too because it offers more choice but I agree that I might be overkill. 

So I would go with solution 1 (and put the robot window to generic and so on).

But I would put a message to inform the user that controller/robot window will not be copy. Otherwise I am afraid that they will download a robot, change its color and not understand why the robot is not moving anymore.

ygoumaz:
I believe this would fix issue #4905.

omichel:
I am done with the implementation: the `controller` and `window` fields are set to `<generic>`. However, I didn't change the `controllerArgs` for several reasons:
- It doesn't harm to leave it to the original values, which is in 99% of the cases `[]`. The only exceptions are:
  - `Thymio2.proto: "port=33333" `
  - `InvertedPendulumBenchmark: [ "seed=2" ]`
  - `VisualTrackingBenchmark.proto: [ "hit-error=0.1" ]`
- It is difficult to do it in a generic way and handle multi-line values if any.

BenjaminDeleze (changes requested):
The previous issues seems to have been addressed.

Just two small remarks:

- There is not warning/info that the robot window, controller have been replaced by the generic, do you think it would be worth adding?
- When importing `Darwin-op` and `e-puck` locally and reloading the world, I got the following error: 
```
WARNING: Darwin-op (PROTO): The remote control library has not been found.
```
Because the `remoteControl`field has not been changed. But I don't know if it is worth bothering.

omichel:
I added a message for robots to warn that the controller/window fields will be set to generic.
I believe the remote control problem is not critical. We could possibly add a proto parameter to enable it and leave it disabled by default, so that we don't get this warning.

omichel:
I found a more elegant fix for the E-puck and Darwin-op robots: if the window is set to `<generic>` or `<none>`, the remote control library is not initialized, thus no warning is raised. However, it will raise as soon as the user sets a specific robot window (which should be pretty rare and only for advanced users).

BenjaminDeleze (approved):
Thank you!

I just applied your find to `Darwin-opHinge2`

PR #5004, opened 2022-08-04
Minor improvement to EXTERNPROTO tooltips

omichel:
This PR display the exact `EXTERNPROTO` declaration line in the tooltips.
Moreover, it skips the `tr` as no translation is needed for this.


ad-daniel (approved):
I've taken advantage of the PR to fix the other tooltip as well, thank you

PR #5006, opened 2022-08-04
Hotfix relative PROTO search

ad-daniel:
**Description**

Although downloaded and cached correctly when generating the PROTO tree, when it comes time to instantiate the PROTO node itself, relative-defined nested PROTO weren't found, triggering an assert.

I was under the impression this situation was already part of the test suite, but we only test similar scenarios where a texture within a sub is relative, not the PROTO themselves. In other words:

Situations such as:
`world references remote PROTO` -> `remote PROTO references a relative TEXTURE`
were  tested, but not:
`world references remote PROTO` -> `remote PROTO references a relative PROTO` -> `relative PROTO references a texture`

For these reasons I've added a test.


omichel (approved):
Thank you.

PR #5014, opened 2022-08-05
Fix convert root to basenode missing declaration

ad-daniel:
**Description**

Due to the recent changes, declarations added on the fly by the convert root to basenode mechanism no longer work.
Additionally, ambiguous situations should be notified to the user.


ad-daniel:
Due to the recent change that the extern proto list is purged on every node deletion (rather than just on saves), the declaration of the nodes that will become visible must occur only AFTER the "previous" node (i.e., non converted one) is deleted. Since no instances yet exist of these declarations, deleting the previous node would result in the declarations being purged. I've also fixed this in the PR.

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -790,18 +790,26 @@ void WbProtoManager::exportProto(const QString &path, int category, const QStrin
...
-                                        bool updateContents) {
+                                        bool updateContents, bool isFromRootNodeConversion) {
   for (int i = 0; i < mExternProto.size(); ++i) {
     if (mExternProto[i]->name() == protoName) {
       mExternProto[i]->setImportable(mExternProto[i]->isImportable() || importable);
-      if (mExternProto[i]->url() != protoPath)
-        mExternProto[i]->setUrl(protoPath);
+      if (mExternProto[i]->url() != protoPath) {
+        if (isFromRootNodeConversion) {
+          WbLog::warning(tr("Conflicting declarations for '%1' are provided: %2 and %3, the first will be used. To use the "
```
omichel, line 799:
```suggestion
          WbLog::warning(tr("Conflicting declarations for '%1' are provided: %2 and %3, the first one will be used. To use the "
```

src/webots/vrml/WbProtoManager.hpp
```diff
@@ -34,21 +34,25 @@ class WbVersion;
...
+    mImportable(isImportable),
+    mFromRootNodeConversion(isFromRootNodeConversion) {}
 
   const QString &name() const { return mName; }
   void setUrl(const QString &url) { mUrl = url; }
   const QString &url() const { return mUrl; }
   bool isImportable() const { return mImportable; }
   void setImportable(bool value) { mImportable = value; }
+  bool isFromRootNodeConversion() const { return mFromRootNodeConversion; }
+  void setFromRootNodeConversion(bool value) { mFromRootNodeConversion = value; }
```
omichel, line 49: Since it is always called with a `false` parameter, should probably remove the parameter and rename the method.

omichel (changes requested):

omichel (approved):
Thank you.

PR #5015, opened 2022-08-05
Fix streaming of simulations with relatively defined assets

ad-daniel:
**Description**

The way assets are served to webotsjs has been reworked so as to remove the reliance on an internal map. The changes comprise:
- assets are encoded in the x3d relative to the world
- when streaming, webotsjs knows the path of the world and by combining the two can request webots a url that is relative to the project.
- assets that lay behind the project are not served, so if they want to be provided (either relatively or absolutely), they need to be manually relocated.
- the material for local (webots://) cadshapes sent to webotsjs has now the same prefix of the obj itself.
- since the x3d writer already expresses assets relative to the world, the same logic can be re-used when converting to basenode 

For testing, I've prepared 6 worlds for the different assets (backgrounds/mesh/textures) and different situations, namely: basenode (assets relative to world) o internal to a proto (assets relative to the proto):  ~[urls.zip](https://github.com/cyberbotics/webots/files/9290685/urls.zip)~ more complete tests available [here](https://github.com/ad-daniel/prefab/tree/master/urls)
Note: 
- the black vehicle should not be visible when streaming as it's outside of the scope of the project
- the absolute urls need to be adapted to your computer
- worlds with webots:// can be streamed only if you commit the necessary files


resources/web/wwi/nodes/WbCadShape.js
```diff
@@ -50,18 +67,11 @@ export default class WbCadShape extends WbBaseNode {
 
     super.createWrenObjects();
 
-    if (this.urls.length === 0 || this.scene === 'undefined')
+    if (this.scene === 'undefined')
```
BenjaminDeleze, line 70: Why remove the check on the urls length?

resources/web/wwi/nodes/WbCadShape.js
```diff
@@ -287,38 +297,49 @@ export default class WbCadShape extends WbBaseNode {
       17: ambient occlusion
     */
 
+    let materialPrefix;
+    if (this.isCollada || typeof webots.currentView.stream === 'undefined')
+      materialPrefix = ''; // for animations the texture isn't relative to the material but included in the 'textures' folder
```
BenjaminDeleze, line 302: I think Collada should also have a prefix no?

src/webots/gui/WbX3dStreamingServer.cpp
```diff
@@ -64,10 +65,11 @@ void WbX3dStreamingServer::create(int port) {
 
 void WbX3dStreamingServer::sendTcpRequestReply(const QString &url, const QString &etag, const QString &host,
                                                QTcpSocket *socket) {
-  if (!mX3dWorldTextures.contains(url))
-    WbTcpServer::sendTcpRequestReply(url, etag, host, socket);
+  QFileInfo file(WbProject::current()->dir().absolutePath() + "/" + url);
+  if (file.exists())
+    socket->write(WbHttpReply::forgeFileReply(file.absoluteFilePath(), etag, host, url));
```
BenjaminDeleze, line 70: Did you check that we do not lost performance here?

src/webots/scene_tree/WbSceneTree.cpp
```diff
@@ -754,19 +754,6 @@ void WbSceneTree::convertProtoToBaseNode(bool rootOnly) {
...
-      it.next();
-      const QString destination(WbProject::current()->worldsPath() + it.key());
-      if (!(WbUrl::isLocalUrl(it.key()) || WbUrl::isWeb(it.key()))) {
-        const QFileInfo fileInfo(destination);
-        if (!QDir(fileInfo.absolutePath()).exists())
-          QDir().mkpath(fileInfo.absolutePath());
-        QFile::copy(it.value(), destination);
-      }
-    }
-
```
BenjaminDeleze, line 769: Why is it not needed anymore? (if you want to convert a local proto)

src/webots/nodes/WbMesh.cpp
```diff
@@ -387,9 +387,9 @@ void WbMesh::updateMaterialIndex() {
 }
 
 void WbMesh::exportNodeFields(WbWriter &writer) const {
-  WbBaseNode::exportNodeFields(writer);
+  // WbBaseNode::exportNodeFields(writer);
```
BenjaminDeleze, line 390: debug?

BenjaminDeleze (commented):
Some thoughts I had by reading through the code

resources/web/wwi/Parser.js
```diff
@@ -1211,6 +1211,14 @@ function loadMeshData(prefix, urls) {
   if (typeof urls === 'undefined')
     return;
 
+  let worldsPath;
+  if (typeof webots.currentView.stream === 'undefined')
+    worldsPath = ''
+  else {
+    worldsPath = webots.currentView.stream._view.currentWorld;
```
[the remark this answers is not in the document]
ad-daniel: I assumed so, is it possible to stream without sending a world to webotsjs?

resources/web/wwi/nodes/WbCadShape.js
ad-daniel: sanity checks have been moved to the constructor, but I don't it it hurts to leave it it

resources/web/wwi/nodes/WbCadShape.js
ad-daniel: yep, my bad

src/webots/gui/WbX3dStreamingServer.cpp
ad-daniel: How so? The only difference compared to before is that we check if the file exists rather than checking if the map contains it

src/webots/scene_tree/WbSceneTree.cpp
ad-daniel: Because now the isProto writer leverages the same logic of the x3d writer and expresses the URL relative to the world so, when converting to basenode, locally defined assets are re-written relative to the world therefore there isn't (shouldn't) be a need to create copies of the textures/meshes

resources/web/wwi/Parser.js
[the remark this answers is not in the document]
BenjaminDeleze: I guess that normally not

BenjaminDeleze:
I tried some configurations and it seems to work well. 
I just have two remarks:

- The streaming viewer is not able to load the meshes of `bed.proto`
- Is it a good idea to have assets that are loadable by webots but not by webotsjs (namely local assets that are outside the current project)? It doesn't seem very user friendly to have two different behaviors.

ad-daniel:
> The streaming viewer is not able to load the meshes of bed.proto

Should be fixed now. In the description above, in the repo I mention (under urls/mixed) there are 3 worlds that test different combinations of this additional case (relative proto -> relative asset, web proto -> relative asset, etc. for cadshape, backgrounds, meshes and textures). If you use these worlds for testing note that the behavior of the `webots://` ones varies depending on whether you copy the project inside/outside your webots folder.

> Is it a good idea to have assets that are loadable by webots but not by webotsjs (namely local assets that are outside the current project)? It doesn't seem very user friendly to have two different behaviors.

It's odd I agree, especially since the web interface side lacks an active console where to notify the user (except F12) about why it's missing. However anything behind the current project cannot be requested to webots regardless (whether it's the current or previous approach) since all the `../../` would be consumed before they reach webots. Given the feature choices made with this approach it's just not possible to allow it

I'll wait for the test to finish but in principle the PR can be reviewed

BenjaminDeleze:
> If you use these worlds for testing note that the behavior of the `webots://` ones varies depending on whether you copy the project inside/outside your webots folder.

In practice we should never have world with "webots://" outside of the webots folder right?



> It's odd I agree, especially since the web interface side lacks an active console where to notify the user (except F12)

I will work on a web console (non-interactive) when I have some time, most of the requirement are already done.

ad-daniel:
> In practice we should never have world with "webots://" outside of the webots folder right?

Officially, no. A distribution would not be able to open such a world. If however you are in development mode technically it works anyway by chance, but it isn't really a good way of doing things

> It's odd I agree, especially since the web interface side lacks an active console where to notify the user (except F12)

Awesome!

src/webots/nodes/WbCadShape.cpp
```diff
@@ -575,45 +575,49 @@ const WbVector3 WbCadShape::absoluteScale() const {
...
     else {
-      const QString meshPath(WbUrl::computePath(this, "url", mUrl, i));
-      if (writer.isWritingToFile()) {
-        QString newUrl = WbUrl::exportMesh(this, mUrl, i, writer);
-        dynamic_cast<WbMFString *>(urlFieldCopy.value())->setItem(i, newUrl);
-      }
-
-      const QString &url(mUrl->item(i));
-      writer.addResourceToList(url, meshPath);
+      QString meshPath;
```
omichel, line 593: Useless?
```suggestion
```

src/webots/nodes/WbCadShape.cpp
```diff
@@ -575,45 +575,49 @@ const WbVector3 WbCadShape::absoluteScale() const {
...
+      if (writer.isWritingToFile())
+        dynamic_cast<WbMFString *>(urlFieldCopy.value())->setItem(i, WbUrl::exportMesh(this, mUrl, i, writer));
+      else
+        dynamic_cast<WbMFString *>(urlFieldCopy.value())->setItem(i, WbUrl::expressRelativeToWorld(completeUrl));
```
omichel, line 597: Simpler:
```suggestion
      dynamic_cast<WbMFString *>(urlFieldCopy.value())->setItem(i, writer.isWritingToFile() ? WbUrl::exportMesh(this, mUrl, i, writer) : WbUrl::expressRelativeToWorld(completeUrl));
```

src/webots/nodes/WbCadShape.cpp
```diff
@@ -575,45 +575,49 @@ const WbVector3 WbCadShape::absoluteScale() const {
...
+        dynamic_cast<WbMFString *>(urlFieldCopy.value())->addItem(WbUrl::computeLocalAssetUrl(materialUrl));
+      else if (WbUrl::isWeb(materialUrl))
+        dynamic_cast<WbMFString *>(urlFieldCopy.value())->addItem(materialUrl);
+      else {
+        if (writer.isWritingToFile())
+          dynamic_cast<WbMFString *>(urlFieldCopy.value())
+            ->addItem(WbUrl::exportResource(this, materialUrl, materialUrl, writer.relativeMeshesPath(), writer, false));
+        else
+          dynamic_cast<WbMFString *>(urlFieldCopy.value())->addItem(WbUrl::expressRelativeToWorld(materialUrl));
+      }
```
omichel, line 616: More readable and less error prone:
```suggestion
      WbMFString *urlField = dynamic_cast<WbMFString *>(urlFieldCopy.value());
      if (WbUrl::isLocalUrl(materialUrl))
        urlField->addItem(WbUrl::computeLocalAssetUrl(materialUrl));
      else if (WbUrl::isWeb(materialUrl))
        urlField->addItem(materialUrl);
      else {
        if (writer.isWritingToFile())
          urlField->addItem(WbUrl::exportResource(this, materialUrl, materialUrl, writer.relativeMeshesPath(), writer, false));
        else
          urlField->addItem(WbUrl::expressRelativeToWorld(materialUrl));
      }
```

src/webots/nodes/WbMesh.cpp
```diff
@@ -387,31 +387,27 @@ void WbMesh::updateMaterialIndex() {
...
-        const QString newUrl = WbUrl::exportMesh(this, mUrl, i, writer);
-        dynamic_cast<WbMFString *>(urlFieldCopy.value())->setItem(i, newUrl);
-      }
-
-      const QString &url(mUrl->item(i));
-      writer.addResourceToList(url, meshPath);
+      if (writer.isWritingToFile())
+        dynamic_cast<WbMFString *>(urlFieldCopy.value())->setItem(i, WbUrl::exportMesh(this, mUrl, i, writer));
+      else
+        dynamic_cast<WbMFString *>(urlFieldCopy.value())->setItem(i, WbUrl::expressRelativeToWorld(completeUrl));
```
omichel, line 407: We could also here put the dynamic_cast inside a variable to avoid repeating it four times.

BenjaminDeleze:
It works well!

Just one remark:
When exporting `webots_urls_proto/world/webots_proto.wbt`  and `webots_urls_basenode/worlds/webots_basenode.wbt` to a local scene, a texture folder is created with the textures of the `forklift`. However I think it should not be the case as the textures are taken from a `webots://` cadShape.

src/webots/nodes/WbCadShape.cpp
ad-daniel: Normally I'd agree, but when long expressions are involved I feel like an if-else is much easier to digest than a ternary operator:
a

omichel: We should favor the readability and avoid copy/paste of code as much as possible. In this case version 2 seems more readable.

ad-daniel:
> When exporting webots_urls_proto/world/webots_proto.wbt and webots_urls_basenode/worlds/webots_basenode.wbt to a local scene, a texture folder is created with the textures of the forklift. However I think it should not be the case as the textures are taken from a webots:// cadShape.

That's indeed the case, I forgot to remove the export also from the shallow node, the textures were exported but not actually being used

resources/web/wwi/image_loader.js
```diff
@@ -36,8 +36,17 @@ export function loadTextureData(prefix, url, isHdr, rotation) {
       webots.currentView.branch = 'released';
     url = url.replace('webots://', 'https://raw.githubusercontent.com/' + webots.currentView.repository + '/webots/' + webots.currentView.branch + '/');
   }
-  if (typeof prefix !== 'undefined' && !url.startsWith('http'))
-    url = prefix + url;
+  if (typeof prefix !== 'undefined' && !url.startsWith('http')) {
+    if (['smaa_area_texture.png', 'smaa_search_texture.png', 'gtao_noise_texture.png'].includes(url) || typeof webots.currentView.stream === 'undefined' )
```
BenjaminDeleze, line 40: This line is too long (max 128)

resources/web/wwi/image_loader.js
```diff
@@ -36,8 +36,17 @@ export function loadTextureData(prefix, url, isHdr, rotation) {
...
+      url = prefix + url;
+    else {
+      // for simulations the asset is provided relative to the world, hence it requires resolving the url prior to requesting it to Webots
```
BenjaminDeleze, line 43: This line is too long (max 128)

resources/web/wwi/Parser.js
```diff
@@ -1211,6 +1211,14 @@ function loadMeshData(prefix, urls) {
   if (typeof urls === 'undefined')
     return;
 
+  let worldsPath;
+  if (typeof webots.currentView.stream === 'undefined')
+    worldsPath = ''
```
BenjaminDeleze, line 1216:
```suggestion
    worldsPath = '';
```

BenjaminDeleze (changes requested):
Thank you! 
No more remark on functionality, just some styling remarks:

BenjaminDeleze (approved):
Thank you!

PR #5016, opened 2022-08-05
Robotino meshes

omichel:


BenjaminDeleze (approved):
Thank you

We will need to regenerate the `.x3d` but we can do that once when #5016 is merged and all the meshes extraced (and the script corrected)

PR #5017, opened 2022-08-06
ros controller argument "--robot-description" bug fix

BruceXSK:
`max` causes the parameter `description-prefix` to no work, it should be `min` I suppose

### DIFF
--- a/projects/default/controllers/ros/Ros.cpp
+++ b/projects/default/controllers/ros/Ros.cpp
@@ -152,7 +152,7 @@ void Ros::launchRos(int argc, char **argv) {
       mAutoPublish = true;
     else if (std::string(argv[i]).rfind("--robot-description") == 0) {
       const std::string argument = std::string(argv[i]);
-      const size_t valueStart = std::min(argument.find("="), argument.find(" "));
+      const size_t valueStart = argument.find("=");
       mRobotDescriptionPrefix = (valueStart == std::string::npos) ? "" : argument.substr(valueStart + 1);
       mSetRobotDescription = true;
     } else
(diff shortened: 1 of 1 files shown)


projects/default/controllers/ros/Ros.cpp
```diff
@@ -152,7 +152,7 @@ void Ros::launchRos(int argc, char **argv) {
       mAutoPublish = true;
     else if (std::string(argv[i]).rfind("--robot-description") == 0) {
       const std::string argument = std::string(argv[i]);
-      const size_t valueStart = std::max(argument.find("="), argument.find(" "));
+      const size_t valueStart = std::min(argument.find("="), argument.find(" "));
       mRobotDescriptionPrefix = (valueStart == std::string::npos) ? "" : argument.substr(valueStart + 1);
```
omichel, line 156: I don't believe this patch is correct. If the searched string is missing, the find method will return -1. So, if we use min instead of max, it will return -1 in case either " " or "=" is not found which is obviously not what we want.

Also, I am not sure that we need to search for a space. Normally, the robot description should be separated from the `--robot-description` string with a `=` sign. So, I would suggest to simplify the code instead:

```C++
    else if (strncmp(argv[i], "--robot-description=", 20) == 0) {
      mRobotDescriptionPrefix = std::string(&argv[i][20]);
      mSetRobotDescription = true;
    } else
```

Would this work in your case?

BruceXSK: > 

`rfind` method will return a `size_t` type value, thus the -1 should be $2^{32}-1$ or $2^{64}-1$, `min` is appropriate.

And yeah, I also think this searching for a space is weird. But since I consider it was a slip of the pen, I simply changed the `max` to `min`. 
Your sample code can extract the parameter correctly, but it cannot handle the case where the `=` sign is omitted, unless you add another `if` block, but that's a little bit complicated.

To simplify, I think we can just write this line as:
```cpp
const size_t valueStart = argument.find("=");
```
If the `=` sign is omitted, `valueStart` is `std::string::npos`

omichel: Yes, I agree with you. That looks good to me. Can you please go ahead with the latest version you propose?

omichel (approved):
Thank you!

PR #5023, opened 2022-08-08
Open generated template PROTO sources in read-only mode

stefaniapedrazzi:
Addresses #4955:
I think that opening the generated PROTO sources as read-only is cleaner for the user that can easier understand that any changes to this file won't be taken into account by Webots.
If they want to modify the file, they can use the "Save as" option.
Note that the action is also named "Show" and not "Edit", so it seems more coherent to not allow to modify this file.

Other improvements:
* I moved the definition of the assets cached path to the `WbStandardPaths` namespaces with all the other path declarations
* fixed generated PROTO sources path containing double path separator
* fixed cppcheck errors:
  * `WbTokenizer` class: arguments passed by copy instead of reference
  * `WbNetwork` class:
    * some functions that can be defined as static and return reference to string
    * fix assert causing different behavior in debug and release mode


omichel (approved):
That looks all good to me.
Thank you.

PR #5025, opened 2022-08-08
Remove PROTO nodes copy in test suite

stefaniapedrazzi:
It is no longer needed to copy the PROTO nodes from the `test/default` folder.


BenjaminDeleze (approved):
Thank you

PR #5027, opened 2022-08-08
Cleanup not used function arguments

stefaniapedrazzi:
Ditto.


ygoumaz (changes requested):
I think `WbAddNodeDialog::selectionType()` can also be removed as it no longer seems to be useful.

ygoumaz (approved):
Thank you!

PR #5029, opened 2022-08-08
Fixed Qt plugins path on Windows

omichel:
The previous code was relying on `cygpath` to determine if MSYS2 was installed. This failed on @ygoumaz's machine because he had Git installed in `C:\Program Files\Git` which comes with its own `cygpath` and provides a wrong path for MSYS2. This PR removes the dependency on `cygpath` and simply tests for the existence of the Qt plugins folder in the Webots installation folder. If not found, it assumes we are in the development environment and reverts to the Qt plugins folder of MSYS2.

Also, I removed the override of `QT_QPA_PLATFORM_PLUGIN_PATH` which doesn't work. This variable is taken into account only when set before starting Webots, and not if set from within Webots.


ygoumaz (approved):
Thank you. It now works nicely in the development environment and in the distribution!

PR #5033, opened 2022-08-09
Fix EXTERNPROTO backward compatibility script on Windows

ygoumaz:
On Windows, the URLs of local PROTOs are written with backslashes instead of forward slashes.


ad-daniel (approved):
Thank you

PR #5037, opened 2022-08-09
Tiago arms to meshes

omichel:
This one also uses the `Mesh.ccw FALSE` and negative scales to render the right arm.


BenjaminDeleze (approved):
Seems good to me

PR #5039, opened 2022-08-09
Bed meshes

omichel:
I had to separate the texture of the Bed and the one of the PhotoFrame in different folders for a better organization.


BenjaminDeleze (approved):
Thank you

PR #5040, opened 2022-08-09
Fixed Fanatec steering wheel mesh

omichel:
This PR also re-organize the textures of the vehicle/protos/abstract folder which were all mixed-up.
It also removes rim_texture.jpg which was not used by any proto.
Note: this the last PROTO above 1MB.


BenjaminDeleze (approved):
Seems good to me

Issue #5024, opened 2022-08-08
Scaling issue in simulation UI R2022b


TheMangalex:
**Describe the Bug**
I hope this isn't a duplication, but I didn't find a fix or something related. It might also be an issue with 4k displays.
Since the update to R2022b, the simulation window fills only a quarter of the given screen space. 

The mouse movements and clicks are measured at the whole screen though, making it additionally hard to click on objects.

**Steps to Reproduce**
1. Installing webots from source origin/master (make -j8)
2. Launching with ./webots

**Expected behavior**
Normal screen scaling as seen in 2nd Screenshot

**Screenshots**
R2022b
![webots r2022b](https://user-images.githubusercontent.com/26520790/183404487-50facc06-d5f4-4c3f-9cad-6c4115794dab.png)

R2022a
![webots](https://user-images.githubusercontent.com/26520790/183403796-62094790-b2ca-4978-a705-d321c1c03988.png)

**System**
 - Operating System: Ubuntu 20.04
 - Graphics Card: Intel HD Graphics & Rtx 3070 (only tested on intel currently though)
 - Display Settings: 4k (3840x2160)
 - Window-Manager: i3

**Additional context**

omichel:
That's is strange... Can you try to to set `export QT_USE_PHYSICAL_DPI=1` and run Webots again?
Alternatively, can you also try `export QT_SCALE_FACTOR=2`?
Also can you try to remove `QT_ENABLE_HIGHDPI_SCALING` from https://github.com/cyberbotics/webots/blob/master/src/webots/launcher/webots-linux.sh#L74?

Let me know your findings.

TheMangalex:
Okay so the 1st and 3rd option didn't change anything, but the 2nd option looked like that
![option2](https://user-images.githubusercontent.com/26520790/183430693-f8580ede-b352-4325-abfd-d06ac9fd9345.png)
.
Also it seems like this bug doesn't happen in Gnome for me and only i3, but i3 was working correctly with R2022a. Maybe some variable is incorrectly provided by the X-Server?

omichel:
Maybe [this one](https://github.com/cyberbotics/webots/commit/88e8b517af05661aa55d7e7e2209ba12e04bb894) broke it? Can you try to revert it?

TheMangalex:
Sadly this doesn't seem to change it.

TheMangalex:
Sadly nothing as well.

omichel:
[...]

PR #5043, opened 2022-08-10
Fix high DPI scaling bug on i3 window manager (Linux)

omichel:
This adjusts the Linux launch script to deal with a scaling bug in the i3 window manager and Qt6 on high DPI monitors.
The high DPI handling of Qt6 is disabled and the UI is manually scaled.

Scaling issue in simulation UI R2022b https://github.com/cyberbotics/webots/issues/5024

- [x] A check with non 4k monitor or non high dpi monitor would be interesting

### DIFF
--- a/src/webots/launcher/webots-linux.sh
+++ b/src/webots/launcher/webots-linux.sh
@@ -71,13 +71,16 @@ export WEBOTS_TMPDIR=$WEBOTS_TMPDIR
 # add the "lib" directory into LD_LIBRARY_PATH as the first entry
 export LD_LIBRARY_PATH="$webots_home/lib/webots":$LD_LIBRARY_PATH
 
-#Fix for i3 not working with qt6
+# Fix for i3 window manager not working with Qt6
 if [ "$XDG_CURRENT_DESKTOP" == "i3" ]; then
-export QT_ENABLE_HIGHDPI_SCALING=0
-export QT_SCALE_FACTOR=2
-export QT_FONT_DPI=80
+  DPI=`xrdb -query -all | grep Xft.dpi | awk '{print $2}'`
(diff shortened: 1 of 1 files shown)


omichel:
I just tested on a non-4K screen and it doesn't look good as the fonts and scaling of the UI (icons, etc.) are too big.

omichel:
I had to adapt the script to test whether we are on a high DPI display so that it doesn't break the GUI on low DPI displays (96 DPI).

BenjaminDeleze (approved):
Seems good to me

omichel:
@TheMangalex: can you confirm that this still works and looks good for you?

TheMangalex:
Sadly not, the script goes completely correct by getting the Xft.dpi value (180 for me, because else the font on this laptop is unreadable), but the section below isn't executed.

I think the problem is in line 77. `if [[ "$DPI" > 96 ]]; then`
I am by no means a bash expert, but it seems like > can't be correctly used in [[ ]] and must be replaced with -gt or double round brackets must be used.
So either
`if [[ "$DPI" -gt 96 ]]; then`
or
`if (( "$DPI" > 96 )); then`
seem to do the trick for me and launch with the correct exports.

omichel:
You are totally right! I am fixing this right now.

omichel:
> (180 for me, because else the font on this laptop is unreadable)

That looks odd, wouldn't you get a better performance with a value of 192 instead (as the scale factor should be an int number, e.g., 2)? On my machine, the windows manager explicitly recommends setting it to a multiple of the original DPI, which is 96, to get a better performance.

TheMangalex:
That's actually a good point, I should set it 192. Thanks for the tip.

PR #5044, opened 2022-08-10
Fix web component studio and regenerate robots

BenjaminDeleze:
Fix web component studio to avoid duplicate even so the assets are declared as local in protos.

Waiting for #5003


omichel (commented):
Since what we are doing for meshes and for textures is the same, wouldn't it make sense to write a function for this and call it twice: with the 'meshes' and 'textures' arguments?

scripts/web_component_studio/web_component_studio.py
```diff
@@ -51,6 +52,23 @@ def run_webots():
...
+                                        component['name'] + '.x3d'), '"' + type + '/',
+                           '"' + remote_address + '/' + type + '/')
+    elif os.path.isdir(resources_path2):
+        search_and_replace(os.path.join(WEBOTS_HOME, 'docs', 'guide', 'scenes', component['name'],
+                                        component['name'] + '.x3d'), '"' + type + '/',
+                           '"' + remote_address + '/' + component['proto'] + '/' + type + '/')
+
+    resources_path = os.path.join(WEBOTS_HOME, 'docs', 'guide', 'scenes', component['name'], type)
+    if os.path.isdir(resources_path):
+        shutil.rmtree(resources_path)
```
omichel, line 71: Don't we need to remove also this folder if it exists?
```suggestion
    resources_path = os.path.join(WEBOTS_HOME, 'docs', 'guide', 'scenes', type)
    if os.path.isdir(resources_path):
        shutil.rmtree(resources_path)
```

BenjaminDeleze: This folder does not exist because the textures/meshes folder is always at the same level that the `.x3d` file as it is created like a scene/animation.

BenjaminDeleze:
It seems that with the modifications introduced in #5015 the assets are correctly exported and so the modification of `web-component-studio.py` where not needed anymore.

So this PR only regenerate the scenes

omichel (approved):
Thank you.

PR #5046, opened 2022-08-10
Style webotsJS

BenjaminDeleze:
ditto


omichel (approved):
Thank you.
Did you use a linter to address the styling in this PR?

BenjaminDeleze:
Yes I have the `linter-eslint` in atom, following: https://github.com/cyberbotics/webots/wiki/Javascript-Coding-Style/#cs406-use-eslint-linter-atom-package-linter-eslint.

It works quite well except for:
- global functions/variables (like wren ones), it says that they are unknown
- if there is an `if` that have only one instruction (no need for brackets) but that instruction is on two lines because it is too long, it will think that brackets are needed

PR #5048, opened 2022-08-10
Improved the output of the Makefile

omichel:
To make the output of the main Makefile more readable.


ygoumaz (approved):
Seems better this way indeed!

Issue #4981, opened 2022-07-28
Circular dependency in webots modules


omichel:
As discussed in https://github.com/cyberbotics/webots/pull/4956#discussion_r929773004, #4551 introduced a [circular dependency](https://github.com/cyberbotics/webots/blob/b1c3e4980c73f1a40e06424141e2c56e5884b98e/src/webots/Makefile#L36) that should be fixed as soon as possible. The more we wait, the harder it will be to remove it.

PR #5049, opened 2022-08-10
Exclusive IMPORTABLE list check for supervisor imported nodes

### LINKED ISSUE #5038: Wrong PROTO imported when two PROTOs have the same name
Following tests in #4967, using two PROTOs with the same name but different URLs (one remote and one local), importing one of them with a supervisor by following a specific order of steps doesn't produce the correct result.

The steps to perform are the following.

1. Download the following project: [super.zip](https://github.com/cyberbotics/webots/files/9291536/super.zip). It contains a local PROTO called `Ball` which depends on `LocalBall`. The key point here is that the `Ball` PROTO is also an official one which is available remotely.
2. Open Webots and the supervisor.wbt world.
3. Click on `Add Node` and add the local `Ball`.
4. Click on `IMPORTABLE EXTERNPROTO` -> `Insert new`. Only the local `Ball` is available in the list. Insert it.
5. Save the simulation.
6. Now run the simulation. The correct `Ball` is imported after a few seconds (actually it is a cube visually to differentiate it from the official `Ball`).
7. Reset the simulation.
8. Delete both `Ball`s in the IMPORTABLE list and in the scene tree. Don't save.
9. Click on Insert new. It is now possible to insert the official `Ball`. Do so.
10. Run the simulation. The official `Ball` should be imported, but this is not the case. The local one is imported instead.
11. Now reset the simulation and save the simulation.
12. Run the simulation. The correct `Ball` (official one) is now imported.

ygoumaz:
Fixes #5038.

Confusions between EXTERNPROTOs contained in the clipboard, in the world file or in the IMPORTABLE list brought situations where wrong nodes where imported from a supervisor, as exposed in #5038. 

To avoid these issues, when looking for the URL of a node imported from a supervisor, we now only check the IMPORTABLE list.

Additionally, this PR adds a warning if a node with the same name as the one contained in the clipboard is inserted into the IMPORTABLE EXTERNPROTO list.


src/webots/vrml/WbProtoManager.cpp
```diff
@@ -23,6 +23,7 @@
 #include "WbMultipleValue.hpp"
 #include "WbNetwork.hpp"
 #include "WbNode.hpp"
+#include "WbNodeOperations.hpp"
```
stefaniapedrazzi, line 26: Note that here you are adding an additional circular dependency that we will need to remove (see #4981 ).

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -117,20 +119,26 @@ WbProtoModel *WbProtoManager::findModel(const QString &modelName, const QString
...
+  }
 
   // for IMPORTABLE proto nodes the declaration is in the EXTERNPROTO list, nodes added with add-node follow a different pipe
   if (protoDeclaration.isEmpty()) {
     foreach (const WbExternProto *proto, mExternProto) {
       if (proto->name() == modelName && (proto->isImportable() || proto->isFromRootNodeConversion()))
         protoDeclaration = proto->url();
     }
+    if (mImportedFromSupervisor && protoDeclaration.isEmpty())
+      assert(false);  // for supervisor imported nodes, there should always be a corresponding PROTO in the IMPORTABLE list
```
omichel, line 141: It is better to encapsulate this inside the assert (to ensure it won't affect the release version):
```suggestion
    // for supervisor imported nodes, there should always be a corresponding PROTO in the IMPORTABLE list
    assert(!mImportedFromSupervisor || !protoDeclaration.isEmpty());
```

src/webots/nodes/utils/WbNodeOperations.hpp
```diff
@@ -73,6 +73,7 @@ public slots:
 signals:
   void nodeAdded(WbNode *node);
   void nodeDeleted(WbNode *node);
+  void fromSupervisorChanged(const bool value);
```
omichel, line 76: This doesn't read very smoothly, I would rename it to:
```suggestion
  void changedFromSupervisor(const bool value);
```

omichel (changes requested):
That works nicely and seems well implemented.

omichel (approved):
Thank you.

src/webots/nodes/utils/WbNodeOperations.cpp
```diff
@@ -83,6 +83,8 @@ void WbNodeOperations::cleanup() {
 }
 
 WbNodeOperations::WbNodeOperations() : mNodesAreAboutToBeInserted(false), mSkipUpdates(false), mFromSupervisor(false) {
+  connect(this, &WbNodeOperations::changedFromSupervisor, WbProtoManager::instance(),
+          &WbProtoManager::setImportedFromSupervisor);
```
stefaniapedrazzi, line 87: Do you really need a signal/slot mechanism?
Does not work if you call directly `WbProtoManager::setImportedFromSupervisor`?

ygoumaz: It would indeed work, but is it not clearer this way? Both global variables should have the same value, so a signal/slot mechanism seems the most appropriate to reflect this property.

omichel: I agree with @stefaniapedrazzi. Signal/slot should be used only when we cannot call a method directly. Using signal/slot usually complicates the code and the debugging, so we should avoid it wherever possible.

stefaniapedrazzi: Signal/slot make the codes much more complicated and are heavier to run.
If possible we should call the functions directly and avoid adding extra layers.

I don't see exactly why it should be cleaner to emit a signal instead of calling a function, to me they look exactly the same.
What would make the code cleaner would be to define a function that sets both `WbNodeOperations::mFromSupervisor` and calls `WbProtoManager::setImportedFromSupervisor` to ensure that they always have the same value.

ygoumaz: Yes, we just discussed with @omichel and we agreed on using a function that would set both variables at the same time.

PR #5051, opened 2022-08-10
Fixed local and webots URLs for proto editing

omichel:
Fixes the replacement of local URLs while editing remote protos, as spotted in #5003.

In addition, this PR also fixes a bug in the `webots://` replacement as `WbUrl::remoteWebotsAssetPrefix()` was used instead of the actual `repository/branch-tag-or-hash` URL.


BenjaminDeleze:
It tested a bit for `Edit Proto node` and it seems to work.

However, there is still issues if you do:

- `Add node`
- Click on `Spot`
- `Export`
- Import local `Spot`

omichel:
Radical solution: I removed the "Export" button from the Add Node dialog, as it was redundant with the "Edit PROTO" pop-up menu.

BenjaminDeleze (changes requested):
I got some errors when trying to import a PROTO with the Supervisor but I am not sure it is related with this PR, I will continue testing tomorrow

BenjaminDeleze (approved):
Seems good to me

PR #5055, opened 2022-08-10
Fix condition on high DPI detection in Linux launcher

omichel:
Follow-up on #5043.


TheMangalex (approved):
Seems good and worked for me this way.

BenjaminDeleze (approved):
Thank you

PR #5032, opened 2022-08-09
Add declaration to session after modifying a default PROTO parameter

stefaniapedrazzi:
Fixes #5026:
* [x] when modifying a PROTO parameter the PROTO declaration is added to the world EXTERNPROTO declarations
* [x] when displaying the PROTO URL (Add Node dialog and Node editor), use the node URL directly if the node is not found in the world declarations
* [x] remove declaration when resetting PROTO to default value

Tests:
* [x] open world with local PROTO not declared and select it
* [x] open world with a not declared PROTO node as default PROTO node parameter (not written in WBT file), and select it
* [x] modify default parameter, save world -> PROTO is added to the WBT file & no warnings are printed
* [x] reset default parameter, save world -> PROTO is removed from the WBT file & no warnings are printed
* [x] use different URL in world and PROTO file for default parameter:
  * [x] in case of default parameter  -> PROTO declaration should be used (after save and revert)
  * [x] in case of modified parameter  -> world declaration should be used (after save and revert)
* [ ] ...

PR #5059, opened 2022-08-11
Fix assertion crash when importing some PROTOs with supervisor

ygoumaz:
#5049 introduced a bug when importing PROTOs that contain other nested PROTOs with a supervisor (e.g. RectangleArena), causing an assertion to trigger. The variable `mImportedFromSupervisor` should only be true for level 1 of the PROTO.

The PR also removes the signal/slot mechanism which is not the lightest solution to synchronize the `mImportedFromSupervisor` variable between `WbNodeOperations` and `WbProtoManager`.

### DIFF
--- a/src/webots/nodes/utils/WbNodeOperations.hpp
+++ b/src/webots/nodes/utils/WbNodeOperations.hpp
@@ -62,8 +62,6 @@ public:
   bool isSkipUpdates() { return mSkipUpdates; }
   bool isFromSupervisor() { return mFromSupervisor; }
 
-  void setFromSupervisor(const bool value);
-
   static QString exportNodeToString(WbNode *node);
 
   void enableSolidNameClashCheckOnNodeRegeneration(bool enabled) const;
@@ -84,6 +82,8 @@ private:
   bool mNodesAreAboutToBeInserted;
   bool mSkipUpdates;
   bool mFromSupervisor;
+
+  void setFromSupervisor(bool value);
 
 private slots:
   void resolveSolidNameClashIfNeeded(WbNode *node) const;
--- a/src/webots/nodes/utils/WbNodeOperations.hpp
+++ b/src/webots/nodes/utils/WbNodeOperations.hpp
@@ -62,6 +62,8 @@ public:
   bool isSkipUpdates() { return mSkipUpdates; }
   bool isFromSupervisor() { return mFromSupervisor; }
 
+  void setFromSupervisor(const bool value);
+
   static QString exportNodeToString(WbNode *node);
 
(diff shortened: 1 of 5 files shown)


src/webots/vrml/WbProtoManager.cpp
```diff
@@ -137,6 +137,8 @@ WbProtoModel *WbProtoManager::findModel(const QString &modelName, const QString
       if (proto->name() == modelName && (proto->isImportable() || proto->isFromRootNodeConversion()))
         protoDeclaration = proto->url();
     }
+
+    mImportedFromSupervisor = false;
     // for supervisor imported nodes, there should always be a corresponding PROTO in the IMPORTABLE list
     assert(!mImportedFromSupervisor || !protoDeclaration.isEmpty());
```
omichel, line 143: This assert is obviously always true because `mImportedFromSupervisor` is always false.

ygoumaz: Indeed, I had converted the PR back to draft but you were too fast.

src/webots/nodes/utils/WbNodeOperations.hpp
```diff
@@ -62,6 +62,8 @@ class WbNodeOperations : public QObject {
   bool isSkipUpdates() { return mSkipUpdates; }
   bool isFromSupervisor() { return mFromSupervisor; }
 
+  void setFromSupervisor(const bool value);
```
omichel, line 65: It should be private as it is only called from within the `WbNodeOperations` class.

omichel (changes requested):

src/webots/nodes/utils/WbNodeOperations.hpp
stefaniapedrazzi: Also note that it is useless to use  `const`  when passing the argument by value because the original value is copied and won't be modified.

```suggestion
  void setFromSupervisor(bool value);
```

Same for WbProtoManager::setImportedFromSupervisor`.

BenjaminDeleze (approved):
Thank you

omichel (approved):
Thank you.

stefaniapedrazzi:
Is the assertion failure fixed?
On the master branch I'm still getting the assertion failure: for example when running the test suite.

The last test that is successfully executed is `OK: supervisor_get_velocity_with_joints`

ygoumaz:
> Is the assertion failure fixed? On the master branch I'm still getting this assertion failure: for example when running the test suite.

Yes, the assertion failure in the test suite was fixed in #5073. I confirm that #5032 fixes the second assertion failure.

Issue #4764, opened 2022-06-27
Remove import vrml from Supervisor controller


stefaniapedrazzi:
In R2022b we just removed the functionality to import VRML files from the GUI.
But it is still possible to import a ".wrl" file using the `wb_supervisor_field_import_mf_node`.

This functionality is not documented, but for consistency it might be better to remove it as well from the Supervisor controller and to promote the use of different import methods.

PR #5061, opened 2022-08-11
Remove WBO

### LINKED ISSUE #5056: Remove WBO
The WBO file format was implemented in Webots a long time ago, when there was probably no PROTO support in Webots.
Now, it is obsolete in the sense that everything that could be done with WBO can be done with PROTOs in a much cleaner way.
So, it is not good to keep this WBO feature in Webots because:
1. It makes the code of Webots more complex: it requires that we test it, maintain it and document it regularly.
2. It is confusing, useless and misleading for the users.

So, I would suggest to remove it.

omichel:
Fixes #5056.

- [x] remove the wbo import from the Webots GUI.
- [x] remove `wb_supervisor_import_s/mf_node` API functions.
- [x] clean-up unused functions.
- [x] add a change log item.
- [x] update the ROS repository: https://github.com/cyberbotics/webots_ros/pull/126
- [x] ~update the ROS2 repository.~ Not needed.
- [x] change tutorial 8 to use PROTOs instead of WBO files: https://cyberbotics.com/doc/guide/tutorial-8-the-supervisor?version=fix-remove-wbo


stefaniapedrazzi:
If you will also remove the WBO support from the Supervisor API, then this PR could also address  #4764.

omichel:
> If you will also remove the WBO support from the Supervisor API, then this PR could also address #4764.

Yes, I plan to remove the `wb_supervisor_field_import_sf_node` and `wb_supervisor_field_import_mf_node` API functions as they are anyhow broken if using remote extern controllers.

src/webots/scene_tree/WbAddNodeDialog.hpp
```diff
@@ -81,7 +76,6 @@ private slots:
   QList<WbNode *> mDefNodes;
   int mDefNodeIndex;
   bool mHasRobotTopNode;
-  ActionType mActionType;
   QString mImportFileName;
```
ygoumaz, line 79: I believe `mImportFileName` can be removed too.

src/webots/vrml/WbParser.hpp
```diff
@@ -63,7 +63,7 @@ class WbParser {
   WbTokenizer *mTokenizer;
   int mMode;
 
-  enum { NONE, WBT, VRML, PROTO, WBO, WRL };
+  enum { NONE, WBT, VRML, PROTO, WRL };
```
ygoumaz, line 66: Is there a reason to keep WRL here, as it was removed at multiple other locations in the code?

ygoumaz (changes requested):
I found remaining references to .wbo (and .wrl) files in the code. The respective files are:

- WbConsole.cpp
- WbNodeOperations.hpp
- WbTokenizer.cpp
- WbTokenizer.hpp
- WbNodeReader.hpp
- WbParser.hpp
- WbLanguage.cpp

In most of the cases they are just comments that can be either removed or edited.

src/webots/vrml/WbParser.hpp
omichel: They were used to parse the [resources/nodes/*.wrl](https://github.com/cyberbotics/webots/tree/master/resources/nodes) files. However, I found a more elegant way to handle this. See 610b178d899f314fc274d9c2282cb7e25e74a9a4

ygoumaz (commented):
Overall, everything seems fine to me. Is there a reason to keep the nodes in `resources/nodes` in .wrl format? Could we convert them to PROTO files and remove the last occurrences of wrl files in `WbNodeModel` and `WbParser`?

omichel:
> Overall, everything seems fine to me. Is there a reason to keep the nodes in `resources/nodes` in .wrl format? Could we convert them to PROTO files and remove the last occurrences of wrl files in `WbNodeModel` and `WbParser`?

Yes, because these are not PROTO but basic nodes, so I fear that calling them PROTO would be confusing, thus the .wrl file extension. Also, they do not follow the PROTO syntax as they don't contain a body.

ygoumaz (changes requested):
Thanks for the clarification.

I believe that `WbNodeUtilities::isSingletonTypeName()` is no longer called in the code. While the function may be useful in the future, I don't think it's good practice to keep unused methods.

ygoumaz (approved):
Thank you !

Issue #5058, opened 2022-08-11
Fail to load extern Proto from another github repo


BenjaminDeleze:
Webots crash with the following world:

```
#VRML_SIM R2022b utf8

EXTERNPROTO "https://raw.githubusercontent.com/cyberbotics/webots-cloud-simulation-examples/test-protos-download/7_local_assets/protos/MyProto.proto"

WorldInfo {
}
Viewpoint {
  orientation -0.0316302248353415 0.02843866445440257 0.9990949760863156 1.6779495280745207
  position -0.05753688709726134 2.5600691798403346 9.999992923742619
}
Background {
  skyColor 0.5 0.5 0.8
}

MyProto {
}

```

`MyProto` is a valid Proto and works well if you open this world: https://github.com/cyberbotics/webots-cloud-simulation-examples/blob/main/7_local_assets/worlds/local_assets_example.wbt

ad-daniel:
No crashes for me in development mode
![a](https://user-images.githubusercontent.com/44834743/184083315-a082846c-5b2c-4280-a93d-823b11edaece.png)

BenjaminDeleze:
I think it is a problem with CadShape, can you try with `EXTERNPROTO "https://raw.githubusercontent.com/cyberbotics/webots-cloud-simulation-examples/main/7_local_assets/protos/MyProto.proto"` instead?

If it is the case, #5015 will maybe fix it

ad-daniel:
Yes it crashes, the issue is in WbUrl, it triggers an assert when trying to resolve the obj relative to the PROTO

ad-daniel:
I have an idea why it would fail, the ImageTextures used by CadShape are shallow nodes... We might need to unshallow them a bit and define a parent regardless

BenjaminDeleze:
It seems that it has been fixed by #5015, I can now load remote CadShape correctly

PR #5062, opened 2022-08-11
Set parenthood in shallow nodes

ad-daniel:
**Description**

Fixes: https://github.com/cyberbotics/webots/issues/5058

CadShape nodes are essentially an assembly of multiple geometries and appearances. The geometries are static meshes but what makes CadShapes a bit special is that the list of appearances are initialized from an assimp material rather than the usual pipeline. As such, the resulting PBRApperances/ImageTextures are not fully fledged nodes but rather shallow ones, they have no surface-level existence and can't be interacted with, nor are they saved to file. The recent change to the way relative urls are resolved means that they need to at least have a parent defined so that we can climb back to the CadShape node, or `WbUrl::computePath` won't be able to resolve it.


stefaniapedrazzi:
Did you already try to set the WbNode::gParent variable before instantiating the appearance and image texture node instead of passing the parent node as an argument?

https://github.com/cyberbotics/webots/blob/master/src/webots/vrml/WbNode.hpp#L79

src/webots/vrml/WbNode.cpp
```diff
@@ -172,12 +172,12 @@ void WbNode::init() {
 // special constructor for shallow nodes, it's used by CadShape to instantiate PBRAppearances from an assimp material in
 // order to configure the WREN materials. Shallow nodes are invisible but persistent, and due to their incompleteness should not
 // be modified or interacted with in any other way other than through the creation and destruction of CadShape nodes
-WbNode::WbNode(const QString &modelName, const aiMaterial *material) {
+WbNode::WbNode(const WbNode *parentNode, const QString &modelName, const aiMaterial *material) {
```
stefaniapedrazzi, line 175: This is not new, but the `const aiMaterial *material` argument is not used and should be removed.

ad-daniel:
Both aspects can indeed be simplified, thank you for the suggestion

stefaniapedrazzi (approved):
Thank you!

PR #5065, opened 2022-08-11
Grey out share button when not available

### LINKED ISSUE #5063: Grey out share button when streaming
**Describe the Bug**

When streaming the share button is disabled (rightly so), but it would be better if it's also gets greyed out (as the save button does when there's nothing to save)

BenjaminDeleze:
Fix #5063


omichel (approved):
Thank you.

PR #5069, opened 2022-08-11
Fix PROTO wizard wrong or missing declarations

ygoumaz:
When using the PROTO wizard to create a PROTO from a certain base node, if this base node contains other nested PROTOs, the wizard is using the category (world, project, webots, etc.) to check for the URLs of the nested nodes. The following example gives a wrong output for the declarations:

1. Open Webots and add a Bed node to the world using the `Add Node` button.
2. Create a new PROTO with the PROTO wizard based on the Bed node.
  2.1. If the Webots Bed is used, the output PROTO file is correct.
  2.2. If the Bed of the world is used, the output PROTO file is wrong: the PaintedWood PROTO is declared with an empty URL, because it isn't contained in the World category.

This PR improves the accuracy of the declarations of the created PROTO file by checking directly in the base node file. For remote distributed PROTOs, the prefix of nested nodes must be rewritten accordingly.


src/webots/gui/WbNewProtoWizard.cpp
```diff
@@ -141,9 +142,19 @@ void WbNewProtoWizard::accept() {
             tokenizer.tokenizeString(parameters[i]);
             WbParser parser(&tokenizer);
             foreach (const QString &node, parser.protoNodeList()) {
-              QString url = WbProtoManager::instance()->protoUrl(node, mCategory);
+              const QString parentUrl = url.replace("webots://", WbStandardPaths::webotsHomePath());
+              QString nestedUrl = WbProtoManager::instance()->findExternProtoDeclarationInFile(parentUrl, node);
```
ad-daniel, line 146:
```suggestion
              QString nestedUrl = WbProtoManager::instance()->findExternProtoDeclarationInFile(WbUrl::resolveUrl(url), node);
```

src/webots/gui/WbNewProtoWizard.cpp
```diff
@@ -141,9 +142,19 @@ void WbNewProtoWizard::accept() {
...
+              // Replace local URL of nested nodes in distributed remote parent nodes
+              const QString prefix = WbUrl::computePrefix(parentUrl);
+              if (!prefix.isEmpty() && (WbUrl::isLocalUrl(nestedUrl) || QDir::isRelativePath(nestedUrl))) {
```
ad-daniel, line 149: I haven't tested to see if it's actually an issue in practice, but I just want to point out that if `nestedUrl` happens to be remote (`http://...`), then the `QDir::isRelativePath` activates as well and I don't believe in this case we wish to execute the body of the function. Or rather, if it's already a web-url, we want to use it without any manipulation. In short, I think a condition `WbUrl::isWeb(nestedUrl)` prior to doing anything else might be missing

ad-daniel (changes requested):
Thank you, just 2 comments

ad-daniel (approved):
Works nicely, thank you

PR #5070, opened 2022-08-12
Fix incomplete proto wizard generation

ygoumaz:
PROTOs created with the PROTO wizard using a base node contained in the world or locally to the project are incompletely written. In particular SFNodes only write their names but the content of their edited fields is not copied.

**Example**:
Creating MyProto.proto from Bed.proto (contained in the world) generates the following SFNode fields:
```
PROTO MyProto [
  field SFNode frameAppearance PaintedWood {} 
  field SFNode mattressAppearance PBRAppearance {} 
  field SFNode pillowAppearance PBRAppearance {} 
  field SFNode blanketAppearance PBRAppearance {}
]
```
instead of 
```
PROTO MyProto [
  field SFNode frameAppearance PaintedWood { colorOverride 0.11 0.11 0.11 } 
  field SFNode mattressAppearance PBRAppearance { roughness 1 metalness 0 } 
  field SFNode pillowAppearance PBRAppearance { roughness 1 metalness 0 } 
  field SFNode blanketAppearance PBRAppearance { baseColorMap ImageTexture { url "webots://projects/objects/bedroom/protos/Bed/textures/duvet.jpg" } roughness 1 metalness 0 } 
```

In addition, PROTOs that are detected as official are now better handled in the development environment, as well as in the distributions.


src/webots/vrml/WbProtoManager.cpp
```diff
@@ -735,10 +741,17 @@ WbProtoInfo *WbProtoManager::generateInfoFromProtoFile(const QString &protoFileN
...
-    if (defaultValue->type() == WB_SF_NODE && vrmlDefaultValue != "NULL")
-      vrmlDefaultValue += "{}";
+    QString vrmlDefaultValue;
+
+    if (defaultValue->type() == WB_SF_NODE) {
+      const WbSFNode *sfn = dynamic_cast<const WbSFNode *>(defaultValue);
+      if (sfn->value()) {
+        QString nodeContent = WbNodeOperations::exportNodeToString(sfn->value());
+        nodeContent.replace(QRegularExpression("[\\s\\n]+"), " ");
+        vrmlDefaultValue = nodeContent;
```
ad-daniel, line 751:
```suggestion
        vrmlDefaultValue = nodeContent.replace(QRegularExpression("[\\s\\n]+"), " ");
```

ad-daniel: Can be simplified

src/webots/vrml/WbProtoManager.cpp
```diff
@@ -557,9 +559,13 @@ void WbProtoManager::generateProtoInfoMap(int category, bool regenerate) {
       }
 
       WbProtoInfo *info;
-      const bool isWebotsProto = isProtoInCategory(protoName, PROTO_WEBOTS) &&
-                                 (WbUrl::resolveUrl(protoUrl(protoName, PROTO_WEBOTS)) == WbUrl::resolveUrl(protoPath));
-      if (isCachedProto && isWebotsProto)
+      const bool isWebotsProto =
+        isProtoInCategory(protoName, PROTO_WEBOTS) &&
+        (WbUrl::resolveUrl(protoUrl(protoName, PROTO_WEBOTS)) ==
+         WbUrl::resolveUrl(isCachedProto ? WbNetwork::instance()->getUrlFromEphemeralCache(protoPath) : protoPath));
```
ad-daniel, line 565: since we go through the trouble of requesting the original URL beforehand to extract the PROTO name, I think we can store the URL in a variable at the same time to avoid calling `getUrlFromEphemeralCache` twice. Doing so also removes the need to use the ternary operator here, since by the point this line is reached, it's already either the correct path, or the reversed one

ad-daniel (changes requested):
Works well, just 2 suggestions

ad-daniel (approved):
Thank you

Issue #5031, opened 2022-08-09
Local resources of a derived PROTO cannot be found


omichel:
As reported in https://github.com/cyberbotics/webots/pull/5030#issuecomment-1209155107, the TiagoIron proto is derived from the TiagoBase proto but Webots is unable to load the local resources of the TiagoBase.

Issue #5028, opened 2022-08-08
Nested protos: If the field that references an asset is visible from the scene tree then it must be provided relative to the world.


BenjaminDeleze:
**Steps to Reproduce**
1. Open or create a world.
2. Add a `TruckSimple` proto.
3. Change the url of the baseColorMap ImageTexture of the TruckTrailer to `textures/truck_trailer_webots.jpg`. 
4. See that it works.

**Expected behavior**
Webots should not been able to find the texture as `textures/truck_trailer_webots.jpg` is relative to the `TruckTrailer` proto and not to the world

ad-daniel:
Unfortunately it's the worst case scenario, although the `url` field of the `ImageTexture` node is clearly visible from the scene tree, when querying for the visibility `WbNodeUtilities::isVisible()` returns false, hence why it searches it relative to the PROTO rather than the world. The issue is likely due to `WbNodeUtilities::isVisible()` itself, perhaps it gets stuck somewhere due to there being a `Slot` node involved.

BenjaminDeleze:
I remarked another problem: if I modify something in the slot, for example the color of the `TruckTrailerSimple`, it does not trigger the proto regeneration.
I have to save and reload for the changes to be applied

BenjaminDeleze:
This latter problem was already existing in R2022a

BenjaminDeleze:
Concerning the first issue, it seems that the `slot` is not the problem, the structure of the truck is:
- `TruckSimple.proto`: normal proto (robot)
  - `TruckTrailerSimple.proto`: normal proto (slot)
    - `TruckTrailerMesh.proto`: hidden proto (shape)

The appearance parameter of the `TruckTrailerSimple.proto` is not used in this proto but in the `TruckTrailerMesh.proto` which is hidden.

So when trying to resolve the path to the texture, we conclude that the texture is not visible in the sceneTree as `TruckTrailerMesh` is not visible, so that the texture must be relative to the proto

If I replace `TruckTrailerMesh` by:
```
Shape {
  geometry Box{}
  appearance IS appearance
}

```

Then I get the correct behavior: the texture is visible in the SceneTree and thus must be defined relatively to the world

BenjaminDeleze:
Here is a summary to help me keep track with this issue.

I have three simple PROTOS:

TestAncestor:
```
#VRML_SIM R2022b utf8
# license: Copyright Cyberbotics Ltd. Licensed for use only with Webots.
[...]

PR #5071, opened 2022-08-12
Fix search of relative assets in derived nodes

ad-daniel:
**Description**
Fixes: https://github.com/cyberbotics/webots/issues/5031

Assets relative to a PROTO are determined based on the visibility of the field in question. For nested PROTO, following the aliases allows to determine the scope of a field and therefore relative to which PROTO it must be resolved.
This method however falls flat when derived PROTO are concerned because in the process of building these PROTO nodes, all intermediary nodes are instantiated only temporarily, so notions of visibility can't be relied upon as no contiguous chain of aliases exists anymore.

With this PR, a map of "IS" connections (between the exposed parameter and the internal field) is kept at the PROTO model level so that it's possible to follow this chain by jumping through consecutive ancestors in the derived PROTO chain.


BenjaminDeleze:
With a bit of luck it will also fix #5028 !

ad-daniel:
> With a bit of luck it will also fix #5028 !

eheh, no such luck unfortunately, this issue/fix is just related to derived PROTO, it doesn't touch the visibility condition

src/webots/vrml/WbProtoModel.hpp
```diff
@@ -145,9 +147,11 @@ class WbProtoModel : public QObject {
   QStringList mTags;
   QString mTemplateLanguage;
 
+  QMap<QString, QString> mParameterAliases;  // tracks the connections between an exposed parameter and its internal counterpart
+
   ~WbProtoModel();  // called from unref()
-  void verifyAliasing(WbNode *root, WbTokenizer *tokenizer) const;
-  void verifyNodeAliasing(WbNode *node, WbFieldModel *param, WbTokenizer *tokenizer, bool searchInParameters, bool &ok) const;
+  void verifyAliasing(WbNode *root, WbTokenizer *tokenizer);
+  void verifyNodeAliasing(WbNode *node, WbFieldModel *param, WbTokenizer *tokenizer, bool searchInParameters, bool &ok);
```
omichel, line 154: I would rename these functions to `setupAliasing` and `setupNodeAliasing` as the `verify` prefix is a bit misleading (they also construct the parameter alias map).

omichel (changes requested):

omichel (approved):
Thank you.

PR #5073, opened 2022-08-15
Remove IMPORTABLE assertion in findModel

ygoumaz:
The test suite is broken because of an assertion that is triggered when importing invalid nodes with a supervisor. After some attempts to fix it, I realized that it is actually possible for the `findModel` function (in `WbProtoMananger`) to be called by a supervisor importation even if the IMPORTABLE list is empty. The assertion is therefore not correct and can be removed.


omichel (approved):
Well spotted!

PR #5075, opened 2022-08-15
Move Wizard menu in File/New menu

### LINKED ISSUE #5072: URL relative to an unsaved world
**Describe the Bug**

Based on the visibility and/or while streaming, some URL may be expressed or searched for relative to the world file. This however presupposes that the world itself has a known location. If a user creates a world, adds stuff but never actually saves, the behavior is incorrect.

omichel:
Fixes  #5072.

This PR moves the Wizard menu into the File / New menu as it is a more natural place to create new projects, worlds, controllers, plugins, protos, etc. It also fixes #5072 by enforcing the naming and saving of the world file when creating a new world.


src/webots/gui/WbMainWindow.cpp
```diff
@@ -1419,27 +1414,23 @@ void WbMainWindow::saveWorld() {
...
   }
 
-  QString worldFilename = world->fileName();
-  QString previousWorldFileName = worldFilename;
-  if (!WbProjectRelocationDialog::validateLocation(this, worldFilename)) {
-    simulationState->resumeSimulation();
-    return;
-  }
+  QString worldFileName = world->fileName();
+  QString previousWorldFileName = worldFileName;
```
stefaniapedrazzi, line 1418: It seems that this variable is no longer used

stefaniapedrazzi (changes requested):
Everything works fine.
But the behavior of the **New World File..** menu item is a little bit weird because it opens a save file dialog from which you can choose any folder, but then you get an error message saying that you can just save the file in the current "worlds" folder.
Why did you add the save file dialog if the user can only choose the file name?

omichel:
To be more consistent with other wizards, I remove the file save dialog and replace it by a check on the existence of the chosen file. At the end of the wizard, the exact location of the file is displayed to the user, so I guess this OK.

omichel:
I now trying to factorize the creation of the new world to avoid code duplication between the project wizard and the world wizard.

stefaniapedrazzi (commented):
I saw that in the wizard classes many dialogs doesn't use the translation function.
I'm now reviewing the latest changes.

src/webots/gui/WbNewWorldWizard.cpp
```diff
@@ -138,37 +137,37 @@ bool WbNewWorldWizard::validateCurrentPage() {
...
-
 QWizardPage *WbNewWorldWizard::createWorldPage() {
   QWizardPage *page = new QWizardPage(this);
   page->setTitle(tr("World settings"));
-  page->setSubTitle(tr("Please select the features you want:"));
+  page->setSubTitle(tr("Please choose a name for the new world and select the features you want:"));
+  mWorldEdit = new WbLineEdit(page);
   mBackgroundCheckBox = new QCheckBox(page);
+  mBackgroundCheckBox->setChecked(true);
+  mBackgroundCheckBox->setText("Add a textured background");
```
stefaniapedrazzi, line 159: Don't we need to use `tr()` here as well (and for the other check box labels)?

stefaniapedrazzi (commented):
Everything works fine!

I just have a one last question about the translations.

src/webots/gui/WbNewWorldWizard.cpp
omichel: Yes, I will take care of fixing this.

stefaniapedrazzi (approved):
Thank you!

PR #5087, opened 2022-08-19
Merge worlds warning and worlds update tests in single test

### LINKED ISSUE #4989: `world update` test should check for warnings too
**Describe the Bug**

Since every night this test opens every world anyway, we might as well check for warnings at the same time rather than doing so only before the releases.

Essentially we should merge https://github.com/cyberbotics/webots/blob/master/tests/manual_tests/test_worlds_warnings.py into https://github.com/cyberbotics/webots/blob/master/scripts/packaging/update_all_worlds.py

Between the two I assume `update_all_worlds` detects when the world is actually loaded better than `test_worlds_warnings.py` does (where we just set an arbitrary 20sec timer), so it might be an improvement as well.

ygoumaz:
Fixes #4989.

This PR adds a new `test worlds` test, with its corresponding `test worlds` label. In addition to the old `world-update` test which was saving worlds, this new one also checks for warnings.

Previous scripts, `scripts/packaging/update_all_worlds.py` and `tests/manual_tests/test_worlds_warnings.py`, are removed since `test_worlds.py` is now more complete and does both tasks in equal time.

In addtion, the PR also fixes warnings about missing textures in `visual_tracking.wbt`.


ygoumaz:
The new `worlds test` fails because there are indeed warnings in some worlds, but they are not related to this PR. They should be fixed in other PRs.

tests/test_worlds.py
```diff
@@ -37,15 +36,17 @@ def setUp(self):
             'ComposedShader is experimental.',
```
BenjaminDeleze, line 36:
```suggestion
```

The ComposedShader does not exist anymore

tests/test_worlds.py
```diff
@@ -37,15 +36,17 @@ def setUp(self):
...
             self.crashError  # To remove once #6125 is fixed
         ]
         # Get all the worlds from projects
         self.worlds = []
-        for directory in ['projects']:
-            for rootPath, dirNames, fileNames in os.walk(os.environ['WEBOTS_HOME'] + os.sep + directory):
-                for fileName in fnmatch.filter(fileNames, '*.wbt'):
-                    world = os.path.join(rootPath, fileName)
-                    self.worlds.append(world)
+        root = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
```
BenjaminDeleze, line 45: I am not sure about that.
In the case of tests in the CI, I think that what we do is:
- Download the artifact that contain the built distrib of webots
- Set the $WEBOTS_HOME to point to the distrib
- Run the tests

However as the distrib does not contain the tests, we run the tests from the cloned github repo.
So the $WEBOTS_HOME does not correspond to `../..`

BenjaminDeleze (changes requested):
Seems good, only two remarks

tests/test_worlds.py
ad-daniel: Haven't checked if it's right or wrong, just wanted to point out that contrary to the test_suite, the test_world targets the worlds of the distrib (since they are distributed) as opposed to targetting the worlds from the cloned repo (which we need to do for the test_suite since they aren't included)

ygoumaz: Until https://github.com/cyberbotics/webots/pull/5087/commits/d1562a65877419af4fda5caa5df94a97d3218a94, I was looking for worlds in the `$WEBOTS_HOME/projects` directory. But this was leading to only 294 worlds to be tested. Then I switched to this method, which is the one used for the current `world-update` test (297 worlds). But if we aim at testing only the distributed worlds (which seem to be 294), it would be more correct to use WEBOTS_HOME to find worlds to test.

BenjaminDeleze (approved):
Thank you

PR #5088, opened 2022-08-19
Rework relative URL resolver

ad-daniel:
**Description**

TODO.


BenjaminDeleze:
I think it could be related to this PR (if not, do not hesitate to open a separate issue) but the `MercedesBenzSprinter` proto is completely broken whereas the `MercedesBenzSprinterSimple` is just broken as the other cars.

BenjaminDeleze:
It seems that the texture of the `JetBotRaspberryPiCamera` in the Jetbot was also broken.

PR #5091, opened 2022-08-19
Fix new project wizard wrong world location

### LINKED ISSUE #5089: New project wizard creates the world in the wrong place
**Describe the Bug**

1. Open `spot.wbt`
2. Create a new project with `file > new > new project directory`
3. Set location elsewhere (ex: in documents) and name the world `abc.wbt`
4. Summary says `abc.wbt` will be created in the desired location
5. When clicking finish, the desired location is empty and instead `abc.wbt` is saved in the same location of the `spot.wbt` (also ignoring the fact that this location is read-only)

ygoumaz:
Fixes #5089.

The new project was never set as current WbProject.

### DIFF
--- a/src/webots/gui/WbNewProjectWizard.cpp
+++ b/src/webots/gui/WbNewProjectWizard.cpp
@@ -70,6 +70,7 @@ void WbNewProjectWizard::accept() {
     QDialog::accept();
     return;
   }
+  WbProject::setCurrent(mProject);
   createWorldFile();
   // store the accepted project directory in the preferences
   QDir dir(mProject->path());


ad-daniel (approved):
Works well, thank you

src/webots/gui/WbMainWindow.cpp
```diff
@@ -1956,8 +1956,11 @@ void WbMainWindow::newWorld() {
   WbNewWorldWizard wizard(this);
   wizard.exec();
   simulationState->resumeSimulation();
-  if (!wizard.fileName().isEmpty())
-    loadWorld(WbProject::current()->worldsPath() + wizard.fileName());
+  if (!wizard.fileName().isEmpty()) {
+    QFile newWorldFile(wizard.fileName());
+    if (newWorldFile.exists())
+      loadWorld(WbProject::current()->worldsPath() + wizard.fileName());
+  }
```
stefaniapedrazzi, line 1963: There is no need to define a variable here that is used only once, additionally QFile` class has a static method to check the existence of a file without having to create an instance of `QFile`.
Moreover the nested IF blocks should be merged:
```
if (!wizard.fileName().isEmpty() && QFile::exists(wizard.fileName())
    loadWorld(WbProject::current()->worldsPath() + wizard.fileName());
```

Issue #3979, opened 2021-12-07
Nested PROTO translation/rotation propagation not functional in groups


ad-daniel:
**Describe the Bug**

For a fairly simple nested PROTO, like a `StraightRoadSegment` inside a `Group`, changing the translation/field has no effect until some other field is changed. The loading of non-default values of translation works fine, is just the changes from GUI that are "temporarely" broken.

```
#VRML_SIM R2021b utf8
# template language: javascript

PROTO Test [
  field   SFVec3f              translation               0 0 0
  field   SFRotation           rotation                  0 0 1 0
  field   SFFloat              width                     7                      
]{
  Group {
    children [
      StraightRoadSegment {
        translation IS translation
        rotation IS rotation
        width IS width
      }
    ]
  }
}
```

To reproduce:
   - load the world in [bug_translation_propagation.zip](https://github.com/cyberbotics/webots/files/7666520/bug_translation_propagation.zip)
   - change translation/rotation field, nothing happens.
   - change width, the road widens
   - change translation/rotation field, now it propagates.

EDIT:
it appears to affect `StraightRoadSegment` but not `Road`

stefaniapedrazzi:
Fixed in  #5779.

PR #5092, opened 2022-08-19
Fix nested proto visibility

### LINKED ISSUE #5028: Nested protos: If the field that references an asset is visible from the scene tree then it must be provided relative to the world.

BenjaminDeleze:
**Description**
Fix #5028 

The problem of urls path (`protos/` instead of `worlds/`) and the fact the `TruckTrailerSimple` was never updated if we tried to change its color/texture was due to the fact that the exposed field was not correctly redirected in the case of nested Proto.

The best solution I found is to modify this line: https://github.com/cyberbotics/webots/blob/2307b3f68c99df6faf18fe6ec210e76eb875a94c/src/webots/vrml/WbNode.cpp#L238)

In:
```
      if (gNestedProtoFlag && (!parameterNodeField->alias().isEmpty() || !parameterNodeField->isParameter())) {
```

It fixes the issue, pass the test suite and I detected no side effect.

However, if I just do: 
```
      if (gNestedProtoFlag && !parameterNodeField->alias().isEmpty()) {
```
I get exactly the same result, even if I find it less conceptually correct (put useless redirection on internal nodes)

So I am a bit hesitating on the way to follow.
@stefaniapedrazzi, maybe you have an idea?

Other points that worry me a bit is that issue: https://github.com/cyberbotics/webots/issues/3979 or the fact that the `tiago++` -> `endEffectorRightSlot` -> `TiagoGripper` -> `translation` field does not work until we changed the name.

Both problem are not new (already present in R2022a and even before I think) but they seem really close to the one fixed by this PR so I am a bit afraid that this PR fixes only part of the problem

### DIFF
--- a/src/webots/vrml/WbNode.cpp
+++ b/src/webots/vrml/WbNode.cpp
@@ -235,7 +235,7 @@ WbNode::WbNode(const WbNode &other) :
     // copy fields
     foreach (WbField *parameterNodeField, other.mFields) {
       WbField *field = NULL;
-      if (gNestedProtoFlag) {
+      if (gNestedProtoFlag && (!parameterNodeField->alias().isEmpty() || !parameterNodeField->isParameter())) {
         // create an instance of a nested PROTO parameter node
         // don't redirect PROTO instance fields to PROTO node fields
         // PROTO instance fields will be redirected to PROTO node parameters in function cloneAndReferenceProtoInstance()


BenjaminDeleze:
I just discovered a side effect :(

ygoumaz (changes requested):
I think I found an additional case where the behavior is not correct. Here are the steps to reproduce it.

1. Open a world.
2. Add a PROTO that contains an empty slot (the E-puck for example).
3. Add the TruckSimple PROTO to the slot (turretSlot for the E-puck).
4. Now change the URL in turretSlot > TruckSimple > TruckTrailerSimple > PBRAppearance > ImageTexture > url to anything that should be leading to a missing texture (remove a character for example).
5. Three problems emerge:
  5.1. The first one is that you can put any URL but the PROTO is not automatically regenerated so Webots doesn't show that the url is wrong and the texture is missing.
  5.2. Now if you change the URL to relative format (`textures/truck_trailer_webots.jpg`), save the simulation and reload, the truck will be regenerated but the texture will still be found, although it shouldn't because it is not present at the world level, only at the PROTO level.
 5.3. After the reload, a warning is shown in the console because there is a missing declaration in the world file (the url is non-default anymore).

I think 5.3 is probably related to #5096. 5.1 and 5.2 don't seem to be related to other open issues, but I'm not sure if it's related to what you're trying to fix.

stefaniapedrazzi:
Nodes creation and parameter redirection seems to work fine and for the moment I couldn't find any side effect.
However, when changing a field of the `TruckTrailerSimple`'s `PBRAppearance` no `parameterChanged` signal is emitted as it is the case when changing other parameters field like the ones in the `TruckSimple` PROTO.

There are a couple of functionalities that are based on the `parameterChanged` signal:
* template regeneration: I tested and it seems to still work fine because the `parameterChanged` signal is emitted by the parent `appearance` field if it's a template regenerator field
* update of EXTERNPROTO declaration: doesn't work because there is no notification that the parameter changed

ad-daniel:
Just a note, since the entire logic of the search of relative URLs is gonna change, this PR should only focus on ensuring whether the URL is flagged as visible (with a print) when it actually is. If that alone works as intended and there are no side-effects, then this PR can be merged. Whether the asset can be found or not will be a separate issue which as it stands won't rely on the visibility check directly anymore. A good place to do this check is indeed where we currently do it, namely in WbUrl::computePath

BenjaminDeleze:
@ygoumaz:
I think I fixed 5.1 and 5.2
And yes 5.3 is related to #5096 

@stefaniapedrazzi:
I do not think that this is linked with redirection.
I agree that `parameterChanged` should be emitted but I believe it is because we only connect the signal to the parameter fields of the PROTO, and when fields of subnodes are updated we do not look if they have a proto parameter ancestor

BenjaminDeleze:
So now this PR should also fix #5096

stefaniapedrazzi (approved):
Thank you

ygoumaz (approved):
I tested a few more cases and everything seems to work nicely. Thank you!

PR #5097, opened 2022-08-22
Fix executing color edit dialog when running in fast mode

### LINKED ISSUE #5090: Color editor does not open in fast mode
**Describe the Bug**
When trying to modify the color with the color editor (instead of changing the value, click on the colored square), the editor does not open if we are in fast mode. However, the Webots screen darkens and become non-clickable as if the editor was open.

Not a regression

**Steps to Reproduce**
1. Open `atlas.wbt`
2. Run it on fast mode
3. Go to `TextureBackground` -> `skyColor` -> `0 0 0`
4. Click on the black square
5. The screen should darken

stefaniapedrazzi:
Fix #5090: pause the simulation while a dialog is open to avoid UI refresh issues.

I can reproduce the same issue in R2021b as well.

### DIFF
--- a/docs/reference/changelog-r2022.md
+++ b/docs/reference/changelog-r2022.md
@@ -71,6 +71,7 @@ Released on July, 7th, 2022.
     - Fixed animation time not starting at 0 seconds ([#4659](https://github.com/cyberbotics/webots/pull/4659)).
     - Fixed inverted left and right sound from speaker ([#4847](https://github.com/cyberbotics/webots/pull/4847)).
     - Fixed various crashes with devices whose top node is not a [Robot](robot.md) node ([#4878](https://github.com/cyberbotics/webots/pull/4878)).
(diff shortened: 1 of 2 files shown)


stefaniapedrazzi:
I tried to debug the QtWidgets module but didn't find the reason why the QColorDialog is blocked if the simulation is running in fast mode while the dialog is executed (I also tried to open it in asynchronous mode but nothing changed).
I tried to add additional calls to `QCoreApplication::processEvents` but it didn't help.
Then, it only occurs with the native color picker editor, changing it to the Qt one works correctly.
Pausing the simulation seems the simpler fix.

Just for info I checked and not all the dialogs are pausing the simulation:
* share, take screenshot, record movie, open/save world file pause the simulation seems to  be the only ones that are pausing the simulation
* all the other dialogs (add node,  preferences, help dialogs, etc.) don't stop the simulation and they are working correctly

BenjaminDeleze (approved):
Thank you!

PR #5099, opened 2022-08-22
Silence false GCC 12.2.0 warning

omichel:
GCC 12.1.0 and 12.2.0 seem to display false warnings about dangling pointers (tested on Windows). This PR extends the silencing of the warning to GCC 12.2.0 which came as a MSYS2 update today.

### DIFF
--- a/src/webots/nodes/WbConnector.cpp
+++ b/src/webots/nodes/WbConnector.cpp
@@ -373,7 +373,7 @@ void WbConnector::snapOrigins(WbConnector *other) {
       h[i] /= 2.0;
   }
 
-  // gcc version 12.1.0 and 12.2.0 is raising a false positive warning here about dangling pointers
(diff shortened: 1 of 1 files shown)


src/webots/nodes/WbConnector.cpp
```diff
@@ -373,9 +373,9 @@ void WbConnector::snapOrigins(WbConnector *other) {
       h[i] /= 2.0;
   }
 
-// gcc 12.1.0 is raising a false positive warning here about dangling pointers
+  // gcc version 12.1.0 and 12.2.0 is raising a false positive warning here about dangling pointers
```
BenjaminDeleze, line 376:
```suggestion
// gcc version 12.1.0 and 12.2.0 is raising a false positive warning here about dangling pointers
```

Not sure about the added spaces

BenjaminDeleze (changes requested):

BenjaminDeleze (approved):
Thank you

PR #5103, opened 2022-08-23
Fixed running Webots in UTF-8 path

### LINKED ISSUE #4757: UTF-8 characters not supported on Windows nightly build
Launching Webots when having UTF-8 characters in the account name doesn't work on Windows 11, when installed only for a local account and not for all users.

omichel:
Fix #4757.
I had to fix [libpico.dll](https://github.com/naggety/picotts/commit/049f942c833048a1dbabca3a8a52841618ba69f9) which was having problems with UTF-8 paths on Windows. Therefore, you will have to re-install pico to test this PR on Windows.
I can now run the guided tour without any error with Webots installed in a path with UTF-8 characters (e.g., Chinese characters).
I also tested a few other world files and they all run flawlessly in that path.

In addition to fixing UTF-8 paths, this PR also:
- Fixes a bug with the `mTemplate` field of `WbProtoModel` which was wrongly set: the `textured_proto_shape.wbt` example was causing the initialization of the Lua engine which was completely useless.
- The computation of the hash for the meshes was wrong, causing different UTF-8 URLs to have the same hash.


ygoumaz (changes requested):
I couldn't find any problem on Windows, it seems to work nicely. Just a few suggestions on the code.

ygoumaz (approved):
Thank you!

PR #5106, opened 2022-08-23
Fix tiago++ right arm shoulder

BenjaminDeleze:
ditto


omichel (approved):
Oops...
Thank you for spotting and fixing this.

PR #5107, opened 2022-08-23
Cleanup `WbNetwork` to a singleton class

ygoumaz:
The way `WbNetwork` is used is not consistent. Sometimes it is called as a singleton, sometimes not. This PR cleans up the class to get consistent use as a singleton class. 

The side effect of this inconsistency is that the value of `gCacheMap` was not the same in all calls to the `get`, `isCached` and `getUrlFromEphemeralCache` methods. This is leading to errors when loading worlds that have some PROTOs already present in the assets cache, e.g. the worlds loaded in the `world update` test of the CI (see #5087).

### DIFF
--- a/src/webots/core/WbNetwork.cpp
+++ b/src/webots/core/WbNetwork.cpp
@@ -118,7 +118,7 @@ void WbNetwork::save(const QString &url, const QByteArray &content) {
   }
 }
 
-const QString WbNetwork::get(const QString &url) {
+const QString &WbNetwork::get(const QString &url) {
   if (!mCacheMap.contains(url)) {
     const QString filePath = WbStandardPaths::cachedAssetsPath() + urlToHash(url);
     mCacheMap.insert(url, filePath);
--- a/src/webots/core/WbNetwork.cpp
+++ b/src/webots/core/WbNetwork.cpp
@@ -40,7 +40,6 @@ WbNetwork *WbNetwork::instance() {
 
 WbNetwork::WbNetwork() {
   mNetworkAccessManager = NULL;
-  mCacheMap.clear();
 
   // delete previous caching system folder (< R2022b)
   QDir oldCache(QStandardPaths::writableLocation(QStandardPaths::CacheLocation) + "/network/");
--- a/src/webots/core/WbNetwork.cpp
+++ b/src/webots/core/WbNetwork.cpp
@@ -104,7 +104,7 @@ void WbNetwork::setProxy() {
 }
 
 void WbNetwork::save(const QString &url, const QByteArray &content) {
-  if (!isCached(url)) {
+  if (!isCachedWithMapUpdate(url)) {
     // save to file
     const QString path(WbStandardPaths::cachedAssetsPath() + urlToHash(url));
     QFile file(path);
@@ -127,7 +127,7 @@ const QString &WbNetwork::get(const QString &url) {
   return mCacheMap[url];
 }
 
-bool WbNetwork::isCached(const QString &url) {
+bool WbNetwork::isCachedWithMapUpdate(const QString &url) {
   if (mCacheMap.contains(url))  // avoid checking for file existence (and computing hash again) if asset is known to be cached
     return true;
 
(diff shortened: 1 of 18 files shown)


src/webots/core/WbNetwork.cpp
```diff
@@ -44,7 +40,7 @@ WbNetwork *WbNetwork::instance() {
 
 WbNetwork::WbNetwork() {
   mNetworkAccessManager = NULL;
-  gCacheMap.clear();
+  mCacheMap.clear();
```
omichel, line 43: This is not needed as a `QMap` is initially empty.
```suggestion
```

src/webots/core/WbNetwork.cpp
```diff
@@ -118,28 +114,28 @@ void WbNetwork::save(const QString &url, const QByteArray &content) {
...
       // save reference in internal representation
-      gCacheMap.insert(url, path);
+      mCacheMap.insert(url, path);
     }
   }
 }
 
-const QString &WbNetwork::get(const QString &url) {
-  if (!gCacheMap.contains(url)) {
+const QString WbNetwork::get(const QString &url) {
```
omichel, line 122: Why did you remove the reference here?

omichel (changes requested):

src/webots/core/WbNetwork.cpp
ygoumaz: This reference was not present before the changes applied by @stefaniapedrazzi in #5023. So I thought it was not useful. If we use a reference for the `get` method, should we also use one for `getUrlFromEphemeralCache`?

omichel: Yes, that's probably better.

ygoumaz: Actually, `getUrlFromEphemeralCache` is returning a temporary object, so it is not a good idea to use a reference. I still put it back for the `get` method.

stefaniapedrazzi: We should always return the reference instead of a copy when possible.

omichel:
The CI test failed because:
```
src/webots/nodes/WbCadShape.cpp:233:39: warning: Assert statement calls a function which may have desired side effects: 'isCached'. [assertWithSideEffect]
        assert(WbNetwork::instance()->isCached(adjustedUrl));
```
I believe we wrote this assert mainly because the `isCached()` method name is misleading: this is not a const method as we would expect it with such a name. It does possibly modify the object, which is not expected, hence the warning... So, I believe we should rename this method and probably remove the assert or use a different one.

ygoumaz:
All tests passed, so this PR is ready for review.

src/webots/core/WbNetwork.hpp
```diff
@@ -25,15 +25,20 @@ class WbNetwork {
   QNetworkAccessManager *networkAccessManager();
   void setProxy();
 
-  static bool isCached(const QString &url);
-  static const QString &get(const QString &url);
+  // check the file in the cache and add it to the internal representation. Outside of the assets, this function should be used
+  // above isCachedNoMapUpdate().
+  bool isCachedWithMapUpdate(const QString &url);
+  // check the file in the cache without adding it to the internal representation. This one should only be used in assertions
+  // because they cannot have side effects.
```
omichel, line 32:
```suggestion
  // check the file in the cache without adding it to the internal representation (for use in asserts, no side effect).
```

src/webots/core/WbNetwork.hpp
```diff
@@ -25,15 +25,20 @@ class WbNetwork {
   QNetworkAccessManager *networkAccessManager();
   void setProxy();
 
-  static bool isCached(const QString &url);
-  static const QString &get(const QString &url);
+  // check the file in the cache and add it to the internal representation. Outside of the assets, this function should be used
+  // above isCachedNoMapUpdate().
```
omichel, line 29:
```suggestion
  // check the file in the cache and add it to the internal representation (don't use it in asserts).
```

omichel (changes requested):
Thank you, it looks good to me.
I suggest to slightly improve the comments.

omichel (approved):
Thank you.

PR #5109, opened 2022-08-23
Fix the switching of worlds in webotsJS

BenjaminDeleze:
**Description**
Due to past changes, the list of the world was not sent anymore to webotsJS as `nao_demo.wbt` for example but as `worlds/nao_demo.wbt`.

This causes a bug because when WebotsJS was sending back the name to request to switch the worlds, Webots adds the name from WebotsJS to the path of the current world which would give something like:
```
/home/benjamin/webots/projects/robots/adept/pioneer3/worlds//worlds/pioneer3dx_gripper.wbt
```

### DIFF
--- a/src/webots/gui/WbTcpServer.cpp
+++ b/src/webots/gui/WbTcpServer.cpp
@@ -613,10 +613,11 @@ void WbTcpServer::sendWorldToClient(QWebSocket *client) {
 
   QString worlds;
   foreach (const QFileInfo item, worldFiles)
-    worlds += QDir(WbProject::current()->dir()).relativeFilePath(item.absoluteFilePath()) + ";";
+    worlds += QDir(WbProject::current()->dir()).relativeFilePath(item.absoluteFilePath()).split('/')[1] + ";";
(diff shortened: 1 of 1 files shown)


src/webots/gui/WbTcpServer.cpp
```diff
@@ -613,10 +613,11 @@ void WbTcpServer::sendWorldToClient(QWebSocket *client) {
 
   QString worlds;
   foreach (const QFileInfo item, worldFiles)
-    worlds += QDir(WbProject::current()->dir()).relativeFilePath(item.absoluteFilePath()) + ";";
+    worlds += QDir(WbProject::current()->dir()).relativeFilePath(item.absoluteFilePath()).split('/')[1] + ";";
```
omichel, line 616: Why is that so complicated? Can we simply do it this way?
```suggestion
    worlds += item.fileName() + ";";
```

omichel (approved):
Better this way indeed.

ad-daniel (approved):
Looks fine, thank you

PR #5113, opened 2022-08-24
Fix additional mesh loading

### LINKED ISSUE #5111: Meshes are loaded too many times
If you create a very simple world with only a `RubberDuck` proto and if you add a qDebug() statement in https://github.com/cyberbotics/webots/blob/b42cbe6c9ae0c009c0b11436843325e8508f1a01/src/webots/nodes/WbMesh.cpp#L162, you will see that each `Mesh` node of the `RubberDuck` proto is created 3 times by calling the assimp importer function. I believe we should fix this so that they are created only once, as parsing an OBJ file is a costly operation.

BenjaminDeleze:
Fix #5111 

I just replace `updateUrl()` at its initial place.

I could not find any side effect to that. Either the function was moved by error or the thing it was supposed to fix was also fixed by something else in the meantime.


omichel (approved):
Thank you.

PR #5116, opened 2022-08-24
Fix URL format when converting to base node

ygoumaz:
Currently, when converting a PROTO to base node from the scene tree, the URLs of `ImageTextures`, `CadShapes`, `Backgrounds` and `Meshes` are automatically converted to web format (`raw.githubusercontent.com/`). This should not be the case if we are not in a distribution and not streaming in x3d mode.


ygoumaz:
When an asset URL is defined in relative format, the conversion to base node computes the complete URL (absolute or web) from the URL of the parent. In the process, the filename is removed from the parent URL string. On Windows, the method of the QUrl class is formatting the URL with specific rules, including the hostname to be lower case. This is leading to absolute URL to be in the following format: `c:/path/to/parent/proto` instead of `C:/path/to/parent/proto`. Therefore the comparison with the WEBOTS_HOME path was returning a wrong result, leading to weird relative URLs after the conversion to base node.

omichel (approved):
It now works as expected on Windows as well.
Thank you.

PR #5124, opened 2022-08-26
Fix switch of worlds when creating a new one

### LINKED ISSUE #5123: Switch to world when creating a new one
**Describe the Bug**

To reproduce:
1. Create a new project locally
2. Add random stuff to the world and save
3. Click `New > New World File` and inputting valid settings
4. The world is indeed created in the local directory, however Webots doesn't switch to it.

BenjaminDeleze:
Fix #5123


ygoumaz (approved):
Thank you!

PR #5125, opened 2022-08-26
Fix converting robot windows and controllers to generic with 'convert to base node'

ygoumaz:
When converting a distributed PROTO to base node, the controllers and robot windows are often not present locally in the current project. They are often located either remotely or in the Webots installation folder.

This PR aims at converting the `window` and `controller` fields  to `<generic>` when a robot converted to base node can't find its previous controller or window in the local project.


src/webots/nodes/WbRobot.hpp
```diff
@@ -74,6 +74,7 @@ class WbRobot : public WbSolid {
   bool isWaitingForWindow() const { return mWaitingForWindow; }
   void setWaitingForWindow(bool waiting);
   void addNewlyInsertedDevice(WbNode *node);
+  void updatingMissingResources() const override;
```
omichel, line 77: We usually don't use the progressive tense form (-ing) for methods, I would suggest to rename this method with a present tense form, e.g. without -ing:
```suggestion
  void fixMissingResources() const override;
```

omichel (changes requested):

omichel (approved):
Thank you.

ad-daniel:
I'm a bit late, should we perhaps notify the user in the console that the change took place? Before it was evident from the error, now maybe not so much

ygoumaz:
I think it is indeed more user-friendly for the user. For consistency, I would also display these warnings when editing remote PROTOs locally  and when creating a derived PROTO from the PROTO wizard.

Issue #5076, opened 2022-08-15
Broken textures


BenjaminDeleze:
Some textures are currently broken on master, for example in `highway_overtake` and `village_realistic`.

It seems that some textures where broken by https://github.com/cyberbotics/webots/pull/5003
And some other by https://github.com/cyberbotics/webots/pull/5071

BenjaminDeleze:
I just check a bit the first part of the issue:
Since #5003, webots does not found anymore the textures for the left and right indicators of the `lincoln` car like `textures/lincoln_mkz_details_base_color.jpg`

It is because, webots try to retrieve those textures from the `vehicles/protos/abstract/textures/` folder instead of the `vehicles/protos/lincoln` one.
This could be partially explained because the indicators are parameters of the the `car` proto which is in the `abstract` folder. However it is somehow more specific because the `textures/lincoln_mkz_details_base_color.jpg` image is used at another place in the `lincoln` proto and is retrieved correctly in that case.

ygoumaz:
The issue is probably mainly coming from #5071. The implemented condition in `WbUrl::computePath` seems wrong when dealing with derived PROTOs. The alias value is set to the lowest field, which is (in this case) "url". When looking in the first ancestry of the list, if the "url" field is not directly exposed, the first ancestry will always be chosen. This way the `TiagoIron` works because `TiagoBase` should be chosen. But in the case of the `ToyotaPriusWheel`, the `VehicleWheel` should not be chosen but it actually is.

Two things should probably be adapted:
- The alias to look for. Currently, the "url" field is not correct, because there is only little chance that it is contained in the parameterAliases list of the protoModel.
- In the case of Slots, there is probably a special case to handle, but not completely sure.

Currently, I don't see any easy solution to solve this problem.

ad-daniel:
I think fundamentally it isn't wrong to follow the aliases from the deepest ancestor. Derived PROTO by design build upon the PROTO they depend on, in most cases the fields remain common to all ancestors (and new ones are added on top), the logic handles the case if at different levels the name of the field itself changes, which is also possible although less common.

[...]

PR #5126, opened 2022-08-26
Rework relative URL resolver

ad-daniel:
**Description**

Fixes https://github.com/cyberbotics/webots/issues/5076

Continuation of https://github.com/cyberbotics/webots/pull/5088. As mentioned there, the previous approach didn't work for relative urls in several more complex cases (relative URL in PROTO header, derived PROTO with multiple ancestors, slots, ...)


tests/test_worlds.py
```diff
@@ -36,11 +36,11 @@ def setUp(self):
             # the ros controller of complete_test.wbt is started when loading the world because the robot-window is open
             'Failed to contact master at',
             'Cannot initialize the sound engine',
-            'System below the minimal requirements',
             self.crashError  # To remove once #6125 is fixed
         ]
+        # Set empty.wbt as the first world (to trigger the 'System below the minimal requirements' message)
+        self.worlds = [os.path.join(os.environ['WEBOTS_HOME'], 'resources/projects/empty.wbt')]
```
stefaniapedrazzi, line 42: This path is wrong and it is missing the `worlds` folder:
```suggestion
        self.worlds = [os.path.join(os.environ['WEBOTS_HOME'], 'resources/projects/worlds/empty.wbt')]
```

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -1268,6 +1268,42 @@ bool WbNodeUtilities::isVisible(const WbField *target) {
...
+  const WbNode *parameterNode = proto;
+  while (parameterNode) {
+    const WbField *parameter = field;
+    const QVector<WbField *> parameterList = parameterNode->parameters();
+    while (parameter) {
+      if (parameterList.contains(const_cast<WbField *>(parameter))) {
+        if (parameter->isDefault())
+          return true;
+        else
+          return false;
```
omichel, line 1284: Simpler:
```suggestion
        return parameter->isDefault();
```

tests/rendering/controllers/complex_relative_asset_resolution_checker/complex_relative_asset_resolution_checker.c
```diff
@@ -0,0 +1,57 @@
...
+static WbDeviceTag cameras[NB_CAMERAS];
+
+void test_camera_color(int i, const int expected_color[3]) {
+  int r, g, b;
+  const int width = wb_camera_get_width(cameras[i]);
+
+  const unsigned char *image = wb_camera_get_image(cameras[i]);
+  r = wb_camera_image_get_red(image, width, 0, 0);
+  g = wb_camera_image_get_green(image, width, 0, 0);
+  b = wb_camera_image_get_blue(image, width, 0, 0);
```
omichel, line 25:
```suggestion
  const int width = wb_camera_get_width(cameras[i]);

  const unsigned char *image = wb_camera_get_image(cameras[i]);
  const int r = wb_camera_image_get_red(image, width, 0, 0);
  const int g = wb_camera_image_get_green(image, width, 0, 0);
  const int b = wb_camera_image_get_blue(image, width, 0, 0);
```

tests/rendering/controllers/complex_relative_asset_resolution_checker/complex_relative_asset_resolution_checker.c
```diff
@@ -0,0 +1,57 @@
...
+
+#include "../../../lib/ts_assertion.h"
+#include "../../../lib/ts_utils.h"
+
+#define TIME_STEP 64
+#define NB_CAMERAS 12
+
+static int red[3] = {203, 0, 0};
+static int green[3] = {35, 203, 0};
+static int blue[3] = {0, 18, 203};
```
omichel, line 14:
```suggestion
static const int red[3] = {203, 0, 0};
static const int green[3] = {35, 203, 0};
static const int blue[3] = {0, 18, 203};
```

tests/rendering/controllers/texture_color_checker/texture_color_checker.c
```diff
@@ -7,7 +7,16 @@
...
+#define NB_CAMERAS 11
+
+static int black[3] = {0, 0, 0};
+static int red[3] = {203, 0, 0};
+static int green[3] = {35, 203, 0};
+static int blue[3] = {0, 18, 203};
+static int yellow[3] = {203, 196, 0};
+static int cyan[3] = {6, 176, 203};
+static int orange[3] = {203, 85, 0};
+static int pink[3] = {203, 0, 175};
```
omichel, line 19:
```suggestion
static const int black[3] = {0, 0, 0};
static const int red[3] = {203, 0, 0};
static const int green[3] = {35, 203, 0};
static const int blue[3] = {0, 18, 203};
static const int yellow[3] = {203, 196, 0};
static const int cyan[3] = {6, 176, 203};
static const int orange[3] = {203, 85, 0};
static const int pink[3] = {203, 0, 175};
```

tests/rendering/protos/depth_2/NestedOverwrittenExposedTexture.proto
```diff
@@ -0,0 +1,17 @@
+#VRML_SIM R2022b utf8
+
+EXTERNPROTO "./depth_3/ExposedTexture.proto"
```
omichel, line 3: We should also test without the `./` prefix which is a bit superfluous.
```suggestion
EXTERNPROTO "depth_3/ExposedTexture.proto"
```

tests/rendering/protos/derived_proto/IntermediaryAncestor/IntermediaryAncestorNoChildren.proto
```diff
@@ -0,0 +1,14 @@
+#VRML_SIM R2022b utf8
+
+EXTERNPROTO "./LowestAncestor/SolidAncestorInternalTexture.proto"
```
omichel, line 3:
```suggestion
EXTERNPROTO "LowestAncestor/SolidAncestorInternalTexture.proto"
```

tests/rendering/protos/derived_proto/MiddleProtoUpperScope/MiddleProtoUpperScope.proto
```diff
@@ -1,6 +1,6 @@
 #VRML_SIM R2022b utf8
 
-EXTERNPROTO "../LowerProto/LowerProtoUpperScope.proto"
+EXTERNPROTO "./LowerProtoUpperScope/LowerProtoUpperScope.proto"
```
omichel, line 3:
```suggestion
EXTERNPROTO "LowerProtoUpperScope/LowerProtoUpperScope.proto"
```

tests/rendering/protos/derived_proto/TextureInternalIntermediaryAncestor.proto
```diff
@@ -0,0 +1,14 @@
+#VRML_SIM R2022b utf8
+
+EXTERNPROTO "./IntermediaryAncestor/IntermediaryAncestorWithInternalTexture.proto"
```
omichel, line 3:
```suggestion
EXTERNPROTO "IntermediaryAncestor/IntermediaryAncestorWithInternalTexture.proto"
```

tests/rendering/protos/derived_proto/TextureInternalTopmostAncestor.proto
```diff
@@ -0,0 +1,25 @@
+#VRML_SIM R2022b utf8
+
+EXTERNPROTO "./IntermediaryAncestor/IntermediaryAncestor.proto"
```
omichel, line 3:
```suggestion
EXTERNPROTO "IntermediaryAncestor/IntermediaryAncestor.proto"
```

tests/rendering/protos/derived_proto/UpperProtoMiddleScope.proto
```diff
@@ -1,6 +1,6 @@
 #VRML_SIM R2022b utf8
 
-EXTERNPROTO "../MiddleProto/MiddleProtoMiddleScope.proto"
+EXTERNPROTO "./MiddleProtoMiddleScope/MiddleProtoMiddleScope.proto"
```
omichel, line 3:
```suggestion
EXTERNPROTO "MiddleProtoMiddleScope/MiddleProtoMiddleScope.proto"
```

tests/rendering/protos/derived_proto/UpperProtoWorldScope.proto
```diff
@@ -1,6 +1,6 @@
 #VRML_SIM R2022b utf8
 
-EXTERNPROTO "../MiddleProto/MiddleProtoWorldScope.proto"
+EXTERNPROTO "./MiddleProtoWorldScope/MiddleProtoWorldScope.proto"
```
omichel, line 3:
```suggestion
EXTERNPROTO "MiddleProtoWorldScope/MiddleProtoWorldScope.proto"
```

omichel (changes requested):

tests/rendering/protos/depth_2/NestedOverwrittenExposedTexture.proto
ad-daniel: I've added it precisely because it is superfluous, since the relative path needs to be combined with the location path, it's to ensure it does the correct job of concatenating even if people happen to add it.

>We should also test without the `./`

So should I duplicate all tests and check both versions? Or just mix up the existing tests (so not all are `./`)

omichel: I believe that simply mixing-up the existing tests is enough.

omichel:
FYI, I compared the loading time for `CH_Morges.wbt` on this branch versus the master branch and the result is very much the same on both branches (around 2 minutes and 10 seconds on my laptop under Windows). So this PR doesn't seem to introduce any performance penalty at load time.

src/webots/vrml/WbNode.cpp
```diff
@@ -1116,6 +1120,24 @@ QStringList WbNode::listTextureFiles() const {
   return list;
 }
 
+const WbNode *WbNode::containingProto(bool skipThis) const {
+  const WbNode *n = this;
+  while (n) {
+    const WbProtoModel *protoModel = n->proto();
+    if (protoModel && (!skipThis || (skipThis && n != this)))
+      return n;
```
BenjaminDeleze, line 1128:
```suggestion
    if (protoModel && (!skipThis || n != this))
      return n;
```

`n != this` will be evaluated only if `skipThis` is `true`

src/webots/vrml/WbNode.cpp
```diff
@@ -1116,6 +1120,24 @@ QStringList WbNode::listTextureFiles() const {
...
+    else {
+      const WbNode *ppn = n->protoParameterNode();
+      if (ppn && ppn->proto() && (!skipThis || (skipThis && ppn->proto() != proto())))
```
BenjaminDeleze, line 1131:
```suggestion
      if (ppn && ppn->proto() && (!skipThis || ppn->proto() != proto()))
```

same

BenjaminDeleze (changes requested):

src/webots/vrml/WbNode.cpp
ad-daniel: I don't think I understand, or at least, what I understood is intentional.
Basically the idea is: if I pass a node n to this function, and let's assume this node happens to be a PROTO already:
- if skipThis is false: I can return n immediately since it's already a PROTO (the second part of the OR isn't considered since the first is already TRUE)
- if skipThis is true: then I return n only if it differs (i.e. it would fail the check for n, but not for an upper PROTO of n)

ad-daniel: ah nevermind, I see what you mean

BenjaminDeleze: It simplifies the expression, but I don't know if it is maybe clearer to let the `(skipThis &&`

ad-daniel:
Locally the `test_worlds` test passes without any warning, I'll let it run in the CI but I think the PR can be reviewed already in the meantime.

Here's the sample case found by Benjamin where the logic failed, if needed: [slot2.zip](https://github.com/cyberbotics/webots/files/9451554/slot2.zip)

BenjaminDeleze (dismissed):
Thank you!

omichel (dismissed):
Looks good to me as well.

stefaniapedrazzi:
How should relative URLs in default parameters be handled?
Should they be resolved from the PROTO path or the world path?
I still testing but none of them seems to work in a custom project.

omichel:
> How should relative URLs in default parameters be handled? Should they be resolved from the PROTO path or the world path? I still testing but none of them seems to work in a custom project.

They should be resolved relative to the PROTO path. If that doesn't work for you, could you post a example here?

ad-daniel:
> I still testing but none of them seems to work in a custom project.

What do you mean by custom project? The 3 worlds used for testing and the zip above are custom projects too. Can you provide an example?

stefaniapedrazzi:
Here is the procedure to reproduce the error:
1. Open `highway_driving.wbt`
2. Edit `LincolnMKZ.proto` in the Text Editor and accept relocation
3. Change the `plate` URL to the relative path `textures/plate.jpg` and copy the image in the `<current_project>/protos/textures/` folder
4. Save and Reload the world
Error:
```
ERROR: Impossible to infer URL from 'textures/plate.jpg' and ''
ERROR: Impossible to infer URL from 'textures/plate.jpg' and ''
ERROR: Impossible to infer URL from 'textures/plate.jpg' and ''
WARNING: DEF WEBOTS_VEHICLE0 LincolnMKZ (PROTO) > Group > Transform > DEF PLATES Shape > PBRAppearance > ImageTexture: Texture 'textures/plate.jpg' not found.
ERROR: Impossible to infer URL from 'textures/plate.jpg' and ''
ERROR: Impossible to infer URL from 'textures/plate.jpg' and ''
```

ygoumaz:
I don't know if it's supposed to behave this way, but I think I found a problem:

1. Open Webots and add a Bed to the world.
2. In the scene tree, expand Bed>blanketAppearance.
3. Change the roughness to 0.5.
4. Now save and reload. The texture of the blanket is no longer found (which is correct, as it should now be relative to the world). The problem is that the texture was still found before saving and reloading. Shouldn't the missing texture already be displayed before saving and reloading?

omichel:
> I don't know if it's supposed to behave this way, but I think I found a problem:
> 
>     1. Open Webots and add a Bed to the world.
> 
>     2. In the scene tree, expand Bed>blanketAppearance.
> 
>     3. Change the roughness to 0.5.
> 
>     4. Now save and reload. The texture of the blanket is no longer found (which is correct, as it should now be relative to the world). The problem is that the texture was still found before saving and reloading. Shouldn't the missing texture already be displayed before saving and reloading?

I agree, this behavior is wrong. However, I am not sure we want to fix it. Here is why. If some PROTO parameters are used in JavaScript code, they cause the re-generation of the whole PROTO when they are changed. This was the example I tested (TruckTrailer transparency) and it worked as expected because the whole PROTO was regenerated after changing the default parameter. However, in this simpler case, the roughness is not used in JavaScript code and changing it doesn't trigger the re-regeneration of the PROTO. Thus, the texture is not re-loaded. To fix this issue, we should enforce the re-generation of the PROTO in every case where we change the default parameter. However, re-generating a PROTO is a costly operation and there are use cases where it could be a real pain. For example, if a supervisor changes regularly the default value of a PROTO to a different value and then back to the default value, the re-generation will take place every time step, causing an important performance drop. Therefore, I suggest that we simply document the fact that textures used in PROTO parameters may not be properly updated when changing other PROTO parameters until the world is saved and reloaded.

stefaniapedrazzi:
> > I don't know if it's supposed to behave this way, but I think I found a problem:
> > ```
> > 1. Open Webots and add a Bed to the world.
> > 
> > 2. In the scene tree, expand Bed>blanketAppearance.
> > 
> > 3. Change the roughness to 0.5.
> > 
> > 4. Now save and reload. The texture of the blanket is no longer found (which is correct, as it should now be relative to the world). The problem is that the texture was still found before saving and reloading. Shouldn't the missing texture already be displayed before saving and reloading?
> > ```
> 
> I agree, this behavior is wrong. However, I am not sure we want to fix it. Here is why. If some PROTO parameters are used in JavaScript code, they cause the re-generation of the whole PROTO when they are changed. This was the example I tested (TruckTrailer transparency) and it worked as expected because the whole PROTO was regenerated after changing the default parameter. However, in this simpler case, the roughness is not used in JavaScript code and changing it doesn't trigger the re-regeneration of the PROTO. Thus, the texture is not re-loaded. To fix this issue, we should enforce the re-generation of the PROTO in every case where we change the default parameter. However, re-generating a PROTO is a costly operation and there are use cases where it could be a real pain. For example, if a supervisor changes regularly the default value of a PROTO to a different value and then back to the default value, the re-generation will take place every time step, causing an important performance drop. Therefore, I suggest that we simply document the fact that textures used in PROTO parameters may not be properly updated when changing other PROTO parameters until the world is saved and reloaded.

I agree that the PROTO should not be regenerated.
But for example when converting to base node, the relative URLs are modified to work even after the scope changed from PROTO to world. Maybe the same mechanism could be applied here.

stefaniapedrazzi:
I did some other tests and everything else seems to work nicely.

ad-daniel:
The issue can be reproduced more easily by adding a Lincoln to an empty world and changing the plate url from `webots://projects/vehicles/protos/textures/plate.jpg` to `../textures/plate.jpg`.

The issue stems from this line: https://github.com/cyberbotics/webots/blob/bd75fc8daba67f665f4b60b65c0e07a46c5515e0/src/webots/nodes/utils/WbUrl.cpp#L91

Basically the procedure works in terms of finding that the relevant parameter is `plate` and that `Lincoln` is the relevant PROTO, however since the vehicle is also a derived PROTO, the scope is supposed to be defined/stored in the `mScope` variable while the derived PROTO is built (since the intermediary nodes that comprise the derivation are not kept) and when resolving the relative URL, in order to know which of the ancestors actually defines it, we should only need to read `f->scope()`. Or at least, that's the idea. What happens is that the line shown above returns empty so either it was never set, or at some point it gets overwritten (usually when fields are copied/replaced in the build process)

ad-daniel:
Probably found it, but will need to check if it creates side-effects with other derived PROTO but so far doesn't seem to

ygoumaz:
Maybe this is more related to #5092, but the following behavior seems to be wrong:
1. Open Webots and add an official `LincolnMKZ`.
2. In one of the slots (e.g. `sensorsSlotCenter`), add an official `TruckSimple`.
3. In one of the slots of the Truck (e.g. sensorsSlotCenter), add an official `LincolnMKZ`.
4. Now change the URL of the plate in the Lincoln added in step 3 to something invalid. The PROTO is not regenerated and no missing texture is displayed.
5. If you apply the same URL update to the Lincoln of step 1, the behavior is correct.

ad-daniel:
> Maybe this is more related to #5092, but the following behavior seems wrong:
>     1. Open Webots and add an official `LincolnMKZ`.
>     2. In one of the slots (e.g. `sensorsSlotCenter`), add an official `TruckSimple`.
>     3. In one of the slots of the Truck (e.g. `sensorsSlotCenter`), add an official `LincolnMKZ`.
>     4. Now change the URL of the plate in the Lincoln added in step 3 to something invalid. The PROTO is not regenerated and no missing texture is displayed.
>     5. If you apply the same URL update to the Lincoln of step 1, the behavior is correct.

It's odd yes, but considering the behavior is identical while the doing the same procedure but with non-relative urls (invalid `webots://` URL for example), I suspect the issue isn't related to this PR

omichel:
> It's odd yes, but considering the behavior is identical while the doing the same procedure but with non-relative urls (invalid `webots://` URL for example), I suspect the issue isn't related to this PR

Assuming this behavior shows up on the master branch as well with non-relative URLs, we should open a new issue about this.

ygoumaz (approved):
I have opened an issue: #5148. Apart from that, I haven't found any further issues.

omichel (approved):

ad-daniel:
Personally I haven't encountered side effects with the fix so far, so unless there are other problematic cases I think we can merge this

@stefaniapedrazzi are you still testing this PR?

stefaniapedrazzi:
I will test now the latest changes.

stefaniapedrazzi (approved):
Seems to work fine now.

omichel:
🚀

PR #5129, opened 2022-08-29
Print postponed log messages when load is cancelled

### LINKED ISSUE #5127: No errors explaining why world cannot be loaded
When opening an invalid world the error explaining that the world is invalid and cannot be loaded is not immediately printed in the console and Webots seems buggy.
The error message(s) is printed in the console only when loading a different world.

Steps to reproduce:
1. Start Webots
2. Try to open `projects/languages/matlab/worlds/e-puck_matlab.wbt`
    -> nothing happens
3. Open a different world or reload the current one
    -> the error messages related to the matlab world are now printed in the console.    
    
 This is a regression and it is working as expected in R2022a.

The texture URL in `e-puck_matlab.wbt` should be changed to `textures/logo.jpg` (not existing) to reproduce the load failure.

stefaniapedrazzi:
Fix #5127:
always unset the postponed logs flag and print queued messages when canceling a world load.

### DIFF
--- a/src/webots/app/WbApplication.cpp
+++ b/src/webots/app/WbApplication.cpp
@@ -160,11 +160,11 @@ void WbApplication::cancelWorldLoading(bool loadEmpty, bool deleteWorld) {
     mWorld = NULL;
   }
 
-  if (loadEmpty) {
-    WbLog::setConsoleLogsPostponed(false);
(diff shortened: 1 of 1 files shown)


omichel (approved):
Thank you.

PR #5132, opened 2022-08-29
Improve warning for missing resources with absolute path display

ygoumaz:
Defining resources (textures, sounds, meshes, ...) with relative URLs is a very good and efficient practice. However, sometimes it may not be clear to the user which path Webots takes as reference for the relative URL. Sometimes it is relative to the world, while sometimes it is relative to the PROTO file.

This PR adds a new warning showing the user where its texture should be saved when it can't be found.


ad-daniel (changes requested):
Is it intended that the warning messages show up as 2 separate ones?
For example I get:
```
WARNING: stuff > ImageTexture: Unable to find resource at '/home/daniel/Desktop/slot2/slot/protos/textures/truck_trailer_webots22.jpg'.
WARNING: stuff > ImageTexture: Texture 'textures/truck_trailer_webots22.jpg' not found.
```
Can it be made a single one? Or is it too complicated (since they are printed at different places)?

I've also noticed that CadShape prints the `Unable to find resource ...` twice, seems to be the only case.

It's less important, and not related to this PR but when setting an invalid sound URL (for example, change the `bumpSound` of `asymmetric_friction.wbt` to something relative yet invalid, the message shown is:
```
WARNING: WorldInfo > ContactProperties: Unable to find resource at '/home/daniel/webots_master/projects/default/worlds/sounds/bumps.wav'.
WARNING: WorldInfo > ContactProperties: First item of 'bumpSound' field is empty.
```
The second message should probably behave like it does for everything else (meshes, imagetexture, ...). It's empty because it was invalid, not because it's actually empty.

Overall though I agree that this approach is more user friendly

ygoumaz:
Regarding the 2 separate messages, there are 3 options:

1. The current approach. I think that having 2 messages is not really a problem, because one gives the path information and the second one gives more precision on the type of missing resource (texture, mesh, ...). So (apart from the node reference) there is no redundant information. Moreover, the necessary changes are very small.
2. Merge the two warnings into one by getting the "wrong URL" from `WbUrl` in all other files (`WbImageTexture`, `WbMesh`, ...). But this would probably require using a global variable to keep track of the bad URL in addition to the "missing texture" URL that is returned. I've tried this approach before but it was clearly more complex, less clean and I'm not really sure it's worth it.
3. Pass the resource type to `computePath` and display a different warning for each type. Again, I'm not sure if this extra complexity is really worth it.

I will remove the redundancy with CadShapes.

As for the sound, it should indeed show a different warning if a URL is given but invalid. I will fix it in this PR.

omichel:
I believe it is not nice to have two warnings for a single problem. So, I believe it is worth trying to display only one warning (displaying only the first one would be sufficient I think).

ygoumaz:
I removed the double warnings. Now only the one showing the absolute path to the resource should now be displayed.

`CadShapes` should now also display a single `Unable to find resource ...`.

An invalid sound URL is not showing an empty warning anymore.

ad-daniel (changes requested):
I don't know if it's related to the changes done in this PR or if it was already the case:
- load `asymmetric_friction.wbt` and set invalid relative URLs and press play => still can hear the sound despite the console saying they couldn't be found.
- Create a `Robot > Skin`, set `modelUrl` to `a.fbx`, Webots crashes

src/webots/nodes/WbCamera.cpp
```diff
@@ -1156,7 +1156,7 @@ void WbCamera::updateNoiseMaskUrl() {
     noiseMaskPath = completeUrl;
 
   const QString error = mWrenCamera->setNoiseMask(noiseMaskPath);
-  if (!error.isEmpty())
+  if (!completeUrl.isEmpty() && !error.isEmpty())
```
ad-daniel, line 1159: Just wondering, in theory the `noiseMarkUrl` is also a `jpg` file and if that's the case, couldn't `completeUrl` value be `missingTexture()` too ?

src/webots/nodes/WbContactProperties.cpp
```diff
@@ -243,8 +243,10 @@ void WbContactProperties::loadSound(int index, const QString &sound, const QStri
   if (WbUrl::isWeb(completeUrl)) {
     assert(WbNetwork::instance()->isCachedNoMapUpdate(completeUrl));  // by this point, the asset should be cached
     *clip = WbSoundEngine::sound(WbNetwork::instance()->get(completeUrl), extension);
-  } else
-    *clip = WbSoundEngine::sound(WbUrl::computePath(this, name, completeUrl), extension);
+  } else {
+    if (!completeUrl.isEmpty())
```
ad-daniel, line 247: ditto (at least for the backgrounds, irradiances probably wouldn't (since they're `.hdr` and `missingTexture` only checks jpg and png)

ad-daniel (changes requested):

ygoumaz:
I fixed a lot of issues in the last commits. Here is a non-exhaustive list:

- An invalid sound in `WbContactProperties` was leading to a warning being displayed but the previous sound was still there.
- I have added exceptions for double warnings when an invalid .png or .jpg file was used in the field leading to the `missing_texture.png` to be used.
- `WbMotor`, `WbSkin`, `WbContactProperties` and `WbCamera` were crashing in some cases (use of .jpg or .png files, invalid URLs, passing a real file with unsupported extension).

The PR is ready for review. The following things should be tested:
- `WbContactProperties` sound fields
- `WbBackground` URLs
- `WbCadshape` URL
- `WbImageTexture` URL
- `WbMesh` URL
- `WbMotor` sound
- `WbSkin` modelUrl
- `WbCamera` noiseMaskUrl

For each of these fields, different URLs should be tested:
- an invalid file (other than jpg or png)
- an existing file but unsupported extension
- an existing png or jpg
- an invalid png or jpg
- a valid file which shouldn't trigger any warning

docs/reference/changelog-r2022.md
```diff
@@ -80,7 +80,9 @@ Released on July, 7th, 2022.
     - Fixed drag events when using the orthographic projection mode ([#5080](https://github.com/cyberbotics/webots/pull/5080)).
     - Fixed "show resize handles" check box status in the Field Editor ([#5080](https://github.com/cyberbotics/webots/pull/5080)).
     - Fixed the background reflection on objects when the `skyColor` is updated ([#5133](https://github.com/cyberbotics/webots/pull/5133)).
+    - Fixed multiple crashes when using invalid URLs in the fields of [Skin](skin.md), [Motor](motor.md), [Camera](camera.md) and [ContactProperties](contactproperties.md) ([#5132](https://github.com/cyberbotics/webots/pull/5132)).
     - Fixed the sanitization of `WbPbrAppearance` to also trigger at the creation of the node ([#5139](https://github.com/cyberbotics/webots/pull/5139)).
+
```
ad-daniel, line 85:
```suggestion
```

src/webots/nodes/WbMotor.cpp
```diff
@@ -313,8 +313,10 @@ void WbMotor::updateSound() {
     const QString extension = completeUrl.mid(completeUrl.lastIndexOf('.') + 1).toLower();
     if (WbUrl::isWeb(completeUrl))
       mSoundClip = WbSoundEngine::sound(WbNetwork::instance()->get(completeUrl), extension);
-    else
+    else if (!(completeUrl.isEmpty() || completeUrl == WbUrl::missingTexture()))
```
ad-daniel, line 316: Why does the motor need to check for missing texture?

ad-daniel (changes requested):
Works very well! Just one last question

src/webots/nodes/WbMotor.cpp
ygoumaz: This is in the case the user inputs an invalid .png or .jpg file in the `sound` field. This would put the `missing_texture.png` URL in `completeUrl`. If you don't add this condition here, warnings will be triggered but the `mSoundClip` won't be updated to NULL, which would cause the sound to still work even if an invalid file was given in the field.

omichel: Maybe we should add a comment in the code explaining this, as it is not obvious.

ad-daniel (approved):
Thank you

ygoumaz:
FYI I have added an additional comment in `WbContactProperties`, as it has a similar case as `WbMotor`.

PR #5133, opened 2022-08-29
Fix apply skyColor to appearance on background updates

### LINKED ISSUE #5105: Changing Skycolor does not update appearance
**Describe the Bug**
When having a simple `Background` node (not a `TextureBackground`) and an object with a `PBRAppearance`, the object will have some reflections that correspond to the `skyColor` field of the `Background`. However, when changing this color, the reflections do not change until save + reload.

**Steps to Reproduce**
1. Create a new empty world.
2. Insert a `Background` node.
3. Insert a `Tiago++` robot. The `Tiago++` will have blue reflections (as the default `skyColor` is blue).
4. Change the `skyColor` to yellow.
5. See that the `Tiago++` still has blue reflections.
6. Save and reload: now the `Tiago++` has yellow reflections.

**Expected behavior**
The reflections of objects should immediately be updated.

**Additional context**
Already present in R2021b

BenjaminDeleze:
Fix #5105

### DIFF
--- a/docs/reference/changelog-r2022.md
+++ b/docs/reference/changelog-r2022.md
@@ -79,6 +79,7 @@ Released on July, 7th, 2022.
     - Fixed unwanted position jump when moving [Solid](solid.md) objects by dragging them in the 3D Scene while holding the SHIFT key ([#5080](https://github.com/cyberbotics/webots/pull/5080)).
     - Fixed drag events when using the orthographic projection mode ([#5080](https://github.com/cyberbotics/webots/pull/5080)).
     - Fixed "show resize handles" check box status in the Field Editor ([#5080](https://github.com/cyberbotics/webots/pull/5080)).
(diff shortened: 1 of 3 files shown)


ad-daniel:
Just FYI CadShapes also have instances of PBRAppearances, so if the background changes they too must change. In theory the connections do exist, but we should make sure they adapt on-the-fly as well

BenjaminDeleze:
It should work fine, but I will check

BenjaminDeleze:
> Just FYI, CadShapes also have instances of PBRAppearances, so if the background changes they too must change. In theory the connections do exist, but we should make sure they adapt on-the-fly as well

It works

BenjaminDeleze:
The test that fails is `FAILURE with supervisor_import_robot_node: The robot node has not been imported correctly
` and is not related to this PR

ygoumaz (approved):
Works nicely, thank you.

PR #5134, opened 2022-08-29
Fix conversion to `<generic>` and warn user

ygoumaz:
The conversion to `<generic>` controllers and robot-windows when opening a remote Robot PROTO in the text editor is only working in distributions, but not in the development environment. Moreover, relative and `webots://` URLs in remote PROTOs were not correctly converted when editing them locally from the development environment.

This PR fixes these issues, and adds `INFO` logs in the console to warn the user that `controller` and `window` fields have been updated.

### DIFF
--- a/src/webots/gui/WbMainWindow.cpp
+++ b/src/webots/gui/WbMainWindow.cpp
@@ -2352,7 +2352,7 @@ void WbMainWindow::openFileInTextEditor(const QString &fileName, bool modify, bo
           const int commentIndex = line.indexOf("#");
           const int start = binaryMatch.capturedStart(2);
           const int length = binaryMatch.capturedEnd(2) - start;
-          QString replacement = "<generic>";
+          const QString replacement = "<generic>";
           line.replace(start, length, replacement);
           // adjust comment indentation
           const int offset = commentIndex - line.indexOf("#");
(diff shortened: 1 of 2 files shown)


src/webots/gui/WbMainWindow.cpp
```diff
@@ -2315,61 +2315,64 @@ void WbMainWindow::openFileInTextEditor(const QString &fileName, bool modify, bo
...
-              const int n = (end < commentIndex) ? -offset : commentIndex - offset - end - 1;
-              if (n > 0)
-                line.replace(end, n, "");
-            }
+        const QRegularExpressionMatch binaryMatch = binary.match(line);
+        if (binaryMatch.hasMatch()) {
+          const int commentIndex = line.indexOf("#");
+          const int start = binaryMatch.capturedStart(2);
+          const int length = binaryMatch.capturedEnd(2) - start;
+          QString replacement = "<generic>";
```
omichel, line 2355:
```suggestion
          const QString replacement = "<generic>";
```

omichel (changes requested):
Looks good to me.
There is only one `const` missing (but was already missing in previous code).

omichel (approved):
Thank you.

PR #5136, opened 2022-08-29
Fix urls that are send to webotsJS with a space

BenjaminDeleze:
When streaming assets that have spaces in their path (either in some folder name or directly in their filename), the urls are automatically converted by `fetch` such that all spaces are replaced by `%20`.

However, in Webots we need to convert the `%20` back to normal spaces otherwise Webots will not found the files.

### DIFF
--- a/src/webots/gui/WbX3dStreamingServer.cpp
+++ b/src/webots/gui/WbX3dStreamingServer.cpp
@@ -66,7 +66,7 @@ void WbX3dStreamingServer::create(int port) {
 void WbX3dStreamingServer::sendTcpRequestReply(const QString &url, const QString &etag, const QString &host,
                                                QTcpSocket *socket) {
   QString urlWithSpaces(url);
-  urlWithSpaces = urlWithSpaces.replace("%20", " ");
+  urlWithSpaces.replace("%20", " ");
   QFileInfo file(WbProject::current()->dir().absolutePath() + "/" + urlWithSpaces);
   if (file.exists())
     socket->write(WbHttpReply::forgeFileReply(file.absoluteFilePath(), etag, host, urlWithSpaces));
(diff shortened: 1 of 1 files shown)


src/webots/gui/WbX3dStreamingServer.cpp
```diff
@@ -65,9 +65,11 @@ void WbX3dStreamingServer::create(int port) {
 
 void WbX3dStreamingServer::sendTcpRequestReply(const QString &url, const QString &etag, const QString &host,
                                                QTcpSocket *socket) {
-  QFileInfo file(WbProject::current()->dir().absolutePath() + "/" + url);
+  QString urlWithSpaces(url);
+  urlWithSpaces = urlWithSpaces.replace("%20", " ");
```
omichel, line 69:
```suggestion
  urlWithSpaces.replace("%20", " ");
```

omichel (changes requested):

src/webots/gui/WbX3dStreamingServer.cpp
```diff
@@ -65,9 +65,11 @@ void WbX3dStreamingServer::create(int port) {
 
 void WbX3dStreamingServer::sendTcpRequestReply(const QString &url, const QString &etag, const QString &host,
                                                QTcpSocket *socket) {
-  QFileInfo file(WbProject::current()->dir().absolutePath() + "/" + url);
+  QString urlWithSpaces(url);
+  urlWithSpaces.replace("%20", " ");
```
omichel, line 69: `%20` is only an example of percent encoding, but there are many more. To be generic, we should probably use:
```c++
const QString decodedUrl = QUrl::fromPercentEncoding(url.toUtf8());
```

omichel (changes requested):

omichel (approved):
Thank you.

PR #5138, opened 2022-08-30
Fix copy of relative texture URLs during project relcoation

### LINKED ISSUE #5131: Textures copied outside current project during relocation
When relocating a world that is using relative URLs for textures, a corresponding folder is created during project relocation at the same path but relative to the new project location.

This is fundamentally wrong and Webots should never modify or copy anything on the system outside of the location chosen by the user.

**Steps to reproduce**
1. Open sample `track.wbt`
2. Add a texture URL pointing to a relative path to the node `DEF TRACKED_ROBOT Robot > children > DEF GEOM Shape > appearance`: for example "../../../default/worlds/textures/square_block_parquetry.jpg" (path defined by Webots when choosing a default texture from the Select File dialog)
3. Save the world and accept the project relocation: for example at `~/Desktop`
-> the project relocation also creates the folder `~/default/worlds/textures/` containing `square_block_parquetry.jpg

stefaniapedrazzi:
Fix #5131: do not copy textures outside the new project folder chosen by the user.
If a world uses relative URLs:
* if they are located inside the new project path: keep the relative URL and copy the texture if needed
* otherwise if the relative path is no longer valid: convert the relative path to the corresponding absolute path

### DIFF
--- a/src/webots/core/WbFileUtil.cpp
+++ b/src/webots/core/WbFileUtil.cpp
@@ -148,11 +148,13 @@ bool WbFileUtil::areIdenticalFiles(const QString &fileAPath, const QString &file
 }
 
 bool WbFileUtil::isLocatedInDirectory(const QString &file, const QString &directory) {
+  return QFileInfo(file).absoluteFilePath().startsWith(QFileInfo(directory).absoluteFilePath(),
 #ifdef _WIN32
-  return file.startsWith(directory, Qt::CaseInsensitive);
+                                                       Qt::CaseInsensitive);
 #else
-  return file.startsWith(directory, Qt::CaseSensitive);
+                                                       Qt::CaseSensitive
 #endif
+  );
 }
 
 bool WbFileUtil::isLocatedInInstallationDirectory(const QString &file) {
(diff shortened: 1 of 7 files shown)


ad-daniel:
Sorry only noticed now you started working on this issue. Yesterday we discussed about this internally a bit, and the best solution to us seems to be to create a local folder where all the textures are copied into (and if name clashes exist, to add suffixes to distinguish them). Basically it's the same principle we already use for animations (and much of the logic exists already). To us it seemed a more portable solution compared to having absolute paths

stefaniapedrazzi:
Ok. Se then I will let you implement it.

stefaniapedrazzi:
However, I personally think that copying textures during project relocation is useless and it will just increase the disk space used by Webots projects.

omichel:
> However, I personally think that copying textures during project relocation is useless and it will just increase the disk space used by Webots projects.

Yes, but using absolute paths is not a good solution either as the project won't be easy to share with others.

ad-daniel:
I've continued the discussion in the issue https://github.com/cyberbotics/webots/issues/5131

PR #5139, opened 2022-08-30
Sanitize PBR fields at creation

BenjaminDeleze:
**Description**
The fields of `PBRApperance` were not sanitized at their creation, only on updates.
So if a field was set to an unsupported value either in:
- a `.proto` or  `.wbt` file.
- in the `sceneTree` of Webots and it was triggering a proto regeneration.

This can lead to surprising graphical results, (try setting the transparency to 5 for example).

This PR fixes that by calling the corresponding `WbFieldChecker` function on each field in the `createWrenObjects` function, like the geometries do.


omichel (approved):
Thank you.

PR #5140, opened 2022-08-30
Fix supervisor import robot node test

omichel:
The `supervisor_import_robot_node` test was failing randomly due to an unfinished string (missing the final `\0`).


BenjaminDeleze (approved):
Thank you

Issue #4111, opened 2022-01-12
"Front View" should be "South View"


Thomas-Feldmeier:
Load sample world "pedestrian", select green pedestrian

 view > change view > front view (alt + 2)

The view is always from south to north. The view is not of the front side of the pedestrian.

![pedestrian](https://user-images.githubusercontent.com/73528085/149146690-6a2e6372-ece1-41b7-952b-ffc9e2e9ec42.png)

Rename
"Front View" -> "South View" (from -y to +y of world EastNorthUp (ENU) coordinate system)
"Back View" -> "North View"
"Left View" -> "West View"
"Right View" -> "East View"

Maybe add a true "Front View" (from +x to -x relative to selected object) etc.

ad-daniel:
Additionally, I think the shortcuts should probably be re-set.

![aa](https://user-images.githubusercontent.com/44834743/149183035-3f6d8362-5f57-4e8b-ad40-bcee970354cc.png)

There isn't much logic behind it currently (to my knowledge), and considering things like ALT + 5 (move viewpoint to object) and ALT + 3 (does nothing) are scattered among them, the overall behavior seems a bit random.

A fix could be setting ALT + 1 to ALT + 6 just for views, or there might be a better mapping. We can perhaps check how it's done in other software.

ShuffleWire:
I think you could use what Blender use, as long as I remember, it was very convenient

stefaniapedrazzi:
I just checked what is done in Blender (I have version 3.0.1):
* Viewpoint standard views:
    `Top`/`Bottom`/`Front`/`Back`/`Front`/`Left views match exactly what we implemented in Webots -> same name and same behavior
* Additionally Blender has a menu item "Align View" > "Align View To Active" with again the 6 options (Top/Bottom/Front/Back/Right/Left)

omichel:
Yes, but blender doesn't have the ENU concepts we have in Webots `WorldInfo.coordinateSystem`, so the cardinal directions seems to be a good alternative to the front/back, left/right terminology for Webots.

PR #5149, opened 2022-08-31
Enhancement viewpoint standard views

stefaniapedrazzi:
Addresses  #4111:
* [x] Rename Viewpoint standard views
* [x] Add actions to move to default standard view related to the selected object
* [x] Improve keyboard shortcuts
* [x] Add ChangeLog entries
* [x] Update documentation


stefaniapedrazzi:
It remains to update the `context_menu.png` documentation image from Windows, but otherwise this PR is ready for review.

stefaniapedrazzi:
Finally, I completely replaced the screenshot with an updated one taken on Linux.

ad-daniel (changes requested):
Overall seems to work well, the only thing I find a bit odd is that when aligning to a face, no "rectification" takes place, what I mean is:

1. Open `skin_animated_humans.wbt`
2. Select robert and `ALT + SHIFT + 2`
3. The view is indeed correct and facing Robert, but it's slanted rather than actually face-on

To make it more evident you can reload, move the camera such that you're looking at Robert from above, and then press `ALT + SHIFT + 2`. Again it's not "wrong" but the slant is quite big

If you reload and do `ALT + 2`, the movement of the camera seems to be a combination of three steps and the last of the 3 takes care of re-aligning with the axes. Is it possible that the object-focused orbits are missing the last step?

stefaniapedrazzi:
I will check because I probably introduce it when cleaning up the code.

stefaniapedrazzi:
It should be fixed now.

ad-daniel:
Perfect!
I might however have found an edge case. It doesn't always happen, so not sure exactly what triggers it, either way I have it often when I:
- load same world: `SHIFT + ALT + 2`
- when done `SHIFT + ALT + 3`
- when done `SHIFT + ALT + 3` yet again.

Although you're already facing from the back, it still triggers an animation and gets back to the same spot. Not the end of the world but doesn't seem to happen in the other cases. It seems like you NEED to move from FRONT to BACK for it to happen, if you get to the BACK from any other direction it doesn't.

stefaniapedrazzi:
I improved the comparison of quaternions and it should solve the problem with the orbit to the same view.
I can no longer reproduce it.

ad-daniel:
Still happens to me unfortunately

stefaniapedrazzi:
Yes, I will commit very a soon a fix.
But I found another orbit behavior exception that I'm fixing too.

ad-daniel:
I don't know if you're still investigating, but I just wanted to say that the last commit solved all issues for me

stefaniapedrazzi:
Perfect!
There is still a weird orbit when moving from back to left view of the human.
It seems a singularity somewhere but I didn't find it yet.

stefaniapedrazzi:
I'm still testing different worlds, but it now seems to work properly.

ad-daniel (approved):
Works very well for me too, thank you!

Issue #5150, opened 2022-08-31
Long Importable EXTERNPROTO panels affect other panels


ad-daniel:
**Describe the Bug**

1. Open a world with many Importable EXTERNPROTO, for an extreme case open `icon_studio.wbt`
2. Select a random field, for instance `WorldInfo > gravity`
3. Scroll for 2 minutes until you find the widgets related to the gravity field panel

The scroll area used to display the Importable EXTERNPROTO seems to remain when changing panel

PR #5151, opened 2022-08-31
Fix Importable EXTERNPROTO panel size

ad-daniel:
**Description**

Fixes https://github.com/cyberbotics/webots/issues/5150.


ygoumaz (commented):
It works fine now ! Just a question: what is the reason to change the signal from `externProtoListChanged` to `worldLoadCompleted` to call  `updateContents()`? Both code fundamentally do the same, which is to call `updateContents()` when loading a world.

ad-daniel:
Basically we don't want to update the contents every time a world is loaded otherwise we would fill the panel even when something else is selected in the scene tree which results in the bug mentioned. For this reason I've removed the `emit externProtoListChanged` during the world load so that it doesn't get filled. However, there is one situation where we do want to update the contents after a world load and is if the EXTERNPROTO panel happened to be open already when we switch world (or do a reload). In this scenario we need to refresh the contents or it would show up as empty and/or outdated info (list from the previous world).

Issue #5148, opened 2022-08-31
No PROTO regeneration in highly-nested PROTOs


ygoumaz:
Changing the URL of a resource in a highly-nested PROTO doesn't trigger the regeneration. A similar problem was reported in [#5092 (comment)](https://github.com/cyberbotics/webots/pull/5092#pullrequestreview-1080487653).

How to reproduce:
1. Open Webots and add an official LincolnMKZ.
2. In one of the slots (e.g. sensorsSlotCenter), add an official TruckSimple.
3. In one of the slots of the TruckSimple (e.g. sensorsSlotCenter), add an official LincolnMKZ.
4. Now change the URL of the plate in the Lincoln added in step 3 to something invalid. The PROTO is not regenerated and no missing texture is displayed.
5. If you apply the same URL update to the Lincoln of step 1, the behavior is correct.

ad-daniel:
The problem appears to be present since R2022a at least, so it isn't a (recent) regression.

However there are some oddities:
- Doing `Lincoln > TruckSimple > Lincoln` behaves as mentioned above. However the same structure can't be created with `Truck` instead of `TruckSimple`, after inserting Truck, no node shows up in the add-node window. This indicates that there are no compatible nodes for this slot, which is odd since the TruckSimple let's you do it.
- The `plate` field of the innermost Lincoln doesn't show a "select" button in order to pick a texture whereas the upper does does. This button is hidden in the following conditions:

`showSelectButton = mesh || sound || texture || !regular || (solidReference && !protoParameter) || (fluidName && !protoParameter) || (referenceArea && !protoParameter)`

The values for which it fails is `texture` and `!regular` (these are true for first Lincoln). Might be a clue.

ad-daniel:
The missing button is not related to the issue. It shows the button only if the parameter name contains the world "url".

PR #5157, opened 2022-09-01
Fix alias not propagating in derived proto and empty add-node

ad-daniel:
**Description**

Fixes: https://github.com/cyberbotics/webots/issues/5148 (both issues)

Due to the lack of a link between field and parameter in a derived PROTO context (without derivation everything works), changes to the parameter didn't propagate (first issue) and it wasn't possible to insert the second Lincoln in a slot of a Truck (second issue, add-node showed up empty whereas it was possible to insert it in a TruckSimple).

Intuitively it seems wrong to me that the alias connection is made only if the parameter belongs to the base PROTO. In the Lincoln case for example, the parameter `plate` is indeed a parameter exclusive to the derived PROTO (Lincoln), not the base ones (Car, Ackermann).

So far I haven't encountered side effects but I'll keep investigating


omichel (approved):
Thank you.
The patch seems good to me and everything seems to work nicely.

Issue #5161, opened 2022-09-02
Noisemask sanity check is done on the cached file rather than the original url


ad-daniel:
**Describe the Bug**

In the guided tour on a distrib I get this error:
`Robot > Camera: Invalid URL '/home/daniel/.cache/Cyberbotics/Webots/assets/d31aa5ff062e9d66bcc4af69d35d1370f309c80c'. The noise mask must be in '.jpeg', '.jpg' or '.png' format.`

Likely introduced here: https://github.com/cyberbotics/webots/pull/5132

PR #5162, opened 2022-09-02
Fix noise mask on distrib

ad-daniel:
**Description**

Fixes: https://github.com/cyberbotics/webots/issues/5161


ygoumaz (commented):
There is an additional occurence of `noiseMaskUrl` at https://github.com/cyberbotics/webots/blob/a876f82fa5edaabf4b15ddfc0e7b87d1136524ea/src/webots/wren/WbWrenCamera.cpp#L368

I believe it should also be replaced by `noiseMaskPath`.

ad-daniel:
Should be addressed now

ygoumaz (approved):
Thank you.

Issue #5164, opened 2022-09-02
Missing relative textures when created from PROTO wizard


ad-daniel:
**Describe the Bug**

There's two separate issues:

1. Create a `CustomBed` from a `Bed` using the PROTO wizard (and select all the fields so the derived one exposes the same parameters as the base PROTO)
   - Since we selected all the fields, a field `field SFNode blanketAppearance PBRAppearance { baseColorMap ImageTexture { url "Bed/textures/duvet.jpg" } }` is created for `CustomBed` too. Now of course this URL is indeed relative to the original Bed, but isn't to the custom locally available one, so the texture will be missing. When creating a PROTO using the Wizard, if the PROTO exposes assets, we should download them to the local folder.

2. Create a `CustomBed` from a `Bed` using the PROTO wizard (and leaving the default fields selected, namely translation, rotation and name)
   - Since now the appearance field is not overwritten by the derived PROTO, the relative URL needs to be resolved with respect to the remote Bed PROTO (base PROTO), which it appears to do, except it uses the disk file rather than the `http` URL: `Unable to find resource at '/home/testr2022b/snap/webots/common/.cache/Cyberbotics/Webots/assets/Bed/textures/duvet.jpg'.`

omichel:
> When creating a PROTO using the Wizard, if the PROTO exposes assets, we should download them to the local folder.

Or we could use the web URL instead of the local one.

ad-daniel:
A third issue surfaced, if the PROTO from which we derive references additional PROTO as parameters, these need to be declared. To know what these declarations should look like, we open the derived PROTO and copy the declarations from there. This process already takes place however if the cache is cleaned and we're on a distrib, when attempting to open the reference PROTO to search for the declarations webots crashes since this file isn't locally available (isn't cached).

I'll fix the issue in #5169 as well.

PR #5169, opened 2022-09-02
Fix issues between relative URL and PROTO wizard

ad-daniel:
**Description**

Fixes: https://github.com/cyberbotics/webots/issues/5164

- [x] Exposed parameter in derived PROTO refers to relative URL
- [x] Scope refers to disk file
- [x] Declaration of non-cached parameters of type PROTO results in a crash

### DIFF
--- a/src/webots/gui/WbNewProtoWizard.cpp
+++ b/src/webots/gui/WbNewProtoWizard.cpp
@@ -203,9 +203,7 @@ bool WbNewProtoWizard::generateProto() {
       }
 
       // if the interface refers to a relative url, turn it into a web one
-      const QRegularExpression resources("^\"([^\"]*)\\.(jpe?g|png|hdr|obj|stl|dae|wav|mp3)$\"",
-                                         QRegularExpression::CaseInsensitiveOption);
-      QRegularExpressionMatchIterator it = resources.globalMatch(interface);
+      QRegularExpressionMatchIterator it = WbUrl::vrmlResourceRegex().globalMatch(interface);
       while (it.hasNext()) {
         const QRegularExpressionMatch match = it.next();
         if (match.hasMatch()) {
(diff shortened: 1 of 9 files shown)


src/webots/gui/WbNewProtoWizard.cpp
```diff
@@ -97,151 +98,185 @@ bool WbNewProtoWizard::validateCurrentPage() {
...
+      QRegularExpressionMatchIterator it = resources.globalMatch(interface);
+      while (it.hasNext()) {
+        const QRegularExpressionMatch match = it.next();
+        if (match.hasMatch()) {
+          QString asset = match.captured(0);
+          asset.replace("\"", "");
+          if (!WbUrl::isWeb(asset) && QDir::isRelativePath(asset))
+            interface.replace(QString("\"%1\"").arg(asset), QString("\"%1\"").arg(WbUrl::combinePaths(asset, info->url())));
+        }
+      }
```
ad-daniel, line 217: note: the diff is big but it's because I've removed most of the lines from the body of a if by testing `if (!success)` rather than `if (success)`. L205-218 is the only actual addition done

src/webots/gui/WbNewProtoWizard.cpp
```diff
@@ -97,151 +98,185 @@ bool WbNewProtoWizard::validateCurrentPage() {
...
+        const QRegularExpressionMatch match = it.next();
+        if (match.hasMatch()) {
+          QString asset = match.captured(0);
+          asset.replace("\"", "");
+          if (!WbUrl::isWeb(asset) && QDir::isRelativePath(asset)) {
+            QString newUrl = QString("\"%1\"").arg(WbUrl::combinePaths(asset, info->url()));
+            interface.replace(QString("\"%1\"").arg(asset), newUrl.replace(WbStandardPaths::webotsHomePath(), "webots://"));
+          }
+        }
+      }
```
ad-daniel, line 217: note: the diff is big but it's because I've removed most of the lines from the body of an if by testing if (!success) rather than if (success). L205-218 is the only actual addition done

omichel (approved):
Looks good to me and worked out for all the cases I tested.

PR #5170, opened 2022-09-05
Add missing declaration in osm_importer

ad-daniel:
**Description**

When roads are written, so are RoadLines which currently aren't declared and therefore show a warning


omichel (approved):
Thank you.

PR #5171, opened 2022-09-05
Added scale update in animations

omichel:
This minor feature is needed for the SimGait project as I need to scale non-uniformly the shape of the muscles.

Although the scale of a Transform should never change while a simulation is running (except if a supervisor changes it explicitly), I needed this feature in my [conversion script](https://github.com/cyberbotics/scone-playback/blob/main/convert.py) that produces Webots animations from OpenSim model and motion files.


ad-daniel (approved):
Thank you

Issue #5173, opened 2022-09-06
Convert root to base-node broken with relative URL


ad-daniel:
**Describe the Bug**

In a distrib, when the conversion takes place, the relative URLs that suddenly get exposed at the world level need to be converted to remote ones.
When converting EVERYTHING to basenode, the process works as intended and they are turned to remote ones.

There's two ways to go about it:
- Simpler but not ideal: do the manipulation manually in the convert function as we do for the PROTO wizard. 
- ~~Better: figure out why certain nodes/urls don't get converted to remote url whereas they do for a full-basenode conversion~~

stefaniapedrazzi:
This issue is referring to images URLS etc or PROTO URLs?
Because I also experienced a couple of different issues with PROTO URLs after converting root to base nodes the `Tractor` PROTO in `boomer.wbt` after relocation:
* "TractorFrontWheel" not found after conversion if "Tractor" PROTO URL is remote
* "ERROR: Missing declaration for 'SickLms291'" after conversion, save and revert if "Tractor" PROTO URL is local and relative

ad-daniel:
It's not related but I've noticed some too on the Lincoln, for which I have a patch already, I'll test if that fixes the boomer too.
So, to reproduce it exactly you:
1. open official boomer
2. do change to trigger relocation message and do so
3. convert tractor's root to basenode?

stefaniapedrazzi:
Yes.
Then depending if the Tractor PROTO is local (for example edit the name to trigger the local copy of the PROTO)  or the default remote one I get one or the other error.

PR #5175, opened 2022-09-06
Fix convert root to basenode and other conversion issues

ad-daniel:
**Description**

Fixes: https://github.com/cyberbotics/webots/issues/5173

Also fixes 2 issues with the conversion:

- If slots with PROTO nodes were present prior to the conversion, when deleting the node (prior to spawning the converted version) the declaration of these slots would be removed and they weren't re-added afterwards
- In the process of recursively exposing the declarations needed when converting derived PROTO to base-node, it could happen that the parent file was a cached file, and its disk location (rather than the original URL) was used when determining the full URL of a relatively-defined PROTO.

Concerning the relative URL appearing after conversion:

It turns out that solution 2 is not really an option, or at least not without rethinking the writer logic from the ground up. The reason is that when converting root to base node, as the name implies, the writer does not progress through nodes which are themselves PROTO (nested PROTO). Since it doesn't progress through it, the corresponding ImageTextures will be skipped. After the conversion however, the visibility will go up one layer and all of a sudden they are world-visible and therefore their URL needs to be converted nonetheless, so solution 1 albeit less optimal, seems to be the only approach.

In the case of the Lincoln for example, `textures/lincoln_mkz_details_base_color.jpg` is initially internal to the Lincoln, but child to `VehicleLights`. After root conversion, the Lincoln disappears but `VehicleLights` remains as a PROTO and therefore doesn't go through the writer. After conversion however, `VehicleLights` went up one layer in visibility, from internal it becomes world-visible, and its URLs therefore cannot remain relative and must be changed regardless.

### DIFF
--- a/src/webots/core/WbProject.cpp
+++ b/src/webots/core/WbProject.cpp
@@ -250,12 +250,7 @@ bool WbProject::createNewProjectFolders() {
 }
 
 bool WbProject::isReadOnly() const {
-#ifdef _WIN32
-  Qt::CaseSensitivity caseSensitivity = Qt::CaseInsensitive;
-#else
-  Qt::CaseSensitivity caseSensitivity = Qt::CaseSensitive;
-#endif
-  return mPath.startsWith(WbStandardPaths::webotsHomePath(), caseSensitivity);
+  return WbFileUtil::isLocatedInDirectory(mPath, WbStandardPaths::webotsHomePath());
 }
 
 QString WbProject::controllerPathFromDir(const QString &dirPath) {
--- a/src/webots/core/WbProject.cpp
+++ b/src/webots/core/WbProject.cpp
@@ -250,7 +250,7 @@ bool WbProject::createNewProjectFolders() {
 }
 
 bool WbProject::isReadOnly() const {
-  return WbFileUtil::isLocatedInDirectory(mPath, WbStandardPaths::webotsHomePath());
+  return WbFileUtil::isLocatedInInstallationDirectory(mPath, true);
 }
 
 QString WbProject::controllerPathFromDir(const QString &dirPath) {
--- a/src/webots/nodes/utils/WbUrl.cpp
+++ b/src/webots/nodes/utils/WbUrl.cpp
@@ -286,6 +286,8 @@ QString WbUrl::combinePaths(const QString &rawUrl, const QString &rawParentUrl) 
   }
 
   if (QDir::isRelativePath(url)) {
+    if (parentUrl.startsWith(WbStandardPaths::cachedAssetsPath()))
+      parentUrl = WbNetwork::instance()->getUrlFromEphemeralCache(parentUrl);
     // if it is not available in those folders, infer the URL based on the parent's url
     if (WbUrl::isWeb(parentUrl) || QDir::isAbsolutePath(parentUrl) || WbUrl::isLocalUrl(parentUrl)) {
       // remove filename from parent url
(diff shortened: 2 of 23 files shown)


src/webots/nodes/utils/WbUrl.cpp
```diff
@@ -286,6 +286,8 @@ QString WbUrl::combinePaths(const QString &rawUrl, const QString &rawParentUrl)
   }
 
   if (QDir::isRelativePath(url)) {
+    if (parentUrl.startsWith(WbStandardPaths::cachedAssetsPath()))
```
stefaniapedrazzi, line 289: It would be more consistent to use
```suggestion
    if (WbFileUtil::isLocatedInDirectory(parentUrl, WbStandardPaths::cachedAssetsPath()))
```

src/webots/vrml/WbNode.cpp
```diff
@@ -1246,6 +1246,16 @@ void WbNode::exportExternalSubProto(WbWriter &writer) const {
...
 
+  // find all proto that were already exposed prior to converting the root (typically, slots with world visibility)
+  const QList<const WbNode *> protos = WbNodeUtilities::protoNodesInWorldFile(this);
+  foreach (const WbNode *p, protos)
+    // the node itself doesn't need to be re-declared since it won't exist after conversion
+    if (p != this) {
+      const QString protoDeclaration = WbProtoManager::instance()->externProtoUrl(p, false);
+      assert(!protoDeclaration.isEmpty());  // since the proto has world-visibility, a declaration for it must exist
+      writer.trackDeclaration(p->modelName(), protoDeclaration);
+    }
```
stefaniapedrazzi, line 1257: Here it would be better to add the brackets to the IF block given that it is a multiline content
```suggestion
  foreach (const WbNode *p, protos) {
    // the node itself doesn't need to be re-declared since it won't exist after conversion
    if (p != this) {
      const QString protoDeclaration = WbProtoManager::instance()->externProtoUrl(p, false);
      assert(!protoDeclaration.isEmpty());  // since the proto has world-visibility, a declaration for it must exist
      writer.trackDeclaration(p->modelName(), protoDeclaration);
    }
  }
```

stefaniapedrazzi (changes requested):
Seems to workd fine but I'm creating a local package to test it properly but I could not do it today.

stefaniapedrazzi:
I finally could test the behavior and all the issues I reported are correctly fixed and works fine.

ad-daniel:
Thank you, I've taken the opportunity to apply the same fix in other similar situations

src/webots/core/WbProject.cpp
```diff
@@ -250,12 +250,7 @@ bool WbProject::createNewProjectFolders() {
 }
 
 bool WbProject::isReadOnly() const {
-#ifdef _WIN32
-  Qt::CaseSensitivity caseSensitivity = Qt::CaseInsensitive;
-#else
-  Qt::CaseSensitivity caseSensitivity = Qt::CaseSensitive;
-#endif
-  return mPath.startsWith(WbStandardPaths::webotsHomePath(), caseSensitivity);
+  return WbFileUtil::isLocatedInDirectory(mPath, WbStandardPaths::webotsHomePath());
```
omichel, line 253:
```suggestion
  return WbFileUtil::isLocatedInInstallationDirectory(mPath);
```
To generalize this replacement, we should probably add a boolean argument to WbFileUtil::isLocatedInInstallationDirectory to ignore the test on `WEBOTS_ALLOW_MODIFY_INSTALLATION` if not wanted, but in this case, I believe we can call this method as is.

ad-daniel: I'm not sure it's a good idea to leave it as-is, as it might vary the behavior significantly. All it takes is for you to have, at any given time, installed the development environment and this variable will exist in your bashrc. This specific case is used (I think) for example to determined the behavior when you right-click a PROTO and which options should be there (view and/or edit). We can do the change you mention (calling isLocatedInInstallationDirectory instead of isLocatedInDirectory), but it feels like altering the logic (i.e., checking also for WEBOTS_ALLOW_MODIFY_INSTALLATION) now is a bit risky

omichel (approved):
That looks good to me.
Thank you.

omichel (approved):
Nice.

PR #5179, opened 2022-09-07
Fix project relocation dialog layout

ygoumaz:
The layout of the project relocation dialog is currently broken on macOS. The `WbLineEdit` for the target path has no minimum width which results in a very small edition area on macOS. The line edit doesn't take the complete width of the window. I can't reproduce the issue on Linux, and the fix doesn't seem to break anything on Linux. I haven't tested on Windows.


omichel (approved):
I tested it on Windows, and everything looks good there.

src/webots/editor/WbProjectRelocationDialog.cpp
```diff
@@ -91,6 +91,7 @@ void WbProjectRelocationDialog::initCompleteRelocation() {
   mSourceEdit->setMinimumWidth(sourcePath.length() * 8);
 
   mTargetEdit = new WbLineEdit(mTargetPath, this);
+  mTargetEdit->setMinimumWidth(mTargetPath.length() * 8);
```
omichel, line 94: I would be better this way:
```suggestion
#ifdef __APPLE__
  mTargetEdit->setMinimumWidth(mTargetPath.length() * 8);
#endif
```
So that we understand the problem is specific to macOS.

ygoumaz:
The PR is again ready for review

omichel (approved):
Thank you.

Issue #5183, opened 2022-09-07
Animation/Scene containing CadShape with textures does not work


BenjaminDeleze:
When exporting a world to an Animation/Scene, all the textures are put in a `textures` folder and the urls of the world are changed such that it worked. However, in the case of `CadShape`, we do not change the urls of the `.dae` or `.mtl` and so the path are broken.

PR #5185, opened 2022-09-07
Fix `CadShape` animations

ad-daniel:
**Description**

Fixes: https://github.com/cyberbotics/webots/issues/5183


ad-daniel:
It also fixes the fact that when creating animations of worlds that include PROTO-derived backgrounds, intermediary folders were created in `textures`.

PS: if you use the repo I created last time for testing, you should pull. I've changed the name of the folders like `textures -> tex` to reduce the chance that it works even when it shouldn't

BenjaminDeleze (approved):
Thank you

PR #5187, opened 2022-09-07
Improved extern controller documentation

omichel:
It is more convenient (and it works) to set environment variables in the controller programs.


ygoumaz (changes requested):
Thank you. However, this new documentation and corrections should probably be extended to Linux and macOS tabs.

omichel:
I moved this information out of the OS tab.

ygoumaz (approved):
Thank you, looks better now !

PR #5193, opened 2022-09-08
Fix field editor size update

### LINKED ISSUE #5189: Field editor area too big for content
On the master branch the field editor is too big and it is not resized to content.
This is really bad because lot of space is added between the elements and very often you also need to scroll to view and used all the fields when there will be enough space to display all the fields without the scrollbar.
I'm on Ubuntu 22.04.

Screenshot from 2022-09-08 10-11-28

stefaniapedrazzi:
Fix #5189:
the `WbExternProtoEditor` content was updated when modifying a new PROTO to the externproto list, even if the editor was not visible, and this was increasing the size of all the field editors.
Given the latest changes in the IMPORTABLE vs other EXTERNPROTO declarations, the importable list can currently only be changed by `WbExternProtoEditor` itself and it is not modified by other user actions.
Thus, we can simplify the code and completely remove the notification when the list of EXTERNPROTO changes.


ad-daniel (approved):
That's a nice cleanup, thank you

Issue #5188, opened 2022-09-08
Webots crashes undoing template parameter change


stefaniapedrazzi:
Steps to reproduce:
1. Add "PicketFence" PROTO to any world
2. Increase "numberOfSegments"
3. Type keyboard shortcut "Ctrl+Z" (undo)

```
Thread 1 "webots-bin" received signal SIGSEGV, Segmentation fault.
0x0000555555ef0eec in WbEditCommand::resetValue (this=0x55557005f210, newValue=...) at user_commands/WbEditCommand.cpp:62
62	  switch (mFieldValue->type()) {
(gdb) bt
#0  0x0000555555ef0eec in WbEditCommand::resetValue(WbVariant const&) (this=0x55557005f210, newValue=...)
    at user_commands/WbEditCommand.cpp:62
#1  0x0000555555ef0e8b in WbEditCommand::undo() (this=0x55557005f210) at user_commands/WbEditCommand.cpp:54
#2  0x00007ffff747b800 in QUndoStack::undo() () at /home/stefi/develop/webots/lib/webots/libQt6Gui.so.6
#3  0x0000555555e64126 in WbUndoStack::undo() (this=0x5555569097b0) at user_commands/WbUndoStack.cpp:62
#4  0x0000555555b99d1f in WbFieldIntSpinBox::keyPressEvent(QKeyEvent*) (this=0x5555567e5790, event=0x7fffffffd210)
    at scene_tree/WbFieldIntSpinBox.cpp:46
#5  0x00007ffff68252ce in QWidget::event(QEvent*) () at /home/stefi/develop/webots/lib/webots/libQt6Widgets.so.6
```

stefaniapedrazzi:
In R2022a, the undo functionality is not enabled after changing the template parameter.

ad-daniel:
So the desired behavior is to restore how it worked in R2022a or to allow to undo it (ideally without crashing) ?

stefaniapedrazzi:
I think there was a reason why the undo was disabled instead of supporting it.
But if we can support it is for sure better.

PR #5196, opened 2022-09-09
Fix UNDO crash

ad-daniel:
**Description**

Essentially the logic seems to be that actions that result in a template regeneration are not even pushed to the undo stack although the action itself takes place otherwise the scene tree wouldn't be updated. 

The function itself seems rather odd since we set `mClearRequest` to false 2 lines above, but I suspect that the execution of `QUndoStack::push(cmd)` calls `setClearRequest` which changes the flag back to positive, ensuring that we enter the `clear()` function. In short we push it, do the action, and remove it from the stack and since it's no longer there neither the menu option nor CTRL+Z have anything to revert to (i.e., the button remains disabled) .

These two lines appear in the source code of R2022a but no longer do in our repo. I've tried to figure out at which point they were removed but I can't find any indication they ever were, the last change in this function (both on master/develop) dates from 4  years ago, long before R2022a. How can the history have changed without leaving any trace is beyond me.

### DIFF
--- a/src/webots/user_commands/WbUndoStack.cpp
+++ b/src/webots/user_commands/WbUndoStack.cpp
@@ -51,6 +51,9 @@ void WbUndoStack::push(QUndoCommand *cmd) {
 
   QUndoStack::push(cmd);
 
+  if (mClearRequest)
+    clear();
+
   updateActions();
   // notify the scene tree that some fields changed in order to update the
   // field editor if needed
--- a/src/webots/user_commands/WbUndoStack.cpp
+++ b/src/webots/user_commands/WbUndoStack.cpp
@@ -49,7 +49,7 @@ void WbUndoStack::push(QUndoCommand *cmd) {
 
   mClearRequest = false;
 
-  QUndoStack::push(cmd);
+  QUndoStack::push(cmd);  // may change the value of mClearRequest via the clearRequest slot
 
   // cppcheck-suppress knownConditionTrueFalse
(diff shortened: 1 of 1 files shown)


ad-daniel:
Nevermind, it was indeed changed [here](https://github.com/cyberbotics/webots/pull/4202), blame just doesn't mention it. It was removed as part of a cppcheck fix, probably due to the odd behavior of the function I mentioned.

src/webots/user_commands/WbUndoStack.cpp
```diff
@@ -51,6 +51,10 @@ void WbUndoStack::push(QUndoCommand *cmd) {
 
   QUndoStack::push(cmd);
```
omichel, line 52:
```suggestion
  QUndoStack::push(cmd);  // may change the value of mClearRequest via the clearRequest slot
```

omichel (changes requested):
I would add an extra comment to be very clear as this code is very odd.

omichel:
Does this PR aim to fix #5188?

omichel (approved):
Thank you.

ad-daniel:
Well, it restores the behavior of R2022a, but doesn't enable the undo of template regenerators which entails rewriting the undo stack history, so we may keep the issue open since that's a desired feature

omichel:
I would close this issue and open a new one since it is not any more a crash.

PR #5203, opened 2022-09-09
Specify the default value of notify and shareIdleTime

BenjaminDeleze:
I did not put the whole `webots.cloud` address (https://webots.cloud/ajax/server/setup.php) as it is just for information


omichel (approved):
Thank you.

omichel:
We should also report this change [here](https://github.com/cyberbotics/webots-server/blob/main/simulation_server.py#L796-L797).

BenjaminDeleze:
> We should also report this change [here](https://github.com/cyberbotics/webots-server/blob/main/simulation_server.py#L796-L797).

Done here: https://github.com/cyberbotics/webots-server/pull/6

PR #4539, opened 2022-05-17
Feature animals proto

carikaser:
Adding new animal PROTOs

Defining PROTO files for the following animals: cow, horse, dog, cat, rabbit, deer, sheep, fox and farm related objects (barn for now).

PR #5204, opened 2022-09-12
Add animals PROTO to R2022b

stefaniapedrazzi:
Port #4539 to R2022b.

Add just the PROTO models (no documentation or examples) so that they use the R2022b PROTO header version and do not trigger warnings.


ygoumaz (approved):
Thank you !

PR #5206, opened 2022-09-12
Fix incorrect insertion in IMPORTABLE EXTERNPROTO list

ygoumaz:
The clean-up of #5193 removed the wrong boolean in the call to `declareExternProto` in `WbAddNodeDialog`, causing the Add Node button to also add the PROTO to the IMPORTABLE list.


stefaniapedrazzi (approved):
Thank you!

PR #5220, opened 2022-09-14
Fix .deb and .dmg package creation

BenjaminDeleze:
A regex was not correctly detecting if it was a revision or not.


BenjaminDeleze:
The only failing test is the window distrib which is normal

scripts/packaging/generate_asset_cache.py
```diff
@@ -37,9 +37,8 @@
     raise RuntimeError('Error, WEBOTS_HOME variable is not set.')
 
 
-def generate_asset_cache(tag):
-    with open(os.path.join(WEBOTS_HOME, 'resources', 'version.txt'), 'r') as file:
-        folder_name = f'assets-{file.readline().strip()}'
+def generate_asset_cache(tag, version_name):
```
stefaniapedrazzi, line 40: This function is also used at line 79 in this file with just one argument.
It should be fixed accordingly.

scripts/packaging/generic_distro.py
```diff
@@ -54,12 +54,11 @@ def get_webots_version():
...
         with open(os.path.join(os.getenv('WEBOTS_HOME'), 'scripts', 'packaging', 'webots_version.txt'), 'r') as version_file:
-            version = version_file.readline()
-            match = re.search(r'^(.{6,15})( .* (\d+))?', version)
-            global package_version
-            if match.group(2):
-                return match.group(0) + '-rev' + match.group(2)
-            return match.group(0)
+            version = version_file.readline().strip()
+            parts = version.split(' ')
+            if parts[2]:
```
stefaniapedrazzi, line 59: This will throw errors:
```suggestion
            if len(parts) > 2:
```

stefaniapedrazzi (changes requested):
Thank you.
The changes are fine but there are still a couple of issue to solve.

scripts/packaging/generate_asset_cache.py
```diff
@@ -73,8 +72,10 @@ def generate_asset_cache(tag):
...
-    if len(sys.argv) != 2:
-        sys.exit('Missing argument: commit sha or tag.')
+    if len(sys.argv) != 3:
+        sys.exit('''Missing one or more arguments.
+            It should have the form: generate_asset_cache [commit sha or tag] [webots version name]''')
     else:
         tag = sys.argv[1]
-    generate_asset_cache(tag)
+        version_name = sys.argv[2]
+    generate_asset_cache(tag, version_name)
```
stefaniapedrazzi, line 81: Why the webots version cannot be read directly from the file as it was done before?

BenjaminDeleze: It could, as could be the tag, but I think that we never call this script directly in the build process (we call directly the generate_asset_cache function)

So as this function is for debugging purpose and that the first parameter is chosen by the user I thought that he could give the second as well.

stefaniapedrazzi (approved):
Thank you

PR #5224, opened 2022-09-15
Fix lidar rotating head export to x3d

BenjaminDeleze:
Before the rotating head was export side-by-side with the lidar and not _in_ the lidar.

The consequence was that if the lidar has a translation/rotation, it was not applied to the rotating head once exported in x3d.


ygoumaz (approved):
Thank you!

src/webots/nodes/WbLidar.cpp
```diff
@@ -222,14 +222,20 @@ void WbLidar::postPhysicsStep() {
 }
 
 void WbLidar::write(WbWriter &writer) const {
-  if (writer.isWebots())
-    WbBaseNode::write(writer);
-  else {
+  if (writer.isWebots() || writer.isUrdf()) {
```
omichel, line 225: We should remove unneeded curvy brackets here.

PR #5231, opened 2022-09-16
fix WbLightSensor: collect behavior of recieving light of SpotLight

tsubota-kouga:
In `WbLightSensor::computeLightMeasurement`, `R` is directly used from `spotLight->direction()`.
So `WbLightSensor` class does not work correctly when spot light object is moved.

I think `R` should be calculate using `spotLight->upperTransform()->rotation()`.

Thank you for making nice application, and maintenance!


BenjaminDeleze (changes requested):
Thank you for you contribution!
It seems indeed to work better.
May I ask you to add an entry in the [changelog](https://github.com/cyberbotics/webots/blob/master/docs/reference/changelog-r2022.md) under `R2022 revision 1` `bug fixes` ?

BenjaminDeleze (approved):
Thank you!

PR #5236, opened 2022-09-16
Fix labels reset

### LINKED ISSUE #5227: Webots animation: labels not automatically reset
If in the simulation a label is only added during the simulation run, when the animation is reset to timestep 0 the label in the animation is not deleted.

**Steps to Reproduce**
1. Go to 'propeller.wbt'
2. Add ``wb_robot_step(1000);`` at line 43 in `axial_and_trail_rotors.c` and recompile controller (red label)
3. Record and save the animation
4. When running the animation for the first time, the red label is not displayed at time 0, but after running it for the first time the label never disappears at time 0

BenjaminDeleze:
At setup we check if all labels are present in the first frame to have a reference point when jumping backward and forward in the animation.

If it is not the case, create a label instruction in the first frame which set the label to transparent (and empty)

Fix #5227

### DIFF
--- a/resources/web/wwi/Animation.js
+++ b/resources/web/wwi/Animation.js
@@ -40,6 +40,27 @@ export default class Animation {
     this._labelsIds = typeof this.data.labelsIds === 'undefined' ? [] : this.data.labelsIds.split(';')
       .filter(Boolean).map(s => parseInt(s));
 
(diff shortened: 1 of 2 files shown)


stefaniapedrazzi (approved):
Thank you!

PR #5237, opened 2022-09-16
Fix webotsJS follow viewpoint

### LINKED ISSUE #5228: Problem with animation and viewpoint follow
**Describe the Bug**
If the viewpoint follow is activated in animation some issues could happen:

- If you move the viewpoint _before_ starting the animation, there will be a wrong jump of the viewpoint when you start the animation.
- When the animation is on loop, when it starts again, the viewpoint is sometime wrong.

BenjaminDeleze:
Fix #5228 

I tried several things but finally went for a very simple solution:

- If the viewpoint follow is not enabled, we never touch the viewpoint position (except if the user reset it manually).
- If the viewpoint follow is enable, we reset it **but** we preserve the transformations (translation and rotation) done by the user.
  We are force to kind of reset the position otherwise we have some problems with the follow (it does not like big jump).

### DIFF
--- a/docs/reference/changelog-r2022.md
+++ b/docs/reference/changelog-r2022.md
@@ -4,6 +4,7 @@
  - Bug Fixes
    - Fixes the export of lidar's rotating head to X3D. ([#5224](https://github.com/cyberbotics/webots/pull/5224)).
    - Fixed behavior of `WbLightSensor::computeLightMeasurement` when spotlight is rotated ([#5231](https://github.com/cyberbotics/webots/pull/5231)).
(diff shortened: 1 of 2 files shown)


omichel:
Can I test it online? Is it on the beta or testing branch of webots.cloud?

BenjaminDeleze:
You can test it on [testing](https://testing.webots.cloud) or directly https://testing.webots.cloud/Ac9FYE5

omichel (approved):
That works nicely now. Thank you.

PR #5243, opened 2022-09-19
added section to correct corners that are separated by commas instead…

RugnirViking:
… of newlines

**Description**
Encountered an issue in which when upgrading some SimpleBuildings were converted correctly but others appeared flipped. upon investigation it was found that those SimpleBuildings were using comma separated vectors for its corners instead of newline separated like the script expects. This PR goes through and changes any comma-separated lists of corners into newline separated before the file is parsed by the main script

**Additional context**
Persuant to a conversation had in the webots discord on 19th september 2022

### DIFF
--- a/scripts/converter/convert_nue_to_enu_rub_to_flu.py
+++ b/scripts/converter/convert_nue_to_enu_rub_to_flu.py
@@ -187,8 +187,6 @@ def convert_nue_to_enu_world(filename, mode='all', objects_pi=[], objects_pi_2=[
     rotation_next_object = []
     miss_rotation = False
     last_type = ''
-
-
     for line in fileinput.input(filename, inplace=True):
 
         type = [x for x in re.compile('\n|,| ').split(line) if (any(x_char.isalpha() for x_char in x))]
--- a/scripts/converter/convert_nue_to_enu_rub_to_flu.py
+++ b/scripts/converter/convert_nue_to_enu_rub_to_flu.py
@@ -218,7 +218,7 @@ def convert_nue_to_enu_world(filename, mode='all', objects_pi=[], objects_pi_2=[
                 if "," in line:
                     vectors = [x for x in re.compile('[,]').split(line)]
                     for vector in vectors:
-                        print(add_space(line)+vector.strip())
+                        print(add_space(line) + vector.strip())
                 else:
                     print(line, end='')
         else:
--- a/scripts/converter/convert_nue_to_enu_rub_to_flu.py
+++ b/scripts/converter/convert_nue_to_enu_rub_to_flu.py
@@ -187,6 +187,8 @@ def convert_nue_to_enu_world(filename, mode='all', objects_pi=[], objects_pi_2=[
     rotation_next_object = []
     miss_rotation = False
     last_type = ''
+
+
     for line in fileinput.input(filename, inplace=True):
 
         type = [x for x in re.compile('\n|,| ').split(line) if (any(x_char.isalpha() for x_char in x))]
@@ -217,7 +219,7 @@ def convert_nue_to_enu_world(filename, mode='all', objects_pi=[], objects_pi_2=[
                 if "," in line:
                     vectors = [x for x in re.compile('[,]').split(line)]
                     for vector in vectors:
-                        print(vector)
+                        print(add_space(line)+vector.strip())
                 else:
                     print(line, end='')
         else:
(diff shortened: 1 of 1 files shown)


scripts/converter/convert_nue_to_enu_rub_to_flu.py
```diff
@@ -187,6 +187,44 @@ def convert_nue_to_enu_world(filename, mode='all', objects_pi=[], objects_pi_2=[
     rotation_next_object = []
     miss_rotation = False
     last_type = ''
+
+
```
stefaniapedrazzi, line 191:
```suggestion

```

scripts/converter/convert_nue_to_enu_rub_to_flu.py
```diff
@@ -187,6 +187,44 @@ def convert_nue_to_enu_world(filename, mode='all', objects_pi=[], objects_pi_2=[
...
+        if type in ['corners', 'path', 'wayPoints', 'spine', 'startingAngle', 'endingAngle'] or (
+                        type in ['height'] and 'ElevationGrid' in last_type) or (
+                        type in ['shape'] and 'Crossroad' in last_type):
+            if '[]' not in line:  # if type not empty
+                next_line_is_corners = 1
+            print(line,end='')
+        elif next_line_is_corners == 1:
+            if ']' in line:  # we stop when we reach the end of the node 'point' of 'coord'
+                next_line_is_corners = -1
+                print(line, end ='')
```
stefaniapedrazzi, line 216:
```suggestion
                print(line, end='')
```

scripts/converter/convert_nue_to_enu_rub_to_flu.py
```diff
@@ -187,6 +187,44 @@ def convert_nue_to_enu_world(filename, mode='all', objects_pi=[], objects_pi_2=[
...
+            else:  # else we convert the line
+                # fix for issue where corners are comma separated
```
stefaniapedrazzi, line 218:
```suggestion
                # split coordinates separated by commas on new lines
```

scripts/converter/convert_nue_to_enu_rub_to_flu.py
```diff
@@ -187,6 +187,44 @@ def convert_nue_to_enu_world(filename, mode='all', objects_pi=[], objects_pi_2=[
...
+                if "," in line:
+                    vectors = [x for x in re.compile('[,]').split(line)]
+                    for vector in vectors:
+                        print(vector)
```
stefaniapedrazzi, line 222: The new lines are not correctly aligned.
In this file there is the `add_space` function that is used to compute the spaces needed to correctly align the values.

stefaniapedrazzi (changes requested):
Thank you for this fix!
I tested it and it seems to work fine.

If possible it would be better to avoid duplicating the loop through the file and apply all the changes directly the already existing loop.

RugnirViking:
I have made it so that the vectors are padded correctly, I had a go at putting everything into the single loop but it wasn't working quite right and I need to get back to work @stefaniapedrazzi

scripts/converter/convert_nue_to_enu_rub_to_flu.py
```diff
@@ -187,6 +187,43 @@ def convert_nue_to_enu_world(filename, mode='all', objects_pi=[], objects_pi_2=[
...
+        elif next_line_is_corners == 1:
+            if ']' in line:  # we stop when we reach the end of the node 'point' of 'coord'
+                next_line_is_corners = -1
+                print(line, end='')
+            else:  # else we convert the line
+                # split coordinates separated by commas on new lines
+                if "," in line:
+                    vectors = [x for x in re.compile('[,]').split(line)]
+                    for vector in vectors:
+                        print(add_space(line)+vector.strip())
```
stefaniapedrazzi, line 221:
```suggestion
                        print(add_space(line) + vector.strip())
```

stefaniapedrazzi (changes requested):
Thank you for fixing the alignment!

There is still just a minor coding style detail to fix then this PR is ready to be merged.

omichel:
@RugnirViking: I just sent you an invite to become a member of the Webots-committers group so that you should be to merge your PR directly and create new PRs on the cyberbotics repos.

stefaniapedrazzi (approved):
Thank you!

PR #5253, opened 2022-09-21
Fixed an axis typo

Jean-Eudes-le-retour:
Due to the axis convention change

https://cyberbotics.com/doc/guide/tutorial-5-compound-solid-and-physics-attributes?version=doc-axis-typo


BenjaminDeleze (approved):
Good catch

PS: your link is broken because it should be `?version` instead of `&version`

PR #5259, opened 2022-09-23
Better figure order

Jean-Eudes-le-retour:
When following the tutorial step-by-step, the low-level representation seems to come too early as it gives information from the next sections. Placing it at the end of the HingeJoint chapter seems more suited.

https://cyberbotics.com/doc/guide/tutorial-6-4-wheels-robot?version=doc-better-fig-placement


omichel (approved):
Thank you.

PR #5240, opened 2022-09-16
added shader that creates a more promissing spherical image

TheMangalex:
Updates  #2295

This PR provides a fisheye shader adapted from https://www.shadertoy.com/view/lssGD4 to create a spherical image which distorts in both directions instead of only one. 

It currently only works for rgb cameras as the merge_spherical.frag shader is also a basis for the LIDAR class and changing these projection for range images therefore destroys lidar range images and point clouds.

In general a test, if this shader projects the image correct, should be done.

Maybe the two shaders should be split into an equirectangular and a fisheye shader class which can be properly assigned.

PR #5266, opened 2022-09-26
Add Camera fisheye effect

stefaniapedrazzi:
Continues from #5240.

Tasks:
* [x] Improve `merge_spherical.frag` shader to produce fisheye effect
* [x] Replace `Camera.spherical` field with `Camera.projection` to enable both spherical and cylindrical projection
* [x] Remove old deprecated Camera fields
* [x] Update documentation
* [x] Add entry in Changelog
* [x] Update tests
* [x] Add new test


src/webots/nodes/WbCamera.cpp
```diff
@@ -107,16 +107,12 @@ void WbCamera::init() {
...
-    parsingWarn("Deprecated 'colorNoise' field, please use the 'noise' field instead.");
-    if (mNoise->value() == 0.0)
-      mNoise->setValue(colorNoise->value());
-    colorNoise->setValue(0.0);
+  WbSFBool *sphericalField = findSFBool("spherical");
+  if (sphericalField->value()) {  // Introduced in Webots R2023
+    parsingWarn("Deprecated 'spherical' field, please use the 'projection' field instead.");
+    if (isPlanarProjection())
+      mProjection->setValue("cylindrical");
+    sphericalField->setValue(false);
```
omichel, line 115: Why do you set it to false here? If the user saves, it will be FALSE while they didn't changed anything...

src/webots/nodes/WbCamera.cpp
```diff
@@ -107,16 +107,12 @@ void WbCamera::init() {
...
-    parsingWarn("Range finder type is not available for camera since Webots 8.4, please use a RangeFinder node instead.");
-
-  WbSFDouble *colorNoise = findSFDouble("colorNoise");
-  if (colorNoise->value() != 0.0) {  // Introduced in Webots 8.4.0
-    parsingWarn("Deprecated 'colorNoise' field, please use the 'noise' field instead.");
-    if (mNoise->value() == 0.0)
-      mNoise->setValue(colorNoise->value());
-    colorNoise->setValue(0.0);
+  WbSFBool *sphericalField = findSFBool("spherical");
+  if (sphericalField->value()) {  // Introduced in Webots R2023
```
omichel, line 111:
```suggestion
  if (sphericalField->value()) {  // Deprecated in Webots R2023
```

src/webots/nodes/WbLidar.cpp
```diff
@@ -81,6 +81,15 @@ void WbLidar::init() {
 
   mTcpImage = NULL;
   mTcpCloudPoints = NULL;
+
+  // backward compatibility
+  WbSFBool *sphericalField = findSFBool("spherical");
+  if (!sphericalField->value()) {  // Introduced in Webots R2023
```
omichel, line 87:
```suggestion
  if (!sphericalField->value()) {  // Deprecated in Webots R2023
```

src/webots/nodes/WbLidar.cpp
```diff
@@ -81,6 +81,15 @@ void WbLidar::init() {
...
+    parsingWarn("Deprecated 'spherical' field, please use the 'projection' field instead.");
+    if (mProjection->value() == "cylindrical")
+      mProjection->setValue("planar");
+    sphericalField->setValue(true);
```
omichel, line 91: Why this change?

src/webots/nodes/WbRangeFinder.cpp
```diff
@@ -33,6 +33,15 @@ void WbRangeFinder::init() {
   mMinRange = findSFDouble("minRange");
   mMaxRange = findSFDouble("maxRange");
   mResolution = findSFDouble("resolution");
+
+  // backward compatibility
+  WbSFBool *sphericalField = findSFBool("spherical");
+  if (sphericalField->value()) {  // Introduced in Webots R2023
```
omichel, line 39:
```suggestion
  if (sphericalField->value()) {  // Deprecated in Webots R2023
```

src/webots/nodes/WbRangeFinder.cpp
```diff
@@ -33,6 +33,15 @@ void WbRangeFinder::init() {
...
+    parsingWarn("Deprecated 'spherical' field, please use the 'projection' field instead.");
+    if (isPlanarProjection())
+      mProjection->setValue("cylindrical");
+    sphericalField->setValue(false);
```
omichel, line 43: Why this change?

omichel (changes requested):
Generally, it looks very good to me.
I only have a few questions/requests.
Also, I understand it makes no sense to have a spherical projection for a lidar, but for the range-finder, this is more arguable. What is the motivation for making the range-finder only planar or cylindrical and not spherical?

docs/reference/rangefinder.md
```diff
@@ -49,9 +49,10 @@ Since range-finder pixels are squares, the vertical field of view can be compute
 
 - `height`: height of the image in pixels
 
-- `spherical`: switch between a planar or a spherical projection.
-A spherical projection can be used for example to simulate a lidar device.
-More information on spherical projections is provided in the [spherical projection](camera.md#spherical-projection) section of the [Camera](camera.md) node.
+- `projection`: switch between a planar or a cylindrical projection.
+A cylindrical projection can be used for example to simulate a lidar device.
```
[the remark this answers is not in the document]
stefaniapedrazzi: Yes. I agree.
This was already in the documentation and I didn't add it.

Note that there are at least another reference to the lidar in the RangeFinder documentation.
This should be fixed in the released branch.

src/webots/nodes/WbCamera.cpp
stefaniapedrazzi: This is the backward compatibility method.
When loading the node we want to convert the settings of the deprecated `spherical` field to the new `projection` field.
So we exactly want to reset the `spherical` field because should no longer be used.
Even if the user doesn't change nothing, the saved world should be now compatible with R2023c and the `projection` field should be set and the `spherical` field should be reset to default.

stefaniapedrazzi:
> What is the motivation for making the range-finder only planar or cylindrical and not spherical?

I think the shader was only tested with RGB camera.
I will check it for range-finder images too.

src/webots/nodes/WbCamera.cpp
omichel: OK, that makes sense. Thank you for the explanation.

stefaniapedrazzi:
I enabled the spherical projection for the range-finder camera and fixed preventing using it for lidars.

Windows compilation is failing due to errors with OpenAL but they are not related to this PR (same errors in the scheduled nightly tests on master).

docs/reference/rangefinder.md
```diff
@@ -49,9 +49,8 @@ Since range-finder pixels are squares, the vertical field of view can be compute
 
 - `height`: height of the image in pixels
 
-- `spherical`: switch between a planar or a spherical projection.
-A spherical projection can be used for example to simulate a lidar device.
-More information on spherical projections is provided in the [spherical projection](camera.md#spherical-projection) section of the [Camera](camera.md) node.
+- `projection`: switch between a planar or a cylindrical projection.
```
omichel, line 52:
```suggestion
- `projection`: switch between a planar, a cylindrical or a spherical projection.
```

omichel (approved):
Thank you.

omichel:
It's nice to see that the Windows build appear to be repaired 😄.

PR #5272, opened 2022-09-27
Enhance robot window in webots.js for mjpeg mode

BenjaminDeleze:
**Description**
The internal windowing system and the robot windows were not supporting mjpeg streaming.
Moreover, the information regarding the robot window was sent in the `.x3d` file and by websocket messages with `robot window:`.

**Tasks**

- [x] Remove the information concerning the robot window in the `.x3d`
- [x] Modify the windowing system to support mjpeg
- [x] Modify webotsJS to get the information concerning the robot windows from the `robot window:` messages
- [x] Upgrade the `robot window: ` messages to allow
  - [x] The dynamic addition or removal of robot windows.
  - [x] To determine which window is the main on.

### DIFF
--- a/resources/web/wwi/Stream.js
+++ b/resources/web/wwi/Stream.js
@@ -57,7 +57,13 @@ export default class Stream {
 
   _onSocketMessage(event) {
     let data = event.data;
-    if (data.startsWith('robot:') || data.startsWith('robot window:'))
+
+    if (data.startsWith('robot window:')) {
+      console.log(data)
+      const json = JSON.parse(data.substring(14));
+      const robotWindow = json.window === '<generic>' ? 'generic' : json.window;
+      webots.currentView.robots.push({name: json.robot, window: robotWindow});
+    } else if (data.startsWith('robot:'))
       return 0; // We need to keep this condition, otherwise the robot window messages will be printed as errors.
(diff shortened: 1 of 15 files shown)


resources/web/wwi/Stream.js
```diff
@@ -57,7 +57,27 @@ export default class Stream {
...
+        const robots = webots.currentView.robots;
+        let index;
+        for (let i = 0; i < robots.length; i++) {
+          if (robots[i].name === json.robot) {
+            index = i;
+            break;
+          }
+        }
+        if (typeof index !== 'undefined')
+          robots.splice(index, 1);
```
omichel, line 75: Isn't that simpler?
```suggestion
        for (let i = 0; i < robots.length; i++) {
          if (robots[i].name === json.robot) {
            robots.splice(i, 1);
            break;
          }
        }
```

omichel (changes requested):

resources/web/wwi/Stream.js
BenjaminDeleze: I will test it. 
I don't usually like modifying the object I am iterating on, but as we break immediately, it should be ok

omichel (approved):
Thank you.

PR #5284, opened 2022-09-30
Fix controller restart after crash

### LINKED ISSUE #5279: Controller gets stuck after running once
**Describe the Bug**
We are working in the "Complete Apartment" demo and observe that Python controllers get stuck after running once. After pausing or restarting, the time will simply not advance. It works again when setting "synchronization" to FALSE, that is the controller seems to get stuck. Multiple students have this problem (not all), and I could not find commonalities. All of them use Windows with a variety of graphic cards (Intel, Nvidia). 

It seems Python simply does not output the errors more than once. 

**Steps to Reproduce**

We can share the lab / controller that we ship. Its complete apartment with a Turtlebot Burger in there.

omichel:
Fixes  #5279.

### DIFF
--- a/src/webots/control/WbController.cpp
+++ b/src/webots/control/WbController.cpp
@@ -158,22 +158,21 @@ WbController::~WbController() {
     mRobot->removeRemoteExternController();
     if (!mHasBeenTerminatedByItself)
       sendTerminationPacket(mTcpSocket, buffer, size);
-  } else if (mProcess && mProcess->state() != QProcess::NotRunning)
+  } else if (mProcess && mProcess->state() != QProcess::NotRunning) {
     mProcess->terminate();
+    mProcess->deleteLater();
+    mProcess = NULL;
+  }
 
   if (mExtern) {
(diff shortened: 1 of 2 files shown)


src/webots/control/WbController.cpp
```diff
@@ -158,22 +158,21 @@ WbController::~WbController() {
...
-    mHasBeenTerminatedByItself = false;
-  else {
-    delete mSocket;
-    delete mServer;
-    delete mProcess;
-    delete mTcpSocket;
-  }
+  delete mProcess;
+  delete mSocket;
+  delete mTcpSocket;
```
omichel, line 174:
```suggestion
  if (!hasBeenTerminatedByItself)
    delete mTcpSocket;
```

ygoumaz (approved):
Tested most of the cases I had in mind. Everything seems to work fine, expect that interrupting an `extern` controller triggers an assertion and crashes Webots. This problem is already present in Master, so no need to fix it in this PR. I will open an issue about that.

PR #5286, opened 2022-10-03
Fix camera plane in documentation

stefaniapedrazzi:
If I'm not wrong the camera plane should be `xy` and not `xz`.
It is referenced correctly just after at line 185.

### DIFF
--- a/docs/reference/camera.md
+++ b/docs/reference/camera.md
@@ -180,7 +180,7 @@ So this mode is costly in terms of performance! Reducing the resolution of the c
 
 When the camera is spherical, the image returned by the `wb_camera_get_image` function is a 2-dimensional array (s,t) in spherical coordinates.
 
-Let `hFov` be the horizontal field of view, and let `theta` be the angle in radian between the `(0, 0, -z)` relative coordinate and the relative coordinate of the target position along the `xz` plane relative to the camera, then `s=0` corresponds to a `theta` angle of `-hFov/2`, `s=(width-1)/2` corresponds to a `theta` angle of 0, and `s=width-1` corresponds to a `theta` angle of `hFov/2`.
+Let `hFov` be the horizontal field of view, and let `theta` be the angle in radian between the `(0, 0, -z)` relative coordinate and the relative coordinate of the target position along the `xy` plane relative to the camera, then `s=0` corresponds to a `theta` angle of `-hFov/2`, `s=(width-1)/2` corresponds to a `theta` angle of 0, and `s=width-1` corresponds to a `theta` angle of `hFov/2`.
 
 Similarly, let `vFov` be the vertical field of view (defined just above), and `phi` the angle in radian between the `(0, 0, -z)` relative coordinate and the relative coordinate of the target position along the `xy` plane relative to the camera, `t=0` corresponds to a `phi` angle of `-vFov/2`, `t=(height-1)/2` corresponds to a `phi` angle of 0, and `t=height-1` corresponds to a `phi` angle of `vFov/2`).
 


omichel:
I don't agree. According to [this figure](https://raw.githubusercontent.com/cyberbotics/webots/released/docs/reference/images/camera.png), it looks correct. The vertical and horizontal fields of view are not the same and are defined from different planes. So, I believe that the current documentation is correct. If that is not clear for you, this probably means that we need a better explanation here. Maybe an additional figure could help?

stefaniapedrazzi:
You're right my fix is wrong.
However, then the camera image plane accordingly to the figure should be `yz` and not `xz`.
I don't see why it is correct the `xz` plane on line 183 and `xy` on line 185 as well the (0, 0, -z) coordinate if the camera axis is aligned with the x axis.

omichel:
> However, then the camera image plane accordingly to the figure should be `yz` and not `xz`.
> I don't see why it is correct the `xz` plane on line 183 and `xy` on line 185 as well as the (0, 0, -z) coordinate if the camera axis is aligned with the x axis.

On lines 183 and 185, we are not speaking about the camera image plane, which is indeed `yz`. We are speaking about planes relative to the camera, but not the image plane. That looks fairly clear to me. But if it doesn't look clear to you, could you please rephrase it so that it clearer for you and hopefully for users reading this documentation?

omichel (commented):
I know we have a bug with the spherical camera projection, but adding a picture now with a cylindrical projection image and labeling it spherical camera image is also very confusing.

docs/reference/camera.md
```diff
@@ -178,11 +178,17 @@ Moreover an algorithm computing the spherical projection is applied on the resul
...
+%figure "Spherical camera image"
+
+![spherical_camera_image.png](images/spherical_camera_image.png)
+
 When the camera is spherical, the image returned by the `wb_camera_get_image` function is a 2-dimensional array (s,t) in spherical coordinates.
 
-Let `hFov` be the horizontal field of view, and let `theta` be the angle in radian between the `(0, 0, -z)` relative coordinate and the relative coordinate of the target position along the `xz` plane relative to the camera, then `s=0` corresponds to a `theta` angle of `-hFov/2`, `s=(width-1)/2` corresponds to a `theta` angle of 0, and `s=width-1` corresponds to a `theta` angle of `hFov/2`.
+Let assume a 3D target point is located at coordinates `(X, Y, Z)` relative to the camera origin.
+
+Let `hFov` be the horizontal field of view, and let `theta` be the angle in radian between the `(X, 0, 0)` relative coordinate and the relative coordinate of the target position along the `xy` plane relative to the camera, i.e., `(X, Y, 0)`, then `s=0` corresponds to a `theta` angle of `hFov/2`, `s=(width-1)/2` corresponds to a `theta` angle of 0, and `s=width-1` corresponds to a `theta` angle of `-hFov/2`.
```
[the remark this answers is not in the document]
stefaniapedrazzi: Yes, it is the angle between the vector from point (0, 0, 0) to point (X, 0, 0) and the vector from point (0, 0, 0) and point (X, Y, 0).

I agree that an angle between coordinates doesn't make sense.
This was already the case before. I will improve it.

I would keep the reference to the plane projection because I think it is useful.

stefaniapedrazzi:
> I know we have a bug with the spherical camera projection, but adding a picture now with a cylindrical projection image and labeling it spherical camera image is also very confusing.

Ok. So I will just fix the text here and add the picture in the other PR where the camera projection and all the relative documentation is updated.

stefaniapedrazzi:
I removed and image for this version of the documentation and tried to improve the text.

docs/reference/camera.md
```diff
@@ -180,9 +180,13 @@ So this mode is costly in terms of performance! Reducing the resolution of the c
...
 
-Let `hFov` be the horizontal field of view, and let `theta` be the angle in radian between the `(0, 0, -z)` relative coordinate and the relative coordinate of the target position along the `xz` plane relative to the camera, then `s=0` corresponds to a `theta` angle of `-hFov/2`, `s=(width-1)/2` corresponds to a `theta` angle of 0, and `s=width-1` corresponds to a `theta` angle of `hFov/2`.
+Let assume a 3D target point is located at coordinates `(X, Y, Z)` relative to the camera origin.
 
-Similarly, let `vFov` be the vertical field of view (defined just above), and `phi` the angle in radian between the `(0, 0, -z)` relative coordinate and the relative coordinate of the target position along the `xy` plane relative to the camera, `t=0` corresponds to a `phi` angle of `-vFov/2`, `t=(height-1)/2` corresponds to a `phi` angle of 0, and `t=height-1` corresponds to a `phi` angle of `vFov/2`).
+Let `hFov` be the horizontal field of view, and let `theta` be the angle in radian between the vector to `(X, 0, 0)` and the vector to `(X, Y, 0)`, i.e., the projection of the target point position along the `xy` plane relative to the camera.
```
omichel, line 185: Honestly, this is simpler and clearer to me:
```suggestion
Let `hFov` be the horizontal field of view, and let `theta` be the angle in radian between vector `(X, 0, 0)` and vector `(X, Y, 0)`.
```
@BenjaminDeleze & @ygoumaz: what do you think?

ygoumaz: I think this version is better. Calling a vector by its components is totally correct and better than using the formulation 'to <coordinates>'. Also I believe the part of the sentence about the projection is a bit complicated and not very understandable.  I would also remove it.

However, I would add a simple diagram containing the camera, a sphere, the point (X,Y,Z) and the 2 angles `theta` and `phi`, because any textual formulation here would be complicated to really understand what is going on.

stefaniapedrazzi:
I just added the node about the spherical camera actually using a cylindrical projection.
I will add the figure about the spherical/cylindrical coordinates as suggested in #5266  when fixing cylindrical/spherical projection mismatch.

omichel (changes requested):

omichel (approved):
Thank you.

PR #5287, opened 2022-10-04
Fix pointSet shader for web

BenjaminDeleze:
The PointSet shader was working fine on desktop, but in some situations on the web it was crashing the simulation.

It is because webgl is less permissive than openGL when it comes to shaders.

The issue was that the draw buffer is expecting fragment output: `color` and `normal`, as for `Phong` shader for example.
However, the PointSet shader was only returning a `color` output because the `normal` one is not used.

I added a null `normal` output to fix the issue.

### DIFF
--- a/resources/wren/shaders/point_set.frag
+++ b/resources/wren/shaders/point_set.frag
@@ -4,7 +4,8 @@ precision highp float;
 
 in vec3 color;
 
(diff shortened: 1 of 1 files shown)


omichel (approved):
Thank you.

PR #5297, opened 2022-10-05
Non-Swig Python API

omichel:
It turns out it is way easier to use and maintain a Python wrapper which relies on the `ctypes` python interface with native shared libraries, rather than Swig. The advantages are the following:

- no dependency on Swig any more.
- no need to create and distribute many versions of the Python wrappers (for Python 3.7, 3.8, 3.9, 3.10, 3.7-brew, 3.8-brew, 3.9-brew, 3.10-brew, etc.)
- no need to create a new version of the Python wrappers when a new release of Python is available and deprecate an old one.
- support for any python distribution (including conda, anaconda, etc.)
- easier to debug.

The only disadvantage is that we have to maintain it manually, but even with Swig we had to sometimes maintain it manually (with some special API functions) and we had to test it anyhow, so I believe this is a small drawback.

This PR also adds python controller examples to most of the [project/samples/devices/controllers/](https://github.com/cyberbotics/webots/tree/feature-new-python-api/projects/samples/devices/controllers)* folders. The examples not converted in this PR will be converted in a subsequent PR.

Completion of the API:

- [x] Robot
- [x] Supervisor
- [x] Accelerometer
- [x] Altimeter
- [x] Brake
- [x] Camera
- [x] Compass
- [x] Connector
- [x] Display
- [x] DistanceSensor
- [x] Emitter
- [x] Gps
- [x] Gyro
- [x] InertialUnit
- [x] Led
- [x] Lidar
- [x] LightSensor
- [x] Motor
- [x] Pen
- [x] PositionSensor
- [x] Radar
- [x] RangeFinder
- [x] Receiver
- [x] Skin
- [x] Speaker
- [x] TouchSensor

Additionally more Python wrappers have to be implemented:

- [x] motion API
- [x] libCar
- [x] libDriver
- [ ] ~manager [Darwin-OP]~ The managers library is written in C++, it would be a big effort to write a Python wrapper to it, moreover the Robotis OP2 robot is now unsupported (not sold any more), so it is not needed to invest such an effort here. The only example that uses the Python interface to the Robotis OP2 managers is the one from the humanoid marathon in robotbenchmark. Since we are going to allow C++ programming in robotbenchmark, it is better to convert the robotbenchmark humanoid marathon to C++ instead.
- [x] Convert robotbenchmark humanoid marathon to C++.

Test the following examples:

- [x] projects/languages/python/
- [x]  projects/sample/devices/
- [x] projects/sample/robotbenchmark/

Tests on:

- [x] Windows
- [x] Linux
- [x] macOS


BenjaminDeleze (changes requested):
I just began testing and it seems to work well, however I think that some functions are missing:

The `getModel`, `getName` and `getNodeType` functions of `device` are missing.
The `getSamplingPeriod` function of at least the `accelerometer` and `altimeter` is missing.

It' i a good idea to have both `c` and `python` controllers for the devices. However is it possible to specify in webots that we want to use the Python controller? The only way I found was to delete everything related to the `.c` controller in the folder, which is not very handy.

projects/samples/devices/controllers/altimeter/altimeter.py
```diff
@@ -0,0 +1,53 @@
...
+        self.timeStep = int(self.getBasicTimeStep())
+        self.altimeter = self.getDevice('altimeter')
+        self.altimeter.enable(self.timeStep)
+        self.left_motor = self.getDevice('left wheel motor')
+        self.right_motor = self.getDevice('right wheel motor')
+        self.left_motor.setPosition(float('inf'))
+        self.right_motor.setPosition(float('inf'))
+
+        self.left_motor.setVelocity(0.5)
+        self.right_motor.setVelocity(-0.5)
```
BenjaminDeleze, line 34:
```suggestion
```

lib/controller/python/controller/camera.py
```diff
@@ -0,0 +1,249 @@
...
+    def setFov(self, f: float):
+        self.fov = f
+
+    def getHeight(self):
+        return self.height
+
+    def getImage(self) -> bytes:
+        return self.image
+
+    def getImageArray(self) -> List[List[List[int]]]:
```
BenjaminDeleze, line 63: This function throw an error: 
```
AttributeError: 'LP_c_ubyte' object has no attribute 'append'
```

lib/controller/python/controller/camera.py
```diff
@@ -0,0 +1,249 @@
...
+        return wb.wb_camera_recognition_has_segmentation(self._tag) != 0
+
+    def isRecognitionSegmentationEnabled(self) -> bool:
+        return wb.wb_camera_recognition_is_segmentation_enabled(self._tag) != 0
+
+    def getRecognitionSegmentationImage(self) -> bytes:
+        return wb.wb_camera_recognition_get_segmentation_image(self._tag)
+
+    def getRecognitionSegmentationImageArray(self) -> List[int]:
+        return [x for x in self.getRecognitionSegmentationImage()]
```
BenjaminDeleze, line 246: I get:
```
WARNING: camera_segmentation: The process crashed some time after starting successfully.
WARNING: 'camera_segmentation' controller crashed.
```

projects/samples/devices/controllers/camera_recognition/camera_recognition.py
```diff
@@ -0,0 +1,60 @@
...
+    def run(self):
+        while self.step(self.timeStep) != -1:
+            # Get current number of object recognized
+            number_of_objects = self.camera.getRecognitionNumberOfObjects()
+            print(f'Recognized {number_of_objects} objects.')
+            print(' ')
+            objects = self.camera.getRecognitionObjects()
+            counter = 1
+            for object in objects:
+                print(f' Object {counter}/{number_of_objects}: {object.model} (id = {object.id})')
```
BenjaminDeleze, line 45: As we access the attribute directly (instead of user `getModel()` for example), we should update accordingly.

BenjaminDeleze (changes requested):
`setExposure` is missing in `Camera`

lib/controller/python/controller/camera.py
```diff
@@ -0,0 +1,249 @@
...
+
+    def setFocalDistance(self, d: float):
+        self.focal_distance = d
+
+    def setFov(self, f: float):
+        self.fov = f
+
+    def getHeight(self):
+        return self.height
+
```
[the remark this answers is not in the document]
omichel: I just forgot this one...

BenjaminDeleze (commented):
The new implementation of `getName` and `getModel` does not have the same return type as the old one.
New -> b'altimeter'
Old -> altimeter

lib/controller/python/controller/brake.py
```diff
@@ -0,0 +1,57 @@
...
+    def motor(self):
+        from controller.motor import Motor
+        tag = wb.wb_brake_get_motor(self._tag)
+        return None if tag == 0 else Motor(tag)
+
+    @property
+    def position_sensor(self):
+        from controller.position_sensor import PositionSensor
+        tag = wb.wb_brake_get_position_sensor(self._tag)
+        return None if tag == 0 else PositionSensor(tag)
```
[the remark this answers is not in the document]
omichel: I did not added them here because it would cause a circular dependency problem which I was unable to solve.
However, these type annotations are not mandatory, they are just a help for the user to know which data type is returned.

projects/samples/devices/controllers/connector/connector.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def run(self):
+        robot_number = int(self.getName()[6])
+        hop_counts = 0
+        direction = 1
+        state = 'CONNECTOR_IN_PLACE'
+        left_speed = 0
+        right_speed = 0
+        while self.step(self.timeStep) != -1:
+            if state == 'CONNECTOR_IN_PLACE':  # wait for the motor to be in place
```
BenjaminDeleze, line 50:
```suggestion
        while self.step(self.timeStep) != -1:
            left_speed = 0
            right_speed = 0
            if state == 'CONNECTOR_IN_PLACE':  # wait for the motor to be in place
```

Otherwise the robot does not stop while rotating its arm, resulting in a clanky execution.

lib/controller/python/controller/display.py
```diff
@@ -0,0 +1,127 @@
...
+    def drawLine(self, x1: int, y1: int, x2: int, y2: int):
+        wb.wb_display_draw_line(self._tag, int(x1), int(y1), int(x2), int(y2))
+
+    def drawOval(self, cx: int, cy: int, a: int, b: int):
+        wb.wb_display_draw_oval(self._tag, int(cx), int(cy), int(a), int(b))
+
+    def drawPixel(self, x: int, y: int):
+        wb.wb_display_draw_pixel(self._tag, int(x), int(y))
+
+    def drawPolygon(self, x: List[int], y: List[int]):
```
BenjaminDeleze, line 55: Throw `TypeError: 'list' object cannot be interpreted as an integer` when doing:
```
p = [50, 60, 70, 650]
q = [10, 40, 50, 300]
self.camera_display.drawPolygon(p, q)
```
Which is working in develop

lib/controller/python/controller/display.py
```diff
@@ -0,0 +1,127 @@
...
+    def drawRectangle(self, x: int, y: int, width: int, height: int):
+        wb.wb_display_draw_rectangle(self._tag, int(x), int(y), int(width), int(height))
+
+    def drawText(self, text: str, x: int, y: int):
+        wb.wb_display_draw_text(self._tag, str.encode(text), int(x), int(y))
+
+    def fillOval(self, cx: int, cy: int, a: int, b: int):
+        wb.wb_display_fill_oval(self._tag, int(cx), int(cy), int(a), int(b))
+
+    def fillPolygon(self, x: List[int], y: List[int]):
```
BenjaminDeleze, line 70: Same error as `drawPolygon`

projects/samples/devices/controllers/connector/connector.py
omichel: Well spotted!

projects/samples/devices/controllers/emitter_receiver/emitter_receiver.py
```diff
@@ -0,0 +1,109 @@
...
+    def __init__(self):
+        super(Controller, self).__init__()
+
+        # Get a handler to the motors and set target position to infinity (speed control).
+        self.left_motor = self.getDevice('left wheel motor')
+        self.right_motor = self.getDevice('right wheel motor')
+        self.left_motor.setPosition(float('inf'))
+        self.right_motor.setPosition(float('inf'))
+        self.left_motor.setVelocity(0.0)
+        self.right_motor.setVelocity(0.0)
```
BenjaminDeleze, line 36:
```suggestion
```

projects/samples/devices/controllers/emitter_receiver/emitter_receiver.py
```diff
@@ -0,0 +1,109 @@
...
+        while self.step(self.timeStep) != -1:
+            # The emitter simply sends the message but the receiver
+            # has to check if there is something before it can reads the buffer.
+            if self.robot_type == 'EMITTER':
+                self.communication.send("Hello!")
+            else:
+                # is there at least one packet in the receiver's queue?
+                if self.communication.getQueueLength() > 0:
+                    # read current packet's data
+                    buffer = self.communication.getData().decode()
```
BenjaminDeleze, line 71: This line throws the following error:
```
AttributeError: 'str' object has no attribute 'decode'. Did you mean: 'encode'?
```

Moreover `Receiver.getData` seems to be deprecated and should not be used in an example

lib/controller/python/controller/emitter.py
```diff
@@ -0,0 +1,87 @@
...
+        wb.wb_emitter_set_channel(self._tag, c)
+
+    def getRange(self) -> float:
+        return self.range
+
+    @property
+    def range(self) -> float:
+        return wb.wb_emitter_get_range(self._tag)
+
+    def setRange(self, range: float):
```
BenjaminDeleze, line 82: I don't know why but the `setRange` function does not seems to work.
If I do:
```
self.communication.setRange(4)
self.communication.getRange()
```
I get 0 whereas it works on develop

lib/controller/python/controller/gps.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def getCoordinateSystem(self) -> int:
+        return self.coordinate_system
+
+    @property
+    def coordinate_system(self) -> int:
+        return wb.wb_gps_get_coordinate_system(self._tag)
+
+    @staticmethod
+    def convertToDegreesMinutesSeconds(decimalDegrees):
```
BenjaminDeleze, line 43: Missing return type

BenjaminDeleze (commented):
Nao_demo_python `Home` and `End` keyboard keys result in an error

projects/samples/devices/controllers/lidar/lidar.py
```diff
@@ -0,0 +1,66 @@
...
+        speed = [0, 0]
+        us_value = [0, 0]
+        n = self.lidar.getNumberOfPoints()
+        while self.step(self.timeStep) != -1:
+            value = self.lidar.getRangeImage()
+            i = random.randrange(n)
+            print(f'range[{i}] = {value[i]}')
+            points = self.lidar.getPointCloud()
+            if points:
+                print(f'point[{i}] = {points[i].x} {points[i].y} {points[i].z} {points[i].layer} {points[i].time}')
```
BenjaminDeleze, line 54: Was it for debugging or is it on purpose?
Because it is not in the C code and it is flooding the terminal.

lib/controller/python/controller/lidar.py
```diff
@@ -0,0 +1,155 @@
...
+    def getRangeImage(self) -> List[float]:
+        return self.range_image[:self.horizontal_resolution * self.number_of_layers]
+
+    def defRangeImageArray(self) -> List[List[float]]:
+        array = []
+        for i in range(self.number_of_layers):
+            array.append(self.getLayerRangeImage(i))
+        return array
+
+    def getLayerRangeImage(self, layer) -> List[float]:
```
BenjaminDeleze, line 81: This function throws: `TypeError: 'LP_c_float' object is not callable`

lib/controller/python/controller/lidar.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def getPointCloud(self) -> List[LidarPoint]:
+        number_of_points = self.number_of_points
+        data = bytes(wb.wb_lidar_get_point_cloud(self._tag)[:number_of_points * 20])
+        list = []
+        for i in range(number_of_points):
+            list.append(LidarPoint(data, i))
+        return list
+
+    def getLayerPointCloud(self, layer: int) -> List[LidarPoint]:
```
BenjaminDeleze, line 149: Error: 
```
Error: wb_lidar_get_layer_point_cloud() called with a 'layer' argument (426052200) bigger or equal to the number of layers of this lidar (6).
```
When call like: `self.lidar.getLayerPointCloud(0)`

lib/controller/python/controller/lidar.py
```diff
@@ -0,0 +1,155 @@
...
+    def getHorizontalResolution(self) -> int:
+        return self.horizontal_resolution
+
+    def getNumberOfLayers(self) -> int:
+        return self.number_of_layers
+
+    def getFrequency(self) -> float:
+        return self.frequency
+
+    def setFrequency(self, frequency: float):
```
BenjaminDeleze, line 66: Return the wrong value:
```
self.lidar.setFrequency(2)
print(self.lidar.getFrequency())
```
return 0 instead of 2 (correct in develop)

lib/controller/python/controller/light_sensor.py
```diff
@@ -0,0 +1,42 @@
...
+class LightSensor(Sensor):
+    wb.wb_light_sensor_get_value.restype = ctypes.c_double
+    wb.wb_distance_sensor_get_lookup_table.restype = ctypes.POINTER(ctypes.c_double)
+
+    def __init__(self, name: Union[str, int], sampling_period: int = None):
+        self._enable = wb.wb_light_sensor_enable
+        self._get_sampling_period = wb.wb_light_sensor_get_sampling_period
+        super().__init__(name, sampling_period)
+
+    def getLookupTable(self) -> List[float]:
```
BenjaminDeleze, line 30: Error:
```
TypeError: 'int' object is not subscriptable
```

lib/controller/python/controller/motor.py
```diff
@@ -0,0 +1,238 @@
...
+    def getMaxVelocity(self) -> float:
+        return self.max_velocity
+
+    def getAcceleration(self) -> float:
+        return self.target_acceleration
+
+    def getAvailableForce(self) -> float:
+        return self.available_force
+
+    def getMaxForce(self) -> float:
```
BenjaminDeleze, line 82: Return `None`

lib/controller/python/controller/motor.py
```diff
@@ -0,0 +1,238 @@
...
+        return self.max_force
+
+    def getAvailableTorque(self) -> float:
+        return self.available_torque
+
+    def getMaxTorque(self) -> float:
+        return self.max_torque
+
+    def getMultiplier(self) -> float:
+        return self.multiplier
```
BenjaminDeleze, line 92: Both function return `None`

lib/controller/python/controller/mouse.py
```diff
@@ -0,0 +1,71 @@
...
+    def enable(self, p: int):
+        wb.wb_mouse_enable(p)
+
+    def disable(self):
+        wb.wb_mouse_disable()
+
+    def getSamplingPeriod(self) -> int:
+        return self.sampling_period
+
+    def getState(self) -> MouseState:
```
BenjaminDeleze, line 49: It causes the controller to crash.
I used this code in the run function of a device example:
```
mouse = Mouse()
mouse.enable(self.timeStep)
while self.step(self.timeStep) != -1:
    print(mouse.getState())
```

lib/controller/python/controller/position_sensor.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def getValue(self) -> float:
+        return self.value
+
+    def getBrake(self):
+        return self.brake
+
+    def getMotor(self):
+        return self.motor
+
```
BenjaminDeleze, line 41: missing the getType(self) function

BenjaminDeleze:
I finished a first review, let me know when you are done with the corrections.

omichel:
Thank you for this deep review!
I believe I fixed all the problems you reported.

lib/controller/python/controller/robot.py
```diff
@@ -0,0 +1,395 @@
...
+    def getSynchronization(self) -> bool:
+        return self.synchronization
+
+    def getNumberOfDevices(self) -> int:
+        return self.number_of_devices
+
+    def getTime(self) -> float:
+        return self.time
+
+    def getUrdf(self, prefix: str) -> str:
```
BenjaminDeleze, line 295: In the doc, there is a default value for the prefix

BenjaminDeleze (changes requested):
Thanks,
I still have issues with:

- `self.camera.getRecognitionSegmentationImageArray()`, I get 
```
line.append([image[i + 2], image[i + 1], image[i]])  # RGB pixel
ValueError: NULL pointer access
```

- Light sensor `getLookupTable`:
```
TypeError: unsupported operand type(s) for *: 'int' and 'LP_c_double'
```

omichel:
I believe I fixed the remaining issues.

BenjaminDeleze:
I confirm that issues have been fixed.
@ygoumaz I am done with the review, you can go.

ygoumaz (changes requested):
`docs/guide/running-extern-robot-controllers.md` should be updated because currently it gives information about Python3X folders, which won't exist anymore.

src/webots/control/WbLanguageTools.cpp
```diff
@@ -69,12 +69,11 @@ QString WbLanguageTools::pythonCommand(QString &shortVersion, const QString &com
     "To fix the problem, you should set the full path of your python command in "
     "Webots->preferences->python command.\n";
 #else
-    QObject::tr("Webots requires Python version 3.10, 3.9, 3.8 or 3.7 from python.org in your current PATH.\n"
+    QObject::tr("Webots requires Python version 3.7 or ealier in your current PATH.\n"
```
ygoumaz, line 72:
```suggestion
    QObject::tr("Webots requires Python version 3.7 or newer in your current PATH.\n"
```

I think 'earlier' here would mean 'previous' instead of 'more recent'.

ygoumaz (changes requested):
I still need to test the car and driver libraries.

lib/controller/python/vehicle/driver.py
```diff
@@ -0,0 +1,230 @@
...
+            path = os.path.join('lib', 'controller')
+            car = 'car.dll'
+            driver = 'driver.dll'
+        elif sys.platform == 'darwin':
+            path = os.path.join('Contents', 'MacOS', 'lib', 'controller')
+            car = 'libcar.dylib'
+            driver = 'libdriver.dylib'
+        ctypes.cdll.LoadLibrary(os.path.join(os.environ['WEBOTS_HOME'], path, car))
+        self.api = ctypes.cdll.LoadLibrary(os.path.join(os.environ['WEBOTS_HOME'], path, driver))
+        self.api.wbu_driver_get_brake_intensity.restype = ctypes.c_double
```
ygoumaz, line 49:
```suggestion
        self.api.wbu_driver_get_brake_intensity.restype = ctypes.c_double
        self.api.wbu_driver_get_current_speed.restype = ctypes.c_double
```

The following snippet of code:
```
while self.step() != -1:
  print(self.getCurrentSpeed())
```
produces incorrect large int values instead of floats without this additional line.

ygoumaz (approved):
My last tests on the libcar and libdriver were successful. I believe the PR is ready to be merged. Thank you.

BenjaminDeleze (approved):
Thank you!

PR #5305, opened 2022-10-06
Fix lidar reset when extern controller changed

### LINKED ISSUE #5298: Lidar ranges (are again) not available after API restart
This bug came back in 2022b, it had been fixed before.
https://github.com/cyberbotics/webots/issues/3509

By the way, it's exactly the issue described here https://github.com/cyberbotics/webots/discussions/4154

note that even calling wb_supervisor_simulation_reset(); and the start of the controller don't change anything, even if I let the simulation run 1 step between reset and sensor setup.

after a look to the code, it's seems that the two lines in https://github.com/cyberbotics/webots/pull/3544/files#diff-b5827c7dc2bacb4ab37687df23574581283aec96cde3b9ac0143bd9e0d926b72 
have been forgotten...

stefaniapedrazzi:
Fix #5298:
the shared memory file was not sent to the extern controller when connecting in the middle of the simulation if the extern controller was not terminating correctly.
It works correctly for range finders and camera because the WbLidar class overrides the WbAbstractCamera `C_SAMPLING_PERIOD` command.

TODO:
* [ ] fix or remove assertion failure when killing an extern controller

### DIFF
--- a/src/webots/nodes/WbLidar.cpp
+++ b/src/webots/nodes/WbLidar.cpp
@@ -296,6 +296,9 @@ void WbLidar::handleMessage(QDataStream &stream) {
     if (!hasBeenSetup()) {
       setup();
       mSendMemoryMappedFile = true;
+    } else if (mHasExternControllerChanged) {
+      mSendMemoryMappedFile = true;
+      mHasExternControllerChanged = false;
     }
 
     return;


stefaniapedrazzi:
When the controller exits and disconnects normally from Webots, the memory mapped file is immediately re-initialized and the flag used to know if the a new  memory mapped file needs to be sent to the controller is already set to TRUE (same as at the beginning of a simulation).

ygoumaz:
I already noticed the assertion failure in debug mode when reviewing #5284 and forgot to open an issue. The assertion failure is wrong. The assertion itself also didn't make a lot of sense to me, but I haven't checked in depth. Maybe we can just get rid of it. Do you plan to fix this issue in this PR too?

stefaniapedrazzi:
Yes, the plan was to fix it in this PR too. 
I started to check if there is meaningful way to fix it. But I need to do more tests.

ShuffleWire:
As far as I am concerned, the current state of the MR fix my issue, so thanks you :)

ygoumaz (approved):
Thank you!

PR #5307, opened 2022-10-07
WebotsJS private variables

BenjaminDeleze:
The ECMAScript 2022 introduces private variables and classes in JS.
They work as follow:
```
class myClass {
  #myVariable; //private variables must be declared before the constructor
  constructor(value){
    this.#myVariable = value;
  }
  #myPrivateFunction(){
     console.log(this.#myVariable);
  }
  publicFunction(){
    this.#myPrivateFunction();
  }
```
As you can see private methods and variable must be prefixed with `#`
You can read more here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

When trying to access a private field from outside the class, an error will be thrown. 
Warning: the private variable is only accessible in the current class and will not be inherited by any subclasses.
In such case (when we want a parent class to share fields with a child class), we need to use the `_`prefix which does not enforce anything, but symbolize that we should not use that field outside of those classes.

I updated the coding style wiki to add those rules.

Concerning the support by browsers, it is well supported: https://caniuse.com/mdn-javascript_classes_private_class_fields

I also took the opportunity to clean some useless variables/functions.

If you want to test, the code is in the `testing` folder of the server.


omichel (approved):
Very nice. I believe it greatly improves readability and maintenance.
We should apply this to all our JS code, including webots.cloud.

PR #5308, opened 2022-10-07
Fix macOS compilation due to imu.c

ygoumaz:
The controller of the imu sample world (`imu.c`) is using the `abs()` method instead of the `fabs()` one. This triggers a compilation error on macOS.


omichel (approved):
Thank you.

PR #5309, opened 2022-10-07
Translation: Add some in French

ShuffleWire:
Hello,

I'm trying to add french translation to Webots :) 
Before doing more, I wish to know if it's the correct way to do.
Question : have your rules on character escape ? for instance i'm using `&#39;` for " ' ", is it correct ?
Just to be sure why is there &amp; before some terms ? (eg `<source>&amp;Filter</source>`)

If there are some French folks in the wild, please give your taught on my translation (I'm french, but more eye to look at it is better :)  )

Thanks

### DIFF
--- a/resources/translations/wb_fr.ts
+++ b/resources/translations/wb_fr.ts
@@ -5,87 +5,87 @@
     <name>ConsoleEdit</name>
     <message>
         <source>&amp;Filter</source>
-        <translation type="unfinished"></translation>
+        <translation>Filtrer</translation>
     </message>
     <message>
         <source>Display all the logs.</source>
-        <translation type="unfinished"></translation>
+        <translation>Afficher tout les journaux.</translation>
     </message>
     <message>
         <source>Display all the messages from Webots.</source>
-        <translation type="unfinished"></translation>
+        <translation>Afficher tout les messages de Webots.</translation>
     </message>
     <message>
         <source>Display parsing error when editing or loading a world.</source>
-        <translation type="unfinished"></translation>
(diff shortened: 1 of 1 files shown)


omichel:
I believe the `&` character is for keyboard shortcuts (for menus).
Normally, you should HTML for any special character, like `&apos;` for `'`, but maybe simply `'` is working (you should test it).
If it works, then it is probably better, as more readable.

omichel:
If your PR is not ready for review, please turn it into a draft (see right column of this page), and turn it again to "ready for review" once it is.

ShuffleWire:
> I believe the `&` character is for keyboard shortcuts (for menus). Normally, you should HTML for any special character, like `&apos;` for `'`, but maybe simply `'` is working (you should test it). If it works, then it is probably better, as more readable.

i don't know if it work, but it mess up VSCode syntax highlighting ! so I would not go that route...


PS : Gitlab have actually interpreted my escape code.... that make my message meaningless. I will edit that !

ShuffleWire:
Additionnal question : What the meaning of `type="unfinished"` ? I initially thought that it was to indicate missing or WIP translation, but the following seems very fine, for instance...
```
<message>
        <source>Webots cannot compute inertia for ElevationGrid objects.</source>
        <translation type="unfinished">Webots ne peut pas calculer l&apos;inertie pour les objets ElevationGrid.</translation>
    </message>
```

omichel:
I am not sure about this unfinished type, probably we simply forgot to remove it. I believe it is better that you remove it from your translations.

omichel:
I am closing this PR as it was inactive for a long time.

PR #5310, opened 2022-10-07
Fix TCP extern controller connection when world is loading

ygoumaz:
If a TCP controller tries to connect to a Webots instance that is currently loading a world, Webots crashes.


ygoumaz:
I added the connection closing when reloading or loading another world. This way, no crash happens if a remote controller tries to connect during this moment.

omichel (approved):
That looks good to me.
Thank you.

PR #5314, opened 2022-10-10
Improve PROTO web loader

ad-daniel:
**Description**

- [x] Rework execution to support EXTERNPROTO and build PROTO prototypes
- [x] Rework logic such that a node structure is generated (and preserved)
   - [x] Adapt handling of DEF, IS, USE
- [x] Add VRML field instances with appropriate exporters
    - [x] x3d
    - [x] js
    - [ ] vrml ( when writing PROTO ? not needed now )  
- [x] Handle all VRML types 
- [x] Add script to automatically generate FieldModel.js from .wrl files
- [ ] Handle roles when exporting to x3d
   - [x] ImageTexture roles
   - [x] boundingObject role
      - [x] bounding primitives showing up as solid rather than wireframe 
   - [x] jointParameters role  
   - [ ] animated geometry role
- [x] Add Parameter class wrapper
   - [x] Add comparators for all field types 
   - [x] Add default values and limit x3d exporter to write only non-default ones
- [x] Consolidate BaseNode and ProtoNode classes
- [x] Check if WRL or Webots default values are enforced in PROTO (enforces WRL)


resources/web/wwi/protoVisualizer/Tokenizer.js
```diff
@@ -213,7 +220,23 @@ export default class Tokenizer {
...
+      this.skipToken('IS');
+      this.nextToken();
+      return;
+    }
+
+    if (this.peekWord() === 'USE') {
+      this.skipToken('USE');
+      this.nextToken();
+      return;
+    }
```
BenjaminDeleze, line 233: Seems like it could be combine in one `if`

resources/web/wwi/protoVisualizer/Parameter.js
```diff
@@ -0,0 +1,98 @@
...
+  get isTemplateRegenerator() {
+    return this.#isTemplateRegenerator;
+  }
+
+  set isTemplateRegenerator(value) {
+    this.#isTemplateRegenerator = value;
+  }
+
+  setValueFromJavaScript(v) {
+    this.value.setValueFromJavaScript(v);
```
BenjaminDeleze, line 78: Not sure if inside the class we should use the `getter` or directly access the attribute with `this.#value`. Personally I would tend to access directly the value.

BenjaminDeleze (changes requested):
Seems good, just some minor remarks:

resources/web/wwi/nodes/utils/id_provider.js
```diff
@@ -1,8 +1,13 @@
...
 
 function getAnId() {
-  return 'n' + undefinedID--;
+  return 'n' + nodeUndefinedID--;
 }
 
-export {getAnId};
+function getAParameterId() {
+  return 'p' + parameterUndefinedID--;
+}
```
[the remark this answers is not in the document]
ad-daniel: probably not anymore, if yes I'll re-add it later

BenjaminDeleze (approved):
Seems good to me

PR #5328, opened 2022-10-13
Enhancement add new web nodes

BenjaminDeleze:
This PR add the missing nodes (except for `WbSkin`) in WebotsJS to be able to parse the protos.
Some classes are empty and just extends another one. It is to be able to retrieve the name of the device.


omichel:
Will there be an impact on the loading time or execution of WebotsJS?

BenjaminDeleze:
> Will there be an impact on the loading time or execution of WebotsJS?

No there will be no difference for the execution.

The only things that change is that:
-  when streaming from Webots desktop, all `SolidDevices` are turned in `Transform` and all `JointDevices` are ignored.

- when displaying a `.proto`, we keep the different `SolidDevices` nodes as `SolidDevice` and we also get the `JointDevices`.

So what could have an impact on the streaming/animation is that there is some additional conditions in the `Parser.js` to detect the type of the node.

I made some tests with `nao_room.wbt`, `complete_apartement.wbt` and `spot.wbt` and could not see a real impact on the loading time.

resources/web/wwi/Parser.js
```diff
@@ -1,35 +1,68 @@
 import {M_PI_4} from './nodes/utils/constants.js';
+import WbAccelerometer from './nodes/WbAccelerometer.js';
+import WbAltimeter from './nodes/WbAltimeter.js';
 import WbAbstractAppearance from './nodes/WbAbstractAppearance.js';
 import WbAppearance from './nodes/WbAppearance.js';
```
ad-daniel, line 5:
```suggestion
import WbAbstractAppearance from './nodes/WbAbstractAppearance.js';
import WbAccelerometer from './nodes/WbAccelerometer.js';
import WbAltimeter from './nodes/WbAltimeter.js';
import WbAppearance from './nodes/WbAppearance.js';
```

ad-daniel: wrong order

resources/web/wwi/Parser.js
```diff
@@ -1,35 +1,68 @@
...
 import WbBackground from './nodes/WbBackground.js';
+import WbBallJoint from './nodes/WbBallJoint.js';
+import WbBallJointParameters from './nodes/WbBallJointParameters.js';
 import WbBillboard from './nodes/WbBillboard.js';
 import WbBox from './nodes/WbBox.js';
+import WbBrake from './nodes/WbBrake.js';
+import WbCamera from './nodes/WbCamera.js';
 import WbCapsule from './nodes/WbCapsule.js';
 import WbCadShape from './nodes/WbCadShape.js';
```
ad-daniel, line 14:
```suggestion
import WbCadShape from './nodes/WbCadShape.js';
import WbCamera from './nodes/WbCamera.js';
import WbCapsule from './nodes/WbCapsule.js';
```

resources/web/wwi/Parser.js
```diff
@@ -1,35 +1,68 @@
...
+import WbLightSensor from './nodes/WbLightSensor.js';
+import WbLinearMotor from './nodes/WbLinearMotor.js';
 import WbMaterial from './nodes/WbMaterial.js';
 import WbMesh from './nodes/WbMesh.js';
+import WbPen from './nodes/WbPen.js';
+import WbPositionSensor from './nodes/WbPositionSensor.js';
 import WbPbrAppearance from './nodes/WbPbrAppearance.js';
 import WbPlane from './nodes/WbPlane.js';
 import WbPointLight from './nodes/WbPointLight.js';
 import WbPointSet from './nodes/WbPointSet.js';
```
ad-daniel, line 51:
```suggestion
import WbPbrAppearance from './nodes/WbPbrAppearance.js';
import WbPen from './nodes/WbPen.js';
import WbPlane from './nodes/WbPlane.js';
import WbPointLight from './nodes/WbPointLight.js';
import WbPointSet from './nodes/WbPointSet.js';
import WbPositionSensor from './nodes/WbPositionSensor.js';
```

resources/web/wwi/Parser.js
```diff
@@ -180,7 +219,16 @@ export default class Parser {
...
-    } else {
+    } else if (node.tagName === 'Transform' || node.tagName === 'Robot' || node.tagName === 'Solid' ||
+      node.tagName === 'Accelerometer' || node.tagName === 'Altimeter' || node.tagName === 'Camera' ||
+      node.tagName === 'Charger' || node.tagName === 'Compass' || node.tagName === 'Connector' || node.tagName === 'Display' ||
+      node.tagName === 'DistanceSensor' || node.tagName === 'Emitter' || node.tagName === 'GPS' || node.tagName === 'Gyro' ||
+      node.tagName === 'InertialUnit' || node.tagName === 'LED' || node.tagName === 'Lidar' ||
+      node.tagName === 'LightSensor' || node.tagName === 'Pen' || node.tagName === 'Radar' || node.tagName === 'RangeFinder' ||
+      node.tagName === 'Receiver' || node.tagName === 'Speaker' || node.tagName === 'TouchSensor' || node.tagName === 'Track' ||
+      node.tagName === 'TrackWheel')
+      result = this.#parseTransform(node, parentNode, isBoundingObject);
```
ad-daniel, line 230: I'm not sure if it's actually better, but since this if is getting kinda unwieldy, doing it with as a switch wouldn't be more clear? Like in this case:

```
switch(node.tagName) {
   ....
   case 'Accelerometer':
   case 'Altimeter':
   case 'Camera':
   <all others>
   case 'TrackWheel':
      result = this.#parseTransform(node, parentNode, isBoundingObject);

``

File: resources/web/wwi/Parser.js [hunk 5]
```diff
@@ -469,21 +517,112 @@ export default class Parser {
...
+        if (node.tagName === 'Robot' || type === 'robot')
+          name = 'robot';
+        else
+          name = 'solid';
+      }
+      newNode = new WbSolid(id, translation, scale, rotation, name);
+    } else if (node.tagName === 'Accelerometer') {
+      if (name === '')
+        name = 'accelerometer';
+      newNode = new WbAccelerometer(id, translation, scale, rotation, name);
```
ad-daniel, line 544:
```suggestion
      if (name === '')
        name = 'accelerometer';
      newNode = new WbAccelerometer(id, translation, scale, rotation, name === '' ? 'accelerometer' : name);
```

ad-daniel: would save us the ifs and the brackets, same for the others

resources/web/wwi/Parser.js
```diff
@@ -558,6 +697,101 @@ export default class Parser {
...
+
+    WbWorld.instance.nodes.set(jointParameters.id, jointParameters);
+
+    if (typeof parentNode !== 'undefined') {
+      if (parentNode instanceof WbBallJoint){
+        if (jointParameters instanceof WbBallJointParameters)
+          parentNode.jointParameters = jointParameters;
+        else
+          parentNode.jointParameters2 = jointParameters;
+      }else if (parentNode instanceof WbHinge2Joint) {
```
ad-daniel, line 753:
```suggestion
      } else if (parentNode instanceof WbHinge2Joint) {
```

resources/web/wwi/Parser.js
ad-daniel: also I'm not sure to understand why we need for the name to be set, we don't seem to do so for non-devices. I don't believe we use this to distinguish between them (we shouldn't), so I don't understand

ad-daniel (changes requested):
Thank you, just a few doubts/suggestions

resources/web/wwi/Parser.js
BenjaminDeleze: Good idea

resources/web/wwi/Parser.js
BenjaminDeleze: We need the name for the documentation viewer.

ad-daniel (approved):
Thank you!

Issue #3504, opened 2021-07-26
Blocking Webots in fast mode use 100% cpu


ShuffleWire:
**Describe the Bug**
When running in fast mode, with an external controller not running, Webots use 100% of one core.

**Steps to Reproduce**
1. Open the following world :
```
#VRML_SIM R2021c utf8
WorldInfo {
  coordinateSystem "NUE"
}
Viewpoint {
}
Robot {
  controller "<extern>"
}
```

2. Build the following controller : 
```
#include <webots/robot.h>
int main(int argc, char **argv)
{
    wb_robot_init();
    while (wb_robot_step(32) != -1)
    {
    }
    wb_robot_cleanup();
    return 0;
}
```

3. Execute the controller
4. Run the simulation in fast mode
5. Kill the controller
6. The simulation stop (that is expected), but the cpu usage is 100% on one core.

**Expected behavior**
The cpu usage should be near 0%.
This bug is only with fast mode, in normal mode, the cpu load is about 0%

**System**
 - Operating System: Debian 11
 - Build on develop 37d26be17fd33dcb9986a6d1bfc17b8cbffece80

lukicdarkoo:
Did you mean?
```
#VRML_SIM R2021c utf8
WorldInfo {
  coordinateSystem "NUE"
}
Viewpoint {
}
Robot {
  controller "<extern>"
}
```

ShuffleWire:
Yes, I thought I edited the message to fix it... sorry

SammyRamone:
I can confirm this issue. I use the Webots simulator for reinforcement learning with the stable-baselines3 framework. Since the RL process includes phases where the GPU is learning or evaluation is done in a different environment, there are a lot of phases where the controller is not doing anything. Still the simulator is taking up a complete core. This is obviously very unfortunate for performance and power consumption reasons.
I circumvented the issue by using the normal mode. Since my simulation is so complex that it does not simulate in more than real time, it does not pose an issue for me. Still, I think that this issue should be fixed, as it is a typical scenario in learning to not run the controller all the time.
In normal mode, the CPU load behaves as expected and only gets high if the simulator is actually running.
This can also be tested by using an external controller which just calls time.sleep(100) or something similar.

omichel:
I confirm the issue as well. It happens not only with an extern controller, but also with a regular controller.

<details>
<summary>fast.wbt</summary>

```js
#VRML_SIM R2022b utf8
WorldInfo {
}
Viewpoint {
}
[...]

PR #5329, opened 2022-10-14
Fix the external controller connection

ShuffleWire:
Was causing a loop eating 100% CPU.
Fix https://github.com/cyberbotics/webots/issues/3504

It is the correct way to fix it ? It seems to work, but I don't have enough knowledge of the code to say...
I've just noticed that the behavior between `void WbController::start()` (and the print "waiting for connection...") and `void WbController::disconnected()` (and the print "disconnected, waiting for new connection.") was different regarding this issue.

Hence, I've duplicate the code from `disconnected()`, which gave the correct behavior to `start()`.


ygoumaz:
The fix doesn't seem to correct the bug you reported [in this comment](https://github.com/cyberbotics/webots/issues/3504#issuecomment-1276227746), namely:
- Start simulation with extern controller in realtime -> 0% CPU usage
- Activate fast mode -> 100% CPU usage 

Is this the bug you are trying to fix? As you reported this is only true before the controller starts. Once started, and if killed, the CPU goes back to 0% as it should.

ygoumaz:
I tested a little more, and found an additional interesting element. The bug can be extended to the following scenario.
- Start Webots in realtime mode with an extern controller no yet connected.
- Activate fast mode (still no controller) ->100% CPU usage (bug).
- Connect the controller.
- Pause the simulation -> 0% CPU usage (correct behavior)
- Kill the controller -> 0% CPU usage (correct behavior)
- Without quitting Webots, put the simulation back to realtime mode -> still 0% CPU usage (correct behavior)
- Now activate the fast mode again. The CPU goes back to 100% even if no controller is connected (bug)

It seems to be the switch between realtime and fast modes when waiting for an extern controller that is the source of the problem.

ShuffleWire:
Well ! 
You right, my fix didn't work... I don't know what test I've made, but that obviously don't work !
So "Is this the bug you are trying to fix? " Yes.
Your last comment is correct, however I'm not sure about the conclusion. To me it's look like webots is actively looking for a controller, hence eat 100% cpu.
The thing is that it does that in real mode as well, but, because the loop is slow down using the QTimer to maintain real time speed, it's not actually noticeable.

ygoumaz:
I will also investigate to find if we can remove this CPU load.

ygoumaz:
The source of the issue is the following:
- The controllers lists are not making any difference between connected and unconnected extern controllers. 
- Because of that, pressing the play button is moving the extern controller to the "started" list (called `mControllers`).
- The controller is started and `pauseStepTimer()` is called to pause the timer.
- Pressing the fast mode button is restarting the timer in the background, but because the controller is already started, the `pauseStepTimer()` method is never called, causing the `step()` function to be called in loop and use 100% CPU usage.

The proposed solution is the following:
The `mExternControllers` list (which was listing all unstarted extern controllers) is replaced by the `mDisconnectedExternControllers` list which tracks all extern controllers which are waiting for a connection. This way, if a step is executed and `mDisconnectedExternControllers` contains a controller, the timer is paused. 
The TCP server can also benefit from this cleanup.

src/webots/control/WbControlledWorld.hpp
```diff
@@ -42,7 +42,7 @@ class WbControlledWorld : public WbSimulationWorld {
   void step() override;
 
   QList<WbController *> controllers() const { return mControllers; }
-  QList<WbController *> externControllers() const { return mExternControllers; }
+  QList<WbController *> externControllers() const { return mDisconnectedExternControllers; }
```
omichel, line 45: Shall we also rename this method?
```suggestion
  QList<WbController *> disconnectedExternControllers() const { return mDisconnectedExternControllers; }
```

ygoumaz: I think it is a good idea, it will be more explicit.

omichel (approved):
Thank you.

PR #5333, opened 2022-10-17
WebotsJS:  add support for updates

BenjaminDeleze:
**Description**
Until now, only a small subset of update were supported (translation, rotation, scale, color).
To be able to play with protos on webots.cloud, we need to support every type of updates and this is the aim of this PR.

- [x] Geometries
  - [x] Box
  - [x] Cone
  - [x] Capsule
  - [x] Cylinder
  - [x] Elevation grid
  - [x] Indexed line set
  - [x] Indexed face set
  - [x] Plane
  - [x] Sphere
  - [x] Point set
  - [x] Mesh 
- [x] Lights
  - [x] Generic light
  - [x] Point light
  - [x] Spot light
  - [x] Directional light
- [x] Appearances
  - [x] Appearance
  - [x] PBRAppearance
  - [x] ImageTexture
  - [x] TextureTransform
  - [x] Material
- [x] CadShape
- [x] Transform
- [x] Fog
- [ ] Track
- [ ] TrackWheel
- [ ] ...


ad-daniel (changes requested):
Overall looks very nice, I didn't test it extensively since we'll do so as we test the different PROTO later on, but those examples I had on hand worked

ad-daniel (approved):
Thank you!

PR #5337, opened 2022-10-17
Fix black buttons in RangeFinder overlay

stefaniapedrazzi:
Resize and close buttons in RangeFinder overlay are sometimes completely black

Screenshot from 2022-10-17 15-53-47

It seems that this is because the `mix` function cannot properly handle the values (like FLT_MAX) of the RangeFinder image pixel color.
For the close and resize buttons where the alpha channel is 1 or 0 it is simpler to not mix the colors.


BenjaminDeleze (approved):
Thank you, it fixes the bug, but I think that the close button is not very visible when the top part of the image is white.

stefaniapedrazzi:
> Thank you, it fixes the bug, but I think that the close button is not very visible when the top part of the image is white.

Yes, depending on the background color (if it is very similar to the border color) the buttons are not always very visible.
This occurs more often with the range-finder where it is very likely that pixels in the top are white.
I will try to improve it in a different PR.

PR #5341, opened 2022-10-18
Fix PROTO declaration error when copy and pasting inserted nodes

### LINKED ISSUE #5325: Error missing EXTERNPROTO declaration when copying and pasting PROTO node
**Steps to Reproduce**
1. Open world accelerometer.wbt`
2. Add a PROTO from the Webots library, for example `Chair` or `Table`
3. Copy and paste the PROTO node
4. See error:
    `ERROR: Missing declaration for 'Chair', add: 'EXTERNPROTO "../../../objects/chairs/protos/Chair.proto"' to '/home/stefi/webots/projects/samples/devices/worlds/accelerometer.wbt'.`

This is also reproducible on the released R2022b.

stefaniapedrazzi:
Fix  #5325:
generalize the `WbExternProto::isFromRootNodeConversion()` flag to apply for any user action (and including the node insertion) so that the PROTO declaration is correctly found also when copying and pasting a PROTO node inserted in the world from the add node dialog.

### DIFF
--- a/docs/reference/changelog-r2022.md
+++ b/docs/reference/changelog-r2022.md
@@ -5,12 +5,12 @@
    - Fixed controller restart after crash ([#5284](https://github.com/cyberbotics/webots/pull/5284)).
    - Fixed the export of [Lidar](lidar.md)'s rotating head to X3D ([#5224](https://github.com/cyberbotics/webots/pull/5224)).
    - Fixed behavior of `WbLightSensor::computeLightMeasurement` when spotlight is rotated ([#5231](https://github.com/cyberbotics/webots/pull/5231)).
-   - Fixed the reset of the viewpoint in animation when the follow is activated ([5237](https://github.com/cyberbotics/webots/pull/5237)).
-   - Fixed a recursion bug in web animation ([5260](https://github.com/cyberbotics/webots/pull/5260)).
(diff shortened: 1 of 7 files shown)


ad-daniel (changes requested):
Behavior-wise, the only case I got a warning is when I did:

1. add Chair
2. add Table
3. cut Table (ctrl + x)
4. copy Chair
5. paste Chair <--- warning

Since it involves a cut action it might be unrelated to the initial issue.

I'm also a bit unsure about a change, I haven't been able to break it but seems potentially dangerous

stefaniapedrazzi:
I was not sure about what is the better way to fix this because there were two alternatives.
Currently there are two different mechanisms implemented that almost do the same:
* add a flag in WbExternProto for nodes converted base nodes
* buffer storing the cut PROTO declarations

For the copy of PROTO nodes both methods can be used to avoid the warning.
Initially I fixed it using the "cut" buffer and it worked just fine but it requires more computation for each copy.
I will revert to this option.

I would also check if this buffer can also be used for conversion, as it seems better to have a single mechanism instead of many different ones.

ad-daniel (approved):
Definitely cleaner this way, and in my tests I haven't managed to break it. Thank you!

Issue #5317, opened 2022-10-11
Webots should warn when saving a world file outside a worlds folder


omichel:
Some users are having problems due to the fact they saved a world file outside of a `worlds` folder (see below).
To avoid this kind of situation, we should probably warn the user when they are attempting to save a world file in a folder which is not named `worlds`. This may happen from the "File / Save World As..." menu.

> I was having this issue when my world file was under my user directory without the Webots's project directory structure. I was able to solve this issue by first creating a project directory, and putting my world file under 'worlds' directory.

_Originally posted by @tuncalie in https://github.com/cyberbotics/webots/issues/5292#issuecomment-1273951622_

PR #5350, opened 2022-10-19
Improve warnings and suggested path in save and wizard dialogs

stefaniapedrazzi:
1. A user reported issues when creating a new project (see https://github.com/cyberbotics/webots/discussions/5339).

    It seems that the proposed path on some system is not valid and Webots is not able to create the project directories.
    Instead of proposing and invalid path we should check if the suggested path is writable and empty the field in case no valid suggestion are found so that the user have to provide it.

2. Address #5317: display a warning if world file is saved in a folder not named "worlds"

3. The New World wizard currently accepts a path in the world name field but then it cannot create the required folders: display a warning preventing to specify a path containing folders


src/webots/gui/WbMainWindow.cpp
```diff
@@ -1466,6 +1466,13 @@ void WbMainWindow::saveWorldAs(bool skipSimulationHasRunWarning) {
     return;
   }
 
+  const QDir dir(fileName);
+  if (dir.dirName() != "worlds")
+    WbMessageBox::warning(
+      tr("The selected save directory for the world file is not named \"worlds\" and it is thus not located in a valid Webots "
+         "project.\nPlease note that some project-related functionalities may not work."),
```
omichel, line 1473:
```suggestion
      tr("The selected directory for saving the world file is not named \"worlds\".\nThus it is not located in a valid Webots "
         "project.\nAs a consequence, some project-related functionalities may not work."),
```

src/webots/gui/WbNewProjectWizard.cpp
```diff
@@ -44,9 +44,15 @@ WbNewProjectWizard::~WbNewProjectWizard() {
...
-  if (WbProject::current()->isReadOnly())  // propose a new project in user's home dir
-    path = WbPreferences::instance()->value("Directories/projects").toString() + "my_project";
-  else {  // otherwisepropose new project dir as sibling of current project
+  if (WbProject::current()->isReadOnly()) {  // propose a new project in user's home dir
+    path = WbPreferences::instance()->value("Directories/projects").toString();
+    if (!path.isEmpty()) {
+      if (WbFileUtil::isDirectoryWritable(path))
+        path += +"/my_project";
+      else
+        path = QString();  // no valid default path found
```
omichel, line 53:
```suggestion
        path = "";  // no valid default path found
```

omichel (changes requested):
When saving a world file with "Save World As...", we should add a "Cancel" button (and rename the OK button to something like "Save here anyway") to allow the user to change the path where they will save the world file:

image

Also currently, if instead of pressing the "OK" button, we close the window, then the file is saved as if we would have pressed the "OK" button, which seems not very nice.

omichel (approved):
Thank you.

PR #5353, opened 2022-10-20
Fix ChangeLog indent

stefaniapedrazzi:
In R2022b-rev1 changelog the indent in only by 1 space instead of 2 (as in the rest of the file).


omichel (approved):
Well spotted!

PR #5366, opened 2022-10-24
Replace deprecated QPair

### LINKED ISSUE #5349: QPair deprecated in Qt 6
Since Qt 6.0 the `QPair`  class has been deprecated and replaced by `std::pair`.
Old code is still working because `QPair` is currently and alias but it would be better to replace all the occurrences.

stefaniapedrazzi:
Fix #5349


ad-daniel (approved):
Thank you!

Issue #3420, opened 2021-07-14
Robot window doesn't update after restarting an external controller


lukicdarkoo:
**Steps to Reproduce**
1. Open `projects/languages/python/worlds/example.wbt`
2. Change controller to `<extern>`
3. Open the robot window
4. Run the `slave.py` extern controller (see the robot window getting the update)
5. Kill and start the extern controller (no update robot window updates)

**System**
 - Operating System: Linux Ubuntu 20.04

stefaniapedrazzi:
This issue is still occurring in R2022b.
After killing and restarting the extern controller the robot window is not automatically updated.
The user has to reload the robot window page to make it work again.

stefaniapedrazzi:
We discussed that the best solution is probably to close the robot window when the extern controller disconnects.

PR #5367, opened 2022-10-24
Reopen robot window when extern controller reconnects

stefaniapedrazzi:
Addresses  #3420:
close and reopen a new robot window when a new extern controller connects.


ygoumaz (commented):
In the sequence:

1. Open robot window
2. Start extern controller
3. Reset simulation (extern controller gets automatically killed)
4. Restart extern controller

the robot window is reopened at step 3 and 4. Is it possible to avoid one of these reopenings?

Otherwise, this seems the best solution to fix the issue.

stefaniapedrazzi:
Should now be fixed and the robot window is closed when a new extern controller connects only if the simulation has started.

ygoumaz (approved):
Works nicely, thank you!

PR #5368, opened 2022-10-24
Webots.js: remove decimal points from progress value

### LINKED ISSUE #5288: Webots.js: too high precision for progress percentage
It seems that the progress percentage is using too high precision.
This is not very useful and makes difficult to understand what this number `96.2962962629629` means.
Tested on `testing.webots.cloud`:
image

stefaniapedrazzi:
Fix #5288


BenjaminDeleze (approved):
Thank you

PR #5370, opened 2022-10-25
Extend proto visualizer

ad-daniel:
**Description**

Tasks:
- [x] Handle indirect regeneration from interface
- [x] Add export logic that writes .proto based on current parameters
- [x] Add pose-writer (and rename it since now it supports more than just pose) that encodes interface information into a format readable by applyPose (to do the change on the webotsJS side)
 
Less priority:
- [ ] Add role for tracks
- [ ] Fix issues with PROTO that don't load
- [ ] Optimize PROTO to speed up loading
   - [ ] VehicleWheel: store coords in buffer


BenjaminDeleze (approved):
Seems good to me

Issue #5371, opened 2022-10-26
GPS initial speed


ShuffleWire:
When using a GPS, the first (vector) speed return is not correct.
The explanation is straightforward : 
1) in the code WbPGS.cpp, the initial value of `mPreviousPosition` used to estimate the speed don't seems to be initialized (or maybe the constructor set it to zero by default ?)
2) In the reset function, it's clearly resetted to 0

in both case, that yield to wrong speed estimation when the robot is not actually at the 0,0,0 coordinate at the start, (which is almost always the case, robot are moved anywhere in the world)
The correct way to do it it either :  
 - Initialize this value `mPreviousPosition` to the first position of the robot before starting the simulation, hence, the first speed should be 0,0,0, but it's not that bad. However, I don't know if we have access to this info at that moment.
 - Intialize `mPreviousPosition` to nan, and then do something like
 - 
```
if (!mPreviousPosition.isNan())
    mSpeedVector = (t - mPreviousPosition) * 1000.0 / mSensor->elapsedTime();
  else
    mSpeedVector = WbVector3();
```
I could do the fix, but let me know what is the best way to go. 

In the end, the best way to get the GPS speed would be to get the speed from ode directly, but this could be more work...

omichel:
A real GPS usually computes the speed from the change of position only (there is nothing similar to getting the speed from ODE). This means that a GPS needs at least two iterations to compute the speed.
As a consequence of this:
1. It is acceptable to return NAN if the position history is not sufficient. Hence on the first step or after a reset, the GPS should return a NAN value for the speed.
2. The GPS should not get the speed from ODE as this sounds like cheating, e.g., doing something that a real GPS couldn't do. Moreover, if the GPS position is noisy, the speed position should be noisy as well (since its computation rely on the position), which is realistic.

ShuffleWire:
Actually many GPS receiver deduce the speed using Doppler measurement, which give them the speed independently of the (previous/current/next) position.

I will take a moment to think about it

stefaniapedrazzi:
Addressed in #5372.

PR #5372, opened 2022-10-26
Fix gps speed

ShuffleWire:
Discussed in https://github.com/cyberbotics/webots/issues/5371


omichel:
@ShuffleWire: what is the status of this PR? Are you planning to complete it soon?

ShuffleWire:
> @ShuffleWire: what is the status of this PR? Are you planning to complete it soon?

It depends on what we think about https://github.com/cyberbotics/webots/issues/5371#issuecomment-1293248987

omichel:
I believe we could implement a simple version first with the good material already present in this PR (with the NAN values in particular, and getting the velocity from ODE).
I am not sure that adding a different noise for the velocity would be super helpful.
It could be done in a subsequent PR if there is a real use-case for this.

ShuffleWire:
changed "unknow" speed to NAN (initially, when no solid merger is detected)
is that correct for you ?

as it, for most of the case (ie robot with a physic node), we should get accurate ODE speed, and not wrong values at the beginning.

I feel like I should document the case that yield NAN at the start when using a GPS on a non physic robot, but is it actually a useful case ? could it be that upperSolid()->bodyMerger() yield 0 ?

docs/reference/gps.md
```diff
@@ -169,9 +169,9 @@ The returned vector indicates the absolute position of the [GPS](#gps) device.
 This position can either be expressed in the cartesian coordinate system of Webots or using latitude-longitude-altitude, depending on the value of the `gpsCoordinateSystem` field of the [WorldInfo](worldinfo.md) node.
 The `gpsReference` field of the [WorldInfo](worldinfo.md) node can be used to define the reference point of the GPS.
 
-The `wb_gps_get_speed` function returns the current [GPS](#gps) speed in meters per second.
+The `wb_gps_get_speed` function returns the current [GPS](#gps) speed in meters per second. If there is no physic node on the robot, the first returned value will be NAN.
```
stefaniapedrazzi, line 172: Here are some changes to make it more consistent with the existing documentation
```suggestion
The `wb_gps_get_speed` function returns the current [GPS](#gps) speed in meters per second.
If there is no physics node on the robot, the first returned value will be *NaN*.
```

docs/reference/gps.md
```diff
@@ -169,9 +169,9 @@ The returned vector indicates the absolute position of the [GPS](#gps) device.
...
 
-The `wb_gps_get_speed_vector` function returns the current [GPS](#gps) speed vector in meters per second.
+The `wb_gps_get_speed_vector` function returns the current [GPS](#gps) speed vector in meters per second. If there is no physic node on the robot, the first returned vector will be NAN.
```
stefaniapedrazzi, line 174: Here are some changes to make it more consistent with the existing documentation
```suggestion
The `wb_gps_get_speed_vector` function returns the current [GPS](#gps) speed vector in meters per second.
If there is no physics node on the robot, the first returned vector will be *NaN*.
```

docs/reference/gps.md
stefaniapedrazzi, line 172: Note that the sentence `If there is no physics node on the robot` is wrong doesn't match the code.
This new code only checks if the closest parent Solid has physics not if an ancestor or top Robot node has physics.

stefaniapedrazzi (changes requested):
Thank you also for fixing the typos!

Note that the ODE speed is only used if the closest parent Solid node of the GPS has a Physics node but not if any other ancestor Solid/Robot nodes have physics.
So, the new lines added to the document don't seem correct.

Missing changes:
* update the `tests/api/worlds/gps_speed.wbt` test and include also the test for the new computation in case of robot window phyiscs.
* mentioned this change in the ChangeLog:
    https://github.com/cyberbotics/webots/blob/master/docs/reference/changelog-r2022.md

docs/reference/gps.md
ShuffleWire: and what is the correct way to do it ? I would think that having an ancestor or top Robot node with physics is enough, it don't need to be the direct parent. I'm right ? or is the code correct and the documentation to be changed ?

stefaniapedrazzi: The code, i.e. checking only for the closest Solid, is consistent with the Accelerometer device so it would be fine to keep it like this if it is properly documented.
To change the code it should be checked if it is possibile to retrieve the speed from ODE in all the case where an ancestor node has physics, including cases where there are intermediate joint nodes.

ShuffleWire: ok, change the doc from "robot" to "parent node" in the "If there is no physic node on the robot, the first returned value will be NAN." sentence

ShuffleWire: i let you resolve once it's good for you

stefaniapedrazzi:
Thank you! Documentation is fine now.

It just remains to update the `tests/api/worlds/gps_speed.wbt` test.

ShuffleWire:
> Thank you! Documentation is fine now.
> 
> It just remains to update the `tests/api/worlds/gps_speed.wbt` test.

I've taken the opportunity to "improve" the test by building an little routine in charge of making the repetitive test, instead of this code duplication. what do you think ?

also, it doesn't test the case were the robot have a physic node, hence get the speed value straight at the beginning. Should be add another test world ?

tests/api/controllers/gps_speed/gps_speed.c
```diff
@@ -6,45 +6,31 @@
 
 #define TIME_STEP 32
 
+void check_speed(WbDeviceTag gps, float v, float vx, float vy, float vz, char *end_msg) {
+  double speed = wb_gps_get_speed(gps);
```
stefaniapedrazzi, line 10:
```suggestion
  const double speed = wb_gps_get_speed(gps);
```

tests/api/controllers/gps_speed/gps_speed.c
```diff
@@ -6,45 +6,31 @@
...
-    r[0], r[1], r[2]);
-
-  wb_robot_step(2 * TIME_STEP);
+  check_speed(gps, NAN, NAN, NAN, NAN, "before a wb_robot_step is performed");
+  wb_robot_step(TIME_STEP);
 
-  speed = wb_gps_get_speed(gps);
-  ts_assert_double_in_delta(speed, 5.0, 0.0001, "The speed value measured by the GPS should be 5.0 and not %lf after two steps",
-                            speed);
+  check_speed(gps, NAN, NAN, NAN, NAN, "after one step (because no physic node)");
```
stefaniapedrazzi, line 30:
```suggestion
  check_speed(gps, NAN, NAN, NAN, NAN, "after one step (because no Physics node)");
```

tests/api/controllers/gps_speed/gps_speed.c
```diff
@@ -6,45 +6,31 @@
 
 #define TIME_STEP 32
 
+void check_speed(WbDeviceTag gps, float v, float vx, float vy, float vz, char *end_msg) {
+  double speed = wb_gps_get_speed(gps);
+  ts_assert_double_equal(speed, v, "The speed value measured by the GPS should be %f and not %lf %s.", v, speed, end_msg);
```
stefaniapedrazzi, line 11: Double values comparison could depend on the system.
Instead of checking if two values are equals, you should use the `ts_assert_double_in_delta` and `ts_assert_vec3_in_delta` functions instead.

stefaniapedrazzi (changes requested):
It is good to add a function to avoid duplication.

Yes, we should also add tests for the case where the Physics node is defined.
Instead of adding a new test world, a new Robot node should be added in the existing test.
The existing controller could be reused for both case (with/without Physics) by passing a controller argument that switch the expected speed values.

ShuffleWire:
> Yes, we should also add tests for the case where the Physics node is defined. Instead of adding a new test world, a new Robot node should be added in the existing test. The existing controller could be reused for both case (with/without Physics) by passing a controller argument that switch the expected speed values.

I don't get the "switch the expected speed values" part : I've got hard coded values different for each case, but at the end of the day they are check both at the same time

stefaniapedrazzi:
What I meant was to add a new Robot node instead of adding the GPS to the already existing one.
In this case to distinguish between the Robot with and without Physics is was possible to simply set the `controllerArgs` field and use it in the controller to switch the expected speed values.

But if the test is working properly and both implementations are tested even if the two GPS devices are included in the same Robot node, then it is also fine.

ShuffleWire:
> What I meant was to add a new Robot node instead of adding the GPS to the already existing one. In this case to distinguish between the Robot with and without Physics is was possible to simply set the `controllerArgs` field and use it in the controller to switch the expected speed values.
> 
> But if the test is working properly and both implementations are tested even if the two GPS devices are included in the same Robot node, then it is also fine.

oh ok ! I found easier to just move each gps in their own slider + solid and move those solid, and keep the robot actually immobile. Should works like that, I think.

tests/api/worlds/gps_speed.wbt
```diff
@@ -5,34 +5,60 @@ EXTERNPROTO "webots://tests/default/protos/TestSuiteSupervisor.proto"
...
+    SliderJoint {
+      endPoint Solid {
+        children [
+          Shape {
+            geometry Box {
+              size 0.1 0.1 0.1
+            }
+          }
+          GPS {
+            name "gps_with_physic"
```
stefaniapedrazzi, line 49:
```suggestion
            name "gps_with_physics"
```

tests/api/worlds/gps_speed.wbt
```diff
@@ -5,34 +5,60 @@ EXTERNPROTO "webots://tests/default/protos/TestSuiteSupervisor.proto"
...
+          }
+        ]
+        name "solid_with_physic"
```
stefaniapedrazzi, line 52:
```suggestion
        name "solid_with_physics"
```

tests/api/worlds/gps_speed.wbt
```diff
@@ -5,34 +5,60 @@ EXTERNPROTO "webots://tests/default/protos/TestSuiteSupervisor.proto"
...
+          Shape {
+            geometry Box {
+              size 0.1 0.1 0.1
+            }
+          }
+          GPS {
+            name "gps_without_physic"
+          }
+        ]
+        name "solid_without_physic"
```
stefaniapedrazzi, line 36:
```suggestion
        name "solid_without_physics"
```

tests/api/worlds/gps_speed.wbt
```diff
@@ -5,34 +5,60 @@ EXTERNPROTO "webots://tests/default/protos/TestSuiteSupervisor.proto"
...
+      }
+      endPoint Solid {
+        children [
+          Shape {
+            geometry Box {
+              size 0.1 0.1 0.1
+            }
+          }
+          GPS {
+            name "gps_without_physic"
```
stefaniapedrazzi, line 33:
```suggestion
            name "gps_without_physics"
```

tests/api/controllers/gps_speed/gps_speed.c
```diff
@@ -4,47 +4,39 @@
...
-
-  r = wb_gps_get_speed_vector(gps);
-  ts_assert_vec3_in_delta(
-    r[0], r[1], r[2], e[0], e[1], e[2], 0.0001,
-    "The speed vector value measured by the GPS should be [%f, %f, %f] not [%f, %f, %f] after %d wb_robot_step(s)", e[0], e[1],
-    e[2], r[0], r[1], r[2], 2);
+  const int time_step = wb_robot_get_basic_time_step();
+
+  WbDeviceTag gps_with_physic = wb_robot_get_device("gps_with_physic");
+  WbDeviceTag gps_without_physic = wb_robot_get_device("gps_without_physic");
```
stefaniapedrazzi, line 23:
```suggestion
  WbDeviceTag gps_with_physics = wb_robot_get_device("gps_with_physics");
  WbDeviceTag gps_without_physics = wb_robot_get_device("gps_without_physics");
```

stefaniapedrazzi (changes requested):
You should also add this include in the test that is currently throwing a warning
```
#include <webots/device.h>
```

Then the test is failing on my system with the error:
```

FAILURE with gps_speed: The speed vector value measured by the GPS "gps_with_physic" should be [-5.000000, 0.000000, 0.000000] not [0.000000, -0.000098, -5.000000] after one step (because got a Physics node).
```

stefaniapedrazzi (approved):
Thank you!

PR #5375, opened 2022-10-27
Add support for 5+ vertices polygon

BenjaminDeleze:
For example the `shrimp`

This was not needed before as the tessellation and everything was done in Webots, but now it is done in WebotsJS


omichel (approved):
Looks good to me.

PR #5376, opened 2022-10-27
Implement first use case of webots.cloud/proto

BenjaminDeleze:
A user needs a new size of `SoccerBall.proto`
A competition organizer want to develop a soccer competition. It founds the `SoccerBall.proto` in Webots, however the size is a bit too small to work well with the robots he plans to use as players.

He has four solutions to fix his problem:

1. Recreate a new proto from scratch.
2. Edit the `.proto` in a text editor.
3. Use the proto wizard.
4. Use the proto editor of `webots.cloud.

Solution 1 requires some knowledge about how to write proto and this user does not have time to learn them.
Solution 2 is a bit simpler but also require a minimum of knowledge and the interface is not very user-friendly.
Solution 3 is good, but solution 4 is better because the user can see the effect of its modifications in real-time in the viewer of webots.cloud.

For solution 4 to work, it needs:
  - [x] A window to change the radius
     - [x] display `radius` and an `input` field.
     - [x] WebotsJS must support `radius` update.
     - [x] Changing the `radius` must change the visual result.
  - [x] A download button to download the proto with the modified radius.
  - [x] Have a nice user interface:
    - [x] The proto parameter must be open and well positioned when the user first open the proto
    - [x] The download button shoud be displayed
    - [x] The reset button must be implemented to restore the state of the field `radius`

You can see the result here: https://proto.webots.cloud/run?version=proto&url=https://github.com/BenjaminDeleze/sandbox/blob/master/protoExample/protos/SoccerBall.proto


omichel (changes requested):
It is not possible to set a name for the saved proto, it is named CustomProto.
The name field has quotes, e.g., `"`. I believe they should be removed.
The "reset to default value" button should have a tooltip.
The layout should align the field name and field values (e.g., table). I believe that would be more readable especially when the number of exposed parameters is high.

omichel:
Also, I believe we could add a "Download World" button that would download a world file that includes the original proto with the parameters currently set by the user (as well as a light, floor, etc. to match the web view).

Also, maybe the "Download Proto" button could be renamed to "Download Derived Proto", which sounds clearer to me.

BenjaminDeleze:
> It is not possible to set a name for the saved proto, it is named CustomProto.

The decision to open or not the `save as` pop up is taken by the browser so we can really control it. 
User can force their browser to always open the pop up (for example in `Firefox`: `settings/Always ask you where to save files`)

There is one solution to that, it is to use the [File System Access API](https://web.dev/file-system-access/#create-a-new-file). However this feature is quite new and even if is supported in most Chromium base browsers (to some extend), it is on the technology preview step on Safari and not available at all on Firefox. (https://caniuse.com/native-filesystem-api), so I am not totally convinced.

Another solution would be to make a custom field/pop up where the user could choose the name (but not the location) of the file she wants to save.

> The layout should align the field name and field values (e.g., table). I believe that would be more readable especially when the number of exposed parameters is high.

Should the reset buttons also be align or should they be at the end of the field?

> Also, I believe we could add a "Download World" button that would download a world file that includes the original proto with the parameters currently set by the user (as well as a light, floor, etc. to match the web view).

I am not convinced that it is really useful.

BenjaminDeleze:
I think I fixed all the reported issues.

In addition I removed the custom tool-tips to replace them by the browser provided ones.

resources/web/wwi/ProtoManager.js
```diff
@@ -31,23 +31,27 @@ export default class ProtoManager {
...
       // test this using the world: DemoRegeneration.proto in the html
       // setTimeout(() => this.demoRegeneration(), 2000);
 
       // test this using the world: DemoFieldChange.proto in the html
-      setTimeout(() => this.demoFieldChange('color', {r: 0, g: 0, b: 1}), 2000);
-      setTimeout(() => this.demoFieldChange('translation', {x: 0, y: 0, z: 0.5}), 2000);
-      setTimeout(() => this.demoFieldChange('rotation', {x: 0, y: 0, z: 1, a: 0.785}), 2000);
+      // setTimeout(() => this.demoFieldChange('color', {r: 0, g: 0, b: 1}), 2000);
+      // setTimeout(() => this.demoFieldChange('translation', {x: 0, y: 0, z: 0.5}), 2000);
+      // setTimeout(() => this.demoFieldChange('rotation', {x: 0, y: 0, z: 1, a: 0.785}), 2000);
```
omichel, line 47: Can we remove these now?
```suggestion
```

resources/web/wwi/ProtoManager.js
```diff
@@ -31,23 +31,27 @@ export default class ProtoManager {
...
 
       // not implemented yet JS side
       // setTimeout(() => this.demoFieldChange('radius', 0.2), 2000);
       // setTimeout(() => this.demoFieldChange('subdivision', 5), 2000);
 
       // print output proto
-      setTimeout(() => this.exportProto(), 4000);
+      // setTimeout(() => this.exportProto(), 4000);
```
omichel, line 54:
```suggestion
```

resources/web/wwi/FloatingProtoParameterWindow.js
```diff
@@ -0,0 +1,296 @@
...
+
+    const value = document.createElement('p');
+    value.style.gridRow = '' + row + ' / ' + row;
+    value.style.gridColumn = '3 / 3';
+    value.className = 'value-parameter';
+
+    const input = document.createElement('input');
+    input.type = 'text';
+
+    let string = parameter.value.value;
```
omichel, line 199:
```suggestion
    const string = parameter.value.value;
```

resources/web/wwi/FloatingProtoParameterWindow.js
```diff
@@ -0,0 +1,296 @@
...
+    return input;
+  }
+
+  #rotationOnChange(node) {
+    let object = {'x': node.inputs[0].value, 'y': node.inputs[1].value, 'z': node.inputs[2].value, 'a': node.inputs[3].value};
+    node.parameter.setValueFromJavaScript(this.#view, object);
+  }
+
+  #vector3OnChange(node) {
+    let object = {'x': node.inputs[0].value, 'y': node.inputs[1].value, 'z': node.inputs[2].value};
```
omichel, line 174:
```suggestion
    const object = {'x': node.inputs[0].value, 'y': node.inputs[1].value, 'z': node.inputs[2].value};
```

resources/web/wwi/FloatingProtoParameterWindow.js
```diff
@@ -0,0 +1,296 @@
...
+    input.onchange = () => callback(parent);
+
+    span.appendChild(input);
+    parent.appendChild(span);
+
+    return input;
+  }
+
+  #rotationOnChange(node) {
+    let object = {'x': node.inputs[0].value, 'y': node.inputs[1].value, 'z': node.inputs[2].value, 'a': node.inputs[3].value};
```
omichel, line 169:
```suggestion
    const object = {'x': node.inputs[0].value, 'y': node.inputs[1].value, 'z': node.inputs[2].value, 'a': node.inputs[3].value};
```

omichel (commented):
I suggested a few const declarations and comments clean-up. Not sure this is relevant. Let me know.

omichel (approved):
Thank you.

PR #5378, opened 2022-10-27
Resolve circular dependencies

### LINKED ISSUE #4981: Circular dependency in webots modules

stefaniapedrazzi:
Fix #4981:
* Create a `WbVrmlNodeUtilities` namespace in the `vrml` module containing the functions from `WbNodeUtilities` that doesn't rely on WbBaseNode classes:
  -> any suggestion is welcome for improving the two namespace names
*  `WbUrl`:
    * move from `nodes/utils` to `vrml` module
    * world filename used to compute relative URLs is passed as a namespace property
* `WbProtoManager`:
    * pass callback function to compute if PROTO needs a robot ancestor
    * pass world root as argument when needed
    * use signal/slot to notify world load completed
    * fix singleton class structure
* `WbParser`
    * use callback to update parsing progress
* `WbDownloader`:
    * move from `nodes` to `core` module
    * create a `WbDownloadManager` singleton class as an entry point to update the application progress dialog
    * remove `mCopy` class variable that is not used: `WbDownloader::download` is always called only once per instance


ygoumaz (approved):
Nice solution, thank you!

PR #5397, opened 2022-10-31
Fix crash when robot with extern controller is deleted

### LINKED ISSUE #5386: Crash when deleting a robot with an extern controller
Webots crashes if the field of a robot is set to `<extern>` and we try to delete it. The simulation must have run (i.e. the controller must be started).

This implies, among other things, that the `multiple robots` example of `webots_ros2_universal_robot` crashes when the simulation is reset.

ygoumaz:
Fixes #5386.


BenjaminDeleze (approved):
Thank you!

PR #5404, opened 2022-10-31
Web proto: second usecase: joints

BenjaminDeleze:
**Description**
A researcher wants to model an office room in Webots before using it to train a ML model.
He search "chair webots" in a well-know search engine and arrived on `webots.cloud/proto` where he can easily compare the different chairs available in Webots. With the help of the joint tab, he will be able to see that the office chair has a rotating seat and observe how it behave.

**Tasks**
  - [x] Add tabs.
  - [x] Parse the node to find all the joints.
  - [x] Populate the `joints` tab with the joints that were found.
  - [x] Make the user able to move the joints.

**Additional Information**
I encounter some circular dependencies issues will implementing this part so I had to refactor a bit the code


BenjaminDeleze:
You can see the result here: https://proto.webots.cloud/run?version=proto&url=https://github.com/BenjaminDeleze/sandbox/blob/master/protoExample/protos/OfficeChair.proto

This PR also include the additional modifications:

- Support for SFInt32 field
- Refactor of several classes/functions to avoid circular dependencies, it removes mainly :
  - `webots`->`X3dScene`->`ImageTexture`->`image_loader`->`webots` and the corresponding one with `mesh_loader`
  It was solved by turning `image_loader` and `mesh_loader` into classes and putting the global variable `mesh_loader` and `image_loader` needed directly in their classes instead of in `webots`. The same process was done to get rid of the dependency of `CadeShape` on `webots`.
  - `WbBaseNode`->`nodes_utilities`->`WbSolid`->`WbTransform`->`WbGroup`.>`WbBaseNode`
  I had a lot of troubles with this one as, if you begin the loop with `WbSolid`, it will work perfectly well. But if you begin by `WbBaseNode` it will fail to load everything. The solution was to move the functions that were using the functions from `nodes_utilities` directly in `WbGeometry` and `WbShape` were they are used.

Finally please note that the office chair have some `SFNode`  fields for its appearances but they are not yet supported so they will not show up in the proto parameter window.

resources/web/wwi/css/toolbar.css
```diff
@@ -973,3 +998,74 @@ webots-view .reset-field-button {
...
+}
+
+.webots-view .proto-slider::-moz-range-thumb {
+  border: none;
+  height: 15px;
+  width: 15px;
+  background: rgb(0, 122, 204);
+  cursor: pointer;
+  border-radius: 12px;
+}
```
ad-daniel, line 1071:
```suggestion
}

```

resources/web/wwi/nodes/utils/math_utils.js
```diff
@@ -1,9 +1,25 @@
-function clampedAcos(value) {
-  if (value >= 1.0)
-    return 0.0;
-  if (value <= -1.0)
-    return 2.0 * Math.PI;
+export function clampedAcos(value) {
```
ad-daniel, line 1: just a detail, but this file is called `math_utils` whereas for the node we call it `node_utilities`, we should pick one of the two

resources/web/wwi/protoVisualizer/Vrml.js
```diff
@@ -436,10 +436,10 @@ export class SFNode extends SingleValue {
       return false;
 
     for (const [parameterName, parameter] of this.value.parameters) {
-      if (!other.parameters.has(parameterName))
+      if (!other.value.parameters.has(parameterName))
         return false;
 
-      const otherParameter = other.parameters.get(parameterName);
+      const otherParameter = other.value.parameters.get(parameterName);
```
ad-daniel, line 442: I'm not sure to understand why this change was necessary. `other` is supposed to be a Node instance (which can be a PROTO or a base node), and in the case it's a PROTO, then the parameters have to be retrieved directly from it. Accessing the value of the node entails means we are retrieving the parameters of the derived PROTO. It might work, but intuitively seems the wrong approach. Do you remember why it was necessary to do so?

ad-daniel (changes requested):
Just details and a doubt

resources/web/wwi/protoVisualizer/Vrml.js
BenjaminDeleze: Because `other.parameters` was undefined in some cases

ad-daniel: that should be the case only if a PROTO without parameters is instantiated, was it with the officechair or another proto?

BenjaminDeleze: Yes, error when trying to download it

ad-daniel: since it's a SFNode, then yes, the `.value` is necessary, I thought we were passing the value itself already

ad-daniel (approved):
Thank you

PR #5413, opened 2022-11-01
Fix isTemplate flag not passed from base PROTO

### LINKED ISSUE #5214: TexturedBox PROTO parameters changes not working
The changes to the `TexturedBox` PROTO `size` field are not applied in the scene until the world is saved and reverted.

stefaniapedrazzi:
Fix #5214:
this is a regression from R2022a introduced in #5103 where the instruction to pass the `isTemplate` flag from the base to the inherited node has been removed.

Additionally I also updated some debug instructions throwing errors during compilation.


omichel (approved):
Thank you.

Issue #5207, opened 2022-09-12
Copy PROTO icon when relocating the remote PROTO file


stefaniapedrazzi:
During the PROTO relocation triggered when editing a remote PROTO file the icon is currently not copied.
Thus, in the Add Node dialog the local copy doesn't have any icon.
Maybe it would be better to also copy the icon file when copying the PROTO file.

PR #5414, opened 2022-11-01
Copy icon when editing cached remote PROTO

stefaniapedrazzi:
Address #5207:
copy icon when editing a web PROTO and copying it on the current project folder.

To avoid duplicating the code already used in `WbAddNodeDialog` to retrieve the PROTO icon the code is moved in a new class `scene_tree/WbProtoIcon` and reused in `WbMainWindow` and `WbAddNodeDialog`.


BenjaminDeleze (approved):
Thank you!

PR #5415, opened 2022-11-02
Removed obsolete dependency

omichel:
Webots doesn't depend any more on this DLL (probably due to some other DLL dependency change) and it doesn't get installed in the GitHub CI machine, causing the creation of the Windows package to fail.
Removing this obsolete dependency should fix the package creation on Windows in the CI.


BenjaminDeleze (approved):
Thank you

PR #5420, opened 2022-11-04
Update axis orientation in wrl

BenjaminDeleze:
When the FLU conversion took place we forgot to change the description of `Cone`/`Cylinder`/`Capsule` in the `.wrl` files.


omichel (approved):
Thank you

PR #5430, opened 2022-11-07
Fix SFNode field flagged empty in libController

### LINKED ISSUE #5365: Non-empty field is flagged as empty
**Describe the Bug**

Something might be wrong or inconsistent with the remove_sf API.

1. load [debug.zip](https://github.com/cyberbotics/webots/files/9850301/debug.zip) and run
2. the appearance is changed to asphalt correctly
   -  according to [this](https://github.com/cyberbotics/webots/blob/8b8047ead7f7b7f7ca4ed2bd8e0e9aa1a9b3e0ce/tests/api/controllers/supervisor_import_remove/supervisor_import_remove.c#L167-L171) unit test however, this should not be possible because the field is not empty (contains roughcast)
3. When we try to actually remove the roughcast (uncomment the line), we get `wb_supervisor_field_remove_sf() called for an empty field.`

So what would explain both behaviors is that indeed the field is empty (hence why we don't need to remove the existing node before adding a new, and why it doesn't complain when we import without removing). However the field obviously isn't empty.

EDIT: calling removeSF after importing does work as intended, which once again seems to indicate the field was empty the first time

stefaniapedrazzi:
Fix #5365:
use the WbFieldStruct::count variable in the libController to store if the SFNode value is NULL without having to request the field value.

### DIFF
--- a/docs/reference/changelog-r2023.md
+++ b/docs/reference/changelog-r2023.md
@@ -19,6 +19,7 @@ Released on December, 12th, 2022.
   - New Examples
     - Added a sample world showing how to compute the attitude of a robot from IMU sensors ([#5256](https://github.com/cyberbotics/webots/pull/5256)).
   - Bug Fixes
+    - Fixed wrong warnings due to SFNode empty flag in the controller ([#5430](https://github.com/cyberbotics/webots/pull/5430)).
   - Cleanup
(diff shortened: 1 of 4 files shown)


omichel (approved):
Thank you.

Issue #4320, opened 2022-03-04
Use boolean in shaders instead of integers or floats for flags.


BenjaminHug8:
**Describe the Bug**
Since the PR https://github.com/cyberbotics/webots/pull/4302 it is possible to declare uniform bool for shaders in wren.

However, in some shaders, there are still some flags using integers or floats instead of boolean. To increase the code clarity these flags should be converted in boolean uniforms.

PR #5436, opened 2022-11-08
Cleanup boolean values in shaders

stefaniapedrazzi:
Address #4320:
* [x] use bool instead of float or int for uniform values


ad-daniel (approved):
Looks good, thank you

PR #5439, opened 2022-11-09
Fix Windows nightly builds

### LINKED ISSUE #5438: Nightly build 20222b-rev1: WeBots doesn't start (libpcre2-8-0.dll version incompatibility)
After installing the most recent nightly build (from 7.11.22, I think), for Windows, in the form of EXE installer,
WeBots doesn't start because of a missing DLL (see above).  In the C:\Program Files\Webots\msys64\mingw64\bin there exists what I believe is a newer version of the same DLL (libpcre2-16-0.dll).  The same problem existed in some earlier nightly builds (like 4 days ago, I think) 

(A similar problem with this dll in MSYS was reported years ago here:  https://github.com/msys2/MINGW-packages/issues/4808)

**System**

omichel:
Fix #5438.

### DIFF
--- a/scripts/packaging/files_msys64.txt
+++ b/scripts/packaging/files_msys64.txt
@@ -15,6 +15,7 @@
 /mingw64/bin/libmd4c.dll
 /mingw64/bin/libminizip-1.dll
 /mingw64/bin/libopenal-1.dll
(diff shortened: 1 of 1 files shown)


ygoumaz (approved):
Thank you.

PR #5447, opened 2022-11-09
Stop pending animation on close

Jean-Eudes-le-retour:
**Description**
If an animation recording is started and Webots is closed before it is stopped, the json animation file is not created. This adds a small check to stop the animation correctly when Webots is closed by the user in the GUI, with ctrl-C in the console or by a supervisor.


ad-daniel:
The test sources fails because you have spaces before and after the code you inserted, I suggest you activate the `"files.trimTrailingWhitespace": true,` if you have vscode or something equivalent

ad-daniel (approved):
Thank you, seems to be working as intended. I've also fixed the missing dot.

PR #5448, opened 2022-11-09
Improved error message of empty animation

### LINKED ISSUE #4994: Allow creating an animation without controllers in the scene
**Describe the Bug**

If the world has no controllers and a user records an animation, although the process itself is allowed, nothing gets generated in the end since no robot steps are performed (although the simulation is run). Adding a robot with just a generic controller doesn't appear to do the trick either. 

To an extent it kinda makes sense, since if there are no robot steps, making a scene might make more sense than an animation. However it still feels kinda odd that it's allowed and that nothing is generated at the end of the process. Moreover, although a scene might be more appropriate, I can still see the interest of creating an animation (to have a pre-recording of a environment for example), so perhaps a solution to allow it anyway could be found.

Jean-Eudes-le-retour:
**Description**

When an animation is done in a simulation where nothing happens (for example when \<generic\> controllers are used for the robots), the error message incorrectly says:

> Error: No animation content is available because the simulation did not start

even though the simulation did start.

I changed the error message to be more specific and added a recommendation to use a scene instead

**Related Issues**
This pull-request fixes #4994

### DIFF
--- a/src/webots/engine/WbAnimationRecorder.cpp
+++ b/src/webots/engine/WbAnimationRecorder.cpp
@@ -435,7 +435,8 @@ void WbAnimationRecorder::stopRecording() {
   // write initial state
   out << "{\"time\":0,\"poses\":[";
   if (commandsChangedFromStart.isEmpty()) {
-    WbLog::info(tr("Error: No animation content is available because the simulation did not start."));
+    WbLog::info(tr("Error: No animation content is available because no changes occurred in the simulation. "
+                   "If you just want a 3D environment file, consider exporting a scene instead"));
     return;
   }
   foreach (WbAnimationCommand *command, commandsChangedFromStart) {


ad-daniel (approved):
Thank you

src/webots/engine/WbAnimationRecorder.cpp
```diff
@@ -435,7 +435,8 @@ void WbAnimationRecorder::stopRecording() {
   // write initial state
   out << "{\"time\":0,\"poses\":[";
   if (commandsChangedFromStart.isEmpty()) {
-    WbLog::info(tr("Error: No animation content is available because the simulation did not start."));
+    WbLog::info(tr("Error: No animation content is available because no changes occurred in the simulation. "
+                   "If you just want a 3D environment file, consider exporting a scene instead"));
```
omichel, line 439: It would be nice to add a final dot at the end of the second sentence.

Jean-Eudes-le-retour: Ok, Daniel did it in commit https://github.com/cyberbotics/webots/pull/5447/commits/10b706568d9523b3bbcebf17a29d75a9c5497ce7

PR #5450, opened 2022-11-09
Documentation update for R2023a

ygoumaz:
Some thumbnails updates.

`Boomer.wbt` has useless PROTO declarations, which is odd as it should have been detected by the `test_worlds` test in the CI.


ad-daniel (approved):
Thank you

PR #5453, opened 2022-11-10
Fix header required for PROTO files

### LINKED ISSUE #5249: PROTO header should be mandatory
**Describe the Bug**

1. Create an empty project
2. Copy an official PROTO (ex: Chair.proto) to the protos folder
3. The PROTO can be inserted using the Add Node dialog.
4. Remove `#VRML_SIM R2022b utf8`, reload and insert it. You can do so but you get a warning that the header is missing
5. Remove all the remaining comments in the header, you can still insert it, but this time you get no warning.

Since a warning for this existed, it doesn't appear to be technically mandatory to insert it (otherwise it would be an error), in practice however I think we should enforce the rule and make it mandatory since it's one of the very few version indicators we have for PROTO files. Moreover, other mechanisms (such as FLU backwards compatibility mechanism, EXTERNPROTO missing warnings) rely on this information being present (and correct).

ygoumaz:
Fixes #5249.


stefaniapedrazzi (commented):
Nice fix!

I would just also add an entry in the ChangeLog.

stefaniapedrazzi (approved):
Thank you

PR #5454, opened 2022-11-10
Add missing init in libcontroller

ad-daniel:
**Description**

Even after adjusting the CMakeList.txt file on the webots_ros2 side (python38 and python310 folders no longer exist), the compilation still fails due to a missing init. The change can only be done on the webots side and wait for it to be automatically pushed to the libcontroller repo tonight


ygoumaz:
If we do it this way, will we able to import controller and vehicle with `from webots_ros2_driver_webots import controller` and `from webots_ros2_driver_webots import vehicle`?

ad-daniel:
Seemed like it, locally the tests were passing at least, but when I launched the tutorial worlds (like the tesla) the car doesn't actually move. Perhaps your approach is better anyway as it's more granular, you don't always need vehicle. If you're handling it already in the rework then I can close these

ygoumaz:
I agree about the granularity. And yes I think we can close this PR and the other on ROS, as I will handle the compilation problems in the rework!

PR #5457, opened 2022-11-10
Fix crash when deleting some node and being connected to webotsJS

### LINKED ISSUE #5351: Crash when streaming and removing the color/coord node from PointSet
**Describe the Bug**
When removing one of the two nodes of a `PointSet` geometry while streaming, Webots crashes.

It seems that the crash happens in `WbX3dStreamingServer::propagateNodeDeletion` when trying to query the `id` of the deleted node.

**Steps to Reproduce**
1. Launch webots with the `--stream` option.
2. Open `geometric_primitives.wbt`.
3. Delete the `color` node in the `PointSet` geometry.
4. Webots crashes

BenjaminDeleze:
Fix  #5351

### DIFF
--- a/src/webots/gui/WbX3dStreamingServer.cpp
+++ b/src/webots/gui/WbX3dStreamingServer.cpp
@@ -188,9 +188,11 @@ void WbX3dStreamingServer::propagateNodeDeletion(WbNode *node) {
     return;
 
   WbTcpServer::propagateNodeDeletion(node);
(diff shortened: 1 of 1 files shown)


stefaniapedrazzi (commented):
Thank you!

It would be better to also an entry in the ChangeLog.

stefaniapedrazzi (approved):
Thank you!

PR #5460, opened 2022-11-10
Fix crash double `root to base node`

### LINKED ISSUE #5141: Double "Convert Root to Base Node(s)" makes Webots crash in Tiago++
**Steps to Reproduce**
1. Open `Tiago++.wbt`
2. `Convert Root to Base Node(s)` one of the `TiagoGripper`
3. `Convert Root to Base Node(s)` one of the `CapScrew` inside of the `TiagoGripper`
4. Webots crash

**Additional context**
Bug present at least since R2022a.

ygoumaz:
Fixes #5141.

The `updateSelection()` method of the `SceneTree` was called on a deleted node, leading to a crash in some cases (e.g. the one reported in the issue). This can be fixed by using the `mRowsAboutToBeRemoved` boolean, which is already used when simply deleting a node from the `SceneTree`.

In addition, this PR fixes another issue, where the newly created `Node` was not selected after the conversion.


ad-daniel (changes requested):
Crash-wise it works as intended now, however I'm still getting the `Warning: QObject::disconnect: Unexpected nullptr parameter` warning mentioned by Olivier in the issue when converting the gripper's root to basenode

ad-daniel (approved):

ygoumaz:
I will open a new issue for the warning, as it doesn't seem to be related to the crash. It is also less important to fix for the release.

PR #5461, opened 2022-11-11
Fix suggested project path

ad-daniel:
**Description**

1. open spot.wbt (or another project in a protected location)
2. `file > new > new project directory`
3. The path contains two `//`

doing the same in a non-protected location doesn't require the correction (line 58).

### DIFF
--- a/src/webots/gui/WbNewProjectWizard.cpp
+++ b/src/webots/gui/WbNewProjectWizard.cpp
@@ -48,7 +48,7 @@ QString WbNewProjectWizard::proposeNewProjectPath() const {
     path = WbPreferences::instance()->value("Directories/projects").toString();
     if (!path.isEmpty()) {
       if (WbFileUtil::isDirectoryWritable(path))
-        path += +"/my_project";
+        path += "my_project";
       else
         path = "";  // no valid default path found
     }


ygoumaz (approved):
Thank you!

src/webots/gui/WbNewProjectWizard.cpp
```diff
@@ -48,7 +48,7 @@ QString WbNewProjectWizard::proposeNewProjectPath() const {
     path = WbPreferences::instance()->value("Directories/projects").toString();
     if (!path.isEmpty()) {
       if (WbFileUtil::isDirectoryWritable(path))
-        path += +"/my_project";
+        path += "my_project";
```
omichel, line 51: Shouldn't we test if a folder separator is needed here?

ad-daniel: doesn't seem necessary as we set one every time: https://github.com/cyberbotics/webots/blob/3b4b6316f1545b3735ff9b04d5d203cd72ad7ccb/src/webots/gui/WbNewProjectWizard.cpp#L84, https://github.com/cyberbotics/webots/blob/3b4b6316f1545b3735ff9b04d5d203cd72ad7ccb/src/webots/editor/WbProjectRelocationDialog.cpp#L224 https://github.com/cyberbotics/webots/blob/3b4b6316f1545b3735ff9b04d5d203cd72ad7ccb/src/webots/core/WbPreferences.cpp#L108

which is why there were two `//`

Issue #2385, opened 2020-10-25
Type annotations for Python API


PeterJCLaw:
I'm not sure how the Python API is built, however I was wondering whether it would be possible to create type annotations for it. This would help document the APIs as well as allow users to check that their code is using the API correctly (just as, for example, the Java compiler is able to do for users of the Java API).

At the moment I've got a set of [local stubs in my project](https://github.com/srobo/competition-simulator/tree/master/stubs) which I've manually created based on the documented API, with the types captured from a mixture of observed data types and the docs for other languages (mostly Java).

It feels like type annotations could usefully be added to the signature snippets in the docs as well as potentially being published as a stubs package on PyPI for use in type checking and as editor completions.

I'm definitely hoping that the signatures would be generated from the true source, however I'd be happy to contribute the stubs I've created as a starting point if they need to be manually defined.

omichel:
We use swig to generate the Python API from the C++ API and there is no easy way to introduce type annotations there, see https://github.com/swig/swig/issues/735 for a discussion on this topic. So, you local stubs approach may be a better way. I guess we shall it to the Python library build process somewhere here: https://github.com/cyberbotics/webots/tree/master/resources/languages/python
Feel free to propose a solution by submitting a new pull request in this repository. We will be happy to review it.

Alternatively, it shouldn't be that difficult to get rid of SWIG for the Python wrapper and maintain it, based on the C libController library (instead of the C++ one). But that's probably a bit more of work, but could be also very nice and fix other problems caused by SWIG.

omichel:
All what you propose should be easy to implement if we totally get rid of swig for generating the Python API.
I believe more and more that a totally new Python API could be created without swig and maintained manually, which should be easy because it will be very straightforward. This API could rely on the C API (instead of the C++ one) thus, removing the dependency on `libCppController` and other C++ dependencies. It would call the C API function directly from python.
Such a library could easily include python hints and documentation with hyperlinks as described above.
In addition to all these benefits, such a library should be designed to be more "pythonic" with python `.value` getters and setters instead of `.getValue()/.setValue()` methods inherited from C++.

If someone would volunteer to go ahead with such an implementation, I would suggest to start with a minimal proof of concept implementation supporting only one type of sensor and one type of actuator. I would be happy to review it in a PR and if it looks good we could generalize it to the complete API. Ultimately, we could drop the current Python API and replace it with this new one.

omichel:
Your type hinting suggestion sounds very good to me.
[...]

PR #5472, opened 2022-11-11
Experiment towards an auto-geneated type stubs package

PeterJCLaw:
**Description**

This creates a [PEP 561](https://peps.python.org/pep-0561/) type stubs package which can be published to PyPI, perhaps for use in type-checking in CI without needing to install the whole of Webots.

A core limitation of this approach is that since this is a stubs-only package consumers are still unable to import the Webots classes, meaning that this doesn't provide a solution for writing unit-tests of controller code which imports (but does not use in tests) Webots controllers. This stubs package does not preclude (nor would it interfere with) adding a more complete runtime package on PyPI later. Doing so would likely render this package redundant however.

This is very much a first pas and not at all ready for merge yet. Instead this is intended as a contribution to the conversation around Python type stubs for Webots.

Currently this package produces type checking failures in the test project I have (https://github.com/srobo/competition-simulator/tree/webots-published-stubs) which are not found by the local stubs from the project (removed on the linked branch). I'm not sure if that's a result of the project-local stubs being incorrect, whether the API in Webots has moved on since the last release, whether there are bugs in the type annotations in Webots, or something else.

I did notice that the Python API in Webots doesn't have full type coverage, nor is it itself checked by `mypy`. While those aren't things I'd want this PR to introduce, it would probably be better to have at least the checking by `mypy` in place before publishing the stubs package so that the stubs don't themselves contain inconsistencies.

**Related Issues**

Follows from discussion https://github.com/cyberbotics/webots/issues/2385#issuecomment-1308487381

**Documentation**

This would likely need documentation.

**Questions**

- Is this a useful thing to have?

If so:
- Is this the right place for it in the code?
- Where/how should the code to publish it go?


omichel:
My feeling is that use cases for this are pretty limited (type checking in CI), and it adds some complexity to our CI actions (which we will have to maintain on the long run). So, I am not sure this is worth the effort.
I believe that if we would add the libController binaries, it would allow for more use cases, e.g., starting an extern controller on one machine that connects to a Webots instance running on a different machine.
But even for this one, I am not aware of anyone needing this.

omichel:
Note: we are currently working on a new repository that holds only the libController and which you could use in your CI: https://github.com/cyberbotics/webots-libcontroller/tree/R2023a
This repo is kept in sync with the Webots repository every night.
Alternatively we are also exploring having branch on the webots repo that contains only the libController files (including Python files, C/C++ headers, Java files, etc.). This could also be a good solution for your use case. It is here for testing purposes: https://github.com/cyberbotics/webots/tree/libcontroller. This one is also synchronized every night on the master and develop branches.

PeterJCLaw:
Yeah, purely type checking is fairly limited. In my actual use-case we also use our "stubs" to allow our controller code to be imported by unit tests (though we don't touch the webots interface during testing), which your libController package approach would also solve for. If you're planning to/interested in publishing that as a standalone typed package (on PyPI?) then that definitely feels like it would make this stubs-only package redundant.

I'll have a play with one or both of those alternatives at some point and let you know if they work for us.

Issue #5477, opened 2022-11-14
Cannot open any world file on macOS with R2023a nightly build


omichel:
When trying to open the first world of the guided tout, I am getting this error in the console:
```
ERROR: 'https://raw.githubusercontent.com/cyberbotics/webots/028694df3179b81396ee615707832aec654e0161/projects/objects/backgrounds/protos/TexturedBackground.proto': error: Template engine error: Couldn't write jsTemplateFilled to disk..
ERROR: 'https://raw.githubusercontent.com/cyberbotics/webots/028694df3179b81396ee615707832aec654e0161/projects/objects/backgrounds/protos/TexturedBackgroundLight.proto': error: Template engine error: Couldn't write jsTemplateFilled to disk..
ERROR: 'https://raw.githubusercontent.com/cyberbotics/webots/028694df3179b81396ee615707832aec654e0161/projects/objects/trees/protos/Sassafras.proto': error: Template engine error: Couldn't write jsTemplateFilled to disk..
ERROR: 'https://raw.githubusercontent.com/cyberbotics/webots/028694df3179b81396ee615707832aec654e0161/projects/objects/trees/protos/Sassafras.proto': error: Template engine error: Couldn't write jsTemplateFilled to disk..
ERROR: 'https://raw.githubusercontent.com/cyberbotics/webots/028694df3179b81396ee615707832aec654e0161/projects/vehicles/protos/abstract/VehicleWheel.proto': error: Template engine error: Couldn't write jsTemplateFilled to disk..
ERROR: 'https://raw.githubusercontent.com/cyberbotics/webots/028694df3179b81396ee615707832aec654e0161/projects/robots/clearpath/moose/protos/MooseWheel.proto': error: PROTO has unknown base type.
ERROR: 'https://raw.githubusercontent.com/cyberbotics/webots/028694df3179b81396ee615707832aec654e0161/projects/vehicles/protos/abstract/VehicleWheel.proto': error: Template engine error: Couldn't write jsTemplateFilled to disk..
ERROR: 'https://raw.githubusercontent.com/cyberbotics/webots/028694df3179b81396ee615707832aec654e0161/projects/robots/clearpath/moose/protos/MooseWheel.proto': error: PROTO has unknown base type.
[...]

PR #5482, opened 2022-11-14
Add a better error message to help debugging the problem

omichel:
This helped me when I bumped into #5477 and will certainly help users experiencing the same problem.


BenjaminDeleze (approved):
Thank you

PR #5484, opened 2022-11-14
Fixed tmp path on macOS

omichel:
Fixes #5477.

This PR makes that Webots uses the local user tmp folder instead of the global one, thus avoiding problems like the one raised in #5477.

PR #5486, opened 2022-11-14
Fix tmp path on mac os

omichel:
Following #5484, we also need to fix the tmp directory for extern controllers.


ygoumaz (approved):
Thank you!

PR #5487, opened 2022-11-14
remove path normalization + replace optparse

### LINKED ISSUE #5481: Sumo crashing directly after startup
`Sumo-gui` crashes directly after the launch.

1. Open `vehicles/highway.wbt`
2. Run the simulation.
3. Sumo starts but crashes directly.

ygoumaz:
Fixes #5481.

This PR also replaces `optparse` (deprecated in Python 3.10) by `argparse`.


stefaniapedrazzi (changes requested):
Thank you!
It works correctly now, but we should not completely remove the normalization of the path but skip it only if `options.directory`is empty.

PR #5488, opened 2022-11-14
Fix motor sound on macOS

omichel:
The text-to-speech sample demo was generating a motor sound whereas no motor was running.
This PR fix this problem which was noticed on macOS, but could be also on other platforms.


stefaniapedrazzi (changes requested):
This fix is fine.
But I would also add an entry in the ChangeLog.

stefaniapedrazzi (approved):
Thank you!

PR #5494, opened 2022-11-15
Update doc specifying that we should not free some returned string.

ShuffleWire:
for function wb_device_get_name/type

just to be consistent, like the many returned array returned by the api (like vector speed in gps, for instance)


omichel (approved):
Thank you.

PR #5504, opened 2022-11-15
Fix gtao shader in webotsJS

BenjaminDeleze:
**Description**
With some specific combination of graphic card, drivers, ubuntu and google chrome, it was possible that WebotsJS was crashing during the compilation of the shaders.

This crash was due to the `gtao.frag` shader and more specifically to this line:
```
  vec2 noises = texelFetch(inputTextures[2], loc % 4, 0).rg;
```

The issue is coming from the fact that the first parameter of `textelFetch` is coming from an array.

I am not sure why it is a problem. Could be a size issue, but we can access the elements of `inputTextures` in other functions, so it would be strange.

I found another post that had the same issue and reached the same conclusion as me: https://community.amd.com/t5/archives-discussions/bug-report-texelfetch-shader-crash-on-msaa-fbo/td-p/87124

To fix the crash, I took the texture out of the array and added it as another `uniform`

This PR needs to be tested both for webots and WebotsJS.

For WebotsJS, you can use the `https://cyberbotics.com/wwi/testing/WebotsView.js` (to replace in animation/scene/index .html)


src/wren/PostProcessingEffect.cpp
```diff
@@ -118,8 +118,14 @@ namespace wren {
           mInputTextures[i]->setTextureUnit(i);
           mInputTextures[i]->bind(mInputTextureParams[i]);
 
-          const int locationTexture =
+          int locationTexture =
             mProgram->uniformLocation(static_cast<WrGlslLayoutUniform>(WR_GLSL_LAYOUT_UNIFORM_TEXTURE0 + i));
+
+          // Special gtao case to bypass a chrome driver bug where texelFetch does not work with textures comming from array:
```
ygoumaz, line 124:
```suggestion
          // Special gtao case to bypass a chrome driver bug where texelFetch does not work with textures coming from array:
```

ygoumaz (changes requested):
Works nicely now. Just a minor comment.

ygoumaz (approved):
Thank you.

PR #5506, opened 2022-11-15
Rewrite gps position test

ShuffleWire:
**Description**
I was reading trough so of the gps test, and some "clang-format on/off" stuff made me wish to update it a bit.
Here is my take on it, a bit in the spirit of https://github.com/cyberbotics/webots/pull/5372
Maybe not all of it is desirable, lets me know.

Note that clang-format issue were removed by using a "XYZ" string.
But could it be better to use the "vec3" version of those testing function, I don't think that getting different message for each of the axis is very useful...
IDEM in tests/api/controllers/gps_coordinates_north_direction/gps_coordinates_north_direction.c

I want to pinpoint as well the fact that I'm removing the `#define TIME_STEP 32` and replaced with `wb_robot_get_basic_time_step()` because it ensure compatibility with the world (and often the world timestep is 60 != 32, not even a multiple...)

And as usual some code factorization.

I would be willing to do more rewrite like this in the future on others tests, but is it a good idea for you ?


tests/api/controllers/gps/gps.c
```diff
@@ -4,59 +4,32 @@
 #include "../../../lib/ts_assertion.h"
 #include "../../../lib/ts_utils.h"
 
-#define TIME_STEP 32
+void check_position(WbDeviceTag gps, double expected_position[], char *end_msg) {
+  const double *position = wb_gps_get_values(gps);
+  char axisNames[] = "XYZ";
+  for (int i = 0; i < 3; i++)
+    ts_assert_double_in_delta(position[i], expected_position[i], 0.000001,
+                              "The %c value measured by the GPS should be %g and not %f %s.", axisNames[i],
```
ShuffleWire, line 12: I've noticed usage of %g, why is that ? I've never used it...

omichel: It's for double with a more human readable format, see https://en.wikipedia.org/wiki/Printf_format_string for details.

omichel: So we should use the same at the two places, either %g or %f.

tests/api/controllers/gps/gps.c
```diff
@@ -4,59 +4,32 @@
 #include "../../../lib/ts_assertion.h"
 #include "../../../lib/ts_utils.h"
 
-#define TIME_STEP 32
+void check_position(WbDeviceTag gps, double expected_position[], char *end_msg) {
+  const double *position = wb_gps_get_values(gps);
+  char axisNames[] = "XYZ";
```
omichel, line 9:
```suggestion
  const char *axisNames = "XYZ";
```

tests/api/controllers/gps/gps.c
```diff
@@ -4,59 +4,32 @@
 #include "../../../lib/ts_assertion.h"
 #include "../../../lib/ts_utils.h"
 
-#define TIME_STEP 32
+void check_position(WbDeviceTag gps, double expected_position[], char *end_msg) {
```
omichel, line 7:
```suggestion
void check_position(WbDeviceTag gps, double *expected_position, char *end_msg) {
```

tests/api/controllers/gps/gps.c
```diff
@@ -4,59 +4,32 @@
...
-                                "The %c value measured by the GPS should be %g and not %g after %d wb_robot_step(s)",
-                                i == 0 ? 'X' :
-                                i == 1 ? 'Y' :
-                                         'Z',
-                                e[i], r[i], j);
-    // clang-format on
+  for (int j = 1; j <= 10; j++) {
+    wb_robot_step(time_step);
+    char s[64];
+    snprint(s, sizeof(s), "after %d wb_robot_step(s)", i);
```
omichel, line 31: I am surprised this compiled:
```suggestion
    snprintf(s, sizeof(s), "after %d wb_robot_step(s)", i);
```

omichel (changes requested):
Yes, this is good to improve the quality of the code even for the tests.

tests/api/controllers/gps/gps.c
ShuffleWire: in tests/api/controllers/gps_speed/gps_speed.c we use %f that for sure

ShuffleWire:
I don't know how to run the tests on my pc, any hints ?

tests/api/controllers/gps/gps.c
ShuffleWire: I've taken the option of %f

tests/api/controllers/gps/gps.c
```diff
@@ -4,59 +4,33 @@
...
-    ts_assert_double_equal(r[i], NAN, "The %c value measured by the GPS should be NaN and not %g before the device is enabled",
-                           i == 0 ? 'X' :
-                           i == 1 ? 'Y' :
-                                    'Z',
-                           r[i]);
-  // clang-format on
-
-  wb_gps_enable(gps, TIME_STEP);
+  const int time_step = wb_robot_get_basic_time_step();
+  WbDeviceTag gps = wb_robot_get_device("gps");
```
omichel, line 20:
```suggestion
  const WbDeviceTag gps = wb_robot_get_device("gps");
```

omichel (changes requested):
Except for a couple of constness issues, everything looks good to me.

ShuffleWire:
I have a question, why does the "Changes requested" stay, even if I've fixed what you have requested to be changed ?
image

Anyway, it's done !

ShuffleWire:
i see, thank !

PR #5508, opened 2022-11-15
Add `isInitializationPossible` to Python API

ygoumaz:
`isInitializationPossible` allows to know if a robot can use the driver API function or not. It is mainly used by `webots_ros2` driver node.
 
The function was missing in the new Python API. Also it is renamed and refactored in the C and C++ codes, as it isn't part of the public documented function.


include/controller/c/webots/vehicle/driver.h
```diff
@@ -35,9 +35,11 @@ typedef enum { UNDEFINED_CONTROL_MODE = -1, SPEED = 0, TORQUE } WbuDriverControl
 
 typedef enum { DOWN, SLOW, NORMAL, FAST } WbuDriverWiperMode;
 
+// private function for webots_ros2 to identify robots that can use libdriver
+bool initialization_is_possible();
```
omichel, line 39: I would keep the prefix to avoid name clashes:
```suggestion
bool wbu_driver_initialization_is_possible();
```

omichel (changes requested):
I believe it is better to keep the same name for the C function to avoid the name clash and be consistent with other API functions (C++, Python).

omichel (approved):
Thank you.

PR #5511, opened 2022-11-15
Improve the position of the viewpoint in the create.wbt

BenjaminDeleze:
Before, the viewpoint position was not very good with the viewpoint follow


ad-daniel (approved):
it's better indeed, thank you

PR #5512, opened 2022-11-15
Removed obsolete Java controller warning for non-ASCII path

### LINKED ISSUE #5497: Java controller in path with non 8-bit characters on Windows
I tested compiling and running Java controllers from a path containing non 8-bit characters (user name contains non ASCII characters).
Although I get many warnings like:
```
The path to this Webots project contains non 8-bit characters. Webots won't be able to execute any Java controller in this path. Please move this Webots project into a folder with only 8-bit characters.
```

The Java controller could be compiled and works correctly.

Is the warning still up-to-date?

omichel:
Fixes #5497.


stefaniapedrazzi (approved):
Thank you

Issue #5493, opened 2022-11-15
Loading dialog continuously open and closes while loading world


stefaniapedrazzi:
On Windows, it happens that when opening a world the loading dialog continuously disappears and re-appears when the following text is displayed "Open World File".
This blinking effect is a little bit disturbing and should be avoided.

I could reproduce this on Windows (didn't check on Ubuntu).

omichel:
I confirm this problem also shows up on macOS.

ad-daniel:
This might actually be a regression, it doesn't happen with the release of R2022b

ad-daniel:
The problem appears to stem from 2 conflicting progress counters:
- [This](https://github.com/cyberbotics/webots/blob/48444fa1f9d6213ef6efc0f69f63dbbd13db692f/src/webots/app/WbApplication.cpp#L53-L54) check says the progress is 100 and therefore the dialog is closed
- meanwhile, occasionally, [this](https://github.com/cyberbotics/webots/blob/48444fa1f9d6213ef6efc0f69f63dbbd13db692f/src/webots/app/WbApplication.cpp#L142-L147) setter re-creates the dialog because according to the parameter it receives, the progress is not 100

basically the two fight with each other

omichel:
I believe we can close this issue now.

PR #5513, opened 2022-11-15
Fix dialog popup opening/closing

ad-daniel:
**Description**

Fixes https://github.com/cyberbotics/webots/issues/5493


ad-daniel:
I believe I found the source of the problem, basically the conflict arises when doing the [finalize](https://github.com/cyberbotics/webots/blob/48444fa1f9d6213ef6efc0f69f63dbbd13db692f/src/webots/engine/WbSimulationWorld.cpp#L105) step during the loading.

At this step the pop up is supposed to display the progress with respect to the finalization process (create WREN objects, create ODE objects, etc), meanwhile as the ImageTextures/Mesh nodes are finalized, their corresponding Downloader is destroyed, which triggered the destruction of the popup progress (hundreds of times, one for each instance), which doesn't seem necessary since at this point the download is complete anyway (hence why this one kept saying progress was 100).

Removing this progress update in the Downloader destroyer removes the flickering and doesn't appear to have side effects.

ad-daniel:
Note: the flickering is visible only in distrib mode, with a sufficiently complex world (like panda) and with cache cleared

ad-daniel:
Actually there might be a side effect in case of a download failure (line above), then perhaps the update should take place in that specific case. I'll test it quickly, will put the PR back to draft for a moment

omichel (dismissed):
Thank you.
I confirm the problem is fixed on macOS.

ad-daniel:
This approach does indeed seem more appropriate as if the count changes due to unfinished downloads, the progress bar should reflect the change.

omichel (approved):
That seems better indeed.

stefaniapedrazzi (approved):
Thank you!

PR #5516, opened 2022-11-17
Add missing methods in Python API

ygoumaz:
One static method, used by the `webots_ros2` package, was missing in the new Python API.


omichel (commented):
It looks good to me, but I am not sure about the need for `getSupervisorInstance()` and `getDriverInstance()` static methods. Do we really need them? Isn't it clearer and easier to instantiate these objects directly instead of calling these methods?

ygoumaz:
I don't know if these methods are intended to be used outside `webots_ros2`. In C++, the function is looking for an existing cInstance. In the case of the ROS2 package, there is no initialization of the Supervisor or Driver objects through Python before calling these methods, so there is no way to retrieve an existing instance. This is why it is simplified to simply re-creating a new object. Maybe I could add the check for an existing instance to avoid trying to recreate it if it already exists. (This would be only useful if this method is used outside webots_ros2)

omichel (approved):
Very nice.
Thank you.

Issue #5515, opened 2022-11-16
Motion blur test is broken


ad-daniel:
**Describe the Bug**

As emerged in https://github.com/cyberbotics/webots/pull/5505, the motion_blur test occasionally fails but was until now silently ignored (didn't trigger the test suite to fail in the CI). I'm having a hard time trying to reproduce it locally, but I see it appear now and then by:

1. open `camera_color_motion_blur.wbt`
2. set fast mode (or normal mode, not sure which happens more often)
3. keep reloading over and over, might take 10-20 tries before it shows, but it shows

stefaniapedrazzi:
Might be related to #5491

ad-daniel:
Unfortunately it turns out to be more recent, it seems this GTAO shaders fix https://github.com/cyberbotics/webots/pull/5504 introduces the problem

PR #5517, opened 2022-11-17
Fix motion blur shader

### LINKED ISSUE #5491: Motion blur not applied on pasted robot with camera (Windows)
Steps to reproduce:
1. Open `camera_motion_blur.wbt` and run the simulaton
2. Copy the MyBot robot and paste it in the world
-> in the camera overlay of the newly added Robot the motion blur effect is not visible

I get this behavior on Windows (with user name containing UTF-8 characters).
I tested on my development system on Ubuntu 22.04 and the motion blur is applied correctly to the pasted Camera image.

stefaniapedrazzi:
Possibly fix #5491 and #5515.
"firstRender" uniform is created with the wrong type (introduced in #5436).
This might cause the not fully reproducible issues experienced on various systems.


ad-daniel:
I can confirm I'm unable to make the test fail in this branch, nice catch!

PR #5518, opened 2022-11-18
Backport gtao fixes

BenjaminDeleze:
**Description**
Backport the gtao fixes such that animations/simulations using R2022b would also benefit from the fix.

PS: Not sure if it is a good idea to backport as we should release soon, but I am afraid that if we don't (and I just modify the online resources), we could forget this patch in the case were we need to rebuilt the web resources.


omichel (dismissed):
That makes sense.
Thank you.

BenjaminDeleze:
I just correct a style issue

ad-daniel (approved):

PR #5521, opened 2022-11-18
Clean-up python API

omichel:
This PR:

- Simplifies the syntax for the local module import (e.g., remove the useless `controller` prefix in import statements).
- In case the Webots preferences have an empty string for the `Python command`, use the default system python command. The problem is that if you install and run Webots on Windows and there is no Python command in the PATH, Webots will set the `Python command` to an empty string in its preferences. If you install Python afterwards, the `Python command` string being empty and Webots will fail to use Python, unless you explicitly change the preferences.

### DIFF
--- a/src/webots/control/WbLanguageTools.cpp
+++ b/src/webots/control/WbLanguageTools.cpp
@@ -130,7 +130,7 @@ QString WbLanguageTools::pythonCommand(QString &shortVersion, const QString &com
   if (pythonCommand == "!")
     WbLog::warning(QObject::tr("Python was not found.\n") + advice);
 #else  // __linux__
-    shortVersion = checkIfPythonCommandExist(pythonCommand, env, true);
+  shortVersion = checkIfPythonCommandExist(pythonCommand, env, true);
   if (shortVersion.isEmpty()) {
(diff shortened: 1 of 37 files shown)


src/webots/control/WbLanguageTools.cpp
```diff
@@ -124,7 +130,7 @@ QString WbLanguageTools::pythonCommand(QString &shortVersion, const QString &com
   if (pythonCommand == "!")
     WbLog::warning(QObject::tr("Python was not found.\n") + advice);
 #else  // __linux__
-    shortVersion = checkIfPythonCommandExist(pythonCommand, env, true);
+      shortVersion = checkIfPythonCommandExist(pythonCommand, env, true);
```
BenjaminDeleze, line 133: This indentation looks strange to me, but atom auto-correct like that so I guess it is right

BenjaminDeleze (approved):
Thank you

src/webots/control/WbLanguageTools.cpp
omichel: Yes, I had to do this to pass the clang-format test... But I agree, this looks wrong...

PR #5530, opened 2022-11-21
Merge develop into master

ygoumaz:
Merge develop into master for the R2023a release.


ygoumaz:
Should we add the test suite and other tests to make sure everything works fine?

omichel (approved):
I don't think it is needed to add the test suite label here, as basically, the result on the master branch should what we currently have on the develop branch.

PR #5537, opened 2022-11-21
Add automatic branch synchronization

ygoumaz:
This PR adds a workflow to automatically create PRs to synchronize the following branches overnight (if needed):
- `master` into `develop`.
- `released` into `master`.


.github/workflows/sync_develop_master.yml
```diff
@@ -0,0 +1,75 @@
...
+        if: ${{ steps.check-diff.outputs.create_pull_request == 1 }}
+        uses: repo-sync/pull-request@v2.6.2
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          source_branch: released
+          destination_branch: master
+          pr_title: Merge released into master
+          pr_body: Synchronizes changes of released branch into master branch.
+          pr_reviewer: Maintainers
+          pr_label: "synchronization"
```
ad-daniel, line 39: Since the workflows don't appear to run immediately, I think we can avoid setting the label (that way we don't need to remove it, and re-set it)

.github/workflows/sync_develop_master.yml
```diff
@@ -0,0 +1,75 @@
...
+        if: ${{ steps.check-diff.outputs.create_pull_request == 1 }} || ${{ needs.sync-master-with-released.outputs.pr_opened == 1 }}
+        uses: repo-sync/pull-request@v2.6.2
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          source_branch: master
+          destination_branch: develop
+          pr_title: Merge master into develop
+          pr_body: Synchronizes changes of master branch into develop branch.
+          pr_reviewer: Maintainers
+          pr_label: "synchronization"
```
ad-daniel, line 74: ditto

.github/workflows/sync_develop_master.yml
```diff
@@ -0,0 +1,75 @@
...
+          pr_allow_empty: true
```
ad-daniel, line 75: I'm not sure about this option. Usually we only push bug fixes to master, meaning "normally" we are much more active on develop rather than master. Does this option entail that every day we'll have an empty PR popping up telling us there's nothing sync?

ad-daniel (changes requested):
Thank you, just some doubts

.github/workflows/sync_develop_master.yml
ad-daniel: perhaps I misunderstood, the step is run only if there's a diff. In that case then it would never be empty, no?

ad-daniel (approved):
Thank you

PR #5551, opened 2022-11-23
Fix accessibility to view in webotsview

BenjaminDeleze:
The `view` variable should not be private as it is used in `view.js`


omichel (approved):
Thank you.

PR #5555, opened 2022-11-23
Remove downscale of texture quality

BenjaminDeleze:
When the `Texture Quality` parameter is set to low, some images are scaled down.

However they are saved outside of the `project` folder which make them unavailable for the streaming, which was causing issue on `webots.cloud`

After some discussion, it was decided to stream the untouched images, mainly because as most of the textures are hosted on github it does not make a lot of sense to scale down only some of them.


omichel (approved):
Thank you.

PR #5556, opened 2022-11-23
Fix Python import in documentation

stefaniapedrazzi:
Some Python import in documentation should be updated because are triggering errors with the new python API.


ad-daniel (approved):
Thank you

PS: does having an empty class still serve a purpose [here](https://github.com/cyberbotics/webots/blob/1a48493d8d0d1fc1faf237fa53856e5699a84185/lib/controller/python/controller/node.py#L35-L36)?

omichel:
> PS: does having an empty class still serve a purpose [here](https://github.com/cyberbotics/webots/blob/1a48493d8d0d1fc1faf237fa53856e5699a84185/lib/controller/python/controller/node.py#L35-L36)? <-- looks like it is

Yes, if we remove it, it won't work any more. The class somehow needs to be declared as it is used as a return value in the class declared hereafter. Maybe there is a better way to do this, but that's I found so far.

omichel:
Instead of changing the documentation in a somewhat strange way, shouldn't we fix this [__init__.py](https://github.com/cyberbotics/webots/blob/master/lib/controller/python/controller/__init__.py) file instead by adding the missing objects?

stefaniapedrazzi:
> Instead of changing the documentation in a somewhat strange way, shouldn't we fix this [__init__.py](https://github.com/cyberbotics/webots/blob/master/lib/controller/python/controller/__init__.py) file instead by adding the missing objects?

Currently this is the correct import reflecting the implementation of the classes and seems quite simple, even because I'm not sure that there will be many cases in which the inner classes will be needed without having to import the device class itself.

But for me it is fine to change it.
However, we should also change the way these classes are defined and move them outside the parent device class.
For example currently `CameraRecognitionObject` is defined inside the `Camera` class.

stefaniapedrazzi:
Just the declaration of the `CameraRecognitionObject`class needs to be moved.

omichel (approved):
Thank you.
It looks better to me this way.

ad-daniel (approved):

PR #5558, opened 2022-11-23
Updated extern controller setup for Python

omichel:
Due to the change in the Python API, the setup for extern controllers is now simpler as the Python controllers don't need to load the libCppController any more.

This PR also pre-loads the `generic_robot_window.dll` in the Python API on Windows, so that adding `WEBOTS_HOME/lib/controller` to the `Path` is not needed any more.

The e-puck robot is an exception as its remote control library has dependencies on the MSYS2 C++ libraries, so some extra setup is needed in this case which has been documented in this PR.

I successfully tested the following on Windows in a modified installed distribution:
- standard Python controller on a generic robot.
- extern Python controller on a generic robot.
- extern Python controller on an e-puck robot.
- standard Python controller on an e-puck robot.

Everything works nicely as expected.


ygoumaz (changes requested):
I tested on Windows and it works well! Just a minor comment.

ygoumaz (approved):
Thank you!

PR #5560, opened 2022-11-23
Fix automatic protected branches sync

ygoumaz:
Add an intermediary branch for the branches sync


ad-daniel (approved):
Looks good, thank you

PR #5572, opened 2022-11-24
Add default argument for some methods in Python API

### LINKED ISSUE #5569: Wrong default arguments in new Python API
Some default argument specified in the documentation are not specified in the new Python API implementation.
For example this instruction ``object.enablePoseTracking(timestep)`` is throwing the following error:
```
TypeError: Node.enablePoseTracking() missing 1 required positional argument: 'fromNode'
```

These are some functions where the default argument is documented but not implemented:
* Supervisor.enablePoseTracking
* Supervisor.disablePoseTracking
* Supervisor.getPose
* Supervisor.worldSave

Not sure if these were skipped because it is difficult to pass a NULL pointer to the C functions.
But the documentation should match the implementation and if possible the default argument should be kept.

ygoumaz:
Fixes #5569.

### DIFF
--- a/lib/controller/python/controller/node.py
+++ b/lib/controller/python/controller/node.py
@@ -130,15 +130,18 @@ class Node:
         o = wb.wb_supervisor_node_get_orientation(self._ref)
         return [o[0], o[1], o[2], o[3], o[4], o[5], o[6], o[7], o[8]]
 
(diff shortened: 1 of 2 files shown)


omichel (approved):
Thank you.

PR #5591, opened 2022-11-28
Fix Python Node.disablePoseTracking

stefaniapedrazzi:
The wrong C function is called in `Node.disablePoseTracking`.


stefaniapedrazzi:
I also noticed that it is not consistent that the `wb_supervisor_node_enable_pose_tracking` support a NULL `from_node` argument, but the `wb_supervisor_node_disable_pose_tracking` doesn't support it.
I will open an issue for this.

ygoumaz (approved):
Thank you!

Issue #5567, opened 2022-11-24
Add Webots version in web interface


stefaniapedrazzi:
I would be nice if the Webots tab in the Simulation information window of the web interface also mentions which Webots version is used (similarly to the About window of Webots).

https://github.com/cyberbotics/webots/blob/667eba1ee39b9e59b0bc0c1226597697a3014a25/resources/web/wwi/InformationPanel.js#L24-L28

PR #5601, opened 2022-11-29
Add version information in webotsJS

BenjaminDeleze:
**Description**
Add the version number in webotsJS information panel as requested in #5567 

you can test it here https://testing.webots.cloud


BenjaminDeleze:
The loading screen is not correct, because on `testing` webots.cloud is wired only to the testing folder of the server that is on R2023b. So all loading screen will display R2023b (not related with this PR)

The info panel get the version from the `.x3d` file so it get the correct one even on `testing.webots.cloud` which is the version who correspond also to the table.

Also when this PR will be merged, it will work only for `R2023b` files as the older one will load a version of webotsJS that does contain those modifications.

stefaniapedrazzi (approved):
Thank you!

PR #5605, opened 2022-11-30
Add libtess

BenjaminDeleze:
The libtess dependency that is needed for the tesselation was added on the server but not in the branch


ad-daniel (approved):
Thank you

PR #5606, opened 2022-11-30
Add recognitionColors default value for some PROTO models

stefaniapedrazzi:
Add default `recognitionColors` in new animals, Barn PROTO models and a new `recognitionColors` parameter in SolidBox, Tractor and Forest.

### DIFF
--- a/projects/objects/animals/protos/Cat.proto
+++ b/projects/objects/animals/protos/Cat.proto
@@ -229,5 +229,8 @@ PROTO Cat [
         }
       ]
     }
+    recognitionColors [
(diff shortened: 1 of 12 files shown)


ad-daniel (changes requested):
Thank you. We should also regenerated the object docs (although it might not matter much since it won't be visible until the revision is done, if ever)

stefaniapedrazzi:
Done. Only the documentation `SolidBox` changed.

ad-daniel (approved):
Thank you

PR #5610, opened 2022-11-30
Use better data types in the Python API

omichel:
The new Python API converts the data returned by the libController API to Python data types such as lists of standard data types. This operation is sometimes costly (in case of images for example).
This PR aims at performing efficient cast operations rather than costly conversions.


stefaniapedrazzi (approved):
Thank you!

PR #5612, opened 2022-12-01
Enhancement display camera

BenjaminDeleze:
# Use case 4 : Camera
A professor uses Webots in its robotics course. She is preparing a lab about vision and wants to choose the robot that suits her need the best camera-vise.
Then, she wished to modify the proto such that only translation and rotation can be modified.
She has two choices
1) Open Webots and load every robots, then go in the sceneTree or in the proto file and find detail about the camera. Once she finds the right robot, she must find the corresponding proto file and modify it.
2) Go to webots.cloud and open the robots there. I would say that it is simpler to see the robots, cameras frustum and to play with the parameters there but it is debatable. However it is way more intuitive to create a new proto with less parameters directly on webots.cloud than to modify directly the proto file.

What is implemented in this PR:
- Display the list of camera in the devices tab.
- When hovering a camera, it displays its frustum in the 3D view.
- Add support for near, far, height, width and fieldOfView updates for the cameras.
- Add y scrolling when the list of parameters/devices/joints is too long
- Fix the step of the integer fields from 0.1 to 1.

PS: The frustum color is grey/white as I took the color as in Webots but I am not sure it is the best for visibility


resources/web/wwi/nodes/WbAbstractCamera.js
```diff
@@ -0,0 +1,171 @@
...
+      _wr_renderable_set_visibility_flags(this._renderable, WbWrenRenderingContext.VM_REGULAR);
+    else
+      _wr_renderable_set_visibility_flags(this._renderable, WbWrenRenderingContext.VM_REGULAR);
+
+    _wr_node_set_visible(this._transform, true);
+  }
+
+  minRange() {
+  }
+  maxRange() {
```
ad-daniel, line 146:
```suggestion
  
  maxRange() {
```

resources/web/wwi/nodes/WbCamera.js
```diff
@@ -1,5 +1,36 @@
...
+  }
+
+  set far(newFar) {
+    this.#far = newFar;
+  }
+
+  minRange() {
+    return this.#near;
+  }
+  maxRange() {
```
ad-daniel, line 33:
```suggestion
  
  maxRange() {
```

ad-daniel (changes requested):
Insofar as the PROTO viewer is concerned, I think it might be fine to enable the display of the frustums by default whenever we are on the device tab, but potentially I assume we'd like to be able to see them in other situations too like scenes, animations whenever the user selects the robot, in that case however it might be better to have an option in the settings to enable/disable it. Can this rendering be ported to those situations too later on or it would be necessary to do it a different way?

Also I agree that the color of the frustum is a bit hard to see, but other than that it looks fine to me

ad-daniel:
Perhaps @omichel should also take a look at it

omichel:
> Perhaps @omichel should also take a look at it

Sure, can you post a link so that I can quickly test it online?

BenjaminDeleze:
> Insofar as the PROTO viewer is concerned, I think it might be fine to enable the display of the frustums by default whenever we are on the device tab

I think it is better to show it only when hovering the device, otherwise we will not be able to see which device is where.

> Also I agree that the color of the frustum is a bit hard to see, but other than that it looks fine to me

I will try with the pink that we have in Webots when the camera is enable

> Sure, can you post a link so that I can quickly test it online?

I just modify the color of the frustum and post the link after

BenjaminDeleze:
https://proto.webots.cloud/run?version=proto&url=https://github.com/BenjaminDeleze/sandbox/blob/master/protoExample/protos/ProtoCamera.proto

Here is the link to test

omichel:
It looks good to me.
I just noticed that when I change the near, this is taken into account in the frustum representation. However, if I change the field of view, it doesn't seem to be taken into account.
Also, when a camera is selected, it is nice that it remains selected even if we navigate in the 3D scene. But it would be also more convenient if it could remain visible also when we modify the near or fieldOfView parameters. It is unselected as soon as I click the parameters tab...

BenjaminDeleze:
> I just noticed that when I change the near, this is taken into account in the frustum representation. However, if I change the field of view, it doesn't seem to be taken into account.

It should be good now

>Also, when a camera is selected, it is nice that it remains selected even if we navigate in the 3D scene. But it would be also more convenient if it could remain visible also when we modify the near or fieldOfView parameters. It is unselected as soon as I click the parameters tab...

When we discussed yesterday I understood that we did not want that behavior...
Further than the user experience, it raises two issues (but nothing too hard):
- Once you select something, there is no way to unselect it, so we should find one.
- When changing parameters:
  -   if the proto does not need to regenerated, I need to make some adjustment to rebuild the frustum with the new parameter but it should be easy.
  -  if the proto regenerates, we will lose the frustum as the scene and the tabs are rebuilt

omichel:
I confirm the change of `fieldOfView` now works as expected.
We could simply consider that the frustum will remain visible until we click on the device name to un-select it. What do you think? Any other idea?

BenjaminDeleze:
> I confirm the change of `fieldOfView` now works as expected. We could simply consider that the frustum will remain visible until we click on the device name to un-select it. What do you think? Any other idea?

So the behavior would be:

- hover to select
- hover another device to change the showed device
- click on the selected device to unselect 
- Regeneration unselect it as well

ad-daniel:
> We could simply consider that the frustum will remain visible until we click on the device name to un-select it. What do you think? Any other idea?

That wouldn't be a bad idea, especially if the user wants to see more than one device at once, he activates and de-activates them at will

BenjaminDeleze:
> That wouldn't be a bad idea, especially if the user wants to see more than one device at once, he activates and de-activates them at will

Ok so click to enable and click to disable
and hover? enable it but not in a persistent way?

ad-daniel:
> enable it but not in a persistent way?

yes exactly, what do you think?

BenjaminDeleze:
> > enable it but not in a persistent way?
> 
> yes exactly, what do you think?
seems good to me

omichel:
I would add a radio button so that the user understands that she has to click to enable/disable something. Then, we can probably get rid of the hover behavior.

ad-daniel:
> Then, we can probably get rid of the hover behavior.

I think the hover has a purpose, especially on humanoid robots with many devices (with device names that might not be evident), the hover can allow you to search for the ones you're looking for quickly without selecting-deselecting them one at a time until you do find it. But yea, we might not need it immediately

omichel:
That's a good point. So, let's keep the hover. I was just concerned that if a device is already selected (with the radio button) and we hover another one, the two will be selected at the same time. But maybe this is not a problem? In that case the hover should select the device only when the mouse is hovering the device name and it should be un-selected when un-hovering the device name.

omichel:
And if selecting multiple devices is not a problem, we should probably replace the radio buttons (exclusive) with check boxes (to select several devices).

BenjaminDeleze:
> That's a good point. So, let's keep the hover. I was just concerned that if a device is already selected (with the radio button) and we hover another one, the two will be selected at the same time. But maybe this is not a problem? In that case the hover should select the device only when the mouse is hovering the device name and it should be un-selected when un-hovering the device name.

Yes exactly, I think we should allow multiple devices to be selected with `click` and `hover` will just enable the optional rendering of a device while the mouse is houvering this specific device

BenjaminDeleze:
I finished a new version that you can test here: https://proto.webots.cloud/run?version=proto&url=https://github.com/BenjaminDeleze/sandbox/blob/master/protoExample/protos/Nao.proto or with the previous link

I did not put the checkboxes yet as I was not sure they were necessary. 
Can you try it and tell me what do you think @omichel @ad-daniel please?

ad-daniel:
Behavior-wise I think it's great, personally I like it even without the checkboxes, just highlighting seems sufficient. The fact that it updates even when on the parameters tab (with rendering on) is very nice

BenjaminDeleze:
Thanks, the near is now fixed

ad-daniel (approved):
It looks quite nice, thank you

I've noticed with the spot however that the floating window is populated before the spot appears, would it be possible to delay it until the 3d view is ready? (and perhaps show a "loading.." text in the tab meanwhile?
The risks are likely present for other tabs too, I'm just afraid if people start pressing stuff, causing regenerations and so on before everything is ready might lead to problems, but we can address it in a different PR

BenjaminDeleze:
> I've noticed with the spot however that the floating window is populated before the spot appears, would it be possible to delay it until the 3d view is ready? (and perhaps show a "loading.." text in the tab meanwhile?
> The risks are likely present for other tabs too, I'm just afraid if people start pressing stuff, causing regenerations and so on before everything is ready might lead to problems, but we can address it in a different PR

Indeed we should address that but in another PR. My guess is that it happens because we are still doing a double loading (first the empty world and then the proto) so the issue should been gone once we fix that

PR #5613, opened 2022-12-01
Fix two bugs in feature-web-proto

BenjaminDeleze:
- Error when a bounding object was in a slot.
- Issue of position if the joint anchor was not 0


ad-daniel (approved):
nice catch, thank you

PR #5616, opened 2022-12-02
Removed useless Python installations on macOS

omichel:
It seems that Python 3.7 is now deprecated causing the nightly build to fail on macOS.
This PR cleans-up the installation of Python. Since the new Python API was introduced, is not needed any more to install all these versions.

### DIFF
--- a/.github/workflows/test_suite_mac.yml
+++ b/.github/workflows/test_suite_mac.yml
@@ -43,24 +43,7 @@ jobs:
     - name: Install Webots Compilation Dependencies
       run: |
         # swig wget cmake python@3.8 python@3.9 and python@3.10 are already installed
-        if HOMEBREW_NO_INSTALL_CLEANUP=1 brew install python@3.7; then
-          echo "Installation of Python3.7 successful"
-        else
-          echo "Installation of Python3.7 failed"
-        fi
         npm install -g appdmg
-        # install regular Python 3.7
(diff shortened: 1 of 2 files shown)


.github/workflows/test_suite_mac.yml
```diff
@@ -107,10 +90,10 @@ jobs:
         hdiutil mount artifact/webots-*.dmg
         sudo cp -R /Volumes/Webots/Webots.app /Applications
         hdiutil unmount /Volumes/Webots
-    - name: Set up Python 3.9
+    - name: Set up Python 3.10
       uses: actions/setup-python@v2
       with:
-        python-version: 3.9
+        python-version: 3.10
```
ygoumaz, line 96: Is there a reason to use Python 3.10 over 3.11? Is it not available on macOS CI machines yet?

omichel: I have the feeling that 3.11 is still very new and a few modules may be missing (not yet ported to 3.11).

omichel: I just set 3.11. Let's see the tests are passing.

omichel: I seems it worked out, so let's keep Python 3.11.

ygoumaz (approved):
It seems to work with Python 3.11! Thank you!

PR #5619, opened 2022-12-02
Try to get rid of MSYS/git installation on Windows

### LINKED ISSUE #5566: Clean-up Windows CI with respect to Python path
We should fix the Windows CI to get the path to the Python binary.
As seen on this [log](https://github.com/cyberbotics/webots/actions/runs/3530823966/jobs/5923247607), the path is wrongly set on the first place and over-ridden in a second place to be fixed.
We should set it right on the first place.

Also, the CI script install git in MSYS2, which is apparently not necessary as demonstrated [here](https://github.com/omichel/action_test/blob/main/.github/workflows/test.yml).

omichel:
Fixes #5566 (git part).

### DIFF
--- a/.github/workflows/test_suite_windows.yml
+++ b/.github/workflows/test_suite_windows.yml
@@ -29,8 +29,6 @@ jobs:
     - uses: msys2/setup-msys2@v2
       with:
         update: false
-        install: >-
-          git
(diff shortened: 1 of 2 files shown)


ad-daniel (approved):
Thank you

PR #5629, opened 2022-12-05
Feature extern controllers launcher

### LINKED ISSUE #5565: Feature extern controller launcher
Currently the documentation and workflow to run extern controllers is very messy and complicated. There are a lot of cases and environment variables to set which depend on the OS and the language of the controller.

The idea would be automatize the process using a launcher to simplify the life of users willing to run extern controllers. It would also allow to simplify a lot the documentation.

They could run commands like the following:
```
extern_controller_launcher path/to/controller/executable --protocol=ipc --robot_name=MyRobot
```

The launcher, based on the OS and the nature of the controller executable, would take care of setting all the environment variables. Options, like the `protocol` or the `robot-name`, would allow to automatically set the WEBOTS_CONTROLLER_URL variable.

ygoumaz:
Resolves #5565.

The new launcher is compiled to `WEBOTS_HOME/webots-controller`.
Available options are:
- `--protocol`: can be either `ipc` or `tcp`. `ipc` is used for local connections (i.e. Webots is started on the same machine as the controller). `tcp` is used for remote connections (e.g. Webots is started on a different machine than the controller). `ipc` is used by default if this option is not given.
- `--ip_address`: The IP address of the remote machine on which the Webots instance is running. Mandatory for `tcp` connections and can be omitted when using `ipc`. 
- `--port`: 1234 is used by default, as it is the default port of Webots. This parameter allows to connect to a specific instance of Webots if multiple of them are running on the target machine. The port of a Webots instance can be set at its launch.
- `--robot_name`: Target a specific robot by specifying its name in case multiple robots wait for an extern controller in the Webots instance.
- `--matlab-path`: For MATLAB controllers, this option allows to specify the path to the executable of a specific MATLAB version. By default, the launcher checks in the default MATLAB installation folder.
- `--stdout-redirect`: Redirect the stdout of the controller to the Webots console.
- `--stderr-redirect`: Redirect the stderr of the controller to the Webots console.

Example:
```bash
./webots-controller $WEBOTS_HOME/projects/samples/devices/controllers/gps/gps --protocol=tcp --ip_address=localhost --port=1234 --robot_name=MyBot
```

Tasks:
- [x] C / C++ binaries environment
- [x] Java environment
- [x] Python envrionment
- [x] Matlab environment
- [x] Runtime.ini parsing
- [x] Documentation: https://cyberbotics.com/doc/guide/running-extern-robot-controllers?version=feature-extern-controllers-launcher
- [x] Tests on Windows
- [x] Tests on macOS


src/controller/launcher/webots_controller.c
```diff
@@ -0,0 +1,843 @@
...
+  // MATLAB controller
+  else if (strcmp(controller_extension, ".m") == 0) {
+    matlab_config_environment();
+
+    // If no MATLAB installation path was given in command line, check in default installation folder
+    if (!matlab_path) {
+      const bool default_matlab_install = get_matlab_path();
+      if (!default_matlab_install)
+        return -1;
+    }
```
omichel, line 745:
```suggestion
    if (!matlab_path && !get_matlab_path())
      return -1;
```

src/controller/launcher/webots_controller.c
```diff
@@ -0,0 +1,843 @@
...
+#endif
+    char *short_controller_path = strdup(controller_path);
+    short_controller_path[strlen(controller_path) - 1] = '\0';
+#ifdef _WIN32
+    // Write the 'classpath' option (mandatory for java controllers)
+    const size_t classpath_size = snprintf(NULL, 0, "\"%s%s%s\"", lib_controller, jar_path, short_controller_path) + 1;
+    char *classpath = malloc(classpath_size);
+    sprintf(classpath, "\"%s%s%s\"", lib_controller, jar_path, short_controller_path);
+
+    // Write the 'Djava.library.path' option (mandatory for java controllers)
```
omichel, line 796:
```suggestion
    // Write the '-Djava.library.path' option (mandatory for java controllers)
```

src/controller/launcher/webots_controller.c
```diff
@@ -0,0 +1,843 @@
...
+    char *new_argv[] = {"java", "-classpath", classpath, java_library, controller_name + 1, NULL};
+    execvp(new_argv[0], new_argv);
+#endif
+
+    free(lib_controller);
+    free(short_controller_path);
+    free(classpath);
+    free(java_library);
+  } else
+    printf("The file extension '%s' is not supported as webots controller. Supported file types are executable files, '.py', "
```
omichel, line 825:
```suggestion
    fprintf(stderr, "The file extension '%s' is not supported as webots controller. Supported file types are executable files, '.py', "
```

omichel (changes requested):
We should add the old documentation of the environment variables for launching extern controllers as a comment inside the source code of the launcher.

I will now test on Windows.

omichel:
I am testing now on Windows and will test on macOS as well.
@ad-daniel: may I ask you to test it on Linux?

src/controller/launcher/Makefile
```diff
@@ -0,0 +1,48 @@
...
+null :=
+space := $(null) $(null)
+WEBOTS_HOME_PATH?=$(subst $(space),\ ,$(strip $(subst \,/,$(WEBOTS_HOME))))
+endif
+
+default:
+	@echo "Type one of the following (OSTYPE is \""$(OSTYPE)"\"):"
+	@echo "  make debug                 for debug (with gdb symbols)"
+	@echo "  make profile               for profiling (with gprof information)"
+	@echo "  make release               for finale release"
```
ad-daniel, line 27:
```suggestion
	@echo "  make release               for final release"
```

omichel:
When starting an extern controller, the launcher should also "cd" into the controller directory. Otherwise, extern controllers don't have the same environment as regular controller and fail because they don't find their local files, for example:
```
$ webots-controller.exe projects\\robots\\softbank\\nao\\controllers\\nao_demo\\nao_demo.exe
The started controller targets a local instance (ipc protocol) of Webots with port number 1234. Targeting the only robot waiting for an extern controller.

Error: wbu_motion_new(): could not open '../../motions/HandWave.motion' file.
Error: wbu_motion_new(): could not open '../../motions/Forwards50.motion' file.
Error: wbu_motion_new(): could not open '../../motions/Backwards.motion' file.
Error: wbu_motion_new(): could not open '../../motions/SideStepLeft.motion' file.
Error: wbu_motion_new(): could not open '../../motions/SideStepRight.motion' file.
Error: wbu_motion_new(): could not open '../../motions/TurnLeft60.motion' file.
Error: wbu_motion_new(): could not open '../../motions/TurnRight60.motion' file.
Error: wbu_motion_new(): could not open '../../motions/TaiChi.motion' file.
Error: wbu_motion_new(): could not open '../../motions/WipeForehead.motion' file.
----------nao_demo----------
Select the robot and use the keyboard to control it:
(The 3D window need to be focused)
[Up][Down]: move a few steps forward/backwards
[<-][->]: make a few side steps left/right
[Shift] + [<-][->]: turn left/right
[U]: print ultrasound sensors
[A]: print accelerometer
[G]: print gyro
[S]: print gps
[I]: print inertial unit (roll/pitch/yaw)
[F]: print foot sensors
[B]: print foot bumpers
[Home][End]: print scaled top/bottom camera image
[PageUp][PageDown]: open/close hands
[7][8][9]: change all leds RGB color
[0]: turn all leds off
[T]: perform a tai chi move
[W]: wipe its forehead
[H]: print this help message
Error: wbu_motion_set_loop() called with NULL argument.
Error: wbu_motion_play() called with NULL 'motion' argument.
```

omichel:
The forward slashes make the launcher crash on Windows:
```
$ webots-controller.exe projects/robots/softbank/nao/controllers/nao_demo/nao_demo.exe
The started controller targets a local instance (ipc protocol) of Webots with port number 1234. Targeting the only robot waiting for an extern controller.

Segmentation fault
```
We should support both the forward and backward slashes.

docs/guide/running-extern-robot-controllers.md
```diff
@@ -255,13 +159,28 @@ This variable is computed from the `SNAP_USER_COMMON` environment variable which
...
+Matlab controllers can also be started using the launcher.
+By default, the launcher will look for the latest installed version of MATLAB in the following locations, depending on the OS:
+
+- **Windows**: C:\Program Files\MATLAB\R20XXx\bin\win64\MATLAB.exe
+- **Linux**: /usr/local/MATLAB/R20XXx/bin/matlab
+- **MacOS**: /Applications/MATLAB_R20XXx.app
+
+It is also possible to give a custom MATLAB installation path to the launcher, by providing the absolute path to the executable in the `--matlab-path` option.
+
 ## Example Usage
```
ad-daniel, line 174: In these instructions we should perhaps mention that the WEBOTS_HOME variable needs to be set

ad-daniel (changes requested):
I get a segmentation fault on Linux + Development environment for Java & Python & CPP:
- relocate the java sample to documents
- set driver to EXTERN
- `cd` to controller location and launch with `$WEBOTS_HOME/webots-controller Driver.class`

I get: 
```
The started controller targets a local instance (ipc protocol) of Webots with port number 1234. Targeting the only robot waiting for an extern controller.

Segmentation fault (core dumped)
```
Same behavior for python & cpp. Running it with `./` works.

Also, tangentially, if I remember right for ROS2 you had to do some changes to the launch files so that some prefixes are included in case of WSL (I think), couldn't this sort of stuff be moved at the controller launcher level as well?

I'll keep testing afterwards on actual releases

docs/guide/running-extern-robot-controllers.md
```diff
@@ -255,13 +159,28 @@ This variable is computed from the `SNAP_USER_COMMON` environment variable which
...
 4. Save the simulation, restart it and run it.
-5. Open the "WEBOTS\_HOME/projects/robots/softbank/nao/controllers/nao_demo" folder from a terminal.
-6. Setup environment variables needed for a C/C++ controller as explained in the above section.
-7. Start the `nao_demo` controller manually from the terminal.
-8. You should see the Nao robot moving in the simulation, controlled by the `nao_demo` program you just started.
+5. Open a terminal and go to the "WEBOTS\_HOME/projects/robots/softbank/nao/controllers/nao_demo" folder. Note that other controllers could be started from anywhere, but in the example of the Nao, motion files are given relative to this folder. Executing the controller from another location would lead to an error in finding these files.
+6. Start the `nao_demo` controller with:
+
+  ```bash
+  $WEBOTS_HOME/webots-controller ./nao_demo  --protocol=ipc --port=1234 --robot-name=NAO
```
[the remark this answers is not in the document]
omichel: Also, the `--port` and `--robot-name` options can be omitted as they are the default values.

ad-daniel (changes requested):
I've tested the tar on a local distrib and the only further issue I found is that if the robot name contains a space, it doesn't connect to it

As for the snap, I think we need to wait for the PR to be merged and one being generated in the nightly to ensure everything is available

src/controller/launcher/webots_controller.c
```diff
@@ -0,0 +1,843 @@
...
+void remove_comment(char *string) {
+  const char *comment = strchr(string, ';');
+  if (comment) {
+    const size_t comment_size = strlen(comment);
+    const size_t full_line_size = strlen(string);
+    const size_t content_size = full_line_size - comment_size;
+    string[content_size] = '\0';
+    return;
+  }
+  string[strlen(string)] = '\0';
```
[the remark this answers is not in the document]
ygoumaz: For lines without comments, sometimes there are some trailing whitespaces that need to be removed to correctly parse the line further in the code.

docs/guide/running-extern-robot-controllers.md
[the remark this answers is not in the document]
ygoumaz: Yes it is correct that here no option is necessary, as these are default values, but because this is an example I wanted to show the usage of options. Maybe I can change it to tcp, so that `protocol` and `ip-address` options make sense.

src/controller/launcher/webots_controller.c
```diff
@@ -0,0 +1,843 @@
...
+}
+
+// Removes all occurrences of a character from a string.
+void remove_char(char *string, char occurence) {
+  char *removed = string;
+  do {
+    while (*removed == occurence) {
+      ++removed;
+    }
+  } while ((*string++ = *removed++));
```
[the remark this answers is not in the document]
ygoumaz: These additional parenthesis are useful on macOS or warnings are shown at compilation time.

src/controller/launcher/webots_controller.c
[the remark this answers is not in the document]
omichel: `strlen` computes the length of `string`, which means it count characters in `string` until if finds `'\0'`. So, setting the last byte to `'\0'` is useless as the value there should already be `'\'0'`, e.g., `assert(string[strlen(string)] == 0);`.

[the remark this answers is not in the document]
ygoumaz: Yes you're right, I think this code was different at the beginning and was making sense, but not anymore, as it works without and is completely meaningless.

src/controller/launcher/webots_controller.c
```diff
@@ -0,0 +1,834 @@
...
+#ifdef _WIN32
+    char *controller_formated = NULL;
+    if (strstr(controller, " ") != NULL) {
+      const size_t controller_formated_size = snprintf(NULL, 0, "\"%s\"", controller) + 1;
+      controller_formated = malloc(controller_formated_size);
+      sprintf(controller_formated, "\"%s\"", controller);
+    } else
+      controller_formated = strdup(controller);
+
+    const char *const new_argv[] = {"python", controller_formated, NULL};
```
omichel, line 718: If we use the `--stdout-redirect` option, we should add the "-u" option to the python command line so that the `stdout` is actually flushed (e.g., sent to Webots), otherwise it is buffered by default and doesn't show up in Webots console unless the user explicitly flushes `stdout`.

ygoumaz: Nice catch! Thank you!

ygoumaz:
I have resolved all mentioned issues, including the ability to launch the extern controller from any location, as it will automatically execute from its own directory. Running the controller launcher within the target controller's directory can be done without the need for "./" prefix, and Windows users can now use both forward and backward slashes. Robot names with spaces are now supported. Although I have tested in all languages and platforms, there may still be undiscovered edge cases, and I could not retest Matlab as my license has expired since my last test.

ad-daniel (dismissed):
Nothing else from me, works well on linux both for python, C and java

src/controller/launcher/webots_controller.c
```diff
@@ -0,0 +1,877 @@
...
+          break;
+        case Simple:
+          remove_comment(runtime_ini_line);
+          new_env_ptr = strdup(runtime_ini_line);
+          putenv(new_env_ptr);
+          break;
+        case Windows:
+#ifdef _WIN32
+          if (!strchr(runtime_ini_line, '\"')) {
+            fprintf(stderr, "Paths for windows should be written between double-quotes symbols \".\n");
```
omichel, line 625:
```suggestion
            fprintf(stderr, "Windows paths should be written between double-quotes symbols \".\n");
```

omichel (changes requested):
It looks good to me.
I just have a very minor change request.

ad-daniel (approved):

omichel (approved):
All good for me.
Thank you.

PR #5635, opened 2022-12-06
Add pen optional rendering to the web proto viewer

BenjaminDeleze:
Test it here:
https://proto.webots.cloud/run?version=proto&url=https://github.com/BenjaminDeleze/sandbox/blob/master/protoExample/protos/ProtoSensor.proto


ad-daniel (changes requested):
In webots when I change the rotation, the pen ray follows it, here it doesn't appear to be the case. Is that intended?

BenjaminDeleze:
> In webots when I change the rotation, the pen ray follows it, here it doesn't appear to be the case. Is that intended?

The `IS` of the rotation was in the `lightSensor` not in the `robot` that is why the `pen` was not rotating.
It should be good now

ad-daniel (approved):
I see, thank you!

PR #5638, opened 2022-12-07
Improve enable/disable pose tracking API function

### LINKED ISSUE #5590: Webost crash deleting nodes for which pose tracking is enabled
I have robot controller script that imports objects, tracks the pose of these objects, then removes them and generates a set of other object, etc.
 
 At some point during the simulation, Webots crashes with this stack
```
0x0000555555badeb8 in WbBaseNode::isPreFinalizedCalled (this=0x0) at nodes/WbBaseNode.hpp:55
55	  bool isPreFinalizedCalled() const { return mPreFinalizeCalled; }
(gdb) y
Undefined command: "y".  Try "help".
(gdb) bt
#0  0x0000555555badeb8 in WbBaseNode::isPreFinalizedCalled() const (this=0x0) at nodes/WbBaseNode.hpp:55
#1  0x0000555555fbad1b in WbAbstractTransform::matrix() const (this=0x5555808dbf28)
    at nodes/WbAbstractTransform.cpp:353
#2  0x0000555555eee1d5 in WbSupervisorUtilities::pushRelativePoseToStream(WbDataStream&, WbTransform*, WbTransform*)
    (this=0x555560669820, stream=..., fromNode=0x0, toNode=0x5555808dbd80)
```

Webots should properly handle the deletion of nodes and suspend immediately and automatically the pose tracking.

### LINKED ISSUE #5592: Default NULL argument is invalid in `wb_supervisor_node_disable_pose_tracking`
It is not consistent that the `wb_supervisor_node_enable_pose_tracking` supports a NULL `from_node` argument, but the `wb_supervisor_node_disable_pose_tracking` doesn't support it.
It should work the same in enable and disable methods and a NULL `from_node` should be "translated" in both cases to the root node.

stefaniapedrazzi:
Tasks for the function for tracking pose, contact points and field values:
* [x] fix #5590,: Webots crashing when tracked node is deleted
* [x] add additional warning messages if wrong parameters are given and improve documentation.
* [x] extend tests to check Webots robustness to deleting tracked nodes 

Additionally:
* [x] fix #5592 so that the NULL default argument is now supported for disabling pose tracking.
* [x] fix test compilation failure after  #5633


src/controller/c/supervisor.c
```diff
@@ -3015,7 +3015,7 @@ void wb_supervisor_node_disable_pose_tracking(WbNodeRef node, WbNodeRef from_nod
     return;
   }
 
-  if (!is_node_ref_valid(from_node)) {
+  if (from_node != NULL && !is_node_ref_valid(from_node)) {
     if (!robot_is_quitting())
       fprintf(stderr, "Error: %s() called with a NULL or invalid 'from_node' argument.\n", __FUNCTION__);
```
omichel, line 3020: It cannot be `NULL` here as we test explicitly for it:
```suggestion
      fprintf(stderr, "Error: %s() called with an invalid 'from_node' argument.\n", __FUNCTION__);
```

src/webots/nodes/utils/WbSupervisorUtilities.cpp
```diff
@@ -550,6 +550,32 @@ void WbSupervisorUtilities::updateDeletedNodeList(WbNode *node) {
...
+      break;
+    }
+  }
+}
+
+void WbSupervisorUtilities::removeTrackedPoseNode(QObject *obj) {
+  for (int i = mTrackedPoses.size() - 1; i >= 0; --i) {
+    if (mTrackedPoses[i].fromNode == obj || mTrackedPoses[i].toNode == obj) {
+      mTrackedPoses.removeAt(i);
+    }
```
omichel, line 566:
```suggestion
    if (mTrackedPoses[i].fromNode == obj || mTrackedPoses[i].toNode == obj)
      mTrackedPoses.removeAt(i);
```

omichel (changes requested):

src/webots/nodes/utils/WbSupervisorUtilities.cpp
```diff
@@ -1263,29 +1301,55 @@ void WbSupervisorUtilities::handleMessage(QDataStream &stream) {
...
+        } else {
+          mTrackedPoses[index].samplingPeriod = samplingPeriod;
+          mTrackedPoses[index].lastUpdate = -INFINITY;
+        }
+      } else if (index >= 0)
+        mTrackedPoses.removeAt(index);
+      else
+        mRobot->warn(tr("No active pose tracking could be found matching the node '%1' and 'from_node' argument.")
+                       .arg(toNode->modelName())
+                       .arg(fromNode->modelName()));
```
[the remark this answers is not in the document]
stefaniapedrazzi: If we use the `usefulName()` function we do not get the node type in case of a PROTO node.
So even if it will make it easier to identify which node is causing the issue, it won't give much information why the node is invalid.

By the way I used the same style as the other messages in the `WbSupervisorUtilities` class.
In all the other similar warning messages printed in the `WbSupervisorUtilities` class we always use `modelName()` and none of the messages is printing the `usefulName()`.

src/webots/nodes/utils/WbSupervisorUtilities.cpp
```diff
@@ -1263,29 +1303,59 @@ void WbSupervisorUtilities::handleMessage(QDataStream &stream) {
...
+            connect(fromTransformNode, &WbNode::destroyed, this, &WbSupervisorUtilities::removeTrackedPoseNode);
+          connect(toTransformNode, &WbNode::destroyed, this, &WbSupervisorUtilities::removeTrackedPoseNode);
+        } else {
+          mTrackedPoses[index].samplingPeriod = samplingPeriod;
+          mTrackedPoses[index].lastUpdate = -INFINITY;
+        }
+      } else if (index >= 0)
+        mTrackedPoses.removeAt(index);
+      else
+        mRobot->warn(tr("No active pose tracking could be found matching nodes '%1' (to) and '%3' (from) arguments.")
```
omichel, line 1355:
```suggestion
        mRobot->warn(tr("No active pose tracking could be found matching nodes '%1' (to) and '%2' (from) arguments.")
```

omichel (changes requested):
In addition to my change request, the tests are failing.

omichel (approved):
Thank you.

Issue #5586, opened 2022-11-28
Memory leak in the PBRAppearance node


ad-daniel:
**Describe the Bug**

1. Load the world [leak.zip](https://github.com/cyberbotics/webots/files/10102673/leak.zip), play it fast mode
2. The human is removed and spawned over and over
3. RAM consumption increases over time

Surprisingly enough, if you rename the textures folder to something else (so that they aren't found), the leak seems even more evident. Removing the appearance nodes from the PROTO lowers the rate but doesn't extinguish the leak, so there might be something in the IndexedFaceSet node as well. Doing a similar test but with a CadShape in the PROTO doesn't appear to present any leaks (there are RAM jumps, but seems to go back down just as much)

EDIT: CadShape might also increase on average, albeit slower. Having a CadShape but not in a PROTO doesn't appear to though.

ad-daniel:
The leak might actually not be from the PBR node, I've tested with the old FlowerPot proto (18 MB proto + ~10MB of textures, available [here](https://github.com/cyberbotics/webots/pull/4166)) and although the RAM increases slightly, it's nowhere near what can be seen with the human. So the issue might be related to the console rather than the appearance (the human PROTO throws several warnings due to intersecting faces, but it's just a guess)

ad-daniel:
It appears that it doesn't seem to be related to PROTO nor appearances,
It's visible just by spawning: 
`rootChildrenField.importMFNodeFromString(-1, 'DEF BOX Shape { geometry Box {} }')`

If instead we use a mesh node:
`rootChildrenField.importMFNodeFromString(-1, 'DEF BOX Shape { geometry Mesh { url "../protos/human_02/human_02_standing.obj" } }')`

the memory is stable. Other primitives (Cone, Sphere) are almost negligible too, since it seems to be evident only with Box, I suspect it might not be related to the original leak but rather be a separate one...

PR #5639, opened 2022-12-07
Fix memory leaks

ad-daniel:
**Description**

Fixes https://github.com/cyberbotics/webots/issues/5586

The following bugs contributed to the leaks in [leak.zip](https://github.com/cyberbotics/webots/files/10174724/leak.zip). The leak becomes evident when a lot of supervisor actions are involved, primarily when adding and deleting nodes while running the simulation in no-rendering mode.

- PhongMaterial & PBRMaterial: phong especially is created for resize manipulators, which would be left unclean when the nodes are deleted
- StaticMesh: arrays used to define shadows aren't cleared. Currently these arrays are created even for line sets which is unnecessary, this will be addressed in a separate PR.
- 3dview: the problem stems from the BRDF texture created [here](https://github.com/cyberbotics/webots/blob/d02be09b76f545d0bda919b471349af03a04159e/src/webots/nodes/WbPbrAppearance.cpp#L230), or more precisely  TextureRtt [here](https://github.com/cyberbotics/webots/blob/d02be09b76f545d0bda919b471349af03a04159e/src/wren/TextureCubeMapBaker.cpp#L335). When this texture is deleted [here](https://github.com/cyberbotics/webots/blob/d02be09b76f545d0bda919b471349af03a04159e/src/wren/Texture.cpp#L79-L88) it's queued for deletion (L87), but in no-rendering mode the action that is supposed to flush them ([here](https://github.com/cyberbotics/webots/blob/d02be09b76f545d0bda919b471349af03a04159e/src/wren/GlUser.cpp#L32)) never gets called and so the memory keeps rising. Pausing the simulation, running one step, or re-activating the rendering would trigger a clean, but if running continuously in no-rendering mode it won't happen. The proposed fix is to allow for wren cleanups even when the no-rendering option is active.


stefaniapedrazzi (changes requested):
The changes are fine.
I just have a couple of comments.

ad-daniel:
You're right, thank you. I've rechecked if there was a performance impact just in case but it doesn't appear to.

stefaniapedrazzi (approved):
Thank you!

PR #5641, opened 2022-12-07
Display distance sensor in web proto viewer

BenjaminDeleze:
Improve the completness by adding support for distance sensor rays.

- [x] Distance Sensor optional rendering
- [x] Adapt the `updateGl` function of `wren/DynamicMesh.cpp` to support webgl
- [x] Add support for MFString in the parameter tab 

Test it here:
https://proto.webots.cloud/run?version=proto&url=https://github.com/BenjaminDeleze/sandbox/blob/master/protoExample/protos/ProtoSensor.proto


ad-daniel (approved):
Looks good, and the behavior is intuitive, thank you!

PR #5644, opened 2022-12-08
Add delete function in devices

BenjaminDeleze:
The `delete` function was missing and is needing in particular when a proto can be regenerated.


ad-daniel (approved):
Thank you

PR #5660, opened 2022-12-13
Slider joint in web proto viewer

BenjaminDeleze:
Add slider for slider joint

But also modify the loading process such that the world and the proto are loaded at the same time.

Test it here:
https://proto.webots.cloud/run?version=proto&url=https://github.com/BenjaminDeleze/sandbox/blob/master/protoExample/protos/Cabinet.proto

Note: it is normal that the parameters tab does not work, because the IS are not connected to anything in this example (it uses a modified `cabinet.proto`)


omichel (changes requested):
Instead of displaying the Hingejoint number, I would display the name of the endPoint.

Issue #5466, opened 2022-11-11
Need an error when choosing a wrong controller?


omichel:
### Discussed in https://github.com/cyberbotics/webots/discussions/5464

<div type='discussions-op-text'>

<sup>Originally posted by **amrytov** November 11, 2022</sup>
Hi,

This is not an important thing but still may be worth noting (and eventually fixing):

In an educations scene I use a PROTO robot, in which controller names are restricted like this:

      field  SFString{"code_py", "code_c"}  controller "code_c"

The controllers are some unfinished code templates for the students to complete, in their language of choice.  
This works perfectly.  

However,  there is still the [Select...] button below, which knows nothing about the restrictions
and shows a whole bunch of other controllers the students should not even see.  The good thing is when a student
chooses a controller which is not in the allowed list, nothing happens (the controller field does not change).  
The bad thing is that there is absolutely no indication that something wrong has happened.  In similar situations when one tries to enter a wrong field value, there is usually an error message in the console. 

Would be good to add one of the following:

 - Perfect:  if there is a restriction list on the 'controller' field, the "Choose..." button is hidden or disabled
 - Less perfect:  in such situation the "Choose..." button works, but shows only the allowed values
 - Somewhat ok:  no change in the "Choose...",  but at least report an error ("wrong value") to console when trying to assign an invalid value.
</div>

PR #5663, opened 2022-12-14
Disable Select button in WbExtendedStringEditor for restricted fields

stefaniapedrazzi:
Address  #5466:
in the `WbExtendedStringEditor` class check if the current field has restricted values and, if it is the case, hide the `Select...` button.

I evaluated the possibility to check the list of accepted values and remove the ones that are not available.
But it seems clearer and easier to understand for the user if all the accepted values are shown in the combo box and Webots throws an error in the console if the requested resources is not found.
In case the value, for example the controller, is not found or not compiled Webots will still use the default one even if it is not in the list (e.g. the <generic> controller). 

Screenshot from 2022-12-14 11-38-40


ad-daniel (approved):
works well, thank you

PR #5664, opened 2022-12-14
Enhancement hinge2joints slider

BenjaminDeleze:
Test it here:
https://proto.webots.cloud/run?version=proto&url=https://github.com/BenjaminDeleze/sandbox/blob/master/protoExample/protos/Nao.proto


omichel (changes requested):
It looks good to me, but I would use 1 and 2 instead of a and b for the Hinge2Joint axes, since we use the "2" terminology for the second axis in the docs and field names.

omichel (approved):
Thank you.

PR #5669, opened 2022-12-15
Fix crash enabling battery while simulatoin is running

stefaniapedrazzi:
A user reported this crash:
> Everytime I introduce the battery to a new robot Webots crashes and afterwords the charger is not working anymore.

Steps to reproduce:
1. open "battery.wbt" in pause mode
2. reset the `MyRobot.battery` field to default (empty)
3. save and revert the world
4. run the simulation
5. add an item (default value is 0) to the `MyRobot.battery`
6. set the `MyRobot.battery` item to 100
-> Webots crashes

When the battery current energy is greater than 0 It is expected that the motor feedback is enabled, but the code only checks if a battery is used (and this if the motor feedback should be setup) when finalizing the motor node.

Also note that the documentation mentions that the `Robot.battery` field should contain three values.
However, the code is implemented in such a way that even if the `Robot.battery` has just one value, the battery energy is updated correctly. This is just a note, I won't change the documentation.

### DIFF
--- a/src/webots/nodes/WbMotor.cpp
+++ b/src/webots/nodes/WbMotor.cpp
@@ -139,7 +139,7 @@ void WbMotor::preFinalize() {
 void WbMotor::postFinalize() {
   WbJointDevice::postFinalize();
   assert(robot());
-  if (!mMuscles->isEmpty() || robot()->maxEnergy() > 0)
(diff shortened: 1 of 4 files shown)


BenjaminDeleze (approved):
Thank you, I could not reproduce the bug anymore

PR #5671, opened 2022-12-16
Fix wrong assertion failure for console redirection

ygoumaz:
When running an extern controller with WEBOTS_STDOUT_REDIRECT set to 1, the logs are redirected to the Webots console. However, this causes Webots to crash in debug compilation mode because of an assertion. 

If nothing is written to the buffer at a timestep, the buffer was not set back to NULL.

### DIFF
--- a/src/controller/c/robot.c
+++ b/src/controller/c/robot.c
@@ -158,9 +158,12 @@ void stream_pipe_read(int fd, char **buffer) {
   if (len == -1)
     len = 0;
 #endif
-  (*buffer)[len] = '\0';
-  if (!len)
+  if (len != 0)
+    (*buffer)[len] = '\0';
+  else {
+    free(*buffer);
     *buffer = NULL;
+  }
 }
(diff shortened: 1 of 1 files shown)


src/controller/c/robot.c
```diff
@@ -159,6 +159,8 @@ void stream_pipe_read(int fd, char **buffer) {
     len = 0;
 #endif
   (*buffer)[len] = '\0';
+  if (!len)
+    *buffer = NULL;
```
omichel, line 163: That's a memory leak: the pointer to the allocated 1024 bytes buffer is lost, so the memory cannot be free.
```suggestion
  if (len != 0)
    (*buffer)[len] = '\0';
  else {
    free(*buffer);
    *buffer = NULL;
  }
```

omichel (changes requested):

omichel (approved):
Thank you.

PR #5359, opened 2022-10-20
Fix duplicated CadShape mesh load and improve warnings

stefaniapedrazzi:
Address #5356 and ##5313:
* [x] Add the same warning about complex mesh with more than 100'000 vertices in the Mesh node
* [x] Add warning in CadShape node when the imported mesh is fully transparent -> otherwise users may think that the import is buggy
* [x] Avoid creating WREN objects twice when finalizing the CadShape node.

PR #5674, opened 2022-12-19
Fix cadshape move to viewpoint

BenjaminDeleze:
The `recomputeBoundingSphere` function in `WbCadShape` was called in a different place than in `WbGroup` and `WbGeometry` and I could not find out why.

This causes, since https://github.com/cyberbotics/webots/pull/5359, the `move Viewpoint to Object` to not work.

@ad-daniel maybe you remember why it was done differently in `WbCadShape`?

### DIFF
--- a/src/webots/nodes/WbCadShape.cpp
+++ b/src/webots/nodes/WbCadShape.cpp
@@ -162,6 +162,7 @@ void WbCadShape::postFinalize() {
           &WbCadShape::createWrenObjects);
 
   mBoundingSphere = new WbBoundingSphere(this);
+  recomputeBoundingSphere();
 
   // apply segmentation color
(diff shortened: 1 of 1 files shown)


ad-daniel:
No reason that I can think of, likely thought it didn't matter which in practice it seems to

ad-daniel (changes requested):
Actually I think I know why it was defined this way, the reason is that I used WbSkin as a template when defining CadShape, as such if this was broken, that one is likely to be too

BenjaminDeleze:
> Actually I think I know why it was defined this way, the reason is that I used WbSkin as a template when defining CadShape, as such if this was broken, that one is likely to be too

I justed tested with the `sample/devices/worlds/skin.wbt` and I could move to Skin without problem.

ad-daniel (approved):
Thank you

PR #5675, opened 2022-12-19
remove unused variable and fix default axis

BenjaminDeleze:
ditto


ad-daniel (approved):
Nice catch, thank you

Issue #5478, opened 2022-11-14
Robot window plots should ignore NaN values or there shouldn't be NaN values to begin with


ad-daniel:
**Steps to Reproduce**
1. Open `languages > example.wbt`
2. Open the robot window of `ROBOT3` and play
3. Go to `PositionSensor` tab, override controller and enable the devices
4. The plots don't seem to update for a long time, likely due to the presence of NaN values in the buffer (unsure why). When these no longer are present, the plot is shown correctly.

PR #5680, opened 2022-12-20
Improve robustness to NaN values in robot window plots

stefaniapedrazzi:
Address #5478:
ignore NaN values when adapting the plot range in the default robot window, otherwise subsequent valid values are not correctly displayed in the plot until the NaN value goes out of scope (after some time).


ad-daniel (approved):
Thank you

PR #5684, opened 2022-12-21
Enhancement display axis system for other devices

BenjaminDeleze:
[x] Display optional rendering (axis system) at the position of devices that do not have a specific optional rendering (TouchSensor, GPS, Accelerometer, ...)
- [x] Display the bounding object linked to TouchSensor
- [x] Modifiy the Viewpoint according to the size of the initial boundingSphere
- [x] Fix the deletion of node that was not correctly call in `connectors` and `joints`
- [x] Organize the devices in categories
- [x] Joints limits take motor's limits in account
- [x] Display joints details

Test it here: https://proto.webots.cloud/run?version=proto&url=https://github.com/BenjaminDeleze/sandbox/blob/master/protoExample/protos/Pr2.proto or on other protos on [proto.webots.cloud/proto](https://proto.webots.cloud/proto)


ad-daniel (changes requested):
Looks good. I found some odd behaviors:
- Spot: all "solid" joints don't seem to be right
- CoupledMotor: changes of "right pivot" should be coupled

Concerning the viewpoint, since the proto parameter window is open by default, should we perhaps "consider" its presence (the space it takes away) when centering the robot so that it's centered in the visible portion of the 3d view? (just for the first load, if the window is moved around after too bad). Otherwise all robots don't seem centered (although they are). Beyond that I have the feeling the vertical centering isn't correct (for example: tinkerbotscubetriangle), assuming that wasn't intentional

BenjaminDeleze:
> * Spot: all "solid" joints don't seem to be right

I agree that they are strange, but I check in the `.proto` and they are like that. But maybe the proto is wrong

> * CoupledMotor: changes of "right pivot" should be coupled

CoupledMotor support is not implemented in this PR.


> Concerning the viewpoint, since the proto parameter window is open by default, should we perhaps "consider" its presence (the space it takes away) when centering the robot so that it's centered in the visible portion of the 3d view? (just for the first load, if the window is moved around after too bad). Otherwise all robots don't seem centered (although they are). Beyond that I have the feeling the vertical centering isn't correct (for example: tinkerbotscubetriangle), assuming that wasn't intentional

So this is a good idea but I don't think it is going to work because, to determine the viewpoint, I am using the `moveViewpointToObject` function (the same as in Webots) so a pure Webots function that is internal to the canvas  but the proto parameters is on top of the canvas, so I am not sure how to mix the `moveViewpointToObject` with some parameters like the size of the windows, the proto parameter window and so on.
Moreover as the proto parameter window is semi-transparent I think it is ok.

For the vertical alignement, I get the same result in Webots when doing `moveViewpointToObject`, it may feel like not centered because the original viewpoint is not looking straight ahead but is a bit turned toward the ground.

ad-daniel (approved):
Thank you

PR #5689, opened 2022-12-22
Cleanup libcontroller sync in test suite

ygoumaz:
The libcontroller pre-compiled binaries are not used anymore, as they are compiled on the `webots_ros2` side. The synchronization of the libcontroller source files is now performed on the webots_ros2 side too. 

This PR cleans the old sync process to the libcontroller repository.


omichel (approved):
Thank you.

PR #5687, opened 2022-12-21
Troubleshooting for M1 errors caused by Rosetta

kodecreer:
Provided troubleshooting for people who run into issues when trying to run the Webots application on a Mac M1. Specifically errors that are caused due to trying to run the application using Rosetta.

**Description**
Added Troubleshooting for users on Mac M1

**Related Issues**
This pull-request fixes issue #5593
https://github.com/cyberbotics/webots/issues/5593

**Tasks**
Add the list of tasks of this PR.

**Documentation**
This is documentation in. itself.

PR #5691, opened 2022-12-23
Add a troubleshooting section for macOS installation with respect to Rosetta

omichel:
Follow-up of #5687 targeting the released branch instead of the master branch.
See the result here: https://www.cyberbotics.com/doc/guide/installation-procedure?version=doc-macos-rosetta-troubleshooting


BenjaminDeleze (approved):
Thank you

PR #5705, opened 2023-01-03
Fixed crash in Python API when using several cameras with different resolutions

### LINKED ISSUE #5703: Multi-camera system appear to be non-functional.
**Describe the Bug**

Since R2023a, multi-camera system appear to be non-functional.
The robot should have 2 cameras, one with resolution 400x300 and the other 1200x1200.

A minimal example to reproduce the issue is the following controller:
```
from controller import Robot

robot = Robot()

cam = robot.getDevice('camera')
cam2 = robot.getDevice('camera2')

cam.enable(timestep)
cam2.enable(timestep)

robot.step(timestep)

image = cam.getImage()
image2 = cam2.getImage()

  pass
```

When requesting the width/height of the 2 camera, they are correct. The line `image = cam.getImage()` however results in the controller crashing. From user report, it seems that in some situations the controller does not crash however the camera images cannot be resized because the data buffers have the wrong dimensions. What might be happening is that the acquisitions are pointing to the wrong objects.

The same controller above in C doesn't crash, likewise R2022b doesn't appear to have the issue so it might be related to the new Python API

omichel:
Fixes #5703.

### DIFF
--- a/docs/reference/changelog-r2023.md
+++ b/docs/reference/changelog-r2023.md
@@ -10,7 +10,7 @@ Released on ??
     - Disable `Select..` button in SFString editor if the field has restricted values ([5663](https://github.com/cyberbotics/webots/pull/5663)).
     - Improved plot representation in default robot window when a NaN value is received from a device ([#5680](https://github.com/cyberbotics/webots/pull/5680)).
   - Bug Fixes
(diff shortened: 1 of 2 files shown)


ad-daniel (approved):
Works fine on my end, thank you!

PR #5709, opened 2023-01-04
Fix `openai_gym` instructions

ad-daniel:
**Description**

`stable_baselines3` is only compatible with gym up to version 0.21.0, an error is currently thrown when following the instructions.


BenjaminDeleze (approved):
Thank you

PR #5711, opened 2023-01-04
Support coupled motors on slider update

BenjaminDeleze:
- Add support for coupled motors in web proto viewer
- Correct a bug where the rotation of a joint was wrong if the initial position was not 0.

You can test it here:
https://proto.webots.cloud/run?version=proto&url=https://github.com/BenjaminDeleze/sandbox/blob/master/protoExample/protos/Panda.proto


omichel (approved):
Nice.

PR #5715, opened 2023-01-05
Fix error when running individual tests.

brettle:
**Description**
Change test_suite.py's generateWorldsList()  to return a list of strings instead of a list of lists when generating the list from file arguments. Returning a list of lists resulted in this error when running an individual test (e.g. `./test_suite.py api/worlds/center_of_mass_and_contact_points.wbt`):

```
  File "/home/brettle/git/webots/tests/./test_suite.py", line 345, in <module>
    f.write(world + '\n')
            ~~~~~~^~~~~~
TypeError: can only concatenate list (not "str") to list
```


ad-daniel (approved):
nice, thank you!

PR #5716, opened 2023-01-06
Fix race condition when testing node removal.

brettle:
**Description**

Change the controllers associated with the `supervisor_import_remove.wbt` test so that the `SPHERE2` node does not get removed by `wb_supervisor_node_remove()` until a full step after it's been checked for. This is necessary because `wb_supervisor_node_remove()` is immediate. Without this change, the test fails for me approximately 20% of the time with "Wrong DEF name for top node at index 5: found , expected SPHERE2."


ad-daniel (approved):
I've seen it pop up at times, nice catch, thank you.

PR #5720, opened 2023-01-06
Add support for field restrictions to web proto

ad-daniel:
**Description**

This PR adds field restriction support for:
- SFNode & MFNode (when implemented): the restriction is enforced in the node selection window
- SFString, SFFloat, SFInt32, SFVec2f, SFVec3f, SFColor, SFRotation: supported
- MF*: unsupported (no cases in Webots library, not user friendly anyway)
- SFBool: unsupported (makes no sense)

Status:
- [x] Decode restrictions from VRML
- [x] Handle reset
- [x] Handle webotsjs notification
  - [x] SFString
  - [x] SFFloat
  - [x] SFInt32
  - [x] SFVec2f
  - [x] SFVec3f
  - [x] SFColor
  - [x] SFRotation
  - [x] SFNode
    - [x] Filter node list based on   
- [x] Cleanup

Preview:
https://proto.webots.cloud/run?version=proto&url=https://github.com/ad-daniel/prefab/blob/master/protoExamples/protos/TestRestrictions.proto

& 

https://proto.webots.cloud/run?version=proto&url=https://github.com/BenjaminDeleze/sandbox/blob/master/protoExample/protos/Nao.proto


resources/web/wwi/protoVisualizer/Parameter.js
```diff
@@ -28,6 +38,15 @@ export default class Parameter {
     if (v.type() !== this.type)
       throw new Error('Type mismatch, setting ' + stringifyType(v.type()) + ' to ' + stringifyType(this.type) + ' parameter.');
 
+    if (this.restrictions.length > 0) {
+      let isValueAcceptable = false;
+      for (const item of this.restrictions)
+        isValueAcceptable = isValueAcceptable || (v instanceof SFNode && v.value === null) || item.equals(v);
```
BenjaminDeleze, line 44: Wouldn't it be better to break from the loop as soon as `isValueAcceptable` is set to true or maybe use a Set to completely avoid the loop?

resources/web/wwi/FloatingProtoParameterWindow.js
```diff
@@ -306,8 +326,11 @@ export default class FloatingProtoParameterWindow extends FloatingWindow {
     p.style.gridColumn = '2 / 2';
     p.className = 'key-parameter';
 
-    const exportCheckbox = this.#createCheckbox(parent, key);
-    p.checkbox = exportCheckbox;
+    if (this.proto.isRoot) {
+      const exportCheckbox = this.#createCheckbox(parent, key);
+      p.checkbox = exportCheckbox;
+    } else
+      p.style.marginLeft = '20px';
```
BenjaminDeleze, line 333: Isn't this part already in `feature-web-proto`?

BenjaminDeleze (changes requested):
Works well! 
Only two minor remarks:

resources/web/wwi/FloatingProtoParameterWindow.js
ad-daniel: For all others yes, but a couple were missing apparently so I assume they must be node types that have been added later on and didn't adapt them when merging it into my PR

resources/web/wwi/protoVisualizer/Parameter.js
ad-daniel: The loop can indeed be improved, but using a Set might be overkill since in practice we usually have at most a couple restrictions

BenjaminDeleze (approved):
Thank you

PR #5722, opened 2023-01-06
Fix Fabtino wheel orientation

ad-daniel:
**Description**

Fabtino's left wheels are wrong:

before 
b
after
a

It's just a visual bug, the bounding object is correct so should have no impact for the users


omichel (approved):
Thank you.

Issue #4925, opened 2022-07-20
Ease the switch between multiples objects Mass / Position / Velocity


ShuffleWire:
**Describe the Bug**
Given a world where multiples object are moving, I often wish to monitor their Velocity (for instance, but also Position or Mass info) using the GUI and the Node/Mass/Position/Velocity tabs.
But this windows always default to the Node tab each time I change the object selected, which force me to re-select the correct tab. 

**Steps to Reproduce**
1. Open a world with multiple moving object
2. Click on one object
3. Go into the Velocity tab
4. Click on another object
5. See that the GUI come back to the Node tab.

**Expected behavior**
It would be perfect to keep the tab Velocity selected, hence I could easily switch back and forth between object.

**System**
 - Operating System: Debian 11
 - Webots 2022a

PR #5726, opened 2023-01-11
Keep track of selected node pane tab

stefaniapedrazzi:
Address #4925:
keep track of the selected tab in the Field Editor for nodes so that it can automatically restored when selecting a different node.


BenjaminDeleze (approved):
Thank you

PR #5730, opened 2023-01-12
Handle blank field

BenjaminDeleze:
consider fields left empty as 0 instead of NaN (which was throwing exception)


ad-daniel (approved):
Thank you

PR #5731, opened 2023-01-12
Enhancement sphere for appearances

BenjaminDeleze:
For proto web viewer, when the Proto is an `Appearance`/`PBRAppearance` we add a sphere geometry to be able to see the appearance, and a transform such that the sphere is not in the floor.

At first I tought I needed to keep the transform id to replace it on regeneration but finally we alway reuse the same so I think that it is useless.

preview: https://proto.webots.cloud/run?version=R2023b&url=https://github.com/BenjaminDeleze/sandbox/blob/master/protoExample/protos/Grass.proto&type=undefined

### DIFF
--- a/resources/web/wwi/ProtoManager.js
+++ b/resources/web/wwi/ProtoManager.js
@@ -62,8 +62,7 @@ export default class ProtoManager {
     if (this.proto.isRoot && (this.proto.baseType.name === 'PBRAppearance' || this.proto.baseType.name === 'Appearance')) {
       const xml = document.implementation.createDocument('', '', null);
       const transform = xml.createElement('Transform');
-      this.proto.wrapperTransformID = getAnId();
-      transform.setAttribute('id', this.proto.wrapperTransformID);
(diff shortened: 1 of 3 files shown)


resources/web/wwi/ProtoManager.js
```diff
@@ -59,7 +59,27 @@ export default class ProtoManager {
   async loadX3d() {
     let xml = this.getXmlOfMinimalScene();
     const scene = xml.getElementsByTagName('Scene')[0];
-    scene.appendChild(this.proto.toX3d());
+    if (this.proto.isRoot && (this.proto.baseType.name === 'PBRAppearance' || this.proto.baseType.name === 'Appearance')) {
```
ad-daniel, line 62: should we perhaps isolate this into a function the same way  we do for the minimal scene? I believe we said at some point that there would be other exceptions such as these (only hingejointwithbacklash comes to mind atm)

BenjaminDeleze: good idea

PR #5733, opened 2023-01-13
Fix notification after regeneration

ad-daniel:
**Description**

Fixes `colorOverride` parameter not having an effect after regeneration. Specifically, it fixes the `Grass.proto`. The reason the change didn't happen is that the wren objects of the regenerated node were not created (`wrenObjectsCreatedCalled` was always true). Fixing this issue however broke the DEF-USE mechanism for the OfficeChair `legAppearance` (appearance changes no longer propagated to all legs). The issue was that the process of "unfinalizing" the parent (Shape) node to which we're changing the appearance occurred only for the DEF node, and we weren't doing the same for the parent of the USE ones. In short, when finalizing a DEF node we were notifying all USE nodes, however when "unfinalizing" it, we did so only for the DEF and not the USE ones, which resulted in the nodes existing on the webotsjs side but being white

The similar issue in `OfficeChair > BrushedAluminium > colorOverride` has a completely different source, in this case the problem is not the regeneration but rather that a non-existant webotsjs node is being notified, so this fix doesn't solve that. This issue is being handled separately 

https://proto.webots.cloud/run?version=R2023b&url=https://github.com/BenjaminDeleze/sandbox/blob/master/protoExample/protos/Grass.proto&type=undefined

https://proto.webots.cloud/run?version=R2023b&url=https://github.com/BenjaminDeleze/sandbox/blob/master/protoExample/protos/OfficeChair.proto

### DIFF
--- a/resources/web/wwi/X3dScene.js
+++ b/resources/web/wwi/X3dScene.js
@@ -178,9 +178,7 @@ export default class X3dScene {
     let parentNode;
     if (typeof parentId !== 'undefined') {
       parentNode = WbWorld.instance.nodes.get('n' + parentId);
-      parentNode.isPreFinalizedCalled = false;
-      parentNode.wrenObjectsCreatedCalled = false;
-      parentNode.isPostFinalizedCalled = false;
+      parentNode.unfinalize();
     }
 
     const parser = new Parser(webots.currentView.prefix);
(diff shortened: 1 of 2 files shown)


BenjaminDeleze (approved):
Thank you

PR #5735, opened 2023-01-13
Fix transparency in webotsjs

ad-daniel:
**Description**

`ImageTexture` transparency was previously handled on the Webots side, a `isTransparent` field included in the x3d of the `ImageTexture` node. This is however not an option when loading PROTO, resulting in this:

b

The same happens with many logos.

The logic to determine the transparency was already present on the webotsjs side, but wasn't being applied in practice. With this PR:

- Webots no longer sends the `isTransparent` information
- A `checkTransparency` parameter is retained as `highlightX3DElement` explicitly doesn't want any transparency (not sure why)


BenjaminDeleze (approved):
I tested several worlds and it seems good to me!

PR #5736, opened 2023-01-13
Make installation scripts more robust

ad-daniel:
**Description**

Given the recent instabilities with azure, this might help offset some of the random failures.


omichel (approved):
Good idea!

PR #5744, opened 2023-01-16
Use pycodestyle instead of pep8

ad-daniel:
**Description**

According to the [package](https://pypi.org/project/pycodestyle/) info:
> This package used to be called pep8 but was renamed to pycodestyle to reduce confusion.

The change was already done in webots_ros2 and urdf2webots, but not here.

After the change, several other failures were detected, likely because the pep8 packages hasn't been updated since the change

### DIFF
--- a/lib/controller/python/controller/__init__.py
+++ b/lib/controller/python/controller/__init__.py
@@ -49,8 +49,8 @@ from controller.joystick import Joystick                       # noqa
 from controller.motion import Motion                           # noqa
 
 __all__ = [
-    Field, Node, ContactPoint, AnsiCodes, Accelerometer, Altimeter, Brake, Camera, CameraRecognitionObject, Compass, Connector,
-    Display, DistanceSensor, Emitter, GPS, Gyro, InertialUnit, LED, Lidar, LidarPoint, LightSensor, Motor, PositionSensor,
-    Radar, RadarTarget, RangeFinder, Receiver, Robot, Skin, Speaker, Supervisor, TouchSensor, Keyboard, Mouse, MouseState,
-    Joystick, Motion
+    Accelerometer, Altimeter, AnsiCodes, Brake, Camera, CameraRecognitionObject, Compass, Connector, ContactPoint, Display,
+    DistanceSensor, Emitter, Field, GPS, Gyro, InertialUnit, Joystick, Keyboard, LED, Lidar, LidarPoint, LightSensor, Motion,
+    Motor, Mouse, MouseState, Node, PositionSensor, Radar, RadarTarget, RangeFinder, Receiver, Robot, Skin, Speaker,
+    Supervisor, TouchSensor
 ]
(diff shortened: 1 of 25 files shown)


resources/sumo_exporter/crossroad.py
```diff
@@ -24,25 +24,20 @@ def __init__(self, crossroadType):
...
-        except:
-            self.rotation = [0.0, 0.0, 1.0, 0.0]
-        try:
-            self.connectedRoadIDs = [x.replace('"', '') for x in re.findall(r'connectedRoadIDs\s*\[([^\]]*)\]',
-                                                                            wbtString)[0].split()]
-        except:
-            pass
+
+        id = re.findall(r'id\s*"([^"]*)"', wbtString)
+        self.id = id[0] if not id else self.id
```
stefaniapedrazzi, line 29: This seems wrong, shouldn't be
```suggestion
        self.id = id[0] if id else self.id
```

The same wrong conditional is also used in the following lines for `translation`, `rotation`, `connectedRoadIDs`, `roadNumber` and `roadsWidth`.
And in the changes to `resources/sumo_exporter/road.py` file aiming at removing the `try/except` blocks.

stefaniapedrazzi (changes requested):

resources/sumo_exporter/crossroad.py
ad-daniel: yes definitely, got it backwards :cry:

ad-daniel: ops, you're right I got it backwards

stefaniapedrazzi (approved):
Thank you!
It seems that the tests didn't start, but everything seems fine now.

ad-daniel:
when adding the test on the webots_ros2 repo it complained about the issues in the libcontroller. I noticed that we didn't check sources of the libcontroller on the webots side, since we have the new python API we should. I'll make a few tests and put it back to ready for review when done

lib/controller/python/controller/__init__.py
```diff
@@ -47,3 +47,10 @@
 from controller.mouse import MouseState                        # noqa
 from controller.joystick import Joystick                       # noqa
 from controller.motion import Motion                           # noqa
+
+__all__ = [
+    Field, Node, ContactPoint, AnsiCodes, Accelerometer, Altimeter, Brake, Camera, CameraRecognitionObject, Compass, Connector,
+    Display, DistanceSensor, Emitter, GPS, Gyro, InertialUnit, LED, Lidar, LidarPoint, LightSensor, Motor, PositionSensor,
+    Radar, RadarTarget, RangeFinder, Receiver, Robot, Skin, Speaker, Supervisor, TouchSensor, Keyboard, Mouse, MouseState,
+    Joystick, Motion
```
stefaniapedrazzi, line 55: What order did you use to list the classes?
I would rather use the alphabetical order to make this list cleaner.

stefaniapedrazzi (commented):
Everything seems fine.
I just have one comment.

lib/controller/python/controller/__init__.py
ad-daniel: I've used the order as they appear in the imports above

stefaniapedrazzi (approved):
Thank you!

PR #5749, opened 2023-01-18
Enhancement get rid of proto list for webotsjs

BenjaminDeleze:
Remove the use of the proto-list.xml in the web proto viewer to retrieve:
- the documentation
- the nodes that can be inserted in others.

Instead we are using the webots.cloud DB:
- When adding a node to webots.cloud, we retrieve all the necessary information (license, descriptions, base_type, slot_type,...) and we add them in the DB (works also on update)
- When we want to generate a `.md` for the documentation from a proto, we can retrieve the information with one query instead of loading and manipulating the huge `proto-list.xml`.
- When we want to populate the addNodeWindow, we send also just one query that will return only the nodes that can be inserted. The only check that we do on the webotsJS side it the filtering from the `Find:` field. As we need to redo it every time the user types a letter it is better to do it on the js side instead to do multiple queries.
- This system natively supports protos inserted by users.

The drawback is that its add a dependency on webots.cloud. However we would have had it anyway as we had to move the proto-list to webots.cloud.


ad-daniel (changes requested):
Thank you, I found a couple of regressions

- The change breaks slot insertion (tinkerbots), no nodes show up in the list
- For some reason the icons are broken sometimes, when using this PROTO:
```
#VRML_SIM R2023b utf8

EXTERNPROTO "https://raw.githubusercontent.com/cyberbotics/webots/develop/projects/objects/balls/protos/Ball.proto"

PROTO TestProto [
  field MFNode extension [Ball{}]
]
{
  Group {
    children IS extension
  }
}
```
the list is populated correctly but it searches the icons in your sandbox repo
- I believe a couple appearances (GenericTireAppearance, GenericDoorAppearance) and geometries (all) that should appear, no longer do. Can use a PROTO like this to see it:
```
#VRML_SIM R2023b utf8

EXTERNPROTO "https://raw.githubusercontent.com/cyberbotics/webots/develop/projects/objects/balls/protos/Ball.proto"

PROTO TestProto [
  field SFNode app NULL
  fielf SFNode geo NULL
]
{
  Shape {
    appearance IS app
    geometry IS geo
  }
}

```

BenjaminDeleze:
> * The change breaks slot insertion (tinkerbots), no nodes show up in the list

And 

> * I believe a couple appearances (GenericTireAppearance, GenericDoorAppearance) and geometries (all) that should appear, no longer do. Can use a PROTO like this to see it:

It is normal as the tinkerbots nodes, the geometries and so on are not yet on webots.cloud

> * the list is populated correctly but it searches the icons in your sandbox repo

Some protos present on webots.cloud are in my sandbox repo, so it is normal that it tries to look in my sandbox for the corresponding icons.

ad-daniel (approved):

PR #5750, opened 2023-01-18
Cleanup web proto viewer

BenjaminDeleze:
ditto


resources/web/wwi/ProtoManager.js
```diff
@@ -29,7 +29,7 @@ export default class ProtoManager {
       // generate list of exposed parameters (that will be tied to interface elements)
       for (const [parameterName, parameter] of this.proto.parameters) {
         parameter._view = this.#view;
-        this.exposedParameters.set(parameterName, parameter); // TODO: change key to parameter id ?
+        this.exposedParameters.set(parameterName, parameter);
```
ad-daniel, line 32: I might be wrong but I believe we don't use this anymore since we allow to edit sub-proto from the interface, so as long as `this.proto` is set to the root prior to populating the window (might be the case already) this shouldn't be needed

resources/web/wwi/protoVisualizer/Parameter.js
```diff
@@ -211,7 +210,6 @@ export default class Parameter {
         const action = {};
         action['id'] = this.node.id;
         action[this.name] = this.value.toJson();
-        console.log('setPose', action);
         view.x3dScene.applyPose(action);
```
ad-daniel, line 213: I think we should finally rename the apply pose function as we use it for anything now, something like `applyAction` perhaps?

resources/web/wwi/protoVisualizer/Vrml.js
```diff
@@ -381,7 +381,7 @@ export class SFNode extends SingleValue {
 
     // handle exceptions
     if (this.value.name === 'ImageTexture')
-      nodeX3d.setAttribute('role', parameterName.slice(0, -3)); // TODO: rename on the JS side so it matches the field name?
+      nodeX3d.setAttribute('role', parameterName.slice(0, -3));
```
ad-daniel, line 384: Can't we address this? If I remember right, the role of `baseColorMap` is called `baseColor` on the js side. I believe all other roles match the field name. Unless it breaks backwards compatibility or something

resources/web/wwi/test.html
```diff
@@ -1,136 +0,0 @@
-<!DOCTYPE html>
-
-<style>
```
ad-daniel, line 3: perhaps it isn't a good idea, but can't we have a default html included in the repository (but which we don't include in the distro)?
Otherwise every time we want to debug/test something locally we need to create one

ad-daniel (changes requested):
Just a few comments

ad-daniel (approved):
Thank you

PR #5753, opened 2023-01-19
Make the robot window appear by default for competitions

Jean-Eudes-le-retour:
**Description**
This PR make the robot window appear by default for competitions on webots.could "try" page to mimic the behaviour of robotbenchmark.net

### DIFF
--- a/docs/guide/web-streaming.md
+++ b/docs/guide/web-streaming.md
@@ -104,7 +104,6 @@ This is the API of the `webots-streaming` web component:
     * `callback`: the function to be called when a message is received, the text of the message is passed to this function as the only argument.
 * `setWebotsErrorMessageCallback(callback)`: define a function that will be called every time an error is send by Webots.
     * `callback`: the function to be called when an error is received, the text of the error is passed to this function as the only argument.
-* `showRobotWindowOnStartup(name)`: display the desired robot window on simulation startup. The worldInfo robot window is called 'robot'. Must be called before connect. No windows are shown by default.
 
 Moreover, the following attributes are available:
 * `data-server`: URL of the server.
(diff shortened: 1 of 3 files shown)


BenjaminDeleze:
I think it is a good idea however I am not convinced of the approach.

By checking the url we make it dependant to `webots.cloud`. Morehover, if someone want to embed a simulation on its website and as the same variable in its url it will also open the robot window.

I would prefer to have a documented API function in `webotsView.js` to do it.

omichel:
Yes, having a documented API function would be less hacky and also better documented.

Jean-Eudes-le-retour:
Ok, I'll look for a better way of doing it

BenjaminDeleze:
If you add this function as an API one, this is were it should be documented: https://github.com/cyberbotics/webots/blob/master/docs/guide/web-streaming.md

BenjaminDeleze:
Good question,

We could call it `openRobotWindow` or `displayRobotWindow` and maybe rename the `showRobotWindow` into `showRobotWindowButton` to be clearer

Jean-Eudes-le-retour:
But then we would need to append "Button" all other existing attributes and I don't how much this would break existing code?

Jean-Eudes-le-retour:
Maybe we could call it something like `robotWindowOnStartup`?

BenjaminDeleze:
> But then wouldn't we need to append "Button" all other existing attributes? I don't how much this would break existing code

Indeed, if we can avoid that, it may be better.


Just thought about something. How will webots know which robot window to open with an attribute?
It is maybe better to have a function which take the name of the window as parameter no?

Jean-Eudes-le-retour:
> Just thought about something. How will webots know which robot window to open with an attribute?
It is maybe better to have a function which take the name of the window as parameter no?

Good idea, I'll do that

Jean-Eudes-le-retour:
> Just thought about something. How will webots know which robot window to open with an attribute?
It is maybe better to have a function which take the name of the window as parameter no?

I think I found a better solution: we can set an attribute that tells the toolbar to make the main window visible or not

Jean-Eudes-le-retour:
This PR can be tested on: https://testing.webots.cloud/run?version=R2023a&url=https%3A%2F%2Fgithub.com%2Fcyberbotics%2Frobot-programming-competition%2Fblob%2Fmain%2Fworlds%2Frobot_programming.wbt&type=competition&context=try

BenjaminDeleze (approved):
Seems good, thank you

PR #5759, opened 2023-01-20
Turn Pionneer3Gripper into a coupled motor

BenjaminDeleze:
Also fix the corresponding world: since FLU conversion the tie was not perfectly flat and so the ball was rolling.


ad-daniel (approved):
Thank you

PR #5764, opened 2023-01-20
Add a add button for mf field

BenjaminDeleze:
Test it here: https://proto.webots.cloud/run?version=R2023b&url=https://github.com/cyberbotics/webots/blob/feature-web-proto/projects/robots/softbank/nao/protos/Nao.proto&type=undefined


omichel (approved):
Nice, thank you.

Issue #5768, opened 2023-01-22
Contact points (other than the first 10) are silently ignored


brettle:
[Edit: Updated (and inlined) test case to remove some extraneous files and make objects only partially transparent.]

**Describe the Bug**
When 2 geometries collide, only the first 10 contact points discovered are used to create contact joints. Others are silently ignored. This can lead to strange behavior. In particular, if 2 spatially separated regions of a geometry collide, the first 10 contact points found might be in one of those regions and as a result the other region will pass through.

**Steps to Reproduce**
1. Open ContactPointTest.wbt in [ContactPointTest.zip](https://github.com/cyberbotics/webots/files/10473445/ContactPointTest.zip). It contains a traffic cone mesh falling onto a slanted box.
2. Stop and reset the simulation if necessary.
3. Show bounding boxes and contact points if necessary.
4. Step through the simulation until the base of the traffic cone should be colliding with the box.
5. See how there are no contact points shown at the base of the traffic cone but there are many shown at the top of the cone. Note that there are no error or warning messages in the console associated with this behavior.

**Expected behavior**

The base of the traffic cone should also collide with the box. 

At a minimum, a warning message should appear in the console indicating that the maximum allowed number of contact points were found and additional contact points will be ignored. Ideally, all contact points would be used but I realize that is not practical. See below for ideas on strategies that ensure that the base of the traffic cone collides with the box even if not all contact points are used.

**Screenshot**
![ContactPointTest](https://user-images.githubusercontent.com/118192/213894716-72d17f0d-edc2-4b06-94f8-c9113f99e67d.png)

The problem code is [here](https://github.com/cyberbotics/webots/blob/52225cf830de223477da74dc35e0fe17fba48cf2/src/webots/engine/WbSimulationCluster.cpp#L631-L632).

[...]

PR #5769, opened 2023-01-22
Make contact point handling configurable (was: Stop silently ignoring more than 10 contacts)

brettle:
**[Edit: With the merger of PR #5792 into master, this PR now just makes the number of contact points and contact joints configurable via `ContactProperties`. It originally just increased the number of contact points to 100 and warned when that limit was met but it has evolved since then. See conversation below for full details.]**

Look for up to 100 contacts.
Emit a warning if there may be more than 100 contacts.
Create contact joints for the deepest 10 contacts. 
Emit a warning if there are contacts without associated joints. 
Add an associated test.

**Related Issues**
This pull-request addresses issue #5768.

**Additional context**
See issue #5768 for full context. This PR makes it less likely that the contacts will be ignored but it is still possible that with a very detailed mesh there might be more than 100 contacts. The effects of that should be greatly reduced with the mesh triangle shuffling I'm hoping to submit as a separate PR.


stefaniapedrazzi (changes requested):
Thank you for your contribution!
There are some details to adjust to make make it compliant and consistent with this repo, but for the moment the changes are fine.
If as suggested in https://github.com/cyberbotics/webots/issues/5768, you plan to add additional changes in this PR I will fully test it once completed so that we can also evaluate if there is some performance drop.

About the test:
* the STL file should be renamed "cone.stl" and moved in the new folder `tests/physics/worlds/meshes`

stefaniapedrazzi:
I tested if the changes in this PR are affecting the simulation performance, but it doesn't seem there is any significant performance drop.

Seems fine to arbitrary set the contact count to 100.
If for some simulations this value should not be sufficient, then we can modify it at that point and maybe evaluate if it is not better to let specify this value in the world to not penalize other simulations.

Given that this solves the issues you are experiencing, I would wait to add additional shuffle mechanism until there are some real use cases where the current implementation is not sufficient and finalize this PR as it.

brettle:
Thanks for the feedback. In issue #5768, you recommended targeting develop instead of master since the change could affect existing behavior. If I make `MAX_CONTACTS` and `MAX_CONTACT_JOINTS` configurable in `WorldInfo` (as say `maxContacts` and `maxContactJoints`) and have both values default to 10 for now, that would preserve the existing behavior and provide a way to workaround the issue. Would that make this acceptable for master? In develop, the default for `maxContacts` could be increased to make it unlikely that the issue would be encountered in the first place.

brettle:
I think I've addressed your suggestions and I went ahead and added the fields to `WorldInfo` in hopes of getting this merged into master.

Full disclosure: Although the tests are all passing in CI, I'm sometimes (but not always) seeing the following failure locally:
```
FAILURE with rolling_friction_supervisor: BALL_8 position different from expected one.
```

I haven't been able to find a cause yet, but:

1. The problem doesn't seem to occur on master, and
2. I've confirmed that the `maxContacts` and `maxContactJoints` fields are both set to the default value of 10 at the start of the test, as expected.

brettle:
My latest commit fixes the aforementioned failing test.

omichel:
Thank you @brettle for this very nice contribution!

However, I am not sure the `WorldInfo` is the best place to set the `maxContacts` and `maxContactJoints` parameters. Instead, I would set them in the [ContactProperty](https://www.cyberbotics.com/doc/reference/contactproperties) node, which is more modular and would allow a finer tuning of the contact management. Also, I would rename `maxContacts` to `maxContactPoints` which sounds more accurate to me.

stefaniapedrazzi:
Additionally, changes to nodes should be merged to the develop branch and not the master branch because of compatibility issues. So as is this branch cannot be merged to master either.

docs/reference/worldinfo.md
```diff
@@ -136,3 +139,5 @@ The value of the seed should be either non-negative or -1, if the value is -1 a
 Using a time-based seed makes simulations non-reproducible.
 
 - The `contactProperties` field allows to specifiy a number of [ContactProperties](contactproperties.md) nodes that define the behavior when [Solid](solid.md) nodes collide.
+
+- The `maxContacts` and `maxContactJoints` fields control the generation of contact joints during a collision. At most `maxContacts` contact points are detected (the rest are ignored). Of those, contact joints will only be created for the (at most) `maxContactJoints` contact points that are the deepest. If it seems like some points of contact are being ignored, consider greatly increasing `maxContacts`. This will likely have little impact on performance. By contrast changes to `maxContactJoints` may have a significant effect on performance because the computational complexity of the default ODE physics engine scales with the cube of the number of contact joints.
```
stefaniapedrazzi, line 143: In the documentation, each sentence should be written on a new line.

resources/nodes/WorldInfo.wrl
```diff
@@ -25,6 +25,8 @@ WorldInfo {
   field       SFFloat                        dragTorqueScale                5.0        # order of magnitude of the torque applied to an object in the interface
   field       SFInt32                        randomSeed                     0          # defines the seed used internally by the random number generator
   field       MFNode                         contactProperties              []         # contact properties
+  field       SFInt32                        maxContacts                    10         # max number of contacts to consider in a collision
+  field       SFInt32                        maxContactJoints               10         # max number of contact joints to create in a collision
```
stefaniapedrazzi, line 29: In the documentation we try to avoid abbreviations
```suggestion
  field       SFInt32                        maxContacts                    10         # maximum number of contacts to consider in a collision
  field       SFInt32                        maxContactJoints               10         # maximum number of contact joints to create in a collision
```

brettle:
Thanks for the feedback. I've changed this PR to go against the develop branch and will continue working on it. I've also created an API-compatible PR #5776 against master in hopes of getting a hotfix for issue #5768 into the next patch release.

tests/physics/worlds/contact_points.wbt
```diff
@@ -0,0 +1,130 @@
+#VRML_SIM R2023a utf8
```
omichel, line 1: Fixing the test failure:
```suggestion
#VRML_SIM R2023b utf8
```

brettle:
How would you feel about making the default that *all* contact points are considered (i.e. `maxContactPoints` = infinity, probably specified with `maxContactPoints` = -1)? That should make existing simulations with detailed meshes more accurate and reduce the chances of users being surprised by collisions not happening. I don't know if it would slow things down to an appreciable degree, but if it does, they should probably use less detailed meshes anyway. And if they can't do that for some reason, they can set `maxContactPoints` to some smaller value.

I'd implement this as follow:

1. Whenever dCollide()` runs out of space in the contacts array, increase the the size of the array (by say 10x) and re-run `dCollide()`. 
2. To minimize the amount of re-colliding needed over the course of the simulation, I would move the contacts array to be a member of `WbSimulationCluster` so I could use it's most recent size when calling `dCollide()`. On a related note, can I assume that at most one thread is associated with an instance of `WbSimulationCluster`. The comment at the top of `WbSimulationCluster.hpp` describes the class as "single-threaded simulation" but I thought I'd double check.

Thoughts?

omichel:
Your proposal sounds very good to me. As yes, I believe only one thread is associated with an instance of `WbSimulationCluster`.

stefaniapedrazzi:
Yes, seems OK to set the default maximum number of contacts to infinity.
But before merging, we then go through all the distributed example and adjust the when needed.

stefaniapedrazzi (commented):
I just ported the latest changes from master to develop branch.
So that this branch can be synchronized with `cyberbotics:develop`.

brettle:
Some performance testing I'm doing indicates that the string processing associated with generating the warning messages when there are ignored contact points or more contact points than contact joints can slow down the physics processing by large factors when there are a lot of collisions (e.g. the collision_multiple_trimesh test). This leads to 2 questions:

1. Do you have an established way to avoid such a problem? I have some ideas but would rather not reinvent the wheel.
2. This will be an issue in the related hotfix PR #5776 which was recently merged into master. Should I reopen that PR or create a new one to address this in master?

omichel:
> Some performance testing I'm doing indicates that the string processing associated with generating the warning messages when there are ignored contact points or more contact points than contact joints can slow down the physics processing by large factors when there are a lot of collisions (e.g. the collision_multiple_trimesh test). This leads to 2 questions:
>
> 1. Do you have an established way to avoid such a problem? I have some ideas but would rather not reinvent the wheel.

Yes, we should actually print this kind of message only once, and mention that it may occur again but won't be printed again. If you have a better idea, however, feel free to discuss it here.

> 2. This will be an issue in the related hotfix PR #5776 which was recently merged into master. Should I reopen that PR or create a new one to address this in master?

You should create a new one targeting the master branch.

brettle:
> Yes, we should actually print this kind of message only once, and mention that it may occur again but won't be printed again. If you have a better idea, however, feel free to discuss it here.

I'm assuming once per timestep is acceptable since that's what happens when the ODE LCP solver doesn't converge. Even once per timestep is a little tricky (but possible) to do in a way that doesn't break ABI-compatibility with `WbSimulationCluster`. Is that class considered part of the public ABI which which we want to maintain compatibility in master?

> You should create a new one targeting the master branch.
Will do.

omichel:
We can afford to break the ABI in Webots on the master branch as long we don't break the libController ABI, this is OK.
Printing once per time step may still have some significant impact on the performance. You should evaluate it before going ahead.
Another alternative would be print once it every virtual second if the problem has occurred in the past second.

brettle:
If PR #5792 is merged into master, I'll update this PR to only be about making `maxContactPoints` and `maxContactJoints` configurable via `ContactProperties`.

src/webots/engine/WbSimulationCluster.cpp
```diff
@@ -653,21 +668,29 @@ void WbSimulationCluster::odeNearCallback(void *data, dGeomID o1, dGeomID o2) {
...
+  if (findAllContactPoints)
+    maxContactPoints = std::max<size_t>(100, cl->contactVector.capacity());
 
   int n;
   dContact *contact;
   while (true) {
-    contactVector.reserve(maxContactPoints);
-    contact = contactVector.data();
+    cl->contactVector.reserve(maxContactPoints);
+    contact = cl->contactVector.data();
```
brettle, line 684:
```suggestion
    contactVector.reserve(maxContactPoints);
    contact = contactVector.data();
```

brettle (commented):
Make suggested fix.

brettle:
Do we still want to make `maxContactPoints` and `maxContactJoints` configurable? I can see how allowing `maxContactJoints` to be changed enables trading accuracy for performance, but I'm not sure setting `maxContactPoints` to anything other than -1 (i.e. all) is ever the best choice. Wouldn't it always be better for the user to use a simpler model than limit contact points, since limiting contact points can lead to such unexpected behavior?

omichel:
Yes, I agree. I believe we could get rid of the `maxContactPoints` parameter (and consider it is always unlimited) and expose only the `maxContactJoints` parameter which makes more sense in term of tuning the performance versus accuracy. Nevertheless, I believe it would be useful to warn the user if a large number of contact points is found as this would certainly be a symptom of a poorly designed model which could affect the performance.

brettle:
I'll remove `maxContactPoints`, but how many contact points is enough to warn about? Note that there will already be a warning if there are more than `maxContactJoints` (default 10) contacts.

omichel:
> I'll remove `maxContactPoints`, but how many contact points is enough to warn about? Note that there will already be a warning if there are more than `maxContactJoints` (default 10) contacts.

OK, then a single warning should be sufficient.

brettle:
Is there anything else that needs to be addressed to get this approved?

docs/reference/changelog-r2023.md
```diff
@@ -9,6 +9,7 @@ Released on ??
   - Enhancements
     - Improved the robots' grippers such that they use coupled motors ([#5694](https://github.com/cyberbotics/webots/pull/5694)).
     - Improved the [Charger](charger.md) behavior ([#5771](https://github.com/cyberbotics/webots/pull/5771)).
+    - Added`maxContactJoints` field to `ContactProperties` node to give users control over the tradeoff between collision accuracy and performance ([#5769](https://github.com/cyberbotics/webots/pull/5769)).
```
omichel, line 12:
```suggestion
    - Added `maxContactJoints` field to `ContactProperties` node to give users control over the tradeoff between collision accuracy and performance ([#5769](https://github.com/cyberbotics/webots/pull/5769)).
```

src/webots/engine/WbSimulationCluster.hpp
```diff
@@ -24,6 +24,8 @@
 #include <QtCore/QList>
 #include <QtCore/QMutex>
 
+#include <vector>
+
```
omichel, line 28: This doesn't seem to be used.
```suggestion
```

omichel (changes requested):
Why did you change webots_ros? It doesn't seem related with this PR.

brettle:
> Why did you change webots_ros? It doesn't seem related with this PR.

I think that was an unintended side effect of retargeting this PR from master to develop. Should be fixed now. Thanks for pointing it out.

omichel (approved):
All good for me. Thank you.

Issue #5765, opened 2023-01-20
"Battery.wbt" charger discharges to 0 when a robot leaves the charger radius


AghiadRaslan:
When a robot enters the radius of the charger, it starts charging. The robot's battery level starts to increase, and the charger's battery level starts to decrease. However, the robot might have consumed only 10% of the charger's battery, once it leaves the radius of the charger, the charger's battery drops to 0 immediately, then it starts to reproduce energy as it is supposed to do.

I have tried to decrease the charging rate of the robot to get a better diagnostic of the issue, and the bug is still there.   

**Steps to Reproduce**
1. Open to 'Battery.wbt' sample world.
2. Adjust the battery's charging rate of the "MyBot" robot to get a clearer look at the issue.
3. Open the TreeView and expand 'Charger "charger0" '.
4. Scroll down to "Battery" field and expand it.
5. Run the simulation and observe the first value of the charger's battery when the robot enters and leaves the charger's radius.

**Expected behavior**
The charger's battery decreases by the amount the robot consumed, and when the robot leaves the radius, it starts to recharge itself starting from the remaining capacity,

**Screenshots**
![Screenshot from 2023-01-20 19-06-38](https://user-images.githubusercontent.com/93394103/213775361-ad088755-967f-41a2-804c-6ce1b8be7263.png)

![Screenshot from 2023-01-20 19-09-20](https://user-images.githubusercontent.com/93394103/213775421-612e87a8-db87-44eb-99a0-bd941481c5a4.png)
![Screenshot from 2023-01-20 19-09-51](https://user-images.githubusercontent.com/93394103/213775554-5d7bcea7-bea5-4c25-8a3f-34005b7a961b.png)

**System**
 - Operating System: [Linux Ubuntu 22.04.1 LTS
 - Graphics Card: [NVIDIA GeForce GTX 1060 Mobile

stefaniapedrazzi:
This is fixed in #5771 alongside with some other improvements to the Charger device.
The fix will be available in the nightly build of Webots R2023b starting from tomorrow.

PR #5771, opened 2023-01-23
Improve Charger behavior

stefaniapedrazzi:
Addresses #5765:
continues with the Charger behavior refactor (see also  #2879) and improve the discharging behavior.
Changes:
* [x] remove reset of Changer battery level to 0 when a robot disconnects
* [x] remove transfer of energy from Changer to Robot if Robot battery is fully loaded
* [x] continue to transfer energy from Charger to Robot even after Robot was fully loaded
* [x] allow the Charger to connect to a Robot for transferring energy even if the Charger is not fully loaded


omichel (approved):
Thank you.

PR #5772, opened 2023-01-23
Fix floating window resize

### LINKED ISSUE #5761: Web: Floating window resize with small window
**Describe the Bug**
When the webotsView is relatively small and centered in the page, the floating windows (robot window, proto parameter,...) do not resize correctly.

If you try to drag them to change their size (especially when trying to increase the left size) the size will "jump" and increase linearly as it should.

Particularly visible on Firefox, but can also be reproduce in chrome.

BenjaminDeleze:
The problem was that we were calculating the remaining width relatively to the size of the 3d window but we were getting the absolute position of the mouse.

Fix #5761 

Note: we cannot use `document.getElementsByClassName('webots-view')[0].offsetLeft` to solve the issue because if the `webots-view` element is included in another one which is not the full width of the screen we will have the same problem.


omichel (approved):
Nice, thank you.

PR #5774, opened 2023-01-24
Fix retrieving cached PROTO internal fields in Supervisor API

### LINKED ISSUE #5757: calling wb_supervisor_node_get_field with proto internal field causes performance degradation
**Describe the Bug**
The problem lies in `webots/src/controller/c/supervisor.c`
When calling `wb_supervisor_node_get_field` (https://github.com/cyberbotics/webots/blob/master/src/controller/c/supervisor.c#L2367) with a proto internal node as `WbNodeRef` the field list is searched (https://github.com/cyberbotics/webots/blob/master/src/controller/c/supervisor.c#L2384) with the parameter `is_proto_internal` set to false by default.
As the node is proto internal, this will always fail.
Therefore, the field is newly requested from the simulator and the first element of the field_list is returned.

However, this causes the field list to grow.
If this is done often, the field_list grows to an unmanageable size causing significant performance degradation.

This function may be called often, for example when getting contact points using `wb_supervisor_node_get_contact_points` with include_descendants = true, then getting the `WbNodeRef` using  `wb_supervisor_node_get_from_id` using the node_id in he contact point struct (https://cyberbotics.com/doc/reference/supervisor#contact-point) to get the `WbNodeRef`. When calling `wb_supervisor_node_get_field` on this reference (for example to get the name of the solid which made contact), the above described behavior occurs.

**Expected behavior**
performance should not degrade 

**System**
 - Operating System: Linux Ubuntu 22.04
 - should affect all systems

**Additional context**
I found a workaround which is implemented which is implemented here: https://github.com/RoboCup-Humanoid-TC/hlvs_webots/pull/7

stefaniapedrazzi:
Fix #5757: add a new flag to `WbFieldStruct` to mark if the field is a PROTO field or parameter.

In the Supervisor API the the `WbFieldStruct.is_proto_internal` flag was used to determine if the cached field was read only (i.e. not visible from the scene tree).
Then in #4363, it was added a check was added to make sure that if a PROTO field and parameter with the same name exist the correct one is returned. This check is fine and is needed to retrieve the correct field.
However, the value contained in the former `WbFieldStruct.is_proto_internal` (renamed `WbFieldStruct.is_read_only`) cannot be used to determine the field is a PROTO field or parameter.

### DIFF
--- a/docs/reference/changelog-r2023.md
+++ b/docs/reference/changelog-r2023.md
@@ -27,6 +27,7 @@ Released on ??
     - Fixed crashes resulting from converting to base nodes a PROTO containing DEF and USE nodes ([#5676](https://github.com/cyberbotics/webots/pull/5676)).
     - Fixed crashes resulting from updating a DEF node whose USE node is contained in a PROTO field triggering the regeneration ([#5676](https://github.com/cyberbotics/webots/pull/5676)).
     - Fixed crashes resulting from copy or DEF/USE update in [`Solid.boundingObject`](solid.md) field ([#5686](https://github.com/cyberbotics/webots/pull/5686)).
+    - Fixed performance issues when retrieving multiple times a node's field defined in a PROTO body using the [Supervisor API](supervisor.md) ([#5774](https://github.com/cyberbotics/webots/pull/5774)).
 
 ## Webots R2023a
 Released on November 29th, 2022.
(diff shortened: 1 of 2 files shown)


BenjaminDeleze (approved):
Seems good to me

PR #5787, opened 2023-01-26
Web Proto: revert change

BenjaminDeleze:
**Description**
This PR implements a feature discussed some time ago: when someone reset a field, the reset button is change to a revert one (instead of being disabled).
When clicking on it, it will revert to latest value set by the user.
If the user enters a new value after the reset, the revert button is replaced by the reset button.
The feature is available for all fields except for SFBool, SFNode and MFNode

You can test it here: https://proto.webots.cloud/run?version=R2023b&url=https://github.com/cyberbotics/webots/blob/feature-web-proto/projects/robots/softbank/nao/protos/Nao.proto&type=undefined

### DIFF
--- a/resources/web/wwi/FloatingProtoParameterWindow.js
+++ b/resources/web/wwi/FloatingProtoParameterWindow.js
@@ -465,7 +465,6 @@ export default class FloatingProtoParameterWindow extends FloatingWindow {
     const hideShowButton = this.#createHideShowButtom(currentMfId, addButton);
 
     const resetButton = this.#createResetButton(parent, p.style.gridRow, parameter.name);
-    console.log(p)
     resetButton.onclick = () => {
       const nodesToRemove = document.getElementsByClassName('mf-id-' + currentMfId);
       let maxRowNumber = 0;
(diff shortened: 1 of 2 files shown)


ad-daniel:
Personally I find it a bit non-intuitive to be honest, particularly the fact that after changing a value I'm unable to get back to the "grey out" state the panel starts with (unless I'm missing something). The button tooltip should probably show something different as well

omichel:
Yes, I agree with @ad-daniel that it is a bit confusing...
If I understand well, basically, we have two possible actions:

- reset to default value
- undo this reset and revert to the edited value

Although, this could be useful, it is not very standard and will probably confuse most users.
I am not sure if we should go ahead with this...

BenjaminDeleze:
> If I understand well, basically, we have two possible actions:
> 
> * reset to default value
> * undo this reset and revert to the edited value

Yes exactly.

I agree that it is uncommon and for me we can drop it. 
I implemented it because you told me (maybe with Yannick) that it would be a nice feature to have when we first discussed the reset button before Christmas.

omichel:
OK... I am sorry, this was probably not a very good idea...

PR #5789, opened 2023-01-26
Feature robot window on startup for stream

Jean-Eudes-le-retour:
**Description**
This PR makes the robot window visible on the streaming client on startup if they have been set to open in the wbproj file.

This would make the web streaming experience more coherent with the local Webots experience. This would be an easier way than #5753 to set robot windows to open on webots.cloud and would make it obsolete. This would also be more flexible, as it would allow people to submit simulations in webots.cloud with robot windows opened by default.

It can be tested locally by calling webots with the --stream argument, opening a robotbenchmark sample world and going to http://localhost:1234/index.html to see the robot windows being visible.

**Tasks**

- [x] replace "../resources/web/wwi" with https://cyberbotics.com/wwi/R2023b before merge


resources/web/wwi/Toolbar.js
```diff
@@ -553,8 +555,6 @@ export default class Toolbar {
       if (rw.isMainWindow) {
         rw.setSize(2 * robotWindowWidth, 2 * robotWindowHeight);
         rw.setPosition(margin, margin);
-        if (this.parentNode?.openMainRobotWindow === true)
-          rw.changeVisibility();
```
BenjaminDeleze, line 557: You removed what was introduced in #5753 right? In this case, we should also remove the corresponding documentation

BenjaminDeleze (changes requested):
Seems good, just a question:

resources/web/wwi/Toolbar.js
Jean-Eudes-le-retour: Oh yeah, right! I'll do that

Jean-Eudes-le-retour:
I left some comments in the code for me to understand how some part of the code worked, should I leave them or remove them ?

BenjaminDeleze (approved):
Thank you
Comments are fine for me

PR #5776, opened 2023-01-24
Hotfix to use more contacts

brettle:
**Description**
API-compatible hotfix to issue where only the first 10 contact points are used and others are silently ignored.

**Related Issues**
This pull-request fixes issue #5768.

**Additional context**
A more complete fix which involves API changes is in PR #5769.

PR #5790, opened 2023-01-26
Speed up contact processing.

brettle:
Rate limit warnings.
Use std::nth_element() instead of std::sort().

**Description**
PR #5776 caused excessive warnings when there were more than 10 contact points and those warnings greatly slowed down physics . This PR rate limits those warnings and also adds a small optimization to the algorithm used to find the deepest contact points.

The remaining increase in physics time is overwhelmingly due to the time required for ODE to find the additional contact points. [Edit: See next comment for updated performance data.] ~A particularly challenging test is `collision_multiple_trimesh.wbt`. With this PR, the physics time for that test only increases by approximately 2.3x when using 100 (or even 1000) vs 10 for MAX_CONTACTS. On my machine, the speed factor goes from 22x to 11x.~

### DIFF
--- a/src/webots/engine/WbSimulationCluster.cpp
+++ b/src/webots/engine/WbSimulationCluster.cpp
@@ -668,11 +668,13 @@ void WbSimulationCluster::odeNearCallback(void *data, dGeomID o1, dGeomID o2) {
   if (!p1 || !p2) {
     WbRobot *const robot1 = dynamic_cast<WbRobot *>(s1);
     WbRobot *const robot2 = dynamic_cast<WbRobot *>(s2);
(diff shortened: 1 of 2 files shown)


brettle:
Correction: Due to a bug I was only rate-limiting 1 of the 2 warnings. With that now fixed, the physics time for `collision_multiple_trimesh.wbt` only increases by approximately 1.2x and my speed factor goes from 26x to 22x, and those differences might just be measurement error.

brettle:
Moving this to draft, because I'd personally prefer to see PR #5792 merged instead.

brettle:
Closing because PR #5792 has been merged.

PR #5791, opened 2023-01-27
Hotfix to find all contact points quickly

brettle:
The commit history for this PR is suspect, so I've replaced it with PR #5792.


brettle:
Hmm... Looks like my cherry picking from my other PR branch has messed up the commit history in a scary way...
I'm going to try again.

PR #5792, opened 2023-01-27
Hotfix to find all contacts quickly

brettle:
**Description**

This builds on PR #5790 by finding *all* of the contact points instead of just the first 100. It uses the same technique as PR #5769 to ensure that all contact points are found, but doesn't make the ABI-incompatible changes in that PR that are needed to add `maxContactPoints` and `maxContactJoints` to ContactProperties. Instead it hard-codes those to -1 (i.e. all) and 10, respectively.

**Additional context**
If this PR is merged, PR #5790 can be closed and PR #5769 can be easily changed to just add the new fields to `ContactProperties`.

### DIFF
--- a/src/webots/engine/WbSimulationCluster.cpp
+++ b/src/webots/engine/WbSimulationCluster.cpp
@@ -686,11 +686,13 @@ void WbSimulationCluster::odeNearCallback(void *data, dGeomID o1, dGeomID o2) {
   if (!p1 || !p2) {
     WbRobot *const robot1 = dynamic_cast<WbRobot *>(s1);
     WbRobot *const robot2 = dynamic_cast<WbRobot *>(s2);
+    // Ensure that the deepest contact point is first for calls to collideKinematicRobots() below.
+    std::nth_element(contact, contact, contact + n,
+                     [](const dContact &c1, const dContact &c2) { return (c1.geom.depth > c2.geom.depth); });
     if (robot1 && !p1 && !isRayGeom2 && robot1->kinematicDifferentialWheels()) {
       wg1->setColliding();
       cl->mCollisionedRobots.append(robot1->kinematicDifferentialWheels());
       if (robot2 && !p2 && robot2->kinematicDifferentialWheels()) {
-        dCollide(o1, o2, MAX_CONTACTS, &contact[0].geom, sizeof(dContact));  // we want only one contact point
         wg2->setColliding();
         cl->mCollisionedRobots.append(robot2->kinematicDifferentialWheels());
         collideKinematicRobots(robot1->kinematicDifferentialWheels(), true, contact, true);
@@ -700,7 +702,6 @@ void WbSimulationCluster::odeNearCallback(void *data, dGeomID o1, dGeomID o2) {
       return;
     }
     if (robot2 && !p2 && !isRayGeom1 && robot2->kinematicDifferentialWheels()) {
-      dCollide(o1, o2, MAX_CONTACTS, &contact[0].geom, sizeof(dContact));
       wg2->setColliding();
       cl->mCollisionedRobots.append(robot2->kinematicDifferentialWheels());
       collideKinematicRobots(robot2->kinematicDifferentialWheels(), false, contact, false);
--- a/src/webots/engine/WbSimulationCluster.cpp
+++ b/src/webots/engine/WbSimulationCluster.cpp
@@ -466,18 +466,12 @@ void WbSimulationCluster::handleCollisionIfSpace(void *data, dGeomID o1, dGeomID
 
 void WbSimulationCluster::warnMoreContactPointsThanContactJoints() {
   static std::mutex mutex;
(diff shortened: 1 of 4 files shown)


omichel (changes requested):
I only have a request regarding the warning method.

omichel:
[test-worlds](https://github.com/cyberbotics/webots/actions/runs/4022790233/jobs/6913342026) is failing. I believe we should fix the two world files [summit_xl_steel.wbt](https://github.com/cyberbotics/webots/blob/master/projects/robots/robotnik/summit_xl_steel/worlds/summit_xl_steel.wbt) and [fabtino.wbt](https://github.com/cyberbotics/webots/blob/master/projects/robots/rec/fabtino/worlds/fabtino.wbt) by using simpler bounding objects (e.g. primitives) so that no warning is generated and the [worlds test](https://github.com/cyberbotics/webots/blob/master/tests/test_worlds.py) passes.

omichel:
I believe I fixed the worlds test by using a cardboard box with a simple bounding object instead of a plastic crate which has a problematic bounding object for this kind of use.

brettle:
`WbSimulationCluster::odeNearCallback()` is called from ODE. Are you confident that when ODE is running with multiple threads, only one thread will be in `odeNearCallback` at a time? Or will each thread have a pointer to a different `WbSimulationCluster` instance?

brettle:
Also, with your recent changes to [summit_xl_steel.wbt](https://github.com/cyberbotics/webots/blob/master/projects/robots/robotnik/summit_xl_steel/worlds/summit_xl_steel.wbt)[summit_xl_steel.wbt](https://github.com/cyberbotics/webots/blob/master/projects/robots/robotnik/summit_xl_steel/worlds/summit_xl_steel.wbt), I'm now getting:
`WARNING: CardboardBox (PROTO): 'name' field value should be unique: 'cardboard box' already used by a sibling Solid node.`

omichel:
I just fixed the solid name conflicts.

omichel:
> `WbSimulationCluster::odeNearCallback()` is called from ODE. Are you confident that when ODE is running with multiple threads, only one thread will be in `odeNearCallback` at a time? Or will each thread have a pointer to a different `WbSimulationCluster` instance?

Yes, I fear that `odeNearCallback` could be called from different threads. So, it should probably be re-entrant.

brettle:
Any problem with using [`thread_local`](https://en.cppreference.com/w/cpp/language/storage_duration) so that each thread has it's own `mContactVector`? It doesn't look like you are using `thread_local` anywhere else in the code, so thought I'd check.

brettle:
I've assumed thread_local is acceptable. Note that I've only ensured thread-safety for the changes associated with this PR. I suspect `odeNearCallback()` wasn't thread-safe to begin with, but fixing that presumably belongs in a separate PR.

src/webots/engine/WbSimulationCluster.hpp
```diff
@@ -23,6 +23,9 @@
 #include <ode/ode.h>
 #include <QtCore/QList>
 
+#include "WbLog.hpp"
+#include "WbSimulationState.hpp"
+
```
omichel, line 28: These include statements are now useless:
```suggestion
```

src/webots/engine/WbSimulationCluster.cpp
```diff
@@ -467,6 +464,17 @@ void WbSimulationCluster::handleCollisionIfSpace(void *data, dGeomID o1, dGeomID
   }
 }
 
+void WbSimulationCluster::warnMoreContactPointsThanContactJoints() {
+  static std::mutex mutex;
+  std::lock_guard<std::mutex> lock(mutex);
```
omichel, line 469: In #5796 and in the rest of the Webots code, we are using the Qt mutexes (QMutex). Is there a particular reason for using std::mutex here?

src/webots/engine/WbSimulationCluster.cpp
```diff
@@ -638,20 +646,31 @@ void WbSimulationCluster::odeNearCallback(void *data, dGeomID o1, dGeomID o2) {
     }
   }
 
-  dContact contact[MAX_CONTACTS];
-  int n = dCollide(o1, o2, MAX_CONTACTS, &contact[0].geom, sizeof(dContact));
+  const int maxContactJoints = MAX_CONTACT_JOINTS;
+  thread_local std::vector<dContact> contactVector;
```
omichel, line 650: Similarly, we tend to use the Qt interface to std::vector, e.g., QVector elsewhere in Webots.
Do you have any reason for not doing the same here?

omichel (changes requested):

src/webots/engine/WbSimulationCluster.cpp
brettle: I just hadn't done any Qt programming before so was going with what I knew. I've switched to the Qt wrappers.

omichel (approved):
Looks good to me.
Thank you.

PR #5794, opened 2023-01-27
Rework node building logic, remove node cloning & fix several issues

ad-daniel:
**Description**

This PR:
- Reworks the creation of PROTO nodes
   - Decouples the creation of the nodes from the creation of the proto models. The `cProtoModels` structure is populated with the information necessary to build the corresponding node (similar to what already occurs for basenodes where this information is hard-coded in FieldModel.js)
      - default values of PROTO parameters are stored in the structure as instances of a Tokenizer. This is unavoidable as it's the only "simple" way of being able to create a node instance from complex nodes like, for example, if the default value looks something like:
    `field SFNode legAppearance  BrushedAluminium { textureTransform TextureTransform { scale 5 5 } }`
    In theory it could've been possible to translate this into a JS object or something, but since we already have the infrastructure to configure a node from tokenizer, this is the most painless approach (which brings enough pain of its own), and spares us yet another encoder/decoder (which we have plenty of as it is)
   - A distinction is made between fields & parameters. PROTO nodes have parameters and a BaseType (which points to either another node), base nodes only have fields.
   - Removes the cloning of nodes. When a new instance needs to be created (for example, multiple IS connected to the same parameter), the parameter is referenced directly every time, which greatly simplifies the resulting structure.
     - One consequence is that the assignment of ids becomes tricky since webotsjs expects multiple instances of the same node to be unique.
- Fixes issues:
   - Changing of fields in MFNodes not having an effect (unknown nodes were being notified, consequence of countless clones happening in complex PROTO, the number of nodes exploded exponentially since cloning a node means cloning its parameters and parameters can be nodes so... you get the idea... in short, way too many clones of the same node existed where in practice we only cared about those belonging to the root PROTO -> basenode tree branch)
   - Some parameters having no effect while others in the same list work
   - Inserting a new node in a MFNode caused existing nodes not to work anymore.
   - Logic problems which resulted in the template engine encoding the parameters prior to these being overwritten by a tokenizer, effectively using the wrong value to generate the VRML

Voluntary regression:
- In this PR the parsing of DEF-USE takes place, but when writing the x3d it doesn't write the USE nodes as USE. I've quickly tried to restore this functionality, but it currently doesn't play nice with the rest and might require additional work. It might not be hard to fix, but in practice I don't believe it's that critical at the moment. The reason is that even though the resulting x3d would be shorter, webotsjs ends up creating a separate node instance anyway so performance-wise, I don't think we really gain much. 
- - 
https://proto.webots.cloud/run?version=R2023b&url=https://github.com/cyberbotics/webots/blob/feature-web-proto/projects/robots/softbank/nao/protos/Nao.proto&type=undefined

https://proto.webots.cloud/run?version=R2023a&url=https://github.com/BenjaminDeleze/sandbox/blob/master/protoExample/protos/OfficeChair.proto&type=undefined

Bonus:
- Cabinet works now (I tried adding it to webots.cloud but can't seem to, but it works locally)


BenjaminDeleze (changes requested):
Seems very good to me just remarks:
- The reset of some (mf) fields is broken: MfColor and MfString for example
- In `OfficeChair` => `BrushedAluminium` => `TextureTransform`, no field of the TextureTransform are modifiable. Is it normal?

ad-daniel:
>     * The reset of some (mf) fields is broken: MfColor and MfString for example

Yes indeed, forgot to remove the clone calls for non-sfnode parameters and adapt those. Should be good now.

>     * In `OfficeChair` => `BrushedAluminium` => `TextureTransform`, no field of the TextureTransform are modifiable. Is it normal?

That's because it's a base-node and not a proto. Basenodes have fields but no parameters, so the list is of parameters to show is empty. For now I've removed the "configure" button if the value of a parameter is a base node. As we add ways of interacting with base nodes on top of PROTO, we can re-enable it on a case by case basis

BenjaminDeleze (approved):
> That's because it's a base-node and not a proto. Basenodes have fields but no parameters, so the list is of parameters to show is empty.

Thanks for the explanation. However this basenode is in a parameter so for me I think we should be able to modify it.
But I guess this can be done later

PR #5795, opened 2023-01-27
Fix north in worlds imported from OpenStreetMap

fparat:
**Description**
A world imported from an OpenStreetMap file has the axis Y to the east and X to the south.
This patch fixes the importer script to have Y to the north and X to the east

Example for this random location: https://www.openstreetmap.org/export#map=15/48.0048/1.6344

Before:
tournoisis_world

After:
tournoisis_world_fixed

**Related Issues**
N.A.

**Tasks**
N.A.

**Documentation**
N.A.

**Screenshots**
N.A.

**Additional context**
I had trouble to put some inclination to the viewpoint, so I used View > Change View > Top View to get the new values.


resources/osm_importer/utils/misc_utils.py
```diff
@@ -80,9 +80,9 @@ def print_header(options, file, minlat, minlon, maxlat, maxlon, elevation=None):
     file.write("  lineScale " + str(round(max(xSize, zSize) / 200.0)) + "\n")
     file.write("}\n")
     file.write("Viewpoint {\n")
-    file.write("  orientation -0.443 0 0.896 3.14102\n")
+    file.write("  orientation -0.57735 0.57735 0.57735 2.0944\n")
     position = round(xSize * math.cos(0.785) * 1.5 + zSize * math.cos(0.785) * 1.5)
-    file.write("  position " + str(position * 0.75) + " 0 " + str(position) + "\n")
+    file.write("  position " + "0" + " 0 " + str(position) + "\n")
```
omichel, line 85:
```suggestion
    file.write("  position 0 0 " + str(position) + "\n")
```

omichel (changes requested):
Thank you for this nice contribution.
I have a request for some simplification in your code.

stefaniapedrazzi:
The changes to the coordinate system of the worlds imported from OSM was changed on purpose to be consistent with the new global coordinate system (ENU) used in Webots since version R2022a.
These changes were introduced in  #1643:
* `projection.py`: https://github.com/cyberbotics/webots/pull/1643/files#diff-e61ad3409535312f1f34fdbc7c663beea73cd513c93a6e03ecd42bc729f51eb4
* `misc_util.py`:https://github.com/cyberbotics/webots/pull/1643/files#diff-806a12a552161fa136924a632fa0b336b0e9bff4048fda655d011a3c12f3b49d

Since Webots R2022a it is strongly suggested to use ENU as global coordinate system for new simulations.

@fparat: could you please further explain why would you like to revert back the orientation of the importer OSM world to the former NUE coordinate system?
Do you have any particular issue with the ENU global coordinate system?

fparat:
This fix intends to respect the ENU system.

In my example you can see that without the fix the objects are placed with X=south Y=east and Z=up.

With the fix, X=East, Y=North and Z=Up.

Maybe there was a regression since #1643, or maybe I am changing the wrong code, in which case I would need some pointers on which part to fix.

stefaniapedrazzi (commented):
You are right!
I just tested and the imported world is not oriented correctly.
So the fix is fine.

About the Viewpoint settings, maybe we could restore a similar view as before using values like the following ones.
```
    file.write("Viewpoint {\n")
    file.write("  orientation -0.3922 0.3922 0.8321 1.7536\n")
    ...
    file.write("  position 0 " + str(-position * 0.85) + " " + str(position) + "\n")
```
Would that work for your use case?

fparat:
I updated the branch with your settings, it is nicer than the top view. 

tournoisis_world_fixed_reviewed

omichel (approved):
Thank you.

PR #5796, opened 2023-01-28
Make robot collisions thread-safe.

brettle:
**Description**
This fixes an issue where it was possible for 2 different ODE cluster threads to simulteously try to append to `mCollisionedRobots`.

### DIFF
--- a/src/webots/engine/WbSimulationCluster.cpp
+++ b/src/webots/engine/WbSimulationCluster.cpp
@@ -456,9 +456,11 @@ static bool needCollisionDetection(WbSolid *solid, bool isOtherRayGeom) {
   return false;
 }
 
-void WbSimulationCluster::appendCollisionedRobot(WbKinematicDifferentialWheels *robot) {
-  QMutexLocker lock(&mCollisionedRobotsMutex);
-  mCollisionedRobots.append(robot);
+  if (WbOdeContext::instance()->numberOfThreads() > 1) {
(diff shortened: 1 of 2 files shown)


brettle:
Note: worlds test will fail until PR #5792 is merged.

src/webots/engine/WbSimulationCluster.cpp
```diff
@@ -455,6 +456,11 @@ static bool needCollisionDetection(WbSolid *solid, bool isOtherRayGeom) {
   return false;
 }
 
+void WbSimulationCluster::appendCollisionedRobot(WbKinematicDifferentialWheels *robot) {
+  QMutexLocker lock(&mCollisionedRobotsMutex);
+  mCollisionedRobots.append(robot);
```
omichel, line 461: Did you evaluate the performance impact of this? Is it worth to add some quick test like:
```suggestion
  if (WbOdeContext::instance()->numberOfThreads() > 1) {
    QMutexLocker lock(&mCollisionedRobotsMutex);
    mCollisionedRobots.append(robot);
  } else
    mCollisionedRobots.append(robot);
```

brettle: I didn't see a noticeable performance degradation, but my testing was not exhaustive and your suggestion adds minimal complexity.

omichel (approved):

Issue #5492, opened 2022-11-15
Laser visualization update rate differs from distance sensor rays


stefaniapedrazzi:
In `laser_pointer.wbt`, the laser point visualization update rate (64 ms) is lower than the update rate of the distance sensor rays (optional rendering updated at basicTimeStep 8ms).
This is not really good because most of the time the laser pointer are noticeable not at the correct position and not aligned with the DistanceSensor rays.

It would be better to increase use the basicTimeStep as sampling rate for the DistanceSensor device so that the visualization of the laser point and sensor rays is always consistent.

![laser_pointer](https://user-images.githubusercontent.com/5910449/201867363-9b221375-1d16-4ea3-b73a-451eef86a676.png)

ShuffleWire:
AFAIK, it remind me https://github.com/cyberbotics/webots/issues/3168 . I'm correct. It should be a good motivation to fix it together

ShuffleWire:
On a broader scale, their is **A LOT** of `#define TIMESTEP 32` thingies in the controllers codebase. (sample, example, tests...), and I feels it should be good to get ride of it altogether...

omichel:
Sure, but let's do it step-by-step, this is more manageable.

stefaniapedrazzi:
There are some good reason to use a different time step in the controller than the world basic time step.
Unless this is causing issues, like in this example, we should not change all of them.

ShuffleWire:
> There are some good reason to use a different time step in the controller than the world basic time step. Unless this is causing issues, like in this example, we should not change all of them.

What example do you have in mind ? Just to be able what to change or not before hitting the review.... Beside case when we specifically want to test time_step behavior, ofc

stefaniapedrazzi:
The choice of the controller time step and basic time step is very often made so that to improve the physics computation precision without slow down too much the simulation.
So, unless it is causing particular issues we should not change the controller step of tests or any other examples in batch.

Modifying the tutorials controllers and the default controllers is the only improvement that seems to be meaningful.

PR #5798, opened 2023-01-30
Laser visualization update rate differs from distance sensor rays #5492

Chandan-Bharadwaj:
**Description**
There is a difference in laser point visualization update rate (64 ms) the update rate of the distance sensor rays (8ms). This is causing a noticable difference in the pointer position and the sensor rays. Updating the laser pointer controller time step to 8ms.

**Related Issues**
This pull-request fixes issue #5492

**Tasks**
  - [ ] Task 1 : changed laser pointer controler time step from 64 to 8


stefaniapedrazzi (changes requested):
Thank you for your contribution!

This PR contains changes that are not related and should be removed before it can be merged.
Additionally, the fix should be applied to the `master` branch.

Given that in this particular simulation the time step choice is critical, it would be better to retrieve the world basic time step using the API function `wb_robot_get_basic_time_step` to avoid this problem to occur again in the future.

Chandan-Bharadwaj:
Thank you for considering my Contribution.

I will make the necessary changes in a new PR.

PR #5800, opened 2023-01-30
Remove unused material references

ad-daniel:
**Description**

When streaming, if an `obj` file references an `mtl` file, this is loaded (or downloaded) from a location that is assumed to be the combination of the URL of the wavefront together with the relative URL of the material. In several cases however we don't distribute/use the material file, which results in 404s.


BenjaminDeleze (approved):
Thank you!
Good initiative, we must be aware that the `.mtl` line will be rewritten every time we save the `.obj` file in Blender so we will need to remove it again.

PR #5801, opened 2023-01-30
Laser visualization update rate differs from distance sensor rays #5492

Chandan-Bharadwaj:
**Description**
There is a difference in laser point visualization update rate (64 ms) the update rate of the distance sensor rays (8ms). This is causing a noticable difference in the pointer position and the sensor rays. Updating the laser pointer controller time step to the basic time step of robot (8ms).

**Related Issues**
This pull-request fixes issue #https://github.com/cyberbotics/webots/issues/5492

### DIFF
--- a/projects/samples/devices/controllers/laser_pointer/laser_pointer.c
+++ b/projects/samples/devices/controllers/laser_pointer/laser_pointer.c
@@ -25,7 +25,7 @@
 #include <webots/motor.h>
 #include <webots/robot.h>
 
-#define TIME_STEP 8
+#define TIME_STEP wb_robot_get_basic_time_step() 
 
 int main() {
   // initialize webots stuff
--- a/projects/samples/devices/controllers/laser_pointer/laser_pointer.c
+++ b/projects/samples/devices/controllers/laser_pointer/laser_pointer.c
@@ -25,20 +25,20 @@
 #include <webots/motor.h>
 #include <webots/robot.h>
 
-#define TIME_STEP wb_robot_get_basic_time_step() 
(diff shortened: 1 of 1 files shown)


projects/samples/devices/controllers/laser_pointer/laser_pointer.c
```diff
@@ -25,7 +25,7 @@
 #include <webots/motor.h>
 #include <webots/robot.h>
 
-#define TIME_STEP 64
+#define TIME_STEP wb_robot_get_basic_time_step()
```
stefaniapedrazzi, line 28: Even if this works correctly it would be better to use the same structure as in the other worlds.
See for example:
https://github.com/cyberbotics/webots/blob/68c3a24f07e68bd364904e5efa3d80f1c59a86d2/projects/samples/devices/controllers/axial_and_tail_rotors/axial_and_tail_rotors.c#L46

stefaniapedrazzi (changes requested):

projects/samples/devices/controllers/laser_pointer/laser_pointer.c
```diff
@@ -25,20 +25,20 @@
 #include <webots/motor.h>
 #include <webots/robot.h>
 
-#define TIME_STEP 64
-
 int main() {
+
```
stefaniapedrazzi, line 29: ClangFormat test is failing because of this new line.
It should be removed to pass the tests.

stefaniapedrazzi (approved):
Thank you!

Issue #5803, opened 2023-01-31
robot_window_html test hangs intermittently


brettle:
**Describe the Bug**
Every once in a while the `robot_window_html` test hangs on my local machine.

**Steps to Reproduce**
1. Run `test_suite.py`
2. The suite will pause when it gets to `robot_window_html` with the robot window displayed in a new Chrome tab.
3. The test will eventually timeout before continuing.

**Expected behavior**
The test should pass consistently.

**System**
 - Fedora 37 Workstation
 - Graphics Card: NV167 / Mesa Intel® UHD Graphics 630 (CFL GT2)
 - Browser: Google Chrome 109.0.5414.74

**Additional context**
I suspect a race condition somewhere between webots, the robot controller, and the javascript running in the browser, but I've tried introducing intentional delays into the code in various intentional locations but I've thus far been unable to reproduce the problem reliably. 

Does this issue occur for anyone else?

omichel:
> Does this issue occur for anyone else?

Not for me, but it looks like a race condition indeed.

brettle:
Found it. Problem was occurring when I ran the tests while an instance of webots was already running. The test webots was listening on the same port and sometimes the original webots would be the one to accept connection.

This will be fixed by PR #5806.

PR #5806, opened 2023-01-31
Don't use port that is already in use

brettle:
**Description**
At least on some versions of Linux, if a Webots process is already running, a new Webots process will try to use the same port instead of trying the next higher port number.

**Related Issues**
This pull-request fixes issue #5803


brettle:
I have a fix ready. I just want to see the test fail in CI first.

brettle:
[Here](https://github.com/cyberbotics/webots/actions/runs/4056247886/jobs/6981387998) is an example of the test failing in CI before the proposed fix. Fyi, it failed on ubuntu-22.04 but not ubuntu-20.04 or macos-11.

ygoumaz (changes requested):
Although the test suite seems to work correctly, warning and info messages in the console of Webots are not correct for normally started instances. 

If I start a first instance of Webots on the `projects/samples/devices/worlds/gps.wbt` world, I get the following INFO message that should not be displayed:
```bash
INFO: Non-streaming server listening on port 1234.
```

Nothing should be displayed in case the port is set by default to 1234 or by the command line option (`--port`) and it could be opened correctly.

When I open a second instance of Webots on the same world, I obtain two Warnings:
```bash
WARNING: Could not listen to extern controllers and robot windows on port 1235. . Using port 1235 instead. This could be caused by another running instance of Webots. You should use the '--port' option to start several instances of Webots.
WARNING: Could not listen to extern controllers and robot windows on port 1234. Port 1234 is already in use. Using port 1235 instead. This could be caused by another running instance of Webots. You should use the '--port' option to start several instances of Webots.
```

The first warning should not be displayed at all. Only the second one should.

brettle:
@ygoumaz, thanks for pointing out the duplicate warnings. I'll look into that. 

Regarding the info message about the server starting, displaying such a message when starting a server is quite common, and in fact webots already displays such a message when starting a streaming server. If such a message was being displayed when I started working on issue #5803, it might have helped me diagnose the problem earlier.

brettle:
I've removed the info message in the interest of getting the actual fix accepted. I can create a separate PR for the info message if others agree it's appropriate.

ygoumaz (commented):
The fix seems to work nicely. However, I don't really understand why we need to connect to the instance to know if it exists. 
Further in the `WbTcpServer::create()` function, we try to listen on the given port. If the port is already in use, shouldn't it always throw the error `"Cannot set the server in listen mode: The bound address is already in use"` ? Do you explain why this is not always the case in the test suite?

brettle:
> If the port is already in use, shouldn't it always throw the error `"Cannot set the server in listen mode: The bound address is already in use"` ? 
The behavior to be platform dependent. On some platforms the socket that Qt creates can reuse the address and Qt does not provide a way to change that before listening.

> Do you explain why this is not always the case in the test suite?
I've added a comment to `WbTcpServer.cpp` which hopefully explains it.

ygoumaz (commented):
Very nice. I wasn't aware of the difference of behavior in function of the platform. However the test suite crashed because of the same issue. Apparently, there seems to be random failures as it was succeeding previously.

brettle:
The test suite succeeded on [a 2nd attempt](https://github.com/cyberbotics/webots/actions/runs/4074377951). I suspected the failure might be do to either tests starting before the background webots has finished starting up or the background webots terminating early for some reason, so I added code to ensure the test wait until the background webots is listening and to detect early termination (and provide diagnostic info). I've been unable to reproduce the failure despite [5 consecutive successful attempts](https://github.com/cyberbotics/webots/actions/runs/4077270904).

My inability to reproduce the problem prompted me to look more carefully at the log and I noticed at the very end of the test suite run it [says](https://github.com/cyberbotics/webots/actions/runs/4074377951/jobs/7020198436#step:7:4380):
```
Warning: The X11 connection broke (error 1). Did the X11 server die?`
```
Some googling indicates that the Xvfb might be getting killed by Linux's OOM do a low memory condition. The [GitHub-hosted runners have 7GB of RAM](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners) and on my machine, one instance of webots is using 3.4GB. So I suspect that by starting a background copy of webots, I'm so close to the memory limit that (very) occasionally Xvfb is getting killed and that is causing the background webots to exit.

So what to do? You could accept this PR as-is (and keep an eye out for a recurrence of the problem), or I could change the PR to start a little dummy python server on port 1234 instead of starting webots. The primary advantage I see to the former is that it would also expose any other problems associated with running multiple copies of webots.

Thoughts?

ygoumaz:
I think we can leave it this way. It is a good thing to detect all issues that could be related to multiple running instances and it seems to work correctly without any failure currently. 

I believe the original failure is indeed the delay for the instance to be ready and listening. I am not really sure that the warning
```
Warning: The X11 connection broke (error 1). Did the X11 server die?
```
is related to the RAM. The dummy instance of Webots is started on an empty world (at least it should be) and should not take up gigabytes of memory. On my machine, an instance opened on the empty world takes only 130MB.

This error may simply be displayed because the process running the dummy instance (from the test suite) is killed in an unclean way.

ygoumaz (changes requested):
I think the PR is almost ready. Just a minor comment.

ygoumaz (approved):
Thank you! The failure of the test suite is not related to this PR.

PR #5807, opened 2023-01-31
Fixed stdout/stderr redirection with extern Python controllers on Windows

omichel:
For some reason the stdout/stderr redirection is not working for Python extern controllers on Windows.
It seems the streams of the libController are not shared with the Python environment using the libController.
This PR fixes the problem by catching the Python stdout/stderr in a pipe, reading them asynchronously in some thread and sending them to the stdout/stderr of the libController.

I tested it successfully from both a MSYS2 console and a DOS console.


ygoumaz (approved):
I tested on Windows (also with the new controller launcher) and it works nicely in all types of consoles! I also tested the UTF-8 characters and it works too now.

Thank you!

PR #5811, opened 2023-01-31
Fix web proto issues related to vehicles

ad-daniel:
**Description**

- [x] Fix Slot issues prevent loading of vehicles
- [x] Improve vehicle loading time ==> Tesla now loads in ~3 seconds rather than ~15
- [x] Fix updating of URL in webotsjs
- [x] Fix vehicle lights being in the wrong place (wrong PROTO) and being rendered in the wrong place (bug when generating the nodes)
- [x] Fix issues with boundingObject

https://proto.webots.cloud/run?version=R2023b&url=https://github.com/ad-daniel/prefab/blob/master/protoExamples/protos/tesla/TeslaModel3.proto&type=undefined


resources/web/wwi/Parser.js
```diff
@@ -1086,11 +1086,10 @@ export default class Parser {
     const radius = parseFloat(getNodeAttribute(node, 'radius', '100'));
     const ambientIntensity = parseFloat(getNodeAttribute(node, 'ambientIntensity', '0'));
     const castShadows = getNodeAttribute(node, 'castShadows', 'false').toLowerCase() === 'true';
-
     const spotLight = new WbSpotLight(id, on, attenuation, beamWidth, color, cutOffAngle, direction, intensity, location,
       radius, ambientIntensity, castShadows, parentNode);
 
-    if (typeof parentNode !== 'undefined' && typeof spotLight !== 'undefined')
+    if (typeof parentNode !== 'undefined')
       parentNode.children.push(spotLight);
```
BenjaminDeleze, line 1093: Do we really want to push `undefined` as children?

BenjaminDeleze (changes requested):
The bounding objects are too big

ad-daniel:
All issues should be addressed now, I've uploaded to the server and this time I actually committed the BO fix to the branch, not just on the server :sweat_smile:

BenjaminDeleze:
I confirm that the previous issues are fixed.

I found a new one but I am not sure it's related to this PR:
When there is a node in the extension slot of the nao or tesla we cannot download the model

ad-daniel:
Yes it isn't related to this PR but it should work now.

BenjaminDeleze:
Did you push the fix to the server? Because I can still reproduce the issue.

I found another as well:
I cannot add anything in the hand slots of the `Nao` whereas in Webots any node can fit.
(If you want to merge this PR and handle the new issues later in another one, we can)

ad-daniel:
> Did you push the fix to the server? Because I can still reproduce the issue.

No, it was local only.

> I cannot add anything in the hand slots of the Nao whereas in Webots any node can fit.

The issue was introduced when removing the proto-list.xml, the check only looked at the first IS link (which in itself is fine), but it didn't go full depth down to the base node. In the case of the Nao, the hands are a separate PROTO and the IS chain is longer.

I've fixed it and pushed files to the server

BenjaminDeleze (approved):
Thank you! well done!

PR #5813, opened 2023-01-31
Log test suite performance

brettle:
**Description**
Automatically log test suite performance to facilitate detection of changes that affect performance.

**Additional context**
I'm requesting this be merged into master so that it can be used when making hotfixes.


omichel (commented):
That looks like a very good idea.
However, it doesn't provide yet an automatic way to detect performance regression.
But that's already a good PR to merge.
However, I am not convinced about the idea of committing the performance_log_develop.txt file for several reasons:
1. This file seems to be specific to your machine (refers to /home/brettle/git/webots/...).
2. The average speed factor is highly dependent on the machine on which it was run.
3. A good workflow to detect performance regression doesn't need this file:
   1. Perform a performance test on the master or develop branch on the developer's machine
   2. Perform a performance test on the new branch on the developer's machine.
   3. Compare the two performance tests.

So, unless there is a good justification to keep this on the repository, I would remove it.

stefaniapedrazzi:
Adding tests about performance is good.
But this PR is not adding such a tests. It is simply adds performance log that will never be analyzed and will probably just slow down the test suite.
So personally as is I don't see the point of adding the performance log to the test suite.
Even because this is something that can be added very easily manually when testing some changes.

brettle:
@omichel `performance_log_develop.txt` was included by mistake and has been removed. Thanks for catching that.

@stefaniapedrazzi I haven't noticed any performance degradation associated with running the test suite with performance logging enabled. This is designed to be used in the workflow @omichel described above. In fact, that's exactly what I did when working on PR #5792. I didn't see any other way to log the performance of the test suite.

stefaniapedrazzi:
This log file will be created each time the tests are run on the CI and it will never be checked or read or stored for reference.
So most of the time the `test_suite` is run the log file is just useless.
I see very few cases where this file will be used if it is not analyzed automatically.
Even because  it is just a starting point in the performance comparison but it is not enough:
* tests needs to be run multiple times in order to make sure to have meaningful measurements
* the tests worlds are not very good example of standard Webots simulation because they are simplified examples

Adding an argument to enable it just when need would make much more sense to me than enabling it everytime by default.
However, given that to generate it you just have to add `--log-perfomance=<log_file>` to the Webots command there are not many advantages of having an argument w.r.t to changing the `test_suite.py` script directly.

omichel:
I agree with @stefaniapedrazzi: logging the performance should be optional as we don't automatically analyze it in the CI.
For me, it is OK to add a command line option to `test_suite.py` for enabling performance logs, as it might be useful for manual generic performance tests.

brettle:
It is now optional.

omichel (approved):
Looks good to me.
Thank you.

PR #5814, opened 2023-01-31
Run httpd for test cache environment.

brettle:
**Description**
Allow cache tests to get files from the local dir instead of web.

**Additional context**
Before this change, the cache tests retrieved the files from GitHub under the branch that the user was in. However, when working on PRs, this cause the cache tests to fail because the branch was not yet (and might never be) in the official repo. To address this and also ensure that the retrieved files are from the current work tree, this change starts a simple python http server on port 8000.

### DIFF
--- a/tests/cache/cache_environment.py
+++ b/tests/cache/cache_environment.py
@@ -14,8 +14,11 @@
 
 """Sets the environment so that the cache tests can be performed in any condition"""
 
+import atexit
+import http.server
 import os
 import sys
+import threading
 from pathlib import Path
 
 if 'WEBOTS_HOME' in os.environ:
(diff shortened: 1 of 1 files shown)


stefaniapedrazzi:
I'm not sure this is really what we want.
The cache tests are explicitly about testing retrieving the assets in many different ways. If the cache test fail locally because the branch doesn't exists the failures can simply be ignored if nothing related to the "cache" or asset mechanism is changed.

brettle:
Yeah, when I started looking into the test failures resulting from this PR I realized I was in over my head. I'll consider submitting another PR with a different workaround.

Issue #5708, opened 2023-01-04
Position mismatch with velocity control of a RotationalMotor on a Hinge2Joint


CoolSpy3:
**Describe the Bug**
When using velocity control with a RotationalMotor attached to a Hinge2Joint (or HingeJoint), there accumulates a mismatch between the position of the joint in the world tree and the position of the joint in the simulation.

**Steps to Reproduce**
1) Open the sample world `samples > devices> hinge_2_joint_with_backlash.wbt`
2) REPLACE line 46 in the controller with:
```
wb_motor_set_position(motor1, INFINITY);
wb_motor_set_velocity(motor1, fabs(cos(0.01 * counter)));
```
3) Save and build the controller, and reload the world
4) Let the world run in fast mode for a while (I waited until an effective time of 24 minutes had elapsed)
5) Pause the world
6) In the world tree, open `DEF WIHTOUT_BACKLASH Robot > children > Solid "solid" > children > Hinge2Joint > JointParameters HingeJointParameters > position`
7) Without changing the value, click in the value entry box and press ENTER
8) The joint moves to a significantly different position

With a HingeJoint, the behavior is a little harder to see visually, but I've been able to reproduce it with:
1) Open the sample world `samples > devices> hinge_joint_with_backlash.wbt`
2) Let the world run in fast mode for a while (I waited until an effective time of 24 minutes had elapsed)
3) Pause the world
4) In the world tree, open `Robot "robot" > children > HingeJointWithBacklash > JointParameters HingeJointParameters > position`
5) Without changing the value, click in the value entry box and press ENTER
6) The joint moves to a significantly different position

**Expected behavior**
There should be no change after forcing an update to the position in the world tree because it should already be the value in the simulation.

**Screenshots**
Hinge2Joint after running simulation:
![webots_pos_bug_before](https://user-images.githubusercontent.com/55305038/210516238-44fa8334-225a-44d2-aa77-965fef169b40.PNG)

Hinge2Joint after running simulation and forcing position update to world tree
![webots_pos_bug_after](https://user-images.githubusercontent.com/55305038/210516318-536f4fc9-def8-48ab-8b53-717237f9221e.PNG)

HingeJoint after running simulation
![webots_pos_bug_before_hj1](https://user-images.githubusercontent.com/55305038/210516370-bc0188a1-2fd8-40b4-a5d9-7a9b2130c971.PNG)

HingeJoint after running simulation and forcing position update to world tree (look at the blue flag)
[...]

PR #5834, opened 2023-02-05
Hotfix for inaccurate HingeJoint, Hinge2Joint, and BallJoint position under velocity control

brettle:
**Description**
This fixes a bug where, when using velocity control with a `RotationalMotor` attached to a `HingeJoint`, `Hinge2Joint`, or `BallJoint`, a mismatch accumulates between the joint position(s) reported by attached `PositionSensors` (or shown in the position field of the `JointParameters`) and the actual position of the joint in the simulation. Tests have been added to verify the fixes.

**Related Issues**
This pull-request fixes issue #5708

**Additional context**
Note that to fix `BallJoints` I needed to make a small fix need ODE's `dJointGetAMotorAngle()`. The issue was that ODE only updated the motor euler angles before the step. As a result, the angles returned by `dJointGetAMotorAngle()` were always one step behind. I don't know whether this bug still exists in the upstream ODE code.


tests/physics/controllers/coupled_motor/coupled_motor.c
```diff
@@ -170,7 +170,7 @@ int main(int argc, char **argv) {
       motors[i] = wb_robot_get_device_by_index(2 * i + 1);
 
     double positions[NB_SENSORS];
-    double tolerance = 1e-10;
+    double tolerance = 1e-4;
```
ad-daniel, line 173: I'm not sure to understand why this specific test required changing the tolerances and not others. As far as I recall (and it has been a while, so I might recall incorrectly), mostly we test that instructions provided to one motor (and relayed to the others in the coupling) result in both joints reaching the same position (at a multiplier coefficient difference, if a multiplier is present). So intuitively we should expect them to be very strictly close to each other. Was this change necessary?

ad-daniel (changes requested):
Thank you for the contribution! Overall I think the change is legitimate, and indeed the deviation is no longer there.
May I ask you to synchronize this branch with the base one? For some reason github doesn't let me. I'll do additional tests after

tests/physics/controllers/coupled_motor/coupled_motor.c
brettle: Yes, these tolerance changes are necessary. They reflect the amounts that the *real* joint positions have always varied when not under position control. We just didn't notice before because the position values reported were integrated estimates based on the joint velocities, not the real joint positions. Note that the tolerance for the position control portion of the test did not change. Make sense?

tests/api/worlds/motor2_velocity.wbt
```diff
@@ -0,0 +1,130 @@
+#VRML_SIM R2023a utf8
+
+EXTERNPROTO "webots://projects/objects/floors/protos/Floor.proto"
+EXTERNPROTO "webots://tests/default/protos/TestSuiteSupervisor.proto"
+EXTERNPROTO "webots://tests/default/protos/TestSuiteEmitter.proto"
+
+WorldInfo {
+  coordinateSystem "NUE"
```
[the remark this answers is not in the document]
brettle: Switched to ENU.

tests/api/controllers/motor2_velocity/motor2_velocity.c
```diff
@@ -0,0 +1,88 @@
...
+  // Run the motor at varying velocities
+  wb_motor_set_position(motor, INFINITY);
+  wb_motor_set_velocity(motor, 0);
+  for (i = 0; i < 100; ++i) {
+    wb_motor_set_velocity(motor, max_velocity * fabs(cos(i * 2 * M_PI / 100)));
+    wb_robot_step(TIME_STEP);
+    quaternion = wb_inertial_unit_get_quaternion(inertial_unit);
+    position = wb_position_sensor_get_value(position_sensor);
+    // printf("position = %g, quaternion = %g, %g, %g, %g\n", position, quaternion[0], quaternion[1], quaternion[2],
+    //        quaternion[3]);
```
ad-daniel, line 38:
```suggestion
```

tests/api/controllers/motor2_velocity/motor2_velocity.c
```diff
@@ -0,0 +1,88 @@
...
+                            sensor_name, quaternion[0], quaternion[1], quaternion[2], quaternion[3], actual_quaternion[0],
+                            actual_quaternion[1], actual_quaternion[2], actual_quaternion[3]);
+
+  // The position itself should not be normalized to be between -pi and pi
+  double min_expected_position = max_velocity / 2 * TIME_STEP / 1000.0 * 100.0;
+  ts_assert_double_is_bigger(position, min_expected_position, "The position measured by %s shoud be at least %g but is only %g",
+                             sensor_name, min_expected_position, position);
+
+  // Returning to original position
+  // printf("Returning to original position\n");
```
ad-daniel, line 61:
```suggestion
```

tests/api/controllers/motor2_velocity/motor2_velocity.c
ad-daniel: sorry, forgot about these 2

ad-daniel (changes requested):
Just a small detail I missed left, other than that I made some tests and it looks good to me, thanks!

ad-daniel (dismissed):
Thank you!

ad-daniel (approved):

PR #5839, opened 2023-02-06
Hotfix for retrieval of sibling devices returning wrong devices for BallJoint's device3 devices.

brettle:
**Description**
Fixes a bug where function calls to retrieve sibling devices for `BallJoint` `device3` devices incorrectly returned `device` devices.


tests/api/controllers/position_sensor_get_motor/position_sensor_get_motor.c
```diff
@@ -0,0 +1,38 @@
...
+#include <webots/robot.h>
+
+#include "../../../lib/ts_assertion.h"
+#include "../../../lib/ts_utils.h"
+
+#define TIME_STEP 32
+
+void test_get_motor(char *sensor_name, char *expected_motor_name) {
+  WbDeviceTag position_sensor = wb_robot_get_device(sensor_name);
+  char *actual_sensor_name = wb_device_get_name(position_sensor);
```
stefaniapedrazzi, line 12:
```suggestion
  const char *actual_sensor_name = wb_device_get_name(position_sensor);
```

tests/api/controllers/position_sensor_get_motor/position_sensor_get_motor.c
```diff
@@ -0,0 +1,38 @@
...
+  ts_assert_string_equal(actual_sensor_name, sensor_name, "wb_device_get_name(wb_robot_get_device(\"%s\")) returned \"%s\"",
+                         sensor_name, actual_sensor_name);
+  WbDeviceTag motor = wb_position_sensor_get_motor(position_sensor);
+  char *motor_name = wb_device_get_name(motor);
```
stefaniapedrazzi, line 16:
```suggestion
  const char *motor_name = wb_device_get_name(motor);
```

tests/api/worlds/position_sensor_get_motor.wbt
```diff
@@ -0,0 +1,143 @@
+#VRML_SIM R2023a utf8
+
+EXTERNPROTO "webots://projects/objects/floors/protos/Floor.proto"
+EXTERNPROTO "webots://tests/default/protos/TestSuiteSupervisor.proto"
+EXTERNPROTO "webots://tests/default/protos/TestSuiteEmitter.proto"
+
+WorldInfo {
+  coordinateSystem "NUE"
```
stefaniapedrazzi, line 8: New test world should use the default ENU coordinate system.

tests/api/controllers/position_sensor_get_motor/position_sensor_get_motor.c
```diff
@@ -0,0 +1,38 @@
...
+  wb_robot_step(TIME_STEP);
+
+  test_get_motor("position sensor1", "rotational motor1");
+  test_get_motor("position sensor2", "rotational motor2");
+  test_get_motor("position sensor3", "rotational motor3");
+
+  // position sensor3 does not have an associated brake. Make sure we don't get
+  // the brake associated with the first position sensor instead.
+  WbDeviceTag brake = wb_position_sensor_get_brake(wb_robot_get_device("position sensor3"));
+  ts_assert_pointer_null(brake, "wb_position_sensor_get_brake(wb_robot_get_device(\"position sensor3\")) != NULL");
```
stefaniapedrazzi, line 34: `WbDeviceTag` is an integer and not a pointer.
```suggestion
  ts_assert_int_equal(brake, 0, "wb_position_sensor_get_brake(wb_robot_get_device(\"position sensor3\")) != NULL");
```


Moreover it would be better to have meaningful error messages than simply printing the test instruction.
Could you please improve all the error messages?

tests/api/controllers/position_sensor_get_motor/position_sensor_get_motor.c
```diff
@@ -0,0 +1,38 @@
...
+int main(int argc, char **argv) {
+  ts_setup(argv[0]);
+  wb_robot_step(TIME_STEP);
+
+  test_get_motor("position sensor1", "rotational motor1");
+  test_get_motor("position sensor2", "rotational motor2");
+  test_get_motor("position sensor3", "rotational motor3");
+
+  // position sensor3 does not have an associated brake. Make sure we don't get
+  // the brake associated with the first position sensor instead.
```
stefaniapedrazzi, line 32: We try to avoid to split sentences over multiple lines.
```suggestion
  // position sensor3 does not have an associated brake.
  // Make sure we don't get the brake associated with the first position sensor instead.
```

stefaniapedrazzi (changes requested):
Thank you for the fix!

I manly have comments about the new test files:
* it seems that the world `tests/api/worlds/motor3_velocity.wbt` is unrelated to this PR and should be removed
* the new controller is throwing many warnings during compilation that have to be fixed.

Just a side note: it would be better to try limit the number of test worlds and merge tests together when possible. So if you plan to add a test using the same world but testing a different functionality, then please merge the two test controllers programs instead of adding two different tests simulation. In particular given that this test about retrieving the sibling devices is very simple and doesn't require any particular environment.

brettle:
I think I've addressed your feedback.

stefaniapedrazzi (approved):
Thank you

stefaniapedrazzi (approved):

PR #5845, opened 2023-02-06
Make bash scripts exit when any command fails.

brettle:
This should make failing builds fail faster.


omichel (approved):
Thank you.

omichel:
This PR is causing the scripts to fail when running them on a Ubuntu 20.04 machine whereas it used to work (as some errors can be safely ignored). I will provide a fix in a new PR.

brettle:
Sorry about that. I'd figured such problems would be caught during CI.

omichel:
No problem. Such things happen every day.

PR #5847, opened 2023-02-07
Update Python receiver get data documentation

stefaniapedrazzi:
In the Python API the `Receiver.getData` function is deprecated but still documented.


omichel (approved):
Oops, I missed this one.
Thank you.

PR #5848, opened 2023-02-07
Fix documentation about wwi

ad-daniel:
**Description**

Some information about wwi wasn't up to date.

### DIFF
--- a/docs/reference/robot.md
+++ b/docs/reference/robot.md
@@ -2370,3 +2370,5 @@ while ((message = wb_robot_wwi_receive_text())) {
 }
 ```
 The reading head will be reset to the beginning of the buffer of received messages each time a time step is performed.
+
+> **note** [Java, Python, MATLAB, ROS]: `wb_robot_wwi_receive` and `wb_robot_window_send` functions are not available in the Java, Python, MATLAB, or ROS API.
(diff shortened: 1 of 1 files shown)


omichel (dismissed):
Thank you.

docs/reference/robot.md
```diff
@@ -2361,13 +2361,12 @@ The message is sent using the `webots.window("<robot window name>").send` method
...
+To read the full buffer, you should call repeatedly this function until it returns `NULL`:
 ```C
 const char *message;
 while ((message = wb_robot_wwi_receive_text())) {
   // do something with each message
 }
 ```
 The reading head will be reset to the beginning of the buffer of received messages each time a time step is performed.
-
-> **note** [Java, Python, MATLAB, ROS]: `wb_robot_wwi_receive` and `wb_robot_window_send` functions are not available in the Java, Python, MATLAB, or ROS API.
```
stefaniapedrazzi, line 2373: Why did you remove this line?
It seems still correct that these functions are not available in Java Python MATLAB and ROS.

ad-daniel: Oh I see, it's the `receive` methods that are missing, all languages only have at least the `receive_text`, I must have read it wrongly then

stefaniapedrazzi (approved):
Thank you!

PR #5850, opened 2023-02-07
Enhancement led support

BenjaminDeleze:
**Description**

Add led support for web protos.

It is possible that the color keyword is present in the x3d but is not associated with any value. It is the case when the LED is a rgb gradual LED. In this case, we set the LED to the default color: red.

Concerning the Tesla (and other vehicles) protos, we can have the impression that the representation of some LEDs does not work. However this is because the vehicles have LEDs without children (all the LEDs are present in the generic car but the Tesla does not fill all the fields)

Test it here:

https://webots.cloud/run?version=R2023b&url=https://github.com/ad-daniel/prefab/blob/master/protoExamples/protos/tesla/TeslaModel3.proto&type=undefined

https://webots.cloud/run?version=R2023b&url=https://github.com/cyberbotics/webots/blob/develop/projects/robots/softbank/nao/protos/Nao.proto&type=undefined


BenjaminDeleze:
As discussed I changed the color to white when there is an empty list.

You can see it in the Nao example

omichel (approved):
Great, thank you.

PR #5851, opened 2023-02-07
Enhancement for simgait

BenjaminDeleze:
This improvement is necessary to make the Simgait animations work.

It allows us to give a json file directly to the animation, instead of supplying an url


omichel (changes requested):

omichel (approved):
Thank you.

PR #5858, opened 2023-02-08
Fix uninitialized set_exposure variable.

brettle:
Valgrind's memcheck tool reported this while I was doing some other testing.

It looks like at most this could cause a little bit of unnecessary communication between the controller and webots.


omichel (approved):
Well spotted!
Thank you.

Issue #5780, opened 2023-01-25
Test `robot_parallel_step` failing on local machine running in fast mode


stefaniapedrazzi:
On my machine (Ubuntu 22.04) the `robot_parallel_step` api test is failing when it is ran in fast mode.
Running it at normal speed works correctly.

I experienced this only in develop branch (R2023b).
I have Webots compiled in debug mode, I will also check compiling in release mode.

stefaniapedrazzi:
This is the error I get in the Webots console:
```
Warning: wb_robot_step() called before calling wb_robot_step_end().
Warning: robot_step_begin() called multiple times before calling wb_robot_step_end().
FAILURE with robot_parallel_step: Camera measurement after first pose update is wrong.
WARNING: 'robot_parallel_step' controller exited with status: 1.
```

ad-daniel:
tests in the CI should also run in fast mode, no?

stefaniapedrazzi:
Yes.
I'm now inspecting what occurs in my machine that makes the test fail on the develop branch only.

brettle:
Fwiw, the first 2 warnings are just a result of the assertion code calling `wb_robot_cleanup()` when an assertion fails. `wb_robot_cleanup()` calls `wb_robot_step()` which produces the warnings when a step is in progress. Of course, the real mystery is why assertion fails to begin with.

brettle:
I'm pretty confident that the issue is that images are being transferred from webots to the controller via a memory mapped file. That means that webots writes and controller reads of the image data could occur in an undefined order. When the error occurs, webots has updated the image before the controller has had a chance to read it and check it. I'm going to investigate using a double-buffering scheme to avoid this.

brettle:
Can I add a private member variable to [`WbAbstractCamera`](https://github.com/cyberbotics/webots/blob/master/src/webots/nodes/WbAbstractCamera.hpp) in a PR targeting master? I just need to avoid changing any wire protocols (e.g. the one between controllers and webots) or making ABI-incompatible changes to shared libs (e.g. `libController.so`), right? I obviously want to preserve compatibility on the master branch, but unless I'm missing something, those `src/webots/nodes/*` classes are only compiled into the webots executable. They aren't compiled into any shared lib where ABI compatibility would be an issue.

brettle:
Alternatively (and much more simply), we just document that the contents of image data is not well-defined between `wb_robot_step_begin()` and `wb_robot_step_end()`.

[...]

PR #5860, opened 2023-02-08
Don't use Camera in robot_parallel_step test.

brettle:
**Description**
Data returned by wb_camera_get_image() is subject to change during a step, so it should not be used for this test.

Changed test to use compass, distance, and GPS sensors instead.

**Related Issues**
This pull-request fixes issue #5780


brettle:
I'm getting some [strange test-sources on windows CI failures](https://github.com/cyberbotics/webots/actions/runs/4129144919/jobs/7134712164) that don't appear to be related to anything in this PR. Any idea what's going on?

brettle:
test-sources passes for PR #5862 so that probably needs to be merged to make it pass.

brettle:
Test-sources still fails on Windows and now also fails on Mac. It is also failing in the latest merge PR to merge master into develop so it isn't specific to this PR. Oddly, it didn't fall for PR #5862. I'm stumped. Anyone else have ideas?

stefaniapedrazzi:
I suspect the cppcheck tests were broken by your fix:
https://github.com/cyberbotics/webots/pull/5862

I checked and in #5862 the source tests didn't run (the job is executed but not the tests step).

stefaniapedrazzi:
Actually it is probably unrelated to the #5862 (even if the tests were skipped for Windows and macOs) as also the nightly tests failed.

tests/api/controllers/robot_parallel_step/robot_parallel_step.c
```diff
@@ -74,29 +63,41 @@ int main(int argc, char **argv) {
...
+  // distance sensor
+  distance = wb_distance_sensor_get_value(ds);
+  ts_assert_double_in_delta(distance, 650.0, 0.001, "Distance sensor measurement after first pose update is wrong.");
 
-  // intense computation between begin and end
-  for (int i = 0; i < 1000000; i++) {
-    r = wb_camera_image_get_red(image, 1, 0, 0);
-    g = wb_camera_image_get_blue(image, 1, 0, 0);
-    b = wb_camera_image_get_green(image, 1, 0, 0);
-  }
```
stefaniapedrazzi, line 88: We should keep this loop (or replace it by something similar) even if the camera values won't be used because it is an important part of the test.
Parallelization only makes sense if intense computation occurs on the controller side and that is the main case we should test.

stefaniapedrazzi (changes requested):

stefaniapedrazzi:
I can no longer reproduce the error with this test on my machine, I had it only last week but it disappeared after a recompilation.
So I cannot check if this is solving the issue.
If you can reproduce it consistently and this works for you, then this fix is fine, but we should provide a proper fix in develop.

tests/api/controllers/robot_parallel_step/robot_parallel_step.c
brettle: I've added some computation along with some sleeping.

brettle:
> I can no longer reproduce the error with this test on my machine.

You should be able to reproduce it reliably by adding `sleep(1)` after `robot_step_begin()`. That gives webots enough time to update the image.

brettle:
Regarding a proper fix for develop, I've opened issue #5870 to discuss options.

tests/api/controllers/robot_parallel_step/robot_parallel_step.c
```diff
@@ -74,29 +65,64 @@ int main(int argc, char **argv) {
...
-    g = wb_camera_image_get_blue(image, 1, 0, 0);
-    b = wb_camera_image_get_green(image, 1, 0, 0);
+  // Give webots plenty of time to finish the step so we can be sure we'll detect any unexpected tampering with our values.
+  usleep(time_step * 1000 + 100000);
+
+  // Do some intensive computation, just for good measure
+  int result = 0xffff;
+  for (int i = 0; i < 0xffff; i++) {
+    result &= ~i;
   }
```
omichel, line 93: [CS360](https://github.com/cyberbotics/webots/wiki/CPP-Coding-Style/#cs360-single-if-while-or-for-statement-must-be-written-without-brackets-the-statement-part-should-be-put-on-a-separate-line-in-case-of-nested-block-keep-brackets):
```suggestion
  for (int i = 0; i < 0xffff; i++)
    result &= ~i;
```

omichel (changes requested):

tests/api/controllers/robot_parallel_step/robot_parallel_step.c
```diff
@@ -1,3 +1,5 @@
+#include <unistd.h>
```
stefaniapedrazzi, line 1: Usually we list these libraries after webots and ts_* libraries.

stefaniapedrazzi (dismissed):
Thank you the test is now fine!
I just have a remaining minor comment about the include style.

tests/api/controllers/robot_parallel_step/robot_parallel_step.c
brettle: Fixed.

omichel (approved):
Thank you.

PR #5863, opened 2023-02-09
Upgrade to cpp-check 2.10

ygoumaz:
ditto.


brettle:
Any idea what triggered test-sources to start failing? The title of this PR implies an update to cpp check, but do you know what caused that?

ygoumaz:
The version of cpp-check used in the CI is now 2.10 instead of 2.9. This is caused by the update on the Github CI side.

brettle:
> The version of cpp-check used in the CI is now 2.10 instead of 2.9. This is caused by the update on the Github CI side.

Ah, I see. Have you considered trying to pin the version of cppcheck used in master and just using the newer version on develop? This just seems like a lot of changes to make to a stable branch with no obvious benefit to the end user.

brettle:
You might want to consider using PR #5865 for the master branch and retargeting this PR at develop.

projects/robots/gctronic/e-puck/plugins/remote_controls/e-puck_wifi/Wrapper.cpp
```diff
@@ -207,42 +207,42 @@ int Wrapper::robotStep(int step) {
...
-        unsigned char *rgb888 = (unsigned char *)malloc(160 * 120 * 3);
+        unsigned char *rgb888 = static_cast<unsigned char *>(malloc(160 * 120 * 3));
         rgb565_to_brg888(image, rgb888, 160, 120);
         save_bmp_image(filename, rgb888, 160, 120);
         free(rgb888);
         image_counter++;
 #endif
         got_camera_image++;
-      } else if (header == 0x02) {  // sensor data
+      } else if (header == 0x02) {  // senreinterpret_castsor data
```
omichel, line 234:
```suggestion
      } else if (header == 0x02) {  // sensor data
```

src/webots/gui/WbMainWindow.cpp
```diff
@@ -2295,7 +2295,8 @@ void WbMainWindow::updateOverlayMenu() {
         action->setCheckable(true);
         action->setChecked(device->isOverlayEnabled());
         action->setEnabled(!device->isWindowActive());
-        action->setProperty("renderingDevice", QVariant::fromValue((void *)device));
+        action->setProperty("renderingDevice",
+                            QVariant::fromValue(static_cast<void *>(const_cast<WbRenderingDevice *>(device))));
```
omichel, line 2299: Doesn't look very nice, but is apparently a good practice. See https://stackoverflow.com/questions/51249130/nested-static-cast-and-const-cast

src/webots/nodes/WbCamera.cpp
```diff
@@ -244,8 +244,7 @@ void WbCamera::initializeImageMemoryMappedFile() {
   if (mImageMemoryMappedFile) {
     // initialize the memory mapped file with a black image
     int *im = reinterpret_cast<int *>(image());
-    const int size = width() * height();
-    for (int i = 0; i < size; i++)
+    for (int i = 0; i < width() * height(); i++)
```
omichel, line 247: I believe we might introduce a performance impact with this change.

src/webots/nodes/WbCamera.cpp
```diff
@@ -259,8 +258,7 @@ void WbCamera::initializeSegmentationMemoryMappedFile() {
     unsigned char *data = (unsigned char *)mSegmentationMemoryMappedFile->data();
     // initialize the memory mapped file with a black image
     int *im = reinterpret_cast<int *>(data);
-    const int size = width() * height();
-    for (int i = 0; i < size; i++)
+    for (int i = 0; i < width() * height(); i++)
```
omichel, line 261: Same question here.

tests/sources/test_cppcheck.py
```diff
@@ -123,10 +123,10 @@ def test_sources_with_cppcheck(self):
             'src/webots/widgets',
             'src/webots/wren'
         ]
-        command = 'cppcheck --enable=warning,style,performance,portability --inconclusive --force -q'
+        command = 'cppcheck --enable=warning,style,performance,portability --inconclusive --force -q --library=qt '
         command += ' -j %s' % str(multiprocessing.cpu_count())
         command += ' --inline-suppr --suppress=invalidPointerCast --suppress=useStlAlgorithm --suppress=uninitMemberVar '
```
omichel, line 128:
```suggestion
        command += ' --inline-suppr --suppress=invalidPointerCast --suppress=useStlAlgorithm --suppress=uninitMemberVar'
```

tests/sources/test_cppcheck.py
```diff
@@ -165,7 +165,7 @@ def test_projects_with_cppcheck(self):
         skippedfiles = [
             'projects/robots/robotis/darwin-op/plugins/remote_controls/robotis-op2_tcpip/stb_image.h'
         ]
-        command = 'cppcheck --enable=warning,style,performance,portability --inconclusive --force -q '
+        command = 'cppcheck --enable=warning,style,performance,portability --inconclusive --force -q --library=qt '
         command += '--inline-suppr --suppress=invalidPointerCast --suppress=useStlAlgorithm -UKROS_COMPILATION '
         command += '--suppress=strdupCalled --suppress=ctuOneDefinitionRuleViolation --suppress=unknownMacro '
```
omichel, line 170: This is not very consistent with the previous code where we put the space first and not last...

omichel: I like putting the space first because it means this is a continuation of a previous string being build. Also putting the space first means we don't put it at the end which means we do not assume there will be some additional content.

src/webots/nodes/WbCharger.cpp
```diff
@@ -199,7 +199,7 @@ void WbCharger::prePhysicsStep(double ms) {
     } else
       currentEnergy -= e;  // transfer
 
-    if (mDone == false && mRobot->battery().size() > 2) {  // else energy is wasted
+    if (mDone == false) {  // else energy is wasted
```
omichel, line 202: I am not sure this is correct. We should probably modify line 190 instead.

src/webots/nodes/WbDisplay.cpp
```diff
@@ -1032,6 +1032,7 @@ void WbDisplay::imageLoad(int id, int w, int h, void *data, int format) {
     assert(0);
 
   mImages.push_back(new WbDisplayImage(id, w, h, clippedImage, isTransparent));
+  delete[] clippedImage;
```
omichel, line 1035: We are introducing a double delete here which is wrong.

src/webots/gui/WbGuiApplication.cpp
```diff
@@ -532,7 +532,6 @@ enum WINDOWCOMPOSITIONATTRIB {
 };
 
 struct WINDOWCOMPOSITIONATTRIBDATA {
-  WINDOWCOMPOSITIONATTRIB Attrib;
```
omichel, line 535: This breaks compilation on Windows:

src/webots/nodes/WbCharger.cpp
ygoumaz: After discussion, this is actually correct.

src/webots/gui/WbGuiApplication.cpp
ygoumaz: This was a false positive then.

ygoumaz:
I have gone through all the new cpp-check errors. There are a few left on macOS and Windows (less than 10) that I tried to fix, but couldn't come with something that works, even though I don't think they are false positives.

Adding the --library=qt option to cpp-check is mandatory to avoid errors on signals. But it seems to create additional errors on Ubuntu, that may be false positives (or at least unwanted, like AssertWithSideEffect ones).

omichel:
I may have found a solution to handle nicely the `assertWithSideEffect` warning, which is actually not a false positive I think, see my previous commits.

omichel:
Test Sources is passing on Linux!

src/webots/scene_tree/WbInsertExternProtoDialog.cpp
```diff
@@ -68,6 +68,10 @@ WbInsertExternProtoDialog::WbInsertExternProtoDialog(QWidget *parent) : mRetriev
 }
 
 WbInsertExternProtoDialog::~WbInsertExternProtoDialog() {
+  delete mTree;
+  delete mSearchBar;
+  delete mInsertButton;
+  delete mCancelButton;
```
stefaniapedrazzi: Qt is already handling the destruction of the objects if you pass a QObject as parent in the constructor.
So these four delete instructions should be removed.
```suggestion
```

ygoumaz: Yes true, these are false positives.

src/webots/nodes/WbIndexedFaceSet.cpp
```diff
@@ -334,12 +334,12 @@ void WbIndexedFaceSet::exportNodeContents(WbWriter &writer) const {
...
   const int n3 = n * 3;
-  int *const coordIndex = new int[n3];
-  int *const normalIndex = new int[n3];
-  int *const texCoordIndex = new int[n3];
-  double *const vertex = new double[n * 9];
-  double *const normal = new double[n * 9];
-  double *const texture = new double[n * 6];
+  int *const coordIndexes = new int[n3];
+  int *const normalIndexes = new int[n3];
+  int *const texCoordIndexes = new int[n3];
```
stefaniapedrazzi, line 339: typo
```suggestion
  int *const coordIndices = new int[n3];
  int *const normalIndices = new int[n3];
  int *const texCoordIndices = new int[n3];
```

src/webots/gui/WbVideoRecorderDialog.cpp
```diff
@@ -99,14 +99,15 @@ WbVideoRecorderDialog::WbVideoRecorderDialog(QWidget *parent, const QSize &curre
...
-  double acceleration;
-  bool caption;
-  WbPreferences::instance()->moviePreferences(resolutionIndex, quality, acceleration, caption);
-  mResolutionList.setCurrentRow(resolutionIndex);
-  mQualitySpinBox->setValue(quality);
-  mAccelerationSpinBox->setValue(acceleration);
-  mCaptionCheckBox->setChecked(caption);
+  int qualityPreference, resolutionIndexPreference;
+  double accelerationPreference;
+  bool captionPreference;
```
stefaniapedrazzi, line 104: These variable names doesn't seem very well chosen.
They are way longer than before without adding any additional information.

In general I  have the feeling that the code became worse and less clear after resolving the shadow variable cppcheck error rather than improving the quality and readability.

src/webots/gui/WbSimulationView.cpp
```diff
@@ -642,9 +642,8 @@ void WbSimulationView::writeScreenshot() {
   const QImage &image = mView3D->grabWindowBufferNow();
   disconnect(mView3D, &WbView3D::screenshotReady, this, &WbSimulationView::writeScreenshot);
 
-  QString filename;
   while (!mScreenshotFileNameList.isEmpty() && !mScreenshotQualityList.isEmpty()) {
-    filename = mScreenshotFileNameList.takeFirst();
+    QString filename = mScreenshotFileNameList.takeFirst();
```
stefaniapedrazzi, line 646:
```suggestion
    const QString filename = mScreenshotFileNameList.takeFirst();
```

src/webots/gui/WbGuidedTour.cpp
```diff
@@ -172,10 +172,9 @@ void WbGuidedTour::setTitleText(const QString &title) {
 static QString formatInfo(const WbMFString &info) {
   QString outputText;
   QString item;
-  QString lowerCaseItem;
   for (int i = 0; i < info.size(); i++) {
     item = info.item(i);
-    lowerCaseItem = item.toLower();
+    QString lowerCaseItem = item.toLower();
```
stefaniapedrazzi, line 177:
```suggestion
    const QString lowerCaseItem = item.toLower();
```

src/webots/scene_tree/WbInsertExternProtoDialog.hpp
```diff
@@ -39,13 +39,15 @@ private slots:
   void updateSelection();
 
 private:
-  bool mRetrievalTriggered;
+  // cppcheck-suppress unsafeClassCanLeak
```
omichel, line 42:
```suggestion
  // cppcheck 2.10 seems to give a series of false positives here that we want to silence:
  // cppcheck-suppress unsafeClassCanLeak
```

stefaniapedrazzi (commented):
I have the feeling that the code clarity is often worse than before after resolving the shadows variable cppcheck error for example when replacing the meaningful variable names by abbreviations when loop is more than a couple of lines long or adding long suffixes.
Not always is easy to find a good name.

src/webots/gui/WbRobotWindow.cpp
```diff
@@ -57,7 +57,7 @@ void WbRobotWindow::setClientID(const QString &clientID, const QString &robotNam
 
 bool WbRobotWindow::openOnWebBrowser(const QString &url, const QString &program, const bool newBrowserWindow) {
   QString systemProgram;
-  QStringList arguments;
+  QStringList arguments = QStringList();
```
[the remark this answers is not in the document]
ygoumaz: cppcheck gives an UnInitializedVariable error without this assignement.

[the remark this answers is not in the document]
stefaniapedrazzi: Yes, but it is a wrong error.
It is useless to assign a variable to the same default value.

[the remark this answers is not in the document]
stefaniapedrazzi: I saw another similar assignment to default value in this branch but I cannot find it again.

src/webots/gui/WbVideoRecorderDialog.cpp
stefaniapedrazzi: Personally, I think that the new variable names chosen by @ygoumaz are a better option in this particular case.

Other abbreviations that were used but didn't really improve the readability of the code where for examples `sc` meaning solid children, `tb` meaning toolbar, etc. 
If the definition is very close to the instruction it uses it, then it is fine, but this is not always the case in this PR.

ygoumaz: I think I agree with @stefaniapedrazzi on the proximity of the declaration and where it is used. Abbreviations should be used if the declaration is close to where the variable is used. In other cases, I think it would be more convenient to have a longer variable name (even a little redundant), like `name` becomes `nameString` or `nameArray`.

stefaniapedrazzi (approved):
Thank you!
The changes are fine for me.

omichel (approved):
It looks to me as well.
However, I just activated the test suite to check if we didn't break it with some functional change.

omichel (approved):
Thank you.

PR #5862, opened 2023-02-09
Add missing space to cppcheck command line.

brettle:
**Description**
Add missing space to separate options on cppcheck command line. I'm wondering if this is related to some of the test-sources errors I'm seeing in CI.

PR #5865, opened 2023-02-09
Pin cppcheck version at 2.10 for windows and macos

brettle:
This might be an alternative to PR #5863 for the master branch.


brettle:
@ygoumaz, this is an alternative to PR #5863, just for master. It:

* Suppresses cppcheck of projects because they weren't really being checked prior to to PR #5862. Some of the cppcheck errors we're now seeing are caused by those checks now being run. This can be reenabled and the associated cppcheck errors addressed in a separate PR targeting master.
* Explicitly removes the --std=c++03 cppcheck option which was effectively missing prior to PR #5862.
* Pins the cppcheck version at 2.9 for windows to avoid the rest of the errors on that platform.
* Skips source testing entirely for macos entirely because brew doesn't have a 2.9 version of cppcheck.

Alternatives to skipping all source testing on macos include just skipping cppcheck on macos or trying to build cppcheck 2.9 from source during CI.

Regardless, I'm proposing merging something like this into master and then after it gets merged into develop, revert these changes in develop (i.e. use the latest cppcheck on all platforms) and apply the fixes you have in PR #5863 to develop.

Thoughts?

brettle:
Update: I have macos using cppcheck 2.9 now.

ygoumaz:
In my opinion it's not worth going to all that much trouble to force the version of cpp-check and avoid checking the code in part of the repo (the `projects` folder in this case).

Especially since if we will have to make the changes in develop anyway, the amount of work that needs to be done now won't change. Making changes to master (and potentially introducing new bugs) is not necessarily a problem, as master is necessarily unstable between releases. Develop is rather reserved for new features.

Changes made to master will be backported to develop right away, which will keep both branches up to date and avoid having to revert changes made in this PR at the time of the next release.

Therefore I suggest we continue with the implementation of #5863 instead.

brettle:
Would you at least consider merging this PR as a stop-gap until PR #5863 is ready. As it stands, no other PRs can pass CI. This PR just gets CI back to where it was before the new version of cppcheck was released.

ygoumaz:
#5863 is almost complete. It should be merged today or tomorrow. I don't think it is worth merging this one now. We can simply by-pass the tests in case sources are failing. All the tests (including the `projects` directory) will be fixed in #5863.

ygoumaz:
#5863 is now merged so I think we can close this PR. Thanks for the help though!

brettle:
Reopening to pin it at 2.10 so that future cppcheck releases aren't as disruptive.

ygoumaz:
I don't think it's a good idea to pin the version now. That way we won't see future updates that might discover new problems in the code.

I would rather make the following compromise: leave it as it is and wait until a new version is available. I agree that this time it took almost a week to fix the bugs and it broke other PRs tests. So next time, we will pin the version directly, wait for the problems to be fixed, and then unpin (which is the original purpose of this PR #5865).

brettle:
If you want to see future update, subscribe to this feed: https://sourceforge.net/p/cppcheck/news/

brettle:
If you don't want to subscribe to that news feed, there are other ways to automate testing with the latest version that don't reduce the reproducability of CI by generating spontaneous errors in required workflows. Options include:

Option 1: Create a separate "use-latest-cppcheck-version" branch that does not include this PR and periodically merges from master (or develop) and tests the sources.

Option 2: Create a copy of the `tests_sources.yml` that does not include this PR and is not a required workflow but runs periodically.

Option 3: Create a new scheduled non-required workflow that runs the existing `test_sources.yml` workflow with a parameter that tells it to not pin the cppcheck version.

ygoumaz:
@omichel any thoughts? Probably option 2 is the best compromise as it is very easy to implement and would detect new problems when cpp-check is updated.

omichel:
Yes, option 2 seems the best to me as well.

brettle:
Option 2 has been implemented.

.github/workflows/tests_sources.yml
```diff
@@ -40,9 +40,9 @@ jobs:
           - os: ubuntu-20.04
             DEPENDENCIES_INSTALLATION: "sudo apt -y install clang-format-10 cppcheck"
```
omichel, line 41: It seems we are not enforcing cppcheck version 2.10 on Linux, is that on purpose?

brettle: I don't object to trying to do that but I'm not running Ubuntu so wasn't immediately familiar with how to do it and also it seemed unlikely that Ubuntu would upgrade the cppcheck version between Ubuntu releases.

omichel (approved):
Thank you for the explanations.

PR #5867, opened 2023-02-09
Fix delete physic

BenjaminDeleze:
In the kukubox example (https://webots.cloud/run?version=R2023b&url=https://github.com/cyberbotics/webots/blob/doc-add-robots/projects/robots/kuka/youbot/protos/KukaBox.proto&type=undefined) we can delete and reset the physic node.

However this have no influence over WebotsJS as the physic node is just ignored by it.
So it results no new node and we need to check it before attempting to `delete` or `finalize` it.


ad-daniel (approved):
Thank you

PR #5868, opened 2023-02-09
Fix cppcheck 2.9 errors in projects

brettle:
**Description**
Prior to PR #5862 projects were not being checked with cppcheck. This PR both ensures that a similar bug won't cause a similar result in the future, and fixes the cppcheck 2.9 errors that were latent in the projects (by applying only the project fixes from PR #5863) . Since it only touches projects (and `test_cppcheck.py`) it should be safe to apply to master.

If this PR is merged, the temporary disabling of cppchecking of projects that is in PR #5865 will no longer be needed. Either that PR could be updated, or if it gets merged first, a new PR can be created to reenable cppchecking of projects.


brettle (commented):
The test-sources workflow should pass once PR #5865 or PR #5863 is merged.

brettle:
Moot with the merge of PR #5863.

PR #5869, opened 2023-02-09
Hotfix-pin-cppcheck-version-on-macos-and-windows

brettle:
Force use of cppcheck 2.9 on macos and windows.


brettle:
Merged into PR #5865.

PR #5877, opened 2023-02-10
Fix MFNode insertion in web PROTO

ad-daniel:
**Description**

Nodes inserted in a MFNode were being inserted at the root level rather than respecting the IS chain link


BenjaminDeleze (approved):
Works nicely thank you!
(Strange I am sure that I had approved this one on Friday)

PR #5878, opened 2023-02-10
Add BaseNode support to web PROTO

ad-daniel:
**Description**

- [x] Handle boundingObjects (and all variants like `Box {}`, `Transform { Box {} }`, `Group { Shape { Box {} } }`, ...) 
- [x] Show compatible BaseNodes in node selector window
   - [x] Physics
   - [x] Damping
   - [x] ImmersionProperties
   - [x] BoundingObject
   - [x] Appearance/PBRAppearance
   - [x] Geometries
   - [ ] ??
- [x] Add node info (icon + description) when a base-node is selected in the selector window
- [x] Add hasShapeAncestor method for more complex node structures (ex: to a Solid children field add a Shape node, then a PBRAppearance, then an ImageTexture => the addition of the ImageTexture needs to trigger a unfinalize/finalize of the Shape)
- [x] Cleanup 

https://testing.webots.cloud/run?version=R2023b&url=https://github.com/ad-daniel/prefab/blob/master/protoExamples/protos/SolidWithAShape.proto

https://testing.webots.cloud/run?version=R2023a&url=https://github.com/BenjaminDeleze/sandbox/blob/master/protoExample/protos/OfficeChair.proto


BenjaminDeleze:
This PR will break the display of some protos, namely: `teapot.proto` and ` TorusBoundingObject.proto`.
It is because they are either geometry or bounding objects.

Note: I think this PR is doing the right thing and should not be modified to support again those protos. However, once this PR is merged we should thing on how we want to display pure BO or geometry on webots.cloud

ad-daniel:
where should I upload the files now that feature-web-proto was merged?

ad-daniel:
@BenjaminDeleze

BenjaminDeleze:
You can put them on testing such that we can test on testing.webots.cloud

ad-daniel:
note: testing.webots.proto doesn't have much in terms of PROTO, that's why when the node selection window is opened very few are shown.

resources/web/wwi/NodeSelectorWindow.js
```diff
@@ -173,8 +193,23 @@ export default class NodeSelectorWindow {
...
 
-    const compatibleNodes = [];
-    for (const [name, info] of this.nodes) {
+    const compatibleBaseNodes = [];
+    for (const [name, info] of this.baseNodes) {
+      // filter nodes based on restrictions
+      if (!this.doFieldRestrictionsAllowNode(name))
+        continue;
+
+      // don't display PROTO nodes which have been filtered-out by the user's "filter" widget
```
BenjaminDeleze, line 202:
```suggestion
      // don't display base nodes nodes which have been filtered-out by the user's "filter" widget
```

resources/web/wwi/protoVisualizer/test_proto_viewer.html
```diff
@@ -25,5 +25,5 @@
 
   const webotsView = new WebotsView();
   document.getElementById('webots-container').appendChild(webotsView);
-  webotsView.loadProto("https://raw.githubusercontent.com/cyberbotics/webots/released/projects/robots/softbank/nao/protos/Nao.proto")
+  webotsView.loadProto("https://raw.githubusercontent.com/ad-daniel/prefab/master/protoExamples/protos/SolidWithAShape.proto")
```
BenjaminDeleze, line 28: I think we should revert that

BenjaminDeleze (changes requested):
I played a bit with it and it is really impressive!

So far I encounter two errors:

- When I try to add a textureTransform to an appearance Proto (grass for example):
```
The parent field should be defined when attempting a node insertion.
```

- When resetting a MfField inside a base node:
```
tokenizer.rewind is not a function
``` 
and after that the `+` button does not work.

- Nothing can be inserted in `rotatingHead` of `Lidar` (not directly related with this PR)

ad-daniel:
I think I've addressed all the issues, including the Lidar one

BenjaminDeleze (approved):
Thank you!

PR #5885, opened 2023-02-13
Prepend libController path to environment

ygoumaz:
Currently the libController path is appended at the end of the existing `LD_LIBRARY_PATH`, `DYLD_LIBRARY_PATH` and `Path` (respectively on Linux, macOS and Windows). This can lead to the wrong libController being used by Webots.


omichel (changes requested):
Thank you.
I would add a line in the change log as this may change the behavior of Webots.

omichel (approved):
Thank you.

Issue #5842, opened 2023-02-06
Displays in extension slots cause warnings


ad-daniel:
**Describe the Bug**

1. Open the `jetbot.wbt` world
2. Add a `Display` node to the `extensionSlot`
3. Save & reload, a warning is printed 
```
WARNING: DEF JETBOT JetBot (PROTO) > DEF BODY Transform > DEF DISPLAY Transform > Display: 'name' field value should be unique: 'display' already used by a sibling Solid node.
WARNING: DEF JETBOT JetBot (PROTO) > Display: 'name' field value should be unique: 'display' already used by a sibling Solid node.
```
4. Two display entries are also shown under `Overlay > display devices`

Although there is only one instance of the node, multiple are detected. I supposed because it's a PROTO parameter.

stefaniapedrazzi:
The error is correct.
The problem is not that there is multiple instances of the node, but that the JetBot PROTO already contains in body a Display device called "display" and both nodes have the same Solid ancestor.
When searching for siblings Transform and Group nodes are expanded but "ignored" as parent. 

However, the error should be printed just after inserting the new Display in the `extensionSlot` and not only after reloading.

PR #5779, opened 2023-01-25
Refactor node and PROTO creation

stefaniapedrazzi:
Simplify the creation of nodes by removing various global flags and always correctly redirecting fields to parameters.

TODO:
* [x] all the tests should successfully pass
* [x] check distributed worlds run correctly
* [x] check #3979 is fixed
* [x] evaluate changes in loading performances and minimize them.
* evaluate if it is necessary to remove intermediate connection between the visible PROTO parameter and the finalized instance for performance reasons (_edit: moved to a separate PR_)
* [x] after template PROTO regeneration it could occur that multiple nodes had the same unique id (for example in the test `template_nested_slot_container.wbt`): this was apparently not causing issues before but in this branch the wrong node was retrieved. This should now be fixed (71cfd7) in this PR.

PR #5886, opened 2023-02-13
Fix issues resolving Solid name clash

### LINKED ISSUE #5874: Crash on inserting the same solid twice
**Describe the Bug**
When inserting a Solid Proto with the same name as another Solid Proto that is already present, if its trigger the proto regeneration, Webots will crash.

The problem seems to come from #5779 

**Steps to Reproduce**
1. Open `Panda.wbt` on develop
2. Click on endEffectorSlot of the `Panda`
3. Add a `PandaHand` to it (one is already present by default)
4. Webots crash

**Expected behavior**
Webots should not crash and add the other PandaHand.

**Additional context**

- The crash happens in `WbSolid::resolveNameClashIfNeeded`. It seems to happen because `mSolidChildren` is corrupted in some way: I tried to query the size of `mSolidChildren` and just before the crash it goes crazy like 3612067496776
- I tried to comment some part of the code, and the crash was still occurring but not always at the same place, so I think that there is a memory leak or something along those line.
- Before #5779 I could insert the second PandaHand and I would get a warning stating that two solids have the same name. Now it seems like Webots is trying to rename the second one in `panda hand(1)`.
- In `Panda.proto`: 
```
endPoint DEF panda_link7 Solid {
  translation 0.088 0 0
  rotation 0.8614190923211578 -0.3591359294777342 0.3591358121139644 1.7194199439154274
  children [
    CadShape {
      url [
         "meshes/link7.dae"
       ]
    }
    %< if (fields.endEffectorSlot.value.length > 0) { >%
    DEF panda_link8 Transform {
      translation 0 0 0.107
      children IS endEffectorSlot
    }
    %< } >%
    ]
    name "panda_link7"
```
If I remove the templating (line beginning with `%<`), I got neiter a warning nor a crash nor a rename.

stefaniapedrazzi:
Fix #5874 and #5842:
* crash occurs because fields/nodes causing PROTO regeneration are not correctly detected
* fix name clash not resolved (name is not updated) when inserting a PROTO parameter
* fix error in WbNode::fullPath overwriting the path (introduced in #5779)


src/webots/nodes/utils/WbTemplateManager.cpp
```diff
@@ -101,17 +101,20 @@ void WbTemplateManager::subscribe(WbNode *node, bool subscribedDescendant) {
...
+  if (subscribedDescendant)
+    mNodesSubscribedForRegeneration.insert(node);
   recursiveFieldSubscribeToRegenerateNode(node, subscribed, subscribedDescendant);
+  connect(node, &QObject::destroyed, this, &WbTemplateManager::unsubscribe, Qt::UniqueConnection);
 }
 
 void WbTemplateManager::unsubscribe(QObject *node) {
-  disconnect(static_cast<WbNode *>(node), &WbNode::regenerationRequired, this, &WbTemplateManager::nodeNeedRegeneration);
-  mTemplates.removeAll(static_cast<WbNode *>(node));
+  WbNode *n = static_cast<WbNode *>(node);
```
BenjaminDeleze, line 114:
```suggestion
  const WbNode *n = static_cast<WbNode *>(node);
```

Would it be possible to add a `const` here?

BenjaminDeleze (commented):
Just a question, otherwise everything seems good to me

BenjaminDeleze:
> status: the fix has to be adjusted because for PROTO nodes often the translation and rotation handles are no longer visible in this branch

What is the status of this?

I remarked that in `panda.wbt` the handles are now missing for the `PandaHand.proto`. They are present in master and develop but trying to move them result in a crash.

stefaniapedrazzi:
The problem with the handles was fixed.
But I will check the Panda case.

stefaniapedrazzi:
I just checked the PandaHand and it is correct that the handles are not shown.
Currently (also in master and develop but was buggy) the rule is that the handles are disabled if the node is included in a PROTO regeneration field.

The reason is that given that we don't know if the translation/rotation fields are changing the structure of the PROTO, changing translation and rotation with the handles triggers continuous PROTO regeneration while dragging and this causes the crashes or (if this would be addressed differently) very slow reaction.

Unrelated to this PR we could think about implementing a different approach and better identify if the translation/rotation fields (or in general PROTO parameter node's field) should trigger the regeneration.
Note that in this branch I didn't modify the regeneration rules, but just improved the method inspecting if a node or field change is triggering it.

BenjaminDeleze (approved):
Thank you!

PR #5887, opened 2023-02-14
Make supervisor_export_image test image content.

brettle:
Use exported image as a texture and do a basic check for correctness.

I'm also experiencing some strange behavior with image export on my local machine and this is a way to figure out whether it also happens during CI.


stefaniapedrazzi (commented):
It is good to also test the content of the exported image.
Camera image colors might slightly differ due to the differences in the graphics card.
The test is still failing because of the move handles that are displayed for the selected object.
The selection should be disabled from the supervisor by selecting for example the Viewpoint or the WorldInfo node or it could be set in the WBPROJ file (that in this particular case will have to be committed to the repo).
In the WBPROJ file you simply have to add the line:
```
userInteractions: selectionDisabled
```

Meanwhile, I have a couple of comments about the coding style.

brettle:
Thanks for the feedback.

Why would color differ by graphics card in this case? For the image I'm exporting, I'm using a material that emits the pure color and for the image grabbed by the camera, I'm using such a bright light (with no attenuation) that the image should be saturated. If it's not, I can increase the light intensity until it is.

brettle:
Any idea why this is still failing on ubuntu-20.04? The exported image which is supposed be green is actually the color of the sky. I added additional planes to completely enclose the viewpoint (in colors that aren't the blue of the sky) and I changed the workflow to make the `tests/` directory available as an artifact if the test suite fails. The `GREEN.png` file from that artifact should be green but is this instead:

GREEN

tests/with_rendering/controllers/supervisor_export_image/supervisor_export_image.c
```diff
@@ -0,0 +1,105 @@
...
+#include "../../../lib/ts_assertion.h"
+#include "../../../lib/ts_utils.h"
+
+#include <math.h>
+#include <stdio.h>
+#include <unistd.h>
+
+#define FILE_NAME_PNG "image.png"
+#define FILE_NAME_JPG "image.jpg"
+
```
omichel, line 12: This doesn't seem to be used.
```suggestion
```

tests/with_rendering/controllers/supervisor_export_image/supervisor_export_image.c
```diff
@@ -0,0 +1,105 @@
...
+bool file_exists(const char *filename) {
```
omichel, line 13:
```suggestion
static bool file_exists(const char *filename) {
```

tests/with_rendering/controllers/supervisor_export_image/supervisor_export_image.c
```diff
@@ -0,0 +1,105 @@
...
+  remove(file_name_png);
+  remove(file_name_jpg);
+
+  ts_assert_boolean_not_equal(file_exists(file_name_png), "Impossible to remove the PNG file");
+  ts_assert_boolean_not_equal(file_exists(file_name_jpg), "Impossible to remove the JPG file");
+
+  // Turn to face the requested direction
+  WbFieldRef rotation_field = wb_supervisor_node_get_field(wb_supervisor_node_get_from_def("Test"), "rotation");
+  double rotation[4] = {0.0, 0.0, 1.0, 0.0};
+  rotation[3] = angle_radians;
```
omichel, line 42:
```suggestion
  double rotation[4] = {0.0, 0.0, 1.0, angle_radians};
```

tests/with_rendering/worlds/supervisor_export_image.wbt
```diff
@@ -0,0 +1,92 @@
...
+Background {
+  skyColor [
+    0.4 0.7 1
+  ]
+}
+PointLight {
+  ambientIntensity 1
+  intensity 10
+}
+DEF Test Robot {
```
omichel, line 24: We generally use all caps for DEF names.
```suggestion
DEF TEST Robot {
```

tests/with_rendering/controllers/supervisor_export_image/supervisor_export_image.c
```diff
@@ -0,0 +1,105 @@
...
+  sprintf(file_name_jpg, "%s.jpg", color_def);
+
+  remove(file_name_png);
+  remove(file_name_jpg);
+
+  ts_assert_boolean_not_equal(file_exists(file_name_png), "Impossible to remove the PNG file");
+  ts_assert_boolean_not_equal(file_exists(file_name_jpg), "Impossible to remove the JPG file");
+
+  // Turn to face the requested direction
+  WbFieldRef rotation_field = wb_supervisor_node_get_field(wb_supervisor_node_get_from_def("Test"), "rotation");
```
omichel, line 40:
```suggestion
  WbFieldRef rotation_field = wb_supervisor_node_get_field(wb_supervisor_node_get_from_def("TEST"), "rotation");
```

omichel (changes requested):

omichel (approved):
Thank you.

Issue #5332, opened 2022-10-17
Device activation keep it previous state even after reset


ShuffleWire:
Disclaimer: I'm presenting a behavior that should not be encountered very often, and moreover by using a very special setup... But it still show some inconsistencies that could be addressed.

**Describe the Bug**
NB: The following should extend to any sort of Device, not only Lidar.
I'm using costly Lidar (in term of computing power), and they are not always used. Hence I'm actively toggling them on/off during world execution to save time.
My controller assume that if nothing have been done, the Lidar was deactivated, hence saving comuting ressources. And later, I could activate it when needed.
It work actually that way, but there is a caveat : Here is how : 

**Steps to Reproduce**
1. Open world with a Lidar, and a controller which activate the lidar.
2. Run the controller, lidar work fine
3. Change the controller to not activate nor deactivate the Lidar
4. (in a case of an external controller) Restart the controller, without resetting the simulation : the lidar is still running. It seems ok, as long as we consider that we just "resume" the simulation where we left it. But the controller start with an enabled device...
5. Reset the world, re run the world and the controller, and still the lidar is activated ! That don't seems right, because we gone back to the initial state of the simulation. It could potentially to non reproducible simulation, between reset.
6. Reload the world, re run and see that the lidar is deactivated -> that is correct, but not consistent with the reset behavior

**Expected behavior**
I feel like that I least the reset and the reload should behave the same, and ensure that starting condition are the same each time.

**Current fix**
For the moment I just start by enabling or disabling the Lidar, to ensure the activation status.
I guess it should be the correct way to do it, but I find the reset behavior disturbing, however. Or at least add some lines in the documentation, just to be sure...

To make it more persistent, either a reset should reset activation status to false (IMO the best solution), either call `wb_***_disable()` in each `wb_robot_get_device()` (but could add many code...)

Thanks

ShuffleWire:
It's seems to be fixed in 2023b, but I don't know how...
If anyone have a idea before closing, I would be happy :)

omichel:
It seems it was fixed in #5889.

CoolSpy3:
It looks like this is a thing in R2025a? Although, based on [this comment](https://github.com/cyberbotics/webots/pull/5889#issuecomment-1429632238), it might not actually be an issue. Thoughts on if we should still look into this?

omichel:
[...]

PR #5889, opened 2023-02-14
Connector: Fix status after reset

ShuffleWire:
After a reset, the value returned by `wb_connector_is_locked` was not updated, because the controller
didn't received new value.
I can't emphasis enough the fact that I'm asking for help to assess this fix, and moreover ensure that this bug is not present in others node. This could be a pattern issue, and might have been reproduced elsewhere.
However, it's seems that is a specific bug of the Connector : it's the only one to make use of both `mNeedToReconfigure` and a custom `reset()` function. I do think it's because contrary to others node that are pure sensor, this one actively impact the simulation, hence the code pattern change a bit. Am I right ?

**Description**
Ensure that after a reset the correct status is sent to a controller. 

**Related Issues**
This pull-request fixes issue https://github.com/cyberbotics/webots/issues/5890

**Tasks**
  - [ ] Ensure that it's the correct fix
  - [ ] Ensure that the other nodes don't have this issue, and fix them accordingly

### DIFF
--- a/docs/reference/changelog-r2023.md
+++ b/docs/reference/changelog-r2023.md
@@ -11,6 +11,7 @@ Released on ??
     - Improved plot representation in default robot window when a NaN value is received from a device ([#5680](https://github.com/cyberbotics/webots/pull/5680)).
     - Improved default selected tab in Field Editor when nodes are selected ([#5726](https://github.com/cyberbotics/webots/pull/5726)).
   - Bug Fixes
+    - Fix the behavior of the [Connector](connector.md) after a reset to return to the controller the correct status ([#5889](https://github.com/cyberbotics/webots/pull/5889))
     - Fixed redirection of stdout/stderr for Python controllers on Windows ([#5807](https://github.com/cyberbotics/webots/pull/5807)).
     - Fixed crash in Python API when a robot controller was using several cameras with different resolutions ([#5705](https://github.com/cyberbotics/webots/pull/5705)).
     - Fixed Python API `Supervisor.setSimulationMode` which was failing ([#5603](https://github.com/cyberbotics/webots/pull/5603)).
--- a/docs/reference/supervisor.md
+++ b/docs/reference/supervisor.md
@@ -2557,6 +2557,13 @@ The following table summarizes the behavior of different reset functions:
       <td>N/A</td>
     </tr>
     <tr>
+      <td><strong><a href="connector.md">Connector</a></strong></td>
+      <td>Resets</td>
+      <td colspan=3>Resets the <code>isLocked</code> field</td>
+      <td>N/A</td>
+      <td>N/A</td>
+    </tr>
+    <tr>
       <td><strong><a href="display.md">Display</a></strong></td>
       <td>Resets</td>
       <td>Clears</td>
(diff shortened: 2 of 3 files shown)


stefaniapedrazzi:
After adding this functionality to the reset, the documentation should be updated:
https://github.com/cyberbotics/webots/blob/develop/docs/reference/supervisor.md#resetreload-matrix

ShuffleWire:
> After adding this functionality to the reset, the documentation should be updated: https://github.com/cyberbotics/webots/blob/develop/docs/reference/supervisor.md#resetreload-matrix

I'm not sure if it's really necessary :/ what the fix does is just to revert to the correct behavior of getting things fully reset. Moreover I'm not actually changing any behavior of this node listed in this matrix, as I'm changing only the way it interact with the controller. For the actual simulation it should not make any difference.

> And an entry should be added in the ChangeLog: https://github.com/cyberbotics/webots/blob/develop/docs/reference/changelog-r2023.md

 done !

stefaniapedrazzi:
Yes, this is a bug fix and nothing should be added in the table.
So then it should target the master branch instead of the develop branch given that this fix doesn't break the compatibility.

ShuffleWire:
> it should target the master branch instead of the develop branch given that this fix doesn't break the compatibility.
Cleanly force-pushed to rebase on master

ShuffleWire:
Just as a side note, could it help with some issues I've seen about not reproducible run ? Maybe (imo) it could be worth to check if those case were impacted by this bug (ie they use a Connector and act differently based on the locked state returned by the controller)

stefaniapedrazzi:
The fix is fine.
We should wait before the source tests are fixed before merging.

The reset functionality is limited with respect to the reload one and it doesn't resets all the devices status.
The list of what is reset can be found [here](https://github.com/cyberbotics/webots/blob/develop/docs/reference/supervisor.md#resetreload-matrix).
So rather than a bug, what is described in https://github.com/cyberbotics/webots/issues/5332 is more a missing functionality of the reset action.

ShuffleWire:
To be fair, I was expecting that the reload would bring the simulation to a state from which the simulation would be reproducible. Basically, if the user **have changed nothing** in the world, I would expect that the reset would be equivalent to the reload, in term of reproducibility of (the starting point of) the simulation.

To be clear, my use case is that I've a robot and an external controller that I'm working on, so I'm often restarting the simulation. For the most part I don't change the simulation (otherwise, yes, simulation could obviously not be reproducible), but I'm just restarting/running/stopping the controller/restarting etc over and over... And I would expect that the initial state of the simulation would be the same each run.

The reload functionality is fine, but be able to reset make it so much quicker, especially with world with many heavy object (in term of file size).

But whatever ! The fix was necessary, thank for the documentation update, it's just that simulation reproducibility is very important for my work, and webots got mostly it right. It would be sad to let it go on some edges case :)

omichel:
I agree with you @ShuffleWire that we should try to make the reset functionality as close as possible to the reload functionality. Although, this is the ultimate goal, it is not easy to achieve. But your contribution is helping going towards this objective. Nevertheless, meanwhile we have to document what are the differences between reset and reload.

stefaniapedrazzi (approved):

PR #5891, opened 2023-02-14
WebotsJS: fix web mf

BenjaminDeleze:
- [x] Fix MfVector2/3 and rotation. It was broken when we removed the `x: y: z:`
- [x] Fix MFNode reset. It was not offsetting the following rows if the default was not null.

Test it here: https://testing.webots.cloud/run?version=R2023b&url=https://github.com/cyberbotics/webots/blob/doc-add-robots/projects/robots/pal_robotics/tiagopp/protos/Tiagopp.proto&type=undefined


ad-daniel (approved):
Thank you

PR #5894, opened 2023-02-14
WebotsJS: fix parameters bugs

BenjaminDeleze:
**Description**

- Fix viewpoint when there is no boundingSphere
- Fix wrong defaultValues in jointWithBacklash MFNode
- Fix `toJS` method when the node is a baseNode and not a proto

Test it here: https://testing.webots.cloud/run?version=R2023b&url=https://github.com/cyberbotics/webots/blob/fix-web-joint-with-backlash/projects/joints/protos/Hinge2JointWithBacklash.proto&type=undefined

Note that if you try to remove one of the jointsParameters it will break the proto. However it comes from the proto construction.

### DIFF
--- a/resources/web/wwi/nodes/utils/WbBoundingSphere.js
+++ b/resources/web/wwi/nodes/utils/WbBoundingSphere.js
@@ -1,5 +1,6 @@
 import WbVector3 from './WbVector3.js';
 import {findUpperTransform} from './node_utilities.js';
+import WbWorld from '../WbWorld.js';
 
 export default class WbBoundingSphere {
   #center;
@@ -45,6 +46,21 @@ export default class WbBoundingSphere {
     this.boundSpaceDirty = true;
     this.parentCoordinatesDirty = true;
     this.parentUpdateNotification();
+  }
+
+  static addSubBoundingSphereToParentNode(node) {
+    let parentId = node.parent;
+    while (parentId) {
+      const parent = WbWorld.instance.nodes.get(parentId);
+      if (!parent)
+        return;
+
+      if (parent.boundingSphere()) {
+        parent.boundingSphere().addSubBoundingSphere(node.boundingSphere());
+        return;
+      }
+      parentId = parent.parent;
+    }
   }
 
   centerInParentCoordinates() {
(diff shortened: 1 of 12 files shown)


BenjaminDeleze:
Fix a bug where two `ifs` have been converted to an `if, else`

ad-daniel (changes requested):
Thank you
When I add a robot to the endPoint (say Crazyflie), and switch to the Devices tab, several errors related to the bounding sphere are thrown when hovering above some of the devices (gps, gyro, ...)

Similarly, in the Joints tab, everything seems to be working except the root joint (for crazyflie is crazyflie, for nao is nao)

BenjaminDeleze:
Okay I have lost some hairs with those ones:
I found several more bugs:
  - [x] I was calling ._boundingSphere directly instead of the .boundingSphere function. And the ._boundingSphere is not present in all objects
  - [x] When adding objects, the boundingSphere was not recalculated correctly. The solution I found is to recall postFinalize in order to trigger computation.
    It seems a bit overkill, but we will do it only once for each device so I think it is acceptable. I tried to use a modified version of the `recomputeBoundingSphere` function but it is a bit of a mess as it does not exist in all the classes.
  - [x] Some joints were not working because `_endPointZeroRotation` and `_endPointZeroTranslation` were not defined in the case when no endPoint was present at the node creation.
  - [ ] The anchor seems to "jump" in some special case when we change the endPoint.

resources/web/wwi/nodes/utils/WbBoundingSphere.js
```diff
@@ -1,5 +1,6 @@
 import WbVector3 from './WbVector3.js';
 import {findUpperTransform} from './node_utilities.js';
+import WbWorld from '../WbWorld.js';
```
ad-daniel, line 3:
```suggestion
```

ad-daniel: doesn't appear to be used

ad-daniel (changes requested):
Nice, seems to work well, thank you!

ad-daniel (approved):
Thank you

PR #5896, opened 2023-02-14
Add version comparison between libController and Webots

### LINKED ISSUE #3505: Add version check in libController
Following-up #3429, it would be nice to add a version check preventing from running controllers compiled with the wrong version of libController (as it is the case for many common libraries). This would certainly avoid some bug reports about misbehaving controllers.
The Webots version should be included in the compiled controllers and checked against the libController version. If the version do not match, an error message should be displayed, asking users to recompile their controller programs with the current version of the libController and the controller should terminate.

ygoumaz:
Resolves #3505.

With extern controllers (namely TCP ones) and the `webots_ros2` package it happens that the user uses different versions of Webots and of the libController, which is forbidden, as both could be incompatible with each other.

Webots is now sending its version in the configure message of the robot. The version is compared on the libController side and a warning is displayed if they are not equal.

### DIFF
--- a/docs/reference/changelog-r2023.md
+++ b/docs/reference/changelog-r2023.md
@@ -4,6 +4,7 @@
 Released on ??
   - New Features
     - Added a new launcher to simplify the start of extern controllers ([#5629](https://github.com/cyberbotics/webots/pull/5629)).
+    - Added a warning when Webots and controller library versions are different ([#5896](https://github.com/cyberbotics/webots/pull/5896)).
   - Cleanup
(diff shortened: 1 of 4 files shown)


omichel (commented):
It looks good to me.
However, I would pass the version information before the device configure sequence as it may be possible that the libController will crash in the device configure if there is a version mismatch. So, the earlier we can show the warning, the better. I believe we could pass the version immediately after the `C_CONFIGURE` code.

omichel (approved):
Thank you.

PR #5899, opened 2023-02-15
Fix parent when exporting SliderJoint to URDF

### LINKED ISSUE #5866: Wrong parent link for SliderJoint in generated URDF
**Describe the Bug**
The generated URDF from a proto with a Transform->SliderJoint names the parent link of the joint after the Transform's node DEF statement (or the node type if no DEF statement is present), causing an unidentified link name error.

**Steps to Reproduce**
1. Create a robot with 2 Transforms each with a SliderJoint & HingeJoint child 
2. Define a DEF for each transform, and name the solid endPoint for each Joint
3. Export an URDF from the robot using the UI 
4. Check that the parent/children of both joints are properly named

**Screenshots**
An example output of the previous steps is shown next:
protovsurdf

**Additional context**
This behavior does not seem to happen when the joint node's parent is a solid/robot with a established `name` field

ygoumaz:
Fixes #5866.


omichel (approved):
Thank you.
Looks good to me.

PR #5900, opened 2023-02-15
Better reload/reset table

omichel:
This PR tries to make the supervisor reload/reset/etc. table easier to read:
https://cyberbotics.com/doc/reference/supervisor?version=fix-make-table-easier-to-read

### DIFF
--- a/docs/css/webots-doc.css
+++ b/docs/css/webots-doc.css
@@ -463,6 +463,7 @@ body {
   width: 100%;
 }
 .webots-doc #view table th {
+  padding: 6px;
   border-width: 1px;
   border-style: solid;
   border-color: #ddd;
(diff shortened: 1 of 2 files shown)


BenjaminDeleze (approved):
I am not totally convinced that the changes improve the readability but they are correct.

omichel:
> I am not totally convinced that the changes improve the readability but they are correct.

If they don't improve readability, we should not merge this PR.

omichel:
@ygoumaz, @ad-daniel and @stefaniapedrazzi: what do you think?

ygoumaz:
Maybe centering the text in merged cells may look better, but hard to tell without trying.

omichel:
> Maybe centering the text in merged cells may look better, but hard to tell without trying.

Done.

ygoumaz:
It is indeed better imo.

stefaniapedrazzi:
I'm not sure either that the readability is better than before.
I have the feeling that it is even more difficult to understand what is the title of the column by merging all these cells.
I find it easier in the previous version to understand the meaning of the 6 columns.

stefaniapedrazzi:
If it would be possible to add some borders between columns in the header row, then probably it will be much better.

omichel:
I just added header columns, vertical centering and colors... Is that better?

stefaniapedrazzi:
Yes, I think it is better like this.

BenjaminDeleze (approved):

PR #5901, opened 2023-02-15
Fix Python extern controllers on Windows

ygoumaz:
Extern controller launcher is broken for Python controllers on Windows.


omichel (approved):
Well spotted.

PR #5905, opened 2023-02-16
Fix Camera recognition for scaled objects

stefaniapedrazzi:
Detected object center was not computed correctly for scaled objects or in general if the Solid with recognition enabled has a parent Transform node with a non-default scale or rotation.

This is a first bug fix to the Camera recognition but there are other in case the bounding object (and not the bounding sphere) is used to compute the position.
Additionally, the Camera recognition only works for planar projections (not for spherical and cylindrical).


ad-daniel (approved):
Nice catch, thank you

PR #5906, opened 2023-02-16
Robots insertion

BenjaminDeleze:
Allow robots in joints endpoint.

While checking that, I saw that it was also not possible to insert `robot` in 

- `fastHelix` and `slowHelix` of a propeller
- `rotatingHead` of a lidar.

But I am not sure if we want to change it.
I do not really see a utility but on the other point it is strange that:
- `fastHelix` -> `robot` is wrong
- `fastHelix` -> `solid` -> `robot` is correct

What do you think @omichel @stefaniapedrazzi ?

### DIFF
--- a/src/webots/nodes/utils/WbNodeUtilities.cpp
+++ b/src/webots/nodes/utils/WbNodeUtilities.cpp
@@ -762,9 +762,8 @@ bool WbNodeUtilities::isDescendantOfBillboard(const WbNode *node) {
     return true;
 
   WbNode *n = node->parentNode();
-  WbField *field = node->parentField(true);
-  while (n && !n->isWorldRoot() && field) {
-    WbBaseNode *baseNode = dynamic_cast<WbBaseNode *>(field->parentNode());
+  while (n && !n->isWorldRoot()) {
+    WbBaseNode *baseNode = dynamic_cast<WbBaseNode *>(n);
 
     if (!baseNode)
       return false;
@@ -772,7 +771,6 @@ bool WbNodeUtilities::isDescendantOfBillboard(const WbNode *node) {
     if (baseNode->nodeType() == WB_NODE_BILLBOARD)
       return true;
 
-    field = n->parentField(true);
     n = n->parentNode();
   }
 
--- a/src/webots/nodes/utils/WbNodeUtilities.cpp
+++ b/src/webots/nodes/utils/WbNodeUtilities.cpp
@@ -136,7 +136,7 @@ namespace {
 
     // No robot can be inserted in helix of propellers.
     if (WbNodeUtilities::isRobotTypeName(nodeName) && WbNodeUtilities::isDescendantOfPropeller(node))
-        return false;
+      return false;
 
     if (dynamic_cast<const WbSlot *>(node) && (fieldName == "endPoint")) {  // add something in the endPoint field of a slot
       if (dynamic_cast<const WbSlot *>(node->parentNode())) {  // pair of slots, we can add everything that is allowed in the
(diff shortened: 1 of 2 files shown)


stefaniapedrazzi:
`fastHelix -> solid -> robot` is working just because when changing the rules to insert Robot nodes in Solid parent we didn't add this exception.
This is a particular case where slowHelix or fastHelix are shown but not at the same time.
So it seems quite difficult to have a consistent behavior if a Robot is inserted in these fields where it works only if the corresponding parent is active.
To me it seems that it is better to add prevent adding Robot nodes in these two fields.

omichel:
I agree with @stefaniapedrazzi. I don't see any use-case for this and it is really odd. For the rotating head of the lidar, we could allow it if it doesn't cause any trouble, but a robot should never be added as a children or sub-children of an helix (fast or slow).

BenjaminDeleze:
Very good point

> `fastHelix -> solid -> robot` is working just because when changing the rules to insert Robot nodes in Solid parent we didn't add this exception. This is a particular case where slowHelix or fastHelix are shown but not at the same time. So it seems quite difficult to have a consistent behavior if a Robot is inserted in these fields where it works only if the corresponding parent is active. To me it seems that it is better to add prevent adding Robot nodes in these two fields.

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -775,6 +779,36 @@ bool WbNodeUtilities::isDescendantOfBillboard(const WbNode *node) {
...
+
+    if (baseNode->nodeType() == WB_NODE_PROPELLER)
+      return true;
+
+    field = n->parentField(true);
+    n = n->parentNode();
+  }
+
+  return false;
+}
```
stefaniapedrazzi, line 810: I see that this is copied from the `isDescendantOfBillboard` function.
But why do you need to also use the `WbNode::parentField()` (that is much more costly than just calling `WbNode::parentNode()`)?

stefaniapedrazzi (commented):
The change seems fine.
I just have a comment about the code for checking the parent nodes type.

src/webots/nodes/utils/WbNodeUtilities.cpp
BenjaminDeleze: No idea, my guess is that I took example on `WbVrmlNodeUtilities::isFieldDescendant` to first implement `isDescendantOfBillboard`.
I removed the mention to fields in both `isDescendantOfBillboard` and `isDescendantOfPropeller`

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -758,17 +762,43 @@ bool WbNodeUtilities::isDescendantOfBillboard(const WbNode *node) {
...
+    WbBaseNode *baseNode = dynamic_cast<WbBaseNode *>(n);
+
+    if (!baseNode)
+      return false;
+
+    if (baseNode->nodeType() == WB_NODE_PROPELLER)
+      return true;
+
     n = n->parentNode();
   }
```
stefaniapedrazzi, line 803: You could also further simplify this:
```suggestion
  WbNode *n = node;
  while (n && !n->isWorldRoot()) {
    const WbBaseNode *baseNode = dynamic_cast<WbBaseNode *>(n);
    if (!baseNode)
      return false;
    if (baseNode->nodeType() == WB_NODE_PROPELLER)
      return true;
    
    n = n->parent();
  }
```

omichel (approved):
Thank you.

PR #5908, opened 2023-02-16
Fix license

BenjaminDeleze:
**Description**

- Fix the apache license url in some files: `http://www.apache.org/licenses/LICENSE-2.0` instead of `http://www.apache.org/licenses/LICENSE-2.0.` which was resulting in an invalid link
- Took the opportunity to change the urls to https


ygoumaz (approved):
Thank you!

PR #5911, opened 2023-02-17
Slot Tiago

BenjaminDeleze:
**Description**
Turn the `children` field of Tiago's extension to `slot`.

I could be used to make a nice demo video of webots.cloud.


ad-daniel:
I think the change is legitimate, but before making the demo we need to update some of the models as well first (sensors and base are different)

ad-daniel (approved):
Looks good, thank you

PR #5912, opened 2023-02-17
Update car.py inheritance

Minimerl:
According to the documentation the Car lib should inherit from the Driver lib (as seen [here](https://cyberbotics.com/doc/automobile/car-library?tab-language=python#constructor))
Since Driver inherits from Supervisor this does not seem to need additional changes


omichel (approved):
This is correct.
Well spotted!
Thank you.

PR #5914, opened 2023-02-17
WebotsJs: Fix nested slot resolution

BenjaminDeleze:
**Description**
When searching for the type of the slot we were only looking for the parent node.
However, it is possible (for example in the case of RobotisOp2), that the parent node is a proto (the Darwin-op) and so we need to check further to find the type of the slot.

Test it here: https://testing.webots.cloud/run?version=R2023b&url=https://github.com/cyberbotics/webots/blob/doc-add-robots/projects/robots/robotis/darwin-op/protos/RobotisOp2.proto&type=undefined


resources/web/wwi/NodeSelectorWindow.js
```diff
@@ -335,7 +335,23 @@ export default class NodeSelectorWindow {
...
+    while (parentNode.isProto) {
+      let found = false;
+      for (const parameter of parentNode.parameters) {
+        if (parameter[0] === name) {
+          found = true;
+          parentNode = parameter[1].aliasLinks[0].node;
+          name = parameter[1].aliasLinks[0].name;
+          break;
+        }
+      }
```
ad-daniel, line 349: do we need to iterate here? Since parameters is a map (parameterName -> parameter), can't we rely on `has/get` instead?

ad-daniel (changes requested):
Just a doubt

resources/web/wwi/NodeSelectorWindow.js
BenjaminDeleze: Good catch!

ad-daniel (approved):
Thank you!

PR #5915, opened 2023-02-17
WebotsJS: Fix devices position

BenjaminDeleze:
**Description**
The previous approach had two limitations:

- It was not working well when the boundingSphere center was under the floor
- The device "jumped" when first modifying the translation with the parameter window.

Now we are moving the floor under the object, which solved both issues

Try it out: https://testing.webots.cloud/proto


omichel (approved):
Thank you.

PR #5917, opened 2023-02-17
Avoid printing some warnings in CI

omichel:
These warnings are not useful in the CI as they always show up and we know they are harmless.


BenjaminDeleze (approved):
Thank you

PR #5922, opened 2023-02-20
Delete empty shapes folder and icon

BenjaminDeleze:
This is a remnant of the Colladashape proto that has been replaced by the `CadShape` basenode.


ad-daniel (approved):
Thank you

PR #5928, opened 2023-02-21
WebotsJS: Fix nested templateregenerator

BenjaminDeleze:
In the case of the RobotisOp2, the `backlash` parameter was not flagged as template regenerator also it should have.
It was correctly flagged in its subproto (`Darwin-op`) but the information was not transmitted to the upper level.

Test it here: https://testing.webots.cloud/run?version=R2023b&url=https://github.com/cyberbotics/webots/blob/doc-add-robots/projects/robots/robotis/darwin-op/protos/RobotisOp2.proto&type=undefined


ad-daniel (approved):
Good point, thank you!

PR #5929, opened 2023-02-21
WebotsJS: Fix default texture mapping

BenjaminDeleze:
Test it here: https://testing.webots.cloud/run?version=R2023b&url=https://github.com/cyberbotics/webots/blob/doc-add-bo/projects/devices/velodyne/protos/VelodyneHDL-64E.proto&type=undefined


ad-daniel (approved):
Thank you

PR #5934, opened 2023-02-24
Fix render on regeneration

BenjaminDeleze:
Test it here: https://webots.cloud/run?version=R2023b&url=https://github.com/cyberbotics/webots/blob/develop/projects/appearances/protos/Grass.proto&type=undefined


omichel (approved):
Nice. It seems to work well (tested on Chrome where my cache was empty).

ad-daniel:
I suspect something similar might be necessary on line 183 as well, no?

Issue #5937, opened 2023-02-24
Dangling Pointer warning


SGSSGene:
Working on the newest commit d92528e7b4c1d9b6269ee7d378a5575fbb9d9b92 (webots/develop)

While compiling, I got an Error in WbConnector.cpp line 383 about dangling pointers, and I really believe the compiler is right about it. In Line 366 there exists following code:
```
  const dReal *d1 = b1 ? dBodyGetPosition(b1) : matrix().translation().ptr();
```
The function `matrix().translation()` returns a temporary and calling `ptr()` on it will be a dangling pointer.

Here direct link to the code: https://github.com/cyberbotics/webots/blob/95dad26c8789bb854ce026c5abf91a1a6b1bc993/src/webots/nodes/WbConnector.cpp#L365-L387

SGSSGene:
I am wrong, you are right.....

PR #5938, opened 2023-02-24
refactor wbconnector::snapOrigins removing pragmas

SGSSGene:
**Description**
I refactor code, to make it easier to understand for human and compilers alike.

This removes the requirement for compiler specific suppression mechanisms.

I moved the lines `const dReal *d1 = b1 ? dBodyGetPosition(b1) : matrix().translation().ptr();` to where it is  used
and removed the trinary operator.

**Related Issues**
I opened #5937 because I believed there was a bug.

**Additional context**
This cleanup is not critical.

### DIFF
--- a/src/webots/nodes/WbConnector.cpp
+++ b/src/webots/nodes/WbConnector.cpp
@@ -362,10 +362,6 @@ void WbConnector::snapOrigins(WbConnector *other) {
   getOriginInWorldCoordinates(p1);
   other->getOriginInWorldCoordinates(p2);
 
-  // retrieve current body positions
-  const dReal *d1 = b1 ? dBodyGetPosition(b1) : matrix().translation().ptr();
-  const dReal *d2 = b2 ? dBodyGetPosition(b2) : other->matrix().translation().ptr();
-
   // each body must be shifted towards the other by half the distance
   dReal h[3] = {p2[0] - p1[0], p2[1] - p1[1], p2[2] - p1[2]};
   if (b1 && b2) {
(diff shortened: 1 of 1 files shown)


omichel:
This is not equivalent to the previous code as it doesn't handle the case where `b1` is `NULL` and `d1` is set to `matrix().translation().ptr()`. Same for `b2`.

SGSSGene:
> This is not equivalent to the previous code as it doesn't handle the case where `b1` is `NULL` and `d1` is set to `matrix().translation().ptr()`. Same for `b2`.

Setting `d1` to `matrix().translation().ptr()` is not useful, since that would be dangling pointer.

But `d1` is only accessed if `b1` is true. So the dangling pointer of `matrix().translation().ptr()` is never accessed.


In pseudo code, our "original code":
```
   const dReal *d1 = b1 ? dBodyGetPosition(b1) : matrix().translation().ptr(); 
   // ...no accesses to d1...
   if (b1)  {
     dBodySetPosition(b1, d1[0] + h[0], d1[1] + h[1], d1[2] + h[2]); 
  }
  ```
This can be transformed to:
```
   if (b1)  {
     const dReal *d1 = b1 ? dBodyGetPosition(b1) : matrix().translation().ptr(); 
     dBodySetPosition(b1, d1[0] + h[0], d1[1] + h[1], d1[2] + h[2]); 
  }
  ```
And this can be trivially transformed to:
```
   if (b1)  {
     const dReal *d1 = dBodyGetPosition(b1); 
     dBodySetPosition(b1, d1[0] + h[0], d1[1] + h[1], d1[2] + h[2]); 
  }
  ```

omichel (approved):
Doh! You are right. Sorry for overlooking this. And thank you for this great contribution.

PR #5941, opened 2023-02-27
Fix crash when inserting in node in regenerating PROTO paramter

### LINKED ISSUE #5913: Crash with darwin-Op jersey
**Describe the Bug**
This bug seems to be only on develop.

**Steps to reproduce**

- Open Webots
- Insert a `Darwin-op` or `RobotisOp2` (the normal or hinge2 version)
- Try to put a jersey in the jersey slot.
- Webots crash

**Note**
I tried to search for the commit that introduced that bug but could not find it because the but is not always reproducible:

- Most of the time it will crash, but in some worlds it may happen to work
- On some commit it was always crashing, and when I retried to be sure two hours after it was not crashing at all.

Also I tried with other protos that have slot (nao jersey,...) and they seem to work correctly

stefaniapedrazzi:
Fix #5913:
do not reconnect fields of newly inserted PROTO if the insertion trigger an ancestor node regeneration.

Additionally, also apply the same mechanism to postpone the regeneration as done in the mfnode case.

### DIFF
--- a/src/webots/nodes/utils/WbNodeOperations.cpp
+++ b/src/webots/nodes/utils/WbNodeOperations.cpp
@@ -242,11 +242,15 @@ WbNodeOperations::OperationResult WbNodeOperations::initNewNode(WbNode *newNode,
 
   } else {
     WbSFNode *const sfnode = dynamic_cast<WbSFNode *>(field->value());
-    if (sfnode)
-      sfnode->setValue(newNode);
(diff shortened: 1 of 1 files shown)


BenjaminDeleze (approved):
Thank you

PS: the test that failed seems to be unrelated with this PR, I could not reproduce it locally

PR #5944, opened 2023-02-27
Fix regeneration propagation for Web PROTO

ad-daniel:
**Description**

The regeneration flag wasn't being propagated all the way up the IS chain. In this PR we track the reverse of the aliases as well and whenever the regeneration status changes down the line, it gets propagated upwards

https://testing.webots.cloud/run?version=R2023b&url=https://github.com/ad-daniel/prefab/blob/master/protoExamples/protos/RandomBuilding.proto&type=undefined

I will address the issue with the building generating different outcomes after each regeneration as it might be a bit tricky to implement and have other things to work on before that


BenjaminDeleze (approved):
Thank you

PR #5945, opened 2023-02-27
WebotsJS: Fix robot-window when switching worlds

BenjaminDeleze:
The robot windows where not displayed anymore after switching from one world to another.

It was because the robot windows were sent before the world and, when loading a world, we reset the robot windows.
So we need to first load the world and then send the robot windows.


stefaniapedrazzi (approved):
Thank you.

Just a minor note, I saw that there are still some differences in the robot widow button behavior  between the first click and subsequent clicks:
* the first time you click on the button, the robot window is automatically displayed as well as the menu listing all the robot windows
* from the second click on only the menu listing all the robot windows is displayed but the robot window is not shown automatically

BenjaminDeleze:
> Just a minor note, I saw that there are still some differences in the robot widow button behavior between the first click and subsequent clicks:
> 
> * the first time you click on the button, the robot window is automatically displayed as well as the menu listing all the robot windows
> * from the second click on only the menu listing all the robot windows is displayed but the robot window is not shown automatically

Yes this is the normal behaviour as implemented by Thomas
The first time we open all the windows.
But then, we display only the list such that the user can choose which window he wants to open

stefaniapedrazzi:
> Yes this is the normal behaviour as implemented by Thomas
The first time we open all the windows.
But then, we display only the list such that the user can choose which window he wants to open

Ok, then it is fine.
Even if to me this seems a bug.

PR #5946, opened 2023-02-27
Fix PROTO USE info in Add Node dialog

### LINKED ISSUE #5820: Error message when inserting USE node
In a custom test world, when I insert a USE node in the scene tree from Add Node Dialog I get these errors in the console:
```
ERROR: Cannot retrieve proto info list, unknown category '-1'.
ERROR: 'PROTO_IS_1' is not a known proto in category '-1'.
```
(it would also be better to improve the message and change the case of the `proto` word to `PROTO`)

I tested on the develop branch. Not sure if this is also reproducible on the master branch.
Then, if Webots is compiled in debug mode there is also an assertion failure related to this.

Here are the files to reproduce it:
world:
```
#VRML_SIM R2023b utf8

EXTERNPROTO "../protos/PROTO_IS_1.proto"

WorldInfo {
}
Viewpoint {
  orientation -0.5389604751600234 0.49748107171499617 0.6797309684725131 2.0209260413738295
  position 0.07641206827318611 -0.5431454619486344 5.805767039237112
}
DEF A PROTO_IS_1 {
}
Transform {
  children [
  ]
}
```

`PROTO_IS_1.proto`:
```
#VRML_SIM R2023a utf8

PROTO PROTO_IS_1 [
  field SFVec3f t 0 0 4
  field SFVec3f otherParam 0.1 0.1 0.1
]
{
  Transform {
    translation IS t
    children [
      Shape {
        geometry Box {
          size IS otherParam
        }
      }
    ]
  }
}
```

stefaniapedrazzi:
Fix #5820:
in the Add Node dialog, if the selected USE node is a PROTO node it was not defined in which PROTO list the method had to search to retrieve the node information.

In case of USE nodes, the corresponding DEF node is already used by the world, so node info should be looked in the current world PROTO list.

### DIFF
--- a/src/webots/scene_tree/WbAddNodeDialog.cpp
+++ b/src/webots/scene_tree/WbAddNodeDialog.cpp
@@ -321,13 +321,13 @@ void WbAddNodeDialog::showNodeInfo(const QString &nodeFileName, NodeType nodeTyp
     // set icon path
     pixmapPath = "icons:" + fileInfo.baseName() + ".png";
   } else {
-    assert(variant > 0);
+    assert(nodeType == USE || variant > 0);
     // the node is a PROTO
     QMap<QString, WbProtoInfo *> list;
     if (variant == WbProtoManager::PROTO_WEBOTS)
       list = WbProtoManager::instance()->webotsProtoList();
     else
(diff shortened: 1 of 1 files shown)


ygoumaz (approved):
Thank you!

PR #5947, opened 2023-02-27
Fix regeneration seed logic on the web

ad-daniel:
It turns out that for each PROTO instance a new template engine was being created and the id assigned to the `context.id` was set by calling `getAnId()`. This is why changing the height of the building resulted in the textures changing as well, since at each regeneration a new id was used.

The obvious solution would be to try to replicate what is happening on the webots side, but due to the recent rework where all node clones have been removed, this is easier said than done. The reason is that in the current implementation, the node ids (those sent to webotsjs) are set **after** the node structure has been built, whereas the template engine would require this information **during** the building process. Changing this would require a big refactoring, and I don't think the current issue warrants it.

I've considered reworking some of the PROTO that use the `context.id` but I'm not really convinced we'd be improving anything in practice. RandomBuilding is used only in `CH_Vens`, `CH_Morges`,  `village_winter` and `village`, but not OSM. Removing the `context.id` seed entails manually giving an id to hundreds of buildings in the `.wbt` and I don't think that's really any better. `WoodenPalletStack` doesn't really seem to need this time-based variability either and making `SimpleTree` spawning trees in a time-based fashion would make several worlds non reproducible (or again changing hundreds of instances manually just to give them a random seed). Again, it doesn't feel like any of these changes would improve the current situation.

In short, on the web side I've opted instead to ensure that each PROTO instance gets its own "context.id" and that the same value is re-used between regenerations. So if you refresh the page you get a different building (which is consistent with the name of the PROTO), but if you change the height it keeps the same texture.

https://testing.webots.cloud/run?version=R2023b&url=https://github.com/ad-daniel/prefab/blob/master/protoExamples/protos/RandomBuilding.proto&type=undefined


omichel (approved):
That seems to be a reasonable approach.

PR #5950, opened 2023-02-28
Fix WbMFNode::setItem functionality

### LINKED ISSUE #5821: Node disappears from the scene tree when changing referenced DEF node
If in the following world, I set the `Transform` DEF name to `A`, so that the `USE A` node needs to be refreshed to reference to the Transform node instead of the Shape node then the `USE A` line is emptied but still there.

World:
```
#VRML_SIM R2023b utf8

WorldInfo {
}
Viewpoint {
  orientation -0.5773502691896257 0.5773502691896257 0.5773502691896257 2.0944
  position 0 0 10
}
TexturedBackground {
}
DEF A Shape {
}
Transform {
}
Group {
  children [
    USE A
  ]
}
TexturedBackgroundLight {
}
```

Steps:
1. set DEF name of Transform node to `A`
2. click OK in the message box dialog asking to apply changed to subsequent USE nodes

The scene tree is now wrong as depicted in this figure:
Screenshot from 2023-02-01 12-37-54

Tested only on the develop branch .

stefaniapedrazzi:
The `WbMFNode::setItem` method was not fully functional because many signal/slot connections were defined for the `itemRemoved` and `itemInserted` signals but not for the `itemChanged` signal.


omichel:
It looks good to me, but I am wondering why we are targeting the develop branch with this fix.

stefaniapedrazzi:
I'm targeting develop and not master because it is not really a fix but a code cleanup:
the `WbMFNode::setItem` never really worked and was never used (before https://github.com/cyberbotics/webots/pull/5676), the TODO comments were added when writing the Webots 7 code.

I checked deeply all the slots connected to the `itemRemoved` and `itemInserted` signals and I added the connection to the `itemChanged` signal. But given that `WbMFNode::setItem` was never used it might be that I missed something and it seems better to apply this cleanup to the develop branch that will be more tested.

omichel (approved):
Thank you.

BenjaminDeleze (approved):
Thank you

Issue #5882, opened 2023-02-12
Supervisor crash/hang after `enablePoseTracking()`


jaagut:
**Describe the Bug**
The supervisor controller (Python API) seems to crash/hang when using `enablePoseTracking()`.
Running `node1.getPose()` and `node2.getPose()` works just fine.
But running `node2.getPose()` after `node1.enablePoseTracking(<int>)` was called does not work, and the supervisor controller seems to hang forever.

**Steps to Reproduce**
1. Use one of these Webots versions: 
    - R2022b
    - R2023a
    - R2023b Nightly Build 10/2/2023
3. Open [this world](https://gist.github.com/jaagut/2b758d40b7e13b2dcc9020f079295032)
4. Run [this Python minimal example script](https://gist.github.com/jaagut/ae3eeb50650b5d2bd703ab789e7c07c6) `python3 minimal_example_supervisior_controller.py`
5. Examine the output of the script. Probably you will get the following output: _"IMPORTANT!!! Now the supervisor controller will hang forever..."_
6. Just in case the simulator stepping has an impact, the script can be run without stepping by setting `do_step = False` in line 5. Run the script again, results should be the same.
7. Also try `do_pose_tracking_node2 = False` (line 6) to check, whether it helps to enable pose tracking for all nodes we want to get the pose from.

<details><summary>Example output:</summary>
<p>
Getting the poses of both Nodes works:
Pose of Node1 [0.00037402848288182794, 0.1255107792045821, 0.9920921854374989, 0.030007285217697743, 0.9999804787604403, 0.006140897717199582, -0.0011538948251857391, 0.0014632167783275303, -0.006237162875434007, 0.9920732501578126, -0.12550603220793732, 0.23337273269731101, 0.0, 0.0, 0.0, 1.0]
Pose of Node2 [0.00037402848288182794, 0.1255107792045821, 0.9920921854374989, 0.030007285217697743, 0.9999804787604403, 0.006140897717199582, -0.0011538948251857391, 0.0014632167783275303, -0.006237162875434007, 0.9920732501578126, -0.12550603220793732, 0.23337273269731101, 0.0, 0.0, 0.0, 1.0]

Stepping the simulation...

Getting the poses of both Nodes works:
[...]

PR #5952, opened 2023-02-28
Fix `get_pose` requests hanging webots when multiple nodes are being tracked

ad-daniel:
**Description**

Fixes: https://github.com/cyberbotics/webots/issues/5882

The issue ended up being on the libcontroller side, not webots. Basically a chained structure holds the pose information of each tracked node, and whenever a pose request was made by the controller, the libcontroller would go through this chain and return the pose of the matching node (the chain is being fed updated pose information from the tracking). The process however went wrong when two (or more) pose requests occurred (while two nodes were being tracked) because when requesting the pose of the second, that request never had the chance of being met since the matching node could not be found on the chain (or rather, the iteration didn't even get there).
It also explains why tracking two objects but request only the pose of the first appeared to be working, since the loop did explore the first item in the chain at least.

### DIFF
--- a/src/controller/c/supervisor.c
+++ b/src/controller/c/supervisor.c
@@ -2042,15 +2042,16 @@ const double *wb_supervisor_node_get_pose(WbNodeRef node, WbNodeRef from_node) {
   }
 
   WbPoseStruct *tmp_pose = pose_collection;
-  while (tmp_pose)
+  while (tmp_pose) {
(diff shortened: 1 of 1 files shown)


ygoumaz (approved):
Thank you!

jaagut:
Thank you very much for addressing this issue! I meant to test this fix in my setup, but did not get to it in the past months. Still have to stick to R2022b for the foreseeable future…

Issue #5452, opened 2022-11-09
Long non-ASCII robot name crashes the controller.


amrytov:
**Describe the Bug**

See the subj.

**Steps to Reproduce**

1. Create an empty world
2. Put a Robot node in it. No child nodes are needed.
3. Leave the &lt;generic&gt; controller as is.
4. Name the robot with any long non-ascii name ("ЭтоДлинноеИмяДляРобота")
5. Run the simulation

** What happens:

I've just seen each of the following, in different combinations:

- the controller crashes (both mine and the &lt;generic&gt;, but this is not very reproducable, so no screenshot)
- errors starting the controller (see the screenshot)
- errors resetting the simulation (see the screenshot)

**Expected behavior**

Either accept the name as valid and then work correctly, 
or report that the name is too long or contains illegal characters and not accept it.

Suggested possible solution(s):  

(a) Check validity of the string assigned to the 'name' field (don't allow such names)
(b) don't use robot name directly to create pipe names or whatever, instead create some fixed-length harmless hash string from them,
so that any strings can be safely used.

**Screenshots**

![2022-11-09 21_49_30-D__amrytov_documents_NTO_Space_worlds_TestLongNames wbt (NTO_Space) - Webots R20](https://user-images.githubusercontent.com/8374186/200916181-5b638940-bb3f-4458-b395-b8f0a30e0d83.png)
![2022-11-09 21_48_34-D__amrytov_documents_NTO_Space_worlds_TestLongNames wbt (NTO_Space) - Webots R20](https://user-images.githubusercontent.com/8374186/200916182-47811f85-e672-4c61-8449-ce17b35f7491.png)
![2022-11-09 21_48_15-D__amrytov_documents_NTO_Space_worlds_TestLongNames wbt (NTO_Space) - Webots R20](https://user-images.githubusercontent.com/8374186/200916183-16013f67-2fbe-4503-b880-4eecce8adfd8.png)

**System**
 - Operating System: Windows 10

brettle:
I just ran into this bug when running the test suite on a Mac where my username was 11 characters long.
[...]

PR #5953, opened 2023-02-28
Prevent long server names and non-ascii robot names

ad-daniel:
**Description**

Fixes https://github.com/cyberbotics/webots/issues/5452

- Shows an error when the server name is too long (QLocalServer has a hard limit at 106 characters)
- Forces robot names to be ASCII. Connecting to a huge percentage-encoded string seems silly


omichel:
Where did you find that the limit for QLocalSocket is 106 characters?

I believe we should not introduce any constraint on the robot name. Instead, if the string length is greater than 106 characters, we could compute a base64 sha512 string, which should be no longer than [86 characters](https://stackoverflow.com/questions/18236106/what-is-the-length-of-a-hashed-string-with-sha512) and pretty unique. The drawback of this approach is that it should be implemented on both sides (Webots and libController).

ad-daniel:
Changes are required on the launcher as well, so will target develop instead

PR #5956, opened 2023-03-01
Fix basetype for role attribution

BenjaminDeleze:
As it was just checking the name of the current node, it was not attributing the correct role to protos.
For example in the `Eyescrew.proto`, the `TorusBoundingObject.proto` was not given the role `boundingObject` and it was crashing webotsJS.

Test it here: https://testing.webots.cloud/run?version=R2023b&url=https://github.com/cyberbotics/webots/blob/doc-add-objects/projects/objects/factory/tools/protos/EyeScrew.proto&type=undefined

### DIFF
--- a/resources/web/wwi/protoVisualizer/Vrml.js
+++ b/resources/web/wwi/protoVisualizer/Vrml.js
@@ -392,15 +392,7 @@ export class SFNode extends SingleValue {
     const nodeX3d = this.value.toX3d(parameterName);
 
     // handle exceptions
-    let name;
-    if (!this.value.isProto)
-      name = this.value.name;
-    else {
-      let baseType = this.value.baseType;
-      while (baseType.isProto)
-        baseType = baseType.baseType;
-      name = baseType.name;
-    }
+    const name = this.value.getBaseNode().name;
     if (this.value.name === 'ImageTexture')
       nodeX3d.setAttribute('role', parameterName);
     else if (['Shape', 'Group', 'Transform', 'Solid', 'Robot'].includes(name)) {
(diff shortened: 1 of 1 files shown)


resources/web/wwi/protoVisualizer/Vrml.js
```diff
@@ -392,14 +392,23 @@ export class SFNode extends SingleValue {
...
     // handle exceptions
+    let name;
+    if (!this.value.isProto)
+      name = this.value.name;
+    else {
+      let baseType = this.value.baseType;
+      while (baseType.isProto)
+        baseType = baseType.baseType;
+      name = baseType.name;
+    }
```
ad-daniel, line 403: this can be avoided by calling `this.value.getBaseNode().name` directly.
However I have to admit I'm a bit perplexed why we need to do so in the first place. x3d encoding [should only occur for basenodes](https://github.com/cyberbotics/webots/blob/11a426186a8a1ea5d7c28353923937c32a791e1a/resources/web/wwi/protoVisualizer/Node.js#L255-L256) anyway, so I fear we shouldn't end up in this situation in the first place. I have the feeling the problem might be elsewhere, I'll investigate a bit

ad-daniel (changes requested):

ad-daniel (approved):
Ok no it's to be expected. The loop I mention that brings the encoder down at the basenode level works for nodes, here we are dealing with the parameters of a Node (in this case SFNode) and the encoding of it's value comes afterwards (so `nodeX3d` in this function will be correctly brought down to a basenode if it's a PROTO). However the role needs to be defined at the parameter level so it's inevitable

PR #5957, opened 2023-03-01
Fix USE node setup in WbDictionary

### LINKED ISSUE #5821: Node disappears from the scene tree when changing referenced DEF node

stefaniapedrazzi:
Fix #5821 (introduced in #5676 ):
currently the `WbMFNode::setItem` function doesn't work correctly because the `itemChanged` signal is not correctly handled.

A more complete fix for `WbMFNode::setItem` is inspected in #5950 but should target the develop branch.


omichel (approved):
Thank you.

PR #5959, opened 2023-03-01
Fix invalid robot names

ad-daniel:
**Description**

Fixes https://github.com/cyberbotics/webots/issues/5452

This PR:
- If the percentage-encoded robot name length is beyond an arbitrary size of 70 characters, a hash is used instead. The risk is that if the Robot name contains non-ASCII characters, the percentage encoding can result in a huge string which ends up crashing the controller when the connection takes place (QLocalServer has a limit of 106 characters for the server name)
- Reverts https://github.com/cyberbotics/webots/pull/5484 so that "/tmp" is used instead of `TMP_DIR` and changes the permissions so that multiple users can

### DIFF
--- a/src/webots/control/WbController.cpp
+++ b/src/webots/control/WbController.cpp
@@ -441,7 +441,7 @@ void WbController::setProcessEnvironment() {
   QProcessEnvironment env = QProcessEnvironment::systemEnvironment();
 
   // store a unique robot name for the controller
-  env.insert("WEBOTS_ROBOT_NAME", mRobot->encodedName());
+  env.insert("WEBOTS_ROBOT_NAME", mRobot->name());
 
   // Add the Webots lib path to be able to load (at least) libController
   QString ldLibraryPath = WbStandardPaths::controllerLibPath();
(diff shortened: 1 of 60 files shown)


ad-daniel:
I've tested it on linux and macos, couldn't test on windows

omichel (changes requested):
It looks very good to me.
I am now testing on Windows different use-cases, including local and remote controllers.

omichel (dismissed):

omichel:
I would also add a line in the change log.

omichel (approved):
Thank you.

src/webots/control/WbController.cpp
```diff
@@ -314,6 +321,9 @@ void WbController::start() {
   }
   file.write("");
   file.close();
+
+  // FIXME: from Qt 6.4 onwards, QDir::mkdir can be used to set the permissions
+  QProcess::execute("sh", QStringList() << "-c" << QString("chmod -R 777 %1").arg(mIpcPath));
```
omichel, line 326: This code seems wrong: useless and not working on Windows where it is supposed to be executed.

PR #5961, opened 2023-03-02
Fix update of joint endPoint after regeneration

### LINKED ISSUE #5951: Changing a proto parameter stops the motion
While playing with [friction_gears.zip](https://github.com/cyberbotics/webots/files/10846672/friction_gears.zip), I found that if I change the number of teeth of the `Robot / children / DEF PASSIVE HingeJoint / endPoint Gear / teeth` from 20 to 19, the motion of this wheel stops. If I reset the simulation, it doesn't start again. If I save as and reload the simulation, it works.

I can reproduce the problem both on the master and develop branches.

stefaniapedrazzi:
Fix #5951:
listen to changes in descendant joint `endPoint` field and update the children list when needed.

Note that given that the joint solid end point is also added to the `WbSolid::mSolidChildren` list, the `WbSolid::updateChildren` function is automatically called when the end point is deleted. Thus, it is not needed to call `WbSolid::updateChildren` when the `WbBasicJoint::endPointChanged` node argument is NULL.


omichel (dismissed):
Looks good to me.
Thank you.

omichel (approved):

PR #5965, opened 2023-03-06
Fix test_suite overwriting local Webots configuration

### LINKED ISSUE #5960: Test suite changes the graphical options to minimum
Since #5806, running the test suite locally will change all graphical parameters to the minimum (disable gtao, shadows,...) which is pretty annoying when developing.

It is because we are now running "Run a copy of webots in the background to ensure doing so doesn't cause any tests to fail. Use WEBOTS_SAFE_MODE=true to ensure it always runs with the same (empty) project."

The WEBOTS_SAFE_MODE environment variable is responsible for the parameter changes https://github.com/cyberbotics/webots/blob/c759a4dbfac3addd4ded56a9a470c3d5171873fa/src/webots/gui/WbGuiApplication.cpp#L260

As the changes are made by Webots (and not directly by the test suite) it seems more complicated to revert them.

One solution would be to run this specific test only in the CI.

What do you think?

stefaniapedrazzi:
Fix #5960:
start directly the empty world instead of using the `WEBOTS_SAFE_MODE` to avoid overwriting the local Webots configuration.


omichel (commented):
Why did you remove the `--minimize` option? This is pretty convenient when running the test suite locally, otherwise Webots occupies your screen while making tests, which is a bit annoying.

stefaniapedrazzi:
Oops, you're are right. I removed it for debugging and forget to readd it.

omichel (approved):
Thank you.

Issue #2390, opened 2020-10-26
Camera recognition: if Camera is scaled recognized objects bounding boxes are scaled too


stefaniapedrazzi:
A user reported that the bounding box displayed in the camera overlay was not matching the object position in the camera image.
The issue should come from an ancestor scale Transform of the Camera node.

`I have a scale transform to make my camera appear right, and I think it is also somehow having the effect of transforming the locations of the recognition boxes`
`Yep, moving the camera outside the scale transform fixed it right up!  I never would have guessed that a scale transform would affect a camera like that!`

stefaniapedrazzi:
Camera scaling is forbidden since R2023b.

PR #5967, opened 2023-03-06
Fix camera recognition

stefaniapedrazzi:
Detection resulting from the camera recognition functionalities and displayed in the camera image are sometimes wrong.

---
Camera recognition bugs:
* [x] try to reduce extra padding when object size is computed using the bounding sphere
* ~~inspect #2390~~ 
    _-> we will drop scaling of devices in R2023b when inserting the Pose node_
* [x] inspect occlusion preventing detection of crop plants:
      * currently only one ray pointing to the object center is used to check if object is occluded
      * checking the corners of the bounding sphere/box instead of just the center gives better results -> fix in the develop branch
* [x] check and fix (if needed) object detection on image in case
    * [x] object in frustum computation if fovX > PI 
    * [x] spherical projection
    * [x] cylindrical projection 
    * [x] Lens 
    * [x] Zoom
* [x] fix lens distortion in segmented image
* [x] fix recognized objects info not available when mouse is over the overlay image after the controller terminated if occlusion is set to FALSE (work correctly if occlusion is TRUE)
* [x] fix detected size of big objects partly behind planar device
---
Generic bugs to be addressed in a different PR (#6063):
* wrong translation and rotation displayed in Position tab in the Field Editor -> rotation should not be affected by ancestor scale
* handle not working correctly for Transform nodes with a scaled parent
* ~fix `WbAbstractCamera` frustum representation if node is scaled -> frustum should not be affected by scale~ 
    _-> reverted as we will soon no longer support a scaled device_


BenjaminDeleze (commented):
I played with `CameraSegmentation.wbt` and `CameraRecognition.wbt`  and it seems good. However when I modified the parameters of a `Lens` during execution I often get: `OpenGL error: GL_INVALID_VALUE`
`, I also get it sometimes when playing with the field of view.

stefaniapedrazzi:
I only get issues when the segmentation is also enabled. Is it the same for you?

Crashes when changing the field of view are also reproducible in the master branch (see #6033).

BenjaminDeleze:
> I only get issues when the segmentation is also enabled. Is it the same for you?

No, I can have the messages by doing:

1. Open camera recognition (so segmentation is set to false)
2. Run
3. Add a Lens
4. Change tangentialCoefficient to 1

> Crashes when changing the field of view are also reproducible in the master branch

Yes

BenjaminDeleze:
> * Open camera recognition (so segmentation is set to false)
> * Run
> * Add a Lens
> * Change tangentialCoefficient to 1

I still have the error message when following this procedure

stefaniapedrazzi:
> 

This time I cannot reproduce it. Everything seems to work correctly for me.
No crash, and no GL_INVALID_VALUE error.

BenjaminDeleze:
The test suite has failed

BenjaminDeleze:
Strange I just tested at this office and could not reproduce the issue anymore.
However the test suite is still failing.

BenjaminDeleze:
I can reproduce it with this lighten version of  `camera_recognition.wbt`:
[gl_warnings.zip](https://github.com/cyberbotics/webots/files/11201964/gl_warnings.zip)

The warnings always appear at the same timing: when the camera leaves the shortbreads.

I investigate it a bit.

BenjaminDeleze:
After some research, here is what I found:

- The `GL_INVALID_VALUE` persists even after I tried another driver
- I can reproduce the `GL_INVALID_VALUE` on my Windows machine as well
- In [camera.wbt](https://github.com/cyberbotics/webots/blob/9028a89e9ec2aeaf09b20ec667218bdf945415d6/src/webots/nodes/WbCamera.cpp#L805) the result of 
```
// convert uv to range [0, 1]
uv.setX(qMax(0.0, qMin(0.5 + uv.x(), 1.0)));
uv.setY(qMax(0.0, qMin(0.5 + uv.y(), 1.0)));
uv = applyCameraDistortionToImageCoordinate(uv);
```
is sometime something like `[2, 1]` which is out of the range fixed by the comment (except if the comment is only valid for the two first lines)

- Just [below](https://github.com/cyberbotics/webots/blob/9028a89e9ec2aeaf09b20ec667218bdf945415d6/src/webots/nodes/WbCamera.cpp#L808) we have: 
```
// return uv coordinates in range [0, width/height]
return WbVector2((int)((width() - 1) * uv.x()), (int)((height() - 1) * uv.y()));
```
So when uv.x() > 1 (which happens) we have value greater then width (for example 510 instead of 255).

By clamping the value of uv like that:
```
// convert uv to range [0, 1]
uv.setX(qMax(0.0, qMin(0.5 + uv.x(), 1.0)));
uv.setY(qMax(0.0, qMin(0.5 + uv.y(), 1.0)));
uv = applyCameraDistortionToImageCoordinate(uv);
if (uv.x() > 1.0)
  uv.setX(1.0);
if (uv.y() > 1.0)
  uv.setY(1.0);

// return uv coordinates in range [0, width/height]
return WbVector2((int)((width() - 1) * uv.x()), (int)((height() - 1) * uv.y()));
```
I had no more errors but I don't know if it is the right way to patch the issue.

stefaniapedrazzi:
Thank you for debugging the issue!
Yes, it makes sense to clamp the UV coordinates computed from lens distortion.

BenjaminDeleze (approved):
Thank you!

PR #5970, opened 2023-03-07
WebotsJS: Move floor when base type is slot or shape

BenjaminDeleze:
It makes sense because `Slot` will never be a top node and `Shape` will by default be centered at `0 0 0`


omichel (dismissed):
Thank you.

omichel (dismissed):

omichel (approved):

PR #5977, opened 2023-03-09
WebotsJS: Fix mf restrictions

BenjaminDeleze:
**Description**
Implement the restriction mechanism for MFNode as it is needed for the Road protos.

Test it here: https://testing.webots.cloud/run?version=R2023b&url=https://github.com/cyberbotics/webots/blob/doc-add-objects/projects/objects/road/protos/AddLaneRoadSegment.proto&type=undefined

- [x] Parsing of the proto

@ad-daniel for the NodeSelectorWindow I have in mind to just add a check in `fetchCompatibleNodes` of `NodeSelectorWindow.js`, do you think it is a good idea to proceed like that?

### DIFF
--- a/resources/web/wwi/protoVisualizer/Parameter.js
+++ b/resources/web/wwi/protoVisualizer/Parameter.js
@@ -1,6 +1,6 @@
 'use strict';
 
-import {SFNode, stringifyType} from './Vrml.js';
+import {SFNode, MFNode, stringifyType} from './Vrml.js';
 import Field from './Field.js';
 
 export default class Parameter extends Field {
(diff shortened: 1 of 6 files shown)


ad-daniel:
Uhm, restrictions for MFNode was implemented already, unless it broke at some point. Essentially MFNode restrictions are imposed like SFNode restrictions, meaning it's just about only showing the allowed nodes in the NodeSelectorWindow. Basically [this](https://github.com/cyberbotics/webots/blob/ffa2cff35dda621eed5eaf41aa780235c8ca11fa/resources/web/wwi/NodeSelectorWindow.js#L218-L220) check should exclude everything except the allowed ones

BenjaminDeleze:
> Uhm, restrictions for MFNode was implemented already, unless it broke at some point. Essentially MFNode restrictions are imposed like SFNode restrictions, meaning it's just about only showing the allowed nodes in the NodeSelectorWindow. Basically [this](https://github.com/cyberbotics/webots/blob/ffa2cff35dda621eed5eaf41aa780235c8ca11fa/resources/web/wwi/NodeSelectorWindow.js#L218-L220) check should exclude everything except the allowed ones

True, I will check why it failed.

resources/web/wwi/protoVisualizer/Parameter.js
```diff
@@ -27,10 +27,36 @@ export default class Parameter extends Field {
       throw new Error(`Type mismatch, setting ${stringifyType(newValue.type())} to ${stringifyType(this.type)} parameter.`);
 
     if (this.restrictions.length > 0) {
-      for (const item of this.restrictions) {
-        if ((newValue instanceof SFNode && newValue.value === null) || item.equals(newValue)) {
+      if (newValue instanceof MFNode) {
+        if (newValue.value === null) {
```
ad-daniel, line 31: I think this conditional (newValue.value === null) should be checked only for SFNodes, MFNodes can't be `null`. At best they can be `[]`, but that's taken care of already

ad-daniel (changes requested):
Works well, just a possible improvement

PR #5978, opened 2023-03-10
Add Pose node

ad-daniel:
This PR aims at introducing a new node named `Pose` which basically inherits from `Group`, adding only the `translation` and `rotation` field. The `Transform` node inherit from `Pose` adding the `scale` node.
All `Solid` nodes inherit from `Pose` and not from `Transform` any more.
The `Transform` node may only contain `Shape`, `CadShape`, `Group`, `Pose` or `Transform` nodes in its `children` list.
This has 3 mains advantages:
- It makes it clearer for the end-user that it is not possible to scale `Solid` nodes (however PROTOs of `Solid` nodes may still have a working scale parameter, which may be either scalar or vector).
- It simplifies of the management of `Solid` nodes and bounding objects.
- It simplifies the management of `Device` nodes (which inherit from `Solid`).

Scaling `Solid` node is not something realistic as in the real world, it is not possible (or very difficult) to scale objects, whereas it is super easy to change their translation and rotation.

See the most relevant update documentation pages:
- https://cyberbotics.com/doc/reference/transform?version=feature-pose-node
- https://cyberbotics.com/doc/reference/pose?version=feature-pose-node
- https://cyberbotics.com/doc/reference/solid?version=feature-pose-node
- https://cyberbotics.com/doc/reference/group?version=feature-pose-node


resources/nodes/Pose.wrl
```diff
@@ -0,0 +1,15 @@
+# The Transform node is a grouping node that defines a coordinate system for its children that is
+# relative to the coordinate system of its parent.
+# The 'scale' field of a Transform node can be adjusted only in a graphical context and not in a 'boundingObject' context.
+
+Transform {
```
stefaniapedrazzi, line 5: This should be Pose and you should remove the scale. from the fields.

omichel:
@ad-daniel: I am following-up on this PR, but I can't remember what was remaining to be done. As far as I remember, the PR was almost complete, there was just one thing to be completed. Could you please explain here what remains to be done, so that I can follow-up?

omichel:
All the tests are now passing 🎉!

ad-daniel:
Hi, sorry I have been a bit busy. The test suite used to fail consistently on the physics group (due to rolling friction) but passing when running the singular test, but that doesn't seem to be the case anymore as far as I can tell. Beyond that there were only the right-click transform-to rules that needed to be updated and a couple fixes to do on the webotsjs side. I will care of it over the weekend.

There was also the open question regarding whether transform nodes should be selectable from the 3d view. Before, a transform was selectable only if it contained a solid. In this pr solids can't be inserted in transforms, so they can't be selected. In short, the behavior is now consistent with those rules but not necessarily user friendly or desirable

omichel:
Thank you! However, if don't have time, don't worry, I can follow-up on this if you tell me more precisely where/what to patch the code.

> There was also the open question regarding whether transform nodes should be selectable from the 3d view. Before, a transform was selectable only if it contained a solid. In this pr solids can't be inserted in transforms, so they can't be selected. In short, the behavior is now consistent with those rules but not necessarily user friendly or desirable.

I would keep it like that in this PR and maybe we can improve it later in a subsequent PR.

omichel:
I just updated the right-click transform-to rules in faa5082.

omichel:
Hi @ad-daniel, can you tell me what are the "couple fixes to do on the webotsjs side" you were mentioning? I am happy to take of them.

stefaniapedrazzi (commented):
I know that this PR is not completed yet.
But I quickly check (also in the past weeks) and I have a couple of comments.

Probably `Skin.wrl` should be updated because `scale` is no longer an inherited field.
Similarly the documentation pages of Fluid, Solid and Robot (and Skin) should be updated: the base  is not changed from Transform to Pose.

In the Transform documentation are still mentioned limitations in case of boundingObject.

src/webots/nodes/WbTriangleMeshGeometry.cpp
```diff
@@ -129,9 +129,11 @@ void WbTriangleMeshGeometry::createWrenObjects() {
 void WbTriangleMeshGeometry::setResizeManipulatorDimensions() {
   WbVector3 scale(1.0f, 1.0f, 1.0f);
 
-  WbTransform *transform = upperTransform();
-  if (transform)
-    scale *= transform->absoluteScale();
+  const WbPose *const up = upperPose();
```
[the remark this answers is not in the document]
ad-daniel: it shouldn't be necessary has it's already taken care of. If `up` is defined, then `up->absoluteScale` is called (and both poses and transforms define the method)

src/webots/gui/WbDragScaleEvent.cpp
```diff
@@ -246,107 +245,3 @@ void WbRescaleIndexedFaceSetEvent::apply(const QPoint &currentMousePosition) {
   mTotalScaleRatio *= mResizeRatio;
   mManipulator->updateHandleDimensions(mTotalScaleRatio, mViewDistanceUnscaling);
 }
-
-// Moves a scale handle by dragging the mouse and changes the scale field accordingly //
-////////////////////////////////////////////////////////////////////////////////////////
-
-WbDragScaleHandleEvent::WbDragScaleHandleEvent(const QPoint &initialMousePosition, WbViewpoint *viewpoint, int handleNumber,
```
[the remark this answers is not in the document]
ad-daniel: You are right, but in practice to do so would require to create a bit of a mess. Basically the functionality would be needed by the Skin node and the Transform node (and descendants), which would require to have the necessary functionality to be defined at the AbstractPose level (which is the only class common to both). Doing so there however doesn't make much sense conceptually, and likewise the dragscale behavior would need to be deactivated for Pose nodes and re-activated for Transforms and descendants.
In short, after discussing it with olivier it didn't seem worth the exemplification required to achieve it.

src/webots/nodes/utils/WbObjectDetection.cpp
```diff
@@ -189,7 +189,7 @@ bool WbObjectDetection::isWithinBounds(const WbAffinePlane *frustumPlanes, const
         const WbIndexedFaceSet *indexedFaceSet = static_cast<const WbIndexedFaceSet *>(boundingObject);
         const WbCoordinate *coordinates = indexedFaceSet->coord();
         for (int i = 0; i < coordinates->pointSize(); ++i) {
-          points.append(objectRotation * (coordinates->point(i) * mObject->absoluteScale()) + objectPosition);
+          points.append(objectRotation * coordinates->point(i) + objectPosition);
```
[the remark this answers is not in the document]
ad-daniel: It would need to be tested to be 100% sure but by taking a quick look,  `mObject` is of `WbSolid` type (already was in master) therefore in theory there cannot be a Transform node in its ancestry so it should not be possible for it to have a scale. 
In practice however it's possible this was changed prior to the re-implementation a `absoluteScale` at the Pose level which turned out to be necessary only at a later stage of the rework, so technically it would still compile if the change is reverted.

The question therefore is whether it's necessary or not, I don't recall exactly all the cases where the objectdetection class is involved (like, does it need to have recognitioncolors or radarCrossSection, things like that?) If so those are only available to solids indeed, pure geometries wouldn't

src/webots/nodes/utils/WbObjectDetection.cpp
```diff
@@ -201,8 +201,7 @@ bool WbObjectDetection::isWithinBounds(const WbAffinePlane *frustumPlanes, const
         int yDimension = elevationGrid->yDimension();
         for (int i = 0; i < xDimension; ++i) {
           for (int j = 0; j < yDimension; ++j)
-            points.append(objectRotation * (WbVector3(xSpacing * i, elevationGrid->height(i + j * xDimension), ySpacing * j) *
-                                            mObject->absoluteScale()) +
+            points.append(objectRotation * WbVector3(xSpacing * i, elevationGrid->height(i + j * xDimension), ySpacing * j) +
```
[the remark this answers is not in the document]
ad-daniel: ditto, see above

src/webots/nodes/WbSolid.cpp
```diff
@@ -1965,26 +1879,24 @@ void WbSolid::applyPhysicsTransform() {
...
+    const WbMatrix4 &upm = up->matrix();
+    const WbVector3 &prel = upm.pseudoInversed(WbVector3(result));
+    result[0] = prel[0];
+    result[1] = prel[1];
+    result[2] = prel[2];
     // printf("result = %f, %f, %f (apply phy.))\n", result[0], result[1], result[2]);
-    // find rotation difference between upper transform and solid child
-    const WbQuaternion &q = utm.extractedQuaternion(invUtScale);
+    // find rotation difference between upper pose and solid child
+    const WbQuaternion &q = upm.extractedQuaternion();
```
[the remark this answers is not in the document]
ad-daniel: no, `WbQuaternion` doesn't have a constructor from WbMatrix4

ad-daniel:
> Hi @ad-daniel, can you tell me what are the "couple fixes to do on the webotsjs side" you were mentioning? I am happy to take of them.

Basically WbPose.js and WbTransform.js need to be checked to ensure they're consistent with the webots side. One such cases where it isn't currently so is `vrmlMatrix` where this method should be overriden at the WbTransform,js level to take the `scale` into account, but there might be more

src/webots/nodes/utils/WbObjectDetection.cpp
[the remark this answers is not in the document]
stefaniapedrazzi: The mObject is always of type Solid.
But descandant geometries of mObject could still have parent Transform node.

However, I double checked and we are looking only in boundingObject field, so the transform case and scale can be safely ignored.

src/webots/nodes/utils/WbObjectDetection.cpp
```diff
@@ -142,18 +142,18 @@ bool WbObjectDetection::isWithinBounds(const WbAffinePlane *frustumPlanes, const
...
+  assert(pose);
+  WbMatrix3 objectRotation = pose->rotationMatrix();
+  WbVector3 objectPosition = pose->position();
 
   if (nodeType == WB_NODE_SHAPE) {
     const WbShape *shape = static_cast<const WbShape *>(boundingObject);
     boundingObject = shape->geometry();
     return isWithinBounds(frustumPlanes, boundingObject, objectSize, objectRelativePosition);
-  } else if (nodeType == WB_NODE_GROUP || nodeType == WB_NODE_TRANSFORM) {
+  } else if (nodeType == WB_NODE_GROUP || nodeType == WB_NODE_TRANSFORM || nodeType == WB_NODE_POSE) {
```
stefaniapedrazzi, line 156: The test `nodeType == WB_NODE_TRANSFORM` can be removed if Transform nodes can no longer be inserted in the boundingObject.

src/webots/nodes/utils/WbNodeUtilities.cpp
```diff
@@ -473,7 +473,7 @@ namespace {
 
         // if the node itself is a transform or it has a transform ancestor exists, prohibit the insertion of Solids,
         // Robots, Devices, Propellers & Lights
-        if (node->modelName() == "Transform" || WbNodeUtilities::findUpperTransform(node)) {
+        if (isTransformOrTransformDescendant) {
```
omichel, line 476: This broke the compilation:
```
nodes/utils/WbNodeUtilities.cpp:476:13: error: 'isTransformOrTransformDescendant' was not declared in this scope
```

omichel:
> Probably `Skin.wrl` should be updated because `scale` is no longer an inherited field.

The current version looks good to me. What would you change?

src/webots/nodes/WbSkin.cpp
```diff
@@ -70,22 +70,25 @@ void WbSkin::init() {
...
   mBoundingSphere = NULL;
 
+  mPreviousXscaleValue = 1.0;
+
   mName = findSFString("name");
   mModelUrl = findSFString("modelUrl");
   mAppearanceField = findMFNode("appearance");
   mBonesField = findMFNode("bones");
   mCastShadows = findSFBool("castShadows");
+  mScale = findSFVector3("scale");
```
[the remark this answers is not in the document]
omichel: It seems the connect was there, but using the old syntax. I fixed this in 55995c1.

omichel:
I believe I addressed all the comments.

stefaniapedrazzi (commented):
Thank you for the changes!

I started by reviewing the documentation and then I will review the rest.
- The Transform documentation still mentions that Solid is a direct derived node.
    Then, it is a little bit strange that for the other nodes we do not list the derived fields but for the Transform one we do.
- The documentation of the functions `wb_supervisor_node_get_position` and pose_tracking, etc. should be updated because now the expected argument should be a Pose node and not a Transform node.
- The new Pose node has been added in the list of node of the Supervisor for C and MATLAB language, but not for C++, Java and Python.
- It is also missing an entry in the ChangeLog.

src/webots/nodes/utils/WbSupervisorUtilities.cpp
```diff
@@ -1346,7 +1346,7 @@ void WbSupervisorUtilities::handleMessage(QDataStream &stream) {
       }
 
       if (enable) {
-        WbTransform *const fromTransformNode = fromNode ? dynamic_cast<WbTransform *>(fromNode) : NULL;
+        WbPose *const fromTransformNode = fromNode ? dynamic_cast<WbPose *>(fromNode) : NULL;
         if (fromNodeId && !fromTransformNode)
           mRobot->warn(
             tr("Pose tracking can be exclusively used with Transform (or derived) 'from_node' argument, but '%1' (%2) is "
```
stefaniapedrazzi, line 1352: It seems the warnings of the get and track pose functions (and the variable names) are not fully converted to the Pose nodes.
```suggestion
            tr("Pose tracking can be exclusively used with Pose (or derived) 'from_node' argument, but '%1' (%2) is "
```

stefaniapedrazzi:
In this branch it seems it is still possible to add a Solid or passive HingeJoint into a scaling Transform node.
An example is the OfficeChair.
Is this the wanted behavior?

omichel:
> In this branch it is still possible to add a Solid or passive HingeJoint into a scaling Transform node. An example is the OfficeChair. Is this the wanted behavior?

No, it shouldn't be allowed.

stefaniapedrazzi:
I tested the behavior and it working fine apart from the already reported issue with the insertion of joints into Transform nodes everything seems to work as before in Webots (I didn't test webots.js).

omichel:
> In the TrackWheel page it is written that is derived from Group, but this doesn't correspond to the code or node chart where it is derived from Pose.

Yes, but `TrackWheel` has no translation field and only a hidden `rotation` field, so from a user's perspective, it is not derived from `Pose`. We should probably change the node chart about this.

omichel:
Since this is a huge PR with almost 1000 files changed, I suggest that @ygoumaz and @BenjaminDeleze agree together on what they will check/test here among the Webots code, documentation, protos, worlds, webotsjs, etc.

docs/reference/node-chart.md
```diff
@@ -56,36 +56,37 @@ graph LR
...
-        SolidDevice --> RangeFinder[[RangeFinder](rangefinder.md)]
-        SolidDevice --> Receiver[[Receiver](receiver.md)]
-        SolidDevice --> Speaker[[Speaker](speaker.md)]
-        SolidDevice --> TouchSensor[[TouchSensor](touchsensor.md)]
-      Solid --> Track[[Track](track.md)]
-      Solid --> Charger[[Charger](charger.md)]
-      Solid --> Robot[[Robot](robot.md)]
-    Transform --> Fluid[[Fluid](fluid.md)]
-    Transform --> TrackWheel[[TrackWheel](trackwheel.md)]
+  Group --> TrackWheel[[TrackWheel](trackwheel.md)]
```
BenjaminDeleze, line 61: In `TrackWheel.hpp` TrackWheel is derived from Pose

docs/reference/transform.md
```diff
@@ -1,18 +1,14 @@
...
   SFVec3f    scale           1 1 1     # any vector
-  SFFloat    translationStep 0.01
-  SFFloat    rotationStep    0.261799387
 }
 ```
 
-Direct derived nodes: [Solid](solid.md).
+Direct derived nodes: none.
 
 ### Description
```
BenjaminDeleze, line 13: The description contains both the Pose specific field and the inherited ones whereas it should only contain the specific one (scale)

resources/web/wwi/nodes/WbSpotLight.js
```diff
@@ -163,7 +163,7 @@ export default class WbSpotLight extends WbLight {
   }
 
   #attachToUpperTransform() {
-    const upperTransform = findUpperTransform(this);
+    const upperTransform = findUpperPose(this);
```
BenjaminDeleze, line 166: Should be renamed

resources/web/wwi/nodes/utils/WbBoundingSphere.js
```diff
@@ -172,7 +173,7 @@ export default class WbBoundingSphere {
     if (!this.parentCoordinatesDirty)
       return;
 
-    if (this.transformOwner) {
+    if (this.#owner.nodeType === WbNodeType.WB_NODE_TRANSFORM) {
```
BenjaminDeleze, line 176: Seems wrong to check for transform in BO

BenjaminDeleze (changes requested):
Here are some preliminary remarks on the documentation and webotsjs.

1. I guess that the absolute scale is defined in Pose and not in Transform because we can have a Pose inside a Transform. But I wonder if we should not prevent it as I do not see the utility of it.
2. In webots, I could break the insertion rules (I guess) on Webots by doing:
  - Robot -> Transform -> HingeJoint -> Robot

resources/web/wwi/Parser.js
```diff
@@ -768,6 +770,41 @@ export default class Parser {
...
+
+    WbWorld.instance.nodes.set(newNode.id, newNode);
+
+    this.#parseChildren(node, newNode, isBoundingObject);
+
+    if (typeof parentNode !== 'undefined') {
+      newNode.parent = parentNode.id;
+      if (isBoundingObject && parentNode instanceof WbSolid)
+        parentNode.boundingObject = newNode;
+      else if (parentNode instanceof WbSlot || parentNode instanceof WbJoint)
```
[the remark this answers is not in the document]
omichel: Yes, if it doesn't contain any Solid.

docs/reference/node-chart.md
omichel: Yes in the code of Webots (I guess @ad-daniel implemented like this for convenience), but not for the users because it doesn't have a `translation` field and its `rotation` field is hidden. So, it only inherits the `children` field from `Group`.

docs/reference/transform.md
omichel: But that's the case. Are you sure your are up-to-date with this branch?

resources/web/wwi/nodes/WbGeometry.js
```diff
@@ -61,7 +61,7 @@ export default class WbGeometry extends WbBaseNode {
       if (this.isInBoundingObject()) {
         if (parent instanceof WbSolid)
           parent.boundingObject = undefined;
-        else if (parent instanceof WbTransform || parent instanceof WbGroup) {
+        else if (parent instanceof WbPose || parent instanceof WbGroup) {
```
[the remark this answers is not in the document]
BenjaminDeleze: Yes it checks for [inheritance](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof). We actually use that fact quite a lot in here: https://github.com/cyberbotics/webots/blob/eb8acdf9e697e9b208743e2be152568a0d9d70fd/resources/web/wwi/X3dScene.js#L195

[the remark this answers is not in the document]
omichel: Oops, I didn't know that. I will fix it accordingly.

ygoumaz:
The description of the PR should be updated to detail the changes.

docs/reference/transform.md
BenjaminDeleze: Sorry not the description but the field summary

BenjaminDeleze (changes requested):
I think that this PR will break the backward compatibility for Scene and animation. (It is not a problem but we should be aware of it)

Right now webotsJS is not working as the `pose` case is missing from the `parseNode` function of `Parser.js`

docs/reference/transform.md
```diff
@@ -1,18 +1,14 @@
...
-  SFVec3f    translation     0 0 0     # any vector
-  SFRotation rotation        0 0 1 0   # unit axis, (-inf, inf) angle
   SFVec3f    scale           1 1 1     # any vector
-  SFFloat    translationStep 0.01
-  SFFloat    rotationStep    0.261799387
 }
 ```
 
-Direct derived nodes: [Solid](solid.md).
+Direct derived nodes: none.
```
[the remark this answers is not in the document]
omichel: I know, but I did it on purpose to clarify the change we are introducing in this PR.
Before that PR, the Transform node was an ancestor of many nodes (all the Solid, Robot, devices, etc.).
Now it is not any more the case.

BenjaminDeleze:
> > I guess that the absolute scale is defined in Pose and not in Transform because we can have a Pose inside a Transform. But I wonder if we should not prevent it as I do not see the utility of it.
> 
> Good point. However, that would also mean that we cannot insert a Group inside a Transform which is not compatible with the VRML97 specifications (but maybe it is not so important to be compatible with VRML97).

Why it is not possible to allow Group but not Pose?

omichel:
> Why it is not possible to allow Group but not Pose?

Because Group would need the absolute scale which would be inherited by Pose as well...

BenjaminDeleze:
> > Why it is not possible to allow Group but not Pose?
> 
> Because Group would need the absolute scale which would be inherited by Pose as well...

I think I am missing something, because now we can have a Transform that have a Group and the Group does not have the absolute scale, right?

ygoumaz (changes requested):
I have tested a lot of worlds and protos and could not find any big issues.

However, I tried to play a little with the Pose and new Transform nodes and had some difficulties understanding the restrictions:
- When trying to add a Node as child of `Transform`, the list of available ones is truncated a lot compared to the `Pose` node (which is probably due to the fact that you cannot scale all objects). But even for a lot of available nodes it is impossible to add them (error message in the console saying `Cannot insert [...] node in 'children' field of Transform node: only a slot can be added in the parent slot. `). So these nodes should not appear in the list of available nodes at the first place.
- These restrictions doesn't seem to be explained in the documentation. It would be nice to explain why `Pose` and `Transform` nodes are different and what is allowed as `Pose` child and restricted in `Transform` children because of the scale (or other reasons).

resources/web/wwi/nodes/utils/WbBoundingSphere.js
omichel: I am not sure as bounding spheres are not the same as bounding objects. They are used for some visibility computation (at least in Webots), so they should probably consider the case where they are inside a Transform node. However, I am not sure if webots.js also use bounding spheres for this kind of purposes.

stefaniapedrazzi: As far as I know, the implementation of the bounding sphere in webots.js matches exactly the one in Webots.
So it seems correct to handle the transform case, as the bounding sphere is computed for geometries that can have Transform parents.

BenjaminDeleze: Yes my bad, I went to fast and read bounding object instead bounding sphere. It seems correct.

omichel:
> I think I am missing something, because now we can have a Transform that have a Group and the Group does not have the absolute scale, right?

Yes, I believe there is something fundamentally wrong in this PR: the scale stuff should be moved out of the Pose node and put inside the Transform node. I am working on it now.

omichel:
I believe I addressed all your comments, so the PR is ready for review again now.

omichel (approved):

BenjaminDeleze (changes requested):
The streaming viewer / animations are still broken (tested locally):
- spot.wbt: `referenceError: child is not defined
    at WbGroup._updateProgress (WbGroup.js:26:36)`
- nao_roome.wbt: `is[#rotation].normalizeRotation is not a function
    at new WbPose (WbPose.js:22:20)`

docs/reference/changelog-r2023.md
```diff
@@ -3,6 +3,7 @@
 ## Webots R2023b
 Released on ??
   - New Features
+    - Added a new Pose node which inherits from Group and derives to Transform and Solid ([#5978](https://github.com/cyberbotics/webots/pull/5978)).
```
stefaniapedrazzi, line 6: Maybe we could also add some links to the node documentation
```suggestion
    - Added a new [Pose](pose.md) node which inherits from [Group](group.md) and derives to [Transform](transform.md) and [Solid](solid.md) ([#5978](https://github.com/cyberbotics/webots/pull/5978)).
```

docs/reference/transform.md
```diff
@@ -1,65 +1,22 @@
...
 
 ### Description
 
-The [Transform](#transform) node is a grouping node that defines a coordinate system for its children that is relative to the coordinate systems of its parent.
+The [Transform](#transform) node is a grouping node that defines a coordinate system for its `children` that is relative to the coordinate systems of its parent.
+The `children` list of a [Transform](#transform) node may only contain [Transform](#transform), [Pose](pose.md), [Group](group.md), [Shape](shape.md) or [CadShape](cadshape.md) nodes.
+It cannot contain any `Solid` or `Solid` derived nodes.
```
stefaniapedrazzi, line 17: Also here it might be good to link to the node page
```suggestion
It cannot contain any [Solid](solid.md) or `Solid` derived nodes.
```

docs/reference/pose.md
```diff
@@ -0,0 +1,48 @@
...
+```
+Pose {
+  SFVec3f    translation     0 0 0     # any vector
+  SFRotation rotation        0 0 1 0   # unit axis, (-inf, inf) angle
+  SFFloat    translationStep 0.01
+  SFFloat    rotationStep    0.261799387
+}
+```
+
+Direct derived nodes: [Solid](solid.md), [Transform](transform.md).
```
stefaniapedrazzi, line 14: Based on the chart data, we should also add the Fluid node:
```suggestion
Direct derived nodes: [Solid](solid.md), [Transform](transform.md), [Fluid](fluid.md).
```

docs/reference/solid.md
```diff
@@ -28,14 +28,11 @@ A [Solid](#solid) node represents an object with physical properties such as dim
 The [Solid](#solid) class is the base class for collision-detected objects.
 Robots and device classes are subclasses of the [Solid](#solid) class.
 In the 3D window, [Solid](#solid) nodes can be manipulated (dragged, lifted, rotated, etc) using the mouse.
+[Solid](#solid) node may be `children` of [Solid](#solid), [Pose](pose.md) and [Group](group.md) nodes.
+They cannot be `children` of a [Transform](transform.md) node.
```
stefaniapedrazzi, line 32: If we add this, then I would also explain why cannot be children of Transform node.
Also note that the list is complete for the `children` field, but there are other nodes that can be parent of the Solid node.

stefaniapedrazzi (commented):
I re-checked the issue with the insertion of the Transform node in joints endPoint and I confirm that is fixed.

BenjaminDeleze (changes requested):
Concerning Webots:
- Not new but still wrong: it is possible to insert a `Slot` in the BO (BO->group->slot) but it will throw and error
- The duck in `animated_skin.wbt` now have a buggy behavior

resources/web/wwi/Parser.js
```diff
@@ -266,7 +267,12 @@ export default class Parser {
         else
           console.error('This world already has a fog.');
         break;
+      case 'Pose':
+        result = this.#parsePose(node, parentNode, isBoundingObject);
+        break;
```
BenjaminDeleze, line 272:
```suggestion
```

Add `case 'Pose':` below to avoid code duplication

resources/web/wwi/Parser.js
```diff
@@ -292,7 +298,7 @@ export default class Parser {
       case 'TouchSensor':
       case 'Track':
       case 'TrackWheel':
-        result = this.#parseTransform(node, parentNode, isBoundingObject);
+        result = this.#parsePose(node, parentNode, isBoundingObject);
```
BenjaminDeleze, line 301:
```suggestion
    case 'Pose':
        result = this.#parsePose(node, parentNode, isBoundingObject);
```

resources/web/wwi/Parser.js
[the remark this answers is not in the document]
BenjaminDeleze: I am not sure about this as the [joint documentation state](https://cyberbotics.com/doc/reference/joint): 
endPoint: this field specifies which [Solid](https://cyberbotics.com/doc/reference/solid) will be subjected to the joint constraints. It must be either a [Solid](https://cyberbotics.com/doc/reference/solid) child, or a reference to an existing [Solid](https://cyberbotics.com/doc/reference/solid), i.e. a [SolidReference](https://cyberbotics.com/doc/reference/solidreference). Alternatively, a [Slot](https://cyberbotics.com/doc/reference/slot) node can be inserted in the endPoint field, but this [Slot](https://cyberbotics.com/doc/reference/slot) should be connected to another [Slot](https://cyberbotics.com/doc/reference/slot) whose endPoint is either a [Solid](https://cyberbotics.com/doc/reference/solid) or a [SolidReference](https://cyberbotics.com/doc/reference/solidreference).

ygoumaz:
Four nodes give unclear warnings when added as children of `Transform` node:
- `E-puckGroundSensors`
- `JetBotRaspberryPiCamera`
- `Servo`
- `TorusBoundingObject`

The warning is the following `Skipped node: Cannot insert [...] node in 'children' field of Pose node.` which is not correct, as we are in a `Transform`. The 3 first ones give a clearer error when added at the root of the scene tree: `Skipped node: Cannot insert [...] node in 'children' field of Pose node that doesn't have a Robot ancestor.`

`TorusBoundingObject` gives the same unclear warning when added at the root of the scene tree.

stefaniapedrazzi:
The scale of the lines in the RoadIntersection PROTO in the `humanoid_marathon.wbt` simulation is much bigger in this branch than on the develop branch.

humanoid_marathon_branch

develop:
humanoid_marathon_develop

BenjaminDeleze (changes requested):
The `FieldModel.js` file needs to be regenerated (by running the `basenode_encoder.py` script)

stefaniapedrazzi:
Tutorial 5 (https://www.cyberbotics.com/doc/guide/tutorial-5-compound-solid-and-physics-attributes) needs to be updated because describes the use of a Transform node that it now turned into Pose.

BenjaminDeleze:
WebotsJS:

Fix bugs preventing the loading of the word:
  - The variable `child` of the `_updateProgress` of `WbGroup.js` was not defined
  - The argument of `clone` were not in the right order for `WbTransform.js`
  - Solid/Robot,... were still exported as Transform
  - The constructor of the devices still had the scale parameter
  - The web proto parser did not support `Pose`

I also took the opportunity to simplify the x3d export.
Before we were trying to be kind of compliant with VRML: export only nodes that exist within the VRML standard. 
For this reason Solid,Robot,... were exported as Transform and given an additional parameter (type) to allow webotsJS to know if it was a Transform or in fact a Solid/Robot/Track,...
However, since some times, we are not compliant with VRML anymore (Capsule, CadShape, ...). 
Moreover, with the introduction of the web proto parser, we implemented the parsing of the missing nodes (Solid, Robot, Lidar,...) and support both ways: <Transform type="Solid"> and <Solid>
As we are breaking the backward compatibility with the introduction of the Pose node, I tought it was the right moment to drop the "Transform type=solid" and to export directly the real name of the node .

I am still fixing two issues:
- [ ] the bounding sphere seems to be wrong (certainly due to some confusion in upperPose/upperTransform)
- [ ] the tracks are no longer working (even prior to my simplification of the export)

omichel (dismissed):
I tested the webotsjs fixes and it seems to work perfectly for me.
I just found and fixed a bug with the panda demo.

ygoumaz (dismissed):
I haven't found any new problem on my side. Thank you!

omichel:
@stefaniapedrazzi and @BenjaminDeleze, shall we merge this PR now?

omichel (dismissed):

BenjaminDeleze (changes requested):
- The documentation still need some adjustments (reference/BallJointParameters.md, reference/Muscle.md, guide/Ned.md, ...) I found them with `grep` .  Note that if you modify some "robotName.md", you also need to report the changes to `projects/.../robotName/protos/docs/robotName.md`
- Not sure if it is still used but `pose` is missing from here: https://github.com/cyberbotics/webots/blob/f19a55984f7e90f5cdbf2b3973181ce669d9173e/src/webots/core/WbLanguage.cpp#L707

- The `animatedGeometry` field of `Track` should accept the `Pose` node. It is not related to this PR but I think it should also allow `CadShape`.

- The naming of many variables (and some comments) is still wrong. I made the comments for `src/webots/nodes/utils/WbNodeUtilities.cpp` and some other but there is a lot more, for example in: `WbSupervisorUtilities.cpp`, `WbAnimationRecorder.cpp`, `WbPositionViewer.cpp`, `WbPose.cpp`

BenjaminDeleze (changes requested):
The `.x3d`  files of the documentation should be regenerated.

The documentation still contain incorrect references to `Transform`: `/doc/reference/node-chart.md`, `docs/reference/structure-of-ode-objects.md`, and others

Otherwise it seems good to me.

BenjaminDeleze:
The test doc is failing.

BenjaminDeleze:
I have made some corrections in the docs.

Moreover it seems like scaling transform via `alt` + handle does not work anymore, also the `show resize handles` button is not displayed anymore.

Screenshot from 2023-05-10 09-03-53

I could crash Webots by doing:

1. Open a World with a table proto (like PR2.wbt).
2. Convert the table to base node.
3. Select the root node of the table.
4. Convert it to Transform => crash.

BenjaminDeleze (approved):
Seems good to me

omichel (approved):

stefaniapedrazzi:
I didn't find any major issue.

* However, I'm not sure it is good to allow to transform a Solid node into a Transform node (as in the example of the `Table described by @BenjaminDeleze) is the Solid has Solid children because unexpectedly all the children are removed.
I think it would be more user-friendly to prevent this operation.

* The `c3d_viewer.py` controller is throwing an error because it imports a `C3dBocyRepresentation` with the `scale` field that has been removed.

omichel:
> However, I'm not sure it is good to allow to transform a Solid node into a Transform node (as in the example of the Table described by @BenjaminDeleze) if the Solid has Solid children because unexpectedly all the children are removed.
I think it would be more user-friendly to prevent this operation.

Yes, we discussed this point with @BenjaminDeleze yesterday and I believe we could indeed remove this option. However, since there is a warning about loss of information during this operation, I believe it is okay for the user. And also, it seems odd to allow users to transform a Transform into a Solid, but prevent the other way round. For example, with the BiscuitBox, it works better than with the Table because the children of the Solid are Pose nodes. Hence, in that case, you can transform the BiscuitBox to a Solid, and to a Transform and back to a Solid without loosing the geometries (but you will loose the physics, boundingObjects, etc.).

stefaniapedrazzi:
Loosing boundingObject or physics is expected and not as problematic as loosing all the children information.
Personally, I don't think that the info is enough in this case and should be more specific.
It is not possible to undo a node transform so it is really annoying to revert it.
Even more because this is a new behavior and was working the same in previous versions.

ygoumaz:
I agree with @stefaniapedrazzi that the warning should at least be more specific and list the information that will be lost, in particular children Solid objects.

omichel:
I am improving this warning...

omichel (approved):

omichel:
I just fixed the two issues reported by @stefaniapedrazzi. I guess we can merge it now?

ygoumaz:
When changing the scale of a simple Transform node, it affects the translation and rotation arrows in the view in a wrong way that is not present in develop. It makes them unusable.

ygoumaz:
Ok, I still can produce a different behavior between both branches:
1. Adding a Transform at the root of the scene tree.
2. Changing the scale in develop doesn't affect handles, but in feature-pose-node, the handles are scaled too.

However, this additional behavior may be a side effect from the existing bug and not directly related to the addition of the Pose node. We may merge the PR and fix it when addressing #6063.

stefaniapedrazzi (approved):

ygoumaz (approved):

PR #5979, opened 2023-03-13
WebotsJS: Comparison without value

BenjaminDeleze:
Introduce a more permissive definition of `equals` to use in the context of restrictions.

Example:
In the `AddLaneRoadSegment` proto, the `lines` parameter is restricted and accept only one type of node: `RoadLine { }`.

It seems to work as intended at first glance: we can insert new `RoadLine { }`.
However, if we modify a `RoadLine { }`, let's say we change the line `type` from `dashed` to `double` and later we trigger a regeneration of the proto, it will not work.
This is the case because now we are comparing `RoadLine { }` with  `RoadLine { type double }`


ad-daniel (changes requested):
seems overkill to even use `value.equals` at this point, to be honest. In the context of restrictions it should suffice to check that the `.url` coincides between the two nodes. That should ensure that the PROTO being inserted is the same one as the allowed one (same url), and in the case of basenodes, the `url` is the nodes name itself, so it should work as well.

BenjaminDeleze:
> seems overkill to even use `value.equals` at this point, to be honest. In the context of restrictions it should suffice to check that the `.url` coincides between the two nodes. That should ensure that the PROTO being inserted is the same one as the allowed one (same url), and in the case of basenodes, the `url` is the nodes name itself, so it should work as well.

Seems good to me, I will do the change

PR #5980, opened 2023-03-13
Fix some issues related to scaled objects

stefaniapedrazzi:
Some issues arise when an ancestor node has some scaling factor.

Fixes:
* [x] fix bounding sphere's radius scale for Capsule, Cone, Cylinder, and Plane:
  the radius is only scaled to absolute size for these geometries but not for the others (Box, ElevationGrid, IFS, etc.) causing issues and mismatching
  For example the the size of the following object is not correctly detected:
  ``Transform { scale 0.25 0.25 0.25 children [ Solid {children [ Shape { geometry Cylinder { height 4 radius 1.5 } } } } ] }``
* [x] fix computation of bounding sphere of Cone and Cylinder not converted to ENU.
* [x] fix assertion failure in `WbMathsUtilities::clampedAcos` when computing rotation from matrix:
    this could occur if a different scale is computed from node hierarchy than the one defined in the matrix.
    I get this assertion failure when selecting a Solid node with a scaled ancestor Transform node (``Solid -> scaled Trasform > Solid > Solid``)  and clicking on the Position tab in the Field Editor.
* [x] adjust the tests (and check the new behavior is correct) and improve viewpoint position of the checked tests

_Note: Pen ray tracing is wrong with non-uniform scale even in the master branch and will be fixed in a different PR (#5982)._

I didn't add any entry in the ChangeLog because these bugs were not causing any directly visible bugs.


ad-daniel (changes requested):
Looks good to me, just some comments:
- Is it normal that the cone is not entirely encapsulated by the bounding sphere? Isn't the point of the bounding sphere precisely that?
- Not an issue at the moment, but when the synchronization between branches takes place it will introduce a disparity with respect to webotsjs. In fact the develop branch the `recomputeBoundingSphere` methods have been ported to the web, so it needs to be fixed there as well. In short, we can't rely on the automatic sync in this case but should be done manually when this is merged.

stefaniapedrazzi:
> Is it normal that the cone is not entirely encapsulated by the bounding sphere? Isn't the point of the bounding sphere precisely that?

No. The formula to compute the bounding sphere radius is wrong (in master too). I'm now fixing it.

ad-daniel (approved):
Works now, thank you

PR #5981, opened 2023-03-13
Fix `mavic2pro_patrol` controller

stefaniapedrazzi:
The drone should not change altitude when printing the log that the target has been reached.


ad-daniel (approved):
Thank you

Issue #5451, opened 2022-11-09
test_world not working properly


ygoumaz:
`test_world` in the CI should exit with an error if there is a diff between the original world file and the saved one. This doesn't seem to be the case as it was not detecting the diff on the `boomer.wbt` world with useless PROTOs (see #5450).

PR #5983, opened 2023-03-13
Fix test worlds

ad-daniel:
**Description**

- [x] Fixes https://github.com/cyberbotics/webots/issues/5451
- [x] Improve cache test in order to know which worlds have non-cached assets
- [x] Fix recent nightly issues


ygoumaz:
The warnings test fail correctly when unwanted warnings are detected. The cache problem is also solved.

However, `robot_programming.wbt` is changed after save and this is not catched by the test-suite workflow which is supposed to check for the diff.

PR #5994, opened 2023-03-15
Fix bounding sphere radius for Sphere and CadShape

stefaniapedrazzi:
Complete #5980:
fix bounding sphere radius for Sphere and CadShape geometries that are still scaled.


ad-daniel (approved):
Thank you

PR #5995, opened 2023-03-15
Fix webotsjs type in parser

BenjaminDeleze:
**Description**
Some floats/int were passed as string.
It was undetected until now because in most operation JS succeed to use string instead of number, for example:
```
"3" / 2 = 1.5
```

But it results in some `NaN` in #5985


omichel (approved):
Nice.

PR #5999, opened 2023-03-16
Fix instructions for the location of the webots.yml file

omichel:
The `webots.yml` file should be located at the root level of the project, not the repository.


BenjaminDeleze (approved):
Nice catch

PR #6001, opened 2023-03-17
Fix webjs undefined

BenjaminDeleze:
In the case of animation/scene/proto with local textures, we do not have a `worldsPath`


omichel (approved):
We should also do the same for the forged proto URLs from https://webots.cloud/proto

PR #6002, opened 2023-03-17
Simplify tmp directory computation for extern controllers

### LINKED ISSUE #5300: Search for Webots tmp directory in libController is not robust
If a snap installation of Webots is found on the system but another Webots (git clone, tarball, or debian) version is currently executed, the libController uses the snap `tmp` directory for connecting extern controllers and fails to connect instead of checking if the Webots instance files exists and reverting to the system `tmp` folder.

ygoumaz:
Fixes #5300.

The `wbu_system_dir()` function is only used to get the tmp dir when using **ipc extern** controllers. The `WEBOTS_TMPDIR` is not useful and can be discarded in this case. Instead we use `WEBOTS_HOME` to determine if we target a snap installation of Webots. If not we default to `/tmp`.

This PR also updates the `test_sources` to Ubuntu 22 to fix some cppcheck false positives. The clang-format version is locked to 14.0.0.


omichel (approved):
Looks good to me.
Thank you.

src/webots/nodes/WbCharger.cpp
```diff
@@ -104,7 +104,8 @@ void WbCharger::updateMaterialsAndLights(double batteryRatio) {
     WbPbrAppearance *appearance = dynamic_cast<WbPbrAppearance *>(visualElement->node);
     WbLight *light = dynamic_cast<WbLight *>(visualElement->node);
     const WbRgb color(cr, cg, cb);
-    assert(!WbRgb(cr, cg, cb).clampValuesIfNeeded());
+    const bool clampNeeded = WbRgb(cr, cg, cb).clampValuesIfNeeded();
+    assert(!clampNeeded);
```
omichel, line 108: This broke the nightly builds, see https://github.com/cyberbotics/webots/actions/runs/4527046581/jobs/7972592389

src/webots/nodes/WbRobot.cpp
```diff
@@ -158,7 +158,18 @@ WbRobot::WbRobot(WbTokenizer *tokenizer) : WbSolid("Robot", tokenizer) {
   init();
 }
 
-WbRobot::WbRobot(const WbRobot &other) : WbSolid(other) {
+WbRobot::WbRobot(const WbRobot &other) :
+  WbSolid(other),
+  mControllerDir(),
```
stefaniapedrazzi, line 163: All these initialization lines are useless and should be removed because QString are already initialized by default.

src/webots/nodes/WbReceiver.cpp
```diff
@@ -117,7 +117,7 @@ WbReceiver::WbReceiver(WbTokenizer *tokenizer) : WbSolidDevice("Receiver", token
   init();
 }
 
-WbReceiver::WbReceiver(const WbReceiver &other) : WbSolidDevice(other) {
+WbReceiver::WbReceiver(const WbReceiver &other) : WbSolidDevice(other), mTransmissionList(), mWaitingQueue(), mReadyQueue() {
```
stefaniapedrazzi, line 120: The newly added initialization instructions are useless and should be removed.

src/webots/nodes/WbSkin.cpp
```diff
@@ -81,7 +81,20 @@ WbSkin::WbSkin(WbTokenizer *tokenizer) : WbBaseNode("Skin", tokenizer), WbAbstra
...
+  mInitialSkeletonOrientation(),
+  mInitialSkeletonPosition(),
+  mRenderables(),
+  mMaterialNames(),
+  mMaterials(),
+  mSegmentationMaterials(),
+  mEncodeDepthMaterials(),
+  mMeshes(),
+  mBoneTransforms(),
+  mBonesMap() {
```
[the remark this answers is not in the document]
ygoumaz: cppcheck triggers an error if they are missing.

src/webots/nodes/WbReceiver.cpp
ygoumaz: cppcheck triggers an error if they are missing.

src/webots/nodes/WbRobot.cpp
ygoumaz: cppcheck triggers an error if they are missing.

src/webots/nodes/WbLightSensor.cpp
```diff
@@ -134,7 +134,7 @@ WbLightSensor::WbLightSensor(WbTokenizer *tokenizer) : WbSolidDevice("LightSenso
   init();
 }
 
-WbLightSensor::WbLightSensor(const WbLightSensor &other) : WbSolidDevice(other) {
+WbLightSensor::WbLightSensor(const WbLightSensor &other) : WbSolidDevice(other), mRayList() {
```
stefaniapedrazzi, line 137: It useless to explicitly initialize `mRayList` to the default value.

src/webots/nodes/WbLed.cpp
```diff
@@ -38,7 +38,7 @@ WbLed::WbLed(WbTokenizer *tokenizer) : WbSolidDevice("LED", tokenizer) {
   init();
 }
 
-WbLed::WbLed(const WbLed &other) : WbSolidDevice(other) {
+WbLed::WbLed(const WbLed &other) : WbSolidDevice(other), mMaterials(), mPbrAppearances(), mLights() {
```
stefaniapedrazzi, line 41: here as well, the new initialization instructions are useless.

src/webots/nodes/WbRobot.cpp
stefaniapedrazzi: Ok. But the cppcheck error/warning is wrong.

ygoumaz: I don't know if using a lot of suppress tags for cppcheck is really the best solution. Is it really that bad to initialize those QString objects?

stefaniapedrazzi: It is completely useless to explicitly initialize the strings and all the other Qt objects
From my point of view it makes the code less readable than before because we have a long list of useless instructions.
I think that cppcheck should help us writing better code, not worse and lately it is not always the case.
We should try to adjust the configuration to try to avoid these false positives.

PR #6005, opened 2023-03-20
Fix test-suite

BenjaminDeleze:
Error: `AttributeError: module 'lib' has no attribute 'X509_V_FLAG_CB_ISSUER_CHECK'`


ad-daniel (changes requested):
Although we don't enforce anymore the script that ensures master and develop workflows are (nearly) the same, I think we still try to keep as little difference between the two, so the fix should also be done in the develop version of the workflow. Either way, I suspect that develop will have the same issues when uploading the nightly, no?

ad-daniel (approved):
Thank you

PR #6006, opened 2023-03-20
Fix flu vehicles

BenjaminDeleze:
**Description**
When the vehicles where converted in FLU, we introduced additional rotations instead of rotating the meshes/subprotos

For example, in `MotorbikeSimple` we introduced a rotation in the driver slot such that we did not need to convert the driver.

However, as the driver will be a proto that will be displayed on webots.cloud, I fully converted it to be displayed with the correct orientation.

I did that for:

- TruckTrailer
- TruckTank
- ScooterDriver
- MotorbikeDriver

By doing so, I had to touch the `VehicleWheel.proto`... and so basically every other vehicle protos needed to be adjusted.

Note: I expect that some other IndexedFaceSet are not converted in the vehicles but that we only added a rotation to the transform/solid. However, as it has no impact on the rendering, it will stay as is.


stefaniapedrazzi (approved):
Thank you!

stefaniapedrazzi:
oops, I didn't check the test suite results.
Changes are fine, but some tests needs to be adjusted.

stefaniapedrazzi (approved):
Thank you

PR #6010, opened 2023-03-21
Bump windows version of sumo

BenjaminDeleze:
ditto


ygoumaz (changes requested):
It works fine, thank you! I would just add a line in the changelog about this version update.

ygoumaz (approved):

PR #6014, opened 2023-03-22
Merge `master` into `develop`

ad-daniel:
Propagate test worlds fix to develop.

No conflicts.


ad-daniel:
Seems that develop has content which master does not, it makes the test worlds fail

PR #6015, opened 2023-03-22
Merge `master` into `develop`

ad-daniel:
**Description**

No conflicts, but additional fixes were necessary to the worlds.


ad-daniel:
For future reference: although relative assets can be used when defining worlds, the officially distributed ones (i.e. anything under `projects`) should define assets exclusively with `webots://`, since the assets themselves are not distributed, relative urls won't work.

omichel (approved):
Thank you.

PR #6026, opened 2023-03-29
Display in the tree the name of the solid at the extremity of a joint

ShuffleWire:
**Description**
In order to ease the navigation in the scene tree, especially with a lot of element, I propose to add the name of the joint-ed solid, like that : 

image

instead of 

image

in projects/sample/devices/worlds/led.wbt

Also displaying "to ref" instead of "to" when it's a SolidReference


ad-daniel:
as this isn't a bug fix, it would be better to target develop

stefaniapedrazzi (commented):
Thank you for proposing this change.
It is good to add additional information to better identify the joints (and slots).

However, personally I don't like very much the way the endPoint name is displayed in this branch by using "to" and "to ref" the items in the scene tree now become descriptions rather just nodes and fields type names.
I found quite confusing and not very user-friendly that depending on the node settings the node type/name changes.
I think that the node type/name and further info helping identifying it should be better separated in some ways.
In case of Solid nodes it is a more clear because the name is appended in quotes.

Additionally, if an `endPoint node has a DEF names, we should rather use the DEF name instead of the `name` field value to be consisted with the existing code.

ShuffleWire:
> Thank you for proposing this change. It is good to add additional information to better identify the joints (and slots).
> 
> However, personally I don't like very much the way the endPoint name is displayed in this branch by using "to" and "to ref" because the items in the scene tree now become descriptions rather just nodes and fields type names. I found quite confusing and not very user-friendly that depending on the node settings the node type/name changes. I think that the node type/name and further info helping identifying it should be better separated in some ways. In case of Solid nodes it is a more clear because the name is appended in quotes.
> 
> Additionally, if an `endPoint` node has a DEF names, we should rather use the DEF name instead of the `name` field value to be consisted with the existing code.

yes, display was just an idea ! it's just that displaying only the "name" of the endPoint Solid is a bit confusing because it's not the name of the joint... but it's maybe not so important !
Do you want I remove it ?

Yes displaying the def could be nice as well

stefaniapedrazzi:
Yes, I think it would be better to remove it.

Just to summarize, here are the changes needed to complete this PR:
* remove the "to"/"to ref" part
* use the DEF if available instead of the "name" field
* fix the ccpcheck error reported by the `test sources` check
* add an entry in the "Enhancements" section of the ChangeLog:
  https://github.com/cyberbotics/webots/blob/develop/docs/reference/changelog-r2023.md#webots-r2023b

ShuffleWire:
Even if it work, I've notice that changing the DEF of the endpoint don't immediately triggers a change of the displayed name of the joint... don't know why...

stefaniapedrazzi:
> Even if it work, I've notice that changing the DEF of the endpoint don't immediately triggers a change of the displayed name of the joint... don't know why...

This is because we should explicitly notify that a change in DEF name should trigger the scene tree update.
I just added the required instruction in this PR.

Additionally, instead of displaying `DEF_NAME Solid`, in the scene tree it should only display `"DEF_NAME"`.

stefaniapedrazzi:
I finalized this PR.
It would be better if someone else could review it.

ygoumaz (approved):
Works nicely! Thank you.

docs/reference/changelog-r2023.md
```diff
@@ -10,6 +10,9 @@ Released on ??
   - Cleanup
     - Deprecated the C and MATLAB API functions `wb_supervisor_node_enable/disable_contact_point_tracking` in favor of `wb_supervisor_node_enable/disable_contact_points_tracking` to be more consistent with other APIs ([#5633](https://github.com/cyberbotics/webots/pull/5633)).
   - Enhancements
+    - Improved the message error path by including the [Solid](solid.md) name / DEF name ([#6026](https://github.com/cyberbotics/webots/pull/6026)).
```
omichel, line 13: Do you mean the "error message path" instead?

PR #6028, opened 2023-03-29
Display name of object to make path easier to follow

ShuffleWire:
When tree is deep, multiple object with the same type could be found at some level. 
hence log error path is sometimes difficult to follow.

This MR is relevant with https://github.com/cyberbotics/webots/pull/6026
Caution : I've been basically copy pasting some code between those two MR, maybe some code factoring is needed here...

For instance : 

WARNING: Robot "MyBot" > HingeJoint to "left wheel" > Solid "left wheel" > DEF WHEEL Transform > Shape > Cylinder: Invalid 'height' changed to 1. The value should be positive.

should be better than 

WARNING: Robot > HingeJoint > Solid > DEF WHEEL Transform > Shape > Cylinder: Invalid 'height' changed to 1. The value should be positive.

Maybe this example don't emphasis enough the advantage of the additional information, but you should get my point, especially when working with large and deep tree.


stefaniapedrazzi (changes requested):
The idea is nice!
But I would slightly refine it to be consistent with what is displayed in the scene tree:
* if a DEF name is set it doesn't seem necessary to also show the "name" field
* if an "endPoint" node has a DEF name I would use the DEF name instead of the "name" field

Then you should also fix the cppcheck errors and add and entry in the ChangeLog.

ShuffleWire:
it's done, but considering I've been copy pastring code, is there a way to use, for instance QString WbNode::usefulName() in https://github.com/cyberbotics/webots/pull/6026 ?

PR #6035, opened 2023-03-30
Fix setAcceleration error in Python.

songyangZhao:
**Description**
Fix setAcceleration bug in Python.


ygoumaz:
Thanks for the fix! This PR should target the master branch, as it fixes a bug. In addition, you should add an entry in the [changelog](https://github.com/cyberbotics/webots/blob/master/docs/reference/changelog-r2023.md) about this fix.

songyangZhao:
You are welcome! So maybe I should close this PR and open a new one based on master branch.

PR #6043, opened 2023-03-31
Update cppcheck to 2.10 on Ubuntu

ygoumaz:
Update cppcheck to 2.10 on ubuntu to match versions between platforms. Also, restrict cppcheck to given platforms (instead of all configurations on all platforms, which is redundant). 

With this update, we can revert useless empty strings initializations (see https://github.com/cyberbotics/webots/pull/6002#discussion_r1148862522)

### DIFF
--- a/tests/sources/test_cppcheck.py
+++ b/tests/sources/test_cppcheck.py
@@ -35,7 +35,7 @@ class TestCppCheck(unittest.TestCase):
         elif (sys.platform.startswith('win32')):
             self.platformOptions = ' -D_WIN32'
         else:
-            self.platformOptions = ' -U__APPLE__'
+            self.platformOptions = ' -D__APPLE__'
 
     def test_cppcheck_is_correctly_installed(self):
         """Test Cppcheck is correctly installed."""
@@ -172,7 +172,8 @@ class TestCppCheck(unittest.TestCase):
             'projects/vehicles/controllers/ros_automobile/include'
         ]
         skippedfiles = [
-            'projects/robots/robotis/darwin-op/plugins/remote_controls/robotis-op2_tcpip/stb_image.h'
+            'projects/robots/robotis/darwin-op/plugins/remote_controls/robotis-op2_tcpip/stb_image.h',
+            'projects/robots/epfl/lis/plugins/physics/blimp_physics/utils.h'
         ]
         command = 'cppcheck --platform=native --enable=warning,style,performance,portability --inconclusive'
         command += self.platformOptions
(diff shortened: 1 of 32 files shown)


tests/sources/test_cppcheck.py
```diff
@@ -163,9 +172,11 @@ def test_projects_with_cppcheck(self):
             'projects/vehicles/controllers/ros_automobile/include'
         ]
         skippedfiles = [
-            'projects/robots/robotis/darwin-op/plugins/remote_controls/robotis-op2_tcpip/stb_image.h'
+            'projects/robots/robotis/darwin-op/plugins/remote_controls/robotis-op2_tcpip/stb_image.h',
+            'projects/robots/epfl/lis/plugins/physics/blimp_physics/utils.h'
```
stefaniapedrazzi, line 176: Just for info, why this file now needs to be ignored?
Was it working correctly for macOS and Windows cppcheck before, isn't it?

stefaniapedrazzi (commented):
Thank you!
Everything seems fine to me.
I'm now checking if the same is also working locally.

I also found some other files where there is default initialization in the initialization list that could be removed.
Can I add them to this PR?

tests/sources/test_cppcheck.py
ygoumaz: This is because we now only run cppcheck on the source code with `_WIN32`, `__linux__` and `__APPLE__` preprocessor variables, while before it was running all possible configurations, including other variables. In `utils.h`, there is an error thrown if neither `dSINGLE` or `dDOUBLE` are defined. This is why this file is skipped now.

ygoumaz:
Yes, you can directly push your changes to this PR.

stefaniapedrazzi:
I tested locally but I'm receiving many errors (like the following one):
```
src/webots/build/debug/WbRobotWindow.moc.cpp:14:0: error: #error "The header file 'WbRobotWindow.hpp' doesn't include <QObject>." [preprocessorErrorDirective]
#error "The header file 'WbRobotWindow.hpp' doesn't include <QObject>."
```

It seems we should also skip the `build` folder.
I assume these errors don't occur on the CI because Webots is not compiled and the `build` folder is empty.

ygoumaz:
Yes, you are right, the build folder doesn't exist in the CI. I will add the exception on the build folder, as it doesn't make a lot of sense to test those files.

stefaniapedrazzi:
Thank you!

I tested the latest version and I'm still getting some of these errors due to the MOC files for a couple of other folders.
I just commit the changes so that you can review them.

stefaniapedrazzi (approved):
Thank you!

PR #6047, opened 2023-04-03
Fix mesh bounding object

BenjaminDeleze:
**Description**
When using a mesh boundingObject (bad practice), the bounding object is not taken into account if the mesh url is remote and the mesh is not present in the cash because we download the mesh after creating the ode object.

**Fix**

- [x] When the object is in the world when we open it.
- [x] When the object is added after the initial loading.


ygoumaz (dismissed):
Thank you! Is this bug present in R2023a? If yes, an entry should be added in the changelog.

Issue #4906, opened 2022-07-18
Wrong "Bounding Object Based" MoI


ShuffleWire:
**Describe the Bug**
Using a "Bounding object based" MoI give values that are different from the default MoI values.

<details>
<summary>World</summary>
<pre>
#VRML_SIM R2022a utf8
WorldInfo {
  lineScale 10
}
Viewpoint {
  orientation 0.18483272608525197 0.34484097093526966 -0.9202834172863895 1.2065546190350083
  position -8.08416062735125 16.7740818030343 8.768359541015466
}
TexturedBackground {
}
TexturedBackgroundLight {
}
Robot {
  children [
    DEF s Shape {
      appearance PBRAppearance {
      }
      geometry Sphere {
      }
    }
  ]
  boundingObject USE s
  physics Physics {
    density -1
    mass 1
  }
}
</pre>
</details>

**Steps to Reproduce**
1. Open the world, note that in the "Mass" tab of the Solid the MoI is [0.4,0.4,0.4, 0,0,0] 
That is correct
2. RIght click on "physic/inertiaMatrix" -> Add New -> Bounding Object Based -> Add
3. Observe that the MoI in the "Mass" tab of the Solid change to [1675.52,1675.52,1675.52, 0,0,0]
That change should not occurs, the object haven't changed. This is because the inserted matrix is [1675.52,1675.52,1675.52, 0,0,0]

**Expected behavior**
The MoI inserted into the field inertiaMatrix should be [0.4,0.4,0.4, 0,0,0], to stay consistent with the MoI initially displayed in the "Mass" tab

**System**
 - Operating System: Debian 11
 - Webots: 2022a

stefaniapedrazzi:
This seems the exact same issue as reported in  #4684.

ShuffleWire:
I don't think so :  
In #4684 , I explaining a bug that is only visible until reload.
Here, the bug is when **generating** a MoI based on the current Bounding Object using the tool discussed in the OP.
Whenever you reload or not don't impact the MoI (being actually written in the wrl file)

ShuffleWire:
Additionally, changing the mass of the Physic don't affect the inertiaMatrix, even if the code suggest that it should change.
`
void WbPhysics::updateMass()
  ...
  else if (mMode == BOUNDING_OBJECT_BASED)
    emit massOrDensityChanged();
`
But on that part, I will not be as certain...

ShuffleWire:
[...]

PR #6048, opened 2023-04-03
Ensured correct mass is set during MoI matrix generation from bounding object #4906

Chandan-Bharadwaj:
**Description**
The MoI matrix generated based on the bounding box was incorrrect due to incorrect total mass being considered while populating the matrix. This is fixed by rectifying the total mass after the reference mass (dMass object) is calculated. The reference mass is calculated by adding derived masses for each of the primitives inside the bounding object (considering a reference density for each of the primitives).

**Related Issues**
This pull-request fixes issue #4906

### DIFF
--- a/src/webots/nodes/WbSolid.hpp
+++ b/src/webots/nodes/WbSolid.hpp
@@ -355,7 +355,6 @@ private:
   bool mUseInertiaMatrix;  // indicates that the WbSolid uses the latest valid inertia matrix field for ODE physics computation
   WbVector3 mCenterOfMass;
   WbVector3 mScaledCenterOfMass;
-  const double referenceDensity = 1000.0;
 
   // ODE mass adjustments
   void createOdeMass(bool reset = true);
--- a/src/webots/nodes/WbSolid.hpp
+++ b/src/webots/nodes/WbSolid.hpp
@@ -355,6 +355,7 @@ private:
   bool mUseInertiaMatrix;  // indicates that the WbSolid uses the latest valid inertia matrix field for ODE physics computation
   WbVector3 mCenterOfMass;
   WbVector3 mScaledCenterOfMass;
+  const double referenceDensity = 1000.0;
 
   // ODE mass adjustments
   void createOdeMass(bool reset = true);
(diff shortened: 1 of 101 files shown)


stefaniapedrazzi (changes requested):
Thank you for this nice contribution!

I just added some comments about the coding style.
Please also check that the tests pass after the modifications.

From the tests result it also seems that some of the tests needs to adjusted.

Could you please also add an entry in the ChangeLog?
https://github.com/cyberbotics/webots/blob/master/docs/reference/changelog-r2023.md

src/webots/nodes/WbSolid.hpp
```diff
@@ -355,6 +355,7 @@ protected slots:
   bool mUseInertiaMatrix;  // indicates that the WbSolid uses the latest valid inertia matrix field for ODE physics computation
   WbVector3 mCenterOfMass;
   WbVector3 mScaledCenterOfMass;
+  const double referenceDensity = 1000.0;
```
stefaniapedrazzi, line 358: Thank you for this change!

Following our style guidelines (https://github.com/cyberbotics/webots/wiki/CPP-Coding-Style/#cs305-global-constant-names-including-enum-values-must-be-all-uppercase-using-underscore-to-separate-words), this should be changed to uppercase. It should also be moved to the `WbSolid.cpp` file close to the definition of `MASS_ZERO_THRESHOLD`.
```suggestion
  static const double REFERENCE_DENSITY = 1000.0;
```

stefaniapedrazzi (dismissed):
Thank you!

The tests failures seems unrelated. I will check if it will fail again next time.
Could you please add the change log entry directly in this branch?

src/webots/nodes/WbSolid.hpp
Chandan-Bharadwaj: sure, i will make this change.

stefaniapedrazzi:
Thank you!
It just remains to add the ChangeLog to this PR.
@Chandan-Bharadwaj  you can just apply the changes I reported in #6053 merge #6053 into this branch.

Chandan-Bharadwaj:
@stefaniapedrazzi , i hope the changes can be merged now. Please let me know is there is any further revisions pending

stefaniapedrazzi (approved):
Thank you!

ShuffleWire:
It have been tested on Webots 2023a-rev1 and it works, thanks you
I'm closing it.

PR #6051, opened 2023-04-04
Enhancement camera recognition occlusion

stefaniapedrazzi:
The existing object detection code uses just one ray pointing at the object center to determine if an object is occluded or not.
This method is quite limited and in complex scene often results in missing many objects that are visible.

I investigated using multiple rays (8 for bounding boxes and 6 for bounding sphere) for determine the object occlusion and this method gives much better results for example in the following agricultural environment.

Using 1 ray:
Screenshot from 2023-04-04 18-16-32

Using 6/8 rays:
Screenshot from 2023-04-04 18-18-45

However the performance drops from 0.4x to 0.25x.
In this example there are many objects to be detected so a performance drop is quite normal.
But no difference is visible in the simple camera_recognition.wbt simulation.

### DIFF
--- a/docs/reference/camera.md
+++ b/docs/reference/camera.md
@@ -1417,7 +1417,7 @@ structs.WbCameraRecognitionObject.members = struct(
 
 The `id` represents the node id corresponding to the object, and it is possible to use this id directly in the [`wb_supervisor_node_get_from_id`](supervisor.md#wb_supervisor_node_get_from_def) supervisor function.
 The `position` and `orientation` are expressed relatively to the camera (the relative position is the one of the center of the object which can differ from its origin) and the units are meter and radian.
-The `size` represents the X and Y sizes in meters relatively to the camera (it is of course impossible to know the depth of the object).
+The `size` represents the Y and Z sizes in meters relatively to the camera (it is of course impossible to know the depth of the object along the [Camera](camera.md) X axis).
 The `position_on_image` and `size_on_image` can be used to determine the bounding box of the object in the camera image, the units are pixels.
 The `number_of_colors` and `colors` returns respectively the number of colors of the objects and pointer to the colors array, each color is represented by 3 doubles (R, G and B), therefore the size of the array is equal to 3 * `number_of_colors`.
 Finally `model` returns the `model` field of the [Solid](solid.md) node.
--- a/src/webots/nodes/utils/WbObjectDetection.hpp
+++ b/src/webots/nodes/utils/WbObjectDetection.hpp
@@ -23,6 +23,7 @@
 class WbAffinePlane;
 class WbBaseNode;
 class WbSolid;
+class WbOdeGeomData;
 
 class WbObjectDetection {
 public:
@@ -33,14 +34,15 @@ public:
 
   bool hasCollided() const;
   WbSolid *device() const { return mDevice; }
-  dGeomID geom() const { return mGeom; }
+  const QList<dGeomID> &geoms() const { return mRayGeoms; }
+  bool contains(dGeomID rayGeom) const { return mRayGeoms.contains(rayGeom); }
   WbSolid *object() const { return mObject; }
   const WbVector3 &objectSize() const { return mObjectSize; }
   const WbVector3 &objectRelativePosition() const { return mObjectRelativePosition; }
 
-  void setCollided(double depth);
+  void setCollided(dGeomID geom, double depth);
 
-  void deleteRay();
+  void deleteRays();
 
   // Recomputes ray position and direction and returns if the current ray is valid or can be removed.
   bool recomputeRayDirection(const WbAffinePlane *frustumPlanes);
@@ -51,6 +53,9 @@ public:
   // Computes the frustum plane for the given device ray.
   static WbAffinePlane *computeFrustumPlanes(const WbSolid *device, const double verticalFieldOfView,
                                              const double horizontalFieldOfView, const double maxRange);
+
+  // Return corners of the bounding box/sphere of the object
+  QList<WbVector3> computeCorners() const;
 
 private:
   static void mergeBounds(WbVector3 &referenceObjectSize, WbVector3 &referenceObjectRelativePosition,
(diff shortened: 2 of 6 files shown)


docs/reference/camera.md
```diff
@@ -1417,7 +1417,7 @@ structs.WbCameraRecognitionObject.members = struct(
 
 The `id` represents the node id corresponding to the object, and it is possible to use this id directly in the [`wb_supervisor_node_get_from_id`](supervisor.md#wb_supervisor_node_get_from_def) supervisor function.
 The `position` and `orientation` are expressed relatively to the camera (the relative position is the one of the center of the object which can differ from its origin) and the units are meter and radian.
-The `size` represents the X and Y sizes in meters relatively to the camera (it is of course impossible to know the depth of the object).
+The `size` represents the Y and Z sizes in meters relatively to the camera (it is of course impossible to know the depth of the object along the [Camera](camera.md) X axis).
```
stefaniapedrazzi, line 1420: It seems that this was not updated when changing the camera orientation.

docs/reference/recognition.md
```diff
@@ -39,9 +40,14 @@ Objects farther than `maxRange` are not recognized.
 If more objects are visible to the camera, only the `maxObjects` biggest ones (considering pixel size) are recognized.
 
 - The `occlusion` field defines if occlusions between the camera and the object should be checked.
-Disabling the occlusion can be useful to allow the camera to see through thin or transparent objects that may hide the center of the object we are interested in, but it can lead to recognized objects that are not really visible to the camera.
+Disabling the occlusion can be useful to allow the camera to see through thin or transparent objects that may hide the object we are interested in, but it can lead to recognized objects that are not really visible to the camera.
 Additionally, it will slightly speed up the simulation.
 
+- The `occlusionAccuracy` fields defines the accuracy of the occlusion computation.
```
omichel, line 51:
```suggestion
- The `occlusionAccuracy` field defines the accuracy of the occlusion computation.
```

resources/nodes/Recognition.wrl
```diff
@@ -2,10 +2,11 @@
...
-  field SFFloat    maxRange       100     # maximum distance at which objects are detected (m)
-  field SFInt32    maxObjects     -1      # maximum number of objects detected simultaneously
-  field SFBool     occlusion      TRUE    # should occlusions between the radar and the target be checked
-  field SFColor    frameColor     1 0 0   # color of the objects frame in the overlay
-  field SFInt32    frameThickness 1       # thickness of the objects frame in the overlay
-  field SFBool     segmentation   FALSE   # generate segmented ground truth camera image
+  field SFFloat    maxRange          100     # maximum distance at which objects are detected (m)
+  field SFInt32    maxObjects        -1      # maximum number of objects detected simultaneously
+  field SFBool     occlusion         TRUE    # should occlusions between the radar and the target be checked
+  field SFInt32    occlusionAccuracy 1       # should occlusions between the radar and the target be checked
```
omichel, line 8: The comment looks wrong: copy paste of the previous one?

docs/reference/recognition.md
```diff
@@ -2,12 +2,13 @@
...
-  SFFloat  maxRange       100     # [0, inf)
-  SFInt32  maxObjects     -1      # {-1, [0, inf)}
-  SFBool   occlusion      TRUE    # {TRUE, FALSE}
-  SFColor  frameColor     1 0 0   # any color
-  SFInt32  frameThickness 1       # [0, inf)
-  SFBool   segmentation   FALSE   # {TRUE, FALSE}
+  SFFloat  maxRange          100     # [0, inf)
+  SFInt32  maxObjects        -1      # {-1, [0, inf)}
+  SFBool   occlusion         TRUE    # {TRUE, FALSE}
+  SFInt32  occlusionAccuracy 1       # {1, 2}
```
omichel, line 8: I am wondering if we shouldn't also allow a value of 0 for occlusionAccuracy which would allow us to deprecate the occlusion field.

src/webots/nodes/utils/WbObjectDetection.hpp
```diff
@@ -23,25 +23,31 @@
...
 class WbBaseNode;
 class WbSolid;
+class WbOdeGeomData;
 
 class WbObjectDetection {
 public:
   enum FrustumPlane { LEFT = 0, BOTTOM, RIGHT, TOP, PARALLEL, PLANE_NUMBER };
+  // Occlusion accurary:
+  // - ONE_RAY =  only one ray pointing at the center
+  // - MULTIPLE_RAYS =  multiple rays pointing at the BBOX or BS corners (created once object size is determined)
```
omichel, line 33:
```suggestion
  // - MULTIPLE_RAYS =  multiple rays pointing at the bounding box or bounding sphere corners (created once object size is determined)
```

omichel (changes requested):

docs/reference/recognition.md
stefaniapedrazzi: Yes, probably it will be simpler to have a single filed.
Ideally, then it would be better to call is simply `occlusion` but it might cause issues with backwards compatibility.

omichel: I agree it would be better. But since this aims at develop, we can afford to break backwards compatibility. I believe very few Recognition nodes had a non-default value for the occlusion field. So, it shouldn't break many simulations.

stefaniapedrazzi: Ok, so I will apply these changes.

omichel:
Is this PR ready for review again?

stefaniapedrazzi:
This PR is now ready for review.

docs/reference/recognition.md
```diff
@@ -38,9 +43,12 @@ Objects farther than `maxRange` are not recognized.
...
-Disabling the occlusion can be useful to allow the camera to see through thin or transparent objects that may hide the center of the object we are interested in, but it can lead to recognized objects that are not really visible to the camera.
-Additionally, it will slightly speed up the simulation.
+- The `occlusion` field defines if occlusions between the camera and the object should be checked and the accuracy that will be used.
+If the `occlusion` field is set to `0`, then the occlusion computation will be disabled.
+Disabling the occlusion can be useful to allow the camera to see through thin or transparent objects that may hide the object we are interested in, but it can lead to recognized objects that are not really visible to the camera.
+If the `occlusion` field is set to `1`, only the center of the object is taken into account to compute if the object is visible or not.
+Otherwise, if the accuracy is set to `2`, the outbound of the object is used to compute if the object is visible.
+Note the lower the `occlusion` field value is, the higher speed up the simulation achieves.
```
omichel, line 51: This sentence sounds odd to me, we could simplify it:
```suggestion
Note that increasing the `occlusion` field value decreases the simulation speed.
```

omichel (changes requested):
It looks good to me.
Just a little suggestion.

omichel (approved):
Thank you.

PR #6071, opened 2023-04-14
Fixed load progress dialog parenting

omichel:
The load progress dialog was not a child window of the main window. As a consequence, it may appear on a different screen as the main window. This PR fixes this small problem.


BenjaminDeleze (approved):
Thank you

PR #6081, opened 2023-04-17
Fix contact points warnings

### LINKED ISSUE #5819: Unwanted warnings or behavior in distributed simulations
Following the recent changes to the contact points detection these simulations are printing many warnings that should be solved (tested on develop branch):
* `animated_skin.wbt`: many warnings are printed
*  `sojourner.wbt`: cannot climb the rocks
* _continuing_

omichel:
Fix #5819 and display more useful information in the contact point warnings allowing the user to understand which contacts properties should be updated to handle more contact points.


stefaniapedrazzi (approved):
Thank you!

PR #6088, opened 2023-04-18
Fix crash when mTriangleMesh is NULL

stefaniapedrazzi:
Webots crashes when executing the "Convert Root to base node" on PROTO (created with urdf2webots script).

Although adding this additional check is correct and should be commited.
It would also be good to further investigate why this function is called when `mTriangleMesh` is NULL.


stefaniapedrazzi:
The problem occurs during a dictionary update and it is caused by the negative scale `-1 -1 -1`.
the DEF Mesh node is cloned and inserted into the MFNode field before the WbTrianfleMeshGeometry object is prefinalized and the `mTriangleMesh` value is set.
Given that the negative scaling is not valid, adjusting it triggers a scale update that checks if the value is a valid bounding object before the node is finalized.

```
0x0000555555f586cc in WbTriangleMesh::isValid (this=0x0) at nodes/utils/WbTriangleMesh.hpp:48
48	  bool isValid() const { return mValid; }
(gdb) bt
#0  0x0000555555f586cc in WbTriangleMesh::isValid() const (this=0x0) at nodes/utils/WbTriangleMesh.hpp:48
#1  0x0000555555f56ad1 in WbTriangleMeshGeometry::isAValidBoundingObject(bool, bool) const (this=0x55555befab30, checkOde=false, warning=false) at nodes/WbTriangleMeshGeometry.cpp:364
#2  0x0000555555f48995 in WbTransform::isAValidBoundingObject(bool, bool) const (this=0x7fffec008b90, checkOde=false, warning=false) at nodes/WbTransform.cpp:328
#3  0x0000555555f47d9b in WbTransform::applyToScale() (this=0x7fffec008b90) at nodes/WbTransform.cpp:141
#4  0x0000555555fc19d1 in WbAbstractTransform::updateScale(bool) (this=0x7fffec008d38, warning=true) at nodes/WbAbstractTransform.cpp:242
#5  0x0000555555f47dfc in WbTransform::updateScale(bool) (this=0x7fffec008b90, warning=true) at nodes/WbTransform.cpp:146
#6  0x00005555560e179d in WbTransform::qt_static_metacall(QObject*, QMetaObject::Call, int, void**) (_o=0x7fffec008b90, _c=QMetaObject::InvokeMetaMethod, _id=8, _a=0x7fffffffa6a0) at build/debug/WbTransform.moc.cpp:128
#7  0x00007ffff7b29f00 in  () at /home/stefi/develop/webots/lib/webots/libQt6Core.so.6
#8  0x00005555560e39a9 in WbValue::changed() (this=0x55555cfa05b0) at build/debug/WbValue.moc.cpp:169
#9  0x0000555555e6bec4 in WbSFVector3::setValue(WbVector3 const&) (this=0x55555cfa05b0, v=...) at vrml/WbSFVector3.cpp:38
#10 0x0000555555fc1794 in WbAbstractTransform::checkScale(int, bool) (this=0x7fffec008d38, constraintType=1, warning=true) at nodes/WbAbstractTransform.cpp:213
#11 0x0000555555f48651 in WbTransform::createOdeGeom(int) (this=0x7fffec008b90, index=0) at nodes/WbTransform.cpp:279
#12 0x0000555555f4ad0c in QtPrivate::FunctorCall<QtPrivate::IndexesList<0>, QtPrivate::List<int>, void, void (WbTransform::*)(int)>::call(void (WbTransform::*)(int), WbTransform*, void**)
     (f=(void (WbTransform::*)(WbTransform * const, int)) 0x555555f48554 <WbTransform::createOdeGeom(int)>, o=0x7fffec008b90, arg=0x7fffffffa9c0) at ../../include/qt/QtCore/QtCore/qobjectdefs_impl.h:152
#13 0x0000555555f4a8de in QtPrivate::FunctionPointer<void (WbTransform::*)(int)>::call<QtPrivate::List<int>, void>(void (WbTransform::*)(int), WbTransform*, void**)
     (f=(void (WbTransform::*)(WbTransform * const, int)) 0x555555f48554 <WbTransform::createOdeGeom(int)>, o=0x7fffec008b90, arg=0x7fffffffa9c0) at ../../include/qt/QtCore/QtCore/qobjectdefs_impl.h:185
#14 0x0000555555f4a4ed in QtPrivate::QSlotObject<void (WbTransform::*)(int), QtPrivate::List<int>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*) (which=1, this_=0x55555f7eece0, r=0x7fffec008b90, a=0x7fffffffa9c0, ret=0x0)
    at ../../include/qt/QtCore/QtCore/qobjectdefs_impl.h:395
#15 0x00007ffff7b29b8a in  () at /home/stefi/develop/webots/lib/webots/libQt6Core.so.6
#16 0x00005555560c146f in WbMultipleValue::itemInserted(int) (this=0x55555cf9ffe0, _t1=0) at build/debug/WbMultipleValue.moc.cpp:172
#17 0x0000555555d3c007 in WbMFNode::insertItem(int, WbNode*) (this=0x55555cf9ffe0, index=0, node=0x55555befab30) at vrml/WbMFNode.cpp:127
#18 0x0000555555fd931e in WbDictionary::updateDef(WbBaseNode*&, WbSFNode*, WbMFNode*, int, bool, bool&) (this=0x55555f64fee0, node=@0x7fffffffad80: 0x55555673b880, sfNode=0x0, mfNode=0x55555cf9ffe0, index=0, isTemplateRegenerator=false, regenerationRequired=@0x7fffffffb69f: false)
    at nodes/utils/WbDictionary.cpp:174
```

omichel (approved):
Do you know why this method is called when `mTriangleMesh` is `NULL`?

stefaniapedrazzi:
The `mTriangleMesh` value is firstly set during node finalization.
But the DEF Mesh node clone (the problem occurs with the USE instance in boundingObject field) and the insertion in parent node is performed just before node finalization.

omichel:
So would it be better to test the value of `mTriangleMesh` before calling `isAValidBoundingObject` and call it only if non-null?

stefaniapedrazzi:
There is no functional difference in checking the value of `mTriangleMesh` inside the `isAValidBoundingObject` or just before calling it. In both cases the ODE updates are skipped but I checked and this is not a problem because they will be performed during the finalization.

Instead of checking the `mTriangleMesh` value it would be more relevant to check if the node has already been finalized.
In any case in general I think that is bad to use pointers without first checking if they are valid, either with an assertion or with a check.

omichel:
Yes, for sure if we remove the test in the method, we should add an assert. But as you suggest it may be better to check if the node has already been finalized.

omichel (approved):
It looks better to me this way.
Thank you.

PR #6097, opened 2023-04-21
Restore missing const

ygoumaz:
As discussed in #6095, this PR restores the missing `const` keywords.


omichel (approved):

stefaniapedrazzi:
The changes in this PR are wrong.
There is no need to define double arguments as const and this is never done in Webots or in C/C++ because double arguments are const by default.

I will revert it.

ygoumaz:
I believe it is only useful for documentation and indicate that the value is never updated. It improves the global code readability. See the following thread https://stackoverflow.com/questions/117293/use-of-const-for-function-parameters for more details.

stefaniapedrazzi:
Ok.
But then the rest of Webots code never used this.
You can check any function in Webots that uses double arguments and none of them are using the const.

ygoumaz:
Yes, it's true. I agree it is not optimal for consistency right now, but it may be a good starting point to apply this rule everywhere. Of course it would be a huge work to update all functions at the same time but we should start to do it more often and update functions in future PRs.

PR #6100, opened 2023-04-25
Fix computation of Camera recognition object relative orientation

### LINKED ISSUE #6077: Recognition objects' orientation is not purely relative to the camera
## Summary

The [docs indicate](https://www.cyberbotics.com/doc/reference/camera#camera-recognition-object) that the orientation data returned from recognition objects should be purely relative to the camera. It therefore follows that the absolute orientation of the object and the camera should not impact the values returned.

What is instead observed is that the absolute orientation of a camera-recognition-object system in fact does impact the orientation data returned.

It looks to me like the value is at least influenced by the absolute orientation of the token or camera (or both, but in that case not equally).

Notably if I rotate the token in place or the camera in place then in either case the orientation values do also change, which suggests that something is trying to compute the relative orientations but the result is incomplete. 

## Reproduction

I've put a cut-down of the simulation I'm using (including a basic python controller) at https://gist.github.com/PeterJCLaw/58ce5bb8616d7f384fc0d6f18a60836e. It contains a minimal "robot" which is looking at a gold box on which is placed a recognition object which is the analogue of a fiducial marker.

1. launch the world
2. observe the printed orientation data
3. rotate the robot+box system through 90° or 180° (or whatever) on a vertical axis (i.e: change from e.g: looking east to looking west)
4. observe the printed orientation data

This cut-down examines the raw data, though my original controller translated the orientation data into other coordinate formats which had appeared to be correct aside from the issues observed here.

### Expected behaviour

Ideally the printed orientation data would be completely unchanged as the robot+box system rotates (since the relative orientation of the box from the camera is unchanged). I appreciate that there can often be multiple ways that a given orientation can be expressed in various coordinate/orientation systems, however that does not appear to be what is happening.

### Screenshots

Initial world:
image

Moving the token around to the side of the robot, then rotating them each through 90°, so that their relative orientations are the same (but both have moved relative to the world) we get:
image

I had expected that this would result in showing the same orientation data (within tolerances) for the first position as the second, yet they're different.

If I repeat this but put the token the other side of the robot (i.e: to the left of its original position, not right), then we get another different orientation. This time the new orientation is the same as to the right, but with one of the signs flipped:

image

### System
 - Operating System: originally observed on Ubuntu, reproduced on Windows
 - Versions: reproduced on R2022b and R2023a

### Additional context

Initially reported at https://discord.com/channels/565154702715518986/1071170269080719430.
Full simulation from which the cut-down was derived: https://github.com/srobo/competition-simulator/tree/sr2023.6

stefaniapedrazzi:
Fix #6077.

### DIFF
--- a/docs/reference/changelog-r2023.md
+++ b/docs/reference/changelog-r2023.md
@@ -60,6 +60,7 @@ Released on ??
     - Fixed the double downloading of meshes ([#6034](https://github.com/cyberbotics/webots/pull/6034)).
     - Fixed generation of moment of inertia matrix based on bounding box ([#6048](https://github.com/cyberbotics/webots/pull/6048)).
     - Fixed the loading of remote mesh bounding objects ([#6047](https://github.com/cyberbotics/webots/pull/6047)).
+    - Fixed object's relative orientation returned by the [Recognition](recognition.md) functionality ([#6100](https://github.com/cyberbotics/webots/pull/6100)).
 
 ## Webots R2023a
 Released on November 29th, 2022.
--- a/src/webots/nodes/WbCamera.cpp
+++ b/src/webots/nodes/WbCamera.cpp
@@ -822,7 +822,7 @@ bool WbCamera::computeObject(const WbVector3 &cameraPosition, const WbMatrix3 &c
 
   // compute object relative orientation
   WbRotation relativeRotation;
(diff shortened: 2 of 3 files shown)


omichel (approved):
Well spotted!

PR #6103, opened 2023-04-25
Fix tmp path to include username on Linux and macOS

omichel:
See discussion at https://github.com/cyberbotics/webots/pull/6067#issuecomment-1521257898

The problem affects only Linux and macOS where the Webots temporary files are stored in the global `/tmp` folder, which is necessary for shared memory segments and pipes.

On Windows the temporary directory used by Webots already contains the user name (as it is located in the local `AppData` folder).

If user A crashes Webots, it a `/tmp/webots-1234` folder will remain which user B cannot delete and hence they cannot launch a new instance of Webots on this port, although the port is available.

Changing the permissions of this folder and its contents to allow anyone to write/delete is not a good solution because it would allow any user to kill another one's simulation by deleting these files.

It seems that a cleaner solution would be to store the temporary files inside `/tmp/webots/USERNAME/ID/` instead of `/tmp/webots-ID/` and not change the permissions for this folder and its content to prevent others from deleting/modifying it.

This PR implements this cleaner solution.


ygoumaz (approved):
Thank you!

PR #6105, opened 2023-04-26
Update projects to Pose node

stefaniapedrazzi:
Check the distributed simulation in projects have been updated from Transform to Pose node:
* [x] vehicles
* [x] samples
* [x] robots
* [x] objects
* [x] PipeBoundingObject has been fixed so that is it now possible to add it directly in a boundingObject field 
* [x] devices and joints


stefaniapedrazzi:
This is what I found, but it is very likely that I still missed something else that we will find later.

omichel (approved):
Thank you.

PR #6111, opened 2023-04-28
Stop extern controller counter during Webots world loading

ygoumaz:
Connecting an extern controller to a Webots simulation can fail in various ways depending on the chosen protocol and provided information to target a specific Webots instance or robot. This PR addresses a common scenario where Webots is detected but the simulation is currently loading. In some cases, such as ROS 2 or webots.cloud, the controller is automatically started with Webots, but the simulation may take a while to load, causing a timeout in the connection process.

To prevent premature timeouts, this PR introduces a mechanism that stops the timeout counter when the simulation is still loading. This is straightforward for TCP connections, but IPC connections require creating a temporary file in the `tmp` directory of the Webots instance.

Moreover, this PR also improves error messages for other TCP and IPC scenarios, making it easier to identify and resolve connection issues.


src/controller/c/robot.c
```diff
@@ -1196,8 +1196,11 @@ static char *compute_socket_filename() {
     }
     closedir(dr);
     free(WEBOTS_TMP_DIR);
-    if (number == -1)
+    // No Webots instance has been started
+    if (number == -1) {
+      snprintf(error_buffer, ERROR_BUFFER_SIZE, "Cannot connect to Webots instance");
```
omichel, line 1201: We should make this message different from the next one:
```suggestion
      snprintf(error_buffer, ERROR_BUFFER_SIZE, "Cannot find any instance of Webots");
```

src/controller/c/robot.c
```diff
@@ -1221,36 +1224,69 @@ static char *compute_socket_filename() {
...
+    return NULL;
+  }
+  closedir(dr);
+  // check if Webots is currently loading
+  char *loading_file_path = malloc(strlen(webots_instance_folder) + 9);  // webots_instance_folder + '/loading'
+  snprintf(loading_file_path, strlen(webots_instance_folder) + 9, "%s/loading", webots_instance_folder);
+  FILE *loading_file = fopen(loading_file_path, "r");
+  if (loading_file) {
+    fclose(loading_file);
+    snprintf(error_buffer, ERROR_BUFFER_SIZE, "The Webots simulation world is not ready yet");
```
omichel, line 1259:
```suggestion
    snprintf(error_buffer, ERROR_BUFFER_SIZE, "The Webots simulation world is not yet ready");
```

src/controller/c/robot.c
```diff
@@ -1279,26 +1315,30 @@ static char *compute_socket_filename() {
...
         exit(EXIT_FAILURE);
       }
       filenames = r;
       filenames[count++] = strdup(de->d_name);
     }
     closedir(dr);
-    if (count == 0)
+    if (count == 0) {
+      snprintf(error_buffer, ERROR_BUFFER_SIZE,
+               "The Webots simulation has not yet been started or there is no robot set with an <extern> controller");
```
omichel, line 1327: It would be nice here to have a message slightly different from the one at line 1289 (easier to identify where does a problem come from).

omichel (changes requested):
That looks good to me.

omichel (approved):
Thank you.

PR #6112, opened 2023-05-01
Enhancement keyword proto

BenjaminDeleze:
**Description**

Add keywords to every protos. The keywords are used on webots.cloud to navigate quickly through the protos.


omichel (changes requested):
Instead of robot/walking I would prefer robot/legged as it is a more common denomination.

omichel (approved):
Thank you.

PR #6113, opened 2023-05-01
Fix tmp path for extern controllers

ygoumaz:
This update addresses a previously missed case in issue #6103. Specifically, when connecting an external controller and the `WEBOTS_CONTROLLER_URL` is either undefined or only the robot name is provided, the Webots instance information (in particular the port number) needs to be determined. To achieve this, we search for open instances of Webots and select the first available instance. This process relies on the tmp directory, which has undergone changes on Linux and macOS since the implementation of #6103.


src/controller/c/robot.c
```diff
@@ -1130,28 +1130,51 @@ static char *compute_socket_filename() {
...
     if (dr == NULL) {
-      fprintf(stderr, "Error: cannot open directory %s\n", TMP_DIR);
+      fprintf(stderr, "Error: cannot open directory %s\n", WEBOTS_TMP_DIR);
       exit(EXIT_FAILURE);
     }
     struct stat filestat;
     double timestamp = 0.0;
     int number = -1;
     struct dirent *de;
     while ((de = readdir(dr)) != NULL) {
```
omichel, line 1169: We should probably `free(WEBOTS_TMP_DIR);` after this loop.

omichel (changes requested):

src/controller/c/robot.c
ygoumaz: It's weird that cppcheck didn't detect the missing free()...

omichel (approved):
Thank you.

ygoumaz:
Test suite was broken since #6103 for two reasons:

- The required permissions in WbVideoRecorder to record movies were removed.
- The `extern` test was hanging indefinitely because it could not find the tmp directory.

omichel (approved):
Great, thank you.

PR #6115, opened 2023-05-03
Fix joints in pedal_racer.wbt

stefaniapedrazzi:
Fix joints axes in pedal_racer.wbt (probably broken after FLU conversion).


omichel (approved):
Thank you.

PR #6118, opened 2023-05-05
Fix ODE hinge joints limits

### LINKED ISSUE #3947: HingeJoint's minStop and maxStop attributes do not work correctly
Hello, 

I was trying to implement a simple mechanism that requires the use of a joint. The cards, shown on the GIFs, are supposed to be pushed by a robot, and then fall and stay vertically downwards due to gravity. The problem is, that I cannot limit the cards to stay vertically downwards. If I say that the joint's maxStop attribute has a value of π (or a bit less, e.g. 3),  the cards won't stop rotating, as shown on the first GIF. If the maxStop attribute has a value of ~2.6 (the second GIF), the cards rotate once or twice before they get limited by minStop and maxStop. If the maxStop attribute has a value of ~2 or less (the third GIF), I have not encountered an error. 

I would like to be able to set maxStop to be around π, so that the cards can stay vertically downwards. I could not find a reason why my PROTO file creates this kind of problem, so I believe it is a bug. Please take a look at the world on [this link](https://github.com/VladimirVincan/mep3/tree/excavation_squares/mep3_simulation/webots_data/worlds), it is named eurobot_2022.wbt. Also, please take a look at the PROTO file on[ this link](https://github.com/VladimirVincan/mep3/tree/excavation_squares/mep3_simulation/webots_data/protos), it is named ExcavationSquare.proto. Everything is done on the **excavation_squares** branch.

Thank you beforehand!

max_stop_3_3

max_stop_2_6

max_stop_2

ygoumaz:
Fixes #3947.

There was a singularity in ODE, when the min and max limits were set close to PI (or -PI). Depending on the basicTimeStep and on the torque applied, if the next computed angle was above PI or under -PI, it was directly clamped in range [-PI; +PI], switching the sign and messing with the limit computation. The fix uses the delta with the previous angle to correctly compute the limit to use for the LCP contraints algorithm.

### DIFF
--- a/docs/reference/changelog-r2023.md
+++ b/docs/reference/changelog-r2023.md
@@ -63,6 +63,7 @@ Released on ??
     - Fixed crash during conversion of PROTO to base nodes in case of Mesh nodes with invalid scale values ([#6088](https://github.com/cyberbotics/webots/pull/6088)).
     - Fixed object's relative orientation returned by the [Recognition](recognition.md) functionality ([#6100](https://github.com/cyberbotics/webots/pull/6100)).
     - Fixed conversion of joints' axis in `pedal_racer.wbt` to FLU/ENU coordinate system ([#6115](https://github.com/cyberbotics/webots/pull/6115)).
+    - Fixed a bug in ODE that caused minStop and maxStop limits to be violated in HingeJoints ([#6118](https://github.com/cyberbotics/webots/pull/6118)).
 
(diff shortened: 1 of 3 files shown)


omichel (changes requested):
Great. I would simply add a line in the change log.

omichel (approved):
Thank you.

PR #6119, opened 2023-05-05
Update crazyflie.c

knmcguire:
**Description**
The c controller on the crazyflie is broken because of the renaming of the inertial unit, as this was only updated for the python based controller.  Probably this would have been caught with testing later but better to fix it now.


omichel (approved):
Oops. Thank you.

PR #6127, opened 2023-05-09
Feature vacuum suction device

stefaniapedrazzi:
Add new device to simulate vacuum suction.

TODO:
* [x] add node
* [x] add new device API
* [x] update documentation
* [x] update WebotsJs
* [ ]  update webots.cloud
* [ ]  update ROS controller and ROS repo
* [ ]  update ROS2
* [ ] add new sample
* [ ] add tests


stefaniapedrazzi:
I started calling the new device for simulating vacuum grippers "VacuumCup" so that I could start with the development.
But any better suggestion is welcome because it is more the name of the object but doesn't represent the suction functionality.

Other options that I discarded:
`VacuumDevice` doesn't seem very good because it incluses the "device" in the name.
`VacuumGripper` is limited because the device could also be used for walking in robots legs and not only in end-effectors.

docs/reference/vacuumgripper.md
```diff
@@ -0,0 +1,254 @@
...
+  SFFloat  shearStrength     -1            # {-1, [0, inf)}
+  SFInt32  contactPoints     3             # [1, inf)
+}
+```
+
+### Description
+
+[VacuumGripper](#vacuum-gripper) node is used to simulate physical links created by vacuum suction.
+
+[VacuumGripper](#vacuum-gripper) node can only connect to dynamic [Solid](solid.md) nodes.
```
omichel, line 18: This sentence should be clarified. Does it mean it can connect only to Solid nodes with a Physics node? If so, that sounds wrong to me as it could be that a legged robot with vacuum grippers in its feet connect to the ground which has no physics (static environment).

lib/controller/python/controller/vacuum_gripper.py
```diff
@@ -0,0 +1,55 @@
...
+    def turnOff(self):
+        wb.wb_vacuum_gripper_turn_off(self._tag)
+
+    @property
+    def on(self) -> bool:
+        return wb.wb_vacuum_gripper_is_on(self._tag) != 0
+
+    @property
+    def presence(self) -> int:
+        return wb.wb_vacuum_gripper_get_presence(self._tag)
```
omichel, line 51:
```suggestion
    def presence(self) -> bool:
        return wb.wb_vacuum_gripper_get_presence(self._tag) != 0
```

resources/web/wwi/Parser.js
```diff
@@ -633,7 +635,9 @@ export default class Parser {
 
     const id = this.#parseId(node);
 
+    const type = getNodeAttribute(node, 'type', '').toLowerCase();
     const translation = convertStringToVec3(getNodeAttribute(node, 'translation', '0 0 0'));
+    const scale = convertStringToVec3(getNodeAttribute(node, 'scale', '1 1 1'));
```
omichel, line 640: That seems unrelated to this PR, is it fixing a bug?

omichel (changes requested):
Thank you.
It looks very good to me except that I believe a VacuumGripper should be able to connect to a Solid node without any physics (but with a bounding object).
I see no reason why introducing this constraint.
And this prevents use cases of legged robots walking on the ceiling or walls.

docs/reference/vacuumgripper.md
stefaniapedrazzi: Yes, currently it works only for Solid nodes with a Physics node (as it is also the case for the Connector node).
I will check to change it.

resources/web/wwi/Parser.js
stefaniapedrazzi: No, it is a merge issue.
This lines have been removed from the develop branch in the Pose PR.

projects/samples/devices/controllers/vacuum_gripper/vacuum_gripper.c
```diff
@@ -0,0 +1,110 @@
...
+  printf("Turn on vacuum gripper\n");
+  wb_vacuum_gripper_turn_on(vacuum_gripper);
+
+  wb_vacuum_gripper_enable_presence(vacuum_gripper, time_step);
+
+  bool connected = false;
+  for (int i = 0; i < NB_STEPS; i++) {
+    if (wb_vacuum_gripper_get_presence(vacuum_gripper) != connected) {
+      connected = wb_vacuum_gripper_get_presence(vacuum_gripper);
+      printf("Object has been %s", connected ? "picked" : "released");
```
omichel, line 84:
```suggestion
      printf("Object has been %s\n", connected ? "picked" : "released");
```

projects/samples/devices/controllers/vacuum_gripper/vacuum_gripper.c
omichel, line 1: This demo is not very nice.
I tested it on Windows and:
1. On the first run, it badly pick the object on the side in a non-realistic way.
2. It takes very long before the arm move with the grabbed object.
3. On the subsequent runs, it doesn't pick any object at all (apparently due to randomness).

I believe we should not rely on randomness for such a demo as it gives poor results.
vaccum_gripper

omichel (changes requested):

stefaniapedrazzi:
I will improve the demo.
However, there is no randomness involved in the object picking and random positions are only used after the object is picked (can easily be replaced by hard-coded positions but this won't solve the issue on Windows).
On Ubuntu, it works nicely and the object is always picked correctly and there is not long waiting between the actions even after restart or reset.
So the problem seems to be that Webots doesn't behave the same on Windows and Ubuntu and from your description, this particular simulation is not reproducible on Windows.

vaccum_gripper

stefaniapedrazzi:
> That is strange. We should understand why it behaves so differently on Windows. https://github.com/cyberbotics/webots/assets/1264964/921c7615-f338-4482-9fc1-fff679d64118

Yes, I'm now checking on Windows so that I can better understand where is the difference.

omichel:
I confirm the demo now looks way better on Windows.

src/webots/nodes/utils/WbNodeUtilities.hpp
```diff
@@ -21,8 +21,10 @@
 //              For generic node functions please refer to WbVrmlNodeUtilities namespace.
 //
 
-#include <QtCore/QList>
 #include "WbNode.hpp"
+#include "WbOdeTypes.hpp"
+
+#include <QtCore/QList>
```
omichel, line 27: This doesn't seem to be useful.
```suggestion
```

projects/samples/devices/worlds/vacuum_gripper.wbt
```diff
@@ -0,0 +1,285 @@
...
+                                  children [
+                                    Shape {
+                                      appearance DEF PLASTIC_APPEARANCE PBRAppearance {
+                                        baseColor 1 0.666667 0
+                                        roughness 1
+                                        metalness 0
+                                      }
+                                      geometry Cylinder {
+                                        height 0.02
+                                        radius 0.04
```
omichel, line 132:
```suggestion
                                        radius 0.042
```
Just to avoid this not very nice thing:
vacuum_gripper

omichel (changes requested):
Just a minor suggestion. Otherwise, it looks good to me.

omichel (approved):
Thank you.

PR #6134, opened 2023-05-11
Fix ubuntu classic theme

BenjaminDeleze:
The upgrade to Qt 6.4.3 broke the classic theme on ubuntu and macOS.

It copied the stylesheet from the dark theme to standardise the classic theme. However, I had to add a lot of other rules in it.

The result is a mix between the old classic theme and a white version of the dark theme.

- [x] Works on ubuntu
- [x] Does not break windows
- [x] Works on mac


ygoumaz (approved):
It looks nice on all 3 platforms now, thank you! Although we now have more entries in the classic qss than before, we may add the additional ones to Dusk and Dark themes qss files in another PR. If they all completely match (even if some entries are redundant with the default value), it would be easier to make a modification on the layout and back-port it to all themes.

PR #6136, opened 2023-05-12
Enhancement backward compatibility pose

BenjaminDeleze:
The introduction of the `Pose` node breaks the backward compatibility.

This PR adds:

- [x] A mechanism in Webots that convert `Transform` without scale (or with the default one) into `Pose` when parsing a world. It works both for worlds and protos. If the user saves the world, the conversions will be saved for the world, but not for the protos.
- [x] A script to convert the `Transform` withtout scale from proto/worlds to `Pose`.
  - The script has several known limitations: 
    - it does not support protos with templating
    - it requires the protos/worlds to be well formated. 
    - it does not convert the parameters of the protos

I could not see a performance impact.

### DIFF
--- a/scripts/converter/convert_transform_to_pose.py
+++ b/scripts/converter/convert_transform_to_pose.py
@@ -16,7 +16,7 @@
 
 """Convert Transforms with no scale or the default one into Poses"""
 
-# usage: put the script in the root folder of the project you want to convert and run it.
+# usage: copy this script and `webots_parser.py` in the root folder of the project you want to convert and run it.
 from webots_parser import WebotsParser
 from pathlib import Path
 
@@ -52,6 +52,8 @@ def convert_children(node):
             node['name'] = 'Pose'
 
     for field in node['fields']:
+        if field['type'] == 'IS':
+            continue
         if field['name'] in 'children':
             for child in field['value']:
                 convert_children(child)
--- a/scripts/converter/convert_transform_to_pose.py
+++ b/scripts/converter/convert_transform_to_pose.py
@@ -20,6 +20,7 @@
 from webots_parser import WebotsParser
 from pathlib import Path
 
+
 def convert_to_enu(filename):
     world = WebotsParser()
     world.load(filename)
@@ -33,6 +34,7 @@ def convert_to_enu(filename):
 
     world.save(filename)
     print('Conversion done')
+
 
 def convert_children(node):
     if 'USE' in node:
@@ -61,6 +63,7 @@ def convert_children(node):
             convert_children(field['value'])
         elif field['name'] in 'boundingObject':
             convert_children(field['value'])
+
 
 if __name__ == "__main__":
     files = []
--- a/scripts/converter/convert_transform_to_pose.py
+++ b/scripts/converter/convert_transform_to_pose.py
@@ -22,6 +22,7 @@ from pathlib import Path
 
 import sys
 
+
 def convert_to_enu(filename):
     file = WebotsParser()
     try:
--- a/scripts/converter/convert_transform_to_pose.py
+++ b/scripts/converter/convert_transform_to_pose.py
@@ -23,7 +23,7 @@ from pathlib import Path
 import sys
 
 
-def convert_to_enu(filename):
+def convert_to_pose(filename):
     file = WebotsParser()
     try:
         file.load(filename)
--- a/scripts/converter/convert_transform_to_pose.py
+++ b/scripts/converter/convert_transform_to_pose.py
@@ -82,6 +82,6 @@ if __name__ == "__main__":
 
         for file, path in local_files.items():
             print('Loading ' + str(path))
-            convert_to_enu(str(path))
+            convert_to_pose(str(path))
     except IndexError:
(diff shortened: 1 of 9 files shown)


stefaniapedrazzi:
I found that the `scale` field has not been removed from many nodes (mainly the devices) that now inherit from Pose instead of Transform (Accelerometer, Speaker, TouchSensor, etc.) `.wrl`.
I can fix it in a separate PR, but just wanted to mention it because it is somehow related to backward compatibility that will be broken.

scripts/converter/convert_transform_to_pose.py
```diff
@@ -0,0 +1,87 @@
...
+"""Convert Transforms with no scale or the default one into Poses"""
+
+# usage: python convert_transform_to_pose.py /path/to/your/webots/project
+from webots_parser import WebotsParser
+from pathlib import Path
+
+import sys
+
+
+def convert_to_enu(filename):
```
omichel, line 26: This function name is a bit misleading, isn't it?
```suggestion
def convert_to_pose(filename):
```

src/webots/vrml/WbVrmlNodeUtilities.cpp
```diff
@@ -481,3 +482,40 @@ QString WbVrmlNodeUtilities::exportNodeToString(WbNode *node) {
   node->write(writer);
   return nodeString;
 }
+
+// Return true if we can convert the Transform to a Pose.
+bool WbVrmlNodeUtilities::transformBackwardCompatibility(WbTokenizer *tokenizer) {
+  if (tokenizer) {
```
omichel, line 488:
```suggestion
  if (!tokenizer)
    return true;
```

src/webots/vrml/WbVrmlNodeUtilities.cpp
```diff
@@ -481,3 +482,40 @@ QString WbVrmlNodeUtilities::exportNodeToString(WbNode *node) {
...
+        else if (token == "]") {
+          bracketCount--;
+          if (bracketCount == 0)
+            inChildren = false;
+        }
+      } else if (token == "children") {
+        inChildren = true;
+      } else if (token == "scale") {
+        for (int i = 0; i < 3; i++) {
+          if (tokenizer->nextWord() != '1') {
```
omichel, line 506: 1 may also be written "1.0". Shall we transform the token into a float and compare it to 1?

omichel (changes requested):

scripts/converter/convert_transform_to_pose.py
BenjaminDeleze: Ooops a bad copy/paste

src/webots/vrml/WbVrmlNodeUtilities.cpp
```diff
@@ -481,3 +482,41 @@ QString WbVrmlNodeUtilities::exportNodeToString(WbNode *node) {
...
+      else if (token == "]") {
+        bracketCount--;
+        if (bracketCount == 0)
+          inChildren = false;
+      }
+    } else if (token == "children") {
+      inChildren = true;
+    } else if (token == "scale") {
+      for (int i = 0; i < 3; i++) {
+        if (tokenizer->nextWord().toFloat() != '1') {
```
omichel, line 508:
```suggestion
        if (tokenizer->nextWord().toFloat() != 1) {
```

BenjaminDeleze:
The crash I was investigating was finally not due to this PR

src/webots/vrml/WbVrmlNodeUtilities.cpp
```diff
@@ -481,3 +482,41 @@ QString WbVrmlNodeUtilities::exportNodeToString(WbNode *node) {
...
+      else if (token == "]") {
+        bracketCount--;
+        if (bracketCount == 0)
+          inChildren = false;
+      }
+    } else if (token == "children") {
+      inChildren = true;
+    } else if (token == "scale") {
+      for (int i = 0; i < 3; i++) {
+        if (tokenizer->nextWord().toFloat() != 1) {
```
omichel, line 508:
```suggestion
        if (tokenizer->nextWord().toFloat() != 1.0f) {
```

src/webots/vrml/WbNodeReader.cpp
```diff
@@ -49,10 +50,14 @@ WbNode *WbNodeReader::createNode(const QString &modelName, WbTokenizer *tokenize
...
-  if (model)
-    return new WbNode(modelName, worldPath, tokenizer);
-
+  if (modelName == "Transform")
+    return WbVrmlNodeUtilities::transformBackwardCompatibility(tokenizer) ? new WbNode("Pose", worldPath, tokenizer) :
+                                                                            new WbNode("Transform", worldPath, tokenizer);
+  else {
+    WbNodeModel *const model = WbNodeModel::findModel(modelName);
+    if (model)
+      return new WbNode(modelName, worldPath, tokenizer);
```
omichel, line 59:
```suggestion
    if (WbNodeModel::findModel(modelName))
      return new WbNode(modelName, worldPath, tokenizer);
```

PR #6138, opened 2023-05-15
Remove scale field from devices WRL and improve JS basenode encoder

stefaniapedrazzi:
* Remove scale field from WRL files for nodes inheriting from Solid
* Sort nodes in alphabetical order when generating `FieldModel.js` in order to produce the same list independently on the system
* Regenerate `FieldModel.js`

### DIFF
--- a/resources/nodes/Accelerometer.wrl
+++ b/resources/nodes/Accelerometer.wrl
@@ -7,7 +7,6 @@ Accelerometer {
   #fields that inherit from the Solid node:
   vrmlField SFVec3f    translation         0 0 0
   vrmlField SFRotation rotation            0 0 1 0
-  vrmlField SFVec3f    scale               1 1 1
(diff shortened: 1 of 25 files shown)


stefaniapedrazzi:
It seems that the CI tests are currently broken due to failures with the dependencies installation.
I ran the test_suite locally and everything is working correctly.


```
../../include/qt/QtGui/QtGui/qopengl.h:105:13: fatal error: GL/gl.h: No such file or directory
  105 | #   include <GL/gl.h>
```

omichel:
Yes, this happened to me last week and restarting the test resolved the problem. This is likely a GiuHub issue.

stefaniapedrazzi:
After the test was restarted it passed on Ubuntu 20.04 (Ubuntu 22.04 test was probably not restarted).
However, after synchronization now it seems that there are issues with the ROS installation.

PR #6143, opened 2023-05-16
Better fix to comparison issue

BenjaminDeleze:
Better patch than #6141.
The previous fix did not work correctly when the two protos had the same fields, but not the same bodies.


omichel (approved):
Looks good to me.
Thank you.

PR #6150, opened 2023-05-22
Cleanup unused files and fields

BenjaminDeleze:
The `solid` field of `IndexedFaceSet` is a VRML field but, in practice, we do not implement it and just ignore it.
This field was present in `Extrusion.proto` but it was confusing to have it and that nothing was changing when modifying it.

So we decided to remove it.

Also the `WbImportWizard` is not used anywhere and was certainly forgotten when we removed the import feature.


omichel (approved):
Thank you.

PR #6156, opened 2023-05-25
Update using-ros.md

LucasWaelti:
**Description**
Added instructions to implement a ROS node to serve as an extern Webots controller in C++. Only Python has been mentioned so far. 
This only requires some minimal setup in the ROS package's `CMakeLists.txt` and is therefore quite straightforward. 

**Documentation**
This pull-request changes the following file:
https://cyberbotics.com/doc/guide/using-ros?version=released


ygoumaz (changes requested):
Thank you for the enhancement! 

I am wondering if we should not directly provide a sample C++ controller corresponding to `ros_python.py` and call it `ros_cpp.cpp`. The package could also contain the `CMakeLists.txt` file. The controller could be compiled from the `webots_ros` package and a dedicated launch file (`webots_ros_cpp.launch`) would allow to start the same world as the Python example with the compiled C++ controller.

The content of the CMakeLists.txt file could be removed from the .md documentation and we could provide instructions to run the launch file equivalent to the Python one.

LucasWaelti:
> I am wondering if we should not directly provide a sample C++ controller corresponding to `ros_python.py` and call it `ros_cpp.cpp`. 

For me, it makes sense to keep the cmake instructions in the doc because they are generally applicable and precisely show what has to be added. Might be more straightforward than having to look up an example...

ygoumaz:
I agree with you. While it is important to keep the content in the document for reference and quick lookup, I would still include the example in the package. Simply relying on the CMakeLists file may not be straightforward for everyone to implement. Adding the example in the project, along with the detailed file, would help users to understand the differences with the Python implementation (including the controller file), while also being able to run it.

antonkesy (approved):

ygoumaz (changes requested):
This PR is still waiting for the required changes.

LucasWaelti:
Just to clarify, I do not have time to implement myself an example for this (if this is what was understood), sorry!

LucasWaelti:
As you wish, if you prefer to provide no help at all instead of a useful hint, which only requires some basic knowledge about how to build a ROS node, fine by me... 
If I find the time, I'll try and create an example though as I think this really should be part of the documentation. This has definitely been lacking for me at least.

PR #6158, opened 2023-05-26
Fix compilation warnings on macOS

ygoumaz:
`sprintf` is deprecated and `snprintf` should be used instead. This is triggering warnings on macOS preventing the manual creation of packages.


omichel (dismissed):
Thank you.

omichel (approved):
Better this way indeed.

PR #6167, opened 2023-06-05
Add proto formatter script

ygoumaz:
This new PROTO formatter script allows to automatically format the given PROTO file to get the correct block indentations.


BenjaminDeleze (approved):
Thank you

PR #6171, opened 2023-06-06
Harmonize qss instruction for dark and dusk

### LINKED ISSUE #6179: Issues with classic theme
The arrows for increasing/decreasing the double value are no longer visible (same reproducible in both Ubuntu 22.04 and Windows).

Here is a screenshot of the Field Editor:
Screenshot from 2023-06-07 11-13-27

Then, the field editor has no border.
Personally I think that it was good to use the border to separate it from the console widget.
It doesn't change anything when a node is selected because the node editor has his own border, but this is not the case when selecting for example a double value.

For comparison here is the previous appearance:
Screenshot from 2023-06-07 11-30-05

BenjaminDeleze:
Now all three themes have the exact same set of qss instructions.
The only difference is the color.

This PR also fix #6179 

What remains to be done is to find a way to have only one qt file with variables and to change them on loading


omichel (approved):
Thank you.

PR #6185, opened 2023-06-09
Fixed test suite on Windows

omichel:
The output of the test suite on Windows was unreadable because it was mixing the stdout and stderr of Webots to the test results in the console output. This PR fixes the problem so that the output is the same on Windows and Linux.


BenjaminDeleze (approved):
Thank you

PR #6190, opened 2023-06-12
Fix Connector ODE body detection

stefaniapedrazzi:
Use ODE body from `Connector` node if any or search in parent nodes (but stop at joints).
Currently the code only searches for the connector's upper Solid, this means that
* the `Connector` node itself cannot be used in joint's `endPoint` fields otherwise the connected peer won't move with the joint
* if the `Connector` node has a parent `Solid` node without physics, the peer will be connected to the static environment instead of searching for ancestor `Solid` nodes with physics.


src/webots/nodes/WbConnector.cpp
```diff
@@ -407,18 +403,16 @@ void WbConnector::snapNow(WbConnector *other) {
...
 }
 
 // this function must be called once the connectors are aligned
-void WbConnector::createFixedJoint(WbConnector *other) {
-  dBodyID b1 = upperSolid()->bodyMerger();
-  dBodyID b2 = other->upperSolid()->bodyMerger();
+void WbConnector::createFixedJoint(WbConnector *other, const dBodyID b1, const dBodyID b2) {
   if (!b1 && !b2) {
-    warn(tr("Connectors could not be attached because none of their parent nodes have Physics nodes."));
+    warn(tr("Connectors could not be attached because neither of them (nor their parent nodes) has Physics nodes."));
```
omichel, line 415:
```suggestion
    warn(tr("Connectors could not be attached because neither of them (nor their parent nodes) has a Physics node."));
```

src/webots/nodes/WbConnector.cpp
```diff
@@ -456,10 +450,10 @@ void WbConnector::attachTo(WbConnector *other) {
   if (!(mUnilateralLock->isTrue() || other->mIsLocked->isTrue()))
     return;
 
-  dBodyID b1 = upperSolid()->bodyMerger();
-  dBodyID b2 = other->upperSolid()->bodyMerger();
+  const dBodyID b1 = WbNodeUtilities::findBodyMerger(this);
+  const dBodyID b2 = WbNodeUtilities::findBodyMerger(other);
   if (!b1 && !b2) {
-    warn(tr("Connectors could not be attached because none of their parent nodes have Physics nodes."));
+    warn(tr("Connectors could not be attached because neither of them (nor their parent nodes) has Physics nodes."));
```
omichel, line 456:
```suggestion
    warn(tr("Connectors could not be attached because neither of them (nor their parent nodes) has a Physics node."));
```

omichel (dismissed):
It looks good to me.
I just have a minor suggestion for clarity.

omichel (approved):

PR #6192, opened 2023-06-12
Fix classic theme

### LINKED ISSUE #6188: Background of editable fields is gray instead of white
In the classic theme (on Ubuntu 22.04), the background of the editable fields is now grey when the widget doesn't have focus.
When selecting it it correctly turns white.
However, usually the gray background is used for non-editable widget so it might seem that these field cannot be modified so it would be better to restore the white background.

Here are some screenshots:
Screenshot from 2023-06-12 10-18-18

Screenshot from 2023-06-12 10-18-31

Screenshot from 2023-06-12 10-20-41

And here is the comparison with R2022a:
Screenshot from 2023-06-12 10-24-24

Screenshot from 2023-06-12 10-24-04

Screenshot from 2023-06-12 10-23-48

BenjaminDeleze:
Fix #6188

### DIFF
--- a/resources/webots_classic.qss
+++ b/resources/webots_classic.qss
@@ -447,7 +447,7 @@ QScrollArea#editorPane {
 
 #NodePane QLineEdit,
 #LoginTabs QLineEdit {
-  background-color: #e6e6e6;
+  background-color: white;
   color: black;
   border: 1px solid #ababab;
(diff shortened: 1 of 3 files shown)


BenjaminDeleze:
I will merge it now such that it is included in the nightly build.

PR #6198, opened 2023-06-13
Fix background when choosing controller

### LINKED ISSUE #6196: Borders missing in controllers window
Same for robot windows choice, solid name (for SolidReference), etc.
Without borders it is not clear where is the label and which are the items that can be selected, or where is scroll bar.

image

BenjaminDeleze:
Fix #6196


stefaniapedrazzi (approved):
Thank you

PR #6199, opened 2023-06-13
Fix issues with manipulators of scaled Transform nodes

stefaniapedrazzi:
Translation/rotation manipulator of scaled Transform nodes doesn't have the correct size.
* [x] fix manipulator scale for scaled Transform node
* [x] fix manipulator scale for children of scaled Transform node
* [x] fix behavior when translating a node with scaled parent with the handles
* [x] fix handles size not correctly set when selecting node from the scene tree


BenjaminDeleze:
I have rebased your branch from develop to master as we merged develop in master.

BenjaminDeleze (approved):
Thank you!

BenjaminDeleze:
I will merge the PR to recreate the packages

PR #6201, opened 2023-06-13
Disable shadow reception computation on mesh when shadows are globally disabled

omichel:
Follow-up on #3917.

When the shadows are globally disabled, we should skip the shadow computations for meshes.


BenjaminDeleze (commented):
It works really well, but shouldn't we do the same also for dynamic meshes?

BenjaminDeleze (approved):
My bad it was already done

PR #6203, opened 2023-06-14
Fix checkbox check mark first launch

BenjaminDeleze:
**Description**
The checkbox check mark was not displayed during the first launch of Webots: neither in the new version window, neither in the preferences and so on.

It was working in qt 6.2 because we were using the default check mark of qt (not working anymore since 6.4)

The reason of the issue is:

1. We define the path to the check mark icon with a q-property that is in the qss (because the path depends on the theme.
2. We add the path to qt when creating the main window, however, the new version window is created *before* the main window.
3. It seems that once qt tried to access a wrong image (the check mark in the version window) it will still display it wrongly even if, in the meantime, the path became correct: which explain why the check mark does not appear in preferences  for the first launch. (Note that it does not throw any error)
 
The solution to fix this is a bit convoluted:
Attach a name to the checkbox that is in the new version window and build a specific path property for this checkbox while excluding the default one for being applied to it.

Why the solution is not simpler:
a. Put the q-property in `GuiApplication` to initialize it only once for both the new version and the main window: does not work as `GuiApplication` is not a descendant of QWidget.
b. Initialize all q-property in the new-version window: works only if the new version window is opened which happen only during the first launch.
c. Just copy paste the q-property in the new version window and initialize it twice: does not work because, the new version window will opened with the night theme (on ubuntu), but if we select another theme, it will not update the path as it is already set.
d. Why add `WbMainWindow` before the general `QCheckBox` property: even if the image applied in the new version window is the one defined in QCheckBox#telemetryCheckBox:indicator:checked, the qss will go through the generic `QCheckBox:indicator:checked` (if we do not explicitly prevent it with the `WbMainWindow`) and thus will resolve the not yet defined image url and we go back to point 3.

### DIFF
--- a/resources/webots_classic.qss
+++ b/resources/webots_classic.qss
@@ -25,8 +25,6 @@ WbSplashScreen {
 WbNewVersionDialog {
   qproperty-newVersionIconPath: "icons/dark";
 }
-
-
 QCheckBox#telemetryCheckBox:indicator:checked {
   border: 1px solid black;
   image: url(newVersionIconPath:checkboxCheckmark.png);
--- a/resources/webots_classic.qss
+++ b/resources/webots_classic.qss
@@ -28,7 +28,7 @@ WbNewVersionDialog {
 
 
 QCheckBox#telemetryCheckBox:indicator:checked {
-  border: 1px solid dimgrey;
+  border: 1px solid black;
   image: url(newVersionIconPath:checkboxCheckmark.png);
 }
 
--- a/resources/webots_classic.qss
+++ b/resources/webots_classic.qss
@@ -22,6 +22,10 @@ WbSplashScreen {
   qproperty-loadingColor: #666;
 }
 
+WbNewVersionDialog {
+  qproperty-enabledIconPath: "icons/dark";
+}
+
 WbMainWindow {
   qproperty-enabledIconPath: "icons/dark";
   qproperty-disabledIconPath: "icons/light";
--- a/resources/webots_classic.qss
+++ b/resources/webots_classic.qss
@@ -25,6 +25,7 @@ WbSplashScreen {
 WbNewVersionDialog {
   qproperty-newVersionIconPath: "icons/dark";
 }
+
 QCheckBox#telemetryCheckBox:indicator:checked {
   border: 1px solid black;
   image: url(newVersionIconPath:checkboxCheckmark.png);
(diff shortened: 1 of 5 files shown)


resources/webots_classic.qss
```diff
@@ -22,6 +22,16 @@ WbSplashScreen {
   qproperty-loadingColor: #666;
 }
 
+WbNewVersionDialog {
+  qproperty-newVersionIconPath: "icons/dark";
+}
+
+
```
stefaniapedrazzi, line 29:
```suggestion

```

resources/webots_night.qss
```diff
@@ -22,6 +22,16 @@ WbSplashScreen {
   qproperty-loadingColor: #b4b4b4;
 }
 
+WbNewVersionDialog {
+  qproperty-newVersionIconPath: "icons/light";
+}
+
+
```
stefaniapedrazzi, line 29:
```suggestion

```

stefaniapedrazzi (approved):
Thank you!

I noticed that this issue occurs only if the Night theme is chosen.
Otherwise when chosing the classic or dusk theme, the path is probably valid when loading the new theme.
But it is good to have the same for all the themes.

BenjaminDeleze:
> Thank you!
> 
> I noticed that this issue occurs only if the Night theme is chosen. Otherwise when chosing the classic or dusk theme, the path is probably valid when loading the new theme. But it is good to have the same for all the themes.

Yes, because it happened only when the default theme is selected, on Ubuntu the default theme is Night, but on macOS and Windows it is Classic and so the issue was happening also on Classic.
And I implemented it in Dusk such that all 3 themes are coherent.

resources/webots_classic.qss
```diff
@@ -22,6 +22,14 @@ WbSplashScreen {
   qproperty-loadingColor: #666;
 }
 
+WbNewVersionDialog {
+  qproperty-newVersionIconPath: "icons/dark";
+}
+QCheckBox#telemetryCheckBox:indicator:checked {
```
stefaniapedrazzi, line 28: Oops both new lines where removed:
```suggestion
}

QCheckBox#telemetryCheckBox:indicator:checked {
```

resources/webots_dusk.qss
```diff
@@ -22,6 +22,14 @@ WbSplashScreen {
   qproperty-loadingColor: #b4b4b4;
 }
 
+WbNewVersionDialog {
+  qproperty-newVersionIconPath: "icons/light";
+}
+QCheckBox#telemetryCheckBox:indicator:checked {
```
stefaniapedrazzi, line 28:
```suggestion
}

QCheckBox#telemetryCheckBox:indicator:checked {
```

PR #6220, opened 2023-06-15
Fix unneeded regeneration of scene tree items

stefaniapedrazzi:
Issue introduced in #5950.
It is not needed to re-create the scene tree items if the changed field is not a node.
This will also avoid the selection change that it is automatically triggered when a new item is created.


BenjaminDeleze (approved):
Seems good to me

PR #6225, opened 2023-06-16
Fix bounding object scale

### LINKED ISSUE #6221: Pose: wrong bounding object
After the pose node PR, some nodes, that were scaled at root before have a wrong bounding object.
toilet

BenjaminDeleze:
**Description**
I went through all protos (I may have missed some devices and slots) to fix #6221


omichel (approved):
Thank you.

PR #6226, opened 2023-06-16
Fix led crash

### LINKED ISSUE #6213: Webots crashes when deleting children of LED
**Steps to Reproduce**
1. Go to 'led.wbt'
2. Run simulation
3. Delete `Group` node in the children of LED "led1"

BenjaminDeleze:
**Description**
Fix #6213 

When deleting the group present in LED it was crashing because the appearances/lights lists of led was not reset.

This bug was also present in battery.


omichel (approved):
Thank you.

PR #6228, opened 2023-06-17
Fix khr3hv dimensions

### LINKED ISSUE #6207: Bug with size of khr-3hv
The position of the torso of khr-3hv.wbt on webots is set as follows. 

Screenshot from 2023-06-15 12-03-24

However, the size of khr-3hv on the real robot is 401.05mm high (=0.40105m),  so the robot on webots is about twice the size of the real robot. Is it possible to change it to the same scale?

Ref:
- khr-3hv real robot document : [https://kondo-robot.com/d/product/03110e](url)

naos080415:
Description
Fix #6207


omichel (approved):
Thank you.
I just added a line in the change log about it.
I also invited you to become a member of the Webots committers group to facilitate your next contributions if any. You should be able to make new contributions without having to fork our repository, you will be able to merge PR, etc.

Issue #6217, opened 2023-06-15
Missing dependencies for SUMO exporter


stefaniapedrazzi:
Following the instruction on the Q&A and documentation (https://www.cyberbotics.com/doc/automobile/sumo-exporter), the exporter fails on snap/Ubuntu 22.04 because of missing dependencies (like `libfox-1.6.0`, `libxerces-c3.2`).

All these dependencies can be installed with the `sumo` package.
However, this is not documented.

stefaniapedrazzi:
I started to check in details which libraries where missing and I had to install/copy 32 additional libraries (72.9 MB).

The missing libraries can also be installed with the following two packages that we could document:
* libgdal30
* libfox-1.6.0

However, the current SUMO version uses libgdal26 and not libgdal30 (not available by default in apt for Ubutnu 22.04) so it is misleading to install a different version just for getting all the dependencies.

PR #6232, opened 2023-06-19
Document missing SUMO libraries

stefaniapedrazzi:
Address #6217:
document missing system dependencies to run the SUMO exported on Ubuntu 22.04 (and Ubuntu 20.04).


stefaniapedrazzi:
No, I have the exact same issue with the tarball.

stefaniapedrazzi:
It works with the debian package because the package already depends `libgdal-dev` and `libfox-1.6-dev`

BenjaminDeleze:
Did you follow the installation instruction for the `.tar.gz`? 
https://cyberbotics.com/doc/guide/installation-procedure?version=master&tab-os=linux#installing-the-tarball-package
Because it is specified that you need some additional packages for SUMO.

stefaniapedrazzi:
No I didn't. But the same issue occurs with the snap package, and I had in any case to install these two additional libraries to make it work.
Moreover, I didn't need to install the dev packages. So probably also the installation documentation needs to be reviewed.

stefaniapedrazzi:
I will test it, but also for the debian package I don't think we need the to depend on the dev packages.

ygoumaz:
There is also a lot of libraries missing on macOS preventing the use of SUMO. After discussion with @BenjaminDeleze, I wonder here what is our best option with SUMO. By embedding it in the Webots distribution, the goal is to make it work out-of-the-box, so adding documentation (for additional library installation) to make it work doesn't seem to meet the original goal. So here I believe we have two main possibilities:

- We keep SUMO in the package, but this means that we need to provide all necessary libraries to execute it. When we create the SUMO tar.gz, we follow the following procedure: https://github.com/cyberbotics/webots/wiki/SUMO-compilation where we already add some libraries. We should enhance this procedure with currently missing libraries. The consequence would be that SUMO would work out-of-the-box, but the package might significantly grow in size.
- We remove completely SUMO of the package and document its installation (which is very fast using apt). SumoInterface and other components that we distribute should probably be adapted. The pros is that the package size will be significantly reduced.  The drawback is that the user needs to install SUMO by himself, but it would probably be equivalent to the instructions given in this PR.

scripts/packaging/linux_distro.py
```diff
@@ -193,7 +193,7 @@ def create_debian_bundle(self):
                 "libglib2.0-0 (>= 2.10.0), libegl1, libglu1-mesa | libglu1, libgtk-3-0, "
                 "libnss3, libstdc++6 (>= 4.0.2-4), libxaw7, libxrandr2, libxrender1, "
                 "libssh-dev, libzip-dev, xserver-xorg-core, libxslt1.1, "
-                "libxerces-c-dev, libfox-1.6-dev, libgdal-dev, libproj-dev, libgl2ps-dev, "  # SUMO dependencies
+                "libxerces-c3.2, libfox-1.6-0, libgdal-dev, libproj-dev, libgl2ps1.4, "  # SUMO dependencies
```
stefaniapedrazzi, line 196: I kept `libgdal-dev`and `libproj-dev` just because the version changed in Ubuntu 22.04 and this way the version is managed automatically by the packaging tool.

stefaniapedrazzi:
As already mentioned in the issue description (https://github.com/cyberbotics/webots/issues/6217) on Ubuntu 22.04 32 libraries are missing, that is 72.9 MB.

omichel:
I believe we should completely drop SUMO from the package and provide installation instructions instead.

stefaniapedrazzi:
I can update the instructions for Linux and remove SUMO from the packages.

PR #6233, opened 2023-06-19
Fix double Lidar head export

ygoumaz:
As reported in https://github.com/cyberbotics/webots_ros2/pull/770, the Lidar head is written two times when exported to URDF. This PR limits the exports to subnodes for URDFs.


omichel (approved):
Thank you.

PR #6235, opened 2023-06-20
Allow the compilation in non-ASCII paths

omichel:
gcc is now able to compile robot controllers in non-ASCII paths (I tested with Chinese characters).

### DIFF
--- a/src/webots/editor/WbBuildEditor.cpp
+++ b/src/webots/editor/WbBuildEditor.cpp
@@ -304,17 +304,6 @@ void WbBuildEditor::make(const QString &target) {
 
   // find out compilation directory
   QString compilePath = compileDir().absolutePath();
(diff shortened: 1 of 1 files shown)


BenjaminDeleze:
It doesn't seem to work for me:
```
make clean
Clean finished.
make -j 12
# updating 目前页面主页.d
# compiling 目前页面主页.c
Assembler messages:
Fatal error: can't create build/release/??????.o: Invalid argument
make: *** [webots/resources/Makefile.include:626: build/release/目前页面主页.o] Error 1
```

I update msys (`pacman -Syuu`) and qt (`qt_windows_installer.sh`) but it still does not work.

omichel:
I fixed the test and error message to check for non local 8-bit characters in the controller name only.

BenjaminDeleze (approved):
Thank you!

PR #6239, opened 2023-06-21
Doc fix ros tutorial instruction

BenjaminDeleze:
I had to make the change because the old way was not working anymore

Source: https://wiki.ros.org/noetic/Installation/Ubuntu

### DIFF
--- a/docs/guide/tutorial-9-using-ros.md
+++ b/docs/guide/tutorial-9-using-ros.md
@@ -24,7 +24,8 @@ To install the latest version of ROS on Ubuntu use the following commands:
 
 ```sh
 sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros-latest.list'
-sudo curl -sSL https://raw.githubusercontent.com/ros/rosdistro/master/ros.key -o /usr/share/keyrings/ros-archive-keyring.gpg
+sudo apt install curl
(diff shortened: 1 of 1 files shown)


docs/guide/tutorial-9-using-ros.md
```diff
@@ -24,7 +24,8 @@ To install the latest version of ROS on Ubuntu use the following commands:
 
 ```sh
 sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros-latest.list'
-sudo curl -sSL https://raw.githubusercontent.com/ros/rosdistro/master/ros.key -o /usr/share/keyrings/ros-archive-keyring.gpg
+sudo apt install curl
+curl -s https://raw.githubusercontent.com/ros/rosdistro/master/ros.asc | sudo apt-key add -
```
ygoumaz, line 28:
```suggestion
sudo curl -s https://raw.githubusercontent.com/ros/rosdistro/master/ros.asc | sudo apt-key add -
```

I would add sudo to be consistent with other commands.

ygoumaz (changes requested):

docs/guide/tutorial-9-using-ros.md
omichel: I am not sure as in this case the sudo is not necessary and I believe we should use sudo only when necessary.

ygoumaz (approved):

Issue #6063, opened 2023-04-11
Issues with scaled Transform nodes


stefaniapedrazzi:
While testing with scaled objects I found a couple of issues:
* [x] wrong translation and rotation displayed in Position tab in the Field Editor -> rotation should not be affected by ancestor scale
    I would expect to be able to insert the relative translation and rotation directly in a Scene Tree's node to achieve the same object position
* [x] handle not working correctly for Transform nodes with a scaled parent (#6199)
* [x] check if boundingSphere is computed correctly :  the visualized one doesn't seem to take the descendant Transform scale into account, i.e. when selecting the scaled (e.g. scale 0.1, 0.1, 0.1) geometry or Transform node the size of the sphere is correct, when selecting the parent Solid node the bounding sphere is much bigger. (#6243)

Regressions:
* [x] handles of scaled Transform nodes (scale 0.1 0.1 0.1) are very small and unusable (#6199).

stefaniapedrazzi:
After further investigations, I have the impression that the problem is not in the computations in the PositionViewer, but in the way WREN handles the scale.

First, this line seems wrong because the current node and the parent node scale values are multiplied without taking the rotation into consideration.

https://github.com/cyberbotics/webots/blob/f03f2e9745396ad4ba98d662eeca87d8f26053e2/src/wren/TransformNode.cpp#L64

I replaced the transform matrix computations in WREN with the ones used by three.js, and in the PositionViewer I can compute almost correct translation and rotation values with just some small difference from the expected one (might be to computation precision).

If the problem is in WREN, then fixing it would probably cause many other issues.

I'm continuing with the investigation.

PR #6243, opened 2023-06-21
Fix bounding sphere updates

stefaniapedrazzi:
Address part of #6063:
* [x] fix bounding sphere visualization by activating the bounding sphere updates when node position or size is changed
* [x] fix camera recognition functionality by activating the bounding sphere updates when node position or size is changed
* [x] restore performance optimization in bounding sphere checks (see https://github.com/cyberbotics/webots-dev/pull/7570)
* [x] improve performance by skipping parent notification if bound space is already dirty


BenjaminDeleze (approved):
Thank you!

PR #6245, opened 2023-06-21
Remove SUMO from Webots package

stefaniapedrazzi:
As discussed in  #6232, remove SUMO from the Webots package and document how to install it on the system.


projects/default/controllers/sumo_supervisor/sumo_supervisor.py
```diff
@@ -24,22 +24,28 @@
...
-        sumoPath = os.path.join(sumoPath, 'projects', 'default', 'resources', 'sumo')
-        os.putenv("SUMO_HOME", sumoPath)
+        sumoImportError()
+
     if sys.platform.startswith('darwin'):
         libraryVariablePath = 'DYLD_LIBRARY_PATH'
-
         # Fix the following error (shown on "El Capitan"):
         #    ./projects/default/resources/sumo/sumo-gui
         #    Fontconfig error: Cannot load default config file
```
stefaniapedrazzi, line 51: It should be checked if this is still needed on macOS

stefaniapedrazzi:
I completed to remove the sumo files from the Webots package and development system, as well as updated the documentation to refer to the system SUMO installation.

Instructions to install SUMO are currently in these two files:
* projects/default/controllers/sumo_supervisor/sumo_supervisor.py
* docs/automobile/sumo-interface.md

Currently I tested the changes only on Linux.
We should also test them on macOS and Windows.

stefaniapedrazzi:
@ygoumaz could you please check the changes for macOS?

projects/default/controllers/sumo_supervisor/sumo_supervisor.py
ygoumaz: It works on macOS without those lines.

omichel:
I tested on Windows 11 and it works like a charm.

ygoumaz:
I have tested on all platforms and it works nicely. It is very easy to install and use the SUMO package, while decreasing the size of the Webots one.

omichel (approved):
Thank you.

PR #6246, opened 2023-06-21
Fixed scaled rotation matrix for Pose (and derived) nodes

stefaniapedrazzi:
After #5978, if the Pose (or derived) node has a scaled parent Transform then the `WbAbstractPose::rotationMatrix()` function returns the scaled 3x3 submatrix instead of the rotation matrix.

Even for pose, the extracted 3x3 matrix needs to be unscaled in order to return a valid rotation matrix.


omichel (approved):
That looks good to me.
Thank you.

PR #6250, opened 2023-06-23
Unify ROSbots links with rosbot_description

delihus:
**Description**
Unify ROSbots link with https://github.com/husarion/rosbot_ros.
**Related Issues**
https://github.com/cyberbotics/webots_ros2/pull/770


ygoumaz:
Just to get a little more context, is there any benefit to apply these changes? Because we are introducing some less clear and redundant names (namely with the `link` keyword) for the nodes of the robot.

delihus:
> Just to get a little more context, is there any benefit to apply these changes? Because we are introducing some less clear and redundant names (namely with the `link` keyword) for the nodes of the robot.

We switched to publish robot description from .proto file here https://github.com/cyberbotics/webots_ros2/pull/770/. It means that the links names are the same as in the .proto file. 

If there is an user who want to test the software on the simulation and then switch to the real robot nothing has to be changed, any link name, topic and so on. 

I think it is good practice to simulate the real robot.

ygoumaz:
I see the point and I totally agree that we should be able to match the Webots node names with link names of the real robot. However I am not sure changing the names in the PROTO file directly is really the best and cleanest solution. I would instead implement a new parameter in the `webots_ros2_driver` package (`WebotsController` ROS2 node) to pass a dictionary of key-node-name->value-link-name. This way it would be possible to define for necessary nodes the corresponding value of the real robot and have exactly the same topic names. Do you think this solution could work for your use case?

delihus:
Okay! It sounds good especially, taking into account all remappings I made here https://github.com/husarion/webots_ros2/blob/develop-husarion/webots_ros2_husarion/launch/rosbot_launch.py#L72.

ygoumaz:
I will implement it start of next week. Meanwhile I will also check the status of https://github.com/cyberbotics/webots_ros2/pull/770 and try to fix the tests if you are still blocked. It is not mandatory to merge the PR before the release of R2023b. We can also release a new version of `webots_ros2` later.

PR #6251, opened 2023-06-23
Fix right click macos

### LINKED ISSUE #6244: Right click menu stacking up on macOS
On macOS, right-clicking multiple times in the 3D view results in multiple menus instead of removing them at each second click.

BenjaminDeleze:
**Description**
Fix #6244 

This issue is specific to macOS.

Since the upgrade to Qt 6.4.3, the context menu (opened with a right click) of the 3d view was not closed when clicking in the 3d view, worse it was possible to open several context menus.

The issue was not fixed by upgrading to Qt 6.5.1.

Here is what I could gather on the issue:

- Normally, clicking anywhere (`mousePressEvent`) but on the context menu is supposed to close it.
- However, on macOS it seems that clicking on the 3d view does not send the signal to the menu. Instead, the signal is send, as normally, to the 3d view itself (which can result in another menu being opened).
- Clicking on another part of the Webots GUI like the terminal, editor, scene tree will close the menu as expected.
- One hypothesis is that it is working anywhere excepted in the 3d view because it is the only one that is not derived from a QWidget but from a QWindow.

Fix

- Either we revert to 6.2.3 and wait for Qt to fix the bug (as I think that it is a Qt's bug). However, I saw nobody mention this bug for Qt 6 so far. This bug was present in some version of Qt and it took them several years to fix it (see [here](https://bugreports.qt.io/browse/QTBUG-29497)), so I am not sure it is the best idea. 

- Or we apply the fix of this PR. The only strange behavior I remarked with this solution is that if we open a scene tree context menu and then click in the 3d view it will not close it, but if we click in the scene tree it will close it.


src/webots/gui/WbView3D.cpp
```diff
@@ -1567,6 +1569,14 @@ void WbView3D::mousePressEvent(QMouseEvent *event) {
       mDragVerticalAxisRotate)
     return;
 
+
+  // Fix an issue on macOS where the context menu was not closed by a click.
+  #ifdef __APPLE__
+  QMenu *contextMenu = mParentWidget->findChild<QMenu*>("ContextMenu");
+  if (contextMenu)
+    delete contextMenu;
```
omichel, line 1577:
```suggestion
  delete mParentWidget->findChild<QMenu*>("ContextMenu");
```

omichel (changes requested):

omichel (approved):
Thank you.

PR #6255, opened 2023-06-27
Added missing line feed and final dots

omichel:
Added a missing line feed and a couple of missing final dots in stderr messages from libController.


BenjaminDeleze (approved):
Thank you

Issue #6140, opened 2023-05-15
Fail to start on Windows 


SpyrosKou:
**Describe the Bug**
After installing the [webots-R2023a_setup.exe](https://github.com/cyberbotics/webots/releases/download/R2023a/webots-R2023a_setup.exe) I get a small pop up window saying:

> The application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

I see that a similar issue [has been also encountered in the past.](https://github.com/cyberbotics/webots/issues/33) 

**Steps to Reproduce**
1. [Install the R2023a version](https://github.com/cyberbotics/webots/releases/download/R2023a/webots-R2023a_setup.exe)
2. Start the installed version
3. After uninstalling and reinstalling or reinstalling directly, the same behavior is observed.
4. Selecting to install for "All Users" and "For current user" results in the same behavior.

**Expected behavior**
Webots should have started after installation.

**Screenshots**
![image](https://github.com/cyberbotics/webots/assets/21125056/eb8f0b05-6602-42f5-80f9-551e35b5cdd3)

**System**
 - Operating System: Microsoft 11 Pro
 - Graphics Card: NVIDIA RTX A3000 Laptop GPU

SpyrosKou:
@omichel , thank you for your answer.
I  tried [nightly_12_5_2023/webots-R2023a-rev1](https://github.com/cyberbotics/webots/releases/download/nightly_12_5_2023/webots-R2023a-rev1_setup.exe) and the result is the same.

I also did a reinstall of the [nightly_12_5_2023/webots-R2023a-rev1](https://github.com/cyberbotics/webots/releases/download/nightly_12_5_2023/webots-R2023a-rev1_setup.exe) version.

omichel:
That is strange as you are the only one user reporting such a problem. It could come from a conflict with another application installed on your computer. Do you have some other Qt6 libraries installed somewhere on your computer?
To investigate the problem, I would recommend you to install [MSYS2](https://www.msys2.org) on your system, and from the MSYS2 console, type the following:
```bash
$ cd /C/Program\ Files/Webots/msys64/mingw64/bin/
$ ldd webots-bin.exe
```
On my computer I am getting the following output:
```
        ntdll.dll => /c/WINDOWS/SYSTEM32/ntdll.dll (0x7ffc7f9f0000)
        KERNEL32.DLL => /c/WINDOWS/System32/KERNEL32.DLL (0x7ffc7e750000)
[...]

PR #6258, opened 2023-06-27
Removed conflicting Qt5 environment variable on Windows

omichel:
Address #6140.

This should fix a problem that several users are facing.


BenjaminDeleze (approved):
Thank you

PR #6301, opened 2023-07-06
Fix fog billboard js

BenjaminDeleze:
The parsing of these nodes was not adapted when we added a root node in the webotsjs scenetree


omichel (approved):
Thank you.

PR #6302, opened 2023-07-06
Cleanup instanceof

BenjaminDeleze:
Replace `instanceof` by a check with `nodeType` when possible because it:
  - improves the performance
| Node | Instanceof | nodeType | 
| --- | --- | --- |
|Background| 0.06ms | 0.019 ms | 
|Shape| 0.27ms  | 0.077ms |
|Robot| 0.28ms | 0.09ms |
  - moreover using instanceof was the cause of a lot of imports and we had circular dependencies errors by the past.

The only reason to still use `instanceof`sometimes it to check for a node and its derived ones. For example, when checking
```if (node instanceof WbGroup)```it checks against group, but also, pose, transform, solid, and so on.

This PR also adds support for `castShadows` and `isPickable` in WbShape that were missing.


omichel (approved):
Nice clean-up.

Issue #6298, opened 2023-07-05
What is the difference between WbAxisAngle and WbRotation?


stefaniapedrazzi:
Both classes seem to represent a rotation in axis-angle format.
`WbAxisAngle` class has been introduced in Webots R2021a and it is used only by the `WbInertialUnit` class.
Why do we need two classes for this?

Note also that the code for converting a matrix to axis-angle seems to be duplicated in `WbRotation::fromMatrix3` and `WbMatrix3::toAxisAngle`.

omichel:
Yes, some clean-up is probably needed here.

stefaniapedrazzi:
I checked just to make sure that both the methods to compute the axis-angle give the same result and they are equivalent.

PR #6309, opened 2023-07-07
Fix handling scaled nodes in Webots/WREN

stefaniapedrazzi:
**Description**
Scale is integrated in the transform matrix computation differently in Webots and WREN causing mismatches.

**Related Issues**
This pull-request also fixes issues #6298  and #6063.

**Tasks**
  - [x] fix transform matrix computation in Webots to match WREN 
  - [x] fix data displayed in `Node Editor > Position Viewer`(#6063) 
  - [x] cleanup duplicated code to convert from/to different rotation format


stefaniapedrazzi:
Actually, the formulas used in WREN are the common ones used to correctly handle non-uniform scale and correctly apply the scale without shear component.

But then. the formulas in Webots to compute the absolute matrix and absolute position don't match what is displays by WREN and should be updated accordingly to have consistent data.

Instead of multiplying directly the transform matrices to retrieve the absolute pose, the following formulas are used (pseudo-code):
```
absoluteTranslation = parent->absoluteTranslation + parent->absoluteOrientation * (parent->absoluteScale * relativeTranslation);
absoluteRotation = parent->absoluteRotation * relativeRotation;
absoluteScale = parent->absoluteScale() * relativeScale;
matrix.fromVrml(absoluteTranslation, absoluteRotation,absoluteScale);
```

ygoumaz (approved):

PR #6313, opened 2023-07-10
Fix apt installation instructions

### LINKED ISSUE #6267: Update instructions to authenticate apt repository
Currently the instructions from the user guide to authenticate the APT repository work fine but are throwing a deprecation warning
```sh
$ wget -qO- https://cyberbotics.com/Cyberbotics.asc | sudo apt-key add -
Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).
OK
```

omichel:
Fixes #6267.
See it [here](https://cyberbotics.com/doc/guide/installation-procedure?version=fix-apt-installation-instructions#installing-the-debian-package-with-the-advanced-packaging-tool-apt).

### DIFF
--- a/docs/guide/installation-procedure.md
+++ b/docs/guide/installation-procedure.md
@@ -34,6 +34,7 @@ First of all, Webots should be authenticated with the [Cyberbotics.asc](https://
 
 > **Note**: You can check with `apt-key list` if this signature file was already installed using the deprecated `apt-key add` method.
 If so, you should delete it with `apt-key del <keyid>` before proceeding with the re-installation.
+Similarly, if the repository was already listed, you should remove it using `apt-add-repository -y --remove 'deb https://cyberbotics.com/debian/ binary-amd64/'`.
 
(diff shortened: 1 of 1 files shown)


stefaniapedrazzi (commented):
The new code works fine.

However, if the `Cyberbotics.asc` is already installed with the previous method, then after running the new instructions there are errors that completely block the APT package manager.
So it would be nice to also document how to remove the previously added authentication.

Here is the error:
```sh
/etc/apt/keyrings$ sudo apt update
E: Conflicting values set for option Signed-By regarding source https://cyberbotics.com/debian/ binary-amd64/: /etc/apt/keyrings/Cyberbotics.asc != 
E: The list of sources could not be read.
```

omichel:
I just added the requested documentation.

stefaniapedrazzi (dismissed):
Thank you.

Just note that for me it is not enough to delete the key but I also have to open the `Software & Updates` app, and delete the `https://cyberbotics.com/debian/ binary-amd64/` entries from the list in the `Other Software` tab.

omichel:
I just added some instructions on how to remove the repository if installed with the previous method.

PR #6320, opened 2023-07-11
Fix assertion failure with Skin node

stefaniapedrazzi:
Bounding sphere update could also be requested by the Skin node, so the assertion is wrong and results in a failure.
The assertion is not necessary and can be removed.

This issue occurs only when compiling in debug mode on not with the released versions of Webots.


omichel (approved):
Thank you.

PR #6324, opened 2023-07-13
Update rotating lidar angle range

ygoumaz:
The rotating Lidar image is given from 0 to 2 * M_PI. The normal lidar is giving it from -FOV/2 to + FOV/2. For a better consistency and also improve compatibility with ROS 2, the rotating Lidar image should be given from -M_PI to + M_PI. To do this, an offset is added to the current value of the angle to save values at the correct place in the data buffer. The point cloud computation must also be adapted. 

Note: I don't know if this is considered as breaking the backwards compatibility.


BenjaminDeleze (commented):
Thank you,
I would say that it breaks the backward compatibility as the generated "image" is not the same (You can test it by looking at the robot window)
Thus, I would put an entry in the changelog and target develop.

BenjaminDeleze (approved):
Thank you

PR #6327, opened 2023-07-17
Cleanup depecrated fields of display

BenjaminDeleze:
It seems that the related code has already been removed in https://github.com/cyberbotics/webots/pull/5266


ygoumaz (approved):
Thank you!

Issue #6341, opened 2023-07-25
Bad version number


ShuffleWire:
I've notice two things about version number : 

Many are not accurate : in R2023b, there are still many reference to R2023a, and even in translation there are wrong link pointing to the upgrade path from R2022a to R2022b
`<source>Please adapt your project to R2023a following these instructions: https://cyberbotics.com/doc/guide/from-2022a-to-2022b</source>`

I think we should update those, but I don't know on which version to start doing so, and ensuring that next version will still follows the right pattern.

omichel:
Thank you for reporting this.
You should fix it on the master branch and it will be back-ported to the develop branch afterwards.

PR #6342, opened 2023-07-25
Fix version number to 2023b

ShuffleWire:
**Description**
Fix https://github.com/cyberbotics/webots/issues/6341

I've been careful while updating URL and checked if the new link was still ok, so it should be alright. Notice a change of extension of 6 images of a background.

I've not been too aggressive in the compatibility tools (script, world file) that still have old number, because I believe it's better to leave them as it.


ygoumaz (approved):
I restored an URL on a backwards compatibility test. Otherwise it seems fine to me. Thank you!

PR #6355, opened 2023-08-15
Fix ljoint bo

BenjaminDeleze:
The scaling parameter was only applied on the graphical part but not on the bounding object


omichel (approved):
Thank you.

Issue #6362, opened 2023-08-23
Cannot open Matlab desktop when using "batch" option


matl-hsk:
https://github.com/cyberbotics/webots/blob/e9f93aef28dafec9f2864e65de92142d95c1abda/src/controller/launcher/webots_controller.c#L905

@nilsjor 
Unfortunately the change introduced in commit b43d3cd prevents (at least on Mac) to open Matlab desktop.
If Matlab is started with the argument "-batch" the command "desktop" cannot be run and produced the error
> Error using matlab.internal.lang.capability.Capability.require
> Support for user input is required, which is not available on this platform.
> Error in desktop

Hence this should be reverted to the arguments "-nodisplay", "-nosplash", "-nodesktop", "-r" instead.

omichel:
Thank you for this bug report.
Could you please provide a patch for this with a PR?

nilsjor:
First of all, I should mention that I think the fix https://github.com/cyberbotics/webots/pull/6364 is great since it gives the user the option to choose. 

Could you please help me understand what the actual issue is? You mention that `batch` prevents to open Matlab desktop". This is also the point of using `nodesktop`. 

Reading the documentation for [macOS](https://www.mathworks.com/help/matlab/ref/matlabmacos.html) and [Windows](https://www.mathworks.com/help/matlab/ref/matlabwindows.html), we can see that the options 
"-nodesktop" and "-nodisplay" (as well as "nojvm") are only present for macOS. 

My experience on Windows is that the `-batch` is a strict improvement—as described [here](https://github.com/cyberbotics/webots/pull/5943#issuecomment-1451954479). Especially since the other options are deprecated and/or not supported. What functionality is it that you have lost on Mac?

nilsjor:
Upon re-reading your comment a little more, I think it is becoming clearer to me.

Are you calling the command `desktop` as part of your controller code? If so, could you clarify what the use-case for this would be?

matl-hsk:
Several points:
- I completely agree, that in case you just want matlab to interpret the code, the batch option should be used. This is the official way currently recommended from mathworks and seems to work best.
- Background of the issue: I wanted to use the matlab environment (command window and debugging options). The webots documentation says.
> add "desktop; keyboard;" to the beginning of the controller code to bring up matlab.

However, this does not work anymore in batch mode, at least not on Mac.
- Possible solution: Seems to get complicated there since matlab has different startup options on different OS and more over some are not officially supported.


[...]

PR #6364, opened 2023-08-24
FIX:Revert matlab batch option to nodesktop

matl-hsk:
**Description**
See issue #6362

**Related Issues**
This pull-request fixes issue #6362

**Tasks**
  - [x] Revert back to using option "-nodesktop" instead of "-batch" for matlab controller
  - [x] Test on Mac
  - [ ] Test on Linux
  - [ ] Test on Windows
  - [x] Add option in preferences to use batch mode since it seems faster

**Documentation**
If this pull-request mainly changes the code to make it compatible with the documentation again. Added remarks in the matlab examples files that are in this repository.

**Screenshots**
None

**Additional context**
None

### DIFF
--- a/projects/robots/robotis/darwin-op/controllers/symmetry_matlab/symmetry_matlab.m
+++ b/projects/robots/robotis/darwin-op/controllers/symmetry_matlab/symmetry_matlab.m
@@ -2,7 +2,7 @@
 function symmetry_matlab
 
 % uncomment the next two lines if you want to use
-% MATLAB's desktop and interact with the controller
+% MATLAB's desktop and interact with the controller. Must disable the batch option in the preferences.
 %desktop;
 %keyboard;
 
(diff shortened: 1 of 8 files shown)


omichel:
I tested on Windows, but unfortunately, it doesn't work with the batch mode disabled. I see a MATLAB console opening, but the controller is not executed.

omichel:
*Note:* I just fixed the clang-format errors.

matl-hsk:
> I tested on Windows, but unfortunately, it doesn't work with the batch mode disabled. I see a MATLAB console opening, but the controller is not executed.

Thanks for testing!
The "desktop" option is an undocumented feature, which they might have dropped now. See my comment in the issue description for my suggestion how one could fix the issue more cleanly given Matlabs different options support on different OS.

PR #6366, opened 2023-08-28
Enable launching MATLAB desktop from the extern launcher

nilsjor:
**Description**
This will add features as described in https://github.com/cyberbotics/webots/issues/6362#issuecomment-1695794195

**Related Issues**
This pull-request fixes issue https://github.com/cyberbotics/webots/issues/6362

**Tasks**
  - [x] Add the `--interactive` option to the webots-controller which will launch MATLAB in interactive (debug) mode.
  - [x] Remove references to deprecated commands in example files.
  - [x] Change the documentation to reflect the new workflow. 

**Documentation**
The following pages have been changed:
- https://cyberbotics.com/doc/guide/matlab#using-the-matlab-desktop
- https://cyberbotics.com/doc/guide/running-extern-robot-controllers

### DIFF
--- a/lib/controller/matlab/launcher.m
+++ b/lib/controller/matlab/launcher.m
@@ -181,9 +181,11 @@ try
   if ~isvarname(WEBOTS_CONTROLLER_NAME)
     newname = matlab.lang.makeValidName(WEBOTS_CONTROLLER_NAME);
     copyfile(append(WEBOTS_CONTROLLER_NAME, '.m'), append(newname, '.m'), 'f');
+    if desktop('-inuse'), dbstop('in', newname); end
     eval([newname, args]);
     delete(append(newname, '.m')); % delete temporary file
   else
+    if desktop('-inuse'), dbstop('in', WEBOTS_CONTROLLER_NAME); end
     eval([WEBOTS_CONTROLLER_NAME, args]);
   end
 
--- a/src/controller/launcher/webots_controller.c
+++ b/src/controller/launcher/webots_controller.c
@@ -44,6 +44,7 @@ static char *controller;
 static char *controller_path;
 static char *controller_extension;
 static char *matlab_path;
+static char *matlab_args;
 static char *current_path;
 static int nb_controller_arguments;
 static int next_argument_index;
@@ -233,11 +234,12 @@ static void print_options() {
     "connect.\n    1234 is used by default, as it is the default port for Webots.\n    This setting allows you to connect to a "
     "specific instance of Webots if\n    there are multiple instances running on the target machine.\n    The port of a Webots "
     "instance can be set at its launch.\n\n  --robot-name=<robot-name>\n    Target a specific robot by specifying its name in "
-    "case multiple robots wait\n    for an extern controller in the Webots instance.\n\n  --matlab-path=<matlab-path>\n    For "
-    "MATLAB controllers, this option allows to specify the path to the\n    executable of a specific MATLAB version.\n    By "
-    "default, the launcher checks in the default MATLAB installation folder.\n    See "
-    "https://cyberbotics.com/doc/guide/running-extern-robot-controllers#running-a-matlab-extern-controller\n    for more "
-    "information.\n\n  --stdout-redirect\n    Redirect the stdout of the controller to the Webots console.\n\n  "
+    "case multiple robots wait\n    for an extern controller in the Webots instance.\n\n  --matlab-desktop\n    Launch the "
+    "full desktop version of MATLAB.\n    See https://cyberbotics.com/doc/guide/matlab#using-the-matlab-desktop for\n    more "
+    "information.\n\n  --matlab-path=<matlab-path>\n    For MATLAB controllers, this option allows to specify the path to the "
+    "\n    executable of a specific MATLAB version.\n    By default, the launcher checks in the default MATLAB installation "
+    "folder.\n    See https://cyberbotics.com/doc/guide/running-extern-robot-controllers#running-a-matlab-extern-controller\n"
+    "    for more information.\n\n  --stdout-redirect\n    Redirect the stdout of the controller to the Webots console.\n\n  "
     "--stderr-redirect\n    Redirect the stderr of the controller to the Webots console.\n\n");
 }
 
@@ -250,6 +252,7 @@ static bool parse_options(int nb_arguments, char **arguments) {
 
   controller = NULL;
   matlab_path = NULL;
+  matlab_args = NULL;
   char *protocol = NULL;
   char *ip_address = NULL;
   char *port = NULL;
@@ -273,6 +276,10 @@ static bool parse_options(int nb_arguments, char **arguments) {
         const size_t robot_name_size = strlen(arguments[i] + 13) + 1;
         robot_name = malloc(robot_name_size);
         memcpy(robot_name, arguments[i] + 13, robot_name_size);
+      } else if (strncmp(arguments[i], "--matlab-desktop", 16) == 0) {
+        const size_t matlab_args_size = snprintf(NULL, 0, "-nosplash -r") + 1;
+        matlab_args = malloc(matlab_args_size);
+        sprintf(matlab_args, "-nosplash -r");
       } else if (strncmp(arguments[i], "--matlab-path=", 14) == 0) {
         const size_t matlab_path_size = strlen(arguments[i] + 14) + 1;
         matlab_path = malloc(matlab_path_size);
@@ -723,6 +730,7 @@ static char **add_controller_arguments(char **argv, char **controller_argv, size
 static void free_memory() {
   free(WEBOTS_HOME);
   free(matlab_path);
+  free(matlab_args);
   free(current_path);
   free(controller);
 
@@ -885,6 +893,12 @@ int main(int argc, char **argv) {
     if (!matlab_path && !get_matlab_path())
       return -1;
 
+    if (!matlab_args) {
+      const size_t matlab_args_size = snprintf(NULL, 0, "-batch") + 1;
+      matlab_args = malloc(matlab_args_size);
+      sprintf(matlab_args, "-batch");
+    }
+
 #ifdef _WIN32
     const char *launcher_path = "\\lib\\controller\\matlab";
 #elif defined __APPLE__
@@ -902,7 +916,7 @@ int main(int argc, char **argv) {
     char **new_argv = NULL;
     new_argv = add_single_argument(new_argv, &current_size, matlab_path);
     new_argv = add_single_argument(new_argv, &current_size, matlab_command);
-    new_argv = add_single_argument(new_argv, &current_size, "-batch");
+    new_argv = add_single_argument(new_argv, &current_size, matlab_args);
     new_argv = add_single_argument(new_argv, &current_size, "launcher");
     if (nb_controller_arguments)
       new_argv = add_controller_arguments(new_argv, argv, &current_size, true);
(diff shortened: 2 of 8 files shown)


omichel:
It looks very good to me. Remain to address the documentation.

lib/controller/matlab/launcher.m
nilsjor, line 1: Only in interactive mode: add breakpoints to the first line in the controller file with `dbstop`.
Ignored in batch mode since `desktop` returns FALSE

src/controller/launcher/webots_controller.c
nilsjor, line 1: New char array `matlab_args` is initially NULL. Is updated only by passing the appropriate argument to webots-controller.

If provided, `-nosplash -wait -r` is used, which starts interactive mode, without splash screen. Blocks launcher from exiting before MATLAB. Otherwise, `batch` is used like normal.

nilsjor:
@matl-hskcould you please help testing on macOS? 

1. Open the "mir100" world in webots.
2. Change the "controller" field of Mir100 (the white AGV) to `<extern>`.
3. Run the Webots simulation. Nothing should happen until an extern controller has connected.
4. In a new MSYS window, call the following:   
```sh
webots-controller --interactive ~/webots/projects/robots/mir/mir100/controllers/keyboard_matlab/keyboard_matlab.m
  ```
5. Wait until MATLAB opens. Click "Continue" in the debugging toolstrip. 
6. The simulation should start. Click on the Mir100 robot. Try to drive it with the arrow keys of your keyboard.

**Comment by matl-hsk** [comment]
sorry for the delay.

I've tested your branch but it does not open Matlab, neither with the interactive option nor without. I can however
- start Matlab and run the launcher manually (similar to how you described it in the issue)
- start Matlab in batch mode from webots

**Comment by nilsjor** [comment]
> sorry for the delay.
> 
> I've tested your branch but it does not open Matlab, neither with the interactive option nor without. I can however
> 
>     * start Matlab and run the launcher manually (similar to how you described it in the issue)
> 
>     * start Matlab in batch mode from webots

Can you provide the output from the terminal? 

Have you tried using `webots-controller` for any other controller, python, etc.?

**Comment by matl-hsk** [comment]
unfortunately it does not give any hint on the error. I looked at the code but could not find an issue quickly.

> The started controller targets a local instance (ipc protocol) of Webots with port number 1234.
> Targeting the only robot waiting for an extern controller.
> Running MATLAB in interactive mode...

**Comment by nilsjor** [comment]
Note that it takes a while to start Matlab. The launcher is written to wait until the instance closes. After calling the externlauncher, is control ever returned to you, or dou need to Ctrl-C?

What do you get if you run without the interactive flag?

Have you tried providing the matlab path, perhaps it cannot find it?

Mvh
________________________________

**Comment by matl-hsk** [comment]
> Have you tried providing the matlab path, perhaps it cannot find it?

This was it! From webots itself it can find Matlab and from the terminal I can start it as well.

Suggestion: Instead of "Running Matlab ..." it should rather print an error when it cannot find Matlab.

Two more issues
- The "-wait" option for Matlab is only available for windows but not on Linux or Mac.
- For the webots command a small script to start it is created in the webots home folder but not for the webots-controller. For the latter one has to dig into the folder "Contents/MacOS" to find it. I would rather make this consistent. Or is this on purpose?

**Comment by nilsjor** [comment]
> Suggestion: Instead of "Running Matlab ..." it should rather print an error when it cannot find Matlab.

This is a great observation! But the solution lies outside the scope for this PR. As you saw, this bug appears also for non-interactive use... If you you could, please open another Issue; "No error generated when webots-controller is unable to find executable" or similar ;)

For the future, I believe the underlying reason has to do with the configuration of your MSYS. You could check the PATH variables, it is likely that you are missing something. 

> The "-wait" option for Matlab is only available for windows but not on Linux or Mac.

That's interesting. Perhaps it is already the default behaviour on unix that the MATLAB window is blocking... Could you test that for me? Just run `matlab` in a terminal and check if control is returned at any point.

> For the webots command a small script to start it is created in the webots home folder but not for the webots-controller. For the latter one has to dig into the folder "Contents/MacOS" to find it. I would rather make this consistent. Or is this on purpose?

This is also a separate Issue, feel free to report it if you want. It could also have to do with your MSYS installation, it would have to be reproduced by someone else.

Finally—regarding the features of this PR—did you get the interactive debugging to work? Were you able to reproduce [these 6 steps](https://github.com/cyberbotics/webots/pull/6366#issuecomment-1697407770)?

> 1. Open the "mir100" world in webots.
> 2. Change the "controller" field of Mir100 (the white AGV) to `<extern>`.
> 3. Run the Webots simulation. Nothing should happen until an extern controller has connected.
> 4. In a new Terminal window, call the following:
> 
> ```shell
> webots-controller --interactive ~/webots/projects/robots/mir/mir100/controllers/keyboard_matlab/keyboard_matlab.m
> ```
> 
> 5. Wait until MATLAB opens. Click "Continue" in the debugging toolstrip.
> 6. The simulation should start. Click on the Mir100 robot. Try to drive it with the arrow keys of your keyboard.

**Comment by matl-hsk** [comment]
On topic:
- Yes, Matlab blocks the terminal until it is closed.
- It works all as intended when I specify the Matlab path.

Off topic:
- I understand that you want to prevent feature creep in this PR - okay with me to not fix the other issues.
- Regarding the path: Matlab is in the user path (usr/local/bin) but apparently the app is started as another user or does not use the path variable. I am mostly a Linux user and don't know how this is handled on Mac.

**Comment by nilsjor** [comment]
> * Yes, Matlab blocks the terminal until it is closed.
> 
> * It works all as intended when I specify the Matlab path.

Great, thanks! 

As a last touch, I moved the printout so it will only appear if a valid installation is found. I also removed `-wait` from non-windows versions.

Out of curiosity, I had a look at why it does not warn, and I found this https://github.com/cyberbotics/webots/blob/58d1b509a81ad33854b88fcbbec46150dd325814/src/controller/launcher/webots_controller.c#L187-L192

It seems that the error is suppressed only on macOS. I could not tell you why... Please include this in your Issue, if you open one.

File: src/controller/launcher/webots_controller.c
Comment (nilsjor) [reply src/controller/launcher/webots_controller.c:1]: Update: `-nosplash` is removed to ensure that the splash screen is shown. A user who want to use the desktop "like normal" for debugging would likely expect it. 

Additionally, only Windows supports `-wait`, since on unix, MATLAB desktop is blocking by default.

**Comment by nilsjor** [comment]
Hello again. What else is needed to go ahead with this? I see it is still pending reviewers.
@ ygoumaz pehaps you can have a look, since I know you are familiar with the webots-controller launcher?

I have left some explanations on the important files:
* src/controller/launcher/webots_controller.c
* lib/controller/matlab/launcher.m

Other than that, it's just the documentation and removed references to `desktop` in some examples (thanks matl-hsk!)

**Comment by omichel** [comment]
I did a first review a while ago, but my requests for change were not yet taken into account or answered.

File: src/controller/launcher/webots_controller.c [hunk 1]
```diff
@@ -273,6 +276,16 @@ static bool parse_options(int nb_arguments, char **arguments) {
         const size_t robot_name_size = strlen(arguments[i] + 13) + 1;
         robot_name = malloc(robot_name_size);
         memcpy(robot_name, arguments[i] + 13, robot_name_size);
+      } else if (strncmp(arguments[i], "--interactive", 13) == 0) {
+#ifdef _WIN32
+        const size_t matlab_args_size = snprintf(NULL, 0, "-wait -r") + 1;
+        matlab_args = malloc(matlab_args_size);
```
omichel, line 282:
```suggestion
        matlab_args = malloc(strlen("-wait -r") + 1);
```

src/controller/launcher/webots_controller.c
```diff
@@ -273,6 +276,16 @@ static bool parse_options(int nb_arguments, char **arguments) {
...
+        sprintf(matlab_args, "-wait -r");
+#else
+        const size_t matlab_args_size = snprintf(NULL, 0, "-r") + 1;
+        matlab_args = malloc(matlab_args_size);
```
omichel, line 286:
```suggestion
        matlab_args = malloc(strlen("-r") + 1);
```

omichel (changes requested):

src/controller/launcher/webots_controller.c
```diff
@@ -273,6 +276,10 @@ static bool parse_options(int nb_arguments, char **arguments) {
         const size_t robot_name_size = strlen(arguments[i] + 13) + 1;
         robot_name = malloc(robot_name_size);
         memcpy(robot_name, arguments[i] + 13, robot_name_size);
+      } else if (strncmp(arguments[i], "--matlab-desktop", 16) == 0) {
+        const size_t matlab_args_size = snprintf(NULL, 0, "-nosplash -r") + 1;
+        matlab_args = malloc(matlab_args_size);
+        sprintf(matlab_args, "-nosplash -r");
```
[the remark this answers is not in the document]
nilsjor: I agree with the suggestions. The `snprintf` approach was only used to harmonize with the rest of the existing code.

matl-hsk:
Sorry for the late reply - I was too busy with other stuff. Thanks for the reminder.

Code looks good to me but that does not work as intended:
> As a last touch, I moved the printout so it will only appear if a valid installation is found. I also removed `-wait` from non-windows versions.

I just tried without specifying the Matlab path and it still prints "starting matlab" although it cannot find it. Until a proper solution is available, I'd suggest to change the phrase to give a hint that if nothing happens one should try to specify the path to matlab. (Hope that fits into the PR.)

nilsjor:
> I just tried without specifying the Matlab path and it still prints "starting matlab" although it cannot find it. Until a proper solution is available, I'd suggest to change the phrase to give a hint that if nothing happens one should try to specify the path to matlab. (Hope that fits into the PR.)

Thanks again for the feedback! I am still of the opinion that this is outside the scope of this PR. It seems to me that the problem lies somewhere in the function `get_matlab_path()`, which is unchanged in this PR.

To summarize, could you confirm the following:
1) Running webots-controller **without** `interactive` and **without** `matlab-path`: MATLAB never opens. 
2) Running webots-controller **with** `interactive` and **without** `matlab-path`: 
`"Running MATLAB in interactive mode..."`  ...but MATLAB never opens. 
3) Running webots-controller **without** `interactive` and **with** correct `matlab-path`: MATLAB runs in batch mode.
4) Running webots-controller **with** `interactive` and **with** correct `matlab-path`: 
`"Running MATLAB in interactive mode..."` and then MATLAB opens in interactive mode.

It appears that the launcher skips the following lines, even though the function should return `false` if there is not installation in the default path: https://github.com/cyberbotics/webots/blob/58d1b509a81ad33854b88fcbbec46150dd325814/src/controller/launcher/webots_controller.c#L899-L901


The expected behavior here is that the launcher exits and you get:
https://github.com/cyberbotics/webots/blob/58d1b509a81ad33854b88fcbbec46150dd325814/src/controller/launcher/webots_controller.c#L207-L208

Could you please create a [new issue](https://github.com/cyberbotics/webots/issues/new?assignees=&labels=&projects=&template=bug_report.md), call it "**macOS: No error generated when webots-controller cannot find executable**".

There, please list the steps needed to reproduce it on the webots R2023b release (i.e., without this PR), including the path to your installed version of MATLAB, macOS, etc. Also describe if the launcher exits or if you need to Ctrl+C to return control.

Thanks!

matl-hsk:
> 1. Running webots-controller **without** `interactive` and **without** `matlab-path`: MATLAB never opens.
> 2. Running webots-controller **with** `interactive` and **without** `matlab-path`:
>    `"Running MATLAB in interactive mode..."`  ...but MATLAB never opens.
> 3. Running webots-controller **without** `interactive` and **with** correct `matlab-path`: MATLAB runs in batch mode.
> 4. Running webots-controller **with** `interactive` and **with** correct `matlab-path`:
>    `"Running MATLAB in interactive mode..."` and then MATLAB opens in interactive mode.
all correct

matl-hsk:
> Could you please create a [new issue](https://github.com/cyberbotics/webots/issues/new?assignees=&labels=&projects=&template=bug_report.md), call it "**macOS: No error generated when webots-controller cannot find executable**".

Done and assigned you, @nilsjor.

omichel (dismissed):
Thank you both.

omichel (approved):

Issue #6395, opened 2023-09-28
Wrong relative Position of objects in recognition by spherical cameras with pi/2 < fov < pi .


TheMangalex:
**Describe the Bug**
Hello everyone, I have noticed some problems with recognition objects in spherical cameras:

The position of an object in a recognition with a spherical camera becomes wrong if the center of the object/or a bigger part of it isn't  in the camera frustum anymore. This happens for fovs < pi. For fov > pi this seems to be fine. (Tested with fov of 2.14 and 3.0)
The wrong object positions during this projection force wrong bounding boxes and a wrong position in the webots_ros/RecognitionObjects Message.

Potential reason might be the mIsOmniDirectional bool in WbObjectDetection. 
The header has the annotation
`bool mIsOmniDirectional;  // is sensor omnidirectional (horizontal FOV < PI/2)`    (https://github.com/cyberbotics/webots/blob/master/src/webots/nodes/utils/WbObjectDetection.hpp#L101 )
The cpp file implements it with
`mIsOmniDirectional(mHorizontalFieldOfView > M_PI),`   (https://github.com/cyberbotics/webots/blob/master/src/webots/nodes/utils/WbObjectDetection.cpp#L40).

Changing this line to 
`mIsOmniDirectional(mHorizontalFieldOfView > M_PI / 2.0),`
lets it look correct for me, but I'm unsure if it is correct in all cases or just the two I tested.

I have some screenshots showing the problem and how it looks fixed attached below:

**Steps to Reproduce**
1. Use a spherical camera with fov between pi/2.0 and pi
2. Put an object close to the view frustum corner so that parts of it are outside it and parts are inside (as seen in the screenshots)
3. Use a recognition image/ros message
4. You can see the bounding box corners being wrong/the object position and other values being wrong

**Expected behavior**
The bounding box should match the recognition and the recognized object position should be correct. See screenshots below

**Screenshots**
Fov 2.14 before
<img width="374" alt="fov2 14-wrong" src="https://github.com/cyberbotics/webots/assets/26520790/c5196c52-31d6-4d6c-aed7-82e114db4034">

Fov 3.0 before
<img width="374" alt="webots-fov-3-wrong" src="https://github.com/cyberbotics/webots/assets/26520790/b69da7f9-80af-4a41-b93e-9aa7110699b8">

Fov 2.14 after adjustments
[...]

PR #6396, opened 2023-09-29
fixed mIsOmnidirectional in WbObjectDetection

TheMangalex:
**Description**
Changes the calculation of mIsOmniDirectional to horizontalFov > pi / 2.

**Related Issues**
This pull-request fixes issue #6395

**Additional context**
Details on how the spherical projections were wrong before can be seen in #6395

### DIFF
--- a/docs/reference/changelog-r2023.md
+++ b/docs/reference/changelog-r2023.md
@@ -13,6 +13,7 @@ Released on XXX XXth, 2023.
     - Fixed translation, rotation and scale displayed in the Position tab of the Node viewer in the scene tree ([#6309](https://github.com/cyberbotics/webots/pull/6309)).
     - Replaced the [Mesh](mesh.md) bounding object of the ROSbot XL by [Boxes](box.md) ([#6326](https://github.com/cyberbotics/webots/pull/6326)).
     - Fixed a crash when [IndexedLineSet](indexedlineset.md) has `coord` but no `coordIndex` ([#6359](https://github.com/cyberbotics/webots/pull/6359)).
+    - Fixed values returned by the [Receiver.getEmitterDirection](https://cyberbotics.com/doc/reference/receiver?tab-language=python#wb_receiver_get_emitter_direction) Python method ([#6394](https://github.com/cyberbotics/webots/pull/6394)).
 
 ## Webots R2023b
 Released on June 28th, 2023.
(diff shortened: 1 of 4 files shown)


omichel (changes requested):
Thank you!
May I kindly ask you to add a line in the [changelog](https://github.com/Adman/webots/edit/fix-getemitterdirection/docs/reference/changelog-r2023.md) under "Bug fixes"?

TheMangalex:
Does this work for you?
https://github.com/cyberbotics/webots/pull/6396/commits/f650732e9de7fd76454b2c4a310ade49db38ef16

omichel (changes requested):
It looks good to me, but I believe [this test](https://github.com/cyberbotics/webots/blob/f650732e9de7fd76454b2c4a310ade49db38ef16/src/webots/nodes/utils/WbObjectDetection.cpp#L423) could be also simplified now.

TheMangalex:
That seems to be the case. I adapted it and used M_PI_2 instead of M_PI / 2.0 in the initial change now to remove the additional computation and stick to the use of M_PI_2.

omichel (approved):
That looks very good to me now.
Thank you!

omichel:
@TheMangalex: This PR apparently broke the [camera_recognition](https://github.com/cyberbotics/webots/blob/master/tests/api/worlds/camera_recognition.wbt) test, see: https://github.com/cyberbotics/webots/actions/runs/6386382958/job/17333860294#step:7:1254
May I kindly ask you to look into this and check the [camera_recognition](https://github.com/cyberbotics/webots/blob/master/tests/api/controllers/camera_recognition/camera_recognition.c) controller to understand why the test fails and eventually fix it?

TheMangalex:
I'll try to look into it in the following days

TheMangalex:
It seems that this broke the test for cylindrical projections which seem to be handled similar to spherical projections in this case..


Before PR:
webots_test_prev

After PR:
webots_test_fix

It seems to be a different issue which was created by this change, as the issue of close objects also existed for cylindrical cameras:
image


I'll have a deeper look, what causes this

omichel:
Thank you for investigating this.

TheMangalex:
I created #6403 as a draft which seems to fix the test and #6395 
In this PR I also added another object to the recognition test which is perpendicular to the camera sensor plane and only partially seen, as this was a problem in #6395

I want to add another test which specifically checks for the error which was fixed in #6395 , so I'll come back to this tomorrow.

PR #6402, opened 2023-10-04
Fix snap build

BenjaminDeleze:
Do not install wrenJS dependencies when building the snap are they are not needed.


omichel (approved):
Thank you!

Issue #5809, opened 2023-01-31
Missing point on lidar


ShuffleWire:
**Describe the Bug**
When using high quality Lidar, I run into the issue that some points were not shown. i don't know if it a rendering issue of the Lidar by Wren, or a point cloud display issue, didn't tried to inspect that, but I'm confident about the display of the point cloud.

**Steps to Reproduce**
1. Open the world and run two steps
[lidar_cylinder_missing_points.zip](https://github.com/cyberbotics/webots/files/10547127/lidar_cylinder_missing_points.zip)

2. See missing point on the border
![lidar_6](https://user-images.githubusercontent.com/83014015/215783503-7c9e21f0-9a09-4c36-8883-9a25f2f97298.png)

Same with this world : 
[lidar_cylinder_missing_points_2.zip](https://github.com/cyberbotics/webots/files/10547140/lidar_cylinder_missing_points_2.zip)
![lidar_7](https://user-images.githubusercontent.com/83014015/215783593-ee522c03-a16d-458a-b52d-2783fe6f969f.png)

**Expected behavior**
All point should be displayed.

**System**
 - Operating System: Debian 11
 - Graphics Card: None, that could be a source of issue, please someone with one test for me :)  
 - Webots master 045000f6f723f93a57e34434dac05ff74a745013

ShuffleWire:
Still present in R2023b

omichel:
Fixed in #6417.

Issue #6344, opened 2023-07-26
Strange jump in value outputed by Lidar


ShuffleWire:
**Describe the Bug**
Using this world,
[lidar_jump_in_values.zip](https://github.com/cyberbotics/webots/files/12170331/lidar_jump_in_values.zip)

I get that : 
![lidar](https://github.com/cyberbotics/webots/assets/83014015/73e406f1-e18a-42d6-90c9-3954c13c410f)
with those weird jumps.
Displaying the raw depth values show the same issue : 
![image](https://github.com/cyberbotics/webots/assets/83014015/cdea7a3a-7c01-479b-b5df-e3d0dfc89fdf)

Moreover, looking closely, the points are not on the plane... 
![image](https://github.com/cyberbotics/webots/assets/83014015/ff0ed3d8-e293-4c91-9353-63c8413d4069)
in fact, probing computed values from inside Webots show that they are between 2 and 5cm bellow the plane, which is a lot.

I initially thought that the depth buffer was missing resolution, and truncating value, but the outputted value are far from that, and reading the code i understand that the z-buffer is in float, so that should be ok

**Expected behavior**
I would expect to get a smooth curve.

**System**
 - Operating System: Debian 11
 - Graphics Card: GTX 1650
 - Webots 2023b

omichel:
Fixed in #6417.

Issue #6370, opened 2023-09-06
Lidar data incorrect


ShuffleWire:
**Describe the Bug**
Even in a basic configuration I get data from a Lidar being significantly wrong.
I've setup a world where a Lidar look perpendicularly to a flat surface.
I can manually compute the distance I should get, using basic trigonometry.
I'm comparing to the Lidar data and get error of more than 10mm.
Modifying `minRange` and `near` don't change anything.
PS: I llok like it's a duplicate of https://github.com/cyberbotics/webots/issues/4767 but I hope to give more accurate data on this bug, even if the original bug report stay valid.

**Steps to Reproduce**
1. Open world [measure_lidar.zip](https://github.com/cyberbotics/webots/files/12534812/measure_lidar.zip)
2. Build and run controller
3. See output : right values are the error in mm
Plotting it in the included Spreadsheet give : 
![image](https://github.com/cyberbotics/webots/assets/83014015/a501df0c-10f0-40bb-85be-8111f5a9e7ba)

**Expected behavior**
Error should be negligible.

**Screenshots**
The actual display in the GUI confirm that issue :  
![image](https://github.com/cyberbotics/webots/assets/83014015/c95249f9-9914-4a7c-bc65-19cd23882a89)
The point cloud is wrong.

**System**
I will try more system as soon as I have access to them, especially using real GPU (I have a 1650 at home), and I will pleased to hear from people trying on others configurations. I'm getting this exact behavior at least on those : 

- Debian 12 
```
System: Linux 6.1.0-11-amd64 x86_64
OpenGL vendor: Intel
OpenGL renderer: Mesa Intel(R) UHD Graphics 630 (CFL GT2)
OpenGL version: 4.6 (Core Profile) Mesa 22.3.6
Available GPU memory: N/A
```
AND (with software rendering)
```
System: Linux 6.1.0-11-amd64 x86_64
OpenGL vendor: Mesa/X.org
OpenGL renderer: llvmpipe (LLVM 15.0.6, 256 bits)
OpenGL version: 4.5 (Core Profile) Mesa 22.3.6
Available GPU memory: N/A
```

 - WSL : Ubuntu 20
 ```
 System: Linux 5.10.102.1-microsoft-standard-WSL2+ x86_64
OpenGL vendor: Microsoft Corporation
OpenGL renderer: D3D12 (NVIDIA Quadro T1000)
[...]

PR #6417, opened 2023-10-21
Fix missing points in lidar scan

### LINKED ISSUE #5810: Multiple "identical" Lidar give differents results
**Describe the Bug**
I'm playing with Lidar, and encounters this weird behavior : when using two Lidar, they seems to pertubate each other.
I've two identical Llidar, placed at the same position. One "wide" has a fov of 180°, and 512 points. The second "narrow" as a fov 16 times smaller and 16 times as well less points. Hence, in the region in front of the lidar, I was expecting to have exactly the same result, i.e. points cloud generated by those two Lidar should overlap. (of course the larger one will generate additional points, but that not important here).

The thing is, they do not.

Open and run for two step this world : 
[lidar_bad_overlap.zip](https://github.com/cyberbotics/webots/files/10547298/lidar_bad_overlap.zip)

You get : 
lidar_10

Clearly cloud don't overlap : the correct one is emitted by the "wide" Lidar, and the bad one is generated by the "narrow" Lidar
We could see, from the side, in orthographic view, that the bad point are in fact below the ground, as if they "gone trough" the ground, and eventually stopped. Alignment with the origin of the Lidar seems fine, thought.
lidar_8

I've circled in colors the points (broadly each line of the point cloud, but given the POV, they appear as one point each, approximately) : in green the good ones, in red the bad one, the white line is the floor (bold in order to be visible)

What is happening ? Why green and red points don't match ?

**System**
 - Operating System: Debian 11 master and develop 
 - Graphics Card: None, could someone with one could test for me ? EDIT : Tested with Webots 2023a-rev1 + GTX 1650 + Debian 11 on another machine, and same result

TonyWelte:
**Description**

This pull request fixes the issue with missing points in lidar scan.

I was trying (and failing) to fix #5810 and came across this issue. For horizontal FOV higher than around 1 (depends on the vertical FOV), there are corners of points that are missing.

##### Missing points

Looking at `faceCoord` in [merge_spherical.frag](https://github.com/cyberbotics/webots/blob/master/resources/wren/shaders/merge_spherical.frag) I noticed that the values in the corners fall outside the [0, 1] range.
In black pixels where `faceCoord.y` is not within [0, 1]:
Screenshot from 2023-10-21 16-56-19
Depth image:
Screenshot from 2023-10-21 16-54-12

Looking at which face of the cube is used for each pixels:
Screenshot from 2023-10-21 16-54-54

As far as I understand the lidar depth is computed by merge_spherical.frag as a depth image using 6 images computed for each side of a cube. However, the 6 images form a flattened cube (depending on the vertical and horizontal FOV). Also, the shape of the camera used to render the top and bottom images was not square but had the same shape as the sides (leading to parts of space not being rendered).

For now I have forced the rendering of the UP and DOWN part of the cube and reworked the code that finds which face to use, leading to the correct face being found and the missing points being rendered.
Screenshot from 2023-10-21 16-52-13
Screenshot from 2023-10-21 16-58-01

##### Points forming small groups

The issue now is that these points look bad. They form small lines as in (#6344). I think it is because with the resolution used to render the cube multiple points will correspond to the same pixel of the cube (and therefore using the same depth value).

To fix this issue I used `texture` instead of `texelFetch` to get the pixel value with interpolation. This leads to a cleaner point cloud:
Screenshot from 2023-10-21 17-40-46

Remaining issue
- the point at the edges of the cube can still look weird because the interpolation doesn't work on the edges
- the interpolation causes false points where there are discontinuities
Screenshot from 2023-10-21 17-47-38
- forcing the render of the UP and DOWN part of the cube is not ideal. I think it would be better force the cube to be an actual cube (not flattened) in which case the existing condition should work. I haven't quite figured out how the rendering of the sides works. I the sides are rendered by considering the vertical FOV the missing corner should have started for any horizontal FOV.

**Related Issues**

This pull-request fixes issue #5809 and #6344.

**Additional context**

Test world: [lidar_cylinder_missing_points.zip](https://github.com/cyberbotics/webots/files/13061055/lidar_cylinder_missing_points.zip)  (modified from #5809)

### DIFF
--- a/resources/wren/shaders/merge_spherical.frag
+++ b/resources/wren/shaders/merge_spherical.frag
@@ -34,47 +34,124 @@ uniform float fovYCorrectionCoefficient;
 
 uniform sampler2D inputTextures[6];
 
-vec4 smart_texture(sampler2D tex, vec2 p, int bias) {
-    ivec2 texSize = textureSize(tex, 0);
-    ivec2 pixelCoord = ivec2(p.x*(texSize.x-1), p.y*(texSize.y-1));
+// Same as texture but pick the side to used for the interpolation to avoid discontinuities
+vec4 smooth_texture(sampler2D tex, vec2 p, int bias) {
+  ivec2 texSize = textureSize(tex, 0);
+  ivec2 pixelCoord = ivec2(round((2 * texSize.x * p.x - 1) / 2), round((2 * texSize.y * p.y - 1) / 2));
 
-    vec2 delta = vec2(p.x*(texSize.x-1), p.y*(texSize.y-1)) - pixelCoord;
+  float dx = ((2 * texSize.x * p.x - 1) / 2 - pixelCoord.x);
+  float dy = ((2 * texSize.y * p.y - 1) / 2 - pixelCoord.y);
 
-    vec4 pixelCenter = texelFetch(tex, ivec2(pixelCoord.x, pixelCoord.y), bias);
-    vec4 pixelUp = texelFetch(tex, ivec2(pixelCoord.x, pixelCoord.y+1), bias);
-    vec4 pixelDown = texelFetch(tex, ivec2(pixelCoord.x, pixelCoord.y-1), bias);
-    vec4 pixelLeft = texelFetch(tex, ivec2(pixelCoord.x-1, pixelCoord.y), bias);
-    vec4 pixelRight = texelFetch(tex, ivec2(pixelCoord.x+1, pixelCoord.y), bias);
+  // Get depth for each corner
+  float dCenter = texelFetch(tex, ivec2(pixelCoord.x, pixelCoord.y), bias).x;
+  float dUp = texelFetch(tex, ivec2(pixelCoord.x, pixelCoord.y + 1), bias).x;
+  float dDown = texelFetch(tex, ivec2(pixelCoord.x, pixelCoord.y - 1), bias).x;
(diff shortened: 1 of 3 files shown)


TonyWelte:
I managed to fix the way the cube is generated (by modifying the aspect ratio) to avoid requiring the UP and DOWN sides. I had to increase the "mSphericalFovYCorrectionCoefficient" otherwise I still had one corner pixel that was considered in the UP or DOWN face.

I came across https://github.com/cyberbotics/webots/pull/3230#issuecomment-869489788 which explains the reason for choosing `texelFetch` instead `texture`. I have experimented with a custom function that works similarly to `texture` but picks which 4 pixels to used based on their variance. It should avoid interpolating using pixels that don't correspond to the same object. It It might be more demanding, but I'm not sure how to evaluate that.

Screenshot from 2023-11-01 18-35-20
Screenshot from 2023-11-01 18-38-38
Screenshot from 2023-11-01 18-43-57
Screenshot from 2023-11-01 19-05-30

While fixing this I also figured out the issue with #5810 so this pull request should fix #5810 #5809 #6344 #6370
 
Since the code for the lidar is shared with the rangefinder and the camera I still need to do more test to make sure it doesn't brake anything.

omichel:
Indeed, it actually broke a couple of tests, see https://github.com/cyberbotics/webots/actions/runs/6724384236/job/18277537984#step:7:1318

omichel (approved):
Thank you.

PR #6473, opened 2024-01-10
Avoid crashes and better warn when getting PROTO

galou:
Avoid crashes and better warn when getting the PROTO in `WbSupervisorUtilities::getProtoParameterNodeInstance()`.

**Description**
I discovered the bug when we used a cache to save nodes in Python. When calling `getPose()` or `getPosition()` we got an error message that `wb_supervisor_node_get_position() can exclusively be used with Pose (or derived)`. Now, the error message is `Cannot get the PROTO instance for node `LED` (type LED).` plus the old message. This helped me fixing the bug in our code.

**Related Issues**
None

**Tasks**
  - [ ] Translate the strings

**Documentation**
No documentation change

**Additional context**
We probably save the result of `getMFNode()` in Python too soon because the pointed object is not a valid node later in our code (maybe the pointer changes when changing the DEF?).

### DIFF
--- a/src/webots/nodes/utils/WbSupervisorUtilities.cpp
+++ b/src/webots/nodes/utils/WbSupervisorUtilities.cpp
@@ -503,7 +503,15 @@ WbNode *WbSupervisorUtilities::getProtoParameterNodeInstance(int nodeId, const Q
     mRobot->warn(tr("%1: node not found.").arg(functionName));
     return NULL;
   }
-  return static_cast<WbBaseNode *>(node)->getFirstFinalizedProtoInstance();
+  WbBaseNode *proto = static_cast<WbBaseNode *>(node)->getFirstFinalizedProtoInstance();
+  if (!proto)
+  {
+    if (node->modelName() != node->nodeModelName())
+        mRobot->warn(tr("Cannot get the PROTO instance for node '%1' (derived from '%2').").arg(node->usefulName(), node->nodeModelName()));
+      else
+        mRobot->warn(tr("Cannot get the PROTO instance for node '%1'.").arg(node->usefulName()));
+  }
+  return proto;
 }
 
 void WbSupervisorUtilities::changeSimulationMode(int newMode) {
@@ -594,7 +602,8 @@ void WbSupervisorUtilities::handleMessage(QDataStream &stream) {
       stream >> nodeId;
       const QString &stateName = readString(stream);
       WbNode *const node = getProtoParameterNodeInstance(nodeId, "wb_supervisor_node_load_state()");
-      node->reset(stateName);
+      if (node)
+        node->reset(stateName);
       return;
     }
     case C_SUPERVISOR_NODE_SAVE_STATE: {
@@ -602,7 +611,8 @@ void WbSupervisorUtilities::handleMessage(QDataStream &stream) {
       stream >> nodeId;
       const QString &stateName = readString(stream);
       WbNode *const node = getProtoParameterNodeInstance(nodeId, "wb_supervisor_node_save_state()");
-      node->save(stateName);
+      if (node)
+        node->save(stateName);
       return;
     }
     case C_SUPERVISOR_NODE_SET_JOINT_POSITION: {
--- a/src/webots/nodes/utils/WbSupervisorUtilities.cpp
+++ b/src/webots/nodes/utils/WbSupervisorUtilities.cpp
@@ -506,7 +506,8 @@ WbNode *WbSupervisorUtilities::getProtoParameterNodeInstance(int nodeId, const Q
   WbBaseNode *proto = static_cast<WbBaseNode *>(node)->getFirstFinalizedProtoInstance();
   if (!proto) {
     if (node->modelName() != node->nodeModelName())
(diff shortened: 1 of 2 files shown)


omichel:
Can you please fix the source tests which are failing to a wrong coding style?

galou:
When at it. Is the log message on failure of `getFirstFinalizedProtoInstance()` OK?

src/webots/nodes/utils/WbSupervisorUtilities.cpp
```diff
@@ -503,7 +503,13 @@ WbNode *WbSupervisorUtilities::getProtoParameterNodeInstance(int nodeId, const Q
     mRobot->warn(tr("%1: node not found.").arg(functionName));
     return NULL;
   }
-  return static_cast<WbBaseNode *>(node)->getFirstFinalizedProtoInstance();
+  WbBaseNode *proto = static_cast<WbBaseNode *>(node)->getFirstFinalizedProtoInstance();
+  if (!proto)
+  {
+    mRobot->warn(tr("Cannot get the PROTO instance for node `%1` (type %2).").arg(node->modelName(), node->nodeModelName()));
```
omichel, line 509: It would be better to use standard single quotes:
```suggestion
    mRobot->warn(tr("Cannot get the PROTO instance for node '%1' (type '%2').").arg(node->modelName(), node->nodeModelName()));
```

src/webots/nodes/utils/WbSupervisorUtilities.cpp
```diff
@@ -594,15 +600,21 @@ void WbSupervisorUtilities::handleMessage(QDataStream &stream) {
       stream >> nodeId;
       const QString &stateName = readString(stream);
       WbNode *const node = getProtoParameterNodeInstance(nodeId, "wb_supervisor_node_load_state()");
-      node->reset(stateName);
+      if (node)
+      {
+        node->reset(stateName);
+      }
```
omichel, line 606: That should partially fix the clang-format issue:
```suggestion
      if (node)
        node->reset(stateName);
```

omichel:
> When at it. Is the log message on failure of `getFirstFinalizedProtoInstance()` OK?

I am not sure what is the exact difference between modelName and nodeModelName, can you explain it?

galou:
According to the little knowledge I have about Webots, `modelName` is the name of the node and `nodeModelName` is its type. We can maybe remove `nodeModelName`.

omichel:
> According to the little knowledge I have about Webots, `modelName` is the name of the node and `nodeModelName` is its type. We can maybe remove `nodeModelName`.

Yes, I found the confirmation [here](https://github.com/cyberbotics/webots/blob/master/src/webots/vrml/WbNode.hpp#L122-L123). I will propose an improvement to the terminology you used.

src/webots/nodes/utils/WbSupervisorUtilities.cpp
```diff
@@ -503,7 +503,10 @@ WbNode *WbSupervisorUtilities::getProtoParameterNodeInstance(int nodeId, const Q
     mRobot->warn(tr("%1: node not found.").arg(functionName));
     return NULL;
   }
-  return static_cast<WbBaseNode *>(node)->getFirstFinalizedProtoInstance();
+  WbBaseNode *proto = static_cast<WbBaseNode *>(node)->getFirstFinalizedProtoInstance();
+  if (!proto)
+    mRobot->warn(tr("Cannot get the PROTO instance for node '%1' (type '%2').").arg(node->modelName(), node->nodeModelName()));
```
omichel, line 508:
```suggestion
  if (!proto) {
    if (node->modelName() != node->nodeModelName())
      mRobot->warn(tr("Cannot get the PROTO instance for node '%1' (derived from '%2').").arg(node->modelName(), node->nodeModelName()));
    else
      mRobot->warn(tr("Cannot get the PROTO instance for node '%1'.").arg(node->modelName()));
    
```

galou: What I wanted is actually `defName()`, so I changed the code according to this.

galou:
FYI, I probably found the cause of the issue that brought me to this PR: #6474 .

src/webots/nodes/utils/WbSupervisorUtilities.cpp
galou: Did it.

src/webots/nodes/utils/WbSupervisorUtilities.cpp
```diff
@@ -503,7 +503,15 @@ WbNode *WbSupervisorUtilities::getProtoParameterNodeInstance(int nodeId, const Q
     mRobot->warn(tr("%1: node not found.").arg(functionName));
     return NULL;
   }
-  return static_cast<WbBaseNode *>(node)->getFirstFinalizedProtoInstance();
+  WbBaseNode *proto = static_cast<WbBaseNode *>(node)->getFirstFinalizedProtoInstance();
+  if (!proto)
+  {
```
omichel, line 508:
```suggestion
  if (!proto) {
```

omichel (dismissed):

Issue #5181, opened 2022-09-07
Issue with Java on macOS M1


alexvena19:
trying to run a Webots program, and the output is 
Native code library failed to load. See the chapter on Dynamic Linking Problems in the SWIG Java documentation for help.
java.lang.UnsatisfiedLinkError: no JavaController in java.library.path: [/Applications/Webots.app/lib/controller/java]

I'm using a MacBook m1 pro, 64-bit version of java.

<img width="1026" alt="Screen Shot 2022-09-07 at 9 56 18 AM" src="https://user-images.githubusercontent.com/91227417/188896579-b1308f05-c11b-4633-9a67-af564880239b.png">

<img width="566" alt="Screen%20Shot%202022-09-07%20at%209 24 38%20AM copy" src="https://user-images.githubusercontent.com/91227417/188896295-013158b6-af83-4250-9f5d-468231680968.png">

omichel:
Which version of Webots are you using?

alexvena19:
> 

Webots R2022a.

alexvena19:
> Can you try with the latest nightly build of Webots R2022b?
> You can download it from https://github.com/cyberbotics/webots/releases/tag/nightly_5_9_2022
> Webots R2022b introduces native support for M1 macs.
> It could be that the Java version you are using is built for M1 while the libController of Webots R2022a doesn't support M1, thus the error.
> Using Webots R2022b with M1 support should fix the problem.
> Let me know if that works.

Sure I will try that. Which link of these do I download?
<img width="641" alt="Screen Shot 2022-09-07 at 10 04 17 AM" src="https://user-images.githubusercontent.com/91227417/188898557-b05f287d-4e13-410e-869b-a02a7613f34d.png">

omichel:
webots-R2022b.dmg

alexvena19:
> webots-R2022b.dmg

Still not working. I only downloaded the webots-R2022b.dmg

<img width="1100" alt="Screen Shot 2022-09-07 at 10 17 04 AM" src="https://user-images.githubusercontent.com/91227417/188901418-a39f4693-6319-4034-b917-e4d42fc258a1.png">

omichel:
> Still not working. I only downloaded the webots-R2022b.dmg

Did you remove Webots R2022a and installed R2022b instead?

alexvena19:
> > Still not working. I only downloaded the webots-R2022b.dmg
> 
> Did you remove Webots R2022a and installed R2022b instead?

[...]

PR #6501, opened 2024-02-14
Fix jnilibs flags

koenyskout:
**Description**
Ensure that libController.dylib can be found on MacOS by fixing the relative path in JNILIB_FLAGS.

**Related Issues**
This pull-request fixes issue #5181

**Tasks**
  - [X] Update the [changelog](https://github.com/cyberbotics/webots/blob/master/docs/reference/changelog-r2023.md)
  - [ ] Test whether this also fixes the issue when running with other JVM/JDKs


omichel (approved):
That looks great, thank you!

PR #6521, opened 2024-03-13
build with ros in Ubuntu 22.04/23.04 

lucasw:
**Description**

This builds and runs in Ubuntu 23.04, and probably also in 22.04 (it builds in CI).  It loses some functionality in the process which is why this is just a draft, maybe it can be useful for anyone else want to use webots and ros in 22.04 or later.

**Tasks**
  - [ ] Update the [changelog](https://github.com/cyberbotics/webots/blob/master/docs/reference/changelog-r2023.md)
  - [ ] Update the documentation (if needed)

**Additional context**

The packages available in Ubuntu from Debian with apt install ros-* (as opposed to install ros-noetic* in 20.04 with the build farm configured) are almost enough to make all of the webots ros code build, except controller manager and moveit aren't there so I've had to disable the code depending on those- a conditionally building them would be preferable of course.

There are a lot of commits trying to get the github action for 20.04 to build, once it works most of those could be squashed down to clean it up- the 22.04 ci does seem to be passing and is building the ros code (earlier it was skipping over it).

### DIFF
--- a/.github/workflows/test_suite_linux.yml
+++ b/.github/workflows/test_suite_linux.yml
@@ -6,9 +6,10 @@ on:
   schedule:
     - cron:  '0 23 * * *'
   push:
-    tags:
-      - nightly_*
-      - R20*
+  #  tags:
+  #    - nightly_*
+  #    - R20*
+  workflow_dispatch:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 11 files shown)


lukicdarkoo:
@lucasw Thank you very much for the effort, I really appreciate it.

We decided to:
- Remove support for 20.04 and add support for 24.04 (https://github.com/cyberbotics/webots/pull/6704)
- Remove support for ROS (1) (https://github.com/cyberbotics/webots/pull/6697)
- Provide an alternative for ROS (1) users (https://github.com/cyberbotics/webots_ros/pull/153)

I hope this makes sense for you. Maintaining ROS (1) support while ROS (1) doesn't support Ubuntu 22.04 and newer would be kinda hacky.

Therefore, I am closing this PR. If you think there is a better approach feel free to reopen the PR and continue development.

PR #6564, opened 2024-06-20
[Controller API (Java)] Fix Length of Arrays Returned by `CameraRecognitionObject.getColors()`

CoolSpy3:
**Description**
`CameraRecognitionObject.getColors()` is supposed to return an array of length `3*CameraRecognitionObject.getNumberOfColors()`. At the moment, the returned array is truncated to always have a length of `3`. Thus, only the first color can be retrieved. This PR updates the API so that an array of the correct length is returned.

**Tasks**


omichel (approved):
Thank you!

Issue #6573, opened 2024-07-02
Node Ancestry Should be Considered for Node Type Restrictions in PROTOs


CoolSpy3:
**Describe the Bug**
At the moment, Node fields in PROTOs are required to list every type which they can accept (if they are type restricted). However, this us true even if all of the necessary types share a common base type which would be sufficient. This leads to verbose type restrictions where only a single actual type is required, and is unintuitive given the inheritance-style structure of nodes.

**Steps to Reproduce**
Attempt to load the PROTOs:
``` proto
# Base.proto
#VRML_SIM R2023b utf8

PROTO Base [] { Pose {} }
```
``` proto
# ExtendsBase.proto
#VRML_SIM R2023b utf8

EXTERNPROTO "../protos/Base.proto"

PROTO ExtendsBase [] { Base {} }
```
``` proto
# Example.proto
#VRML_SIM R2023b utf8

EXTERNPROTO "../protos/Base.proto"
EXTERNPROTO "../protos/ExtendsBase.proto"

PROTO Example [
    unconnectedField MFNode{Base{},Solid{}} f [ExtendsBase{},LightSensor{}]
]
{ Pose {} }
```

When trying to load `Example.proto`, Webots will state that the default values of `f` are not permitted even though `ExtendsBase` *is* an instance of `Base`, and `LightSensor` *is* an instance of `Solid`.

**Expected behavior**
The entire ancestry of nodes should be considered when determining if they satisfy a node restriction. This would make the above code valid.

**System**
 - Operating System: Windows 10
 - Graphics Card: NVIDIA GeForce 1050 Ti

omichel:
That sounds like a good idea. Please go ahead with the implementation and open a new PR to propose your contribution. We will review it and likely accept it.

CoolSpy3:
Fixed in R2025a

PR #6574, opened 2024-07-03
Enhancement: Consider Node Ancestry in Proto Type Restrictions

CoolSpy3:
**Description**
Currently if a `SFNode` or `MFNode` field is value-restricted, any nodes added to the field, must *exactly* match the restricted type (or in the case of proto-nodes, directly derive from it). This PR updates the behavior so that the entire ancestry of the nodes are searched. (ex. A field restricted to `Pose` nodes can now also contain `Solid` nodes) See the linked issue for more information.

A couple of notes about decisions made in this PR:
* Abstract node classes cannot be represented in the current VRML implementation. Thus, where they occur in the ancestry tree, they have been omitted or jumped-over. Allowing them would require a larger rewrite of the parsing code, which I do not feel is necessary at this time.
  * As a side effect of this choice, there are no cases of multiple-inheritance. I originally planed to implement this to deal with cases such as `SolidDevice`, but rolled it back in favor of #CS41. This means that some of the code will have to be reworked if multiple-inheritance is desired in the future, however, the changes are fairly straight-forward, so I don't have any concerns.
* In order to implement this PR, I had to add a method to determine which nodes defined by Webots extended other nodes. I chose to do this with an additional comment in the `wrl` files. In my mind this makes the most sense because I would expect this information to be contained with the `WbNodeModel` object, which is generated from these files. The only other ways I could think of doing this would be either writing out all of the relationships in a file (which would be more verbose and detached from the actual node definitions) or keeping track of constructor invocations at runtime (which would need to be done on every node instance creation rather than on every model instance creation and would add additional c++ code to convey information which is already expressed in the inheritance structure of the classes. The c++ heirarchy also includes the abstract nodes mentioned above. It would not make sense for these to be included, which is unclear from the c++ code, but is clearer when viewed in the `wrl` context).
* Many of the protos in this project could now be simplified with these changes in mind. I did not do that here because 1) the new behavior is backward compatible in 99% of cases (i.e. definitions which are now *technically* overexplicit will still function without errors), and 2) it may be helpful, especially in places such as the reference docs, for new developers to see multiple examples of nodes that can be used in a specific field even if a more generic definition would suffice.
* There is currently [an edge case which causes `Transform` nodes to be converted to a `Pose` during loading](https://github.com/cyberbotics/webots/blob/608109ca88c8554f48ed79b6996bb20033772607/src/webots/vrml/WbNodeReader.cpp#L53-L55). This could disrupt this system in the case where someone restricts a field to a `Transform`. There are a couple solutions (short of removing the linked code), which all have drawbacks. If I was to fix this, I would do it by interpreting a `Transform` restriction as a `Pose` restriction. However, this would add a bit of extra clutter to the code, and given that 1) I doubt many people are restricting nodes to `Transform`, and 2) this behavior would've already been present in previous versions, I've opted to just ignore it.

If you feel that any of these decisions should be reconsidered, let me know, and I can update the appropriate code.

**Related Issues**
This pull-request fixes issue #6573.

**Tasks**
  - [x] Add a method for querying the node hierarchy at runtime
  - [x] Update type checking code to search the ancestry tree
  - [x] Update the test suite
  - [x] Update the documentation

**Documentation**
[PROTO Definition#field-value-restriction](https://cyberbotics.com/doc/reference/proto-definition?version=DeepBlueRobotics:enhancement-consider-node-ancestry-in-proto-type-restrictions#field-value-restriction)


CoolSpy3:
Note: The test suite is going to fail due to a bug in the `display` test. This is unrelated to this PR and also occurs on `develop`. The important thing is that that parser tests pass.

tests/manual_tests/protos/DerivedApperanceProto.prot
```diff
@@ -0,0 +1,12 @@
+#VRML_SIM R2024a utf8
```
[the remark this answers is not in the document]
CoolSpy3: Ahh, good catch! Fixed!

I incorrectly assumed the test suite would fail on unexpected errors. I've added an additional test case to test the correct behavior.

docs/reference/proto-definition.md
```diff
@@ -77,11 +77,11 @@ PROTO MyProto [
   field SFString                            name          "my proto"
   field SFColor{0 0 0, 0.5 0.5 0.5, 1 1 1}  color         0.5 0.5 0.5
   field SFNode                              physics       NULL
-  field MFNode{Solid{}, Pose{}}             extensionSlot []
+  field MFNode{Solid{}, Transform{}}        extensionSlot []
 ]
 ```
 
-In this example, the `color` field value can only be `0 0 0`, `0.5 0.5 0.5` or `1 1 1` and the `extensionSlot` field can only accept [Solid](../reference/solid.md) and [Pose](../reference/pose.md) nodes.
+In this example, the `color` field value can only be `0 0 0`, `0.5 0.5 0.5` or `1 1 1` and the `extensionSlot` field can only accept [Solid](../reference/solid.md), [Transform](../reference/transform.md) nodes, and nodes that inherit from them.
```
omichel, line 84: This doesn't make sense any more, since `Solid` inherits from `Transform`, so we should write instead:
```js
  field MFNode{Transform{}}                    extensionSlot []
```
Also, I see a more fundamental problem here. In some case we may want to allow a `Transform` or a `Pose`, but not a Solid. The new system doesn't allow this any more, which may be problematic in some cases, like I want to have an optional shape on the top of my proto, which has no bounding object and no physics, thus I need a `Pose`, but not a `Solid`. How could the new system handle this?

CoolSpy3: `Solid` inherits from `Pose` not `Transform`.

In the current model, the desired behavior could be softly enforced by declaring a `Pose` restriction and adding 
``` javascript
%<
// For SFNodes
assert(!(fields.field.boundingBox || fields.field.physics), "field must have no boundingObject or physics!");
// For MFNodes
assert(fields.field.every(node => !(node.boundingBox || node.physics)), "field must have no boundingObject or physics!");
>%
```
to the proto definition.

Alternatively, if you'd rather have the behavior to be supported by the proto syntax, we could define something like
``` js
field SFNode{=Pose{},=Transform{}} shape Pose{}
```
to restore the original behavior.

omichel: > `Solid` inherits from `Pose` not `Transform`.

Oops. Yes, you are right. So the example fully makes sense. Maybe we could also explain that in this specific case we might have simply used `Pose` to encompass both `Transform` and `Solid`. However `Pose` would also allow for `Fluid` which may not be desirable.

> In the current model, the desired behavior could be softly enforced by declaring a `Pose` restriction and adding
> 
> ```js
> %<
> // For SFNodes
> assert(!(fields.field.boundingBox || fields.field.physics), "field must have no boundingObject or physics!");
> // For MFNodes
> assert(fields.field.every(node => !(node.boundingBox || node.physics)), "field must have no boundingObject or physics!");
> >%
> ```
> 
> to the proto definition.

Yes, that's indeed possible, although not very convenient.

> Alternatively, if you'd rather have the behavior to be supported by the proto syntax, we could define something like
> 
> ```js
> field SFNode{=Pose{},=Transform{}} shape Pose{}
> ```
> 
> to restore the original behavior.

Yes, my only remaining concern with this PR is that it breaks the backwards compatibility.

```js
field SFNode{=Pose{},=Transform{}} shape Pose{}
```
in the new version would be equivalent to:
```js
field SFNode{Pose{},Transform{}} shape Pose{}
```
in the old version.

If we want to maintain backwards compatibility, while still introducing your new feature, we could a syntax like:
```js
field SFNode{Solid*{},Transform*{}} shape Pose{}
```
to mean that and `Solid` or any descendant of `Solid` and `Transform` or any descendant of `Transform` is allowed. Though, I am not sure the `*` syntax is the best one (it doesn't look very nice to my eyes). Honestly, I prefer the syntax you propose with the `=` sign even if that breaks the backwards compatibility, because I believe that such a break will only have limited consequences. So, it you go ahead with the proposed `=` syntax, I am OK with that.

CoolSpy3:
Alright, I've updated the code to implement the new parsing behavior. I also updated a bunch of other instances where field values are validated (that I missed before) to point to the new implementation. I still have to test the `=` syntax manually (and then write automatic tests) and update the docs. I've updated the task list to reflect this, and I'll mark this PR as a draft until that gets completed.

docs/reference/proto-definition.md
CoolSpy3: In the latest batch of commits, I added the `=` case, but made the parser revert to the old behavior if the proto version was <R2024a. However, while typing this out, I tried out your suggestion with `+` rather than `*`, and it's starting to grow on me.
``` javascript
field SFNode{Solid+{},Transform+{}} shape Solid{}
```
(explicitly allows descendants)
vs
``` javascript
field SFNode{=Solid{},=Transform{}} shape Solid{}
```
(explicitly disallows descendants)

Ultimately, it might come down to a decision of which default behavior makes more sense. From an object-oriented/inheritance perspective, the `=` syntax makes more sense because, on some level, a descendant of a `Solid` *IS* a `Solid`. However, given that Webots seems to treat inheritance more as "distinct nodes that happen to share some similar features," (i.e. a decendant of a `Solid` just sometimes *LOOKS* like a `Solid`) the `+` syntax might make more sense for this PR.

I've been staring at this too long, so I think I'll sleep on it, but lmk if you have any thoughts.

omichel:
That looks good to me. I do have a preference for the `+` syntax which is backwards compatible as the old syntax would still work as before and the `+` syntax would add more possibilities, e.g., allowing also derived nodes.

CoolSpy3:
Alright, I think this is ready for review again! It looks like the type name and the following `{}` are parsed collectively as a node, so in order to not have a weird edge case in the node parsing code, I switched to the syntax to
```javascript
field SFNode{Solid{}+,Transform{}+} shape Solid{}
```
Notes:
* When testing locally, you have to delete `resources/proto-list.xml` before your first build in order to regenerate it. (I modified the generation script to include some extra info.)
* While debugging I discovered a bug in some of the modified functions which prevents a node from being added to an `unconnectedField` from the GUI. This is not a regression, so I do not believe it has to be addressed in this PR. If you agree, I can migrate it to an issue,

CoolSpy3:
Tests are fixed!

src/webots/vrml/WbProtoModel.cpp
```diff
@@ -417,6 +417,15 @@ WbNode *WbProtoModel::generateRoot(const QVector<WbField *> &parameters, const Q
   return root;
 }
 
+QStringList WbProtoModel::parentProtoNames() const {
+  QStringList parents;
+  const WbProtoModel *parentProtoModel = this;
+  while ((parentProtoModel = parentProtoModel->ancestorProtoModel())) {
+    parents << parentProtoModel->name();
+  }
```
omichel, line 425: Minor styling:
```suggestion
  while ((parentProtoModel = parentProtoModel->ancestorProtoModel()))
    parents << parentProtoModel->name();
```

tests/manual_tests/protos/DerivedPoseProto.proto
```diff
@@ -0,0 +1,14 @@
...
+EXTERNPROTO "webots://tests/manual_tests/protos/PoseProto.proto"
+
+PROTO DerivedPoseProto [
+  field SFVec3f translation 0 0 0
+  field SFRotation rotation 0 0 1 0
+]
+{
+  PoseProto {
+	translation IS translation
+	rotation IS rotation
```
omichel, line 12: Use spaces instead of tabs:
```suggestion
    translation IS translation
    rotation IS rotation
```

omichel (changes requested):
Very good job! Thank you.
I would only request a couple of minor changes.

omichel (approved):

PR #6580, opened 2024-07-10
Update MacOS Runner

CoolSpy3:
**Description**
GitHub recently deprecated the `macos-11` runner, so all of the tests which rely on that os are failing. This PR updates the runner to the ~~latest `macos-14`~~ `macos-13` image. **(There's some bug installing cppcheck on macos-14. I might play around with bumping the cppcheck version later, but I don't have enough info to resolve it as-is.)**

Hopefully, most of the workflow can be ported without too much hassle, however, if that proves to not be the case, I may have to leave this fix for someone else, as I do not have a mac to debug on.


CoolSpy3:
The build passes! I'll come back tomorrow and see if I can get cppcheck working. If not, I'll just revert to this commit.

SIde note: Even though the build runs on macos 13, it still targets macos 11. I assume that we want to keep that as-is because we do not want to drop old support unless it becomes necessary.

omichel:
No, the plan is to support only the maintained versions macOS, that are currently 12 (Monterey), 13 (Ventura) and 14 (Sonoma). So we should drop macOS 11 Big Sur and update the [requirements documentation](https://cyberbotics.com/doc/guide/system-requirements) accordingly.

CoolSpy3:
Update: I've gotten the build to work on macos 14. To do this, I had to bump cppcheck. It works on cppcheck 2.11, but I figured "Why not just update to the latest version, so that we're as up-to-date as we can be?"
The good news: The new versions of cppcheck (2.11+) catch more errors than the older versions.
The bad news: This means that a bunch of new warnings now exist in the codebase.

I could suppress them all globally, but they are (mostly) style ideas which I think *should* be enforced. I've taken a crack at fixing some of them (the most recent commit reverts those changes). However, there are a lot of warnings, so it would probably take me a while to get through them all.

A full list of the warning types (the full warning lists can be viewed in the action logs):
```
[normalCheckLevelMaxBranches]
[duplInheritedMember]
[constParameterCallback]
[passedByValue]
[constVariablePointer]
[variableScope]
[constParameterPointer]
[constParameter]
[cstyleCast]
[knownConditionTrueFalse]
[constVariable]
[returnByReference]
[missingOverride]
[knownPointerToBool]
[truncLongCastReturn]
[iterateByValue]
[ignoredReturnErrorCode]
[unreadVariable]
[uselessOverride]
[constParameterReference]
[truncLongCastAssignment]
[redundantCopyLocalConst]
[functionConst]
[unassignedVariable]
[unusedPrivateFunction]
[incorrectStringBooleanError]
```
We could also consider only upgrading cppcheck to 2.11. That gives the warnings:
```
[constVariablePointer]
[constParameter]
[constParameterPointer]
[cstyleCast]
[knownConditionTrueFalse]
[constVariable]
[functionConst]
[passedByValue]
[variableScope]
[bitwiseOnBoolean]
```
Finally, we could leave the cppcheck version on 2.10 and only upgrade the mac runner to version 14, but in that case, this same problem will come up when GitHub decides to deprecate the macOS 13 runner.

Any thoughts on how to proceed?

omichel:
I believe the best long term solution is to upgrade to the latest version of macOS and cppcheck and fix all the warnings, even if that takes some time.

However I do not have time to do this myself. So if you don't have time either, any solution that would allow us to fix this test on macOS is welcome, even if that is a temporary solution.

CoolSpy3:
Alright! The number of warnings is large, but not insurmountable. I'll work on it on and off for a few days and see what kind of progress I'm making. If it turns out that I'm not making much of a dent, I can temporarily ignore the warnings and open an issue to fix them later.

CoolSpy3:
We've done it! This PR is ready for review!
**A summary of the changes:**
* Updated the macos runner to version 14
  * Once this PR is merged, I can migrate this change to the [wiki](https://github.com/cyberbotics/webots/wiki/CI-Test-Labels/)
* Removed support for macos 11
* Bumped cppcheck to version 2.14.1
* Updated the [system requirements](https://cyberbotics.com/doc/guide/system-requirements?version=DeepBlueRobotics:update-macos-runner) to reflect the above
* Updated the [changelog](https://github.com/DeepBlueRobotics/webots/blob/update-macos-runner/docs/reference/changelog-r2024.md)

**Notes on reviewing/merging:**
* It may be helpful to review [just the config changes](https://github.com/cyberbotics/webots/pull/6580/files?file-filters%5B%5D=.include&file-filters%5B%5D=.md&file-filters%5B%5D=.py&file-filters%5B%5D=.yml&file-filters%5B%5D=No+extension&show-viewed-files=true#) ([also `Serial.cpp`](https://github.com/cyberbotics/webots/pull/6580/files#diff-8d3d8bf406256b7723951184603e8d6b6688a75008c1301f75d2a3c8f01d4217)) and [just the warning fixes](https://github.com/cyberbotics/webots/pull/6580/files?file-filters%5B%5D=.c&file-filters%5B%5D=.cpp&file-filters%5B%5D=.h&file-filters%5B%5D=.hpp&show-viewed-files=true) individually. Luckily, the warning fixes are all very small; there's just a lot of them.
  * You can also review them commit-by-commit, which could make it easier to review some smaller changes separate from the ones which just add `const` in front of everything.
* In order to pass the merge requirments, the required macos-11 "Test Sources" check should be replaced with the new macos-14 variant.

omichel (approved):
Very good job! Thank you!
I could review all the file changes as per your instructions and everything looks very good to me.
I like your approach of adding some `cppcheck-suppress` comments instead of trying to fix the problem when it was not super obvious to fix as it minimizes the risk of introducing some bugs, while still highlighting possible improvements.
I just replaced the mandatory macos-11 "Test Sources" check with the new macos-14 one.
I just have one minor suggestion, but you can ignore it if you prefer.

omichel (approved):

PR #6584, opened 2024-07-25
Fix `Brake`s added to the second axis of a `Hinge2Joint` being applied to the non-transformed axis.

brettle:
**Description**

Fixes bug where`Brake`s added to the second axis of a `Hinge2Joint` were being applied to the non-transformed axis.

**Tasks**

**Additional context**

I'll upload a test world that demonstrates the issue.

### DIFF
--- a/docs/reference/changelog-r2024.md
+++ b/docs/reference/changelog-r2024.md
@@ -14,3 +14,4 @@ Released on December **th, 2023.
     - Fixed length of arrays returned by `getPose()` in Java ([#6556](https://github.com/cyberbotics/webots/pull/6556)).
     - Fixed length of arrays returned by `CameraRecognitionObject.getColors()` in Java ([#6564](https://github.com/cyberbotics/webots/pull/6564)).
     - Fixed handling of device objects with the same name in the controller API ([#6579](https://github.com/cyberbotics/webots/pull/6579)).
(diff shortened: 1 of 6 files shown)


brettle:
[Hinge2Test.zip](https://github.com/user-attachments/files/16384359/Hinge2Test.zip) is a test world that demonstrates the issue. Each of the 2 spinning boxes is connected to it's own Hinge2Joint and both joints have the same configuration except that the second one has the first hinge rotated by 90 deg. As a result both of the spinning boxes *should* be spinning at the same speed throughout the test, but they do not. The second box accelerates/decelerates as though the brake's damping constant was not applied.

brettle:
Converted to draft because although the fix solves the problem for the test case, it does not solve the problem if the first hinge rotates *after* the world loads.

brettle:
My latest commit seems to address the problem when hinge rotates after loading the world.

brettle:
Here's an updated [Hinge2Test.zip](https://github.com/user-attachments/files/16395877/Hinge2Test.zip) which tests the rotate-after-load case.

omichel:
It would be nice, if possible, to integrate this test in the test suite which runs in the CI.

brettle:
> It would be nice, if possible, to integrate this test in the test suite which runs in the CI.

Done.

omichel (approved):
Excellent! Thank you.

Issue #6571, opened 2024-07-01
Importing PROTOs from non-official web sources


gabryelreyes:
## **Describe the Bug**
When using `EXTERNPROTO` to import a PROTO file such as [this one](https://github.com/BlueAndi/RadonUlzer/blob/main/webots/protos/Zumo32U4.proto) into a world, the following message is printed to the console:

```log
ERROR: The cascaded URL inferring mechanism is supported only for official Webots assets.
ERROR: 'https://raw.githubusercontent.com/BlueAndi/RadonUlzer/main/webots/protos/Zumo32U4.proto':115:22: error: Skipped unknown 'Rubber' node or PROTO.
```

In this case, a non-official (custom) PROTO is being loaded, which also has its own dependencies (in this case, the official Rubber PROTO, but it may as well be other custom PROTO).

I traced the log message to [WbProtoManager.cpp](https://github.com/cyberbotics/webots/blob/c2ea36ba57d073cf3f3c6c94d947cd47e6457ffc/src/webots/vrml/WbProtoManager.cpp#L191) which determines that the URL is not from the official repository by using a RegEx expression from [WbUrl.cpp](https://github.com/cyberbotics/webots/blob/master/src/webots/vrml/WbUrl.cpp#L240).

Could you please shed some light onto the reason behind only supporting official Webots assets in this feature?

## **Additional context**
We are working on the [RadonUlzer](https://github.com/BlueAndi/RadonUlzer) project, which now has evolved to support both real hardware and simulated hardware using the same application code. This is achieved by separation of the HALs into different repositories. For example, for Webots there is the [ZumoHALWebots](https://github.com/BlueAndi/ZumoHALWebots). This HAL is dependent on the [Zumo32U4 PROTO](https://github.com/BlueAndi/RadonUlzer/blob/main/webots/protos/Zumo32U4.proto), which currently is found in another repository together with the `wbt` files. 

The main concern is mismatching configurations between the PROTO and the corresponding HAL and drivers. To prevent this the optimal option is to get the PROTO file directly from the GitHub repository based on the HAL that is being processed by the build system. We would also like to avoid manually copying the PROTO file each time from the HAL Repository to the Worlds folder for the same risk of outdated/mismatching configurations after an update.

## **Steps to Reproduce**
1. Create a world.
2. Add line `EXTERNPROTO "https://raw.githubusercontent.com/BlueAndi/RadonUlzer/main/webots/protos/Zumo32U4.proto"`
3. Open wbt file.
[...]

PR #6585, opened 2024-07-26
Remote assets per URL

gabryelreyes:
**Description**
Remote assets do no require any inferring mechanisms from the parent files. This is only valid for local files. Remote asset was being treated as local.

**Related Issues**
This pull-request fixes issue #6571 

**Tasks**
  - [x] Fix handling of remote assets from unofficial sources.

**Additional Information**
`WbNetwork::instance()->isCachedWithMapUpdate(modelPath)`  in line 174 did nothing, as `modelPath` was empty. This was changed to check `WbNetwork::instance()->isCachedWithMapUpdate(protoDeclaration)`, which makes more sense here.


omichel (approved):

PR #6587, opened 2024-07-27
Remove references to unused custom uniforms.

brettle:
**Description**
Some graphics cards/drivers do not allow uniforms that are unused because they are optimized away by the GLSL compiler. The uniforms "subCamerasResolutionX" and "subCamerasResolutionY" were present in merge_spherical.frag but were never used. As a result, when trying to call to glGetUniformLocation() would return -1 for those names, crashing webots.

**Tasks**


omichel (dismissed):
Thank you!

omichel (approved):

PR #6595, opened 2024-07-30
Fix macos CI being slow.

brettle:
**Description**

Treat the "Null Audio Device" returned by Apple's OpenAL during CI the same as NULL so we don't attempt to open it. Attempting to open it was causing the code to hang for 6.5 minutes.

### DIFF
--- a/src/webots/sound/WbSoundEngine.cpp
+++ b/src/webots/sound/WbSoundEngine.cpp
@@ -49,6 +49,7 @@
 #include <ode/ode.h>
 
 #include <cassert>
+#include <cstring>
 
 static WbWorld *gWorld = NULL;
 static bool gOpenAL = false;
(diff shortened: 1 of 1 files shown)


brettle:
Note: The test failures appear to be unrelated to the change. They were already occurring.

src/webots/sound/WbSoundEngine.cpp
```diff
@@ -93,7 +94,7 @@ static void init() {
   WbLog::toggle(stderr);  // we want to disable stderr to avoid warnings in the console
   try {
     const ALCchar *defaultDeviceName = alcGetString(NULL, ALC_DEFAULT_DEVICE_SPECIFIER);
-    if (defaultDeviceName == NULL)
+    if (defaultDeviceName == NULL || !strcmp(defaultDeviceName, "Null Audio Device"))
```
omichel, line 97: We usually prefer this variant which is arguably more understandable:
```suggestion
    if (defaultDeviceName == NULL || strcmp(defaultDeviceName, "Null Audio Device") == 0)
```

omichel (approved):
Thank you.

omichel (approved):

PR #6606, opened 2024-08-01
Fixed bug radar.py

lonely-poppy:
**Description**
Fixed the bug that when using Python to write a controller, using getTargets() could not correctly obtain multiple target data detected by radar


omichel (changes requested):
Thank you!
Could you please also add a line in change log?

omichel (approved):

PR #6608, opened 2024-08-01
Fix slow source testing

brettle:
**Description**

Build optimized/release `cppcheck` so tests run *much* faster.


omichel (approved):
Thank you.

PR #6617, opened 2024-08-08
Fix incomplete loading while minimized under some windowing systems.

brettle:
**Description**

Under some window systems (notably linux wayland), `QWindow::isExposed()` always returns false if the window is minimized (e.g. when running tests). This was preventing the initial non-culling call to `renderNow()` from accomplishing its goal of ensuring that meshes/textures are loaded onto the GPU. (This resulted in the `skin` test failing because the aabb for the skin was not calculated using the skeleton.) This fixes that by allowing the non-culling call to proceed even if the window is not exposed. It also suppresses the resulting errors/warnings *only* for that call because several of the parsing tests fail if the stderr output is not empty.

**Tasks**


src/webots/gui/WbView3D.cpp
```diff
@@ -1111,8 +1112,19 @@ void WbView3D::setWorld(WbSimulationWorld *w) {
 
   WbWrenOpenGlContext::doneWren();
 
-  // first rendering without culling to make sure every meshes/textures are actually loaded on the GPU
+  // first rendering without culling to make sure every meshes/textures are actually loaded on the
+  // GPU If the window is not yet exposed (e.g. minimized under Wayland) this can produce some GL
```
omichel, line 1116:
```suggestion
  // GPU if the window is not yet exposed (e.g. minimized under Wayland) this can produce some GL
```

docs/reference/changelog-r2024.md
```diff
@@ -19,3 +19,4 @@ Released on December **th, 2023.
     - Fixed invalid absolute sound file path resulted in crash ([#6593](https://github.com/cyberbotics/webots/pull/6593))
     - Fixed Speaker relative sound file path not working with external controller ([#6605](https://github.com/cyberbotics/webots/pull/6605)).
     - Fixed the bug that when the language is Python, getTargets() cannot correctly obtain the multi-target data detected by the radar ([#6606](https://github.com/cyberbotics/webots/pull/6606))
+    - Fixed incomplete loading while minimized under some windowing systems ([#6616](https://github.com/cyberbotics/webots/pull/6616)).
```
omichel, line 22:
```suggestion
    - Fixed incomplete loading while minimized under some windowing systems ([#6617](https://github.com/cyberbotics/webots/pull/6617)).
```

src/webots/gui/WbView3D.cpp
```diff
@@ -1111,8 +1112,19 @@ void WbView3D::setWorld(WbSimulationWorld *w) {
...
+  // errors and Qt warnings, so we suppress them just for this one call.
+  bool exposed = isExposed();
```
omichel, line 1118:
```suggestion
  const bool exposed = isExposed();
```

omichel (changes requested):
I am afraid that disabling all GL errors in the CI tests (under wayland) might open the door to some tests passing whereas they should actually fail. Why not disabling minimization in the CI tests?

brettle:
> I am afraid that disabling all GL errors in the CI tests (under wayland) might open the door to some tests passing whereas they should actually fail. Why not disabling minimization in the CI tests?

It's useful to run the CI tests locally while doing other work simultaneously, so I'd really rather not disable minimization. Moreover, if the tests (other than the "with-rendering" tests) fail while minimized, that is arguably a bug.

Note that the suppression is only for the initial post-load rendering pass. Any test which does more than one timestep will do at least one rendering pass without suppression and if the test requires no error output it will still fail.

If that doesn't address you concern, would it be addressed by only doing the suppression if some env var is set and setting that env var in `test_suite.py`? If so, let me know if you have a preferred env var name.

brettle:
I could also *not* set the env var in test_suite.py and instead output a message mentioning it when we detect the situation where it might be useful. Something along the lines of "WARNING: Webots will print errors/warnings because your windowing system does not support rendering to minimized windows. To suppress those, set WEBOTS_QUIET_MINIMIZED_RENDERING=true in your environment."

brettle:
Thinking about this some more, I'm wondering if instead of this PR, we should just make the initial non-culling render target a FBO instead of the window's framebuffer. I *think* that would succeed without errors/warnings even if the window was not exposed. Thoughts?

omichel:
> Thinking about this some more, I'm wondering if instead of this PR, we should just make the initial non-culling render target a FBO instead of the window's framebuffer. I _think_ that would succeed without errors/warnings even if the window was not exposed. Thoughts?

Yes, if that works, it would be ideal.

brettle:
> I am afraid that disabling all GL errors in the CI tests (under wayland) might open the door to some tests passing whereas they should actually fail. Why not disabling minimization in the CI tests?

Latest commit no longer disables error reporting.

brettle:
> > Thinking about this some more, I'm wondering if instead of this PR, we should just make the initial non-culling render target a FBO instead of the window's framebuffer. I _think_ that would succeed without errors/warnings even if the window was not exposed. Thoughts?
> 
> Yes, if that works, it would be ideal.

Afaict, it was already rendering to an FBO, so I mostly just needed to ensure that it would still render if not exposed. The only things that were triggering errors were blitting of the render to the screen, swapping buffers, and clearing the default framebuffer during postprocessing. The first 2 were easy to disable for the initial render. The clearing of the default framebuffer during postprocessing doesn't *appear* to be needed (and seems like it could cause other problems) so this removes it.

omichel (approved):
That looks much better this way to me.
Thank you.

PR #6618, opened 2024-08-11
Fix unitialized sliding friction when using asymmetric rolling friction.

brettle:
**Description**

Previously `mu2` was *not* initialized when using asymmetric rolling friction, but ODE will use asymmetric sliding friction if we set `dContactAxisDep` (because `dContactAxisDep  == dContactMu2`), so we need to make sure `mu2` is set. The uninitialized `mu2` was resulting in the rolling_friction test failing under some circumstances.

This PR also updates the expected values in test to reflect the result of the change.

**Tasks**


omichel (approved):
Well spotted!
Thank you.

PR #6622, opened 2024-08-14
Make supervisor_export_image use FBO instead of screen framebuffer.

brettle:
**Description**

This makes the `supervisor_export_image` test pass under Wayland and has the added benefit of making that test (and the billboard test) work with "--no-rendering --minimize". As a result those tests are moved into the api group and the with_rendering group is removed.

**Tasks**


omichel (approved):
Very nice.
Thank you.

PR #6631, opened 2024-08-19
Fix-test-suite-not-run-on-macos

brettle:
**Description**

Makes the test suite run successfully on macos-13.

- Switched from macos-14/macos-latest to macos-13 because the macos-14 runner crashes during the build, presumably due to the fact that is has half as much memory.
- Increased various timeouts/delays to accomodate the fact that the macos runner is *much* slower than the other runners. Let me know if you want those increases to be restricted to just the macos CI environment.
- Skipped the billboard test during macos CI because it currently fails there but passes on my Mac laptop. Presumably the issue is some combination of the macos runner being headless (and thus only supporting OpenGL 2.1) and the fact that it is run minimized and without rendering. Fwiw that test also fails for me on my local Fedora 40 installation when using x11, but not when using wayland or when running without --minimize.
- Various smaller fixes to allow the tests to run and make it easier to diagnose errors.


tests/default/controllers/test_suite_supervisor/test_suite_supervisor.py
```diff
@@ -240,6 +240,7 @@ def run(self):
 
         # 30 seconds before executing the next world, 60 seconds for the robot_window_html test
         delay = 60 if self.currentSimulationFilename.endswith('/robot_window_html.wbt') else 30
+        delay = delay * 10
```
CoolSpy3, line 243: There's no need to increase the delay during the linux tests, and doing so could make it take longer to find and debug test failures.
```suggestion
        if sys.platform == "darwin":
            # Longer timeout on MacOS because Webots takes longer to start there during CI.
            delay = delay * 10
```

brettle:
The macos build and test takes ~1hr, much longer than Ubuntu. Do we want to only run it when some additional or other label is present? @omichel?

omichel:
> The latest macos build and test-suite runs took nearly 2.5 hours, much longer than Ubuntu. Do we want to only run it when some additional or other label is present? @omichel?

Yes, that seems reasonable. Which label would fire it?

tests/default/controllers/test_suite_supervisor/test_suite_supervisor.py
```diff
@@ -240,6 +240,9 @@ def run(self):
 
         # 30 seconds before executing the next world, 60 seconds for the robot_window_html test
         delay = 60 if self.currentSimulationFilename.endswith('/robot_window_html.wbt') else 30
+        if sys.platform == "darwin":
+            # Longer timeout on MacOS because Webots takes longer to start there during CI.
+            delay = delay * 10
```
omichel, line 245:
```suggestion
            delay *= 10
```

omichel (approved):
That looks good to me.
Thanks.

brettle:
> > The latest macos build and test-suite runs took nearly 2.5 hours, much longer than Ubuntu. Do we want to only run it when some additional or other label is present? @omichel?
> 
> Yes, that seems reasonable. Which label would fire it?

I just added a "test suite mac" label and made the job require it. If that works for you, please reapprove.

omichel (approved):

PR #6635, opened 2024-08-20
Mostly MacOS-related fixes

brettle:
**Description**

These are changes I needed to run make and test_suite.py on my MacOS laptop as a regular user without it crashing.

At least one test (`supervisor_export_image`) still fails for me on MacOS but at least the test suite now runs without crashing.

Notable user-visible changes:

 - Fixed connection errors when using long robot names in some environments (issue #5452). This is not MacOS specific but is more likely to occur on a Mac because it's max socket path length is shorter than on Linux.
 - Fixed crash on MacOS when quitting under some circumstances. One reproducible case was quitting after running `check_saved_hidden_parameters.wbt`.
 - Fixed error when putting displays and cameras in separate windows on MacOS (e.g. with the `mirror.wbt` test). This was caused by using different formats for the different windows.

There are also some minor changes that are not user-visible:

- Use `(mach_port_t)NULL` because it is equivalent to kIOMainPortDefault or kIOMasterPortDefault but works across SDK versions.
- Run `aqt` in a python virtual env.
- Make `test_suite.py` use a new socket for each retry.


omichel (approved):
Thank you.
This PR is small, so it is OK to have several unrelated fixes in it, but in general we prefer to have one fix per PR. That is easier to review and useful to track down bugs afterwards.

PR #6638, opened 2024-08-26
Fix bugs (introduced in PR #6622) when grabbing/recording with devicePixelRatio != 1.

brettle:
**Description**

PR #6622 causes grabs to come from an offscreen fbo instead of the default framebuffer. The latter can be scaled (e.g. on a MacOS Retina display) and the code was still compensating for that, resulting in screen grabs where the screen's image was only in the top left quadrant of the resulting screenshot. There was a similar problem with video recording. This PR addresses both.


omichel (approved):
Thank you.

PR #6640, opened 2024-08-26
Fix-to-include-system-openal

brettle:
**Description**
Don't download openal lib from cyberbotics.com. Use the system's lib instead.


omichel (approved):
Thank you.

PR #6644, opened 2024-08-27
Fix-mirror-crash-under-wayland

brettle:
**Description**
Fixes bug introduced by PR #6635 which forced the format of the *context* for a rendering device window to be the same as the format of the main opengl context.  This resulted in the context not being usable with  rendering device windows created under wayland. This PR fixes that by requesting that the format of the *window" to be the same as the main opengl context.


omichel (approved):
Thank you.

PR #6623, opened 2024-08-15
Bump the Qt version from 6.4.3 to 6.5.3.

brettle:
**Description**
6.4.3 is no longer supported, while 6.5.3 is the latest LTS release and is supported through 31 Mar 2026.

For this to work, the precompiled binaries at https://cyberbotics.com/files/repository/dependencies/ will need to be updated but I don't think I have the ability to do that.

**Tasks**
Add the list of tasks of this PR.
  - [X] Update the [changelog](https://github.com/cyberbotics/webots/blob/master/docs/reference/changelog-r2024.md)

PR #6645, opened 2024-08-27
Fix failure of supervisor_start_stop_movie introduced by PR #6623.

brettle:
If the system ffmpeg links (possibly indirectly) to a different openssl than what we bundle, ffmpeg will fail to start. This fixes that by restoring the original LD_LIBRARY_PATH while running ffmpeg.

### DIFF
--- a/src/webots/gui/WbVideoRecorder.cpp
+++ b/src/webots/gui/WbVideoRecorder.cpp
@@ -459,7 +459,7 @@ QString WbVideoRecorder::nextFileName() {
 
 void WbVideoRecorder::createMpeg() {
 #ifdef __linux__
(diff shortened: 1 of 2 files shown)


omichel (commented):
We try to avoid abbreviations. I would rename `WEBOTS_ORIG_LD_LIBRARY_PATH` to `WEBOTS_ORIGINAL_LD_LIBRARY_PATH`

brettle:
> We try to avoid abbreviations. I would rename `WEBOTS_ORIG_LD_LIBRARY_PATH` to `WEBOTS_ORIGINAL_LD_LIBRARY_PATH`

Fixed.

omichel (approved):
Thank you.

PR #6648, opened 2024-08-31
Run test suite on windows during CI

brettle:
**Description**
Run test suite on windows during CI.

Includes PR #6631.


src/webots/gui/WbSingleTaskApplication.cpp
```diff
@@ -231,10 +231,10 @@ void WbSingleTaskApplication::showSysInfo() const {
   // An OpenGL context is required there for the OpenGL calls like `glGetString`.
   // The format is QSurfaceFormat::defaultFormat() => OpenGL 3.3 defined in main.cpp.
   QOpenGLContext *context = new QOpenGLContext();
-  context->create();
+  assert(context->create());
```
CoolSpy3, line 234: This will get optimized out in a release build. I'm actually not sure why it's not breaking the CI now :/
```suggestion
  if(!context->create())
    assert(false);
```

brettle:
Although this PR currently works in CI, I'm getting a likely unrelated crash on my local machine when running:

```
./tests/test_suite.py --group api tests/api/worlds/robot_urdf.wbt  tests/api/worlds/supervisor_set_hinge_position_dynamic.wbt
```

The robot_urdf test passes but then there is a crash when running the second test. The problem goes away if I remove the `--minimize` at `test_suite.py:246`.

Still investigating but if anyone else can let me know whether or not they can reproduce it, that would be good to know.

brettle:
> Still investigating but if anyone else can let me know whether or not they can reproduce it, that would be good to know.

Further investigation indicates that the crash is unrelated to any changes in this PR. I'll file a separate issue for it.

brettle:
This is now ready for review.

omichel (approved):
Thank you!

Issue #6659, opened 2024-09-12
Crash on windows when combining "--minimize", `animatedGeometry`, and `muscles`


brettle:
**Describe the Bug**
On Windows with the current master branch ([34dfe53](https://github.com/cyberbotics/webots/commit/34dfe53ffc29922e109d15313846d4ce00dfbe03)), when running webots with `--minimize`, if I run a world with a `Track` containing `animatedGeometry`, and then try to load a world with a `muscles` node, webots crashes.

**Steps to Reproduce**
1. Run: 
```
./tests/test_suite.py --group api tests/api/worlds/robot_urdf.wbt  tests/api/worlds/supervisor_set_hinge_position_dynamic.wbt`
```
or
```
./tests/test_suite.py --group api tests/api/worlds/robot_track.wbt  tests/api/worlds/supervisor_set_hinge_position_dynamic.wbt`
```
The first test will pass but webots will crash when running the second test.

The crash does **not** occur if I comment out the `--minimize` option is `test_suite.py`, remove the `animatedGeometry` node from `robot_track.wbt`/`robot_urdf.wbt`, remove the `muscles` node from `supervisor_set_hinge_position_dynamic.wbt`, or when running CI which uses a sofware renderer.

**Expected behavior**
Should not crash.

**System**
```
System              : Windows 10 Version 22H2 64-bit
Processor           : Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
Number of cores     : 6
OpenAL device       : OpenAL Soft
OpenGL vendor       : NVIDIA Corporation (0x10de)
OpenGL renderer     : Quadro T1000/PCIe/SSE2 (0x1fb9)
OpenGL version      : 4.6.0 NVIDIA 538.78
```

**Additional Context**

My best guess is that perhaps a bug in the code which destroys some animated geometry objects is causing memory corruption that is tickling the muscles code.

brettle:
This bug was introduced in PR #6617. Still looking into it, but for reference, here is a minimal test case:
[issue6659.tar.gz](https://github.com/user-attachments/files/17003344/issue6659.tar.gz)

brettle:
Fixed by PR #6661.

PR #6661, opened 2024-09-16
Fix crash when using muscles and animatedGeometry in sequential worlds.

brettle:
**Description**

The issue was that `WbWrenVertexArrayFrameListener` thought that the frame had already been handled because it had been handled in the previous world.

**Related Issues**
This pull-request fixes issue #6659

**Tasks**

No changelog update because the issue did not exist in R2023a.


omichel (approved):
Great! Thank you.

PR #6666, opened 2024-09-20
Ensure that test_suite.py only tries to build controllers which have Makefiles.

brettle:
**Description**

Switching between master and develop branches can result in controller directories sticking around that only contain the results of a build on the previous branch. test_suite.py was trying (and failing) to run make in those directories.


omichel (approved):
Thank you.

PR #6677, opened 2024-10-20
Fix bug where the wrong y coord was used for one corner of the box drawn to represent a box-plane collision.

brettle:
**Description**
Fix bug where the wrong y coord was used for one corner of the box drawn to represent a box-plane collision.

**Tasks**


omichel (approved):
Thank you.

PR #6678, opened 2024-10-20
Show box-plane contact point normals when showing contact points.

brettle:
**Description**
Show box-plane contact point normals when showing contact points. This had been intentionally prevented, but it isn't clear why. I found it confusing that these were missing when I enabled showing contact points.

**Tasks**


omichel (approved):
Thank you.

Issue #5225, opened 2022-09-15
World not flagged as modified after reset


stefaniapedrazzi:
Steps to reproduce:
1. open "accelerometer.wbt"
2. run the simulation
3. Change any field in the world, for example the SLOPE Solid name field
4. Reset simulation
-> the changes to the "name" field is not reset, but the "Save world" button is disabled, i.e., the world modified flag is wrongly reset to FALSE

PR #6686, opened 2024-10-24
Fix issue #5225

MNikoliCC:
**Description**
All modified 'DEF' fields are reset to their initial values when the simulation is reset. The world modified flag is only set when the simulation time is 0 and there is a change between any of the initial and modified values in the 'DEF' fields.

**Related Issues**
This pull-request fixes issue #5225

**Tasks**
  - [x] 'DEF' fields are reset to their initial values when the simulation is reset.
  - [x] The world modified flag is set at appropriate time

### DIFF
--- a/src/webots/scene_tree/WbNodeEditor.cpp
+++ b/src/webots/scene_tree/WbNodeEditor.cpp
@@ -33,6 +33,8 @@
 #include "WbViewpoint.hpp"
 #include "WbVrmlNodeUtilities.hpp"
 #include "WbWorldInfo.hpp"
+#include "WbWorld.hpp"
+#include "WbActionManager.hpp"
 
 #include <QtCore/QDir>
 #include <QtWidgets/QCheckBox>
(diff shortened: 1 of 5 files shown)


omichel (changes requested):
Source code tests are failing.
Can you fix them?

src/webots/scene_tree/WbNodeEditor.cpp
```diff
@@ -87,6 +91,25 @@ WbNodeEditor::WbNodeEditor(QWidget *parent) :
   connect(mPrintUrl, &QPushButton::pressed, this, &WbNodeEditor::printUrl);
   connect(mShowResizeHandlesCheckBox, &QAbstractButton::toggled, WbSelection::instance(),
           &WbSelection::showResizeManipulatorFromSceneTree, Qt::UniqueConnection);
+  connect(worldCheckTimer, &QTimer::timeout, this, &WbNodeEditor::tryConnectToWorld);
+  worldCheckTimer->start(500);
+}
+
+void WbNodeEditor::tryConnectToWorld() {
```
lukicdarkoo, line 98: We cannot have a timer here, we need a cleaner solution

lukicdarkoo (changes requested):
We need to find a simpler and more generic solution (to cover not only DEF but other fields as well). Let's pause on this until we figure out a better solution

PR #6688, opened 2024-10-27
Improve speed, accuracy, and readability of dCollideBoxPlane.

brettle:
**Description**

Improve accuracy by always returning the maxContacts deepest contacts. The old code would produce at most 4 contacts and the 4th contact wouldn't necessarily be the 4th deepest. This could result in oscillations if the entire box went under the plane. Fixing this means that the change isn't perfectly backward compatible. In particular, the values in one test are adjusted because they relied on a thin box resting entirely below the ground for which the new code generates 8 contact points/joints instead of 4.

The new code is significantly faster at colliding random boxes and planes when maxContacts <= 4, slightly faster when colliding boxes that are "settling" on the plane (ie. the most common case), and very close to the same speed in other cases.

Improve readabilitly by avoiding goto statements, using less macros with clearer names, and commenting heavily.

**Tasks**

**Additional context**

A [PR has been submitted to ODE](https://bitbucket.org/odedevs/ode/pull-requests/15) that contains the same implementation plus code for comparing performance and checking accuracy against the original code. See also [this discussion](https://groups.google.com/g/ode-users/c/smY9AioGArk) on the ODE mailing list.


brettle:
I've updated the link link to the ODE PR to point to the correct one.

omichel (approved):
Thank you.

Issue #6675, opened 2024-10-14
Segmentation fault on Windows by following Tutorial


Fredrikhb:
**Describe the Bug**
CTD

**Steps to Reproduce**
1. Follow https://github.com/cyberbotics/webots/wiki/Windows-installation/ to install webots on Windows
2. Run webots through MINGW64 by typing 'webots' and hit enter
3. Add a Robot node having four HingeJoint nodes as children.
4. Add a Shape node to the boundingObject field of the Robot node.
5. In the appearing geometry NULL field, add a Box node.
6. Crash to desktop. 
7. Read off the MINGW64 terminal "Segmentation fault"

**Expected behavior**
A box node to be added to the geometry field of the Robot node's Shape node.

**System**
 - Operating System: Windows 10 Education 64-bit
 - Graphics Card: Intel Iris Xe Graphics

**Additional context**
I simply followed the Tutorial 6: 4-Wheels Robot.

UPDATE: Launching the Webots Windows desktop app does not produce the segmentation fault. It only happens when running the application through the MINGW64 terminal. (However, launching the application directly gives me issues when trying to compile C++ code)

blazaid:
Hi there,

I am experiencing the same segmentation fault described in the original post, but in my case, I'm running Webots on Linux Mint 22 "Wilma". The crash happens under the same conditions:

1. Create a "Robot" node.
2. Under "BoundingBox", add a "Shape" node.
3. Inside the "Shape", under "geometry", add a "Box" node.

Webots crashes immediately with the following error:

`/usr/local/bin/webots: line 105: 33647 Segmentation fault      (core dumped) "$webots_home/bin/webots-bin" "$@"`

**System information:**

- Linux Mint 22 "Wilma"
- Webots version: 2023b
- Graphics card: WhiskeyLake-U GT2 [UHD Graphics 620]

It seems to be the same issue as previously reported.

ShuffleWire:
Reproduced with Webots 2023b, Debian 12
Traceback
````
Thread 1 "webots-bin" received signal SIGABRT, Aborted.
__pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, 
    no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
44	./nptl/pthread_kill.c: No such file or directory.
(gdb) where
[...]

PR #6691, opened 2024-10-29
Fix Bounding Box Initialization when a Geometry is Inserted into a Shape

CoolSpy3:
**Description**
Currently, when `WbMatter` creates an ode geometry from a newly inserted `WbGeometry` in a `Shape` node, `WbGeometry::setOdeData` is called twice by `WbMatter::createOdeGeomFromInsertedShapeItem` (once directly and once through the call to `WbMatter::createOdeGeomFromGeometry`). Because `setOdeData` destroys its old geometry, this has the effect of setting the geometry to an invalid pointer. This PR removes one of these calls so that the geometry is only set once. Additionally, the assumption that `geom != mOdeGeom` in `WbGeometry::setOdeData` is now included in an assertion, so future similar problems should be easier to identify.

I also did a small bit of deduplication in e254766. Those changes can be reviewed independently. If they are determined to be out of scope for this PR, that commit can be safely reverted without affecting the rest of the PR.

**Related Issues**
This pull-request fixes issue #6675.

**Tasks**


omichel (approved):
LGTM. Thank you!

PR #6701, opened 2024-11-16
Remove Old Import Node Methods from WbLanguage

CoolSpy3:
**Description**
As far as I can tell, the `wb_supervisor_field_import_(sf/mf)_node` methods were removed and replaced by their `*_from_string` variants. However, they were never removed from `WbLanguage` so they're still suggested in the Webots editor despite no longer existing in `supervisor.c`. This PR removes them.

**Tasks**


omichel (approved):
Thank you.

PR #6702, opened 2024-11-17
Fix Worlds Test on Forked Repositories

CoolSpy3:
**Description**
Currently, the `test-worlds` job in the Linux CI workflow explicitly refers to the `cyberbotics/Webots` repository when attempting to download the necessary scripts. This causes the workflow to fail when run on a fork. This PR updates the job to use a sparse checkout to retrieve the files, which is repo-agnostic.


omichel (approved):
Thank you.

PR #6704, opened 2024-11-18
Remove Ubuntu 20.04 and add 24.04 support

MNikoliCC:
**Description**
This pull-request focuses on removing support for Ubuntu 20.04 and adding support for Ubuntu 24.04. This involves updating dependencies, testing compatibility, and ensuring smooth packaging for the newer distribution.

**Tasks**
  - [x] Update the documentation
  - [x] Remove Ubuntu 20.04-specific dependencies and configurations.
  - [x] Add Ubuntu 24.04 support in scripts and packaging workflows.
  - [x]  Test and validate compatibility with Ubuntu 24.04.
  - [x]  Update CI/CD pipelines for Ubuntu 24.04.

**Additional context**
This update transitions Webots to the latest supported LTS versions of Ubuntu. While dropping 20.04, the focus is on enhancing compatibility and functionality for users adopting Ubuntu 24.04. This ensures alignment with modern software ecosystems and provides access to newer libraries and features.

The packages: "_libHalf-2_5.so.25_", "_libIexMath-2_5.so.25_" and "_libIlmImf-2_5.so.25_" are missing on Ubuntu 24.04 but the tests are passing.
https://github.com/cyberbotics/webots/blob/d5402dbafc68051a0fd4ba29c85dc7269cb2af8f/scripts/packaging/linux_distro.py#L51-L55


projects/languages/java/controllers/Driver/Driver.java
```diff
@@ -111,6 +114,7 @@ private void displayHelp(){
 
   public static void main(String[] args) {
     Driver controller = new Driver();
+    controller.initialize();
```
MNikoliCC, line 117: This is needed to avoid the "this-escape" warning on Java 21.

tests/test_suite.py
```diff
@@ -206,8 +209,8 @@ def generateWorldsList(groupName):
                     not ('GITHUB_ACTIONS' in os.environ and (
                         filename.endswith('speaker.wbt') or
                         filename.endswith('local_proto_with_texture.wbt') or
-                        (filename.endswith('robot_window_html.wbt') and is_ubuntu_22_04) or
-                        (filename.endswith('supervisor_start_stop_movie.wbt') and is_ubuntu_22_04) or
+                        (filename.endswith('robot_window_html.wbt') and (is_ubuntu_22_04 or is_ubuntu_24_04)) or
```
lukicdarkoo, line 212: This added condition means `supervisor_start_stop_movie.wbt` and `robot_window_html.wbt` will not be tested at all on Linux. We should check if we can test it at least for one version of Ubuntu

scripts/packaging/linux_distro.py
```diff
@@ -67,6 +57,13 @@ class LinuxWebotsPackage(WebotsPackage):
         "libzip.so.4",  # needed by Robotis OP2
         "libx264.so.163"
     ]
+    USR_LIB_X68_64_24_04 = [
+        "libIex-3_1.so.30",
+        "libIlmThread-3_1.so.30",
+        "libwebp.so.7",
```
lukicdarkoo, line 63: Here we are missing some libraries. We discussed them, but just leaving a note

lukicdarkoo: I believe these libraries are related to the tar distribution. It would be good to test the tar distribution on a clean Ubuntu 24.04

lukicdarkoo (changes requested):
In general it looks good, there is one issue I noticed.

Please also check once more the following PR as a reference:
https://github.com/cyberbotics/webots/pull/4502

lukicdarkoo (approved):
Looks good, thank you

lukicdarkoo (approved):

PR #6726, opened 2024-12-10
Fix test sources with latest cppcheck

MNikoliCC:
**Description**
Fix cpp code to pass the test sources with latest cppcheck.

**Tasks**
  - [ ] Fix cpp code issues

### DIFF
--- a/src/webots/nodes/WbRobot.cpp
+++ b/src/webots/nodes/WbRobot.cpp
@@ -286,7 +286,7 @@ void WbRobot::addDevices(WbNode *node) {
       if (renderingDevice) {
         connect(renderingDevice, &WbBaseNode::isBeingDestroyed, this, &WbRobot::removeRenderingDevice, Qt::UniqueConnection);
         mRenderingDevices.append(renderingDevice);
-        WbAbstractCamera *camera = dynamic_cast<WbAbstractCamera *>(renderingDevice);
+        WbAbstractCamera *const camera = dynamic_cast<WbAbstractCamera *>(renderingDevice);
         if (camera) {
           connect(camera, &WbAbstractCamera::enabled, this, &WbRobot::updateActiveCameras, Qt::UniqueConnection);
           if (camera->isEnabled())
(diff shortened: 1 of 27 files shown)


MNikoliCC:
@lukicdarkoo The PR is ready for review. When the PR is approved, I'll revert the changes in the _tests_sources_with_latest_cppcheck.yml_ file before merging if that is okay: https://github.com/cyberbotics/webots/blob/b3a07306e5b440cf9800bd8124bf1ef83bd6e024/.github/workflows/tests_sources_with_latest_cppcheck.yml#L3-L7

lukicdarkoo (approved):
Thank you!

src/webots/nodes/WbRobot.cpp
```diff
@@ -270,7 +270,8 @@ void WbRobot::addDevices(WbNode *node) {
   if (!node)
     return;
 
-  if (node != this && dynamic_cast<const WbRobot *>(node))
+  const WbRobot *robotNode = dynamic_cast<const WbRobot *>(node);
+  if (node != this && robrobotNodeot)
```
omichel, line 274: That looks wrong (typo)

omichel (commented):
Just something looking strange. I am surprised the tests passed.

src/webots/nodes/WbRobot.cpp
lukicdarkoo: This a good catch

MNikoliCC: Thank you for pointing this out

MNikoliCC:
After reviewing the issue with latest cppcheck, I suggest not to make further changes at this time. Addressing the warnings would require adding const to a large number of methods, which involves verifying that each method does not modify any member variables of the passed object.

While it would be possible to add `const` indiscriminately to pass the checks, this approach may not be entirely safe or appropriate without thorough validation. Therefore, I'm opting to leave the current implementation as is.

Issue #6604, opened 2024-08-01
Nodes Cannot Be Inserted into an Unconnected Field of a PROTO


CoolSpy3:
**Describe the Bug**
This is a bug I discovered while working on #6574 and then forgot to open an issue for. If there is an unconnected `SFNode` or `MFNode` field in a proto, no nodes can be inserted into it. The cause appears to be [`WbNodeUtilities:1514`](https://github.com/cyberbotics/webots/blob/9dc232d3e0e177b966c95a2675ae7c31709dbb3e/src/webots/nodes/utils/WbNodeUtilities.cpp#L1514). `isParameter()` is actually checking that `mInternalFields.size() != 0`, which is not true for unconnected fields. The fix should be to fix `isParameter` to work for `unconnectedField`s. The code will then have to be updated as some methods assume that `isParameter() == true => internalFields().size() > 0`. However, I haven't yet thought of a good way for determining whether a field is a parameter. So far my thoughts are either: 1) adding another flag, or 2) maybe something like `parentNode()->isProtoInstance()` (If that doesn't give false positives).

**Steps to Reproduce**
1. Create the following proto:
``` javascript
#VRML_SIM R2023b utf8

PROTO Test [
    unconnectedField SFNode node NULL
]
{
    Solid {
    }
}
```
2. Add it to a world
3. Try to insert something in the `node` field
4. Notice that the "Add a node" dialog is empty

**Expected behavior**
In this case, all available nodes (that do not require a robot ancestor) should be displayed. If there were any node restrictions on the node, those should also be considered.

**System**
 - Operating System: Windows 10
 - Graphics Card: NVIDIA GTX 1050 Ti

PR #6735, opened 2024-12-20
Fix `WbField::isParameter` for Unconnected Fields

CoolSpy3:
**Description**
Updates `WbField::isParameter()` so that it returns true even on `unconnectedFields` (which have no internal fields). This fixes the linked issue.

Some parts of the codebase previously assumed that `field->isParameter() => !field->internalFields().isEmpty()`. Given that that is no longer the case, those sections have been updated to perform that check explicitly. Additionally, `WbTemplateManager::regenerateNodeFromFieldChange` and `WbTemplateManager::regenerateNodeFromParameterChange` appeared to be workarounds to call `WbTemplateManager::regenerateNodeFromField` given that there was no way to directly tell if a field was a parameter, so I updated `regenerateNodeFromField` to make use of `WbField::isParameter` and deleted the other two.

**Related Issues**
This pull-request fixes issue #6604.

**Tasks**


CoolSpy3:
This is ready for review. The ubuntu-20.04 test suite is [currently failing on master](https://github.com/cyberbotics/webots/actions/runs/12422173567/job/34684230618) with the same error, so it's unrelated to this PR. The 22.04 suite passes.

CoolSpy3:
Update: The failure should be fixed by #6736.

omichel (approved):
Thank you.

PR #6719, opened 2024-12-05
Fix main branch builds

MNikoliCC:
**Description**
The PR aims to address issues related to building Debian packages on Ubuntu, as well as Windows installer packages.

**Tasks**
Add the list of tasks of this PR.
  - [x] Fix Linux build
  - [x] Fix Windows build

PR #6736, opened 2024-12-20
Restore Support for Ubuntu 20.04 in Dependency Installation Scripts

CoolSpy3:
**Description**
#6719 removed support for Ubuntu 20.04 in some of the setup scripts. However, because the master branch is still built against 20.04 (iirc only develop has been fully updated for 24.04), this is causing CI failures. IMO, this removal should only target develop and not master. However, given that the other changes from that PR are necessary to fix the Windows and develop builds, it does not make sense to fully revert #6719.

This PR reverts enough to fix the master CI and retains support for 24.04. Given that develop no longer supports 20.04, after this is synchronized, another PR should be opened to fully revert this PR (#6736) *on develop only*. (This can be done by either me or another contributor/maintainer.)


MNikoliCC (approved):
Good point, I agree with the decision to partially revert #6719. The changes look good to me, so please proceed with the plan to synchronize and open a follow-up PR for develop as outlined.
Thank you!

lukicdarkoo (approved):

PR #6717, opened 2024-12-03
Fix Heron warnings

MNikoliCC:
**Description**
The `EXTERNPROTO` import paths for certain PROTO files were incorrectly referencing raw GitHub URLs, such as `https://raw.githubusercontent.com/cyberbotics/webots/R2025a/`.... This has been fixed by replacing them with the correct Webots-specific protocol `webots://`.

**Related Issues**
Fixes #6716

**Tasks**
Add the list of tasks of this PR.
  - [x] Fix bugs

PR #6742, opened 2024-12-22
Fix Heron Warnings (Port #6717)

CoolSpy3:
**Description**
This PR just ports #6717 to `master` to get the worlds test working again. I also had to make an additional two minor changes:
1) I updated the path to the Heron proto to be absolute so that Webots could find it correctly.
2) I removed the `Transform` wrapper around the `Fluid` bounding boxes. Webots was automatically replacing it with a `Pose` node, causing a world update and failing the test. Because it only had a single child, I opted to just remove it.


MNikoliCC (approved):
Looks good, thank you

lukicdarkoo (approved):

PR #6744, opened 2024-12-22
Fix Crash When Loading an Invalid Background HDR Image

CoolSpy3:
**Description**
Currently, if an invalid HDR image is loaded by `WbBackground`, the input data is read as `NULL`, causing a segfault. This PR adds a statement to check for that condition and print an error message which contains the cause from `stb_image`.


docs/reference/changelog-r2024.md
```diff
@@ -34,3 +34,4 @@ Released on December **th, 2023.
     - Fixed a bug where Webots would crash if a geometry was inserted into a `Shape` node used a bounding box ([#6691](https://github.com/cyberbotics/webots/pull/6691)).
     - Removed the old `wb_supervisor_field_import_sf_node` and `wb_supervisor_field_import_mf_node` functions from the list of editor autocomplete suggestions ([#6701](https://github.com/cyberbotics/webots/pull/6701)).
     - Fixed a bug preventing nodes from being inserted into unconnected proto fields ([#6735](https://github.com/cyberbotics/webots/pull/6735)).
+    - Fixed crash when an invalid HDR image was set as a world background ([#6744](httpsjson://github.com/cyberbotics/webots/pull/6744)).
```
lukicdarkoo, line 37: Something is wrong with the URL

lukicdarkoo (changes requested):
Looks good, just check the URL

docs/reference/changelog-r2024.md
CoolSpy3: Good catch! No idea how that happened :P Fixed!

lukicdarkoo (approved):

PR #6749, opened 2024-12-26
Attempting to use cmake instead of make (example)

Kreijstal:
I've managed to compile webots (core) on archlinux
https://github.com/cyberbotics/webots/discussions/6743
```
cmake -GNinja -DCMAKE_C_COMPILER_LAUNCHER=sccache -DCMAKE_CXX_COMPILER_LAUNCHER=sccache -B build -DCMAKE_INSTALL_PREFIX=dummy
cmake --build build
```
The reason I did this is because I believe cmake can speed the compilation a lot... of course there is a lot to do, but I find the curren makefiles are a bit buggy and nonstandard so it's hard to package webots to a distro like debian or even msys2. This might help someone who wants to compile webots on nonubuntu distros and might even replace the current makefiles as they are. Enjoy

In msys2 you can try 
```
pacboy -S ois:p sccache:p qt6:p stb:p pkgconf:p popt:p openvr:p dlfcn:p openal:p cc:p cmake:p --needed

```


Kreijstal:
closing favoring #6832

PR #6697, opened 2024-11-13
Remove ROS1 support

MNikoliCC:
**Description**
Remove ROS 1 support from Webots, including dependencies, code, and documentation references.

**Tasks**
Add the list of tasks of this PR.
  - [x] Remove ROS 1 support from Webots, including dependencies
  - [x] Update documentation to reflect the removal of ROS 1 support

**Additional context**
This PR aims to streamline the codebase by fully removing ROS 1 support. Note any related issues, deprecations, or compatibility changes here.

PR #6757, opened 2025-01-12
Fix `test-suite` Label on Linux

CoolSpy3:
**Description**
#6697 accidently deleted the directive to run the Linux build job when the `test-suite` label is applied to a PR. This PR restores that condition.


MNikoliCC (approved):
Thanks for correcting this, good job

omichel (approved):

PR #6764, opened 2025-01-30
Fix reference s in the URL sample

lukicdarkoo:


projects/samples/howto/url/controllers/url/url.c
```diff
@@ -53,7 +53,7 @@ int main() {
       printf("The texture URL of the robot body was changed from the supervisor to a non-existing file.\n");
     } else if (counter == 160) {
       wb_supervisor_field_set_mf_string(url, 0,
-                                        "https://raw.githubusercontent.com/cyberbotics/webots/R2025a/projects/appearances/"
+                                        "https://raw.githubusercontent.com/cyberbotics/webots/R2023b/projects/appearances/"
```
lukicdarkoo, line 56: @omichel  I don't how else to fix this. We have a circular dependency here (it requires tag, but the tag requires this to pass)

omichel: We should probably merge this PR even if the tests fail. I can do this as admin.

PR #6780, opened 2025-02-22
Adding compilation support for Fedora 41

calvarado2004:
**Description**
Adding compilation support for Fedora 41

**Documentation**
If this pull-request changes the installation scripts for Linux, adding support for Fedora 41, among Ubuntu.

### DIFF
--- a/scripts/install/linux_compilation_dependencies.sh
+++ b/scripts/install/linux_compilation_dependencies.sh
@@ -1,25 +1,66 @@
 #!/bin/bash
 
-# exit when any command fails on CI
+# Exit when any command fails on CI
 if [[ ! -z "$CI" ]]; then
-       set -e
+    set -e
 fi
 
 if [[ $EUID -ne 0 ]]; then
-       echo "This script must be run as root"
-       exit 1
+    echo "This script must be run as root"
+    exit 1
 fi
 
-alias apt='apt --option="APT::Acquire::Retries=3"'
-apt update
-apt install --yes git lsb-release cmake swig libglu1-mesa-dev libglib2.0-dev libfreeimage3 libfreetype6-dev libxml2-dev libboost-dev libssh-gcrypt-dev libzip-dev libreadline-dev pbzip2 wget zip unzip python3 python3-pip libopenal-dev
-
-UBUNTU_VERSION=$(lsb_release -rs)
-if [[ $UBUNTU_VERSION == "22.04" || $UBUNTU_VERSION == "24.04" ]]; then
-       apt install --yes libzip4 openssl
+# Detect the operating system
+if [ -f /etc/os-release ]; then
+    . /etc/os-release
+    OS=$ID
+    VERSION_ID=$VERSION_ID
 else
-       echo "Unsupported Linux version: dependencies may not be completely installed. Only the two latest Ubuntu LTS are supported."
+    echo "Cannot determine the operating system."
+    exit 1
 fi
+
+# Function to install packages on Ubuntu
+install_ubuntu_packages() {
+    alias apt='apt --option="APT::Acquire::Retries=3"'
+    apt update
+    apt install --yes git lsb-release cmake swig libglu1-mesa-dev libglib2.0-dev libfreeimage3 libfreetype6-dev libxml2-dev libboost-dev libssh-gcrypt-dev libzip-dev libreadline-dev pbzip2 wget zip unzip python3 python3-pip libopenal-dev
+
+    if [[ $VERSION_ID == "20.04" ]]; then
+        apt install --yes libzip5 perl libtext-template-perl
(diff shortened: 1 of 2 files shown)


CoolSpy3 (changes requested):
It looks like the build still works (although, as of writing, the test suite is still running). Assuming that passes too, I'm fine to approve this with the one minor correction noted below.

That being said, Fedora is not officially supported by Webots, and I think its unlikely that the maintainers are going to go through the work to change that. Thus, I'm not sure if its wise to add Fedora support as it's unlikely to be maintained when dependency requirements change.

A possible compromise might be to add an echo statement warning of that fact, and let Fedora become a quasi-supported OS. (We can also probably make a note of that somewhere on the wiki.) However, that still leaves a chunk of explicitly-unmaintained code in the repo.

@omichel can probably make the final call here.

omichel:
I agree with @CoolSpy3, we should add an echo statement warning that Fedora support is still experimental and not officially documented. Then, we could merge this PR.

CoolSpy3:
@calvarado2004 I'm trying to test this PR on Fedora 42. Did you manage to find a way to get around the include errors in wren? I'm currently stuck on the one in `ftconfig.h`, but I think there are a couple more arising from the headers detecting my x64 system as 32-bit.

I'm thinking I'm going to have to set another preprocessor directive and include `/usr/include` to get it working, but I wanted to  check if you found a more elegant solution before I start digging around in Makefiles.

CoolSpy3:
Okay; I've managed to get a working Fedora build! Keynotes:
* `lsb_release` doesn't work correctly on Fedora. @calvarado2004 worked around this in his changes, but I went a step further and removed the dependency completely, instead using `/etc/os-release` where applicable. I think those changes slot fairly nicely into the existing code. It does make the test suite setup a *little* more complex, but I think it's still plenty readable.
* The include error I was running into is caused by Emscripten building in 32-bit mode for wasm, which breaks 64-bit systems because Fedora's `libfreetype` has an extra bit of size-checking code which tries to load different headers based on the architecture size. Thus, the Fedora install scripts now explicitly include both the 32 and 64 bit versions of `libfreetype` and `glibc`. (It's possible that this will lead to some extra 64-bit dependencies on 32-bit systems, but I don't think the added effort of working around that is worth saving a couple megabytes of storage for the one person still running a 32-bit system.)
* `libfreetype`'s bit-size checking code depends on `bits/wordsize.h`, which (from what I can tell) is a non-standard header. Thus, the only way I could make it work was by including all of `/usr/include`. Imo, that's a bit of an ugly/heavy-handed solution, but I haven't been able to think of anything better (suggestions welcome). To minimize the impact, I've added some code to detect the Linux distribution and gated that include to Fedora only.
* I also added warnings to the install scripts about Fedora's unofficial status. It would be nice to add something to the end of the build process too, but I couldn't find a nice way to do post-run actions in `make`, and given that it's now part of the installation scripts, I don't think it's as necessary.

As of now, CI is passing, and my build passes all but two of the tests on Fedora. (I'll investigate tomorrow; it's late now.) So, pending that, I'm good to merge this! (Will approve once that's done.)

Keep in mind that I've been testing on Fedora 42. (I'll update the PR title to reflect that.) @calvarado2004 If you want to test against Fedora 41, feel free to do so and report your findings, but I will not be explicitly testing against it otherwise. Thank you for the work you put into getting this started!

**BONUS NOTE: I just realized I actually broke the Ubuntu 22.04 check in the test suite, but the test still passed, so it's possible that that code is no longer needed. I'll wait till tomorrow to remove it so that we get at least 3 CI runs with it (hopefully passing). [The comment says that it's from the 22.04 BETA, so it's possible it's been fixed in the latest release.] If it passes, we can probably just remove the check.**

CoolSpy3 (approved):
Okay, here's what I found with the tests:
`billboard.wbt` - Fails when run in the test suite but passes in a normal Webots session or when the suite is run through `xvfb`. Based on the output, it looks like its failing due to a window-size bug. Given Fedora's unofficial support status and the fact that this is only a problem in the suite, I don't think it's worth fixing.
`supervisor_start_stop_movie.wbt` - This was failing due to a missing video codec. Fedora's repos normally don't include proprietary codecs. I updated the install script to enable RPMFusion and the full (proprietary) version of `ffmpeg`, and that appears to have fixed the problem.

As for the Ubuntu 22.04 test. It does appear to pass consistently (~3 consecutive times so far with no failures), so I think it's good to re-enable. However, even though it's a small change, I think it makes sense to address it in another PR, so I won't back it out here.

Regardless, I'm now happy with the state of this PR, so I'm going to approve!

CoolSpy3 (approved):

omichel (approved):
LGTM. Thank you.

PR #6654, opened 2024-09-06
Bump actions/upload-artifact to v4

CoolSpy3:
**Description**
#6652  bumped `actions/download-artifact` to v4.1.7. This is incompatible with `actions/upload-artifact@v3`, leading to CI failures. This PR bumps `actions/upload-artifact` to the latest version, fixing the CI.

PR #6787, opened 2025-03-08
Bump `geekyeggo/delete-artifact` to v5

CoolSpy3:
**Description**
The CI is currently failing because `geekyeggo/delete-artifact` is unable to list artifacts correctly. My guess is that this is because `v4` artifacts support wasn't introduced until v4 of the action. (Although I'm not sure why this didn't occur when #6654 was originally merged.)

This PR bumps the action version to v5, which should fix this issue.


omichel (approved):
Thank you!

PR #6794, opened 2025-03-20
Fix buffer overflow

Pylgos:
**Description**
We need to allocate an extra byte for null termination.


omichel (approved):
Thank you!

omichel:
We should probably backport this fix on the master branch as well.

Issue #6805, opened 2025-04-10
"Bad CPU type in executable" when launching "webots-controller" for robot external controller in MacOS with M4 chip


kelvin-olaiya:
**Describe the Bug**
Launching the *webots-controller* executable, for external controllers, yields a "Bad CPU type in executable" error

**Steps to Reproduce**
1. Open terminal
2. Execute the $WEBOTS_HOME/Contents/MacOS/webots-controller binary

**Expected behavior**
The help message is displayed

**Screenshots**
![Image](https://github.com/user-attachments/assets/c3ce2796-3870-4d49-bd00-48dc3a2ad604)

**System**
 - Operating System: macOS Sequoia
 - Graphics Card: Apple GPU 16GB unified memory 

**Additional context**
The Mac has an Apple M4 chip

CoolSpy3:
It looks like [the relevant Makefile](https://github.com/cyberbotics/webots/blob/c098bdf742a295a65f5b10f99c8a1a2c483f387a/src/controller/launcher/Makefile#L46) only builds that executable for the host platform. GitHub Actions is currently configured to use an x64 runner, so arm code isn't being generated.

I tried putting together a fix at #6806, but I don't have a mac to test on. If you get the chance, could you try installing the build of that PR once it's ready and seeing if that fixes the problem. It should be available on [this page](https://github.com/cyberbotics/webots/actions/runs/14398324185?pr=6806) in about 40-60 minutes. If the build fails, I'll try to debug it tomorrow.

CoolSpy3:
Great; glad I could help! I'll mark the PR as ready-for-review. Once it's merged, the fix should start appearing in the nightly builds, and eventually, the next major release 👍

PR #6806, opened 2025-04-11
Build `webots-controller` for both x86 and arm on MacOS

CoolSpy3:
**Description**
Currently the `webots-controller` executable is only built with the default cxx options, meaning that on macos, it only contains code for the platform it was built on. This PR updates the relevant `Makefile` to use the same approach as the rest of Webots to generate object code for both platforms.

**Related Issues**
This pull-request fixes issue #6805 

**Tasks**


omichel (approved):
Thank you!

Issue #6841, opened 2025-06-10
Function wb_speaker_stop(speaker, NULL) Crashes Webots


Mirtl02:
**Describe the Bug**
Calling the ```wb_speaker_stop(speaker, NULL)``` function causes Webots (Version R2025a) to crash without any error message in the console.

**Steps to Reproduce**
1. Open example world speaker.wbt
2. Insert line ```wb_speaker_stop(speaker, NULL);``` after line ```wb_speaker_play_sound(speaker, speaker, "sounds/robot_sound.wav", 1.0, 1.0, 0.0, true);```
3. Click play
4. Webots crashes without error message in console.

**Expected behavior**
Example runs without playing the sound "sounds/robot_sound.wav".

**System**
 - Operating System: Windows 11
 - Graphics Card: Nvidia Geforce GTX 1660 Super

**Additional context**
The same example works as expected in the previous version **R2023b**.

Complete code:  
```C
#include <webots/motor.h>
#include <webots/robot.h>
#include <webots/speaker.h>

#include <stdio.h>

#define TIME_STEP 64

int main() {
  WbDeviceTag speaker;

  wb_robot_init();

  /* get the speaker */
  speaker = wb_robot_get_device("speaker");

  /* do a circle around the center of the arena */
  WbDeviceTag left_motor = wb_robot_get_device("left wheel motor");
  WbDeviceTag right_motor = wb_robot_get_device("right wheel motor");
  wb_motor_set_position(left_motor, INFINITY);
  wb_motor_set_position(right_motor, INFINITY);
  wb_motor_set_velocity(left_motor, 8);
  wb_motor_set_velocity(right_motor, 10);

  /* use the speaker to play a sound file */
  wb_speaker_play_sound(speaker, speaker, "sounds/robot_sound.wav", 1.0, 1.0, 0.0, true);
  wb_speaker_stop(speaker, NULL);
  printf("I am going to use the speaker to play a sound file while I am moving in circle around the center of the arena.\n");
  printf("Please make sure the sound is enabled and do not move the viewpoint.\n");

  /* control loop */
  double offset = wb_robot_get_time();
  while (wb_robot_step(TIME_STEP) != -1) {
    // every 5 seconds play a second sound file using the same speaker
    if (wb_robot_get_time() - offset > 5.0) {
      wb_speaker_play_sound(speaker, speaker, "sounds/robot_bip.wav", 0.7, 1.0, 0.0, false);
      offset = wb_robot_get_time();
    }
[...]

PR #6605, opened 2024-08-01
 Improve sound support for external controller (#6591)

nhjschulz:
**Description**
Stream the sound file content from controller to Webots if the sound file path is accesible by the controller process. This allows external controller to use sounds even if the file path is not accessible from Webots (i.e relative to a controller working directory). Controller sends sound data for a file only once per speaker device. It gets cached inside WbSpeaker.

**Related Issues**
This pull-request fixes issue #6591 

**Documentation**
https://cyberbotics.com/doc/reference/speaker=https://github.com/nhjschulz/webots:ControllerSound_6591

**Screenshots**

**Additional context**

The patch changes the format of message  C_SPEAKER_PLAY_SOUND.

PR #6843, opened 2025-06-15
Fix `wb_speaker_play_sound` Request Message

CoolSpy3:
**Description**
As part of #6605, the `C_SPEAKER_PLAY_SOUND` message was updated to allow controllers to upload local audio files. However, that data was setup to be sent for all registered sounds whenever any sound was played. Webots, by contrast, only looked for this data on sounds which were requested to be played [[1]](https://github.com/cyberbotics/webots/blob/238dfd4ed63708b5ed389866caf28715d2b71e5b/src/webots/nodes/WbSpeaker.cpp#L85-L89) [[2]](https://github.com/cyberbotics/webots/blob/238dfd4ed63708b5ed389866caf28715d2b71e5b/src/webots/nodes/WbSpeaker.cpp#L311-L315) [[3]](https://github.com/cyberbotics/webots/blob/238dfd4ed63708b5ed389866caf28715d2b71e5b/src/controller/c/speaker.c#L159-L168). This meant that when a simulation used multiple sounds and `wb_speaker_play_sound` was called, the message would have trailing data corresponding to the other sounds' playback data (often just `\0` because the data was cached). This caused `WbSpeaker` to interpret this data as additional commands, leading to unexpected behavior.

This PR updates the controller API to only send the sound data when a sound object is updated, as expected by Webots.

As a side note: This triggered [this assert](https://github.com/cyberbotics/webots/blob/238dfd4ed63708b5ed389866caf28715d2b71e5b/src/webots/nodes/WbSpeaker.cpp#L136) when running the speaker test world. It might be worthwhile to run future release QAs against a develop build (or a release  build with asserts enabled) in order to catch similar bugs.

**Related Issues**
This pull-request fixes issue #6841

**Tasks**

### DIFF
--- a/docs/reference/changelog-r2025.md
+++ b/docs/reference/changelog-r2025.md
@@ -3,7 +3,7 @@
 ## Webots R2025b
   - Bug Fixes
     - Fixed a bug preventing the `webots-controller` executable from running on arm-based mac devices ([#6806](https://github.com/cyberbotics/webots/pull/6806)).
-    - Fixed a bug causing Webots to crash when multiple sounds were used in a simulation ([#6843](https://github.com/cyberbotics/webots/pull/6843)).
+    - Fixed a bug causing Webots to crash when multiple sounds were used with a [Speaker](speaker.md) node ([#6843](https://github.com/cyberbotics/webots/pull/6843)).
 
 ## Webots R2025a
 Released on January 31st, 2025.


omichel:
>As a side note: This triggered [this assert](https://github.com/cyberbotics/webots/blob/238dfd4ed63708b5ed389866caf28715d2b71e5b/src/webots/nodes/WbSpeaker.cpp#L136) when running the speaker test world. It might be worthwhile to run future release QAs against a develop build (or a release build with asserts enabled) in order to catch similar bugs.

I am not running QAs any more, but @lukicdarkoo might be interested by you suggestion

PR #6844, opened 2025-06-15
Fix Checking of Reload/Reset Selection in WbBuildEditor

CoolSpy3:
**Description**
`QMessageBox::exec()` returns an opaque value when used with custom buttons [[1]](https://doc.qt.io/qt-6/qmessagebox.html#exec). However, `WbBuildEditor` was treating this result as an indication of the user's selection. This lead to implementation-dependent behavior, causing the "Reload" and "Reset" buttons to work on Linux but do nothing on Windows.

This PR updates the relevant code to use `QMessageBox::clickedButton()`, which has predictable behavior on both platforms.

I did a quick scan for similar issues throughout the codebase and didn't find any.

**Tasks**


omichel (approved):
Thank you!

PR #6850, opened 2025-06-24
Use `QImage::flipped` instead of `QImage::mirrored` in `WbWrenTextureOverlay` on Windows

CoolSpy3:
**Description**
`QImage::mirrored` is deprecated in newer Qt versions. Windows always builds against the latest version of Qt, while Linux and Mac versions are pinned. Thus, recent CI runs are failing on Windows. This PR updates the relevant code to use the new function `QImage::flipped`.

Unfortunately, this new function is not available in older versions of Qt, so two separate code paths are necessary until Qt is bumped on the other platforms.


omichel (approved):
Thank you.

PR #6853, opened 2025-06-25
Use `<stdbool.h>` instead of `boolean.h`

CoolSpy3:
**Description**
Recent versions of the msys2 build tools are failing to build the `boolean.h` header because C23 introduced `bool`, `true`, and `false` as C keywords. [[1]](https://en.wikipedia.org/wiki/C23_(C_standard_revision)) (It should maybe be another discussion if the C version standard we use should be rolling [as-is] or pinned.) This PR deletes that header (as it no longer compiles) and swaps the relevant references to use the standard library boolean header (which should support equivalent functionality).


omichel (approved):
Thank you!

PR #6857, opened 2025-07-06
Set `mIsBeingDeleted` before emitting `isBeingDestroyed` in `WbNode`

CoolSpy3:
**Description**
This PR fixes the Windows CI test failures. Here's the rundown:
In the old code, `isBeingDestroyed` was emitted in the destructor of `WbBaseNode` and `mIsBeingDeleted` was set in the destructor of `WbNode`. `WbBaseNode` extends `WbNode`, so its destructor was called first, creating the confusing case where events triggered from `isBeingDestroyed` would see `WbNode::isBeingDeleted() == false`. This PR moves `isBeingDestroyed` to `WbNode` so that that can be avoided.

In the test suite crash, a `WbPose` node was deallocated. `WbPose` is defined as `class WbPose : public WbGroup, public WbAbstractPose`, so `~WbAbstractPose` was called first, causing `WbAbstractPose::mTranslateRotateManipulator` to be deleted and then set to `NULL` [[1]](https://github.com/cyberbotics/webots/blob/7f30a52696bf0ad914d500e3a0a356029a37f86e/src/webots/nodes/WbAbstractPose.cpp#L60). Then, `~WbBaseNode` was called, emitting `isBeingDestroyed`. This called [`WbSelection::clear`](https://github.com/cyberbotics/webots/blob/7f30a52696bf0ad914d500e3a0a356029a37f86e/src/webots/app/WbSelection.cpp#L144), which made its way to [`WbSelection.cpp::74`](https://github.com/cyberbotics/webots/blob/7f30a52696bf0ad914d500e3a0a356029a37f86e/src/webots/app/WbSelection.cpp#L74). The method it's calling checks `WbAbstractPose::mTranslateRotateManipulator`, and calls a method on it if it's not-null. In a debug build, the null-check failed, causing the code to not be run (as expected). However, in a release build, the `mTranslateRotateManipulator = NULL` was optimized out. (Because it's being called in the destructor, the state of the member variables doesn't matter.) Thus, the null check passes, but the variable has been deleted, causing a segfault. There was already a check to prevent this at [`WbSelection:68`](https://github.com/cyberbotics/webots/blob/7f30a52696bf0ad914d500e3a0a356029a37f86e/src/webots/app/WbSelection.cpp#L68), but it was checking `WbNode::isBeingDeleted()`, which (as discussed) was confusingly emitting `false` despite the node being in the process of being deleted. By making it so that that method returns `true` when the corresponding check is fired, the `WbSelection` check can work as-expected, preventing the problem.


CoolSpy3:
Hmmm... Looks like this broke one of the other tests :(
I'll look into it later today.

**Edit:** Fixed it! Just missed one of the signal uses. I also fixed a potential static vs, dynamic cast bug while I was at it :+1:

CoolSpy3:
The linux build failures are unrelated and are patched by #6755.

omichel (approved):
Lgtm. Thank you.

CoolSpy3:
Great :D This is good to merge! I'm just blocked because `windows-2019` is still listed as a required test.

omichel:
> Great :D This is good to merge! I'm just blocked because `windows-2019` is still listed as a required test.

I already changed this in the settings. I guess it will apply to newly created PRs.

Issue #6768, opened 2025-02-03
Python Command preferences crash


AndrewGerstenslager:
This is a problem where I was trying to upgrade to webots 2025a from 2021a so I could use a more modern python version so windows could see the GPU with torch.

So I was trying to point webots to a different python runtime and I set this in the preferences:
Python Command: conda run -n myenv python

Which was a mistake because webots crashed immediately and would only work if I deleted the contents of the world file that webots pointed to last. Even reinstalling webots didn't help because it stored some cache somewhere that I couldn't find to undo my mistake.

I realized this was the way to do it:
Python Command: C:\Users\<Username>\Anaconda3\envs\myenv\python.exe

If there's a way to catch this error before constantly crashing, that could be potentially helpful! Or some guidance on how to point to different python executables such as anaconda environments and virtual python environments.

Davidwarchy:
## Fatal Crash on Wrong Python Command 
I had a similar problem. I want the system to have a virtual environment. Normally one changes it in the preferences menu

![Image](https://github.com/user-attachments/assets/19208a02-6cc1-49b3-b5e7-18502d03d417)

I needed to use a virtual environment to run the python controller, so I put the path to the python.exe in the python command tab: 
`.torchgpu/Scripts/python.exe`

But there was a typo: `.torchgpu/Scripts/pythonexe` (no dot before 'exe')

Thing is that Webots crashed. Each restart ended up in a crash before there was any discernible action in the scene. Also Webots didn't give me enough time to change preferences, so this is a bit of an issue. 

The problem persists even if uninstall and restart the system and install again. I did this a few times, to much chagrin. 

## Solution 
The solution to this was going to the registry and changing the value for Computer\HKEY_CURRENT_USER\Software\Cyberbotics\Webots-R2025a\General\pythonCommand` to the correct path

fallenleaves04:
You could also cancel world objects loading on the startup to prevent the world from opening. This gives you time to change the command to your previous interpreter, thus undoing future crashes.

PR #6870, opened 2025-08-02
Fix Crash when Python is not Found on Windows

CoolSpy3:
**Description**
Currently, if the python command is not found on Windows, Webots crashes when trying to run Python controllers. This is because the version checks in `WbLanguageTools::pythonCommand` attempt to use string parsing techniques on the output of various python commands in order to determine the version. When python was not found, these functions just got the empty string (`""`), leading to unexpected behavior.

This PR updates the function to use safer checking code, preventing the crash.

As a side note: While working on this PR, I also noticed that one of the strings used on macos wasn't being translated. This PR passes that string to `QObject::tr`, allowing it to be translated.

**Related Issues**
This pull-request fixes the crash part of issue #6768. I'm working on another PR to actually support the requested behavior.

**Tasks**


omichel (approved):
Thank you.

PR #6874, opened 2025-08-17
Remove Movie Test Exclusion from Ubuntu 22.04

CoolSpy3:
**Description**
As discussed in [#6780 (comment)](https://github.com/cyberbotics/webots/pull/6780#pullrequestreview-3126168912), this PR removes the test exclusion for `supervisor_start_stop_movie.wbt` on Ubuntu 22.04. This actually removes the only use of `is_ubuntu_22_04` in the suite, which I think is neat. However, I've elected to leave the variable in until Ubuntu 22.04 support is phased out in case it becomes necessary. If maintainers want me to just remove it now, I can.


omichel (approved):
That's fine.

PR #6896, opened 2025-10-04
Print Cause when `WbAbstractCamera` Fails to Create Memory Mapped File

CoolSpy3:
**Description**
Someone [on Discord](https://discord.com/channels/565154702715518986/565154703139405824/1423410946990411856) is running into this "Cannot allocate memory mapped file for camera image." message. Unfortunately, as is, that message does not provide any info about why the map failed, making it hard to debug. This PR adds that info into the message. I'm using this PR to build Webots across all platforms with this patch, however, I'm going to keep this as a draft until we resolve the root cause, and then roll the fix in here if it makes sense.


CoolSpy3:
Looks like the user who originally encountered this has [found a workaround](https://discord.com/channels/565154702715518986/565154703139405824/1425226819682177104) and has moved on to other projects. However, I think I've managed to reproduce their issue (Tracked at #6910). That should probably be addressed in another PR. For now, this is good to merge.

omichel (approved):
Thank you!

Issue #6897, opened 2025-10-06
Simulation is running slower that expected


TonyWelte:
**Describe the Bug**

I've been tracking down an issue we are facing with our simulation running at 0.9 instead of 1.0 (even lower on some machines).

It is not an issue with CPU/GPU load. Webot is using barely 10% on a  16 threads machine. Running in fast mode the simulation is running at 5 to 6 times real time speed. It's only when running in real time mode that it can't keep up a stable 1.0 ratio.

I've tracked down the issue to this piece of code (in WbSimulationWorld.cpp):
```cpp
    // computing the mean of an history of several elapsedTime
    // improves significantly the stability of the algorithm
    // in case of simulations where elapsedTime oscillates often
    // above and below basicTimeStep.
    // Moreover it improves the stability of simulations where
    // basicTimeStep contains significant decimals
    mElapsedTimeHistory.append(elapsed);
    if (mElapsedTimeHistory.size() > qMax(4.0, 128.0 / timeStep))  // history size found empirically
      mElapsedTimeHistory.pop_front();
    double mean = 0.0;
    foreach (const int &v, mElapsedTimeHistory)
      mean += v;
    mean /= mElapsedTimeHistory.size();

    // useful hack: uncomment to run Webots at 90% of the real-time
    //              (if the real-time mode is enabled, of course)
    // mean *= 0.90;

    if (mean > timeStep && mSleepRealTime > 0.0) {
      mSleepRealTime -= 0.03 * timeStep;
      if (mSleepRealTime < 0)
        mSleepRealTime = 0.0;
    } else if (mean < timeStep)
      mSleepRealTime += 0.03 * timeStep;

    mTimer->start(mSleepRealTime);
```

Commenting out the last part that updates the timer period cause the simulation to be mostly stable with a real time ratio of 1.0.

I plotted the values of each variables and it looks like the time period is oscillating +/- 2ms around the expected timeStep.

<img width="1900" height="1287" alt="Image" src="https://github.com/user-attachments/assets/d975a7aa-d4b9-469f-ad7c-5ecb3edba1a7" />

I don't really understand how the adaptation of `mSleepRealTime` is supposed to work. The `mean` (which might be longer that timeStep if the simulation is running slower) is compared to `timeStep` which is constant and not affected by any slow down).

**Expected behavior**

Without issues with CPU load the simulation should be stable with a real time ratio around 1.0

**System**
 - Operating System: Linux Ubuntu 24.04

[...]

PR #6898, opened 2025-10-07
Improve Timestep Algorithm

CoolSpy3:
**Description**
This PR updates the timestep algorithm to calculate the mean of "ideal" delay times to determine how long to wait between timesteps rather than calculating the mean of elapsed times and calculating a delay based on that. This should remove the dependence of each data point on the current sleep time, reducing oscillations in the simulation speed. (See linked issue for more details.)

**Related Issues**
This pull-request fixes issue #6897.

**Tasks**
  - [ ] Update the [changelog](https://github.com/cyberbotics/webots/blob/master/docs/reference/changelog-r2025.md)


CoolSpy3:
Wait... I found an edge case when the timestep is changed while a world is running. (I think I meant to investigate this when I opened this and forgot :P ) Gimme a min, this shouldn't be a hard fix.

CoolSpy3:
Fixed.

omichel (approved):
Great! It both simplifies the code and improves the behavior.
Thank you!

PR #6927, opened 2026-02-05
Fix stb_image optimization bug with gcc 15.2.0

### LINKED ISSUE #6909: `TexturedBackground` Occasionally Creates Lighting Artifacts
**Describe the Bug**
Occasionally (not every time, but not uncommon), when opening a world (any world), there will be strange lighting artifacts. These range from glowing lights to strange reflections to everything being pitch black. (Some examples are shown below.) In every case I've tested, the problem is resolved by deleting the `TexturedBackground` node from the scene tree.

So far, I've been able to reproduce this fairly consistently on the latest nightly, but I haven't been able to reproduce in R2025a, so it appears to be a regression on `master`, but so far, none of the changes stand out to me as something that would cause this.

**Screenshots**

Image

Image

**System**
 - Graphics Card: NVIDIA GeForce GTX 1050Ti

omichel:
This PR fixes a random 3D rendering bug showing up on some Windows 11 (GeForce RTX 3060) in release mode (not in debug). It appears to be caused by gcc 15.2.0 failing to optimize some weird code (with some goto inside a loop) in stb_image.h. See https://github.com/nothings/stb/pull/1884 for details.

This patch is temporary and should be removed whenever https://github.com/nothings/stb/pull/1884 gets merged.

Fixes #6909.

### DIFF
--- a/docs/reference/changelog-r2025.md
+++ b/docs/reference/changelog-r2025.md
@@ -23,6 +23,7 @@
     - Fixed detection of physics plugins in extra projects ([#6880](https://github.com/cyberbotics/webots/pull/6880)).
     - Fixed `addForceWithOffset` and `addTorque` doing the same thing as `addForce` in Python ([#6881](https://github.com/cyberbotics/webots/pull/6881)).
     - Fixed Python controllers on Windows ([#6933](https://github.com/cyberbotics/webots/pull/6933)).
+    - OSM importer no longer crashes when run in 3d mode ([#6935](https://github.com/cyberbotics/webots/pull/6935)).
 
 ## Webots R2025a
 Released on January 31st, 2025.
--- a/resources/osm_importer/webots_objects/webots_object.py
+++ b/resources/osm_importer/webots_objects/webots_object.py
@@ -76,7 +76,7 @@ class WebotsObject(object):
                     # need to add an intermediate point if none of the points are already very close from the border
                     newCoordX = (d2y / (d1y + d2y)) * OSMCoord.coordDictionnary[refs[i]].x +\
                         (d1y / (d1y + d2y)) * OSMCoord.coordDictionnary[refs[i - 1]].x
-                    newCoordy = (d2y / (d1y + d2y)) * OSMCoord.coordDictionnary[refs[i]].y +\
+                    newCoordY = (d2y / (d1y + d2y)) * OSMCoord.coordDictionnary[refs[i]].y +\
                         (d1y / (d1y + d2y)) * OSMCoord.coordDictionnary[refs[i - 1]].y
                     newOSMID = OSMCoord.add_new_coord_to_list(newCoordX, newCoordY)
                     newRefs.insert(i, newOSMID)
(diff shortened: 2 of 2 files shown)


CoolSpy3 (approved):
I haven't reviewed the underlying change in-depth, but it makes sense at a high level, so I'm not opposed to merging this in until stb approves the change to be merged upstream. Out of curiosity, does this fix #6909, or is this a different rendering bug?

I do want to verify that Webots builds against this new version of stb (as it looks like there have been some other changes since then), but if so, this is good to merge.

CoolSpy3:
Hmmm. Looks like the Windows test suite is failing (I ran it twice) because Python can't find `Controller.dll`. Tests pass on the other branches, and this seems unrelated to stb, so I'm fine with merging this. Should maybe look into that tho :P

omichel:
> Out of curiosity, does this fix #6909, or is this a different rendering bug?

Yes, I believe it does as I saw exactly the same kind of weird images appearing randomly (in a non reproducible way).

CoolSpy3:
> > Out of curiosity, does this fix #6909, or is this a different rendering bug?
> 
> Yes, I believe it does as I saw exactly the same kind of weird images appearing randomly (in a non reproducible way).

Unfortunately, I tested this and that bug still appears on my machine when building from this branch, so the root cause is probably something else.

I also did some basic playing around with the test suite, and it works for me locally. In CI, it looks like the dll load is using an absolute path to the `Controller` dll, which should definitely be installed and have no weird dependencies, so idk what's going on there. It's possible that the cause lies with the other error (of failing to allocate a tcp port), which could indicate that Webots is failing to free TCP ports, but imo, it would be more likely for the python error to cause that than for that to cause the python error. In summary, for now, I'm stumped.

Again tho, I doubt that it's a problem related to the stb change.

**Update:** I did a clean build and now it looks like the above issue might be fixed 👀 I'll try to do some better testing later to see for sure. Exciting (for me) if true tho :D

**Update 2:** After doing some better testing, this does appear to be fixed! Amazing :D Idk how you even thought to test for a single-platform, optimization-based bug in an external library, but kudos!

omichel:
On my local Windows 11 machine, this PR breaks some tests in the test suite. I am investigating why.

omichel:
> Hmmm. Looks like the Windows test suite is failing (I ran it twice) because Python can't find `Controller.dll`.

It turns out I had this problem in my development environment as well. I propose this [fix](#6933).

omichel:
> Idk how you even thought to test for a single-platform, optimization-based bug in an external library, but kudos!

I found out that the bug was only showing up in release mode, not in debug. The bug was actually showing up with gcc `-O2` or `-O3` and not with `-O1` optimization level. So I tried to compile some files with `-O1` and some with `-O3`, until I isolated the problem to a single file: [WbBackground.cpp](https://github.com/cyberbotics/webots/blob/master/src/webots/nodes/WbBackground.cpp). Inside this file, I used the `#pragma` statement to compile some functions in `-O1` and some others in `O3`. I found that the culprit was the `stb_image.h` include file. In this file, I isolated the function causing problem. I even further split this function in sub-functions to isolate the lines of codes causing the problem. Then, I found out this very strange `goto` statement which was hard to understand (for me and also for gcc) and that gcc was failing to optimize properly. I fixed the code by getting rid of this odd `goto` and that fixed the problem. It took long, but I believe the fix is clean, as it also makes the code of `stb_image.h` more readable and maintainable. However, I agree the bug is fundamentally in gcc, not in `stb_image.h`. But fixing gcc would have taken me even more time.

CoolSpy3:
Wow! Awesome! Thanks for the explanation! I'll keep some of those tricks in mind for the future :eyes:

PR #6929, opened 2026-02-06
Fixed binaries names for lrelease and lupdate on recent MSYS64

omichel:
On January 6th 2026, Qt6 became the default in MSYS64, see https://github.com/msys2/MINGW-packages/commits/master/mingw-w64-qt6-tools
As a consequence the `lupdate-qt6.exe` and `lrelease-qt6.exe` were renamed to `lupdate.exe` and `lrelease.exe`, thus breaking the build on recent MSYS64 system.

However, in the GitHub CI, we still have the old version where these binary have the `-qt6` suffix. So, we need to distinguish the CI case from the normal up-to-date development environment.
Hopefully this difference will disappear soon, when MSYS64 is updated in the GitHub CI.

### DIFF
--- a/src/webots/nodes/WbBackground.cpp
+++ b/src/webots/nodes/WbBackground.cpp
@@ -638,12 +638,11 @@ void WbBackground::applySkyBoxToWren() {
     } else {  // otherwise, use a small uniform texture with the color of the sky
       cm = wr_texture_cubemap_new();
       size = 2;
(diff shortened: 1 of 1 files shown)


CoolSpy3 (approved):
Ah, the joys of GitHub Actions versioning 🙃 I flagged this for a build to verify that this does, in fact, pass CI, but if that succeeds, this is good.

PR #6941, opened 2026-02-16
Display the number of warnings for each tested world

omichel:
This PR aims at making clearer how many warnings are displayed for each tested world (if any).


CoolSpy3 (approved):
:+1:

PR #6945, opened 2026-02-20
Fix signal handler restoration on exit

### LINKED ISSUE #6944: Webots bypasses python's SIGINT behavior
**Describe the Bug**
Because `quit_controller` gets registered as the signal handler for `SIGINT` it becomes impossible to use normal cleanup code for a python process (and presumably similar in other languages).

**Steps to Reproduce**
1. Write a python script using any controller that tries to do cleanup using `try ... finally` or `atexit`
2. Run the script
3. Try to end it with CTRL+C
4. The script exits without executing the finally.

**Expected behavior**
Webots should not be fully overwriting the signal handler, or at the very least provide an option to not do that. I am honestly not even sure what the point of this code is. The only interesting detail is that a global variable gets set that however AFAICT is not used anywhere.

If this code is for some reason necessary it would be great if this could at least be documented clearly.

**System**
 - Operating System: Linux Ubuntu 24.04
 - Graphics Card: Irrelevant

**Additional context**
Luckily the init code is only executed once and it's possible to overwrite the signal handler back to the python default via `signal.signal(signal.SIGINT, signal.default_int_handler)`, so that's a functional workaround for now. This works perfectly, it doesn't appear that the controller API needs any cleanup done that is disabled by this.

omichel:
Attempt to fix #6944.

### DIFF
--- a/docs/reference/changelog-r2025.md
+++ b/docs/reference/changelog-r2025.md
@@ -25,6 +25,7 @@
     - Fixed Python controllers on Windows ([#6933](https://github.com/cyberbotics/webots/pull/6933)).
     - OSM importer no longer crashes when run in 3d mode ([#6935](https://github.com/cyberbotics/webots/pull/6935)).
     - Fixed Java compilation deprecation warning and run-time warning ([#6936](https://github.com/cyberbotics/webots/pull/6936)).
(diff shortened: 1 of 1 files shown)


CoolSpy3 (approved):
Looks good! Thank you :D I threw on the distribution test tag so the issue submitter can test without having to figure out how to build from source. I'll try to remember to post a link to the download after it finishes :P

**Update:** It's [here](https://github.com/cyberbotics/webots/actions/runs/22216120696?pr=6945) under "Artifacts".

Issue #6976, opened 2026-06-10
macOS 27 Beta: Webots R2025a falsely detects all TCP ports as already in use


shuminzxr:
# macOS 27 Beta: Webots R2025a fails to start due to false TCP port conflict detection

## Environment

- Webots Version: R2025a & R2025b
- Operating System: macOS 27 Beta (26A5353q)
- CPU: Apple Silicon (M1 Max)
- Architecture: ARM64 Native
- Qt Version: bundled Qt 6.5.3

---

## Summary

After upgrading to macOS 27 Beta, Webots R2025a could no longer start correctly.

Two independent issues were observed:

1. Qt platform plugin detection failure (cocoa plugin not found).
2. False detection of TCP port conflicts, causing Webots to terminate immediately during startup.

After investigating the source code and rebuilding Webots from source, I found that the TCP port conflict detection logic appears to behave incorrectly on macOS 27 Beta.

Removing this pre-check allows Webots to launch and function normally.

---

## Initial Startup Failure

The first startup attempt produced a Qt platform plugin error similar to:

text Could not find the Qt platform plugin "cocoa" in "" This application failed to start because no Qt platform plugin could be initialized. 

A temporary workaround was:

bash QT_QPA_PLATFORM_PLUGIN_PATH=/Applications/Webots.app/Contents/lib/webots/qt/plugins/platforms 

After manually setting the plugin path, Webots was able to proceed further during startup.

---

## Main Failure

After bypassing the Qt plugin issue, Webots terminated with messages indicating that every TCP port in its startup range was already in use.

Example:

text Could not listen to extern controllers and robot windows on port 1234. Port 1234 is already in use. Giving up.  Could not listen to extern controllers and robot windows on port 1235. Port 1235 is already in use. Giving up.  ...  failed to open TCP server in the port range [1234-1244] 

The same behavior occurred when specifying completely different port ranges:

bash webots --port=20000 webots --port=30000 

All ports were reported as already occupied.

---

## Verification Performed

I verified that these ports were not actually in use.

### Check 1: Port Scan

bash lsof -iTCP:1234-1244 

No listening processes were found.

### Check 2: IPv4 Socket Binding

Python test:

python import socket  s = socket.socket() s.bind(("127.0.0.1", 20000))  print("OK") 

Result:

text OK 

### Check 3: IPv6 Socket Binding

Python test:

python import socket  s = socket.socket(socket.AF_INET6) s.bind(("::", 20000))  print("OK") 

Result:

[...]

PR #6978, opened 2026-06-10
Work around macOS 27 startup issues

shuminzxr:
This PR is related to issue https://github.com/cyberbotics/webots/issues/6976.

It contains two small changes that allowed Webots R2025a to build and run on macOS 27 Beta:

In WbTcpServer.cpp, the localhost TCP pre-check is disabled because it falsely reports all ports as already in use on macOS 27 Beta.
In g_image.c, stb_image_write.h is included with quotes instead of angle brackets to satisfy the newer macOS/Clang include behavior.
This has been tested by rebuilding Webots from source on Apple Silicon M1 Max. The patched build launches correctly, loads worlds, and runs controllers.

I understand that the TCP change may be considered a workaround rather than the final upstream fix. I am submitting this PR mainly to make the tested patch visible and easier to review.

### DIFF
--- a/src/webots/gui/WbTcpServer.cpp
+++ b/src/webots/gui/WbTcpServer.cpp
@@ -130,24 +130,37 @@ void WbTcpServer::create(int port) {
   // - a websocket on "/"
   // - texture images on the other urls. e.g. "/textures/dir/image.[jpg|png|hdr]"
 
-  // See if a server is already running on port by trying to connect to it.
-  // This is needed because in some environments QTcpServer::listen() uses a socket that is configured to reuse a port [1]
-  // and Qt does not provide a way to configure the socket before calling listen() [2].
-  // [1] https://doc.qt.io/qt-6/qabstractsocket.html#BindFlag-enum
-  // [2] https://stackoverflow.com/questions/47268023/how-to-set-so-reuseaddr-on-the-socket-used-by-qtcpserver
+  // =====================================================
+  // TEMPORARY PATCH FOR macOS 27
+  // -----------------------------------------------------
+  // Webots R2025a performs a pre-check by connecting to
+  // localhost:port and assumes the port is already in use
+  // if the connection succeeds.
+  //
+  // On macOS 27 this check appears to incorrectly report
+  // every port as occupied.
+  //
+  // We disable the pre-check and rely on the real
+  // QTcpServer::listen() call below.
+  // =====================================================
+
+  /*
   QTcpSocket socket;
   socket.connectToHost("localhost", port);
   if (socket.waitForConnected(100)) {
     socket.disconnectFromHost();
     throw tr("Port %1 is already in use").arg(port);
   }
+  */
 
   // Reference to let live QTcpSocket and QWebSocketServer on the same port using `QWebSocketServer::handleConnection()`:
   // - https://bugreports.qt.io/browse/QTBUG-54276
   mWebSocketServer = new QWebSocketServer("Webots Streaming Server", QWebSocketServer::NonSecureMode, this);
   mTcpServer = new QTcpServer();
+
   if (!mTcpServer->listen(QHostAddress::Any, port))
     throw tr("Cannot set the server in listen mode: %1").arg(mTcpServer->errorString());
+
   connect(mWebSocketServer, &QWebSocketServer::newConnection, this, &WbTcpServer::onNewWebSocketConnection);
   connect(mTcpServer, &QTcpServer::newConnection, this, &WbTcpServer::onNewTcpConnection);
   connect(WbSimulationState::instance(), &WbSimulationState::controllerReadRequestsCompleted, this,
--- a/src/controller/c/g_image.c
+++ b/src/controller/c/g_image.c
@@ -25,7 +25,7 @@
 #define STB_IMAGE_WRITE_IMPLEMENTATION
 #define STBIW_WINDOWS_UTF8
 #include <stb_image.h>
-#include <stb_image_write.h>
+#include "stb_image_write.h"
 
 #include <webots/types.h>
 


src/webots/gui/WbTcpServer.cpp
```diff
@@ -130,24 +130,37 @@ void WbTcpServer::create(int port) {
...
+  // =====================================================
+
+  /*
   QTcpSocket socket;
   socket.connectToHost("localhost", port);
   if (socket.waitForConnected(100)) {
     socket.disconnectFromHost();
     throw tr("Port %1 is already in use").arg(port);
   }
+  */
```
omichel, line 154: We should fix this in a different way. The test should be performed on other platforms than macOS. Also, on macOS, we should find another way to check if the port is already in use.

omichel (changes requested):

src/webots/gui/WbTcpServer.cpp
shuminzxr: Thank you for the review.

I agree that simply disabling the check is probably not the best long-term solution and that a platform-specific or more robust implementation would be preferable.

My intention with this patch was mainly to identify the source of the issue and verify that the failure originates from the current localhost connection test.

On macOS 27 Beta, commenting out this pre-check consistently allows Webots to start and operate normally, while the original code reports every tested port as already in use, even when the ports are confirmed to be free.

From my testing, this suggests that the problem is likely located in the current localhost probe rather than in QTcpServer::listen() itself.

I completely agree that the final fix should preserve the port availability check and remain compatible with other platforms.

In the meantime, this workaround may still be useful for affected users running macOS 27 Beta, as it provides a temporary way to bypass what is currently a startup-blocking issue until a more complete solution is implemented.

If it would be useful, I can perform additional tests on macOS 27 Beta and help evaluate alternative approaches for checking port availability on macOS.

shuminzxr: Would a platform-specific approach be acceptable?

For example, we could keep the existing localhost pre-check on non-macOS platforms, but skip it on macOS and rely directly on QTcpServer::listen() there. This would preserve the current behavior on Linux/Windows while avoiding the false-positive localhost connection behavior observed on macOS 27 Beta.

Conceptually:

#ifndef __APPLE__
  QTcpSocket socket;
  socket.connectToHost("localhost", port);
  if (socket.waitForConnected(100)) {
    socket.disconnectFromHost();
    throw tr("Port %1 is already in use").arg(port);
  }
#endif

mTcpServer = new QTcpServer();
if (!mTcpServer->listen(QHostAddress::Any, port))
  throw tr("Cannot set the server in listen mode: %1").arg(mTcpServer->errorString());

This would make the workaround macOS-specific instead of removing the pre-check globally.

CoolSpy3: I was investigating this, and I'm now wondering if the check is even necessary. The comment references [`QAbstractSocket::BindMode`](https://doc.qt.io/qt-6/qabstractsocket.html#BindFlag-enum). Notably the only flag that that claims to set is [`SO_REUSEADDR`](https://man7.org/linux/man-pages/man7/socket.7.html#:~:text=SO_REUSEADDR), which looks like it still blocks connections on ports with active servers. The real worry is setting [`SO_REUSEPORT`](https://man7.org/linux/man-pages/man7/socket.7.html#:~:text=SO_REUSEPORT), which *would* allow that behavior. So, just going by that, we should be fine. Nevertheless, I had some doubts, so here's my in-depth analysis:
* The default bind behavior only sets the `AddressReusable` socket engine option on UNIX [[1]](https://github.com/qt/qtbase/blob/79c5df9a97d8d9d739b9eee3d7b76015107f69a1/src/network/socket/qabstractsocket.cpp#L1539-L1544)
* Same goes for `QTcpServer` [[2]](https://github.com/qt/qtbase/blob/79c5df9a97d8d9d739b9eee3d7b76015107f69a1/src/network/socket/qtcpserver.cpp#L155-L168)
* This option typically only corresponds to `SO_REUSEADDR` [[3]](https://github.com/qt/qtbase/blob/79c5df9a97d8d9d739b9eee3d7b76015107f69a1/src/network/socket/qnativesocketengine_unix.cpp#L154-L155) (called [here](https://github.com/qt/qtbase/blob/79c5df9a97d8d9d739b9eee3d7b76015107f69a1/src/network/socket/qnativesocketengine_unix.cpp#L486))
* On Mac, this can sometimes set `SO_REUSEPORT`, but only on UDP, which doesn't matter here [[4]](https://github.com/qt/qtbase/blob/79c5df9a97d8d9d739b9eee3d7b76015107f69a1/src/network/socket/qnativesocketengine_unix.cpp#L487-L495)

On Windows, there are `SO_REUSEADDR` and `SO_EXCLUSIVEADDRUSE`. If the former is set, we can get multiple servers on the same socket. If the latter is set (admin required), it prevents use of the former. [[5]](https://learn.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse)
* The default bind behavior can set `AddressReusable` and/or `BindExclusively` based on the flags. [[6]](https://github.com/qt/qtbase/blob/79c5df9a97d8d9d739b9eee3d7b76015107f69a1/src/network/socket/qabstractsocket.cpp#L1545-L1554)
* These correspond to the two windows flags [[7]](https://github.com/qt/qtbase/blob/79c5df9a97d8d9d739b9eee3d7b76015107f69a1/src/network/socket/qnativesocketengine_win.cpp#L177-L182) (called [here](https://github.com/qt/qtbase/blob/79c5df9a97d8d9d739b9eee3d7b76015107f69a1/src/network/socket/qnativesocketengine_win.cpp#L531))
* `QTcpServer` sets neither explicitly, so if we run by the `BindMode` defaults (see first link), we'd expect `ShareAddress`, which corresponds to both flags being unset. (see [6])

All of that is to say, just looking at the code, I don't think Qt sets any options which would cause the behavior this check is trying to prevent, so we could try starting the TCP server first and checking the error code. If it fails with an in-use conflict, return the "in use" error. Otherwise, fall back to the generic version. Of course, we should probably check that behavior before merging it in.

Also, is it just me, or does `create` have a memory leak on the second throw whereby `mWebSocketServer` and `mTcpServer` are never deleted? `start()` looks like it just recursively calls `create` on the same `WbTcpServer` object, so the destructor is never called to delete the intermittent members.

CoolSpy3: (I am still curious as to why the current behavior is failing on mac though 👀 )

shuminzxr: "only apple can do" lol

shuminzxr: I think that makes sense. My original intention was mainly to identify the source of the startup failure and provide a temporary workaround. Using the result of QTcpServer::listen() as the source of truth seems cleaner than relying on a localhost connection probe. From my testing on macOS 27 Beta, removing the pre-check consistently restores normal startup behavior, which would also be consistent with the idea that the check itself may no longer be necessary.

omichel: > Also, is it just me, or does create have a memory leak on the second throw whereby mWebSocketServer and mTcpServer are never deleted? start() looks like it just recursively calls create on the same WbTcpServer object, so the destructor is never called to delete the intermittent members.

Yes, I believe you are right. The minimal fix is to clean up before throwing (and ideally parent mTcpServer too for symmetry):
```c++
mWebSocketServer = new QWebSocketServer("Webots Streaming Server", QWebSocketServer::NonSecureMode, this);
mTcpServer = new QTcpServer(this);
if (!mTcpServer->listen(QHostAddress::Any, port)) {
  const QString error = mTcpServer->errorString();
  delete mTcpServer;       mTcpServer = NULL;
  delete mWebSocketServer; mWebSocketServer = NULL;
  throw tr("Cannot set the server in listen mode: %1").arg(error);
}
```
This also keeps `isActive()` `(mWebSocketServer != NULL)` honest after a failed attempt.

shuminzxr: That makes sense.

I hadn't noticed the cleanup issue, but I agree that deleting the server objects before throwing would avoid leaving the object in a partially initialized state and would keep isActive() consistent after a failed startup attempt.

It also seems compatible with using QTcpServer::listen() as the primary source of truth for port availability.

shuminzxr: Would you like me to update the PR in this direction?

I can replace the current workaround with a cleaner implementation that removes the localhost pre-check, relies on QTcpServer::listen() as the source of truth, and cleans up mTcpServer / mWebSocketServer before throwing if listen() fails.

I’m actually an MSc student rather than a regular contributor to the Webots codebase, so I may not be fully familiar with all the design considerations here. If this direction is consistent with the intended architecture, I’d be happy to implement and test it on macOS 27 Beta lol

CoolSpy3: Yes; that would be awesome! Please go-ahead make those changes!

PR #6980, opened 2026-06-11
Use C23 bool

ShuffleWire:
**Description**
I'm using the Webots API using a custom Makefile to integrate into a project which is build using C23 standart.
But if I end up with 2 definition of bool, depending of the include I use in different c files. So I have conflict between "native' bool and Webots bool which is char.

**Additional context**
To get is right, and because I'm using gcc 14, I've found a defect, as indicated into the code as comment, but you can find disccution about that here : https://stackoverflow.com/questions/78581920/what-is-the-stdc-version-value-for-c23/78582932#78582932

### DIFF
--- a/include/controller/c/webots/types.h
+++ b/include/controller/c/webots/types.h
@@ -48,21 +48,7 @@ typedef struct WbFieldStructPrivate *WbFieldRef;
 #define INFINITY (1.0 / 0.0)
 #endif
 
-#ifndef bool
-#define bool char
-#endif
-
-#ifndef true
-// clang-format off
-#define true ((bool)1)
-// clang-format on
-#endif
-
-#ifndef false
-// clang-format off
-#define false ((bool)0)
-// clang-format on
-#endif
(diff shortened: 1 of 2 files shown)


CoolSpy3:
Rather than doing this, does it make sense to just include `stdbool.h` and let gcc define everything as-appropriate?

ShuffleWire:
you are right, and stdbool.h have this line,
`#if defined(__STDC_VERSION__) && __STDC_VERSION__ > 201710L`
so....
funny enough it look like they know that gcc 14 define c23 as 202000L otherwise they would have went with the proper < 202311L

omichel:
@ShuffleWire: will you modify this PR to use `stdbool.h` instead?

ShuffleWire:
yes sorry ! it's done

CoolSpy3 (approved):
Thank you! (The build failure should be addressed by #6985)

omichel (approved):
Thank you.

PR #6988, opened 2026-06-28
Remove Unnecessary `termio.h` Include

### LINKED ISSUE #6982: unable to find termio.h (fixed by renaming -> termios.h)
**Describe the Bug**
During compilation, can't find `termio.h` on Fedora 44. I was able to resolve this by changing it to `termios.h`, though I guess we'll find out if there are any long-term implications there.

**Steps to Reproduce**
Follow install instructions at https://github.com/cyberbotics/webots/wiki/Linux-installation/ up to "Build Webots".

**Expected behavior**
Compile finishes

Instead:

```
# compiling Serial.cpp
Serial.cpp:48:10: fatal error: termio.h: No such file or directory
   48 | #include <termio.h>
      |          ^~~~~~~~~~
compilation terminated.
```

**System**
 - Operating System: Fedora 44
 - Graphics Card: AMD Radeon™ 840M (Framework laptop)
  
**Additional context**
I was able to solve this - at least for compilation - by:

```
find . -name "Serial.cpp" | xargs grep -l "termio.h"
vim ./projects/robots/gctronic/e-puck/plugins/remote_controls/e-puck_bluetooth/Serial.cpp
```

... and change line 48 from `#include <termio.h>` to `#include <termios.h>`. I guess I'll find out later if that breaks things but it got compilation to finish so hey, I'll take it.

CoolSpy3:
**Description**
As far as I can tell, `termio.h` has been superseded by `termios.h`. Newer versions of glibc have stopped including this library altogether, so removing this include is necessary to build on Ubuntu 26 and Fedora 44 (which include this newer version). Additionally, glibc versions going back as far as Ubuntu 22 appear to just define termio.h as
```c
#include <termios.h>
#include <sys/ioctl.h>
```
[ioctl only defines `ioctl()`](https://www.man7.org/linux/man-pages/man2/ioctl.2.html), which is unused in this file. Thus, this include is equivalent to only including `termios`, which is already done above, so this change should not impact functionally on any currently-supported systems.

**Related Issues**
This pull-request fixes issue #6982

### DIFF
--- a/projects/robots/gctronic/e-puck/plugins/remote_controls/e-puck_bluetooth/Serial.cpp
+++ b/projects/robots/gctronic/e-puck/plugins/remote_controls/e-puck_bluetooth/Serial.cpp
@@ -43,10 +43,6 @@
 #include <IOKit/serial/IOSerialKeys.h>
 #include <IOKit/serial/ioss.h>
 #include <sys/param.h>
-
-#else  // __linux__
-#include <termio.h>
-
 #endif
 #endif
 


omichel (approved):
This makes senses. Thank you!

PR #6994, opened 2026-07-03
Fix MacOS Sources Test

CoolSpy3:
**Description**
The macos sources test was complaining about the custom brews we're using to pin clang and cppcheck being untrusted. Not sure why this is showing up now given that they were working earlier this week, and I don't see a recent runner update, but this PR adds the command to trust them. While I was in there, I also swapped those sections to use multiline strings for readability. I'll revert if it breaks anything.


omichel (approved):
Excellent, thank you!

Issue #6414, opened 2023-10-14
Crash (segfault) when boundingObject Group contains Solid with physics


PeterJCLaw:
**Describe the Bug**

Launching webots R2023b with the following world results in some warnings and a segfault. This world is a cut-down. I hit this while developing the larger world, which doesn't normally have warnings but did in the failure mode.

The warnings in the original world before the cut-down was just:
> Pose: A Pose node inside a 'boundingObject' can only contain one Shape or one Geometry node. The child node is ignored.

and then a line like:
```
/usr/local/bin/webots: line 105: 33888 Segmentation fault      (core dumped) "$webots_home/bin/webots-bin" "$@"
```

The warnings observed in the cut-down are different, so I'm not completely sure if the cause is the same.

**Steps to Reproduce**

``` console
$ webots --stdout --stderr --mode=pause $PWD/worlds/Arena.wbt
```

`Arena.wbt`:
``` python
#VRML_SIM R2022b utf8

WorldInfo {
  basicTimeStep 8
}
Viewpoint {}

Solid {
  name "outer"
  children [
    DEF BODY Group {
      children [
        Solid {
          name "inner"
          physics Physics {}
        }
      ]
    }
  ]
  boundingObject USE BODY
  physics Physics {}
}
```

**Expected behavior**

For other invalid worlds webots seems to emit warnings and refuse to run a simulation, but not crash. Matching that would be fine.

**System**
- webots: R2023b, other versions untested
 - Operating System: Ubuntu 22.04.3 LTS
 - Graphics Card: NVIDIA Corporation GP107 [GeForce GTX 1050 Ti]

ShuffleWire:
I don't know what your use case is, but I've been playing lately with Solid into Solid (especially with physic for each of them), and in the end, even if the simulation "could" work on smaller system, it doesn't scaled well and give many weird behavior (the internal solver seems to struggle to follows relationship and so on...)
I would advice (and it's what I've done, and happy with that) to stick to the Solid -> Joint -> Solid architecture, which Webots is much happier to work with.
Anyway, Webots should not crash, but in the end even it didn't I would not be confident in my simulation setup.
My two cents

ShuffleWire:
An assertion if falling 
```
#5  0x00007fb24a053eb2 in __GI___assert_fail
[...]

PR #6997, opened 2026-07-05
Validate USE Nodes

CoolSpy3:
**Description**
The linked issue was caused (in part) because while the scene tree is validated as it's being loaded, the validation check explicitly skips USE nodes. In my analysis of the code, I wasn't able to find a good reason for this [1], so this PR simply removes that check. (Let this be a lesson to everyone to *comment their code*.) Normally, this is caught once the DEF dictionary is updated after the whole tree is loaded, but the inclusion of Physics nodes causes a warning check to be triggered before that can run. The check assumes that the USE node is either "real" or not, but because its in a quasi (i.e. illegal) state, this assumption fails. crashing Webots.

Some quick notes on other options that I considered:
It would also be possible to move that check that's causing this crash. Unfortunately, it's triggered by post-finalization, so that would involve either wiring a new check event (which seems messy) or moving the post finalization routine. The latter can't be done easily because the two triggers occur in different translation units, and I don't really want to go messing that much with the world-load-order semantics.

We could also opt to just add an edge case to that check that explicitly checks for the broken state, but I think this is a more-complete solution as it catches any other issues that are being skipped by this USEe-skipping behavior.

[1] Maybe the code just assumes that USE nodes are valid because their DEF nodes are already initialized? I was also thinking that maybe it's to prevent weird interactions where USEd fields are still linked to the DEF nodes, but everything looks cleanly-enough separated that I don't think that should be a problem, so I'm at a loss.

FWIW, the USE node is created via deep-copy of the original DEF node, so deletes to its members by `validate` don't propagate to the original node. This *could* lead to a state where the two nodes become out-of-sync with the USE node structure being inaccessible in the Scene Tree because it's still linked to the DEF node. However, in at least the provided test case, the code already catches that in `WbDictionary::updateDef` and converts the node into a DEF. (Simply because it's an invalid insert, not specifically because of the desync.) I can't verify that this will work in all edge cases, but we can hope 🤞 In any case, I still think this is preferable to Webots entering an invalid state because the check is skipped.

(Side note, while writing this, I caught a similar 'ignore use nodes' check in `WbDictionary::updateProtosDef`. This one explicitly says that it's skipping because the nodes should've already been validated by the original DEF. I think it might make more sense here because in the nested case, context is better captured. In order to break it you'd have to *at a minimum* have a proto that loads a DEF containing a USE to a node that's invalid in the context of the new USE but not the original DEF. In any case, I'll leave it for now as it doesn't seem to be causing problems.) ~~[Maybe I'll test it later.]~~

**Related Issues**
This pull-request fixes issue #6414

### DIFF
--- a/.github/workflows/test_suite_linux.yml
+++ b/.github/workflows/test_suite_linux.yml
@@ -162,7 +162,7 @@ jobs:
         if [[ "$(diff -qr artifact/ untouched-artifact/ | wc -l)" -ne "0" ]]; then echo Some world and/or wbproj files are not up to date: "$(diff -qr artifact/ untouched-artifact/)"; exit 1; fi
   delete-artifacts:
     needs: [build, test-suite, test-worlds]
-    if: ${{ always() && needs.test-suite.result != 'failure' && needs.test-worlds.result != 'failure' && !contains(github.event.pull_request.labels.*.name, 'test distribution')  && !contains(github.event.pull_request.labels.*.name, 'test webots build') }}
+    if: ${{ always() && !contains(github.event.pull_request.labels.*.name, 'test distribution')  && !contains(github.event.pull_request.labels.*.name, 'test webots build') }}
     strategy:
       matrix:
         os: [ubuntu-22.04, ubuntu-24.04]
--- a/.github/workflows/test_suite_linux.yml
+++ b/.github/workflows/test_suite_linux.yml
@@ -162,7 +162,7 @@ jobs:
         if [[ "$(diff -qr artifact/ untouched-artifact/ | wc -l)" -ne "0" ]]; then echo Some world and/or wbproj files are not up to date: "$(diff -qr artifact/ untouched-artifact/)"; exit 1; fi
   delete-artifacts:
     needs: [build, test-suite, test-worlds]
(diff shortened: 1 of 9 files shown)


CoolSpy3:
Update: I tested the proto case with
```js
#VRML_SIM R2025a utf8

PROTO BadProto6997 [
]
{
	Solid {
		children [
			DEF BadSolid Solid {
				physics Physics{}
			}
			DEF BadGroup Group {
				children [
					USE BadSolid
				]
			}
		]
		boundingObject USE BadGroup
		physics Physics{}
	}
}
```
and it appears to be caught by this fix. After a closer inspection of that block, it also looks (by my quick glance) like it's just doing some more-primitive type checking, so I think we're good!

CoolSpy3:
Between the two runs, all windows tests have passed.

omichel:
@copilot: Your patch [a49f905](https://github.com/cyberbotics/webots/pull/6997/commits/a49f905d524fbfda82d02abd91b140e5e7d50452) has nothing to do with this PR, please implement it as a separate PR targeting the master branch.

omichel:
@copilot: Please revert also this commit [a49f905](https://github.com/cyberbotics/webots/pull/6997/commits/a49f905d524fbfda82d02abd91b140e5e7d50452) from this PR.

omichel (approved):
Thank you for the detailed explanations and proposed fix. It looks very good to me.

PR #7000, opened 2026-07-16
Add Missing `QActionGroup` Creation

CoolSpy3:
**Description**
Someone forgot to create a new `QActionGroup`, so the "Follow Object" buttons got put in the same group as the "Rendering Mode" options, so only one could be selected at a time. From what I can see though, this didn't trigger any events, so the only impacts were that some of the buttons were visually unchecked while they were active.

Also, Happy 7000th Issue/PR 🎉 🥳 🚀


omichel (approved):
Well spotted!
Thank you.

PR #7002, opened 2026-07-18
Reference the howto/docker sample in the documentation

### LINKED ISSUE #5588: `projects/samples/howto/docker` not referenced anywhere in the documentation
Recently it was added the `projects/samples/howto/docker` folder.
However, this is not visible from the Open Sample World dialog and it is not referenced in the documentation so it is invisible to users.
It should be added in the list of samples:
https://www.cyberbotics.com/doc/guide/samples-howto

or moved in a different location

ismetvahidzade986-sketch:
`projects/samples/howto/docker` wasn't linked anywhere in the docs (#5588).

This adds an entry to the "How to" sample list describing it — running Webots and an extern controller in separate Docker containers, which is the setup used to isolate competitors' controllers on a competition server — and links to the folder and its README. I placed it alphabetically and left out a thumbnail since the sample doesn't have one.

Fixes #5588


CoolSpy3 (changes requested):
Thanks for looking into this! This looks good to me, just two notes.

1\)
It looks like the doc tests are failing because the new title 'docker' isn't capitalized. It doesn't trip on the other names because they refer to world files. The relevant file here is `.../docker/simulation/worlds/camera.wbt`. In-keeping with the naming scheme throughout the rest of this file would have us name this `camera.wbt`, which I don't think makes much sense.

Imo, the best option here is to rename the world to `docker.wbt` (and link to it) and the controller to `docker.c` (making sure that any references to those in the surrounding files also get updated) and then we can note in the linked README that the world is a copy of the camera device sample. So if that sounds okay, could you please make those changes.

2\)
Could you include a sample image to line up with the formatting on the other worlds. I know the sample doesn't provide one, but you should be able to just take a screenshot of the world and crop it down. Or if you want, I can do it.

Also, @omichel could you skim through the text and make sure the semantics line up. It looks reasonable to me, but since you contributed the original sample, you might catch something I missed.

Lastly, for anyone reviewing, [here](https://www.cyberbotics.com/doc/guide/samples-howto?version=ismetvahidzade986-sketch:docs-reference-docker-sample)'s the link to the updated doc page.

ismetvahidzade986-sketch:
Done, pushed the rename. The world is now `docker.wbt` and the controller is `docker` (`docker.c`), with the launchers, the controller Dockerfile, `.gitignore`, and the README all updated to match. I also added a line in the README noting the world is a copy of the camera device sample. The doc entry now links the world file, so `test-doc` should go green.

On the sample image: I don't have a working Webots + Docker setup to capture a clean shot of the world running, so if that offer still stands I'd appreciate you grabbing it. If you'd rather I do it, point me at how you usually capture the sample images and I'll sort it out.

CoolSpy3 (dismissed):
Awesome! Thank you! This all looks good to me! I went ahead and added the image as well.

CoolSpy3 (approved):

omichel (approved):
That looks all good to me.
Thank you!

Issue #6800, opened 2025-03-28
App crushed when set 'w' equal to -0.5 in  Quaternions mode


peterwarlg:
**Describe the Bug**
I add an GPS in Spot robot's middleExtension, and change rotation mode to Quaternions, once i set 'w' to any negative number and confirmed, the App crushed.

![Image](https://github.com/user-attachments/assets/079e0402-1ba0-44b5-837f-74e4b1bd651c)

**Steps to Reproduce**
1. add a Spot Robot in an empty world
2. add a GPS node in Spot->middleExtesion (i'm sure other nodes such as imu and Gyro have the same problem)
3. change GPS's rotation mode to Quatersions, and make sure w, x, y, z is zero as default
4. set 'w' to any minus number,  click anywhere else to confirm
5. app crush

**Expected behavior**
I know that setting ’w‘ to a minus number is an illegal and incrrect operation, however i suppose webots can prevent those bad operations.

**Screenshots**
see picture abow

**System**

 - Operating System: Windows 10
 - Graphics Card: NVIDIA GeForce GTX 1650
 - software version: R2025a

lukicdarkoo:
I confirm this is a bug

PR #7009, opened 2026-08-24
Fix negative identity quaternion conversion

tandede:
**Description**

Entering a quaternion such as `(-0.5, 0, 0, 0)` in the rotation editor normalizes it to `(-1, 0, 0, 0)`. This is a valid representation of the identity rotation, but `WbRotation::fromQuaternion()` previously converted its scalar component to an angle of `2π` and then attempted to normalize its zero-length vector component. The resulting division by zero produced NaN axis values that could later crash the scene tree or rendering path.

This change computes the quaternion vector length once and treats a normalized quaternion with no usable vector component as the identity rotation. Both positive and negative identity quaternions therefore produce the existing finite canonical representation `(0, 0, 1, 0)`. Non-degenerate negative-scalar quaternions continue through the existing axis-normalization path and retain their original rotation angle.

**Related Issues**

This pull request fixes issue #6800.

**Tasks**

- [x] Handle negative identity quaternions without dividing by zero.
- [x] Preserve the existing behavior for non-degenerate quaternions.
- [x] Update the changelog.
- [x] Validate the conversion with degenerate and non-degenerate inputs.
- [x] Run the project source-format and static-analysis checks.

**Documentation**

No user-facing documentation change is required. The fix is recorded in `docs/reference/changelog-r2025.md`.

**Screenshots**

Not applicable. The change prevents invalid numerical state rather than changing the expected visual output.

**Additional context**

Before the fix, a standalone C++ reproduction for `(-0.5, 0, 0, 0)` produced:

```text
nan nan nan 6.28319
```

After the fix, the regression probe covers:

- `(-0.5, 0, 0, 0)`
- `(-1, 1e-16, 0, 0)`
- `(1, 0, 0, 0)`
- a non-degenerate negative-scalar quaternion

All cases produce finite results, the three identity inputs use the canonical identity rotation, and the non-degenerate case retains its expected axis and `1.5π` angle.

Validation completed:

- standalone C++ regression probe with AddressSanitizer and UndefinedBehaviorSanitizer
- ClangFormat 14.0.6, matching the source CI major version
- repository `TestCppCheck` with Cppcheck 2.14.2
- `git diff --check`

### DIFF
--- a/docs/reference/changelog-r2025.md
+++ b/docs/reference/changelog-r2025.md
@@ -35,6 +35,7 @@
     - USE nodes are now validated before being inserted into the Scene Tree. This fixes some crashes when loading invalid world files ([#6997](https://github.com/cyberbotics/webots/pull/6997)).
     - Fixed a bug causing the "Plain/Wireframe Rendering" and "Follow Object > ..." buttons to incorrectly be shown as unchecked in certain circumstances ([#7000](https://github.com/cyberbotics/webots/pull/7000)).
     - SVG files are now served with the `image/svg+xml` MIME type instead of being rejected as an unsupported file type, so robot windows can use SVG images ([#7007](https://github.com/cyberbotics/webots/pull/7007)).
+    - Fixed a crash when entering a quaternion with a negative scalar and a zero vector in the rotation editor ([#6800](https://github.com/cyberbotics/webots/issues/6800)).
 
 ## Webots R2025a
 Released on January 31st, 2025.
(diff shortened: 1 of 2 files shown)


docs/reference/changelog-r2025.md
```diff
@@ -35,6 +35,7 @@
     - USE nodes are now validated before being inserted into the Scene Tree. This fixes some crashes when loading invalid world files ([#6997](https://github.com/cyberbotics/webots/pull/6997)).
     - Fixed a bug causing the "Plain/Wireframe Rendering" and "Follow Object > ..." buttons to incorrectly be shown as unchecked in certain circumstances ([#7000](https://github.com/cyberbotics/webots/pull/7000)).
     - SVG files are now served with the `image/svg+xml` MIME type instead of being rejected as an unsupported file type, so robot windows can use SVG images ([#7007](https://github.com/cyberbotics/webots/pull/7007)).
+    - Fixed a crash when entering a quaternion with a negative scalar and a zero vector in the rotation editor ([#6800](https://github.com/cyberbotics/webots/issues/6800)).
```
omichel:
```suggestion
    - Fixed a crash when entering a quaternion with a negative scalar and a zero vector in the rotation editor ([#7009](https://github.com/cyberbotics/webots/pull/7009)).
```

omichel (changes requested):
Thank you.
It looks good to me, except for the reference in the change log.

docs/reference/changelog-r2025.md
tandede: Thanks, fixed. The changelog entry now references this pull request, #7009, instead of the original issue.

omichel (approved):
Thank you!

Issue #3168, opened 2021-06-12
Timestep in the default controler doesn't match the default timestep world


ShuffleWire:
**Describe the Bug**
When creating an new world, and adding the default **C** or **MATLAB** controller using the Wizard "New Robot Controller", timestep (=64) don't match with the new created world (=32). 
A better way to do is like in others default controller (C++, Python, Java) to use `getBasicTimeStep` .
I've no way to test for MATLAB a modification, so I let someone take it in charge.

**Steps to Reproduce**
1. Create a new world
2. See that basicTimeStep is 32
3. Create a new Matlab controller
4. See on line 13 `TIME_STEP = 64;`
5. Create a new C controller
6. See on line 18 `#define TIME_STEP 64`

**Expected behavior**
Default timestep should match. Use of `getBasicTimeStep` should be nice.

**System**
 - Operating System: Debian 11
 - Webots from master c3c2f47a39f0ac4eeffb27f4074c164421f49e3e

ShuffleWire:
Additonnaly, webots/docs/guide/tutorial-1-your-first-simulation-in-webots.md should be fixed the same way.

PR #7011, opened 2026-09-01
Use WorldInfo.basicTimeStep in new C and MATLAB controllers

yurekami:
Summary: use the world basic time step in the C and MATLAB controller templates copied by the new controller wizard; add a regression test. Testing: python -m unittest tests.sources.test_controller_templates

### DIFF
--- a/tests/sources/test_controller_templates.py
+++ b/tests/sources/test_controller_templates.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python
+
+# Copyright 1996-2024 Cyberbotics Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


tests/sources/test_controller_templates.py
CoolSpy3, line 1: Not sure it makes sense to test this. In general, I think tests should test behavior agnostic of the implementation details. Just grepping for a call to `wb_robot_get_basic_time_step()` only ensures that someone doesn't straight-up delete that code (which should be caught in manual review anyways). Someone could still (say) replace the value in the actual call to `wb_robot_step()` without failing the test, making it a misleading metric.

A proper test for this behavior would plug the template controllers into a sample world and check that they execute at the correct rate, which imo is more trouble than it's worth for this simple change. I'd just delete it outright.

CoolSpy3 (changes requested):
Awesome! Personally, I don't think the test is necessary (see comment below). If you want, you can also patch `docs/guide/tutorial-1-your-first-simulation-in-webots.md` so that we can close #3168, but we could also just save that for a later PR if you want.

PR #7012, opened 2026-09-01
Rescan embedded display textures after live scene-tree edits

### LINKED ISSUE #6964: `WbDisplay` Doesn't Detect Embedded Displays Until World is Reloaded
**Describe the Bug**
Currently, `WbDisplay::findImageTextures()` is only called during prefinalization, so updates to its structure are not propagated to the simulation. Thus, if someone sets up a display to work in embedded mode, those changes are not applied until the world is reloaded. This is in contrast to most other Webots nodes which have changes that are applied immediately.

While working on this, it might also be nice to support shapes/groups in all the `children` slots rather than just the first one. That should be a simple fix as it just involves changing the top-level `if` to a loop.

**Steps to Reproduce**
1. Open the `display.wbt` sample world
2. Open `MyBot > EMOTICONS Display`
3. Clear the `appearance` field of its first child
4. Follow [the documentation instructions](https://cyberbotics.com/doc/reference/display#description) to add back either an `Appearance` or a `PBRAppearance` with the correct structure
5. Run the simulation
6. Notice that the emoticons are not displayed

**Expected behavior**
We should listen for changes to the children and automatically rerun/incrementally update `findImageTextures` when required. Unfortunately, because that involves checking for both recursion on Groups AND changes in Shape.appearance AND THEN appearance.\<textureNode\> AND THEN \<textureNode\>.url, I'm having trouble thinking of a way to do it that doesn't get messy fast. (Maybe why it was written this way in the first place.)

yurekami:
## Summary

- rescan embedded image textures when display descendants change at runtime
- track nested `Group` / `Pose` children and appearance changes with stale-safe Qt connections
- scan every top-level child instead of only the first
- extend the display API regression to remove and reinsert a nested texture-bearing shape without reloading the world

## Why

`WbDisplay` only discovered embedded displays during prefinalization. Live scene-tree edits therefore left new or replaced embedded textures undiscovered until the world reloaded, as reported in #6964.

## Verification

- MinGW controller syntax check
- recursive signal/lifetime review against `WbGroup` ownership
- `git diff --check`

The full Webots world test could not run locally and remains CI-bound.

Fixes #6964

### DIFF
--- a/src/webots/nodes/WbDisplay.cpp
+++ b/src/webots/nodes/WbDisplay.cpp
@@ -26,6 +26,7 @@
 #include "WbRobot.hpp"
 #include "WbShape.hpp"
 #include "WbSimulationState.hpp"
+#include "WbSFNode.hpp"
 #include "WbStandardPaths.hpp"
 #include "WbWrenOpenGlContext.hpp"
 #include "WbWrenRenderingContext.hpp"
@@ -94,6 +95,7 @@ void WbDisplay::init() {
   mRequestImages = false;
   mAttachedCamera = NULL;
   mNeedToSetExternalTextures = false;
+  mIsUpdatingImageTextures = false;
 
   mDisplayFont = new WbDisplayFont();
   QString error = mDisplayFont->error();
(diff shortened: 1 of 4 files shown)


yurekami:
Closing this campaign's later Webots submission because another distinct Webots contribution from the same account (#7011) was opened minutes earlier. Keeping one active contribution per project as intended; the branch remains available if this issue needs a future follow-up.

PR #7014, opened 2026-09-10
Add support for Ubuntu 26.04

### LINKED ISSUE #6837: Missing `libsndio7.0` dependency
**Describe the Bug**
I'm installing Webots 2025a on Debian 13, and I've found that libsndio7.0 is missing from the dependencies. I'd to install it manually.

**Steps to Reproduce**
1. Get Webots 2025a .deb from the release page
2. Install it on a fresh Debian 13 (apt install ....deb)
3. Installation work correctly
4. Start webots, it complains about a missing library.
5. Install libsndio7.0 (apt install libsndio7.0)
6. Start webots, everything work fine.

**Expected behavior**
I wish Webots would pull that library as well during install.
I guess that on Ubuntu, this library is somewhat already installed.

Having this dependency added would ensure unofficial support OOTB for the Debian platform, and I would argue that it's a **very** minor thing to add.
For information, I've been running Webots : 
- Personally and professionally for 5 yes,
- Using 4 different machines, 
- Using Debian 11, 12, 13
- Using many version from 2021b to 2025a
- Using a various combination of CPU / GPU, even a 12 years old iGPU. 
- Using manually build from source and the deb packages
I would argue that it make a good coverage of the issues would might meet when running more officially on Debian, and I think that could enlarge the user base.
I think that Webots could almost claim ((un)official) Debian compatibility, but it's at least at least fix the small bit that make it work OOTB, imo

**System**
 - Operating System: Debian 13
 - Webots 2025a

omichel:
### Overview

This PR aims at introducing support for Ubuntu-26.04 to build Webots. `make` currently fails from a clean checkout on Ubuntu 26.04. This PR makes the Linux build work there and adds Ubuntu 26.04 to the CI.

### Details

**PEP 668.** `qt_linux_installer.sh` ran `pip install aqtinstall` and `linux_web_viewer_dependencies.sh` ran `pip3 install pyclibrary` against the system Python. Ubuntu 26.04 ships Python 3.14 and marks the system installation as externally managed, so both abort with `error: externally-managed-environment` and the build stops at the very first dependency. Each now installs into its own virtual environment under `dependencies/` (`aqt-venv`, `pyclibrary-venv`), added to `.gitignore`, with `python3-venv` added to the package lists.

Since `src/wren/Makefile` probed for pyclibrary with `pip list | grep` and then ran the parser via the bare system `python3`, a virtual environment would have been invisible to it. It now resolves the venv interpreter first and falls back to the system one, so macOS and Windows are unaffected, and probes by importing the module rather than grepping `pip list`.

**Emscripten pin.** `linux_web_viewer_dependencies.sh` used to install `emsdk latest`, so builds were not reproducible. It now pins `EMSDK_VERSION=6.0.9`, the version this PR was tested with. Both `emsdk install` and `emsdk activate` are idempotent, so rerunning the script is safe. The script is anchored on its own location rather than the current working directory, and only appends the activation line to `~/.bashrc` when it is not already there.

**Package renames on 26.04.** `libzip4` → `libzip5`, `libfreetype6-dev` → `libfreetype-dev`, and the gcrypt flavor of libssh is gone, leaving the OpenSSL-flavored `libssh-dev`. `execstack` is no longer packaged; nothing in the build invokes it, so it is simply dropped on 26.04. The packages that differ between releases moved into the per-version branches; every package the scripts install was checked to resolve on 26.04.

**Runtime dependency on libsndio.** The `libopenal.so.1` bundled in the tarball and Debian package is the system library of the 22.04 build machine, which links `libsndio.so.7`. Up to 24.04 that library came in implicitly through ffmpeg, but ffmpeg 8 on 26.04 no longer depends on it, so `webots-bin` from the 22.04 tarball fails to start on 26.04. `linux_runtime_dependencies.sh` now installs `libsndio7.0` explicitly and the Debian `Depends` lists it.

**Packaging.** `linux_distro.py` gets a 26.04 library list (`libzip.so.5`, `libx264.so.165`, ...) and falls back from `libssh-gcrypt.so.4` to `libssh.so.4` when the former is absent. An empty `exist_in_projects_linux_26.04.txt` is added, matching the 22.04 and 24.04 files.

**clang-format.** `dependencies/Makefile.linux` only matched 22.04/24.04, so 26.04 silently fell back to the legacy `.clang-format-6`. Added 26.04 to that filter.

**wrenjs link.** `resources/Makefile.include` linked the wren object files with `emcc`. Emscripten infers C++ from source file extensions, which works when compiling but not when linking `.o` files, so recent releases link against the C runtime and fail with hundreds of `undefined symbol: operator new`, `std::__2::...` errors. Emscripten's own diagnostic recommends `em++`, which is what it now uses. This is an Emscripten-version issue rather than an Ubuntu one — it reproduces with any `emsdk install latest` (6.0.9 here).

**`webots_controller.c`.** Ubuntu 26.04 ships glibc 2.43 and GCC 15, where `strchr` in C has a const-correct prototype: given a `const char *` it returns a `const char *`. `replace_char` took a `const char *` and wrote through the `strchr` result, which now fails with `-Werror=discarded-qualifiers`. The function is rewritten as a plain loop over a `char *`, which is what it always mutated. No cppcheck suppression is needed.

**cppcheck.** Ubuntu 26.04 ships cppcheck 2.19, which already reports several hundred new findings on `master` (mostly `nullPointerOutOfMemory` in `src/controller/c`), so `tests/sources/test_cppcheck.py` fails locally on 26.04 regardless of this PR. CI pins cppcheck 2.14.2 on 22.04 and is unaffected. Not addressed here.

**CI and docs.** `ubuntu-26.04` added to the `webots-build`, `build`, `test-suite` and `delete-artifacts` matrices of both Linux workflows, with Java 21 to match the `openjdk-21-jdk` installed by `linux_optional_compilation_dependencies.sh`. `actions/setup-python` publishes no Python 3.9 build for 26.04, so the `test-suite` job now selects its Python version per OS through the matrix, like Java: 3.9 on 22.04/24.04 and 3.11 on 26.04, matching the macOS and Windows test suites. The develop workflow's corresponding step is bumped from `setup-python@v2` to `v4`. The system requirements page and the R2025b changelog mention 26.04.

**Test suite hang.** The nightly Linux test-suite jobs on `master` have been hanging until GitHub's 6-hour limit since September 6 (24.04 every night, 22.04 intermittently), and the first 26.04 run here did the same. The log shows the suite reaching "Test suite complete" after 40 minutes and then never exiting: some test groups leave their Webots instance running, `Command.runTest` in `tests/command.py` never assigned `self.mainProcess` so its 10-minute timeout could not terminate anything, and its non-daemon reader threads then blocked the interpreter at exit. The runner now tracks the process, starts it in its own session and kills the whole process group on timeout (escalating to SIGKILL), and uses daemon reader threads; the Linux test-suite jobs get `timeout-minutes: 90` as a backstop. This turns a 6-hour hang into a 10-minute failure that names the group. Why Webots stops exiting at group end since September 6 is not addressed here.

### Testing

On Ubuntu 26.04.1 (Python 3.14, Emscripten 6.0.9, OpenJDK 21):

- `make` from a clean checkout completes, exit status 0, `bin/webots-bin` links with no missing shared libraries.
- `make` with `EMSDK` set builds `wrenjs.js`/`.wasm`/`.data`.
- `make` with `JAVA_HOME` set builds `Controller.jar` and `libJavaController.so`.
- Every package name and library SONAME in the 26.04 branches was checked against a 26.04.1 installation.
- Both workflow files parse as valid YAML with the expected matrices; the `build` job passes on all three Ubuntu runners.

### Points worth reviewing

- **`libssh-gcrypt-dev` → `libssh-dev` on 26.04 is a crypto-backend swap, not a rename.** The gcrypt flavor is no longer available. `linux_optional_compilation_dependencies.sh` already installed `libssh-dev` on every version (the two conflict), so developers with the optional dependencies have been building against the OpenSSL flavor all along.
- **The 22.04/24.04 package lists were restructured**, moving `libfreetype6-dev`, `libssh-gcrypt-dev` and `execstack` into per-version branches. Behavior should be identical, which CI confirms for the build.
- **The `ubuntu-26.04` runner image is still flagged as preview** by `actions/runner-images`.
- **cppcheck 2.19 on 26.04 fails on `master`** (see above); a follow-up is needed before the sources test can be run locally on 26.04.
- **`.clang-format-9` contains `AllowAllConstructorInitializersOnNextLine`**, deprecated in clang-format 14 in favor of `PackConstructorInitializers`. CI pins clang-format 14.0.0 and is unaffected, but 26.04's default is clang-format 21, so local formatting runs may need a follow-up. Not addressed here.


scripts/install/linux_runtime_dependencies.sh
```diff
@@ -24,18 +24,20 @@ fi
...
+    apt install --yes g++ make libavcodec-extra libglu1-mesa libegl1 libsndio7.0 \
         libxkbcommon-x11-dev libxcb-keysyms1 libxcb-image0 libxcb-icccm4 libxcb-randr0 \
         libxcb-render-util0 libxcb-xinerama0 libxcomposite-dev libxtst6 libnss3 libxcb-cursor0
 
     if [[ -z "$DISPLAY" ]]; then
         apt install --yes xvfb
     fi
 
-    if [[ $VERSION_ID == "22.04" || $VERSION_ID == "24.04" ]]; then
+    if [[ $VERSION_ID == "22.04" || $VERSION_ID == "24.04" || $VERSION_ID == "26.04" ]]; then
```
CoolSpy3: This if statement now applies to all supported versions. We can just remove this if statement and add `ffmpeg` to the normal dependencies.

scripts/install/linux_optional_compilation_dependencies.sh
```diff
@@ -27,14 +27,17 @@ install_ubuntu_optional_compilation_packages() {
...
-    apt install --yes curl python3.7-dev python3.8-dev python3.9-dev python3.10-dev dirmngr execstack libxerces-c-dev libfox-1.6-dev libgdal-dev libproj-dev libgl2ps-dev libssh-dev
+    apt install --yes curl python3.7-dev python3.8-dev python3.9-dev python3.10-dev dirmngr libxerces-c-dev libfox-1.6-dev libgdal-dev libproj-dev libgl2ps-dev libssh-dev
 
     if [[ $VERSION_ID == "22.04" ]]; then
-        apt install --yes openjdk-18-jdk
+        apt install --yes execstack openjdk-18-jdk
     elif [[ $VERSION_ID == "24.04" ]]; then
+        apt install --yes execstack openjdk-21-jdk
+    elif [[ $VERSION_ID == "26.04" ]]; then
+        # 26.04 no longer packages execstack, which is not used by the build anyway
```
CoolSpy3: If we're not using `execstack`, could we just delete it rather than keeping it as a dependency?

scripts/install/linux_compilation_dependencies.sh
```diff
@@ -24,12 +24,16 @@ fi
...
-    apt install --yes git cmake swig libglu1-mesa-dev libglib2.0-dev libfreeimage3 libfreetype6-dev libxml2-dev libboost-dev libssh-gcrypt-dev libzip-dev libreadline-dev pbzip2 wget zip unzip python3 python3-pip libopenal-dev
+    apt install --yes git cmake swig libglu1-mesa-dev libglib2.0-dev libfreeimage3 libxml2-dev libboost-dev libzip-dev libreadline-dev pbzip2 wget zip unzip python3 python3-pip python3-venv libopenal-dev
 
     if [[ $VERSION_ID == "22.04" || $VERSION_ID == "24.04" ]]; then
-        apt install --yes libzip4 openssl
+        apt install --yes libzip4 libfreetype6-dev libssh-gcrypt-dev openssl
+    elif [[ $VERSION_ID == "26.04" ]]; then
+        # 26.04 renamed libzip4 to libzip5 and libfreetype6-dev to libfreetype-dev,
+        # and dropped the gcrypt flavor of libssh in favor of the OpenSSL one
+        apt install --yes libzip5 libfreetype-dev libssh-dev openssl
```
CoolSpy3: I'm wondering if it's possible to unify some of these dependencies. From what I can see, on Ubuntu 22.04, [`libfreetype-dev`](https://packages.ubuntu.com/jammy/libfreetype-dev) and [`libfreetype6-dev`](https://packages.ubuntu.com/jammy/libfreetype6-dev) both map to FreeType 2.11.1. [I can't find `libfreetype6-dev` for Ubuntu 24.04](https://packages.ubuntu.com/search?keywords=libfreetype&searchon=names&suite=all&section=all), but there is [`libfreetype-dev`](https://packages.ubuntu.com/noble/libfreetype-dev).

Given that `libssh-gcrypt-dev` is just a different backend, and [`libssh-dev` exists across all supported versions](https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=libssh-dev&searchon=names), could we just swap the backend on all versions?

And, of course, `openssl` is explicitly the same across this statement.

resources/Makefile.include
```diff
@@ -550,7 +550,7 @@ endif
 
 ifdef WREN
 $(WREN_TARGET): $(OBJECTS_EMCC)
-	$(SILENT) emcc -O3 -s WASM=1 -s EXIT_RUNTIME=1 -s USE_WEBGL2=1 -s FULL_ES3=1 -s ALLOW_MEMORY_GROWTH=1 --preload-file ../../resources/wren/shaders@../../resources/wren/shaders $(OBJECTS_EMCC) -s EXPORTED_FUNCTIONS='[$(shell cat functions_to_export.txt)]' -s 'EXPORTED_RUNTIME_METHODS=["ccall", "getValue"]' -o wrenjs.js
+	$(SILENT) em++ -O3 -s WASM=1 -s EXIT_RUNTIME=1 -s USE_WEBGL2=1 -s FULL_ES3=1 -s ALLOW_MEMORY_GROWTH=1 --preload-file ../../resources/wren/shaders@../../resources/wren/shaders $(OBJECTS_EMCC) -s EXPORTED_FUNCTIONS='[$(shell cat functions_to_export.txt)]' -s 'EXPORTED_RUNTIME_METHODS=["ccall", "getValue"]' -o wrenjs.js
```
CoolSpy3, line 553: From the introducing commit:
> emcc infers C++ from the source
file extensions, so linking object files falls back to C and leaves the
C++ standard library symbols undefined with recent Emscripten releases.

Could you clarify this? Given that wren is written in C++, I agree that `em++` probably makes more sense here, but I wasn't aware of any problems with the standard library. (Just trying to understand the changes.)

scripts/install/linux_runtime_dependencies.sh
```diff
@@ -24,18 +24,20 @@ fi
 install_ubuntu_runtime_packages() {
     alias apt='apt --option="APT::Acquire::Retries=3"'
     apt update
-    apt install --yes g++ make libavcodec-extra libglu1-mesa libegl1 \
+    # libsndio7.0 is needed by the libopenal.so.1 bundled in the tarball, which is built on Ubuntu 22.04;
+    # up to Ubuntu 24.04 it was pulled in by ffmpeg, but ffmpeg 8 on Ubuntu 26.04 no longer depends on it
```
CoolSpy3: I don't think we need this long comment for one dependency.

CoolSpy3: also, need install in client?

dependencies/Makefile.linux
```diff
@@ -26,15 +26,15 @@ cleanse: $(PACKAGES_CLEAN)
 
 ../.clang-format:
 	@echo "# copying clang-format configuration file"
-ifeq ($(filter $(UBUNTU_VERSION), 22.04 24.04), $(UBUNTU_VERSION))
+ifeq ($(filter $(UBUNTU_VERSION), 22.04 24.04 26.04), $(UBUNTU_VERSION))
```
CoolSpy3: At this point, this check applies to all supported Ubuntu versions. Could we just drop clang-format-6 entirely?

scripts/packaging/linux_distro.py
```diff
@@ -207,10 +214,20 @@ def create_tarball_bundle(self):
         if distro.version() == '24.04':
             usr_lib_x68_64 += self.USR_LIB_X68_64_24_04
             usr_lib_x68_64.append('libraw.so.23')
+        if distro.version() == '26.04':
+            usr_lib_x68_64 += self.USR_LIB_X68_64_26_04
+            usr_lib_x68_64.append('libraw.so.23')
```
CoolSpy3: Is there a good reason for all of these append calls? Could we just add the relevant libraries to the relevant USR_LIB arrays?

scripts/packaging/linux_distro.py
```diff
@@ -207,10 +214,20 @@ def create_tarball_bundle(self):
...
         for lib in usr_lib_x68_64:
-            shutil.copy(os.path.join(system_lib_path, lib), package_webots_lib)
+            candidates = lib if isinstance(lib, tuple) else (lib,)
+            for candidate in candidates:
+                lib_path = os.path.join(system_lib_path, candidate)
+                if os.path.exists(lib_path):
+                    shutil.copy(lib_path, package_webots_lib)
+                    break
+            else:
+                shutil.copy(os.path.join(system_lib_path, candidates[0]), package_webots_lib)
```
CoolSpy3: Unnecessary resolution system. A versioning system for this already exists.

CoolSpy3:
Just a heads-up: I'm working on this (probably a little-over halfway though), but it might be another day or two before I can get a full review in. (Worst case, end of the weekend.)

scripts/install/qt_linux_installer.sh
```diff
@@ -4,18 +4,33 @@
...
+# aqtinstall is installed in a local virtual environment: recent Linux distributions mark
+# the system Python installation as externally managed (PEP 668) and refuse a plain pip install.
+AQT_VENV="$WEBOTS_HOME/dependencies/aqt-venv"
+if [ ! -x "$AQT_VENV/bin/python3" ] || ! "$AQT_VENV/bin/python3" -m pip --version > /dev/null 2>&1; then
+  rm -rf "$AQT_VENV"
+  python3 -m venv "$AQT_VENV"
+fi
+if ! "$AQT_VENV/bin/python3" -m aqt version > /dev/null 2>&1; then
+  "$AQT_VENV/bin/python3" -m pip install --no-input aqtinstall
+fi
```
CoolSpy3: Not necessarily wrong, but this kind of check isn't really idiomatic in these scripts. The user is expected to run the installer script once during repo setup (and maybe when we change it), so they generally just run the install scripts rather than checking if they're already installed (except for some long-running operations like file downloads, which this is not).

scripts/install/qt_linux_installer.sh
```diff
@@ -4,18 +4,33 @@
...
+
+QT_INSTALLATION_PATH="$QT_OUTPUT_DIR/${QT_VERSION}/gcc_64"
 QT_INSTALLATION_BIN_PATH=${QT_INSTALLATION_PATH}/bin
-QT_INSTALLATION_LIBEXEC_PATH=${QT_INSTALLATION_PATH}/libexec
 QT_INSTALLATION_LIB_PATH=${QT_INSTALLATION_PATH}/lib
 QT_INSTALLATION_INCLUDE_PATH=${QT_INSTALLATION_PATH}/include
+if [ ! -x "${QT_INSTALLATION_BIN_PATH}/qmake" ] || [ ! -d "${QT_INSTALLATION_INCLUDE_PATH}/QtWebSockets" ] || \
+   ! compgen -G "${QT_INSTALLATION_LIB_PATH}/libQt6WebSockets.so*" > /dev/null; then
+  "$AQT_VENV/bin/python3" -m aqt install-qt --outputdir "$QT_OUTPUT_DIR" linux desktop ${QT_VERSION} gcc_64 -m qtwebsockets
+fi
```
CoolSpy3: See my previous comments on non-idiomatic checks.

This script is already gated behind make's dependency checks, so I don't think gating the download doesn't make a ton of sense (especially with a condition as unruly as this). I'd also move it back up to where it was before (or at least add some line breaks around it. As-is, a quick scan makes it look like another variable definition.

src/wren/Makefile
```diff
@@ -40,19 +40,19 @@ endif
...
-	)
+# pyclibrary is installed in a local virtual environment by linux_web_viewer_dependencies.sh,
+# but it may also be installed system-wide, e.g. on macOS and Windows
+PYTHON = $(WEBOTS_DEPENDENCY_PATH)/pyclibrary-venv/bin/python3
+ifeq ($(wildcard $(PYTHON)),)
+PYTHON = python3
 endif
+
+# Check if pyclibrary is present
+PYC = $(shell "$(PYTHON)" -c "import pyclibrary; print('pyclibrary')" 2>/dev/null | tr -d '\r\n')
```
CoolSpy3: Bit of a nit-pick, but I prefer the old command here. `list > grep > format > check for grepped string` just reads nicer to me than `init python > import library > print random string > format > check for random string` (which will only have been printed if we didn't crash the interpreter).

scripts/install/linux_web_viewer_dependencies.sh
```diff
@@ -43,19 +46,49 @@ case "$OS" in
...
+# Activate emsdk in interactive shells (same approach as the macOS and Windows bash profiles)
+EMSDK_SOURCE_LINE='source "'$EMSDK_PATH'/emsdk_env.sh" >/dev/null 2>&1'
+BASHRC="$TARGET_HOME/.bashrc"
+if [ ! -f "$BASHRC" ]; then
+  touch "$BASHRC"
+  chown "$TARGET_USER":"$TARGET_GROUP" "$BASHRC"
+fi
+if ! grep -qxF "$EMSDK_SOURCE_LINE" "$BASHRC"; then
+  echo "$EMSDK_SOURCE_LINE" >> "$BASHRC"
+fi
```
CoolSpy3: Yes, this preserves (and improves upon) existing behavior, but in my mind, scripts shouldn't modify `.bashrc` without explicit consent. I recommend we instead move this to `bashrc.linux` as a commented out line with comment.

This would also allow us to remove the dependency on `TARGET_HOME`.

scripts/install/linux_web_viewer_dependencies.sh
```diff
@@ -43,19 +46,49 @@ case "$OS" in
...
+# pyclibrary is installed in a local virtual environment: recent Linux distributions mark
+# the system Python installation as externally managed (PEP 668) and refuse a plain pip install.
+PYCLIBRARY_VENV="$WEBOTS_HOME/dependencies/pyclibrary-venv"
+if [ ! -x "$PYCLIBRARY_VENV/bin/python3" ] || ! "$PYCLIBRARY_VENV/bin/python3" -m pip --version > /dev/null 2>&1; then
+  rm -rf "$PYCLIBRARY_VENV"
+  python3 -m venv "$PYCLIBRARY_VENV"
+fi
+if ! "$PYCLIBRARY_VENV/bin/python3" -c "import pyclibrary" > /dev/null 2>&1; then
+  "$PYCLIBRARY_VENV/bin/python3" -m pip install --no-input pyclibrary
+fi
```
CoolSpy3: Not necessarily wrong, but this kind of check isn't really idiomatic in these scripts. The user is expected to run the installer script once during repo setup (and maybe when we change it), so they generally just run the install scripts rather than checking if they're already installed (except for some long-running operations like file downloads, which this is not).

tests/command.py
```diff
@@ -156,11 +188,16 @@ def mainTarget():
             outFile = open(self.outFileName, "w")
             errFile = open(self.errFileName, "w")
 
-            p = subprocess.Popen(self.cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            # start the command in its own session so that the timeout handling can terminate its whole process tree
+            # (on POSIX; the option is ignored on Windows)
+            p = subprocess.Popen(self.cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, start_new_session=True)
+            self.mainProcess = p
+            self.mainProcessOwnsSession = hasattr(os, 'killpg')
```
CoolSpy3: This variable's essentially a no-op. The only time it's used, it's and-ed with the same condition. This assignment doesn't make sense to me anyways, so I'd just remove the variable completely.

scripts/packaging/linux_distro.py
CoolSpy3: There's no reason for this new resolution system. A versioning system for this already exists with the os-specific arrays above.

tests/command.py
```diff
@@ -54,12 +56,42 @@ def isRunning(self):
         """Detect if the command is running."""
         return self.mainProcess is not None
 
+    def signalMainProcess(self, force):
+        """Send SIGTERM (or SIGKILL if force) to the main process, and to its whole process group if it owns one."""
+        process = self.mainProcess
+        if not process:
+            return
+        if self.mainProcessOwnsSession and hasattr(os, 'killpg'):
```
CoolSpy3: I'd add a comment here. Something to the effect of `Posix systems only`. Otherwise, checking for the existence of a method in a standard library doesn't seem to me like a normal practice.

CoolSpy3 (changes requested):
Sweet! Looking forward to Ubuntu 26 support! Mostly cleanup notes. I'm not sure why the test suite is failing. Could be related to the changes here; could be something else. In any case, it looks like multiple test failures some of which are in tests I'd consider to be fairly reliable, so I think we should investigate that before merging this.

resources/Makefile.include
omichel: Emscripten 6.0.6 disabled DEFAULT_TO_CXX, so emcc no longer links libc++ and em++ is now required when linking C++ programs, as with clang and gcc. I should have mentioned this in the commit message.

omichel:
> I'm not sure why the test suite is failing. Could be related to the changes here; could be something else. In any case, it looks like multiple test failures some of which are in tests I'd consider to be fairly reliable, so I think we should investigate that before merging this.

It seems the test suite became flaky at some point. It is not specific to this PR. It randomly fails on all supported Ubuntu versions. We should probably address this in a separate PR.

omichel:
Thank you for the review! I believe I addressed all the issues you reported in separate commits. Let's see if it builds from the CI now.

scripts/install/bashrc.linux
```diff
@@ -5,6 +5,8 @@
 
 # export JAVA_HOME=/usr/lib/jvm/java-16-openjdk-amd64  # Defines the path to Java home.
 
+# source $HOME/webots/dependencies/emsdk/emsdk_env.sh >/dev/null 2>&1 # Activate PATH and other environment variables needed by Emscripten
```
CoolSpy3: Should be
```suggestion
# source $WEBOTS_HOME/dependencies/emsdk/emsdk_env.sh >/dev/null 2>&1 # Activate PATH and other environment variables needed by Emscripten
```

I think we'll have to move the `WEBOTS_HOME` definition up to make this work.

scripts/install/linux_compilation_dependencies.sh
```diff
@@ -24,15 +24,9 @@ fi
...
-    if [[ $VERSION_ID == "22.04" || $VERSION_ID == "24.04" ]]; then
-        apt install --yes libzip4 libfreetype6-dev libssh-gcrypt-dev openssl
-    elif [[ $VERSION_ID == "26.04" ]]; then
-        # 26.04 renamed libzip4 to libzip5 and libfreetype6-dev to libfreetype-dev,
-        # and dropped the gcrypt flavor of libssh in favor of the OpenSSL one
-        apt install --yes libzip5 libfreetype-dev libssh-dev openssl
-    else
+    apt install --yes git cmake swig libglu1-mesa-dev libglib2.0-dev libfreeimage3 libxml2-dev libboost-dev libzip-dev libfreetype-dev libssh-dev openssl libreadline-dev pbzip2 wget zip unzip python3 python3-pip python3-venv libopenal-dev
+
+    if [[ $VERSION_ID != "22.04" && $VERSION_ID != "24.04" && $VERSION_ID != "26.04" ]]; then
```
CoolSpy3, line 29: Just a note: this no longer installs `libzip4`/`libzip5`. The relevant libraries are still pulled in as dependencies of `libzip-dev`, so I'm okay with this, but I wanted to add a record of this change somewhere in this PR.

resources/Makefile.include
CoolSpy3: :+1: Sounds good! Thanks for the clarification!

CoolSpy3 (changes requested):
Thank you! LGTM! Just one small note for the bashrc file.

I agree that it looks like the tests are failing the same way on other branches, so we can investigate separately.

CoolSpy3 (dismissed):
Great! Thank you!

CoolSpy3 (approved):

PR #7022, opened 2026-09-16
Add OneRobotics A1 robot arm

uniquemelody:
**Description**

This adds a self-contained Webots model of the OneRobotics A1 independent right arm.

- 7-DoF `Robot` PROTO with motors, position sensors, physical properties, simplified box collisions, and an end-effector slot
- a Python motion demo and sample world
- model documentation, a rendered 128 x 128 icon, publication metadata, and a changelog entry

The model is derived from [`a1_r.urdf`](https://github.com/katazen/onerobot_h1/blob/ecf530911284ba0e559f7a24dc222fd8e60d31ed/source/h1_reach/h1_reach/assets/urdf/A1_2026/a1_r.urdf) at source commit `ecf530911284ba0e559f7a24dc222fd8e60d31ed`. The kinematics and physical properties retain the published values, except that visual meshes are optimized derivatives of the eight source STL files, collision meshes are approximated with boxes, and the published hardware motor limits are applied. Each OBJ records the source STL SHA-256. The robot assets are © 2026 OneRobotics and licensed under CC BY 4.0.

**Review update**

Following the maintainer's mesh/material review, including the close-up report of holes in the first optimized OBJ revision:

- Rebuilt all eight OBJ meshes from the pinned original STLs. The earlier camera-sampled, per-triangle deletion caused visible holes and has been discarded. The replacement conservatively simplifies complete connected parts only when edge topology remains sound, then unions overlapping closed parts to remove internal intersection surfaces.
- Reduced the visual meshes from 143,638 to 82,450 triangles (42.6% fewer); the largest individual mesh has 18,857 triangles, below the shadow-casting warning threshold observed in Webots R2025a.
- Replaced faceted STL visuals with OBJ meshes containing crease-aware smooth vertex normals and UV coordinates.
- Enabled shadows on every link and switched to Webots' existing `Plastic` PBR appearance.
- Corrected the sample world's default viewpoint so the robot is visible on opening.
- Renamed the PROTO documentation page to lowercase and removed unused `DEF` names.

**Related Issues**

No linked issue.

**Documentation**

[OneRoboticsA1 model documentation](https://github.com/uniquemelody/webots/blob/feature-onerobotics-a1/projects/robots/onerobotics/a1/protos/docs/oneroboticsa1.md)

**Screenshot**

OneRobotics A1

**Validation**

- Coordinate-welded OBJ topology checks show no increase in boundary edges relative to any original STL. A denser 48-view, 160 × 160 ray comparison against each original STL found a maximum bounding-box difference of 0.111 mm, maximum silhouette miss rate of 0.386%, and maximum 95th-percentile visible-surface depth difference of 0.481 mm.
- Applicable Webots source checks pass: project structure, PROTO fields, icons, licenses, line endings, PEP 8, header versions, and textures.
- Webots R2025a runtime smoke test loads the final geometry and `Plastic` material, runs the demo controller for 450 steps, and renders the whole arm plus close-up front/underside base and front/end flange views without the previously reported holes or mesh/shadow warnings. The isolated test invocation only reports R2025b forward-compatibility and launcher warnings; an R2025b runtime was not available locally.

**Additional context**

The published independent-arm revision does not include a gripper. The source repository also marks Link7 mass and inertia as placeholder data; both limitations are documented in the model page.


projects/robots/onerobotics/a1/protos/docs/OneRoboticsA1.md
```diff
@@ -0,0 +1,73 @@
...
+The robot assets are © 2026 OneRobotics and licensed under the
+[Creative Commons Attribution 4.0 International license](https://creativecommons.org/licenses/by/4.0/).
+
+## OneRoboticsA1 PROTO
+
+Derived from [`Robot`](https://cyberbotics.com/doc/reference/robot).
+
+```proto
+OneRoboticsA1 {
+  SFVec3f     translation       0 0 0
```
omichel:
```suggestion
  SFVec3f    translation       0 0 0
```

projects/robots/onerobotics/a1/protos/OneRoboticsA1.proto
```diff
@@ -0,0 +1,435 @@
...
+# license url: https://creativecommons.org/licenses/by/4.0/
+# documentation url: https://webots.cloud/run?url=https://github.com/cyberbotics/webots/blob/released/projects/robots/onerobotics/a1/protos/OneRoboticsA1.proto
+# keywords: robot/arm
+# template language: javascript
+# Model of the fixed-base OneRobotics A1 right arm.
+# Source: https://github.com/katazen/onerobot_h1/tree/ecf530911284ba0e559f7a24dc222fd8e60d31ed/source/h1_reach/h1_reach/assets/urdf/A1_2026
+# Changes: converted from URDF, approximated collision meshes with boxes, renamed mesh files to lowercase, and applied the published hardware motor limits.
+
+PROTO OneRoboticsA1 [
+  field SFVec3f     translation        0 0 0        # Is `Pose.translation`.
```
omichel:
```suggestion
  field SFVec3f    translation        0 0 0        # Is `Pose.translation`.
```

projects/robots/onerobotics/a1/protos/OneRoboticsA1.proto
```diff
@@ -0,0 +1,435 @@
...
+  field SFString   name               "OneRobotics A1" # Is `Solid.name`.
+  field SFString   controller         "<generic>"  # Is `Robot.controller`.
+  field MFString   controllerArgs     []           # Is `Robot.controllerArgs`.
+  field SFString   window             "<generic>"  # Is `Robot.window`.
+  field SFString   customData         ""           # Is `Robot.customData`.
+  field SFBool     supervisor         FALSE        # Is `Robot.supervisor`.
+  field SFBool     synchronization    TRUE         # Is `Robot.synchronization`.
+  field SFBool     selfCollision      FALSE        # Is `Robot.selfCollision`.
+  field MFNode      endEffectorSlot   []           # Extends the robot at the Link7 flange frame.
+  field SFBool      staticBase        TRUE         # Defines whether the base is fixed to the static environment.
```
omichel:
```suggestion
  field MFNode     endEffectorSlot    []           # Extends the robot at the Link7 flange frame.
  field SFBool     staticBase         TRUE         # Defines whether the base is fixed to the static environment.
```

omichel (changes requested):

omichel (changes requested):
Thank you for this contribution.

However, this robot model doesn't meet the quality requirements to be included in the official Webots repository. You may want to include it as is in the [community repository](https://github.com/cyberbotics/community-projects/tree/master/robots) or modify this model as follows:

1. The meshes contain many parts which are not visible (inner parts of the links). You should delete these part to reduce the number of triangles.
2. A consequence of reducing the number of triangles will be that shadows will be cast for every link of the robot (which is not the case right now), in addition to other benefits (loads faster, reduce download time, etc.)
3. The meshes are not smoothed. We can see many small faces on them. You should convert them to .obj files (instead of .stl) and add normals so that faces are smooth.
4. The appearance of the robot should rely on a existing PBR material proto, such as Plastic (or another one if you believe it is best suited).

All the mesh work could be done manually with Blender: deleting non-visible inner parts, generating normals where it makes sense and converting to .obj format.

uniquemelody:
Thank you for the review, @omichel. I pushed 6c52f7d addressing the four points: removed occluded interior faces and reduced visual triangles by 63.3%, converted the eight visual STLs to smooth-shaded OBJ meshes with UVs, enabled shadows on every link, and switched the materials to the existing Plastic PROTO. I also corrected the sample-world viewpoint and updated the model documentation/PR description to identify these meshes as derivatives. The applicable source checks and an R2025a runtime smoke test pass. Could you please take another look when convenient?

omichel (changes requested):
Thank you for making the requested changes.
The shadow is now cast for every part of the robot arm which is great.
However, the meshes are now broken, see this screenshot:
Image
Moreover, there are still many inner triangles which are not visible from the outside and should be deleted.

omichel:
The base link also has the same problem:
image
Before (left), After (right)

omichel:
Doc filename breaks convention. [protos/docs/OneRoboticsA1.md](https://github.com/cyberbotics/webots/blob/master/protos/docs/OneRoboticsA1.md) should be oneroboticsa1.md — all 90 existing PROTO doc pages in the repo are lowercase (TurtleBot3Burger.proto → turtlebot3burger.md, Irb4600-40.proto → irb4600-40.md; 0 of 90 contain an uppercase letter). This likely breaks the documentation url lookup declared in the PROTO header.

omichel:
Unused DEFs: BASE_APPEARANCE and WHITE_APPEARANCE are never USEd (only LINK_APPEARANCE is), likewise DEF JOINT*_A1_R / DEF LINK*.

uniquemelody:
@omichel Thank you for the close-up screenshots. You were right: the previous camera-sampled deletion of individual faces created visible holes. I have replaced all eight meshes in c958e9c24 using the pinned source STLs, topology-checked simplification of complete parts, and solid unions for overlapping closed parts; the per-face deletion has been removed. I checked the base from the front and underside and the flange from the front and end in Webots, with no visible slits or missing surfaces. The meshes now have 82,450 triangles (42.6% below source); the largest is 18,857, and Webots reports no shadow/mesh warnings. I also renamed the documentation page to oneroboticsa1.md and removed the unused DEFs. The PR description now records the corrected method and validation. Please let me know if you still see a particular internal region that should be removed.

omichel (changes requested):
Thank you for the fixes.
The missing triangles are now back which is fine.
The doc file name and unused DEF are fixed.
However, there are still too many non-visible triangles in the inner parts of the robot that we should delete to make the model lighter.
I recommend using Blender (or similar 3D editing software) to edit the obj files.

uniquemelody:
Updated the branch onto the current `develop` baseline (`264c9457f`) so the PR is no longer behind. The current mesh revision remains `c958e9c24`: it restores watertight visual topology after the earlier broken-face attempt, uses smooth OBJ/PBR meshes, and passes the applicable source checks. The remaining requested work is specifically removal of additional non-visible inner geometry; I have not applied another unverified face-deletion pass that could recreate holes. Please re-review the current geometry and point out any particular inner regions that should be removed.

omichel:
If you open link1.obj in Blender, you will see many inner triangles that are not visible in the simulation:
image
You should delete them manually in Blender to reduce the size of the mesh.
Other OBJ files have the same problem.

